PR #9, opened 2015-08-25
Switched the IDs in the documentation to Strings.

SiliconMeeple:
This reflects what's expected and returned through the resources.


mikemey:
Test data in tests `ChargeRequestResourceITest` and `ChargeDaoITest` also requires the change of the gateway account id to a String.

mikemey:
:+1:

PR #13, opened 2015-09-01
PP-12 Confirm a card payment (capture)

chaphilippe:
Allow to capture a payment for a charge that has been authorised successfully beforehand.
If the card has not bee authorised successfully beforehand, an error is returned. 

for more info see https://payments-platform.atlassian.net/browse/PP-12


chaphilippe:
rrun

chaphilippe:
rrun

chaphilippe:
rrun

chaphilippe:
rrun

ijmad:
rrun

README.md
```diff
@@ -240,3 +241,36 @@ Content-Type: application/json
...
+
+### POST /v1/frontend/charges/{chargeId}/capture
+
+This endpoint proceeds to the capture of the card for the specified charge. The charge needs to have been previously authorised for this call to succeed.
+
+#### Request example
+
+```
+POST /v1/frontend/charges/1/capture
+Content-Type: application/json
```
zadsas, line 254: We need to make sure that a "Content-Length: 0" header is also included in the request because some proxies complain about it.

src/main/java/uk/gov/pay/connector/resources/ChargeCaptureResource.java
```diff
@@ -0,0 +1,62 @@
...
+import static uk.gov.pay.connector.model.ChargeStatus.CAPTURED;
+import static uk.gov.pay.connector.util.ResponseUtil.badResponse;
+import static uk.gov.pay.connector.util.ResponseUtil.responseWithChargeNotFound;
+
+@Path("/")
+public class ChargeCaptureResource {
+
+    private ChargeDao chargeDao;
+    private final Logger logger = LoggerFactory.getLogger(ChargeCaptureResource.class);
+    private static final String STATUS_KEY = "status";
```
zadsas: Should this constant live within ChargeStatus.java? That way we keep both the charge statuses and the key to query them in the same place.

src/main/java/uk/gov/pay/connector/resources/ChargeCaptureResource.java
```diff
@@ -0,0 +1,62 @@
...
+        if (!maybeCharge.isPresent()) {
+            return responseWithChargeNotFound(logger, chargeId);
+        }
+
+        Map<String, Object> charge = maybeCharge.get();
+        if (!isAuthorized(charge)) {
+            return responseWithChargeStatusIncorrect((String) charge.get(STATUS_KEY));
+        }
+
+        chargeDao.updateStatus(chargeId, CAPTURED);
```
zadsas, line 51: Why does the jira ticket mention a CAPTURE SUBMITTED status?

chaphilippe, line 51: CAPTURE SUBMITTED is to acknowledge the reception of the request -  could be needed to cope with a problem with external providers which would prevent us to update to CAPTURED somehow. There is no point on setting CAPTURE SUBMITTED here as we always return sequencially the status CAPTURED for the scope of this story (and this status would override the CAPTURE SUBMITTED). If we were to keep track of a list of statuses, it would make sense, but for the moment we keep only the last one.

README.md
chaphilippe, line 254: ok, good to know - here, it is an internal call and the fact that the body is empty will probably change in the near future (card tokenisation,...) when implementing next stories

zadsas, line 254: We might eventually have internal proxies. In any case, if the body is likely to be non-empty in the near future then let's not bother about it.

src/main/java/uk/gov/pay/connector/resources/ChargeCaptureResource.java
zadsas, line 51: :+1:

PR #22, opened 2015-09-09
PP-111: Citizen user sees summary of their service and charge before payment steps

kakumara:
Upon successful charge creation, connector provides a link specific to the chargeId to the card capture view in Frontend. PublicApi will use this in order to redirect the caller to the relevant link
see https://payments-platform.atlassian.net/browse/PP-111


ijmad:
:+1: pending other reviews

PR #29, opened 2015-09-22
PP-120 Add a confirm page

SiliconMeeple:
Changed to the connector to enable the confirmation page to work.
This is pretty much just an extra endpoint in the charge links.

(with @zadsas )


src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -68,4 +69,9 @@ private URI cardAuthUrlFor(UriInfo uriInfo, String chargeId) {
         return uriInfo.getBaseUriBuilder()
                 .path(CARD_AUTH_FRONTEND_PATH).build(chargeId);
     }
+
+    private URI cardCaptureUrlFor(UriInfo uriInfo, String chargeId) {
+        return uriInfo.getBaseUriBuilder()
+                .path(CARD_CAPTURE_FRONTEND_PATH).build(chargeId);
+    }
```
rahul-vinayak: Can potentially make these three methods into one by extracting a method Param "path" and removing duplicacy
chargeLocationFor()
cardAuthUrlFor()
cardCaptureUrlFor()

kakumara: Addressed

PR #27, opened 2015-09-21
PP-121 Authorise and Capture Worldpay payments (@kakumara, @chaphilippe)

chaphilippe:
- Ability to create accounts for sandbox or worldpay (for the moment, the credentials are read from config and setup in pay-script - dockercompose).
- authorise and capture are implemented for both sandbox (unmodified flow) and worldpay.
- worldpay responses are validated and errors are handled (no specific error handling - unique error response for all).
- worldpay and sandbox are creating gatewayTransactionId that are sent to the external payment providers as reference, and stored in the charge table.
- address is now mandatory in the card payload (address line 1,  postcode, city and  country are minimum fields)
- new status introduced: CAPTURED (after capture request succeeded)

NB: Currently we are unsure how to get the charge status from worldpay (synchronously), and we will have a call with them to clarify this. Doing a second request to get the order status, if needed, will be handled in a subsequent story.

PR #30, opened 2015-09-24
[PP-127]: update external status mapping.

mikemey:
the mapping for the internal status `AUTHORIZATION REJECTED` was missing, is now mapped to the external status `FAILED`.


kakumara:
Couple of minor points.
- It'd be nice to have unit tests covering some of these new statuses. However I can see the ChargesResource do not have unit tests and testing this via INT tests will be tricky. I'm not too hung up on this as this just to finish off this story. Perhaps we should do this when we touch this resource in the next story
- I am changing all Statues for AUTHORIZATION to AUTHORISATION (Uk style) as part of world pay story. Can you change (just the new one) to AUTHORISATION

mikemey:
@First point: I'd agree, it's out of scope for this story, let's add is as a tech debt?

@Second point: This PR didn't add the status 'AUTHORIZATION_REJECTED' - this change can be done when rebasing https://github.com/alphagov/pay-connector/pull/27 after this one is merged - plus the end-to-end tests needs this update as well, because they specify that status.

PR #31, opened 2015-09-28
PP-135 @chaphilippe @acceleanu Add smartpay to list of available providers

chaphilippe:
The scope of this story is only to be able to create an account for smartpay in the connector.

see https://payments-platform.atlassian.net/browse/PP-135


mikemey:
:+1:

PR #32, opened 2015-09-30
PP-141 authorise via smartpay

acceleanu:
Service can authorise a payment via smartpay
- handle smartpay success and failure responses (return same error message and code than other providers)
- reading credentials from the config (temporarily) : changes have been made to pay-scripts (docker-compose file)
  more info see: https://payments-platform.atlassian.net/secure/RapidBoard.jspa?rapidView=3&projectKey=PP&view=detail&selectedIssue=PP-141

@chaphilippe @acceleanu


src/main/java/uk/gov/pay/connector/service/PaymentProviders.java
```diff
@@ -1,30 +1,46 @@
...
     private final Map<String, PaymentProvider> providers;
 
-    public PaymentProviders(WorldpayConfig config) {
+    public PaymentProviders(ConnectorConfiguration config) {
         providers = new HashMap<String, PaymentProvider>() {{
-            put(WORLDPAY_PROVIDER, new WorldpayPaymentProvider(config));
+            put(WORLDPAY_PROVIDER, createWorldpayProvider(config.getWorldpayConfig()));
+            put(SMARTPAY_PROVIDER, createSmartPayProvider(config.getSmartpayConfig()));
             put(DEFAULT_PROVIDER, new SandboxPaymentProvider());
         }};
```
zadsas: What about avoiding this static initializer block?
We could just loop over the VALID_PROVIDERS List defined in PaymentProviderValidator.
Something like...

PaymentProviderValidator.VALID_PROVIDERS.stream().forEach( provider -> providers.put(provider, paymentProviderFactory(provider)));

?

src/main/java/uk/gov/pay/connector/app/SmartpayConfig.java
```diff
@@ -0,0 +1,29 @@
...
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+public class SmartpayConfig extends Configuration {
+
+    @Valid
+    @NotNull
+    private String url;
+
+    private String username;
```
zadsas: Do we have username and password non validated on purpose?

src/main/java/uk/gov/pay/connector/service/PaymentProviders.java
zadsas: Another option could be to use EnumMaps?
http://javarevisited.blogspot.co.uk/2012/09/what-is-enummap-in-java-example-tutorial.html

acceleanu: - 1. does not really work as each of the providers is constructed in a different way
- 2. what would be the benefit?

zadsas: - 1. It would work if you create a Payment Provider Factory. The factory would also remove the duplicated code of the factory methods.
- 2. That initialisation block is an overhead for the class loader as it creates more classes than needed. Also it can be a potential memory leak as there is a hidden pointer to the enclosing class.

src/main/java/uk/gov/pay/connector/service/OrderSubmitRequestBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+
+
+import uk.gov.pay.connector.model.domain.Card;
+import uk.gov.pay.connector.util.templates.TemplateStringBuilder;
+
+import java.util.Map;
+
+import static com.google.common.collect.Maps.newHashMap;
+
+public class OrderSubmitRequestBuilder {
```
zadsas: Do we need to have this class and "WorldpayOrderCaptureRequestBuilder"? Looks like they could be bundled into one single class.

src/main/java/uk/gov/pay/connector/app/SmartpayConfig.java
```diff
@@ -0,0 +1,29 @@
...
+        return url;
+    }
+
+    public String getUsername() {
+        return username;
+    }
+
+    public String getPassword() {
+        return password;
+    }
```
rahul-vinayak: I see SmartpayConfig and WorldpayConfig as having exactly same classes with same prop, can we do a common config extraction here. (DRY principle)

src/main/java/uk/gov/pay/connector/service/PaymentProviders.java
SiliconMeeple: On double-brace initialisation: http://blog.jooq.org/2014/12/08/dont-be-clever-the-double-curly-braces-anti-pattern/

src/main/java/uk/gov/pay/connector/app/SmartpayConfig.java
acceleanu: Fixed.
Created GatewayCredentialsConfig to use in both cases.

src/main/java/uk/gov/pay/connector/service/OrderSubmitRequestBuilder.java
acceleanu: They are different and they serve different purposes.
OrderSubmitRequestBuilder is for authorise whereas 
WorldpayOrderCaptureRequestBuilder is for capture.

src/main/java/uk/gov/pay/connector/service/PaymentProviders.java
acceleanu: While I agree that in general it is a bad practice to instantiate a Hashmap like that 
and this implementation can be better, 
I would challenge anyone to show how this code, as it is here, causes a memory leak.

src/main/java/uk/gov/pay/connector/service/OrderSubmitRequestBuilder.java
zadsas: :+1:

src/main/java/uk/gov/pay/connector/service/PaymentProviders.java
SiliconMeeple: We shouldn't need bugs (memory leaks) as a justification to perform useful refactoring.

src/main/java/uk/gov/pay/connector/app/SmartpayConfig.java
acceleanu: We are going to get these configs from another place anyway when we implement account management and have different checks anyway.

PR #33, opened 2015-10-01
PP-150 cancel worldpay charge

mikemey:
Cancellation of Worldpay charge. See JIRA: https://payments-platform.atlassian.net/browse/PP-150

Since asynchronous update of `CANCELLED` state isn't in the scope of this story, the state is updated when the response from Worldpay was OK (which is always the case, as long as the XML is well-formatted).

Added dummy implementation in Sandbox provider - also always returns an OK response.
Added `not yet implemented` exception in SmartPay provider.

~~wait for merge of #32 - this PR is based on #32 - ready after a potential rebase~~


mikemey:
rebased from master, ready to review

kakumara:
Unit tests vs integration tests
- Cannot see any unit tests for cancel logic in CardService.java 
- It seems all scenarios tested by IT tests.
- Understand we used these bit mixed at the moment, but my preference is more unit tests (covering pretty much every line) and less (but enough) INT tests. Happy to discuss.

mikemey:
As discussed in tech-huddle: in this scenario it's ok to not have a unit test.

PR #35, opened 2015-10-02
Refactoring

zadsas:
- There is no point to put the token into the map in order to remove it
  immediately after


src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -10,12 +10,21 @@
 import uk.gov.pay.connector.model.api.ExternalChargeStatus;
 import uk.gov.pay.connector.util.ResponseUtil;
 
-import javax.ws.rs.*;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
```
acceleanu: we have agreed to all have the same intellij settings - defaults

zadsas: :+1:

PR #37, opened 2015-10-06
Add live dev support

ijmad:
With the changes in PP-102 on scripts & endtoend, this allows running of the connector outside of docker, talking to other services still within docker. Technique could be extended across the other services to allow running arbitrary ones outside of docker for 'live' dev purposes with local changes and hot reloading (when supported).

Looking more for feedback rather than a merge at this point - I'd want to tidy this up and make an equivalent PR against other services, particularly frontend.

Please don't merge this until https://github.com/alphagov/pay-scripts/pull/68 & https://github.com/alphagov/pay-endtoend/pull/26 have been looked at.


ijmad:
rebuild

PR #38, opened 2015-10-07
PP-148 make docker CMD wait for postgres to be available before running

surfraz:
The endtoend tests are randomly failing due to postgres not being available
at the time DB migrations start


bazbremner:
As there are multiple issues preventing successful builds, I'm not surprised to see the build fail. However, this change fixes one of the issues @surfraz has identified, getting us closer to a working system (yay!).

Merging.

PR #43, opened 2015-10-13
Pp 172

acceleanu:
Partial implementation for https://payments-platform.atlassian.net/browse/PP-172.
It will handle
- invalid response codes
- dns lookup failures 
- malformed responses from gateway


src/main/java/uk/gov/pay/connector/model/CaptureResponse.java
```diff
@@ -33,4 +36,13 @@ public static CaptureResponse captureFailureResponse(Logger logger, String error
 
         return new CaptureResponse(false, baseGatewayError(errorMessage));
     }
+
+    public static CaptureResponse errorCaptureResponse(Logger logger, Response response) {
+        logger.error(format("Error code received from provider: response status = %s.", response.getStatus()));
+        return new CaptureResponse(false, unexpectedStatusCodeFromGateway("Error processing capture request"));
```
kakumara: (minor)
- perhaps chaining method calls instead of 'new' ing up

src/main/java/uk/gov/pay/connector/model/GatewayErrorType.java
```diff
@@ -1,7 +1,10 @@
 package uk.gov.pay.connector.model;
 
 public enum GatewayErrorType {
-
     ChargeNotFound,
-    GenericGatewayError;
+    GenericGatewayError,
+    UnexpectedStatusCodeFromGateway,
+    MalformedResponseReceivedFromGateway,
+    UnknownHostException,
```
kakumara: 'UnknownHostException' sounds too generic compared to the other error types declared here. Perhaps more specific, so that its inline with the rest. (e.g. UnableToLocateGateway, etc)

src/main/java/uk/gov/pay/connector/model/AuthorisationResponse.java
```diff
@@ -42,6 +42,10 @@ public static AuthorisationResponse authorisationFailureResponse(Logger logger,
         return new AuthorisationResponse(false, baseGatewayError("This transaction was declined."), AUTHORISATION_REJECTED, transactionId);
     }
 
+    public static AuthorisationResponse authorisationFailureResponse(GatewayError gatewayError) {
+        return new AuthorisationResponse(false, gatewayError, null, null);
```
kakumara: may be nicer to have an overloaded constructor, so that no need to pass false, null, null which looks a bit creepy

src/main/java/uk/gov/pay/connector/service/CardService.java
```diff
@@ -83,20 +83,23 @@ public CardService(GatewayAccountDao accountDao, ChargeDao chargeDao, PaymentPro
...
-
         String transactionId = String.valueOf(charge.get(GATEWAY_TRANSACTION_ID_KEY));
 
         CaptureRequest request = captureRequest(transactionId, String.valueOf(charge.get(AMOUNT_KEY)));
-        CaptureResponse response = paymentProviderFor(charge).capture(request);
+        CaptureResponse response = paymentProviderFor(charge)
+                .capture(request);
+
+        ChargeStatus newStatus =
+                response.isSuccessful() ?
```
kakumara: As per this logic the real 'CAPTURE UNKNOWN' (errors specified in story) and possibly KNOWN errors (like auth failures or gateway properly responds with an error stating not captured) are clubbed together in on global 'CAPTURE UNKOWN' status? 
Shouldn't that be a different status like CAPTURE FAILED?

acceleanu: I would say that is out of scope. The right moment to implement this would be when we decide how to use the CAPTURE_UNKNOWN.

src/main/java/uk/gov/pay/connector/model/GatewayErrorType.java
acceleanu: fixed

src/main/java/uk/gov/pay/connector/model/AuthorisationResponse.java
acceleanu: fixed

src/main/java/uk/gov/pay/connector/model/CaptureResponse.java
acceleanu: fine as it is - they're all factory methods

PR #44, opened 2015-10-13
PP-163

rahul-vinayak:
PP-163 connector changes to introduce new charge state “ENTERING_CARD_DETAILS” when user lands on card details page. Introduced a new PUT endpoint for charge state


acceleanu:
rebuild

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -37,20 +44,47 @@ public ChargesFrontendResource(ChargeDao chargeDao) {
...
+                .orElseGet(() -> responseWithChargeNotFound(logger, chargeId));
+    }
 
-                    return ok(responseData).build();
-                })
+    @PUT
+    @Path(PUT_CHARGE_STATUS_FRONTEND_PATH)
+    @Produces(APPLICATION_JSON)
+    public Response updateChargeStatus(@PathParam("chargeId") String chargeId, Map newStatusMap) {
+        Optional<Map<String, Object>> maybeCharge = chargeDao.findById(chargeId);
```
acceleanu: - input validation missing
- please have a look at ChargesApiResource.createNewCharge

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -37,20 +44,47 @@ public ChargesFrontendResource(ChargeDao chargeDao) {
...
+        return maybeCharge
+                .map(charge -> updateStatus(charge, ChargeStatus.chargeStatusFrom(newStatusMap.get("new_status").toString())))
                 .orElseGet(() -> responseWithChargeNotFound(logger, chargeId));
     }
 
+    private Response updateStatus(Map<String, Object> charge, ChargeStatus chargeStatus) {
+        String charge_id = charge.get("charge_id").toString();
+        if (!hasStatus(charge, CREATED, ENTERING_CARD_DETAILS)) {
+            return ResponseUtil.badRequestResponse(logger, "charge with id: " + charge_id + " cant be updated to the state: " + ENTERING_CARD_DETAILS.getValue());
```
kakumara: Shouldn't this be a Conflict(409) than a 400 ?

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
kakumara: (Suggestion) Would it be better to do in one update? UPDATE 'something' where STATUS is in (something) and check if the the updated no of rows not 0 ? (Just to avoid the transaction problem updated by another thread between the calls)

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -37,20 +44,47 @@ public ChargesFrontendResource(ChargeDao chargeDao) {
...
 
-                    return ok(responseData).build();
-                })
+    @PUT
+    @Path(PUT_CHARGE_STATUS_FRONTEND_PATH)
+    @Produces(APPLICATION_JSON)
+    public Response updateChargeStatus(@PathParam("chargeId") String chargeId, Map newStatusMap) {
+        Optional<Map<String, Object>> maybeCharge = chargeDao.findById(chargeId);
+        return maybeCharge
+                .map(charge -> updateStatus(charge, ChargeStatus.chargeStatusFrom(newStatusMap.get("new_status").toString())))
```
kakumara: (may be more of a question)
How good is to expose actual status value to be taken from outside directly in to the database? isn't it better to get an expectation from the outside and let the connector decide the actual status value to be stored in the database?

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
rahul-vinayak: @acceleanu added missing input validation, and a new test to cover that.

rahul-vinayak: @kakumara yes it would be a good idea, however as we discussed we would be looking at making all the updates and inserts transactional at some point this would be addressed like all the other places it is needed.

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
rahul-vinayak: @kakumara good point but looking up the explanation of where should you do 409 conflict is mostly where an update is done on a set of data that is outdated, or old, and my understanding is that it would mean the client should get the latest data and then try to send the same modification request again which would solve it (409 is mostly used with version mismatch to signify outdated data), however here no matter if they fetch latest data and send the same modification request it would not succeed, so a Bad request 400 seems more appropriate.

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
rahul-vinayak: Yes there was a discussion about it whether Frontend should have knowledge of the charge statuses of what should it ask connector to set it to, and we decided Frontend is another app hosted by us so is going to be a safe source, for now atleast. Also the string value coming from frontend will be updated only if it matches a lookup in the charge status Enum in connector, so it wont be just any string.

rahul-vinayak: Although it reminds me I should write a test which tries to set any arbitary string as charge status and see how it fails

rahul-vinayak:
rebuild

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
SiliconMeeple: "How can you have a problem with us delivering this new window with a giant crack in it? All your other windows are broken!"

If it's reasonably straightforward to fix, could we not commit known bugs please?

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -48,28 +55,65 @@ public ChargesFrontendResource(ChargeDao chargeDao, GatewayAccountDao accountDao
...
         return reduce(validateGatewayAccountReference(gatewayAccountId)
                 .bimap(handleError, listTransactions(gatewayAccountId)));
+    }
+
+    private Response updateStatus(Map<String, Object> charge, ChargeStatus chargeStatus) {
+        String charge_id = charge.get("charge_id").toString();
+        if (!hasStatus(charge, CREATED, ENTERING_CARD_DETAILS)) {
+            return ResponseUtil.badRequestResponse(logger, "charge with id: " + charge_id + " cant be updated to the state: " + ENTERING_CARD_DETAILS.getValue());
+        }
+        chargeDao.updateStatus(charge_id, chargeStatus);
```
ijmad: There is an issue with transactionality here. We get the charge on Line 75, and then update it on Line 99 without checking if someone has changed it in the meantime, concurrently with us. One way to fix this would be to wrap the entire interaction in a DB transaction, another would be to ensure the updateStatus checks the existing status before changing it with a WHERE clause, and checks the number of affected rows afterwards.

rahul-vinayak: Applied the second way of updating now, an update method that takes in a new status and a list of present valid statuses on which the update is applied upon, it returns the number of rows updated.

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
rahul-vinayak: implemented as discussed in the broader discussion of applying updates where.......

PR #46, opened 2015-10-19
pp-147 Run test apps on random unused ports.

SiliconMeeple:
This fixes the spurious test failures caused if something (for example, the
pay-connect tests running in parallel on jenkins) happens to already
have port 9000/9001 at the start of the tests.


acceleanu:
Well spotted - good fix that we may need to apply to the other components. 
Discussed this approach vs the more general one of isolating each the services via docker-compose when running their own integration tests (that would require a different test setup because of the way we start Postgres right now in the integration tests).

PR #48, opened 2015-10-21
pp-147 Receive notifications for the sandbox.

SiliconMeeple:
In order for the end to end tests to treat the sandbox payment flow in
the same way as worldpay and smartpay, it's necessary for the sandbox to
receive notifications.

This also lets us test other behaviour through the sandbox.

Additionally, we added a worldpay capture contract test - more contract
test to follow in a following PR.

(with @kakumara)


README.md
```diff
@@ -618,3 +618,30 @@ Content-Type: text/plain
 
 [accepted]
 ```
+
```
acceleanu: line 571, 598 should be 'described'

src/main/java/uk/gov/pay/connector/resources/NotificationResource.java
```diff
@@ -45,8 +48,16 @@ public Response handleNotification(@PathParam("provider") String provider, Strin
             return Response.status(BAD_GATEWAY).build();
         }
 
-        response.forEachStatusUpdate(chargeDao::updateStatusWithGatewayInfo);
```
acceleanu: line 41 @Path("v1/api/notifications/{provider}")
- since there is no authentication should consider making the endpoint unpredictable
- add explanation as to why the endpoint is not secured and how we intend to mitigate this

acceleanu:
rebuild

src/main/java/uk/gov/pay/connector/resources/NotificationResource.java
SiliconMeeple: Why does it matter if the connector endpoint is predictable? It won't be exposed to the internet.

I've added some more documentation on securing the notification endpoint, and where we expect that to happen. Not quite sure the connector readme's the best place for that, but I can't think of anywhere better.

README.md
SiliconMeeple: Ineded.

src/test/java/uk/gov/pay/connector/unit/sandbox/SandboxPaymentProviderTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class SandboxPaymentProviderTest {
+
+    private SandboxPaymentProvider sandboxClient = new SandboxPaymentProvider(new ObjectMapper());
+
+    @Test
+    public void shouldSuccessfullyParseANotification() throws Exception {
+        StatusUpdates statusUpdates = sandboxClient.newStatusFromNotification(toJson(ImmutableMap.of(
+                "transaction_id", "transaction",
+                "status", "AUTHORISATION_REJECTED")));
```
acceleanu: we should reuse the ChargeStatus enum; this also applies to all other notification tests

SiliconMeeple: This would fail the test, as the toString method returns `AUTHORISATION REJECTED`, but the enum value is actually called `AUTHORISATION_REJECTED` and the sandbox reconstructs the status by calling ChargeStatus.valueOf.

README.md
```diff
@@ -568,7 +568,7 @@ HTTP/1.1 204 No Content
 
 ### POST /v1/api/notifications/worldpay
```
zadsas: The link in the summary to here is broken due a missing 's'

README.md
```diff
@@ -576,8 +576,8 @@ This endpoint handles a notification from worldpays Order Notification mechanism
 POST /v1/api/notifications/worldpay
```
zadsas: The link in the summary to here is broken due a missing 's'

SiliconMeeple: Oops.

SiliconMeeple:
rebuild

src/test/java/uk/gov/pay/connector/unit/sandbox/SandboxPaymentProviderTest.java
SiliconMeeple: ... On the other hand, it's just the sandbox, so why not change how it materializes those statuses? :)

PR #52, opened 2015-11-05
PP-323 Fix worldpay contract test.

acceleanu:
- Fixed test that relied on a transaction done in the past being still available

with @holyhaddock


src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
```diff
@@ -94,19 +70,13 @@ public void shouldBeAbleToSendCancelRequestForMerchant() throws Exception {
 
     @Test
     public void shouldBeAbleToSendOrderInquiryRequestWhenStatusNotificationComesIn() throws Exception {
```
zadsas: Should we make this name more meaningful? There is notification happening here.

acceleanu: Fixed

PR #55, opened 2015-11-11
PP-101 quicker feedback

ijmad:
These are the counterpart changes to https://github.com/alphagov/pay-scripts/pull/94


kakumara:
rebuild

PR #61, opened 2015-11-24
PP 307 - Add payment reference

acceleanu:
https://payments-platform.atlassian.net/browse/PP-307
- Added a payment reference field
- Tolerates the field being absent for the moment on the idea that once the publicapi is going to send id, we will do proper checks


acceleanu:
rebuild

PR #63, opened 2015-11-27
Pp 197 update account credentials

kakumara:
This is a connector part of the user story PP-197
The story in brief: A Government service administrator can view/update their credentials to payment provider through self service portal.
- credentials are stored as a JSON in database.
- connector provides a "frontend" api's to update credentials.
- at account creation, connector will store an empty "{}" credentials object.
- GET gateway account api will never return the password element of credentials.

NOTE: This PR does not include the use of these credentials in the actual gateway interactions. This is to be done later.


src/main/java/uk/gov/pay/connector/dao/GatewayAccountDao.java
```diff
@@ -35,13 +37,32 @@ public boolean idIsMissing(String gatewayAccountId) {
...
+                .map(new GatewayAccountMapper())
                 .first());
 
         return Optional.ofNullable(gatewayAccount);
     }
 
+    public void saveCredentials(String credentialsJsonString, String gatewayAccountId) {
+        jdbi.withHandle(handle -> handle
+                .createStatement("UPDATE gateway_accounts SET credentials = :credentials WHERE gateway_account_id = :id")
+                .bind("credentials", createPostgresCredentials(credentialsJsonString))
```
rahul-vinayak: So I believe the credentials together with the Password is getting saved in DB in plain text, which is visible, which I think should be avoided at all costs. We can either save password as a separate column and hash it, or store the json string after atleast hashing the password part. Or if there is a need to retrieve the password to pass on to provider we can think of encrypting it.

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -26,33 +39,44 @@
...
     public Response getGatewayAccount(@PathParam("accountId") String accountId) {
 
         logger.info("Getting gateway account for account id {}", accountId);
 
-        return gatewayDao.findById(accountId)
-                .map(account -> Response.ok().entity(account).build())
+        return gatewayDao
+                .findById(accountId)
+                .map(gatewayAccount -> {
+                    gatewayAccount.remove("credentials");
```
rahul-vinayak: I would prefer not getting the credentials out of DAO in the first place rather than a resource removing it later on, probably having another version of findById() which doesn't select the credentials, and one that does.

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -64,19 +88,69 @@ public Response createNewGatewayAccount(JsonNode node, @Context UriInfo uriInfo)
...
+    @GET
+    @Path(ACCOUNT_FRONTEND_RESOURCE)
+    @Produces(APPLICATION_JSON)
+    public Response getGatewayAccountWithCredentials(@PathParam("accountId") String gatewayAccountId) throws IOException {
+        return gatewayDao
+                .findById(gatewayAccountId)
+                .map(gatewayAccount ->
+                {
+                    Map<String, String> credentialsMap = (Map<String, String>) gatewayAccount.get("credentials");
+                    credentialsMap.remove("password");
```
rahul-vinayak: Same thing I would advise here as well, i.e. not reading the password in the first place in the DAO layer. Could be achieved by some filter logic in DAO or another orchestration layer, or if the password was a separate column..... comments??

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -64,19 +88,69 @@ public Response createNewGatewayAccount(JsonNode node, @Context UriInfo uriInfo)
...
+        return Response.ok().build();
+    }
+
+    private boolean gatewayAccountIdExists(String gatewayAccountId) {
+        if (!isNumber(gatewayAccountId) ||
+            !gatewayDao.findById(gatewayAccountId).isPresent()) {
+            return false;
+        }
+        return true;
+    }
```
rahul-vinayak: Could be simplified:
private boolean gatewayAccountIdExists(String gatewayAccountId) {
         return isNumber(gatewayAccountId) && gatewayDao.findById(gatewayAccountId).isPresent();
}

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -64,19 +88,69 @@ public Response createNewGatewayAccount(JsonNode node, @Context UriInfo uriInfo)
...
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    public Response updateGatewayCredentials(@PathParam("accountId") String gatewayAccountId, JsonNode credentialsPayload) {
+
+        if (!gatewayAccountIdExists(gatewayAccountId)) {
+            return notFoundResponse(logger, format("The gateway account id '%s' does not exist", gatewayAccountId));
+        }
+
+        String provider = (String) gatewayDao.findById(gatewayAccountId).get().get("payment_provider");
+
```
rahul-vinayak: there is findById() call been made twice and hence the same select query getting fired twice:
once in this method call !gatewayAccountIdExists() and then again few lines after, I would combine them in one, may be use:

gatewayDao.findById(gatewayAccountId).isPresent()

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -64,19 +88,69 @@ public Response createNewGatewayAccount(JsonNode node, @Context UriInfo uriInfo)
...
+        List<String> missingFieldsInRequestPayload = getMissingFieldsInRequestPayload(credentialsPayload, provider);
+        if (!missingFieldsInRequestPayload.isEmpty()) {
+            return badRequestResponse(logger, format("The following fields are missing: [%s]", on(",").join(missingFieldsInRequestPayload)));
+        }
```
rahul-vinayak: For the sake of lexical correctness:
a space after the comma `on(", ").join`

src/main/java/uk/gov/pay/connector/dao/GatewayAccountDao.java
kakumara: Yes. This was discussed in detail as part of the story kick-off. And also with the PCI/DSS assessor.
The decision was there is no need of this encryption in app level given that;
- Amazon RDS provides disk encryption by default which we use.
- All out inter process communications will be https and TLS enabled.
- We cannot use hashing, as we need them back in plain text during the actual communication.

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
kakumara: :+1:

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
kakumara: :+1: refactored it optimise the findById to be called just once.

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
kakumara: :+1:  removed this altogether as part of the refactoring

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
kakumara: If we never need the credentials, yes I agree perhaps we can suppress it in the DAO call itself.
But we will need this in the next story where we need to get this from DAO during AUTH, CAPTURE calls. 
Re: discussion, introduced 2 methods. one with creds and the other without creds.

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
kakumara: Same reason above

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -64,19 +85,67 @@ public Response createNewGatewayAccount(JsonNode node, @Context UriInfo uriInfo)
...
+                    return Response.ok(gatewayAccount).build();
+                })
+                .orElseGet(() -> notFoundResponse(logger, format("Account with id '%s' not found", gatewayAccountId)));
+    }
+
+    @PUT
+    @Path(ACCOUNT_FRONTEND_RESOURCE)
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    public Response updateGatewayCredentials(@PathParam("accountId") String gatewayAccountId, JsonNode credentialsPayload) {
```
rahul-vinayak: Only thing I see here and we need to be mindful of (if not already discussed while kick off) that you can over-write anybody's credentials by passing someone else's accountID, we should think on the lines of using a signature or token to verify the authenticity of the request, if not as part of this story then need to discussed its taken into account later on.

kakumara: This is only exposed to self-service. Not public API. So we do not get any account tokens to verify against.

rahul-vinayak: As discussed we should still think of building a trust between different services by way of using signatures/tokens so that connector can do a sanity check as well along with trusting the source, but that needs a larger discussion to cover most of our endpoints, so happy to merge for the time being.

PR #65, opened 2015-11-27
PP-101 Remove explicit postgres wait from default server run.

ijmad:
Instead, create a separate DropWizard command, dbwait, that uses settings from config.yaml that can be put before the migrations run. Change Dockerfile to reflect this.

The rationale for this is that the netcat, while clever, was only validating the TCP socket was accepting connections, not that Postgres was really available and answering queries. This ensures Java can actually connect which requires a fuller handshake.

Plus, having "java server" wait for Postgres wasn't useful if we need the migrations to run first, which weren't doing this wait.

Identikit PR to https://github.com/alphagov/pay-publicauth/pull/12


bazbremner:
As with https://github.com/alphagov/pay-publicauth/pull/12, making it obvious what that 30 means would be nice, but yay! Merging!

PR #68, opened 2015-12-01
Pp 101 quicker feedback final

ijmad:
P-101 quicker feedback revised instructions and scripts for using a devhelper and running locally.
Improve README.md & tidy Dockerfile of old things the new scripts don't use.


ijmad:
Needs to wait until scripts & jenkins get merged

PR #70, opened 2015-12-03
PP-101 quicker feedback revised instructions

ijmad:
PP-101 quicker feedback revised instructions and scripts for using a …


Dockerfile
```diff
@@ -5,7 +5,6 @@ ENV PORT 8080
 ENV ADMIN_PORT 8081
 ENV ENABLE_NEWRELIC no
 
-EXPOSE 8080 8081
```
bazbremner: Required - this is how production gets at the container.

ijmad: As discussed - EXPOSE doesn't map the port, it only indicates to Docker the app might listen on the port. Removing this because our ports are dynamic does not harm dev or production because they both pass in an explicit --publish (-p) to docker either directly or via docker-compose.

PauloPortugal:
:+1:  In terms of local setup this is working really well and was a major improvement.

Dockerfile
bazbremner: :+1:

PR #75, opened 2015-12-10
PP-197 Use credentials from gateway account

PauloPortugal:
This PR depends on [EndToEnd PR](https://github.com/alphagov/pay-endtoend/pull/65).

When interacting with a payment gateway, use the credentials related to the relevant gateway account, instead of a global credential.
- Use gateway account credentials during "AUTH", "CAPTURE", "CANCEL" and "GETSTATUS" flow.
- Refactoring of the notification logic.


src/main/java/uk/gov/pay/connector/service/PaymentProviders.java
```diff
@@ -8,10 +8,13 @@
 import uk.gov.pay.connector.service.worldpay.WorldpayPaymentProvider;
 
 import static java.lang.String.format;
-import static uk.gov.pay.connector.model.domain.GatewayAccount.gatewayAccountFor;
 import static uk.gov.pay.connector.resources.PaymentProviderValidator.*;
 import static uk.gov.pay.connector.service.GatewayClient.createGatewayClient;
 
+/**
+ * TODO: This class seems to be singleton all the way down to GatewayClient
```
rahul-vinayak: Are you sure about it being a singleton, I don't see what will prevent it from instantiated as many a times as you like.

PauloPortugal: Once PaymentProviders is instantiated it will always return the same providers(objects).
@kakumara is that correct?

kakumara: Yes. @rahul-vinayak If you look at carefully, you'll see PaymentProviders only instantiated at ConnectorApp.run() during startup. Once the 3 provider objects have generated at the time (with built in GatewayClient objects) the same ref is being returned.

src/main/java/uk/gov/pay/connector/service/CardService.java
```diff
@@ -100,9 +102,19 @@ private GatewayResponse captureFor(String chargeId, Map<String, Object> charge)
         return response;
     }
 
+    private Optional<GatewayAccount> findAccountByCharge(String chargeId) {
+        Optional<Map<String, Object>> charge = chargeDao.findById(chargeId);
+        if (!charge.isPresent()) {
+            String errorMessage = String.format("No charge exists for this charge id %s.", chargeId);
+            logger.error(errorMessage);
+            throw new IllegalStateException(errorMessage);
```
rahul-vinayak: Is there a better exception to be thrown here, IllegalState is used to signify a diff meaning, whereas in this case it could be a wrong input from the end point

src/main/java/uk/gov/pay/connector/service/CardService.java
```diff
@@ -125,13 +137,24 @@ private GatewayResponse cancelFor(String chargeId, Map<String, Object> charge) {
...
+        return new AuthorisationRequest(chargeId, card, amountValue, "This is the description", getValidAccountForCharge().apply(chargeId));
+    }
+
+    private Function<String, GatewayAccount> getValidAccountForCharge() {
+        return chargeId -> {
+            Optional<GatewayAccount> optionalServiceAccount = findAccountByCharge(chargeId);
+            if (!optionalServiceAccount.isPresent()) {
+                String errorMessage = String.format("No account exists for this charge %s.", chargeId);
+                logger.error(errorMessage);
+                throw new IllegalStateException(errorMessage);
```
rahul-vinayak: again a similar comment on usage of illegalState exception as above

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayNotificationResourceITest.java
```diff
@@ -65,28 +81,47 @@ public void shouldNotAddUnknownStatusToDatabase() throws Exception {
...
 
-        notifyConnector(transactionId)
+        notifyConnector(transactionId, "GARBAGE")
                 .statusCode(502);
 
         assertFrontendChargeStatusIs(chargeId, AUTHORISATION_SUCCESS.getValue());
     }
 
-    private ValidatableResponse notifyConnector(String transactionId) throws IOException {
+    private ValidatableResponse notifyConnector(String transactionId, String status) throws IOException {
```
rahul-vinayak: Consider using enums for passing the Status instead of String

src/test/java/uk/gov/pay/connector/unit/sandbox/SandboxPaymentProviderTest.java
```diff
@@ -19,11 +21,16 @@
...
     @Test
     public void shouldSuccessfullyParseANotification() throws Exception {
-        StatusUpdates statusUpdates = sandboxClient.newStatusFromNotification(toJson(ImmutableMap.of(
+        StatusUpdates statusUpdates = sandboxClient.handleNotification(toJson(ImmutableMap.of(
                 "transaction_id", "transaction",
-                "status", AUTHORISATION_REJECTED.getValue())));
+                "status", AUTHORISATION_REJECTED.getValue())),
+                x -> aServiceAccount(),
+                accountUpdater
+                );
```
rahul-vinayak: can simplify for sake of readability

src/main/java/uk/gov/pay/connector/service/CardService.java
kakumara: removed exception and returning an Optional.empty() which is safer in this case.

src/main/java/uk/gov/pay/connector/service/CardService.java
kakumara: Changed to RuntimeException

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayNotificationResourceITest.java
kakumara: This is a utility method for mocking notifications from Worlpay. And need to mock non existent stuff like "GARBAGE" too. So not seem to worth using a Enum here. But I have introduced a WorldpayPaymentStatus for Tests and refactored whatever the valid statues in to it.

src/test/java/uk/gov/pay/connector/unit/sandbox/SandboxPaymentProviderTest.java
kakumara: :+1:  refactored the inbound notification mocking

src/main/java/uk/gov/pay/connector/service/PaymentProviders.java
rahul-vinayak: Yes thats correct we are making just one instance of this class at the start up, but that still doesn't make this class Singleton as there is nothing that stops us creating another instance of it. So putting that in the comment can be misleading and cause confusions.

kakumara: :+1:  I've updated the comment to be more explicit about the concern.

PR #76, opened 2015-12-11
[REVIEW ONLY] PP-409 refactore get create charge apis

kakumara:
This needs to wait till (PP-197)[https://github.com/alphagov/pay-connector/pull/75] merged in and rebase with master.
- Copied both GET charge and CREATE charge api's under Accounts Resource.
- Kept the old API's for backwards compatibility. (To be removed)

PR #77, opened 2015-12-15
[REVIEW ONLY] PP 409 remove backward compatibility

kakumara:
This PR depends on this (previous)[https://github.com/alphagov/pay-connector/pull/76] PR to be merged first.  
- This PR completes this PP-409 story
- Removes the duplicate methods in ChargesApiResource kept for backward compatibility
- Removed all "OLD_xxxx" path references in ApiPaths.java so that we no longer have API paths with "old" styled.


PauloPortugal:
:+1:

PR #78, opened 2015-12-17
PP-197 Remove formerly used worldpay/smartpay username, password

ijmad:
These are now set through REST APIs (called by selfservice) and configured on a per-account basis


kakumara:
Looks ok :+1:

PR #79, opened 2015-12-21
Pp 415 fix specialchars in credentials

kakumara:
- Added couple of tests to prove our credentials API's are special character safe
- Minor refactoring to ServiceAccountMapper to GatewayAccountMapper (to be consistent)


PauloPortugal:
:+1:

PR #80, opened 2015-12-23
Pp 334 transaction history

kakumara:
Dependent PR -> https://github.com/alphagov/pay-endtoend/pull/72

 This PR includes the all the connector changes necessary to capture the payment state change history in the connector
- charges table structure remains as it is .(still holds the 'current state' of a given payment)
- new table 'charge_events' introduced, which adds a record for each state change.
- Uses observer pattern (ChargeEventListener.java) to emit events between the Daos.
- Events API double checks if a charge ID belongs the correct account before responding with transaction events.

NOTE: ChargeDao is kind of messy at the moment and not entirely happy with the current state of it. Needs a refactoring at some point


src/main/java/uk/gov/pay/connector/dao/EventDao.java
```diff
@@ -0,0 +1,39 @@
...
+    public EventDao(DBI jdbi) {
+        this.jdbi = jdbi;
+    }
+
+    public List<ChargeEvent> findEvents(Long accountId, Long chargeId) {
+        return jdbi.withHandle(handle ->
+                handle.createQuery("SELECT cs.id, cs.charge_id, cs.status, to_char(cs.updated,'DD/MM/YYYY HH24:MI:SS') as updated FROM charge_events AS cs WHERE cs.charge_id IN " +
+                        "(SELECT ch.charge_id FROM charges AS ch" +
+                        " WHERE ch.charge_id=:chargeId AND" +
+                        " ch.gateway_account_id=:accountId)")
```
PauloPortugal: Assuming that the charge_id is unique in the charges table, this select query is not needed.
Is my assumption correct?

src/main/resources/migrations.xml
```diff
@@ -79,4 +79,22 @@
...
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="charge_id" type="bigserial">
+                <constraints foreignKeyName="fk__charges_events" referencedTableName="charges"
+                             referencedColumnNames="charge_id" nullable="false"/>
+            </column>
+            <column name="status" type="text">
+                <constraints nullable="false"/>
+            </column>
+            <column name="updated" type="timestamp without timezone" defaultValueComputed="(now() at time zone 'utc')">
```
PauloPortugal: :+1:  for the defaultValueComputed

src/main/java/uk/gov/pay/connector/resources/ApiPaths.java
```diff
@@ -8,7 +8,7 @@
     String FRONTEND_RESOURCE = "/v1/frontend";
 
     String CHARGES_FRONTEND_PATH = "/v1/frontend/charges/";
```
PauloPortugal: I remember this refactor with Dan. Seems like everytime we pay a visit to the ApiPaths we end up changing.
Possibly we can delete CHARGES_FRONTEND_PATH and just use FRONTEND_RESOURCE + CHARGES_RESOURCE instead

src/test/java/uk/gov/pay/connector/it/resources/ChargesFrontendResourceITest.java
```diff
@@ -35,9 +35,7 @@
     private String returnUrl = "http://whatever.com";
     private long expectedAmount = 6234L;
 
-    private RestAssuredClient chargeCall = new RestAssuredClient(app, accountId, CHARGES_API_PATH);
-    private RestAssuredClient frontendChargesCall = new RestAssuredClient(app, accountId, CHARGES_FRONTEND_PATH);
-    private RestAssuredClient frontendChargeCall = new RestAssuredClient(app, accountId, GET_CHARGE_FRONTEND_PATH);
+    private RestAssuredClient connectorRestApi = new RestAssuredClient(app, accountId);
```
PauloPortugal: I am not sure we have gained much with this refactoring.
Again this is open for debate and I will not hold the ticket back, but I refactored it with Dan and even then it was difficult to reach a state that both were entirely happy. We are now shadowing the frontend endpoints by calling the variable connectorRestApi but having a test class called ChargesFrontendResourceITest.java.
Just some food for thought

src/main/java/uk/gov/pay/connector/dao/EventDao.java
PauloPortugal: Ignore the above comment. Spoke with @kakumara and this is just a double check to ensure that whoever wants to get more information about a charge ID needs to provide the  account_id :+1:

src/test/java/uk/gov/pay/connector/it/resources/ChargesFrontendResourceITest.java
PauloPortugal: Spoke with @kakumara about this.
Pair programming would have been useful in this case.

src/main/java/uk/gov/pay/connector/dao/EventDao.java
kakumara: yes, correct :+1:

src/test/java/uk/gov/pay/connector/it/resources/ChargesFrontendResourceITest.java
kakumara: My personal opinion is that; 
- providing once client object with separate methods for each endpoint is more readable + maintainable + concise code 
  compared with
- Having to declare client objects for each API path and (thereby exposing unusable methods) for usable for one operation.

I'm happy to debate over this; but agree having paired together could have made this more agreeable.

src/main/java/uk/gov/pay/connector/resources/ApiPaths.java
kakumara: make sense. Done :+1:

PR #84, opened 2016-01-08
PP-279 Search & Filter Transactions using Staff Pages

PauloPortugal:
- The staff should be able to:
  - search by a full or partial reference
  - filter by:
    - status
    - created date, by providing any of 'From' and 'To' dates
- This ticket is not backward compatible wiith E2E tests and depends on https://github.com/alphagov/pay-endtoend/pull/74.


PauloPortugal:
rebuild

davidillsley:
Why is it not backwards compatible? In particular, why because of only tests?

src/main/java/uk/gov/pay/connector/resources/ApiPaths.java
```diff
@@ -14,6 +14,7 @@
     String FRONTEND_AUTHORIZATION_RESOURCE = FRONTEND_RESOURCE + CHARGE_RESOURCE + "/cards";
     String FRONTEND_CAPTURE_RESOURCE = FRONTEND_RESOURCE + CHARGE_RESOURCE + "/capture";
 
+    String CHARGES_SEARCH_API_PATH = ACCOUNT_API_RESOURCE + "/search";
```
davidillsley: Why a new endpoint, and not have this at CHARGES_API_PATH ?

PauloPortugal:
@davidillsley _"Why is it not backwards compatible? In particular, why because of only tests?"_
The returned statuses have changed. 
The statuses we are returning are the high level ones, instead of the inner ones.
This only affects E2E because both public api and selfservice are OK with both statuses being returned, as we can mock the behaviour for their own unit tests. On the application they are also OK because both public-api and self-service handle them with on problems.

To make this change backwards compatible we would have to refactor to the SelfServiceTransactionSteps and the TransactionRowCheck to ensure we expect at least one of two values.
Before Christmas the team had a chat at a tech-retrospective about this, that if the backwards compatibility problem arises **only** because of the E2E tests, we could merge with master as long as the reviewer tests locally all the branches related to that store.

src/main/java/uk/gov/pay/connector/resources/ApiPaths.java
PauloPortugal: You mean a GET to the ACCOUNT_API_RESOURCE?

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -138,6 +143,26 @@ public Response getEvents(@PathParam("accountId") Long accountId, @PathParam("ch
         return responseBuilder.build();
     }
 
+    @GET
+    @Path(CHARGES_SEARCH_API_PATH)
+    @Produces(APPLICATION_JSON)
+    public Response getCharges(@PathParam("accountId") String accountId,
```
davidillsley: Should be documented in the README

src/main/java/uk/gov/pay/connector/util/ChargesSearch.java
```diff
@@ -0,0 +1,95 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+
+public class ChargesSearch {
+
+    public static List<Map<String, Object>> createQueryHandle(Handle handle, String query, String gatewayAccountId,
```
davidillsley: I'm a bit nervous about getting into a habit of doing string manipulation of queries because of the chance of a SQL injection vulnerability slipping in.

src/main/java/uk/gov/pay/connector/resources/ApiPaths.java
davidillsley: I meant GET at CHARGES_API_PATH

PauloPortugal: :blush: Yes that makes sense!
I was reading CHARGE_API_PATH. Should have gone to specsavers

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
PauloPortugal: Yes I will document it as part of the API endpoint path change.

src/main/java/uk/gov/pay/connector/util/ChargesSearch.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    private static String constructSearchTransactionsQuery(String reference, ExternalChargeStatus status, String fromDate, String toDate) {
+        StringBuffer subQuery = new StringBuffer();
+        String AND = " AND ";
+
+        // Filter by reference or chargeId
+        if (isNotBlank(reference)) {
+            subQuery.append(AND);
+            if( StringUtils.isNumeric(reference)) {
```
davidillsley: I don't understand this... if we want to support searching by charge id, why isn't is a separate param?

PauloPortugal:
rebuild

README.md
```diff
@@ -281,6 +282,64 @@ Location: http://connector.service/v1/api/charges/1
...
+
+```
+
+##### Query Parameters description
+
+| Field                    | required | Description                               |
+| ------------------------ |:--------:| ----------------------------------------- |
+| `reference`              | X | There (partial or full) reference issued by the government service for this payment. |
+| `status`                 | X | The transaction status |
+| `fromDate`               | X | The initial date to search transactions |
```
kakumara: Shouldn't we be using snake_case in URL query params  (e.g. from_date, to_date) ? (see https://github.com/paypal/api-standards/blob/master/api-style-guide.md#filtering)

README.md
```diff
@@ -281,6 +282,64 @@ Location: http://connector.service/v1/api/charges/1
...
+| `toDate`                 | X | The end date we should search transactions|
+
+#### Response example
+
+```
+201 Created
```
kakumara: Think this should be '200 OK' instead of 201.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -78,6 +80,26 @@ public Response getCharge(@PathParam("accountId") String accountId, @PathParam("
 
     }
 
+    @GET
+    @Path(CHARGES_API_PATH)
+    @Produces(APPLICATION_JSON)
+    public Response getCharges(@PathParam("accountId") String accountId,
+                               @QueryParam("reference") String reference,
+                               @QueryParam("status") String status,
+                               @QueryParam("fromDate") String fromDate,
```
kakumara: think query params names should be snake_case

src/main/java/uk/gov/pay/connector/util/ChargesSearch.java
kakumara: I second this too. Although we may have to construct a sql string eventually, a slightly safer pattern  (builder or similar) could make this a bit nicer.

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -80,6 +87,102 @@ public void insertAmountAndThenGetAmountById() throws Exception {
     }
```
kakumara: A general comment on the tests for search & filter.
It seems all tests here written for existing 1 charge that matches (or not match) the criteria. It would be nicer if a more richer test data set is used (e.g. 2 matching records and 1 unmatching record and then assert for result size is 2 and not including the 3rd one)

just to make sure we don't have spooky behaviour when multiple records exists.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -78,6 +80,26 @@ public Response getCharge(@PathParam("accountId") String accountId, @PathParam("
 
     }
 
+    @GET
+    @Path(CHARGES_API_PATH)
+    @Produces(APPLICATION_JSON)
+    public Response getCharges(@PathParam("accountId") String accountId,
```
kakumara: We now have two almost identical GET transactions endpoints
1. ChargesFrontendResource.getCharges()
2. ChargesApiResource.getCharges()

I feel it can be merged to one endpoint. Open to discuss.

README.md
PauloPortugal: :+1:

README.md
PauloPortugal: :+1:

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
PauloPortugal: :+1:

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
PauloPortugal: :+1:  
I have refactored **pay-self-service** and pay-connector to only use _ChargesApiResource.getCharges()_

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
PauloPortugal: :+1: This has been addressed!

src/main/java/uk/gov/pay/connector/util/ChargesSearch.java
PauloPortugal: @kakumara  
As we have discussed I have removed the construction of the IN clause.
I believe the others cannot be removed or changed. 
I have tried using a Builder Pattern here but I am not gaining much.

PR #87, opened 2016-01-15
PP-334 Adding the payment provider to the response

PauloPortugal:
- Capturing the payment provider name as part of the information
  that is sent back to selfservice.


davidillsley:
Looks good to me. I think you mean alphagov/pay-selfservice#37 is dependent on this?

PauloPortugal:
@davidillsley _"Looks good to me. I think you mean alphagov/pay-selfservice#37 is dependent on this?"_
You are correct.

PR #89, opened 2016-01-19
PP 424 - adding create date

PauloPortugal:
We need to return the CREATION event date for each payment. This was asked by Rory as based on Stripe functionality.


ijmad:
As discussed, what happens if the charge row goes in but the charge_event for created does not? We are not transactional enough in connector at the moment to guarantee this would be the case, which means if there was any delay between the two INSERTs (e.g. database locking the table) there would be a situation where you could get a 404 on a charge briefly that had legitimately been created (and received a 200 earlier) until charge_events catches up.

One option would be to change the join to a LEFT JOIN, but what would we do if we didn't get an event? We couldn't send back a legitimate created date at this point because we wouldn't have it, which could lead to a complicated API edge condition (e.g. created may not be returned). 

We could make the connector more transactional about its insert, but programming defensively we should guard against this here anyway if we can.

A better option might be to put a created date on the charge row, so maybe we could do this?

ijmad:
:+1: once the above is addressed.

PauloPortugal:
rebuild

PauloPortugal:
@ijmad  I have added a created date on the charge row.

src/main/resources/migrations.xml
```diff
@@ -97,4 +97,12 @@
         </createTable>
     </changeSet>
 
+    <changeSet id="add created date to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="created_date" type="timestamp without timezone">
+                <constraints nullable="true"/>
```
kakumara: Why should this be nullable=true ?

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -86,6 +88,7 @@ public void insertAmountAndThenGetAmountById() throws Exception {
         assertThat(charge.get("status"), is(CREATED.getValue()));
         assertThat(charge.get("gateway_account_id"), is(GATEWAY_ACCOUNT_ID));
         assertThat(charge.get("return_url"), is(RETURN_URL));
+        assertThat(charge.get("created_date"), is(createdDate));
```
kakumara: Isn't this kind of cheating (like assert true=true) where you use the  chargeDao.findById(chargeId).get() in @Before and in @Test? 
I think its better to use LocalDateTimeMatchers.within() here. Have a loot the assertions in test EventDaoTest.shouldRetrieveAllEventsForAGivenCharge()

src/test/java/uk/gov/pay/connector/util/DatabaseTestHelper.java
```diff
@@ -44,6 +50,7 @@ public void addGatewayAccount(String accountId, String paymentProvider) {
 
     public void addCharge(String chargeId, String gatewayAccountId, long amount, ChargeStatus status, String returnUrl, String transactionId) {
         addCharge(chargeId, gatewayAccountId, amount, status, returnUrl, transactionId, "Test description", "Test reference");
```
kakumara: Shouldn't this addCharge() implementation take care of the created_date now? (I think it works because the column is nullable now. Perhaps it should change)

src/test/java/uk/gov/pay/connector/it/dao/EventDaoTest.java
```diff
@@ -49,7 +49,7 @@ public void shouldRetrieveAllEventsForAGivenCharge() throws Exception {
         setupLifeCycleEventsFor(app, CHARGE_ID, statuses);
         List<ChargeEvent> events = eventDao.findEvents(GATEWAY_ACCOUNT_ID, CHARGE_ID);
 
-        assertThat(events.size(), is(statuses.size()));
+        assertThat(events.size(), is(statuses.size() + 1));
```
kakumara: may be better to put a developer comment here what this +1 is for (to avoid any confusion).

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
PauloPortugal: We are doing the same thing with the chargeId. 
Basically we create a charge, then we need to assert that the select statement returns the exact date in the correct format. 
I do not believe it is cheating.

src/main/resources/migrations.xml
PauloPortugal: That is a good question. 
We already have existing records that do not have the "created_date" field, so if I make them non nullable I would need to provide a default value.

The other option would be to do the following:

```
<column name="created_date" type="timestamp without timezone" defaultValueComputed="(now() at time zone 'utc')">
                <constraints nullable="false"/>
            </column>
```

but will this deal with the pre-existing records?

src/test/java/uk/gov/pay/connector/util/DatabaseTestHelper.java
PauloPortugal: This code should not be here. It was removed.

src/test/java/uk/gov/pay/connector/it/dao/EventDaoTest.java
PauloPortugal: This code should not be here. It was removed.

src/main/resources/migrations.xml
kakumara: we don't have live data, do we? so I don't this anyone is going to complain modifying test data with now(). But yes using now() by default seems like a good idea

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
kakumara: chargeId is taken from chargeDao.saveNewCharge() (not from chargeDao.findById()).
Isn't findById is the method we are testing here?

PauloPortugal: You are right, that is cheating.
:+1: for the LocalDateTimeMatchers. 
Code has been refactored.

src/main/resources/migrations.xml
PauloPortugal: As discussed. :+1:

kakumara:
Looks good to me :+1:

PR #91, opened 2016-01-26
Try alpine jdk

bazbremner:
@ijmad @joshmyers We've been talking about smaller images for a while, here's a very quick stab at it. Probably not the right image to chose, but it'll prove the point.

185MB vs 350MB; I'm interested to know how this affects deploy times and - of course - if it actually works in a real environment.

I've also no idea how devs give themselves confidence in changes to services - this gives me an excuse to find out.


ijmad:
docker-startup.sh needs bash. We should probably change the app startup to handle newrelic a little more sensibly, but let's try installing bash for now.

bazbremner:
Closing for now - still not working and not something I'm focusing on at the moment. I'll leave the branch.

PR #92, opened 2016-01-29
PP-495 fix external status mappings

kakumara:
- Mapped all internal statuses correctly to an external status
- Added a test to make sure the integrity between the two status mappings so that
  it makes sure there is an external status for all internal statuses


rahul-vinayak:
:+1:

bazbremner:
Nice, thanks for the speedy update. :+1: for the additional unit test.

PR #93, opened 2016-02-02
Pp 280 Download list of transactions in CSV format

rahul-vinayak:
- Connector changes with a new end point which accepts content type "text/csv" and returns list of transactions in a CSV format. 
- updated sql query to return created date coming from the charges table, as per the DB changes, in the field named date_created
- also get the above in a field name updated for the sake of backward compatibility for the view transactions on self service


rahul-vinayak:
rebuild

kakumara:
The datetime format needed to be changed as per the tech discussion couple of days ago. 
At storage/DAO level we should be storing/retrieving ISO-8601 UTC ("yyyy-MM-dd'T'HH:mm:ssZ") format. (We can use `java.time.format.DateTimeFormatter.ISO_INSTANT.withZone("Z")` and `java.time.ZonedDateTime` to achieve this )

But we may have a problem right now as we are storing timezoneless datetimes in the DB which needs a DB structure change and data change.
Let's discuss if this needs to be done in this PR or in a separate story

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -166,17 +166,17 @@ public int updateNewStatusWhereOldStatusIn(String chargeId, ChargeStatus newStat
...
                                                String fromDate, String toDate) {
-        String query =
-                "SELECT DISTINCT c.charge_id, c.gateway_transaction_id, c.status, c.amount, " +
-                        "c.description, c.reference, to_char(ce.updated, 'YYYY-MM-DD HH24:MI:SS') as updated " +
-                        "FROM charges c " +
-                        "INNER JOIN charge_events ce ON (c.charge_id = ce.charge_id AND ce.status = 'CREATED')" +
-                        "WHERE c.gateway_account_id=:gid " +
-                        "%s " +
-                        "ORDER BY c.charge_id DESC";
+        String query = "SELECT DISTINCT c.charge_id, c.gateway_transaction_id, c.status, c.amount, " +
```
kakumara: Perhaps format this as per Barry's suggestion?

bazbremner: Barrie.

rahul-vinayak:
rebuild

kakumara:
Looks ok :+1:

PR #94, opened 2016-02-04
PP 421

kakumara:
[PP-421 This needs to merge first]
Extended connector search/filter functionality to support filter by ISO_8601 UTC formatted datetime
- Supporting ISO_8601 UTC datetime format `yyyy-MM-ddTHH:mm:ssZ`  
- Keeping the "TimeZone less" datetime format supported for backward compatibility.
- Added integration tests to check search filter by from_date and to_date fields in above format.
- When retrieving created_date from database, converting to a ZonedDateTime assuming the stored datetime is in UTC.


src/main/java/uk/gov/pay/connector/resources/ApiValidators.java
```diff
@@ -0,0 +1,48 @@
...
+import static fj.data.Either.right;
+import static java.lang.String.format;
+import static org.apache.commons.lang3.StringUtils.*;
+import static org.apache.commons.lang3.math.NumberUtils.isNumber;
+import static uk.gov.pay.connector.resources.ChargesApiResource.FROM_DATE_KEY;
+import static uk.gov.pay.connector.resources.ChargesApiResource.TO_DATE_KEY;
+
+public class ApiValidators {
+
+    public static Optional<String> validateQueryParams(List<Pair<String, String>> queryParams) {
```
ckalista: Since the aim of this method is to validate date format, are we not better off providing a list of parameter names we want to enforce validation against. (FROM_DATE_KEY and TO_DATE_KEY references tightly couple with ChargeApiResource here) Also the method name should be a little more explicit then i.e. validateDateQueryParams.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -99,28 +99,47 @@ public Response getCharge(@PathParam("accountId") String accountId, @PathParam("
...
-                                   @QueryParam("to_date") String toDate,
+                                   @QueryParam(REFERENCE_KEY) String reference,
+                                   @QueryParam(STATUS_KEY) String status,
+                                   @QueryParam(FROM_DATE_KEY) String fromDate,
+                                   @QueryParam(TO_DATE_KEY) String toDate,
                                    @Context UriInfo uriInfo) {
 
-        return reduce(validateGatewayAccountReference(accountId)
-                .bimap(handleError, listChargesAsJsonResponse(accountId, reference, status, fromDate, toDate)));
+        Optional<String> errorMessageOptional = ApiValidators.validateQueryParams(ImmutableList.of(
```
ckalista: Is there a technical reason for not using Bean Validator to do the query param validation? We could introduce a new Annotation @UTCZonedDateTime.

See the 'email' example here: https://jersey.java.net/documentation/2.13/bean-validation.html

src/main/java/uk/gov/pay/connector/resources/ApiValidators.java
```diff
@@ -0,0 +1,48 @@
...
+                });
+
+        if (invalidQueryParams.size() > 0) {
+            return Optional.of(format("query parameters [%s] not in correct format",
+                    join(invalidQueryParams, ", ")));
+        }
+        return Optional.empty();
+    }
+
+    public static Either<String, Boolean> validateGatewayAccountReference(String gatewayAccountId) {
```
ckalista: Nice to see all the functional programming goodies been used properly :+1:

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -99,28 +99,47 @@ public Response getCharge(@PathParam("accountId") String accountId, @PathParam("
...
-                                   @QueryParam("from_date") String fromDate,
-                                   @QueryParam("to_date") String toDate,
-                                   @Context UriInfo uriInfo) {
+                                  @QueryParam(REFERENCE_KEY) String reference,
+                                  @QueryParam(STATUS_KEY) String status,
+                                  @QueryParam(FROM_DATE_KEY) String fromDate,
+                                  @QueryParam(TO_DATE_KEY) String toDate,
+                                  @Context UriInfo uriInfo) {
+
+        Optional<String> errorMessageOptional = ApiValidators.validateQueryParams(ImmutableList.of(
```
ckalista: Same comment as above. If we manage to introduce this new annotation we could avoid a lot of code duplication.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -267,8 +286,8 @@ private boolean isFieldSizeValid(Map<String, Object> chargeRequest, String field
         }
         List<Map<String, Object>> charges = chargeDao.findAllBy(accountId, reference, chargeStatus, fromDate, toDate);
         charges.forEach(charge -> {
-                    charge.put(STATUS_KEY, mapFromStatus(charge.get(STATUS_KEY).toString()).getValue());
-                    charge.put(CREATED_DATE, DATE_FORMAT.format(charge.get(CREATED_DATE)));
+            charge.put(STATUS_KEY, mapFromStatus(charge.get(STATUS_KEY).toString()).getValue());
+            charge.put(CREATED_DATE, toUTCDateString((ZonedDateTime) charge.get(CREATED_DATE)));
```
ckalista: I wonder why we don't have a DTO for charge? If we had one then this sort of transformation could be taken care of there with a JsonSerialize annotation?

ckalista:
Code review completed.

src/main/java/uk/gov/pay/connector/resources/ApiValidators.java
kakumara: :+1:  Addressed

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
kakumara: Yes, this seems like a reasonable suggestion. I will include this in the next pull request. For this I have to support backward compatibility temporarily. When I remove that I'll use this bean validation approach.

src/main/java/uk/gov/pay/connector/resources/ApiValidators.java
kakumara: :+1:

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
kakumara: Valid concern. This is currently like this due to historical reasons and this need has been identified for sometime in few tech-retrospectives. This will be addressed as part of the planned connector refactoring.

PR #97, opened 2016-02-08
[REVIEW ONLY] PP-438 pass HTTP proxy gubbins to JVM

bazbremner:
Testing this is likely going to require changes to our build pipeline, but opening for visibility, since this has been lurking on a branch for too long.


bazbremner:
rebuild

ijmad:
Looks sensible to me :+1:

docker-startup.sh
```diff
@@ -1,5 +1,26 @@
...
+# It'd be nice if Java could digest the standard HTTP_PROXY and
+# HTTPS_PROXY environment variables, but no. Rather than do too much
+# munging of variables in this script, I expect whatever is handling
+# the containers to set the mulitude of variables required.
+
+if [ -n "${HTTP_PROXY_HOST}" && -n "${HTTP_PROXY_PORT}" ] ; then
+    HTTP_PROXY_SETTINGS="-Dhttp.proxyHost=${HTTP_PROXY_HOST} -Dhttp.proxyPort=${HTTP_PROXY_PORT}"
+    HTTPS_PROXY_SETTINGS="-Dhttps.proxyHost=${HTTPS_PROXY_HOST} -Dhttps.proxyPort=${HTTPS_PROXY_PORT}"
+
+    if [ -n "${}" ] ; then
```
bazbremner: Looks like I missed a variable name here - will fix shortly.

ijmad:
BTW, another option here would be to parse more standard options in Java and have it set its own -D's as a result. Do you want to look at doing this instead?

bazbremner:
@ijmad Let's stick with what's here - I've already got the Ansible changes deployed, it's not that complex and the startup shell script needs re-visiting at some point anyway.

joshmyers:
LGTM!

PR #99, opened 2016-02-11
PP-482 made changes to connector public API for events so that consecutive repetitive external statuses dont show up

rahul-vinayak:
- Bug fix, we have more than one internal charge statuses mapped to a single external charge status, which meant that internal charge state transition lead to showing of same external status events more than once both in public api and the self service, look into the Jira [PP-482](https://payments-platform.atlassian.net/browse/PP-482) for more description about the bug.
- Fix is to purge the external event if it was the same as immediate previous event. e.g. You won’t have events like created, in-progress, in-progress; but you can have events like created, in-progress, cancelled, in-progress.


bazbremner:
Sorry, I don't really understand the _why_ of this change from the commit message. Could you elaborate and clarify your PR commits and preserve them for future generations in the commit message, please?

I remember from some meatspace conversations as to why this problem occurs, but it's not clear from the commit message and I imagine we'll all have forgotten about it in a few months.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventExternal.java
```diff
@@ -0,0 +1,87 @@
...
+        ChargeEventExternal that = (ChargeEventExternal) o;
+
+        if (!chargeId.equals(that.chargeId)) return false;
+        return status == that.status;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = chargeId.hashCode();
+        result = 31 * result + status.hashCode();
```
bazbremner: Why is there a magic number (31) here? What relevance does it have? If it's important, I'd suggest getting it into a constant with a helpful name and - if required - comment appropriately. At first glance it looks concerning.

PauloPortugal:
@rahul-vinayak and @bazbremner 
I agree with the commit higiene, although the title is almost self-explanatory.
I always prefer to have a descriptive text, possibly with an example as you provided in this PR

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventExternal.java
PauloPortugal: @bazbremner 
The number 31 is not a problem, as this is generated by the IDE, a standard practice in Java projects and this is the hashCode() method. A Java developer should know what is there for and that it is generated code. _Not an issue_.

_I have realized the status is being used. All good :smile: _

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -185,10 +187,36 @@ public Response createNewCharge(@PathParam("accountId") String accountId, Map<St
...
+
+        return externalEvents;
+    }
+
+    private List<ChargeEventExternal> getNonRepeatingChargeEvents(List<ChargeEventExternal> externalEvents) {
+        ListIterator<ChargeEventExternal> iterator = externalEvents.listIterator();
+        ChargeEventExternal prev = null;
+        while (iterator.hasNext()) {
+            ChargeEventExternal current = iterator.next();
+            if (current.equals(prev)) { // remove any immediately repeating statuses
```
PauloPortugal: :+1:

rahul-vinayak:
Updated the PR commit message to a bit more explanatory now, more details on the Jira that should describe the bug.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventExternal.java
bazbremner: @PauloPortugal sorry, but your explanation doesn't give me any additional information. I can see the method is called `hashCode()`, but that doesn't help me understand what's actually going on.

I'm still twitchy about magical/autogenerated numbers, be they generated by a human or their IDE.

I'm not a "Java developer", but I am a developer that would like to work on Java code bases and I'll also have to debug this application when it's running in production, so a passing familiarity with the codebase is handy. To that end, I'll occasionally poke through commits and I'll need help understanding idioms, but I _will_ need to understand the reasoning.

I don't want to derail this PR - I'll catch up in meatspace.

rahul-vinayak: Its the industry best practice to use a prime number to multiply to other hashcodes, which promises distinct hash codes for distinct objects. Only magical powers 31 here have is that its a prime number big enough to ensure distinction and is small enough not to generate obscenely large numbers. That is why a prime number in that range has become industry standard, but it could be any prime number. Its not a constant because it has no significance out of this method and no one cares about it out of this method.
http://www.ibm.com/developerworks/library/j-jtp05273/

bazbremner:
@rahul-vinayak, @PauloPortugal sorry, I don't think I made my point about commit messages clearly enough.

The version control system (in this case git) is close to the code, commits should be granular, and can be mined for information (for example, with the [pick-axe](http://www.philandstuff.com/2014/02/09/git-pickaxe.html)). 
PRs are coarser (often many commits), ephemeral and difficult to search and contextualise as they're a little further from the code. Thanks to some magic, information from commits find their way into PRs, so if you write comprehensive commit messages, you'll get comprehensive PRs; the reverse is not true.
Bug trackers/ticketing systems come and go, are very coarse (story or epic level) and are detached from the code base.

The commit history of a code base is highly likely to be migrated over time, whereas PRs and ticketing systems generally fall by the wayside.

So, my point was that I'd much rather that the detail you added to the PR was actually added to the commit message. It'll be useful for longer. Your PR will fade from our memories quickly.

I frequently refer back to older commit messages and appreciate the information that is present there, but it often takes weeks or months before that benefit reveals itself.

A good PR message is for Christmas, a good commit message is for life.

rahul-vinayak:
@bazbremner thats a fair point regarding a more descriptive commit messages. Point noted.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventExternal.java
bazbremner: @rahul-vinayak excellent, thanks for the reading material! :+1:

PR #100, opened 2016-02-14
PP-506 Nullpointer on worldpay notification

ckalista:
1. Rename worldpay variables to smartpay
2. Blacklist all Worldpay notifications of type 'AUTHORISED'
   - WorldpayStatusesBlacklist, enlist all notification statuses currently black listed i.e. ‘AUTHORISATION’.
   - Short-circuit WorldpayProvider to ignore handling blacklisted statuses with a log.
   - Unit tests and integration tests.
3. Blacklist all Smartpay notifications of type 'AUTHORISED'
   - SmartpayStatusesBlacklist, enlist all notification statuses currently black listed i.e. ‘AUTHORISATION’.
   - Short-circuit SmartpayProvider to ignore handling blacklisted statuses with a log.
   - Unit tests and integration tests.


kakumara:
Just had a quick look through the story and the fix.
- The fix in overall seems fine for what was discussed.

However, when I looked at the original issue I'm not sure if the original Null Pointer issue is fixed. My gut feel is its thrown from `ChargeDao.findAccountByTransactionId` line 208 
`return Optional.ofNullable(data.get("gateway_account_id").toString());` 
where the result of `data.get("gateway_account_id")` is null we are trying to convert to string. This can happen if the transactionId returned from gateway does not exist and we have not been defensive here. 
{Also looks like we have not tested this method properly in `ChargeDaoITest` in the past :-(  . Just wrote the below test and its throwing the null pointer

```
@Test
    public void should_findAccountByTransactionId() throws Exception {
        chargeDao.findAccountByTransactionId(PAYMENT_PROVIDER,"unknown-reference");

    }
```

 }

ckalista:
As discussed, the issue spotted above is a different one than the one addressed by this PR. A new JIRA will be raise to handle the scenario outline in a more defensive manner.

JIRA updated and new ticket created: PP-553 Nullpointer exception thrown when notification is sent with wrong transactionId

kakumara:
OK :+1:

PR #102, opened 2016-02-17
Pp 553/nullpointer on notification with wrong txid

ckalista:
- Introduce preliminary payload checks
  - Moved transactionId check and ChargeStatus checks out to its own function.
  - Changed handler interface to take a payloadChecks callback, to normalise behaviour across Providers.
  - Fix unit tests
- Defensive handling of Nullpointer Exception in ChargeDao
  - Null checks on data to avoid Nullpointer exception in the following two DAO methods
    - findChargeByTransactionId
    - findAccountByTransactionId
    - Update test coverage.
- Defensive check to ensure transactionId is not empty
- Refactor paylodCheck in its own class
  - Move the payload checks in its own class for easier testing in isolation.


src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -205,7 +205,11 @@ public int updateNewStatusWhereOldStatusIn(String chargeId, ChargeStatus newStat
                         .first()
         );
 
-        return Optional.ofNullable(data.get("gateway_account_id").toString());
+        if(data != null && data.get("gateway_account_id") != null) {
+            return Optional.ofNullable(data.get("gateway_account_id").toString());
```
rahul-vinayak: If data.get("gateway_account_id") is never null by the line 209 we may not need to do Optional.ofNullable

rahul-vinayak: In fact looking at the code that calls this method, returning Optional wont solve the purpose as the calling code will still throw NoSuchElementException instead of NullPointer as it stands right now.... although that code wasnt part of this PR but still rung a bell.....up for a discussion ()

rahul-vinayak:
Overall looks good :+1:

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
ckalista: Fix applied.

Excellent point!!! Actually I was meant to address this in the first place but somehow left this bit behind. Thanks for pointing it out.

ckalista:
Code review changes applied.

PR #103, opened 2016-02-29
PP-542 Migrate to JPA

rauligs:
- Changes don't include refactoring, so it is still pending but not done as part of this pull request since it contains lots of changes and it can unblock current stories.
- For pay-endtoend to work it must be also merged into master.


src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -1,33 +1,34 @@
 package uk.gov.pay.connector.app;
 
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.google.inject.Provider;
```
ckalista: Remove unused import.

src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
```diff
@@ -4,6 +4,7 @@
 import io.dropwizard.Configuration;
 import io.dropwizard.client.JerseyClientConfiguration;
 import io.dropwizard.db.DataSourceFactory;
+import io.dropwizard.db.DatabaseConfiguration;
```
ckalista: Remove unused import.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.pay.connector.app;
+
+import com.google.inject.AbstractModule;
+import com.google.inject.persist.PersistFilter;
```
ckalista: Remove unused import.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -0,0 +1,103 @@
+package uk.gov.pay.connector.model.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.persistence.*;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.util.Optional;
```
ckalista: Remove unused import.

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
```diff
@@ -0,0 +1,59 @@
...
+
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+//TODO: rename this when we can get rid of Jpa
+@Transactional
+public class JpaDao<T> {
```
ckalista: :+1:

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
```diff
@@ -0,0 +1,59 @@
...
+import com.google.inject.Provider;
+import com.google.inject.persist.Transactional;
+
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+//TODO: rename this when we can get rid of Jpa
```
ckalista: Does this TODO make sense? Why would we want to get rid of Jpa?

rauligs: That must be a typo, I'll remove it

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+    private GatewayAccountJpaDao gatewayAccountDao;
+
+    @Inject
+    public ChargeJpaDao(final Provider<EntityManager> entityManager, ChargeEventJpaListener eventListener, GatewayAccountJpaDao gatewayAccountDao) {
+        super(entityManager);
+        this.eventListener = eventListener;
+        this.gatewayAccountDao = gatewayAccountDao;
+    }
+
+    public void create(ChargeEntity charge) {
```
ckalista: This method is only used for testing purposes and has no other function other than actioning the eventListener to notify of the change. In the future the notification could be taken care in a much cleaner way by using JPA listeners (see @PrePersist in JPA) hence this method would have no value at all. For now I would suggest maybe moving the method somewhere in our tests so not to contaminate the DAO api.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+        super.persist(charge);
+        eventListener.notify(ChargeEventEntity.from(charge, CREATED, charge.getCreatedDate().toLocalDateTime()));
+    }
+
+    public ChargeEntity update(final ChargeEntity charge) {
```
ckalista: Same comment as above.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Agree. It will be removed, if needed it will be created as part of the refactoring (with event notification likely in a JPA listener as previously discussed)

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Same as my previous response.

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -37,31 +38,31 @@ public void initialize(Bootstrap<ConnectorConfiguration> bootstrap) {
...
         environment.healthChecks().register("ping", new Ping());
 
-        jdbi = new DBIFactory().build(environment, dataSourceFactory, "postgresql");
-
-        EventDao eventDao = new EventDao(jdbi);
-        ChargeEventListener chargeEventListener = new ChargeEventListener(eventDao);
-        ChargeDao chargeDao = new ChargeDao(jdbi, chargeEventListener);
-        TokenDao tokenDao = new TokenDao(jdbi);
-        GatewayAccountDao gatewayAccountDao = new GatewayAccountDao(jdbi);
+        IEventDao eventDao = injector.getInstance(EventJpaDao.class);
```
kakumara: I don't think this is the right approach to inject Dao's in to resources. Why not use guice to instantiate the resources below with the injected jpaDao's? 
And inject the right dao in the jpa module.

```
environment.jersey().register(injector.getInstance(ChargesApiResource.class));
```

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -37,31 +38,31 @@ public void initialize(Bootstrap<ConnectorConfiguration> bootstrap) {
...
-        bootstrap.addCommand(new DbWaitCommand());
     }
 
     @Override
     public void run(ConnectorConfiguration conf, Environment environment) throws Exception {
+
+        final Injector injector = Guice.createInjector(new ConnectorModule(conf, environment), createJpaModule(conf.getDatabaseConfig()));
+        environment.servlets().addFilter("persistFilter", injector.getInstance(PersistFilter.class))
+                .addMappingForUrlPatterns(EnumSet.of(DispatcherType.REQUEST), true, "/*");
+        environment.jersey().register(injector.getInstance(GatewayAccountJpaResource.class));
```
kakumara: We did these 2 Jpa resources with a different refactoring in mind. Now that you have gone for a Dao interface, they are no longer needed ??

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
rauligs: Agree with the comment but it will be done in PP-576 as part of the refactoring, I'll put this comment in the refactoring story since connector will be tidied up.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+                new ChargeEntity(new Long(charge.get("amount").toString()),
+                        CREATED.getValue(),
+                        null,
+                        charge.get("return_url").toString(),
+                        charge.get("description").toString(),
+                        charge.get("reference").toString(),
+                        gatewayAccountEntity);
+
+        super.persist(chargeEntity);
+        entityManager.get().flush();
```
kakumara: This `flush` bit sucks in here. Normally yon't need it unless the behaviour of the EntityManager having a spooky behaviour. We should avoid using it unless there's no other way.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+        eventListener.notify(ChargeEventEntity.from(chargeEntity, newStatus, LocalDateTime.now()));
+    }
+
+    @Override
+    public List<Map<String, Object>> findAllBy(String gatewayAccountId, String reference, ExternalChargeStatus status, String fromDate, String toDate) {
+        ChargeSearchQuery searchQuery = new ChargeSearchQuery(new Long(gatewayAccountId));
+        searchQuery.withReferenceLike(reference);
+        searchQuery.withExternalStatus(status);
+        searchQuery.withCreatedDateFrom(fromDate);
+        searchQuery.withCreatedDateTo(toDate);
```
kakumara: (minor) isn't chaining these `with` methods together is bit nicer than this?

src/main/java/uk/gov/pay/connector/dao/ChargeSearchQuery.java
```diff
@@ -0,0 +1,92 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.TypedQuery;
+import java.time.ZonedDateTime;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+
+public class ChargeSearchQuery {
```
kakumara: Nice :+1:

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
rauligs: They will be used, the interface will be removed and these JpaResources potentially will replace the existing resources, since they are already tested is a step closer to have the refactoring done.
Added as a comment in the PP-576 story to make sure these resources are also removed/reused depending on the approach taken.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    public Optional<ChargeEntity> findByGatewayTransactionIdAndProvider(String transactionId, String paymentProvider) {
+        TypedQuery<ChargeEntity> query = entityManager.get()
+                .createQuery("select c from ChargeEntity c where c.gatewayTransactionId = :gatewayTransactionId and c.gatewayAccount.gatewayName = :paymentProvider", ChargeEntity.class);
+
+        query.setParameter("gatewayTransactionId", transactionId);
+        query.setParameter("paymentProvider", paymentProvider);
+
+        return Optional.ofNullable(query.getSingleResult());
```
ckalista: getSingleResult is less likely to ever return null although if no record can be found a EntityNotFoundException will be thrown. This needs to be handled here I guess. We also need a test for that particular case.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Absolutely, this will disappear from this method as soon is refactored, so we need to make sure we use the chained way (already supported). So to cover as part of PP-576. Also the search method will receive the SearchQuery, at the moment not doing it so to conform to the legacy interface.

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public <T> void remove(final T object) {
+        entityManager.get().remove(object);
+    }
+
+    public <T, ID> void removeById(final Class<T> clazz, final ID id) {
+        remove(findById(clazz, id));
+    }
+
+    public <T> List<T> findAll(final Class clazz) {
```
kakumara: Unused method

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
```diff
@@ -0,0 +1,59 @@
...
+        final String query = String.format(QUERY_SELECT_ALL, clazz.getSimpleName());
+        return entityManager.get().createQuery(query).getResultList();
+    }
+
+    public <T> List<T> find(final Class<T> clazz, final String namedQuery, final Map<String, Object> paramsMap) {
```
kakumara: unused method

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public <T> T merge(final T object) {
+        return entityManager.get().merge(object);
+    }
+
+    public <T> void remove(final T object) {
+        entityManager.get().remove(object);
+    }
+
+    public <T, ID> void removeById(final Class<T> clazz, final ID id) {
```
kakumara: unused method

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    public List<ChargeEntity> findAllBy(ChargeSearchQuery searchQuery) {
```
ckalista: This method is only used by the tests so could be moved to the tests for now. Then during the refactoring part (part II of the user story) the method is likely to resurface replacing List<Map<String, Object>> findAllBy

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Yes, it will be covered as part of the PP-576 refactoring, saveNewCharge will disappear and proper listeners must be in placed to handle notifications. This is captured in PP-576 comments anyway.

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
kakumara: It is a typo. I wrote this comment as we copied this `JpaDao.java` from somewhere. May be we should rename it to something more generic to us and we dont have to sa 'Jpa' as thats our default approach. (e.g `BaseDao`)

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventEntity.java
```diff
@@ -0,0 +1,69 @@
...
+    private ChargeEntity chargeEntity;
+
+    @Convert(converter = ChargeStatusConverter.class)
+    private ChargeStatus status;
+
+    @JsonDeserialize(using = LocalDateTimeDeserializer.class)
+    @JsonSerialize(using = LocalDateTimeSerializer.class)
+    @JsonFormat(pattern="yyyy-MM-dd HH:mm:ss")
+    @Convert(converter = LocalDateTimeConverter.class)
+    private LocalDateTime updated;
```
kakumara: we should avoid using LocalDateTime. Instead try using ZonedDateTime (see ChargeEntity) 
Use the DateTimeUtils for necessary conversions.

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
rauligs: I'll remove all the unused methods, as part of the refactoring we can see what could be better to do (it might be added methods or just remove it).

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
ckalista: Agreed - seems like a code smell but as you said it should vanish soon.

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
rauligs: Removed

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
rauligs: Removed

src/main/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,23 @@
...
+        <provider>org.eclipse.persistence.jpa.PersistenceProvider</provider>
+        <class>uk.gov.pay.connector.model.domain.ChargeEntity</class>
+        <class>uk.gov.pay.connector.model.domain.ChargeEventEntity</class>
+        <class>uk.gov.pay.connector.model.domain.GatewayAccountEntity</class>
+        <class>uk.gov.pay.connector.model.domain.TokenEntity</class>
+        <class>uk.gov.pay.connector.model.domain.LocalDateTimeConverter</class>
+        <class>uk.gov.pay.connector.model.domain.ChargeStatusConverter</class>
+        <class>uk.gov.pay.connector.model.domain.CredentialsConverter</class>
+        <class>uk.gov.pay.connector.model.domain.UTCDateTimeConverter</class>
+        <exclude-unlisted-classes>true</exclude-unlisted-classes>
```
kakumara: may be better to scan the package here rather than listing specific classes? Coz we'll have to keep maintaining this list now with every change (add/removal)

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+                        gatewayAccountEntity);
+
+        super.persist(chargeEntity);
+        entityManager.get().flush();
+        eventListener.notify(ChargeEventEntity.from(chargeEntity, CREATED, chargeEntity.getCreatedDate().toLocalDateTime()));
+        return chargeEntity.getId().toString();
+    }
+
+    @Override
+    public Optional<Map<String, Object>> findChargeForAccount(String chargeId, String accountId) {
```
ckalista: In genera I don't think we should convert types internally in our DAOs. Can we not expose a Long for id and let the caller do whatever appropriate?

src/main/resources/config/config.yaml
```diff
@@ -44,6 +44,12 @@ jerseyClient:
 customJerseyClient:
   readTimeout: 20000ms
 
+databaseJpa:
```
kakumara: may be we should rename this to just `database`. (originally we introduce `databaseJpa` to coexist with DBI stuff, but no longer needed.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+        Optional<ChargeEntity> chargeEntityOpt = findById(new Long(chargeId));
+        if (chargeEntityOpt.isPresent() && chargeEntityOpt.get().getGatewayAccount().getId().toString().equals(accountId)) {
```
ckalista: Maybe we could rewrite this in a more functional way with .map. Also same comment as above, we should have to convert gatewayAccount to string.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -0,0 +1,222 @@
...
+            return Optional.of(buildChargeMap(chargeEntityOpt.get()));
+        }
+        return Optional.empty();
+    }
+
+    @Override
+    public Optional<Map<String, Object>> findById(String chargeId) {
```
ckalista: All parameters should be of exact same type as entity attribute, so here should be a Long. Same applies for everywhere else in the code.

src/test/java/uk/gov/pay/connector/dao/ChargeSearchQueryTest.java
```diff
@@ -0,0 +1,176 @@
...
+        verify(queryMock).setParameter("gatewayAccountId", GATEWAY_ACCOUNT_ID);
+        verify(queryMock).setParameter("toDate", TO_DATE);
+        verifyNoMoreInteractions(queryMock, entityManagerMock);
+    }
+
+    @Test(expected = NullPointerException.class)
+    public void shouldFailWhenGatewayAccountIsNotSpecified() {
+        new ChargeSearchQuery(null).apply(entityManagerMock);
+    }
+}
```
kakumara: Fix line ending. try `git config --global core.autocrlf input` and hopefully git will sort this out

src/test/java/uk/gov/pay/connector/it/dao/EventDaoJpaTest.java
```diff
@@ -90,11 +96,9 @@ public void describeTo(Description description) {
...
     }
 
-
     public static void setupLifeCycleEventsFor(DropwizardAppWithPostgresRule app, Long chargeId, List<ChargeStatus> statuses) {
         statuses.stream().forEach(
                 st -> app.getDatabaseTestHelper().addEvent(chargeId, st.getValue())
         );
     }
-
 }
```
kakumara: fix line ending

src/test/java/uk/gov/pay/connector/it/resources/ChargesFrontendResourceITest.java
```diff
@@ -275,4 +273,10 @@ private ValidatableResponse validateGetCharge(long expectedAmount, String charge
     private String expectedChargeUrl(String chargeId, String path) {
         return "http://localhost:" + app.getLocalPort() + CHARGE_FRONTEND_PATH.replace("{chargeId}", chargeId) + path;
     }
+
+    private static void setupLifeCycleEventsFor(DropwizardAppWithPostgresRule app, Long chargeId, List<ChargeStatus> statuses) {
+        statuses.stream().forEach(
+                st -> app.getDatabaseTestHelper().addEvent(chargeId, st.getValue())
+        );
+    }
 }
```
kakumara: fix line ending

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: This method is being used internally, making it private (will be potentially used public as part of the refactoring)

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventEntity.java
rauligs: This change is commented in the PP-576 story so it will be covered

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendJpaResourceITest.java
```diff
@@ -0,0 +1,170 @@
...
+
+    private Gson gson =  new Gson();
+
+    @Test
+    public void shouldGetCredentialsForExistingAccount() {
+        String accountId = createAGatewayAccountFor("worldpay");
+        ImmutableMap<String, String> credentials = ImmutableMap.of("username", "a-username", "password", "a-password", "merchant_id", "a-merchant-id");
+
+        updateCredentialsWith(accountId, credentials);
+        // TODO investigate why does not work updating db with databaseHelper
```
kakumara: Is this TODO still valid here? If we cannot fix it may be we should remove the comment

src/test/java/uk/gov/pay/connector/rules/PostgresDockerRule.java
```diff
@@ -75,4 +75,8 @@ public void stop() {
         container.stop();
         container = null;
     }
+
+    public String getDriverClass() {
+        return "org.postgresql.Driver";
+    }
 }
```
kakumara: Fix line ending (probably my fault when I did this code)

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Indeed, this must be covered in PP-576 so I'll make sure it is in the list of TODOS

src/main/resources/config/config.yaml
rauligs: This is already removed, pending being pushed after the review

src/test/resources/config/test-config.yaml
```diff
@@ -27,6 +27,12 @@ smartpay:
     username: smartpay_notification_username
     password: smartpay_notification_password
 
+databaseJpa:
```
kakumara: rename to 'database'

src/test/resources/config/test-contract-config.yaml
```diff
@@ -27,6 +27,13 @@ smartpay:
     username: not-contract-tested
     password: not-contract-tested
 
+databaseJpa:
```
kakumara: rename to `database`

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Yes, agreed, this will be disappear as part of PP-576

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Yes, this is a pending change for PP-576 (this was just to conform to the legacy interface).

src/main/java/uk/gov/pay/connector/dao/ChargeSearchQuery.java
```diff
@@ -0,0 +1,92 @@
...
+        if (status != null) queryParameters.put(STATUSES, Arrays.asList(status.getInnerStates()));
+        return this;
+    }
+
+    public ChargeSearchQuery withCreatedDateFrom(ZonedDateTime fromDate) {
+        if (fromDate != null) queryParameters.put(FROM_DATE, fromDate);
+        return this;
+    }
+
+    public ChargeSearchQuery withCreatedDateTo(ZonedDateTime toDate) {
```
ckalista: We shouldn't do the conversion here. What if the date format is not as expected? 

This is because the DAO expects the date as a String in its API instead of ZoneDateTime.

src/test/java/uk/gov/pay/connector/it/resources/ChargesFrontendResourceITest.java
rauligs: Done

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendJpaResourceITest.java
rauligs: Not necessary, were introduce doing testing for caching issues. So removed.

src/test/java/uk/gov/pay/connector/rules/PostgresDockerRule.java
rauligs: Done

src/test/resources/config/test-config.yaml
rauligs: Done

src/test/resources/config/test-contract-config.yaml
rauligs: Done

src/main/java/uk/gov/pay/connector/dao/ChargeSearchQuery.java
rauligs: Yes this is already something to be changed as part of PP-576

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
```diff
@@ -0,0 +1,67 @@
...
+    @Inject
+    public GatewayAccountJpaDao(final Provider<EntityManager> entityManager) {
+        super(entityManager);
+    }
+
+    @Override
+    public String createGatewayAccount(String paymentProvider) {
+        GatewayAccountEntity gatewayAccount = new GatewayAccountEntity(paymentProvider, new HashMap<>());
+        super.persist(gatewayAccount);
+        entityManager.get().flush();
```
ckalista: Flush here should be removed (for same reasons as above).

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
```diff
@@ -0,0 +1,67 @@
...
+@Transactional
+public class GatewayAccountJpaDao extends JpaDao<GatewayAccountEntity> implements IGatewayAccountDao {
+
+    @Inject
+    public GatewayAccountJpaDao(final Provider<EntityManager> entityManager) {
+        super(entityManager);
+    }
+
+    @Override
+    public String createGatewayAccount(String paymentProvider) {
```
ckalista: Method only used in tests, hence as above probably better to move it as part of tests.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rauligs: Added failing test and so catches NotFoundException for now (this should be removed as soon the refactoring takes place)

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
rauligs: Yes it will be removed as part of PP-576

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
rauligs: It is used in GatewayAccountResource

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public boolean idIsMissing(String gatewayAccountId) {
+        return !super.findById(GatewayAccountEntity.class, Long.valueOf(gatewayAccountId)).isPresent();
+    }
+
+    @Override
+    public Optional<GatewayAccount> findById(String gatewayAccountId) {
+        return super.findById(GatewayAccountEntity.class, Long.valueOf(gatewayAccountId)).map(gatewayAccountEntity -> {
+            GatewayAccount gatewayAccount = null;
+            if (gatewayAccountEntity != null) {
```
ckalista: Why do we need to check if gatewayAccountEntity != null here? It should be guaranteed to be not null as per Optional contract.

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
```diff
@@ -0,0 +1,67 @@
...
+        });
+    }
+
+    @Override
+    public Optional<GatewayAccountEntity> findById(Long gatewayAccountId) {
+        return super.findById(GatewayAccountEntity.class, gatewayAccountId);
+    }
+
+    @Override
+    public void saveCredentials(String credentialsJsonString, String gatewayAccountId) {
```
ckalista: Wow this looks wrong, having to parse a Json string in the DAO is definitely not the right thing to do. I understand this is legacy code so we should create a tech debt item for it.

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
ckalista: As discussed I think these method even thought not used now are very likely to be needed in the future and therefore better leave them here to give general direction of what should be in this class.

src/main/java/uk/gov/pay/connector/resources/GatewayAccountJpaResource.java
```diff
@@ -0,0 +1,158 @@
...
+import static uk.gov.pay.connector.util.ResponseUtil.notFoundResponse;
+
+@Path("/")
+public class GatewayAccountJpaResource {
+    public static final String ACCOUNTS_API_JPA_RESOURCE = "/v1/api/jpa/accounts";
+    public static final String ACCOUNT_API_JPA_RESOURCE = ACCOUNTS_API_JPA_RESOURCE + "/{accountId}";
+
+    public static final String ACCOUNTS_FRONTEND_JPA_RESOURCE = "/v1/jpa/frontend/accounts";
+    public static final String ACCOUNT_FRONTEND_JPA_RESOURCE = ACCOUNTS_FRONTEND_JPA_RESOURCE + "/{accountId}";
+    private static final Logger logger = LoggerFactory.getLogger(GatewayAccountResource.class);
```
ckalista: Logger should log against GatewayAccountJpaResource.class

src/main/resources/META-INF/persistence.xml
ckalista: Agree on that.

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
rauligs: It is removed, the renaming will be done as part of PP-576

src/main/java/uk/gov/pay/connector/dao/JpaDao.java
rauligs: Agree, I'll delete only the "remove" methods, as part of the refactoring the others will be tackled in PP-576.

src/main/resources/META-INF/persistence.xml
rauligs: No package scanning, it scans for all the sources of the project, added as part of having weird behaviours finally due to cache issues. Removed in favour of scan sources.

src/test/java/uk/gov/pay/connector/it/dao/EventDaoJpaTest.java
rauligs: Done

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
rauligs: Agreed! Method changed.

src/main/java/uk/gov/pay/connector/dao/GatewayAccountJpaDao.java
rauligs: Added to PP-576

src/main/java/uk/gov/pay/connector/resources/GatewayAccountJpaResource.java
rauligs: Yes, changed.

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
rauligs: Done

src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
rauligs: Done

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
rauligs: Done

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
rauligs: Done

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
rahul-vinayak: EntityManager need not be flushed if its managing the entities, and depending upon the cache strategy it will be either flushed or cache updated as it leaves the method boundary, so there shouldn't be a need to do explicit flushes, hopefully to be addressed in refactoring story

ckalista:
rebuild

src/main/java/uk/gov/pay/connector/dao/ChargeSearchQuery.java
rahul-vinayak: So like we discussed the resource will be building the SearchQuery object and passing on to the DAO instead of the individual search params as it is doing currently...

PR #106, opened 2016-03-03
PP-DB_MIGRATION_FAILURED_AFTER_DROPWIZARD_UPGRADE

imranbohoran:
- The DB migrations are flooding build logs with warnings since the dropwizard upgrade. This appears to be because the 0.9.2 dropwizard-migrations library uses the 3.4.1 version of liquibase-core.
- To avoid these warnings, we are excluding maven using 3.4.1 of liquibase-core for 0.9.2 dropwizard-migrations and then using 3.3.5 of liquibase-core instead.
- There were discussions about perhaps not using liquibase in the future, so this seems a sufficient fix for now


bazbremner:
Makeshifts last longest - I'd much rather run with the version of Liquibase that ships with Dropwizard, rather than go off menu.

What is the actual problem that Liquibase 3.4.x is complaining about and how difficult is that to fix?

imranbohoran:
We haven't spent a lot of time working out the root cause of our migrations throwing warnings with 3.4.1 of liquibase-core. In fact we tried the latest version of liquibase-core (3.4.2), which fails the migration. So its either the migration xml (which we haven't found what it could be or its the dropwizard integration with liquibase 3.4.\* resulting this behaviour.
We have looked at our migration xml to see if there's anything obvious that's wrong and might not be compatible (based on this - https://github.com/tananaev/traccar/issues/1674). But there's nothing in our migration.xml that we've come across. So it could be a combination of dropwizard+liquibase versions that's causing this.

kakumara:
Looks :+1:

PR #107, opened 2016-03-03
PP-576 Refactoring resources

rauligs:
- GatewayAccountResource  - Use new methods provided by JPA daos.
- ChargesResource - Use new methods provided by JPA daos.
- ChargeEventsResource - Extracted from ChargesResource and changed to use new JPA dao methods.
- SecurityTokensResource - Use new methods provided by JPA daos (and refactored TokenDao)
- Removed ChargeEventsListener. This class wasn't a Listener so removed this class and use EventDao instead (since it was a simple delegation).
- Removed 'Jpa' as part of the classNames.


kakumara:
:+1:

PR #109, opened 2016-03-04
PP-438 enable proxy for connector outbound traffic

kakumara:
This completes the code of this CR (there may be Ops part for this)
- Done by optionally setting `ClientProperties.PROXY_URI` in JerseyClient
  Note: 
- Enabling/disabling can be done by env var `HTTP_PROXY_ENABLED` [true | false]
- Added a mockerserver based integration test to test traffic via proxy
- Removed conflicting old servlet-api version (`2.x`) and upgraded to `3.x`

Note:
The reason for doing this using native jersey settings is because, Dropwizard jersey client doesn't seem to work as per http://www.dropwizard.io/0.9.2/docs/manual/configuration.html#proxy where setting the proxy config in client configuration should just work. But that doesn't seem to work during testing. However by setting the native jersey proxy config as per this implementation seems to work

similar problem discussed in here -> https://groups.google.com/forum/#!topic/dropwizard-user/AbDSYfLB17M


src/test/java/uk/gov/pay/connector/it/gatewayclient/ClientFactoryTest.java
```diff
@@ -0,0 +1,95 @@
...
+        mockServer.stop();
+    }
+
+    /**
+     * DropwizardAppWithPostgresRule must be constructed during Test construction,
+     * hence this Parameterised Test and the conditional assertion below.
+     *
+     */
+    @Test
+    public void shouldCreateAJerseyClientWithProxyEnabled() throws Exception {
```
rauligs: Nice! :+1:

rauligs:
:+1:

PR #110, opened 2016-03-04
PP-576 Migrating to JPA - Refactor CardService

rauligs:
```
* Use JPA methods instead of legacy Jdbi ones.
* Improved test coverage.
```


src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -99,4 +99,8 @@ public void setStatus(ChargeStatus status) {
     public void setGatewayTransactionId(String gatewayTransactionId) {
         this.gatewayTransactionId = gatewayTransactionId;
     }
+
+    public boolean isAssociatedTo(String accountId){
+        return this.getGatewayAccount().getId().toString().equals(accountId);
```
rahul-vinayak: Doing a long comparison is preferable than converting to String here, and hence pass the Long all the way from the resource.

src/main/java/uk/gov/pay/connector/service/OrderCaptureRequestBuilder.java
```diff
@@ -35,7 +37,7 @@ public OrderCaptureRequestBuilder withMerchantCode(String merchantCode) {
     }
 
     public OrderCaptureRequestBuilder withTransactionId(String transactionId) {
-        this.transactionId = transactionId;
+        this.transactionId = firstNonNull(transactionId, "");
```
rahul-vinayak: I see what this is doing but we know second arg is never null so why to check it, 
I would rather use, but its very minor:
`this.transactionId=defaultString(transactionId);`

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
rauligs: Yes but that is not covered just yet it will come in a later pull request (that is one of the pending TODOS not covered as part of this).

src/main/java/uk/gov/pay/connector/service/OrderCaptureRequestBuilder.java
rauligs: I didn't know about that method! :) Changed!

rahul-vinayak:
Good to go :+1:

PR #111, opened 2016-03-05
Pp 543/jpa transaction around auth

ckalista:
- Add transaction boundaries
  - Remove session-per-http-request filter which provides the wrong transactional strategy for our MicroService endpoints as typically we want more control over the transaction boundaries than having a single global one spanning over the entire lifecycle of the request.
  - Start the PersistenceService as early as possible with a PersistenceServiceInitialiser injected once the injector is ready.
  - Refactor findChargeForAccount in ChargeDao to be JPA friendly.
  - Refactor CardResource to use the new JPA friendly DOA’s returning entities instead of HashMaps object.
  - Add version attribute in ChargeEntity to enable optimistic locking.
  - Update database schema in migrations.xml to reflect new version attribute in charge table.
  - Fix unit and integration tests.
- Refactor to split code logically into pre/post auth
  - Split the code in 3 logical blocks: preAuthorise, authorise and postAuthorise which represent the transaction boundaries.
  - In preAuthorise, checks are made to ensure the charge internal state is correct prior to sending an authorisation to the Gateway. If all preliminary checks pass, the state is set to AUTHENTICATION_READY.
  - In authorise, carry out the authorisation operation with the provider.
  - In postAuthorisation, further checks ensure the internal state is correct prior to the final update with the authorisation outcome.
  - Unit tests + Integration tests.
- Introduce new state AUTHORISATION_SUCCESS
  - Created new state internal charge status state AUTHORISATION_SUCCESS.
  - Removed unused class SmartpayStatusesBlacklist
  - Deprecate charge state AUTHORISATION_SUBMITTED; to be removed after refactoring PP-543


kakumara:
I guess you mean AUTHORISATION_READY (not AUTHENTICATION_READY as in PR description). Code seems correct.

ckalista:
Indeed, you should read AUTHORISATION_READY in place of AUTHENTICATION_READY.

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
```diff
@@ -94,6 +92,24 @@ public String saveNewCharge(String gatewayAccountId, Map<String, Object> charge)
...
+                        + "join GatewayAccountEntity ga "
+                        + "where c.id = :chargeId "
+                        + "and ga.id = :accountId", ChargeEntity.class);
+
+        query.setParameter("chargeId", chargeId);
+        query.setParameter("accountId", accountId);
+
+        try {
+            return Optional.ofNullable(query.getSingleResult());
+        } catch (NoResultException e) {
```
kakumara: As per API there are other possible exceptions from this. Out of which `NonUniqueResultException` and `QueryTimeoutException` might be something we need to worry about?

rauligs: This is a task being done as part of PP-576. Expect it not to be there when PP-576 is finished.

src/main/java/uk/gov/pay/connector/dao/IJpaDao.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.connector.dao;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public interface IJpaDao {
```
kakumara: I'm not sure If this interface provides a purpose. Nobody will use it directly.

src/main/resources/migrations.xml
```diff
@@ -119,4 +119,12 @@
                       tableName="charges"/>
     </changeSet>
 
+    <changeSet id="add version to charges table" author="">
+        <addColumn tableName="charges">
```
kakumara: You might want to consider a default=0 to this column. In fact it might be better to do at next to the `@Version` annotation so that jpa is aware of this.
e.g. 

`````` @Version
    @Column(name = "version", columnDefinition = "integer DEFAULT 0", nullable = false)```
``````

File: src/main/java/uk/gov/pay/connector/service/CardService.java [hunk 4]
```diff
@@ -23,74 +23,114 @@
...
                 .findById(chargeId)
-                .map(authoriseFor(chargeId, cardDetails))
+                .map(doAuthorise)
                 .orElseGet(chargeNotFound(chargeId));
     }
 
-    public Either<GatewayError, GatewayResponse> doCapture(String chargeId) {
+    @Transactional
+    public Either<GatewayError, ChargeEntity> preAuthorise(ChargeEntity charge) {
+        ChargeEntity reloadedCharge = chargeDao.merge(charge);
```
kakumara: I wonder if this is bit dangerous. The merge performs an update and creates a new managed copy of the entity (which came out of a transactional context). What if another thread (just before the merge) has moved the state to AUTH_READY this will override it again to ENTERING_CARD_DETAILS isn't it? Wouldn't it be better to getCharge here inside the Transaction ?

src/main/java/uk/gov/pay/connector/service/CardService.java
```diff
@@ -23,74 +23,114 @@
...
+    private Either<GatewayError, AuthorisationResponse> authorise(ChargeEntity charge, Card cardDetails) {
+        PaymentProvider paymentProvider = paymentProviderFor(charge);
+        AuthorisationRequest request = authorisationRequest(charge.getId(), charge.getAmount(), cardDetails);
+        AuthorisationResponse response = paymentProvider.authorise(request);
+        return right(response);
+    }
+
+    @Transactional
+    public Either<GatewayError, GatewayResponse> postAuthorise(ChargeEntity charge, AuthorisationResponse response) {
+        ChargeEntity reloadedCharge = chargeDao.merge(charge);
```
kakumara: This `merge()` might have the same issue above.

src/main/java/uk/gov/pay/connector/service/CardService.java
```diff
@@ -23,74 +23,114 @@
...
+                        reloadedCharge.getId())));
+            }
+            logger.error(format("Charge with id [%s] and with status [%s] should be in [ENTERING CARD DETAILS] for authorisation.",
+                    reloadedCharge.getId(), reloadedCharge.getStatus()));
+            return left(illegalStateError(format("Charge not in correct state to be processed, %s", reloadedCharge.getId())));
+        }
+        reloadedCharge.setStatus(AUTHORISATION_READY);
+
+        return right(reloadedCharge);
+    }
```
kakumara: In an unlikely event of `OptimisticLockException` during change state should we consider a specific error instead of internal server error? (e.g. 409)

src/main/resources/migrations.xml
kakumara: Need to make sure the existing records in integration environments gets populated with a default version, or things may break for existing records. (populate with default `0`)

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -45,11 +45,18 @@
     @Convert(converter = UTCDateTimeConverter.class)
     private ZonedDateTime createdDate;
 
-    protected ChargeEntity() {
+    @Version
+    @JsonIgnore
+    @Column(name = "version")
+    private Long version;
```
kakumara: version can go the `AbstractEntity` so that all entities by default will get optimistic locking. But I can understand it can remain here for this PR until other entities are fully migrated to JPA.

src/main/java/uk/gov/pay/connector/service/CardService.java
rahul-vinayak: Thats the whole idea of using versioning on ChargeEntity was, due to the versioning of ChargeEntity the above scenario can not happen as there will be an OptimisticLockException thrown before second thread tries to update it again.

src/main/java/uk/gov/pay/connector/service/CardService.java
rahul-vinayak: same explanation as above

src/main/java/uk/gov/pay/connector/model/GatewayErrorType.java
```diff
@@ -2,11 +2,14 @@
...
+    IllegalStateError,
     GenericGatewayError,
     UnexpectedStatusCodeFromGateway,
     MalformedResponseReceivedFromGateway,
     GatewayUrlDnsError,
     GatewayConnectionTimeoutError,
     GatewayConnectionSocketError,
-    ;
+
+    OperationAlreadyInProgress
```
rahul-vinayak: enums in UPPER_SNAKE_CASE please following the conventions.

src/main/java/uk/gov/pay/connector/service/CardService.java
kakumara: spoke to @ckalista and heard this scenario tested manually and works. :+1:

src/main/java/uk/gov/pay/connector/service/CardService.java
rahul-vinayak: Probably not, cause throwing a 409 conflict response can be interpreted by consumer as fetch the latest data and do the modification again, which is not what should happen here, but may be something for a debate.

src/main/java/uk/gov/pay/connector/dao/IJpaDao.java
ckalista: Agreed this should be removed just like the other interfaces that were introduced just to help during the refactoring.

src/main/resources/migrations.xml
ckalista: Agreed, I have actually encountered that issue while testing.

src/main/java/uk/gov/pay/connector/dao/IJpaDao.java
rauligs: Will be removed as part of PP-576 after this is merged

src/main/java/uk/gov/pay/connector/service/CardService.java
ckalista: We should indeed! After debating about it it appears that 409 could be a good error status code for this type of exception.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
ckalista: This will be addressed by the migration to JPA user story.

src/main/java/uk/gov/pay/connector/model/GatewayErrorType.java
ckalista: :+1:

src/main/java/uk/gov/pay/connector/dao/ChargeJpaDao.java
ckalista: :+1: I will let this been addressed by PP-576.

src/main/java/uk/gov/pay/connector/dao/IJpaDao.java
ckalista: :+1: I won't change here, will leave it for PP-576.

src/main/java/uk/gov/pay/connector/model/GatewayErrorType.java
ckalista: Done, now following project convention.

src/main/java/uk/gov/pay/connector/service/CardService.java
ckalista: As discussed during the transaction huddle, this scenario is now handled by returning a 409 error.

src/main/java/uk/gov/pay/connector/service/CardService.java
ckalista: In this scenario we still want to throw an internal error hence no change required.

src/main/resources/migrations.xml
ckalista: Added default 0 to the column definition for version field in liquidbase.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
rauligs: Yes will be there in PP-576

ckalista:
rebuild

kakumara:
:+1:

PR #113, opened 2016-03-08
PP-438 pass HTTP proxy gubbins to JVM

bazbremner:
PCI 1.3 prohibits direct access between the internet and anything in the
cardholder data environment, which the connector is a key part of.

I believe, but haven't yet proved, that it's just the JVM proxy settings
that are required; I don't know if there's separate configuration for
any client library used in the codebase.

Ansible, which is currently used to drive the deployment of containers
should set the HTTP__PROXY_HOST, HTTP__PROXY_PORT and NON_PROXY_HOSTS
variables in a format required by the JVM.

https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html
for all the painful details.


bazbremner:
rebuild

bazbremner:
Closing for now - this isn't actually required at the moment and I'm concentrating on other areas.

PR #114, opened 2016-03-09
PP-596 Disable external DTD validation

rauligs:
Disable external DTD validation when we unmarshall incoming XMLs from payment providers:
- Changed JAXB to use SAX parser to be able to set new properties.
- Added features to disable DTD validations.
- Set namespaceAware to true to be able to parse SOAP XMLs.


bazbremner:
Looks like a very clear change; nice test case too. Merging; let's see how it fares in a prod-like environment :)

PR #115, opened 2016-03-09
PP-590 Connector spamming logs with JDBC operations

imranbohoran:
Our logs are been spammed by JDBC operations due to the eclipselink log configurations. 
These configurations are now moved to the config.yml and the log levels can now be toggled through environment variables.

The available log levels for eclipselink are at - https://wiki.eclipse.org/EclipseLink/Examples/JPA/Logging

This change set also contains a change as to how the `EnvironmentVariableSubstitutor` instance is created. A `false` flag is passed into this, which makes it _non-strict_ so that we can have defaults in the config.yml (http://www.dropwizard.io/0.9.2/docs/manual/core.html#environment-variables). By making this _non-strict_ we are able to provide the default log levels, which then could be overridden by environment variables.


bazbremner:
Can you use words Rather_Than_A_Snakey_Camel please?

Also, as I've mentioned to other members of the team (and I feel like there's a blog post or similar required reading): please put the information you're putting in the PRs into the commit messages - they live closer to the code and generally survive longer.

PR comments tend to get harder and harder to find with time (as I'm discovering trying to find my previous comments on this matter).

If you write a good commit message, the PR UI will lift the text from it and it'll live on in the code base. The opposite is not true for PR comments.

imranbohoran:
@bazbremner When you say use words, are you referring to the PR title in this case?

I thought adding some additional description about the `EnvironmentVariableSubstitutor` in this PR comment in as just extra information. I was expecting the boolean passed in as a named variable was sufficient in the code and the commit message was covering what was done. I can amend the commits to have the same if necessary. But if you feel the commit message is not clear enough, happy to change it have more information.

bazbremner:
@imranbohoran aye, I meant in the title of the PR (which, more importantly) also the first line of your commit message.

The information you've provided in the PR is great, but it's not present in your commit message(s).

When debugging issue and spelunking through a codebase, the commits are invaluable - you're already going to the effort to give good context, just capture it in _commits_, please.

imranbohoran:
@bazbremner no problem. will do. The e2e tests seem to be broken (actually for all branches with PRs for connector at the moment). I'll have a look at that and see there's anything else that's needed and then will change the commits.

imranbohoran:
@bazbremner I've added more details into the commit message from the PR message.

kakumara:
Looks okay :+1:

PR #116, opened 2016-03-11
PP-576 Migrating to JPA - Clean up

rauligs:
```
 * Removed JPA interfaces in favour of classes.
 As consequence ConnectorModule doesn't need to be declared explicitly.
 * Refactored Daos to remove legacy methods. Some methods are being override (such as merge and persist in ChargeDao since Notifications are not in listeners just yet, parto of another story)
 * Changed naming removing 'Jpa' as part of the names of the classes. Only super Class of the daos contains Jpa as literal since is implementation specific.
 * Refactored ChargeResource.
```


src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -23,12 +22,6 @@ public ConnectorModule(final ConnectorConfiguration configuration, final Environ
     protected void configure() {
         bind(ConnectorConfiguration.class).toInstance(configuration);
         bind(Environment.class).toInstance(environment);
-
```
ckalista: :+1:

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.persistence.EntityManager;
+import javax.persistence.TypedQuery;
+import java.time.LocalDateTime;
+import java.util.List;
+import java.util.Optional;
+
+import static uk.gov.pay.connector.model.domain.ChargeStatus.CREATED;
+
+@Transactional
+public class ChargeDao extends JpaDao<ChargeEntity> {
```
kakumara: +1 for Typed Daos

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -119,7 +99,7 @@ public void setGatewayAccount(GatewayAccountEntity gatewayAccount) {
         this.gatewayAccount = gatewayAccount;
     }
 
-    public boolean isAssociatedTo(String accountId){
-        return this.getGatewayAccount().getId().toString().equals(accountId);
+    public boolean isAssociatedTo(Long accountId) {
+        return this.getGatewayAccount().getId().compareTo(accountId) == 0;
```
kakumara: :) seriously?

src/main/java/uk/gov/pay/connector/model/domain/AbstractEntity.java
```diff
@@ -22,6 +22,14 @@ public AbstractEntity(Long id) {
     @JsonIgnore
     private Long id;
 
+    @Version
```
ckalista: :+1:

src/main/java/uk/gov/pay/connector/resources/ChargesResource.java
```diff
@@ -67,28 +59,24 @@
...
-        this.tokenDao = tokenDao;
         this.gatewayAccountDao = gatewayAccountDao;
         this.chargeService = chargeService;
-        this.linksConfig = configuration.getLinks();
     }
 
     @GET
     @Path(CHARGE_API_PATH)
     @Produces(APPLICATION_JSON)
     public Response getCharge(@PathParam("accountId") String accountId, @PathParam("chargeId") String chargeId, @Context UriInfo uriInfo) {
```
ckalista: Should we not change the PathParam of accountId instead to expect a Long and let Jersey deal with the conversion? In case of any problem during .valueOf() we end up with a 500 error which is probably not what we want.

Same comment goes for the other parameters converter from string to Long.

src/test/java/uk/gov/pay/connector/fixture/ChargeEntityFixture.java
```diff
@@ -0,0 +1,77 @@
+package uk.gov.pay.connector.fixture;
```
ckalista: :+1:

ckalista:
Code review completed. Just a small comment worth discussing.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
rauligs: Ups

src/main/java/uk/gov/pay/connector/resources/ChargesResource.java
rauligs: Completely agree!
I have changes in a separated branch with tests to be fixed by this change (only for ids but charge_id since charge_id will be changed when PP-572 is played).
So I'll create a separate pull request with these changes.

kakumara:
Looks ok :+1:

PR #117, opened 2016-03-11
PP-193 New admin task endpoint to expire charges.

rahul-vinayak:
- Added a new dao method to query charges before a date and by charge
  statuses.
- added a new drop wizard task that could be called from the admin port.
  which will process expired charges.
- added tests around dao and the IT for the task
- Logic to send a cancel request for charges with AUTH_SUCCESS state
  and handling the success and failure responses for the cancellation.
- Updated README

with @imranbohoran and @PauloPortugal


rahul-vinayak:
rebuild

PR #118, opened 2016-03-14
PP-193 Introducing a admin task endpoint to expire charges

rahul-vinayak:
- Added a new dao method to query charges before a date and by charge statuses.
- Added a new drop wizard task that could be called from the admin port. which will process expired charges.
- Added tests around dao and the IT for the task
- Logic to send a cancel request for charges with AUTH_SUCCESS state and handling the success and failure responses from the providers for the cancellation.
- Updated README for the new admin task endpoint


README.md
```diff
@@ -83,6 +88,26 @@ The command to run all the tests is:
 |[```/v1/frontend/tokens/{chargeTokenId}```](#get-v1frontendtokenschargetokenid)                                  | GET |  Retrieve information about a secure redirect token.            |
 |[```/v1/frontend/tokens/{chargeTokenId}```](#delete-v1frontendtokenschargetokenid)                                  | DELETE |  Delete the secure redirect token.            |
 
+### POST /tasks/charges-expiry-sweep
+
+This end point is accessible only through the admin port (9301 in case of docker).
+
+This starts the task to expire the charges with a default window of 1 Hr. The default value can be over-ridden by setting an environment variable CHARGE_EXPIRY_WINDOW_SECONDS in seconds. 
+
+Also can be overridden by adding a query param "CHARGE_EXPIRY_WINDOW_SECONDS=xx", mainly intended for testing, xx being numeric value in seconds (i.e. /tasks/charges-expiry-sweep?CHARGE_EXPIRY_WINDOW_SECONDS=60)
```
kakumara: perhaps `snake_case` here for query param

src/test/java/uk/gov/pay/connector/it/task/ChargeExpiryTaskITest.java
```diff
@@ -0,0 +1,40 @@
...
+
+    private RestAssuredClient chargeExpiryClient = new RestAssuredClient(app, ACCOUNT_ID);
+
+    @Before
+    public void setupGatewayAccount() {
+        app.getDatabaseTestHelper().addGatewayAccount(ACCOUNT_ID, PROVIDER_NAME);
+    }
+
+    @Test
+    public void shouldGetSuccessWhenPostChargeExpiryTask() throws Exception {
```
rauligs: Being an integration test, what about checking the status of the row in the database?

src/main/java/uk/gov/pay/connector/task/ChargeExpiryTask.java
```diff
@@ -0,0 +1,97 @@
...
+    @Inject
+    public ChargeExpiryTask(ChargeDao chargeDao, CardService cardService) {
+        super(EXPIRED_CHARGES_SWEEP);
+        this.chargeDao = chargeDao;
+        this.cardService = cardService;
+    }
+
+    @Override
+    @Transactional
+    public void execute(ImmutableMultimap<String, String> immutableMultimap, PrintWriter printWriter) throws Exception {
```
rauligs: If the inmutableMultimap represents something I would change the name to whatever it is (is it request parameters I guess)

src/test/java/uk/gov/pay/connector/it/task/ChargeExpiryTaskITest.java
```diff
@@ -0,0 +1,40 @@
...
+        String chargeId = ((Integer) RandomUtils.nextInt(99999999)).toString();
+        app.getDatabaseTestHelper().addCharge(chargeId, ACCOUNT_ID, AMOUNT, AUTHORISATION_SUCCESS, "/something", null);
+
+        chargeExpiryClient.postAdminTask("", EXPIRED_CHARGES_SWEEP)
```
rauligs: Only used in one place and body is empty, why no inline it?

src/main/java/uk/gov/pay/connector/task/ChargeExpiryTask.java
```diff
@@ -0,0 +1,97 @@
...
+        logger.info(format("%s charges expired out of total %s charges found", expiredCharges.size(), charges.size()));
+    }
+
+    private int getChargeExpiryWindowSeconds(ImmutableMultimap<String, String> immutableMultimap) {
+        //default expiry window, can be overridden by env var, which can be further overridden by query param for ease of testing
+        int chargeExpiryWindowSeconds = ONE_HOUR;
+        if (StringUtils.isNotBlank(System.getenv(CHARGE_EXPIRY_WINDOW))) {
+            chargeExpiryWindowSeconds = Integer.parseInt(System.getenv(CHARGE_EXPIRY_WINDOW));
+        }
+        ImmutableList<String> expiryWindowQueryParam = immutableMultimap.get(CHARGE_EXPIRY_WINDOW).asList();
```
rauligs: For a task maybe is not a big deal but creating/modifying production code for ease of testing is not a recommended approach in my opinion, more because if code is driven by tests it should be already design to be testable. Unless this is already an AC which would made the comment unnecessary.

src/main/java/uk/gov/pay/connector/task/ChargeExpiryTask.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Inject
+    public ChargeExpiryTask(ChargeDao chargeDao, CardService cardService) {
+        super(EXPIRED_CHARGES_SWEEP);
+        this.chargeDao = chargeDao;
+        this.cardService = cardService;
+    }
+
+    @Override
+    @Transactional
```
rauligs: _Transactional_ for the whole lot is too much, or is this temporary?

src/main/java/uk/gov/pay/connector/task/ChargeExpiryTask.java
```diff
@@ -0,0 +1,97 @@
...
+        if (expiryWindowQueryParam.size() > 0 && StringUtils.isNotBlank(expiryWindowQueryParam.get(0))) {
+            chargeExpiryWindowSeconds = Integer.parseInt(expiryWindowQueryParam.get(0));
+        }
+        return chargeExpiryWindowSeconds;
+    }
+
+    private boolean filterExpirableCharges(ChargeEntity ce) {
+        if (ce.getStatus().equals(AUTHORISATION_SUCCESS.getValue())) {
+            logger.debug(format("sending a cancel request for charge ID: %s", ce.getId()));
+            Either<GatewayError, GatewayResponse> cancelResponse = cardService.doCancel(ce.getId(), ce.getGatewayAccount().getId());
```
kakumara: This is a remote call (over the wire) and looks like its within a Transaction boundary. This could be dangerous as this is a batch job with possibly many calls/locks.

src/main/java/uk/gov/pay/connector/task/ChargeExpiryTask.java
imranbohoran: The only reason this is added is so that we can have e2e tests that will expire charges. There are other complicated ways to do this, but we felt that this is a bit more sensible.

src/main/java/uk/gov/pay/connector/task/ChargeExpiryTask.java
imranbohoran: Given that this is a resource that will be invoked by a cron (without multiple requests hitting it) we felt it didn't warrant more complex transaction suspension-resumption logic. We should ideally take another look at usage of @Transactional annotations as it might be worthwhile doing more programmatic transaction handling without depending on the annotations. But that's a different matter.

imranbohoran:
Based on our team discussion today, we'll be addressing some of these comments in accordance to our discussions. We'll capture the decisions in the story as well.

imranbohoran:
Closing this PR as we are addressing changing few things on how this is done. Latest discussions and decisions are captured in the story in Jira.

PR #119, opened 2016-03-14
PP-572 add new random charge_id column to charges table

kakumara:
The purpose of this column is to store the new unguessable random id to a charge
Initially keeping it `null` until the data migration is complete for existing data.
Using a 24^36 random String generator for ids.

with @rauligs


src/main/java/uk/gov/pay/connector/util/RandomIdUtil.java
```diff
@@ -0,0 +1,23 @@
...
+
+    public static final String DICTIONARY = "abcdefghijklmnopqrstuvwxyz0123456789";
+    private static final int ID_LENGTH = 24;
+
+    /**
+     * This method will generate a URL safe random string.
+     * The randomness strength is 24^36 (which should be good enough for our ID needs)
+     * @return
+     */
+    public static String newId() {
```
PauloPortugal: Why not use RandomStringUtils.randomAlphanumeric(ID_LENGTH)?
or 
UUID.randomUUID().replace("-", "");

src/main/resources/migrations.xml
```diff
@@ -151,4 +152,17 @@
...
 
+    <changeSet id="add randomized charge id to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="charge_id" type="char(24)">
+                <constraints nullable="true" unique="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="migrate existing id data" author="">
```
PauloPortugal: :+1:

PauloPortugal:
:+1: great solution

PR #120, opened 2016-03-16
PP-572 Propagate external_id for charges up to resources

rauligs:
**This needs to be merged after  _https://github.com/alphagov/pay-endtoend/pull/113_**
- Resources with charge ids as a parameter or payload property has been changed to String value for the obfuscated value.
- Added dao method to retrieve charge by this obfuscated id represented in a new column called external_id


PauloPortugal:
Looks good :+1:

PR #121, opened 2016-03-16
PP-572 Clean up changes for charge_id as obfuscated id

rauligs:
```
* Changed tests so only daos have acknowledge of an external id
```


PauloPortugal:
:+1:

PR #122, opened 2016-03-16
Pp 546/jpa transaction around capture

ckalista:
- Split CardService in separate class respectively to the type of operation: Authorise, Capture and Cancel.
- New error status codes for authorisation and capture.
  - New error status code specific for authorisation, AUTHORISATION_ERROR (deprecating the former SYSTEM_ERROR which was too general purpose).
  - New error status code specific for capture, CAPTURE_ERROR (deprecating CAPTURE_UNKNOWN).
  - Add status to CaptureResponse for consistency with AuthorisationResponse.
- Apply transaction boundaries around capture
  - Rename status READY_FOR_CAPTURE to CAPTURE_READY so to be more consistent with other READY statuses.
  - Apply the same pattern used for authorisation i.e. preCapture, capture and postCapture blocks.
  - Split CardResourceITest into CardCaptureResourceITest and CarAuthoriseResourceITest


ckalista:
rebuild

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.function.Function;
+
+import static fj.data.Either.left;
+import static fj.data.Either.right;
+import static java.lang.String.format;
+import static uk.gov.pay.connector.model.GatewayError.*;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_READY;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
+
+public class CardAuthoriseService extends CardService {
```
imranbohoran: It's nice we break the responsibilities into in their own classes. :+1: However, why do we need to create a hierarchy of services by extending the CardService. Can't we get rid of the CardService?

ckalista: Agreed - this is a bigger refactoring planned to be carried out with user story PP-607.

PR #125, opened 2016-03-21
PP-567 Fixing non deterministic tests

imranbohoran:
Tests keep failing non deterministically when the accept header is not set with the content type we are expecting.


PauloPortugal:
:+1:

PR #127, opened 2016-03-22
PP-568 Place JTA transactions around Notifications

PauloPortugal:
- Stop using the ChargeDao in favour of ChargeService, as
  is already has Transactional methods

with @imranbohoran


src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -74,6 +74,10 @@ public ChargeResponse create(Map<String, Object> chargeRequest, GatewayAccountEn
                 });
     }
 
+    public Optional<ChargeEntity> findByProviderAndTransactionId(String provider, String transactionId) {
```
ckalista: Generally there are very few use cases for pass-thru methods; they add no value and contribute to clutter the code. I can see why this was done here though I don't see why in this case we couldn't inject the DAO directly in the Resource and call findByProviderAndTransactionId from there?

ckalista:
I don't really see the advantage of doing this guys - the service layer adds unnecessary complexity in this case as it merely acts as a pass through to DAO without adding any business logic for this two methods.

src/main/java/uk/gov/pay/connector/service/ChargeService.java
PauloPortugal: As explained at this point it was the case of avoiding having in the same contructor a ChargeService and a ChargeDAO at the expense of having a more spaghetti code. We thought that a pass through method was not that bad.

The "chargeService.updateStatus" feedback was more interesting because that is a smell, as we should not create specific DAO setters for individual attributes, but instead update the entity and send them to a more generic DAO, although the "ChargeDao.mergeAndNotifyStatusHasChanged" is another smell, but as we discussed it should be removed in the near future.

ckalista: :+1:

PR #128, opened 2016-03-22
Pp 607/place transactions around cancel

ckalista:
- Common interface GatewayRequest
  - New common interface for AuthorisationRequest, CancelRequest and CaptureRequest.
  - getPaymentProviderFor method in CardService abstract class to retrieve the provider for a given charge entity.
- CardService abstract class generalise preOperation
  - Introduce new charge statuses CANCEL_READY and CANCEL_ERROR.
  - Tighten up CancelResponse to force the object to be created with the ErrorResponse and ChargeStatus set.
  - Reinstate CardService abstract class which introduces a hierarchy between CardAuthoriseService, CardCancelService and CardCaptureService with the purpose to share common preOperation logic.
  - Move hasStatus method to ChargeEntity where it belongs.
  - Map CANCEL_READY as IN_PROGRESS external status.
  - Map CANCEL_ERROR as FAILED external status.
- Refactor CardService abstract class into TransactionalGatewayOperation Interface
  - Refactor CardService abstract class into an Interface called TransactionalGatewayOperation with has one default method implementation executeGatewayOperation taking care of orchestrating the operations: preOperation, operation and postOperation.
  - Refactor CardAuthorisationService, CardCaptureService and CardCancelService to comply to the new TransactionalGatewayOperation interface.


imranbohoran:
rebuild

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -88,7 +88,7 @@ public void persist(ChargeEntity chargeEntity) {
         return query.getResultList();
     }
 
-    public void mergeAndNotifyStatusHasChanged(ChargeEntity chargeEntity) {
+    public void notifyStatusHasChanged(ChargeEntity chargeEntity) {
```
rauligs: Why this change? is it not merging too?
reading notifyStatusHasChange means to me only notify (and not update) 
(Maybe the super.merge is not necessary since the transaction boundary changed?)

src/main/java/uk/gov/pay/connector/model/AuthorisationRequest.java
```diff
@@ -4,7 +4,7 @@
 import uk.gov.pay.connector.model.domain.ChargeEntity;
 import uk.gov.pay.connector.model.domain.GatewayAccountEntity;
 
-public class AuthorisationRequest {
+public class AuthorisationRequest implements GatewayRequest {
```
rauligs: :+1:

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
```diff
@@ -2,106 +2,65 @@
...
-                    if (postCapture.isLeft())
-                        return left(postCapture.left().value());
-
-                    return right(captured.right().value());
-                };
-
         return chargeDao
                 .findByExternalId(chargeId)
-                .map(doCapture)
+                .map(TransactionalGatewayOperation.super::executeGatewayOperationFor)
```
rauligs: :+1:

src/main/java/uk/gov/pay/connector/service/TransactionalGatewayOperation.java
```diff
@@ -0,0 +1,45 @@
...
+import uk.gov.pay.connector.model.domain.ChargeEntity;
+
+import javax.persistence.OptimisticLockException;
+
+import static fj.data.Either.left;
+import static fj.data.Either.right;
+import static java.lang.String.format;
+import static uk.gov.pay.connector.model.ErrorResponse.conflictError;
+
+interface TransactionalGatewayOperation {
```
rauligs: - This is clearly a Template Method pattern and it seems the intention is not for executeGatewayOperationFor to be override. It looks to me that _TransactionalGatewayOperation_ should be an abstract class and _executeGatewayOperationFor_ a public final method (a template method).

src/test/java/uk/gov/pay/connector/unit/service/CardCancelServiceTest.java
```diff
@@ -31,24 +31,26 @@ public void shouldCancelACharge() throws Exception {
         Long accountId = 1L;
 
         ChargeEntity charge = createNewChargeWith(chargeId, ENTERING_CARD_DETAILS);
+        ChargeEntity reloadedCharge = mock(ChargeEntity.class);
 
-        when(mockedChargeDao.findByExternalIdAndGatewayAccount(charge.getExternalId(), accountId)).thenReturn(Optional.of(charge));
-        when(mockedAccountDao.findById(charge.getGatewayAccount().getId())).thenReturn(Optional.of(charge.getGatewayAccount()));
+        when(mockedChargeDao.findByExternalIdAndGatewayAccount(charge.getExternalId(), accountId))
+                .thenReturn(Optional.of(charge));
+        when(mockedChargeDao.merge(any()))
```
rauligs: I do prefer not the use of _any()_ but to know which element is being passed (could be done in a verify too). This applies anywhere where _any()_ is being used.

src/main/java/uk/gov/pay/connector/service/TransactionalGatewayOperation.java
ckalista: That was also what I was leaning towards, in fact this is how it was implemented first - Imran has a different view point regarding this and has a few arguments against it - maybe have a quick chat with him about it. Personally I am happy either way.

rauligs: Anyway is much much cleaner and nice to see! :+1: 
I'm happy to merge with this approach and have a discussion if necessary afterwards.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
ckalista: Indeed the merge should not be needed although looking into this it appears that there is still one place where it is necessary - this could be refactored however we decided not to spend time in this since the Auditing is going to be improved in the near future making this whole method redundant.

Renaming back to its original name for clarity.

src/test/java/uk/gov/pay/connector/unit/service/CardCancelServiceTest.java
ckalista: Updated accordingly.

ckalista:
rebuild

PR #129, opened 2016-03-29
PP-554 Improve handling of Worldpay notifications

imranbohoran:
Improving how we handle worldpay notifications. The decisions on how to handle the notifications are captured in the card - https://payments-platform.atlassian.net/browse/PP-554.
- This change avoids updating the charge status twice and instead only updates the charge status based on the status received from the inquiry of the payment status from worldpay.
- Additional logging added where necessary.


rahul-vinayak:
looks good to go :+1:

PR #130, opened 2016-03-31
PP 486 Allow next_url to be a POST url

rauligs:
- Added _next_url_post_ link to charge response payload, so clients have an extra option to integrate with _Pay_


src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -35,13 +37,14 @@
...
     private static final Logger logger = LoggerFactory.getLogger(ChargeService.class);
+
     public static final String EXPIRY_SUCCESS = "expiry-success";
     public static final String EXPIRY_FAILED = "expiry-failed";
 
-    ChargeDao chargeDao;
-    TokenDao tokenDao;
-    LinksConfig linksConfig;
-    CardCancelService cardCancelService;
+    private ChargeDao chargeDao;
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -163,15 +169,16 @@ private URI selfUriFor(UriInfo uriInfo, Long accountId, String chargeId) {
...
+                .queryParam("chargeTokenId", tokenId)
+                .build();
+    }
+
+    private URI nextUrlPost(String chargeId) {
+        return UriBuilder.fromUri(linksConfig.getCardDetailsBaseUrl())
+                .path(chargeId)
+                .build();
     }
 }
```
imranbohoran: :+1:

src/test/java/uk/gov/pay/connector/matcher/RegexMatcher.java
```diff
@@ -0,0 +1,31 @@
...
+        Pattern compiledPattern = Pattern.compile(pattern);
+        Matcher matcher = compiledPattern.matcher(input);
+        return matcher.matches();
+    }
+
+    @Override
+    public void describeTo(Description description) {
+        description.appendText("A String matching pattern: " + pattern);
+    }
+}
```
imranbohoran: I was fairly certain hamcrest had a regex matcher. Looking at why we can't find it, it looks we are using the hamcrest-library. That doesn't have the pattern matcher. So instead if we use the java-hamcrest library, we would have access to the regex pattern matcher. 
        <dependency>
            <groupId>org.hamcrest</groupId>
            <artifactId>java-hamcrest</artifactId>
            <version>2.0.0.0</version>
        </dependency>

rauligs: We can add it If it becomes very used and this doesn't fit our requirements, but take a whole testing library for one matcher maybe can't be done in other time when needed. For our purposes now I think is fine.
As discussed I'll bring into the project the library we encourage ppl to use what is already built and could contain more useful matchers that could be brought to other projects so we also avoid duplication.

PR #131, opened 2016-04-04
PP-548 upgrade govukpay/postgres image version

kakumara:
As part of PP-548 a new databsase was introduced for selfservice and the default postgres image we have been using is upgraded to govukpay specific versioning.


rauligs:
rebuild

kakumara:
rebuild

PR #132, opened 2016-04-04
PP-631 search by date

rauligs:
- Changed query to search charges so that 'to date' is exclusive


rauligs:
rebuild

src/main/resources/migrations.xml
```diff
@@ -179,4 +179,19 @@
         </createIndex>
     </changeSet>
 
+    <changeSet id="createIndex charges.reference" author="">
```
kakumara: 👍

kakumara:
Looks okay

PR #133, opened 2016-04-06
PP-539 Making charge events use ZoneDateTime

imranbohoran:
- Updated date was using LocalDateTime which is not consistent with the
  format used in other entities in the connector.


src/main/java/uk/gov/pay/connector/model/ChargeEvent.java
```diff
@@ -49,12 +42,9 @@ public String getExternalStatusString() {
         return status.getValue();
     }
 
-    public LocalDateTime getUpdated() {
-        return updated;
-    }
-
-    public void setUpdated(LocalDateTime updated) {
-        this.updated = updated;
+    @JsonProperty("updated")
```
PauloPortugal: In the PublicApi we were adding the JsonProperties in the attribute declaration. Could we maintain that consistency?

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventEntity.java
```diff
@@ -15,13 +15,13 @@
     @Convert(converter = ChargeStatusConverter.class)
     private ChargeStatus status;
 
-    @Convert(converter = LocalDateTimeConverter.class)
-    private LocalDateTime updated;
+    @Convert(converter = UTCDateTimeConverter.class)
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/connector/model/ChargeEvent.java
```diff
@@ -49,12 +42,9 @@ public String getExternalStatusString() {
...
+    public String getUpdated() {
+        return DateTimeUtils.toUTCDateString(updated);
```
PauloPortugal: :+1:

PauloPortugal:
:+1:

PR #134, opened 2016-04-06
Pp 565 creates a generic ThreadExecutor service to do async Authorisation

rahul-vinayak:
- Authorisation response returns back a 202 (Accepted) when the gateway response is not achieved back within 10 seconds (configurable) 
- Creates a generic CardExecutorService to execute auth call in an async way, this is generic and can be used to execute other operations.
- The env variable AUTH_READ_TIMEOUT can be passed from env variables to override the default value of 10 seconds, i.e. the timeout before the resource responds with an awaited auth response (202)
- Updated the README with the new env variable description

with @PauloPortugal


README.md
```diff
@@ -33,6 +33,10 @@ Important configurations.
         username: <Worldpay merchantID | this will be removed when gateway accounts have the capability to store username/passwords >
         password: <Worldpay merchat password for integration | this will be removed when gateway accounts have the capability to store username/passwords>
   ```
+  
+#### Environment Variables
+
+`AUTH_READ_TIMEOUT_SECONDS`:  The env variable AUTH_READ_TIMEOUT_SECONDS can be passed into the app to override the default value of 10     seconds, i.e. the timeout before the resource responds with an awaited auth response (202), so that frontend can choose to show a spinner and poll for auth response.
```
imranbohoran: Should we have this in the config.yaml. We seem to have all configurable parameters through the dropwizard configuration approach. This seem to deviate from it. Would it make sense to make this available through dropwizard configuration as well for consistency.

Also is this something that's going to be made available in pay-ansible and/or pay-scripts?

PauloPortugal: Good point @imranbohoran!
We spoke about it and the resolution was that this was an environment variable and that enviroment variables would go in the connector.env file under pay-scripts
Do you mean that we should also make this variable available for configuration in the config.yaml file?

rahul-vinayak: Yes, as @PauloPortugal suggests, its an environment variable and its used purely for the purpose of overriding the defaults, it does not qualify as a config parameter as its not needed for making things work. It will go in scripts/connector.env only if you need to override the default behaviour locally.

imranbohoran: ok - I considered any configurable element of the application to be sourced through config.yaml. I also viewed it as a source of documentation of what kind of configurations that could be done for a given application. If however this is the common understanding of the team, I'm cool with it.

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
```diff
@@ -0,0 +1,56 @@
...
+            return Pair.of(COMPLETED, futureObject.get(timeout, TimeUnit.SECONDS));
+        } catch (InterruptedException | ExecutionException e) {
+            return Pair.of(FAILED, null);
+        } catch (TimeoutException e) {
+            return Pair.of(IN_PROGRESS, null);
+        }
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
```
imranbohoran: Do we really need a finalizer here? It appears we are doing this so that we shutdown the executor and the only time we want that to happen is when the application shuts down (as this is a singleton). Wouldn't the executor be killed with the application shutdown anyway? If we really need executor service shutdown, perhaps a shutdownhook makes sense(?)

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
```diff
@@ -0,0 +1,56 @@
...
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static uk.gov.pay.connector.resources.CardExecutorService.ExecutionStatus.*;
+
+// this service runs the supplied function in a new Thread
+public class CardExecutorService<T> {
+
+    private static final Logger logger = LoggerFactory.getLogger(CardExecutorService.class);
+    private static int timeout = 10;
+    private ExecutorService executor = Executors.newCachedThreadPool();
```
imranbohoran: Perhaps it might be good to use a Queued/Restricted executor. The newCachedThreadPool will keep creating threads if threads are busy and in the rare situation that this might happen could have an affect on the entire jvm. Having a restricted/queued executor will limit the problem to this part of the system and also would allow us to monitor the active threads against the size through jmx.

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -2,40 +2,62 @@
...
         this.cardDetails = cardDetails;
-
-        return chargeDao
-                .findByExternalId(chargeId)
-                .map(TransactionalGatewayOperation.super::executeGatewayOperationFor)
-                .orElseGet(chargeNotFound(chargeId));
+        Optional<ChargeEntity> chargeEntity = chargeDao.findByExternalId(chargeId);
+
+        if (chargeEntity.isPresent()) {
+            Supplier<Either<ErrorResponse, GatewayResponse>> authorisationSupplier = () -> TransactionalGatewayOperation.super.executeGatewayOperationFor(chargeEntity.get());
```
imranbohoran: Why can't this just be `executeGatewayOperationFor(chargeEntity.get())`

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -22,6 +24,8 @@ public ConnectorModule(final ConnectorConfiguration configuration, final Environ
     protected void configure() {
         bind(ConnectorConfiguration.class).toInstance(configuration);
         bind(Environment.class).toInstance(environment);
+        bind(CardExecutorService.class).in(Singleton.class);
```
imranbohoran: Do we need `in(Singleton.class)`. Wouldn't Guice operate on singleton mode by default?

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
imranbohoran: One other suggestion on creating an executor service. Can we build it with a custom ThreadFactory? This will allow us to have meaningful log lines showing what the thread is. By default IIRC, in logs the threads would show as pool-x-thread-y, which becomes useless when diagnosing issues. A ThreadFactory can be passed into the executor service with a thread name which helps with the logs and also thread dumps. 
Our log line format already includes the thread, so having this capability in the executor service will be valuable when we have to diagnose issues.

README.md
PauloPortugal: @imranbohoran  this has been addressed

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
PauloPortugal: Please be more specific

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
rahul-vinayak: The purpose the executor.shutdown is called is it tries to stop all threads running gracefully, so its advisable to do a proper shutdown, but if there is a concern on reliability of the finalize method we can change it to a shutdown-hook (which again isn't 100% reliable to be called in case of JVM crashes), but still can be relied more than the GC finalization.

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
rahul-vinayak: Not sure we need to deal with all that complexity at the moment (KISS + YAGNI) as we are not even doing any thread management or looking for future objects. Our truth will be the charge status at any given point of time. Threads are just making network calls they will either succeed, fail or timeout and DB will be updated accordingly

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
rahul-vinayak: @imranbohoran it can be, but bear in mind its a default method in one of the interfaces and it might lead to clashes if you have more than one interface implemented in this class. It doesn't even need to be a default defined in an interface in the first place, I think.

imranbohoran: You're assuming we might have it implementing another interface with the same method name? There might be other issues in the design if that's true. I'm ok with this, just seemed weird having it this way. It was more of an observation.

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
imranbohoran: Don't think this adds much complexity and I'm fairly certain it will be needed when someone is looking at production logs/ thread dumps. The amount of times I've had to look at threads of a pool (in  logs and dumps) and not knowing what thread is performing what operation made me make this comment. Its creating a new ThreadFactory and passing into the executore service. If you both consider this is not useful then you can skip it.

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
rahul-vinayak: its a shutdown hook in place of finalize now

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
rahul-vinayak: I wasn't too sure about the default injection scope, however even if its the default behaviour I prefer to keep it explicit here as it adds clarity and documents it for sake of readability

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
rahul-vinayak: Ok we will remove it for the sake of argument that a case may not arise that we have the same method names being implemented in favour of better readability.

README.md
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
imranbohoran: 👍 I guess you could also have also just done this.<method_name>, but this is fine

rahul-vinayak: Not clear about this last comment, elaborate please

imranbohoran: We have discussed this and decided to leave it as is.

src/main/java/uk/gov/pay/connector/resources/CardExecutorService.java
rahul-vinayak: As agreed in further discussions, although there are benefits for using Custom ThreadFactory, which are around thread management, setting priorities, and thread naming, the current story doesn't require that level of thread management, for the cost of added code complexity, so leaving it as is till we feel the need to deal with it.

PR #135, opened 2016-04-08
Pp 210/enable resume of payments [REVIEW ONLY]

ckalista:
- Updated SecurityToken resource to return the full charge object for the given token.
- Map ChargeEntity as a many to one in TokenEntity.
- generateNewTokenFor static factory to create a new TokenEntity, generate a random security token and associate it to a given ChargeEntity.
- Remove a lot of code duplication in DAO integration tests with the introduction of DatabaseFixtures that facilitates the creation of test accounts and test charges.
- New ChargeEntity method hasExternalStatus returns true if the charge status matches any of the given external statuses.
- New TokenEntity static method generateTokenFor, factory to create TokenEntity with auto-generated random token.
- ChargeService return a next URL for charges that are in CREATE or IN PROGRESS.


src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -52,6 +52,17 @@ public ChargeDao(final Provider<EntityManager> entityManager, EventDao eventDao)
...
+        String query = "SELECT te.chargeEntity FROM TokenEntity te WHERE te.token=:tokenId";
+
+        return entityManager.get()
+                .createQuery(query, ChargeEntity.class)
+                .setParameter("tokenId", tokenId)
+                .getResultList()
+                .stream()
+                .findFirst();
+    }
+
```
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/dao/TokenDao.java
```diff
@@ -26,7 +26,7 @@ public TokenDao(Provider<EntityManager> entityManager) {
 
     public Optional<TokenEntity> findByChargeId(Long chargeId) {
         return entityManager.get()
-                .createQuery("SELECT t FROM TokenEntity t WHERE t.chargeId = :chargeId", TokenEntity.class)
+                .createQuery("SELECT t FROM TokenEntity t WHERE t.chargeEntity.id = :chargeId", TokenEntity.class)
```
imranbohoran: charge_id is a column in the tokens table, so what was it that prompted this change? Wouldn't this result in 2 queries to load the Charge as well? I'm not sure I see the reason why we have to do this.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventEntity.java
```diff
@@ -8,6 +10,7 @@
 @SequenceGenerator(name = "charge_events_charge_id_seq", sequenceName = "charge_events_charge_id_seq", allocationSize = 1)
 public class ChargeEventEntity extends AbstractEntity {
 
+    @JsonIgnore
```
imranbohoran: I'm assuming you put this because charge_id was not available and json parsing failed(?). If that was case is that even possible? More to the point isn't that invalid if that's happening?

src/main/java/uk/gov/pay/connector/model/domain/TokenEntity.java
```diff
@@ -1,34 +1,43 @@
...
     }
 
-    public TokenEntity(Long chargeId, String token) {
-        this.chargeId = chargeId;
+    public static TokenEntity generateNewTokenFor(ChargeEntity chargeEntity) {
+        TokenEntity tokenEntity = new TokenEntity();
+        tokenEntity.setChargeEntity(chargeEntity);
+        tokenEntity.setToken(UUID.randomUUID().toString());
+        return tokenEntity;
+    }
```
imranbohoran: I quite like the static builder for this. However given that we are dealing with entities here, do you think we might have a risk here where someone calls this method without an non-existing ChargeEntity (non-existing in the DB) which would result in a failure when a persist is called for the TokenEntity retrieved from this static method?

src/main/java/uk/gov/pay/connector/resources/SecurityTokensResource.java
```diff
@@ -31,22 +31,18 @@ public SecurityTokensResource(TokenDao tokenDao, ChargeDao chargeDao) {
     }
 
     @GET
-    @Path(TOKEN_VALIDATION_PATH)
+    @Path(GET_CHARGE_BY_TOKEN_PATH)
```
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -100,6 +104,13 @@ public void updateStatus(List<ChargeEntity> chargeEntities, ChargeStatus status)
         return ImmutableMap.of(EXPIRY_SUCCESS, expiredSuccess + successFailPair.getLeft(), EXPIRY_FAILED, successFailPair.getRight());
     }
 
+    @Transactional
```
imranbohoran: Do we need @Transactional here at all. private scope is not picked up for putting it in a transactional context for guice AFAIK. We expect the caller to be in the transactional context isn't it?

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventEntity.java
ckalista: This is to avoid a infinite recursion during serialisation of a ChargeEntity. 

ChargeEntity has a one to many relationship with ChargeEventEntity. Since ChargeEventEntity also has a relationship with its ChargeEntity we end up with circular loop, hence the @JsonIgnore to avoid this.

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -170,18 +182,16 @@ private URI selfUriFor(UriInfo uriInfo, Long accountId, String chargeId) {
...
-                .path("charge")
-                .path(chargeId)
-                .queryParam("chargeTokenId", tokenId)
+                .path("secure")
+                .path(tokenId)
                 .build();
     }
 
-    private URI chargeUrl(String chargeId) {
+    private URI nextUrl() {
```
imranbohoran: 👍 For making it clearer by just overloading.

src/main/java/uk/gov/pay/connector/model/domain/TokenEntity.java
ckalista: In fact ChargeEntity doesn't even have to be created by JPA, it can be a plain-old-java instance created manually as long as the charge Id exists in the DB. This problem is true for any entity though i.e. if we manually link a ChargeEntity using the public setter setChargeEntity, we have exactly the same concern.

src/main/java/uk/gov/pay/connector/service/ChargeService.java
ckalista: ah! Good point - that's a mistake that I have addressed. Thx!

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -513,11 +584,29 @@ public void testFindByDateStatus_findsNoneForValidStatus() throws Exception {
...
         ArrayList<ChargeStatus> chargeStatuses = Lists.newArrayList(CREATED, ENTERING_CARD_DETAILS, AUTHORISATION_SUBMITTED, AUTHORISATION_SUCCESS);
 
         List<ChargeEntity> charges = chargeDao.findBeforeDateWithStatusIn(now().minusHours(1), chargeStatuses);
 
         assertThat(charges.size(), is(0));
     }
+
+    @Test
+    public void testFindChargeByTokenId() {
+        app.getDatabaseTestHelper().addCharge(100L, "ext-id", String.valueOf(defaultTestAccount.getAccountId()), 300L, CREATED, "return-url", "", "reference", now());
```
imranbohoran: This also could be using the DatabaseFixtures you introduced isn't it? I think it reads better when using the DatabaseFixtures

src/test/java/uk/gov/pay/connector/it/dao/DatabaseFixtures.java
```diff
@@ -0,0 +1,136 @@
+package uk.gov.pay.connector.it.dao;
+
+import uk.gov.pay.connector.model.domain.ChargeStatus;
+import uk.gov.pay.connector.util.DatabaseTestHelper;
+
+import java.time.ZonedDateTime;
+
+public class DatabaseFixtures {
```
imranbohoran: 👍 Nice. Reads better when setting up charges.

src/main/java/uk/gov/pay/connector/dao/TokenDao.java
ckalista: This is AFAIK, how many to one relationship should work in JPA. It was previously done by explicitly referring to the foreign key in the TokenEntity whereas instead, in good old JPA fashion, the foreign entity should simply be embedded as it is now.

src/test/java/uk/gov/pay/connector/it/dao/TokenDaoJpaITest.java
```diff
@@ -46,29 +71,31 @@ public void tearDown() {
...
-        tokenDao.persist(new TokenEntity(chargeId, tokenId));
+        ChargeEntity defaultChargeTestEntity = new ChargeEntity();
+        defaultChargeTestEntity.setId(defaultTestCharge.getChargeId());
+
+        TokenEntity tokenEntity = TokenEntity.generateNewTokenFor(defaultChargeTestEntity);
+
+        tokenDao.persist(tokenEntity);
 
-        assertThat(databaseTestHelper.getChargeTokenId(chargeId), is(tokenId));
+        assertThat(databaseTestHelper.getChargeTokenId(defaultChargeTestEntity.getId()), is(tokenEntity.getToken()));
```
imranbohoran: I see how this is working now. But I'm not sure if that's how we want it. The way we've mapped our relationships we are basically saying the TokenEntity owns the relationship to the ChargesEntity. So you can create a Charge by creating a token. Does that sound right? I thought a ChargeEntity _should_ exist to be able to create a token. But don't think creating a token means a charge can be created automatically.

src/main/java/uk/gov/pay/connector/model/domain/TokenEntity.java
```diff
@@ -1,34 +1,43 @@
...
 
-    @Column(name = "charge_id")
-    private Long chargeId;
-
     @Column(name = "secure_redirect_token")
     private String token;
 
+    @ManyToOne
+    @JoinColumn(name = "charge_id", nullable = false)
+    ChargeEntity chargeEntity;
```
imranbohoran: Are we certain this is a correct relationship. I didn't think a TokenEntity owns the relationship of tokens and charges.

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -195,17 +238,18 @@ public void shouldFindChargeForChargeIdAndAccountIdWithoutNextUrlWhenNoTokenExis
...
         when(chargeDao.findByExternalIdAndGatewayAccount(externalId, accountId)).thenReturn(chargeEntity);
-        Optional<TokenEntity> nonExistingToken = Optional.empty();
-        when(tokenDao.findByChargeId(chargeId)).thenReturn(nonExistingToken);
 
         Optional<ChargeResponse> chargeResponseForAccount = service.findChargeForAccount(externalId, accountId, uriInfo);
 
+        verify(tokenDao, never()).persist(any());
+
         ChargeResponse.Builder expectedChargeResponse = chargeResponseBuilderOf(chargeEntity.get());
         expectedChargeResponse.withLink("self", GET, new URI(SERVICE_HOST + "/v1/api/accounts/1/charges/" + externalId));
```
imranbohoran: Do you think it make sense to make assertions on the ChargeResponse for the non-existence of for next_url and next_url_post?

src/main/java/uk/gov/pay/connector/model/domain/TokenEntity.java
ckalista: As discussed this is the correct way to model the relationship between those two entities here.

src/main/java/uk/gov/pay/connector/dao/TokenDao.java
imranbohoran: I see now why this change was prompted based on the review of the TokenEntity. The inclusion of the Many-To-One means you need this. Unfortunately this now results in 3 queries as oppose to 1 query when retrieving a TokenEntity, but that's how JPA works, unless we make it lazy if we find it to be an overhead.

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
ckalista: Indeed, I have changed that. Thx.

src/test/java/uk/gov/pay/connector/it/dao/TokenDaoJpaITest.java
ckalista: As discussed and as we discovered together it actually doesn't create a new ChargeEntity in this case. If that's what we wanted then we would have to specify the JPA cascading rule.

src/main/java/uk/gov/pay/connector/model/domain/TokenEntity.java
ckalista: As discussed, this is correct.

imranbohoran:
Review completed.

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
ckalista: That assertion is already implicitly in place.

ckalista:
rebuild

ckalista:
rebuild

ckalista:
rebuild

ckalista:
rebuild

PR #136, opened 2016-04-13
PP-13 Frontend cancel endpoint

mrlumbu:
- adds /v1/frontend/charge/{chargeId}/cancel endpoint, and amends existing cancel functionality to reflect different status (USER CANCELLED vs SYSTEM CANCELLED)
- Ensures gateway isn't called when charge has 'pre-authorised' status


src/main/java/uk/gov/pay/connector/service/CardCancelService.java
```diff
@@ -21,6 +22,9 @@
             CREATED, ENTERING_CARD_DETAILS, AUTHORISATION_SUCCESS, AUTHORISATION_READY, CAPTURE_READY
     };
 
+    private static ChargeStatus[] localStatuses = new ChargeStatus[]{
+            CREATED, ENTERING_CARD_DETAILS
+    };
```
imranbohoran: Do you think we could give this a different name? the term localStatuses doesn't give much meaning as to what it is. And the statuses in the 'localStatuses' is a subset of what we have in legal statuses. 
Going through the code, how I understand this is that we are trying avoid going to the gateway providers if the charge is in the status of CREATED or ENTERING_CARD_DETAILS. 
We could consider changing the names and also perhaps using a different data structure, such as a list instead of an array to represent this.

src/main/java/uk/gov/pay/connector/service/CardCancelService.java
```diff
@@ -30,12 +34,32 @@ public CardCancelService(ChargeDao chargeDao, PaymentProviders providers) {
     }
 
     public Either<ErrorResponse, GatewayResponse> doCancel(String chargeId, Long accountId) {
-        return chargeDao
-                .findByExternalIdAndGatewayAccount(chargeId, accountId)
-                .map(TransactionalGatewayOperation.super::executeGatewayOperationFor)
-                .orElseGet(chargeNotFound(chargeId));
+        Optional<ChargeEntity> charge = chargeDao
```
imranbohoran: As the resources calls doCancel directly and since its possible to update the status of the charge without going to the gatway provider, we now seem to update the charge status without any transactions. This possibly could result in overriding an authorised charge. Can we consider doing the updates within a transaction as well?

src/main/java/uk/gov/pay/connector/service/CardCancelService.java
```diff
@@ -30,12 +34,32 @@ public CardCancelService(ChargeDao chargeDao, PaymentProviders providers) {
...
+                .findByExternalIdAndGatewayAccount(chargeId, accountId);
+        if (charge.isPresent()){
+            return globalCancel(charge.get(), ChargeStatus.SYSTEM_CANCELLED);
+        } else {
+            return chargeNotFound(chargeId);
+        }
+    }
+
+    Either<ErrorResponse, GatewayResponse> globalCancel(ChargeEntity charge, ChargeStatus status) {
```
imranbohoran: Do you think we could have a different name instead of globalCancel here?

src/main/java/uk/gov/pay/connector/service/CardCancelService.java
```diff
@@ -30,12 +34,32 @@ public CardCancelService(ChargeDao chargeDao, PaymentProviders providers) {
...
+            if (charge.hasStatus(localStatuses)) {
+                return localCancel(charge, status);
+            } else {
+                return TransactionalGatewayOperation.super.executeGatewayOperationFor(charge);
+            }
     }
 
+    private Either<ErrorResponse, GatewayResponse> localCancel(ChargeEntity chargeEntity, ChargeStatus status) {
```
imranbohoran: ditto

src/main/java/uk/gov/pay/connector/resources/CardResource.java
```diff
@@ -74,19 +77,30 @@ public Response cancelCharge(@PathParam("accountId") Long accountId, @PathParam(
         );
     }
 
+    @POST
+    @Path(FRONTEND_CANCEL_RESOURCE)
+    @Produces(APPLICATION_JSON)
+    public Response userCancelCharge(@PathParam("chargeId") String chargeId) {
```
imranbohoran: If we are having a separate resource for user cancellations does that mean we'd be having a separate user cancellation publicapi endpoint as well?

src/main/java/uk/gov/pay/connector/service/CardCancelService.java
```diff
@@ -30,12 +34,32 @@ public CardCancelService(ChargeDao chargeDao, PaymentProviders providers) {
...
+        } else {
+            return chargeNotFound(chargeId);
+        }
+    }
+
+    Either<ErrorResponse, GatewayResponse> globalCancel(ChargeEntity charge, ChargeStatus status) {
+            if (charge.hasStatus(localStatuses)) {
+                return localCancel(charge, status);
+            } else {
+                return TransactionalGatewayOperation.super.executeGatewayOperationFor(charge);
```
imranbohoran: Is there a reason why this is `TransactionalGatewayOperation.super.executeGatewayOperationFor(charge)` instead of just `executeGatewayOperationFor(charge)`?

src/main/java/uk/gov/pay/connector/resources/CardResource.java
kakumara: We already have one for public api. I wonder why we cannot use the same connector cancel resource for this too? (unless there is very different business logic here)

imranbohoran: Yes, so if are allowing user cancellation from publicapi, this would be duplicated and perhaps confusing for the users (publicapi users) too.

mrlumbu: Maybe there is a nicer way of doing this, but the story requires that when a user cancels from front end (either at entering card details or confirm stage), the status of the charge must be set to USER_CANCELLED (as opposed to SYSTEM_CANCELLED). I couldn't come up with a better way than wrapping the existing cancel functionality in an extra layer, which takes care of setting different status (bit like how expire works). Any better suggestions welcome!

kakumara: My bad. Just realised this is the citizen frontend cancelling charges (thought it is from a selfservice user). You defintely need a separate resource.

src/main/java/uk/gov/pay/connector/service/CardCancelService.java
```diff
@@ -1,64 +1,112 @@
...
             return chargeNotFound(chargeId);
         }
     }
 
-    Either<ErrorResponse, GatewayResponse> globalCancel(ChargeEntity charge, ChargeStatus status) {
-            if (charge.hasStatus(localStatuses)) {
-                return localCancel(charge, status);
+    Either<ErrorResponse, GatewayResponse> cancelCharge(ChargeEntity charge) {
+            if (charge.hasStatus(nonGatewayStatuses)) {
+                return nonGatewayCancel(charge);
```
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/service/CardCancelService.java
```diff
@@ -1,64 +1,112 @@
...
             } else {
-                return TransactionalGatewayOperation.super.executeGatewayOperationFor(charge);
+                return executeGatewayOperationFor(charge);
```
imranbohoran: 👍 I think this reads well now.

src/main/java/uk/gov/pay/connector/resources/ChargesResource.java
```diff
@@ -162,7 +165,7 @@ public Response createNewCharge(@PathParam("accountId") Long accountId, Map<Stri
     public Response expireCharges(@Context UriInfo uriInfo) {
         List<ChargeEntity> charges = chargeDao.findBeforeDateWithStatusIn(getExpiryDate(), NON_TERMINAL_STATUSES);
         logger.info(format("%s charges found expiring since %s", charges.size(), getExpiryDate()));
-        Map<String, Integer> resultMap = chargeService.expire(charges);
+        Map<String, Integer> resultMap = cardCancelService.expire(charges);
```
imranbohoran: Was there a reason why the expiry was moved to the cardCancelService? The expire to me felt to be part of ChargeService. But you probably had a good reason.

mrlumbu: Well, the initial reason was I ended up with some circular dependencies when mocking for tests - cancel service relied on the charge service, and the charge service relied on the cancel service. This led me to think that grouping 'cancel-like' behaviour in cancel service was a nice way out. Guice documentation recommends removing cycles

imranbohoran: ok makes sense. It's unfortunate we have to move the different concepts of the domain into services. But that's a different matter. 👍

imranbohoran:
Review done

PR #139, opened 2016-04-18
PP 651 bug fix, internal charge id being displayed in events

rahul-vinayak:
- changed internal charge id to external id in the events object
- JsonIgnored charge_id in the son response for events as its wrapped
  inside charge_id, so returning charge_id with every event wasn’t
  required
- moved the charge events tests to Eventsresource IT from
  ChargeResource IT.


src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceITest.java
```diff
@@ -86,6 +94,28 @@ public void shouldGetAllEventsForAGivenCharge() throws Exception {
...
+
+        connectorApi
+                .getEvents(externalChargeId)
+                .body("charge_id", Matchers.is(externalChargeId))
+                .body("events.status", hasSize(4))
+                .body("events.status[0]", Matchers.is(EXT_CREATED.getValue()))
+                .body("events.status[1]", Matchers.is(EXT_IN_PROGRESS.getValue()))
+                .body("events.status[2]", Matchers.is(EXT_SYSTEM_CANCELLED.getValue()))
+                .body("events.status[3]", Matchers.is(EXT_IN_PROGRESS.getValue()))
+                .body("events.chargeId[0]", isEmptyOrNullString()); // chargeId should not be there in json response for every event
```
rauligs: Only checked for the first event? worth checking for all?

rahul-vinayak: ya can do that, although will do a shorter test by doing a size check

rahul-vinayak: unfortunately the size check doesn't work with  weird rest assured, it still comes back as size 4, so doing as you originally suggested as this is the only way it seems to work

PR #140, opened 2016-04-18
PP-598 Handle rollbacks correctly in transactional methods

PauloPortugal:
- Created instances of `WebApplicationRuntime` exception, as it may
  be thrown by a resource method as we want to show specific `HTTP`
  error responses. For that we reused the `ResponseUtil` methods
- With the introdcution of exceptions being throwned in the service
  layer we only need to return a success response from the '_preOperation_',
  '_operation_' and '_postOperation_' and any wrapper methods, therefore
  there is no need to continue using `Either<T,V>`
- Fixed the tests to address the new expectations

with @ckalista


src/main/java/uk/gov/pay/connector/exception/ConnectorRuntimeException.java
```diff
@@ -0,0 +1,17 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+
+public abstract class ConnectorRuntimeException extends WebApplicationException {
+
+    protected final Logger logger = LoggerFactory.getLogger(this.getClass());
```
rahul-vinayak: logger seems to be unused here

PauloPortugal: @rahul-vinayak it is being used by the concrete classes that extend `ConnectorRuntimeException`

src/main/java/uk/gov/pay/connector/exception/GenericGatewayRuntimeException.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.pay.connector.exception;
+
+import static uk.gov.pay.connector.util.ResponseUtil.serviceErrorResponse;
+
+public class GenericGatewayRuntimeException extends ConnectorRuntimeException {
+    public GenericGatewayRuntimeException(String message) {
+        super(serviceErrorResponse(message));
+    }
+}
```
rahul-vinayak: Is there a specific reason why we needed ConnectorRuntimeException class to be introduced, rather than extending directly from WebApplicationException

PauloPortugal: Ah! Good point!
It started as a class that would carry some shared behaviour, but in the end it is more a proxy like class. The only shared "resource" is the `logger` attribute. 
For that alone I would keep it, avoids us having to declare and instanciate a `logger` in every concrete class. @ckalista your opinion, Sir?

ckalista: I would agree, no need to contaminate the codebase with purposeless abstract classes.

src/main/java/uk/gov/pay/connector/exception/ConnectorRuntimeException.java
rahul-vinayak: I don't see it being used by the concrete classes, unless I am missing something

src/main/java/uk/gov/pay/connector/exception/GenericGatewayRuntimeException.java
rahul-vinayak: @PauloPortugal again I didn't see the logger getting used in concrete classes so if there is no usage for that as we agreed probably we should remove an abstraction

PauloPortugal: @rahul-vinayak this has been addressed.

src/main/java/uk/gov/pay/connector/exception/ConnectorRuntimeException.java
PauloPortugal: @rahul-vinayak  It has been deleted.

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
```diff
@@ -29,38 +28,34 @@ public CardCaptureService(ChargeDao chargeDao, PaymentProviders providers) {
         this.providers = providers;
     }
 
-    public Either<ErrorResponse, GatewayResponse> doCapture(String chargeId) {
+    public GatewayResponse doCapture(String chargeId) {
         return chargeDao
                 .findByExternalId(chargeId)
                 .map(TransactionalGatewayOperation.super::executeGatewayOperationFor)
-                .orElseGet(chargeNotFound(chargeId));
+                .orElseThrow(() -> new ChargeNotFoundRuntimeException(format("Charge with id [%s] not found.", chargeId)));
```
rahul-vinayak: Now that we have very specific custom built exceptions, could the String messages be pushed into the exception classes themselves so that service and resource classes become much cleaner and can stay away from building an error message, and lead to less duplicacy in building messages

e.g.
new ChargeNotFoundRuntimeException(chargeId);

and rest remains in the custom class

PauloPortugal: This has been refactored.

PauloPortugal:
rebuild

rahul-vinayak:
🎱 👍

PR #143, opened 2016-04-21
PP-598 Test the DB rollback upon thrown exception

PauloPortugal:
- Changed unit tests to ensure that after an exception is thrown the charge remains in the same state previously was.


rahul-vinayak:
👍 good to go

PR #146, opened 2016-04-24
PP-182 certificate-related fixes.

ijmad:
Switch from DB_URL to DB_HOST
Set required complex options on the connection string to use
truststore of loaded certificates provided by CERTS_PATH.

This param is already provided by ansible. Yay!

Fix HTTPS use of certificates
Make Jersey's REST client use certificates loaded from
custom truststore that is generated for certificates
found at CERTS_PATH.

Merge after https://github.com/alphagov/pay-scripts/pull/175


rauligs:
rebuild

rauligs:
👍

PR #147, opened 2016-04-25
PP 209 [REVIEW ONLY] Change service name

ckalista:
# WHAT
- Add service name field in GatewayAccountEntity.
- Bind updateGatewayAccountCredentials endpoint to different URL _/v1/frontend/accounts/:accountId/credentials_ and structural change of the message payload that now wraps 'username' and 'password' inside of parent element 'credentials'.
- New updateGatewayAccountServiceName endpoint:
  - Bind to URL _/v1/frontend/accounts/:accountId/servicename_.
  - Checks on the message payload to ensure service_name field is present.
  - Checks on the service_name field to ensure its length doesn't exceed 50 characters.
- Upgraded liquibase migration script to add new service_name field.
- Unit tests.
# WHO

Any dev


rauligs:
rebuild

kakumara:
Nice PR description 👍

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -135,10 +144,70 @@ public Response updateGatewayCredentials(@PathParam("accountId") Long gatewayAcc
                 .filter(requiredField -> !credentialsPayload.has(requiredField))
                 .collect(Collectors.toList());
     }
+    // <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< BACKWARD COMPATIBLE FIX ENDS HERE
+
```
kakumara: It seems with this change we have 2 separate endpoints for updating gateway accounts. I'm not sure if having micro update endpoints (per field) is the right REST approach for resources. I'd rather:
 a. have one PUT resource for the whole entity updates. (with optional fields, and validate on availability)
OR
b. Have micro PATCH endpoints (only for selected fields e.g. creds) still not for each field

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
```diff
@@ -14,24 +16,102 @@
 
 public class GatewayAccountFrontendResourceITest extends GatewayAccountResourceTestBase {
 
+    static public class GatewayAccountPayload {
+        String userName;
```
kakumara: Minor: I'd rather choose a Map for credentials instead of flat fields (given the cred structure is different from each provider)

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
```diff
@@ -55,97 +135,111 @@ public void shouldReturn404IfGatewayAccountIsNotNumeric() {
...
+                .then()
+                .statusCode(200);
+
+        Map<String, String> currentCredentials = app.getDatabaseTestHelper().getAccountCredentials(Long.valueOf(accountId));
+        assertThat(currentCredentials, is(gatewayAccountPayload.getCredentials()));
     }
 
     @Test
-    public void shouldSaveSpecialCharactersInUserNamesAndPassword() throws Exception {
+    public void shouldUpdateGatewayAccountCredentialsWithCharactersInUserNamesAndPassword() throws Exception {
```
kakumara: may be `special characters` in the test name rather than just characters

kakumara:
Other than that 👍

PR #148, opened 2016-04-25
PP-182 Use DB_SSL_OPTION

ijmad:
This will selectively enable/disable SSL when we're ready
Setting ssl=true for everywhere was a little hasty.

Option not being present defaults to SSL off which is right for prod for now.

Depends on https://github.com/alphagov/pay-scripts/pull/179/files


ijmad:
rebuild

PR #152, opened 2016-04-29
PP-670 fixes the illegal state exception and optimistic lock exception bug while cancelling AUTH_SUCCESS charges

rahul-vinayak:
- Added EXPIRE_CANCEL_PENDING as one of the legal states for cancelling
  a charge.
- Expanded the transaction, to cover up for optimistic lock exception
  being thrown earlier while doing charge.update calls on a detached
  objects when expiring a charge with AUTHORISATION_SUCCESS state.
- Update and merge charge calls now return the merged charge collection
  as opposed to voids, so that versions get updated in the managed charge
  entities
- updated the tests to cover this bug fix


src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelFrontendResourceITest.java
```diff
@@ -43,7 +43,6 @@
             CAPTURED,
             CAPTURE_SUBMITTED,
             CAPTURE_ERROR,
-            EXPIRE_CANCEL_PENDING,
             EXPIRE_CANCEL_FAILED,
```
mrlumbu: Should EXPIRE_CANCEL_PENDING not be added to cancellable statuses?

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelResourceITest.java
```diff
@@ -69,7 +69,6 @@ public void respondWith400_whenNotCancellableState() {
                 CAPTURED,
                 CAPTURE_SUBMITTED,
                 CAPTURE_ERROR,
-                EXPIRE_CANCEL_PENDING,
```
mrlumbu: As before, should EXPIRE_CANCEL_PENDING not be added to cancellable statuses?

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelFrontendResourceITest.java
rahul-vinayak: Probably yes if there was an ITest for that, and thats the reason I don't like separate statuses being maintained in the code and test, should be driven off the same List coming from the actual code. Will try and refactor the code a bit here

mrlumbu:
👍

rahul-vinayak:
rebuild

PR #153, opened 2016-04-29
PP-647 Waiting for db to be ready

imranbohoran:
## WHAT

This PR introduces a dropwizard managed lifecycle object that dropwizard would run prior to jetty starting to accept HTTP requests. When there is no connectivity to DB the main thread that starts the application sleeps for 5 seconds and checks again. And during that time if we have DB connectivity then the application continues its startup. The waiting is not on constant intervals. The check intervals increase by 5 seconds every time it finds out that there is no DB connectivity. 
This will also allow us to get rid of the 20 seconds sleep we currently have in our Dockerfile
## HOW TO REVIEW

To see this working; 
- Bring up all the services (msl -a up) and start connector from your IDE
- Then go into the connector folder and run msl debug. 
- Now shutdown the connector you started from your IDE.
- Kill the pay_connector_db_1 container (i.e. docker kill pay_connector_db_1)
- Now start the connector form your IDE.
- You will see the application logging saying that its waiting for x seconds till the database is available.
- Now bring up the connector db container (you can use the files in the attached zip). Run ./db.sh up and that would bring up the db container for connector.
  [connector-db-up.zip](https://github.com/alphagov/pay-connector/files/242785/connector-db-up.zip)
- As the db container comes up you would see that the container proceeds its startup and will be available to accept requests.
## WHO
- Any Dev


imranbohoran:
rebuild

imranbohoran:
rebuild

src/main/java/uk/gov/pay/connector/managed/DependentResourceChecker.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public void stop() throws Exception {}
+
+    private void waitingForDatabaseConnectivity() {
+        logger.info("Checking for database availability >>>");
+        boolean databaseAvailable = isDatabaseAvailable();
+
+        long timeToWait = 0;
```
PauloPortugal: I am not that sure about increasing the waiting time.

> | Attempts | 5 sec delay | 5 sec increment delay |
> | --- | --- | --- |
> | 1 | 5 sec | 5 sec |
> | 2 | 5 sec | 10 sec |
> | 3 | 5 sec | 15 sec |
> | 4 | 5 sec | 20 sec |

After three attempts the application will have to wait 30 seconds.
After four attempts instead of waiting 20 seconds we will wait 50 seconds

Is it really worth waiting that much, can't we just continue to retry every 5 seconds?

src/main/java/uk/gov/pay/connector/managed/DependentResourceChecker.java
```diff
@@ -0,0 +1,56 @@
...
+import io.dropwizard.lifecycle.Managed;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.util.DependentResource;
+
+import javax.inject.Inject;
+import java.sql.Connection;
+import java.sql.SQLException;
+
+public class DependentResourceChecker implements Managed {
```
PauloPortugal: :+1:

src/test/resources/config/test-config.yaml
```diff
@@ -13,7 +13,7 @@ logging:
         threshold: ALL
         timeZone: UTC
         target: stdout
-        logFormat: "[%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n"
+        logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n"
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/connector/managed/DependentResourceChecker.java
imranbohoran: The reason this is done is that we didn't want to always keep hitting the db on a constant duration should the db be down. It could just keep increasing the load if we just keep on hitting it. This way we spread the interval so we give it enough time. It does mean that there would be a delay in the application coming up even though the db might be accessible. i.e. After 4 attempts we'd wait for 25 seconds. But if the DB was accessible soon after 4 attempts, we'd still wait for 25 seconds. We could probably make it a bit more smarter to say after waiting for 1 minute, reset the time interval. Also the 5 second delay can be changed if we think 5 seconds is too much to be incrementing on. I'd like to hear @bazbremner 's thoughts on this as well.

imranbohoran:
rebuild

imranbohoran:
rebuild

imranbohoran:
rebuild

imranbohoran:
rebuild

bazbremner:
Looks fine to me; my only concern about the sleeps between retries is that we don't have multiple applications hammering on the RDS instance and/or pegging CPU whilst spinning/hammering the DB. The current implementation seems like a sensible starting point.

The only thing that's missing is the removal of the sleep 20 at https://github.com/alphagov/pay-connector/blob/master/Dockerfile#L18 - if this code really works, let's exercise it immediately.

imranbohoran:
@bazbremner Once these are merged into master, removing the sleep 20 is a item on the list before the story is completed.

PR #154, opened 2016-04-29
PP-656 Improve healthchecks so they are consistent and more useful

PauloPortugal:
## WHAT
- The DropWizard healthcheck endpoint is on the admin port and this
  creates much more complexity when routing ports through nginxes, proxies
  and security groups.
## HOW
- The Healthcheck should be available on the application port, making use
  of the existing healthchecks that follow Dropwizard's standards.
## WHO
- Developers and DevOps


imranbohoran:
As discussed, in addition to returning a 200 when everythings running for a node, we should also return a json response with what is failing if the response is a 500.

Also can we consider the new threadpool introduced for processing gateway calls as part of this work?

src/main/java/uk/gov/pay/connector/resources/HealthCheckResource.java
```diff
@@ -0,0 +1,38 @@
...
+    Environment environment;
+
+    @GET
+    @Path(HEALTHCHECK)
+    @Produces(APPLICATION_JSON)
+    public Response captureCharge(@PathParam("chargeId") String chargeId) {
+        healthy = environment.healthChecks().getNames().size() == environment.healthChecks().runHealthChecks()
+                .values()
+                .stream()
+                .filter(healthcheck -> healthcheck.isHealthy() == true)
```
imranbohoran: we could possibly simplify this by using a method reference like .filter(HealthCheck.Result::isHealthy)

PauloPortugal:
@imranbohoran I have addressed the 200 vs 500 response and I have also created a new ThreadPool healthcheck that checks if the Thread Pool is exhausted.

src/main/java/uk/gov/pay/connector/healthcheck/CardExecutorServiceHealthCheck.java
```diff
@@ -0,0 +1,28 @@
...
+    private ThreadPoolExecutor threadPoolExecutor;
+
+    @Inject
+    public CardExecutorServiceHealthCheck(CardExecutorService cardExecutorService) {
+        this.threadPoolExecutor = (ThreadPoolExecutor)cardExecutorService.getExecutor();
+    }
+
+    @Override
+    protected Result check() throws Exception {
+        if (threadPoolExecutor.getActiveCount() < threadPoolExecutor.getMaximumPoolSize()) {
```
imranbohoran: As discussed, can make this check a bit more smarter. So instead of checking for the active count can we check the size of the queue instead. We can mark this as unhealthy if the queue size is greater than say 10. 
Also in the response we send back for this healthcheck, can we add the active-thread-count, pool-size, core-pool-size and queue-size. This way the monitoring tools can display the data.

src/main/java/uk/gov/pay/connector/resources/HealthCheckResource.java
```diff
@@ -0,0 +1,54 @@
...
+
+        boolean healthy = results.size() == results.values()
+                .stream()
+                .filter(HealthCheck.Result::isHealthy)
+                .count();
+
+        if(healthy) {
+            return Response.ok().entity(response).build();
+        }
+        return status(503).entity(response).build();
```
imranbohoran: Can we have a test to cover this as well?

imranbohoran:
👍

PR #155, opened 2016-05-05
PP 335 uses CriteriaBuilder as opposed to jpa queries for ChargeSearches

rahul-vinayak:
- got rid of ChargeSearch class and its test
- the Charge search uses ChargeSearchParams class and now find charges
  using a CriteriaBuilder, as its a much cleaner approach when accepting optional search params from the user, builds predicates as opposed to String appending to construct a Query as the ChargeSearch was presently doing
- Refactored other code to make use of new way of searching, and fixed
  other tests
- Introduces the concept of limit the searchResult and the page number (pagination still in progress, however this change will limit the search results to 100 by default). But the focus of this PR should ignore pagination aspect as it is still under play, this PR should just focus only on the CriteriaBuilder aspect only.


src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -103,14 +90,87 @@ public void searchChargesByGatewayAccountIdOnly() throws Exception {
...
+                .withGatewayAccountId(defaultTestAccount.getAccountId())
+                .withPage(3L)
+                .withDisplaySize(3L);
+        charges = chargeDao.findAllBy(params);
+        // then
+        assertThat(charges.size(), is(2));
+        assertThat(charges.get(0).getId(), is(300L));
+        assertThat(charges.get(1).getId(), is(200L));
+    }
+
```
mrlumbu: Is it worth having one more pagination test where you set pagination params and other params (date, ref...) at same time?

rahul-vinayak: I should have made it clear earlier, but the pagination story is still under play and will have more tests and coverage around that aspect in PRs coming on the way, This pullRequest is to just focus around the new change around the CriteriaBuilder and replacement of ChargeSearch class only. I have updated the PR comment for that now.

mrlumbu: Ok cool. Looks good otherwise

mrlumbu:
👍

PR #156, opened 2016-05-06
PP-335 this changes the charges being returned in order of creation date

rahul-vinayak:
- changed the clause of order by for find charges in descending order of
  CREATION_DATE in DAO as opposed to order by charge_id (which is internal ID and that was never exposed to outside world)
- fixed resource and dao tests
- also added a couple of charge resource tests to test the pagination, i.e. setting the page and display_size params while getting transactions


src/test/java/uk/gov/pay/connector/it/resources/ChargesResourceITest.java
```diff
@@ -299,6 +300,78 @@ public void shouldFilterTransactionsBasedOnFromAndToDates() throws Exception {
     }
 
     @Test
+    public void shouldGetAllTransactionsForDefault_page_1_size_100_inCreationDateOrder() throws Exception {
```
rauligs: 👍

PR #157, opened 2016-05-09
PP-541 lifecycle

ijmad:
Adds new lifecycle to the connector, replacing 'status' with a states structure. 

New states are shown in diagram at https://docs.google.com/drawings/d/1a5eFLDs2U9_vwLwpTtZ5Arln5fiIUYZg7ex5LFjsqfk

I have added some more information to the returned 'state' structure, whether the payment is 'finished', whether it was a 'success', and if not, some further information about what went wrong. Documented in README.md.

Keeping the 'status' parameter for now so tests and other modules remain backwards compatible. Will be removed as soon as the first traunche of PRs are open.

Now depends on https://github.com/alphagov/pay-endtoend/pull/166 to go green


src/main/java/uk/gov/pay/connector/model/api/ExternalChargeStateSerializer.java
```diff
@@ -0,0 +1,38 @@
+package uk.gov.pay.connector.model.api;
+
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+
+import java.io.IOException;
+
+public class ExternalChargeStateSerializer extends JsonSerializer<ExternalChargeState> {
```
kakumara: Is this serialiser really needed? (why not `@JsonProperty` with `@JsonInclude(Include.NON_NULL)`). Also all states looks pretty immutable with values already nicely set.

kakumara:
More of a question on API design of the new state structure. Feels like fields `state`, `finished`, `success` carries too much information if it goes all the time, meaning api integrators have to check few things in order to understand this as well? Can't we pass the same understanding in a little bit more concise way? 

_Looks like all these stuff going in CSV as well, including error messages_

kakumara:
Assume this PR doesn't cover the `state machine` we discussed sometime ago, where ChargeEntity enforcing movement of states?

src/test/java/uk/gov/pay/connector/model/api/ExternalChargeStateTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsNull.notNullValue;
+import static org.hamcrest.core.IsNull.nullValue;
+import static org.junit.Assert.assertThat;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.*;
+
+public class ExternalChargeStateTest {
+    @Test
+    public void shouldMapAnInternalStatusToExternalStateCorrectly() throws Exception {
```
kakumara: (minor) Feels like this could be a `@Parameterized` test ..

src/test/java/uk/gov/pay/connector/model/api/LegacyChargeStatusTest.java
```diff
@@ -0,0 +1,19 @@
...
+public class LegacyChargeStatusTest {
+    @Test
+    public void shouldMapAnInternalStatusToALegacyStatusCorrectly() throws Exception {
+        assertThat(CAPTURE_READY.toLegacy(), is(LEGACY_EXT_IN_PROGRESS));
+        assertThat(CREATED.toLegacy(), is(LEGACY_EXT_CREATED));
+        assertThat(AUTHORISATION_ERROR.toLegacy(), is(LEGACY_EXT_FAILED));
+        assertThat(CAPTURE_SUBMITTED.toLegacy(), is(LEGACY_EXT_SUCCEEDED));
+        assertThat(SYSTEM_CANCELLED.toLegacy(), is(LEGACY_EXT_SYSTEM_CANCELLED));
+    }
+}
```
kakumara: new line

src/main/java/uk/gov/pay/connector/model/api/ExternalChargeStateSerializer.java
ijmad: Couldn't figure out how to get Jackson to play nice with Enums, is there a better way?

ijmad:
@kakumara State machine is PP-547, yes

ijmad:
@kakumara 

On the API design, I agree it's perfectly possible to derive 'finished' and 'success' from the current lifecycle state, they are 'derived' characteristics. However, given the feedback we've had which is that partners want to know for sure if a payment has terminated (which I've renamed 'finished' because I think the language is friendlier), and if it's been successful, I thought this was good information to provide rather than hiding it away in docs.

I would be ok with reconsidering this, possibly removing 'success' might be sensible. I am fairly strong that 'finished' is useful though.

ijmad:
@kakumara , as discussed I have removed 'success' but kept 'finished'. Let me know your thoughts now.

src/test/java/uk/gov/pay/connector/model/api/ExternalChargeStateTest.java
ijmad: Have changed this around a bit.

src/test/java/uk/gov/pay/connector/model/api/LegacyChargeStatusTest.java
ijmad: It's a rename and will be getting deleted shortly...!

src/main/java/uk/gov/pay/connector/model/api/ExternalChargeStateSerializer.java
kakumara: if you annotate each getter in `ExternalChargeState` with `@JsonProperty` and have a class level annotation `@JsonInclude(Include.NON_NULL)` this should serialise as per your expectation. (since you are only ignoring properties if they are null)

src/test/java/uk/gov/pay/connector/model/api/LegacyChargeStatusTest.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/model/api/ExternalChargeStateSerializer.java
ijmad: Sorted.

ijmad:
Added 3e20c678dc072749974d3e7e08848e26c4ff0769, see if you think this is the correct solution.

As a result of this commit, https://github.com/alphagov/pay-endtoend/pull/166 must be merged first.

ijmad:
rebuild

PR #158, opened 2016-05-09
PP-647 Waiting for the Database as a command

imranbohoran:
## WHAT
- Introducing a dropwizard command to wait for the database to be
  available.
- The 20 second sleep in the Dockerfile is removed as the command will
  take care of this.
- Calling the dropwizard command before running migrations and starting
  the HTTP server. This way the command will wait till the database is
  available. When it's available it would continue with migrations and
  starting up the HTTP server to accept requests.
## HOW TO REVIEW

To see this working;
- Remove the connector_db entry in $WORKSPACE/pay-scripts/lib/msl/msl.sh under the msl_database_names section
- Bring up all the services (msl -a up)
- You will see the connector logging saying that its waiting for x seconds till the database is available. (run msl logs from $WORKSPACE/pay-connector)
- Now bring up the connector db container (you can use the files in the attached zip). Run ./db.sh up and that would bring up the db container for connector. 
  [connector-db-up.zip](https://github.com/alphagov/pay-connector/files/254633/connector-db-up.zip)
- As the db container comes up you would see that the connector app proceeds its startup and will be available to accept requests.
## WHO

Any Dev or WebOps


src/main/java/uk/gov/pay/connector/util/ApplicationStartupDependentResourceChecker.java
```diff
@@ -1,34 +1,26 @@
...
-import javax.inject.Inject;
 import java.sql.Connection;
 import java.sql.SQLException;
 
-public class DependentResourceChecker implements Managed {
+public class ApplicationStartupDependentResourceChecker {
 
-    public static final int PROGRESSIVE_SECONDS_TO_WAIT = 5;
-    private static final Logger logger = LoggerFactory.getLogger(DependentResourceChecker.class);
+    private static final int PROGRESSIVE_SECONDS_TO_WAIT = 5;
```
rauligs: Is this something we want to externalise as env variable? (maybe as part of other PR but probably worth checking with webops?)

rauligs:
👍

src/main/java/uk/gov/pay/connector/util/ApplicationStartupDependentResourceChecker.java
imranbohoran: Don't think we need to externalise this as this is not something that would be tweaked by webops

PR #159, opened 2016-05-09
PP-275 HTTP connection leak in Connector

ckalista:
# WHAT
- New GatewayClient.Response pre-fetches the entity of the response it wraps and deletegates all other operation.
- Enforce the response close in a finally before returning to protect against memory leaks.
# WHO

Any dev


src/main/java/uk/gov/pay/connector/service/GatewayClient.java
```diff
@@ -77,11 +84,15 @@ public static GatewayClient createGatewayClient(Client client, String gatewayUrl
         } catch (Exception e) {
             logger.error(format("Exception for gateway url=%s", gatewayUrl), e);
             return left(baseError(e.getMessage()));
+        } finally {
```
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
```diff
@@ -91,4 +102,98 @@ public static GatewayClient createGatewayClient(Client client, String gatewayUrl
             return left(malformedResponseReceivedFromGateway("Invalid Response Received From Gateway"));
         }
     }
+
+    static public class Response {
```
imranbohoran: I still can't think of a good reason to keep these unused methods exposed form the new Response object. We only ever seem to be using getEntity which is what is used to unmarshall into the object, and the getStatus method. We can add them if needed (although I don't see a need arising for them to be added)

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
```diff
@@ -91,4 +102,98 @@ public static GatewayClient createGatewayClient(Client client, String gatewayUrl
...
+        private final javax.ws.rs.core.Response delegate;
+        private final String entity;
+
+        protected Response(final javax.ws.rs.core.Response delegate) {
+            this.delegate = delegate;
+            this.entity = delegate.readEntity(String.class);
+        }
+
+        public int getStatus() {
+            return delegate.getStatus();
```
imranbohoran: Shouldn't this be initialised at the constructor, as the delegate (whichi is the response) might be closed by the time this method is invoked?

ckalista: My understanding is that only the entity methods where the entity is backed by an input stream are affected by the close.

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
ckalista: If we didn't need some of those methods we could have probably gotten away with returning a String. However we do seem to need more control. For instance here: 

```
    private InquiryResponse inquire(String transactionId, GatewayAccountEntity gatewayAccount) {
        return reduce(
                client
                        .postXMLRequestFor(gatewayAccount, buildOrderInquiryFor(gatewayAccount, transactionId))
                        .bimap(
                                InquiryResponse::inquiryFailureResponse,
                                (response) -> response.getStatus() == OK.getStatusCode() ?
                                        mapToInquiryResponse(response) :
                                        errorInquiryResponse(logger, response)
                        )
        );
    }
```

imranbohoran: Yes, but isn't that just getStatus(). So we only expose getEntity() and getStatus() and if we initialise the entity and status in the constructor there's no need to do any delegation.

ckalista: The goal was not to close the API but as discussed I am not fussed so happy to change that.

ckalista:
rebuild

imranbohoran:
👍 Looks good

PR #160, opened 2016-05-10
PP-615 'XML Entity' related attacks mitigation

rauligs:
# WHAT

Since payment providers send us notifications (no whitelisting), we need to make sure XML processing is secure by mitigating different possible _XML ws_ related attacks. This is one of a serie of several PR's covering a range of different XML attacks, more information can be found in [PP-615](https://payments-platform.atlassian.net/browse/PP-615), where all possible vulnerabilities and mitigations (plus some sources) can be found.
- _XML Entity Expansion_ is being mitigated by setting explicitly _FEATURE_SECURE_PROCESSING_ flag to the parser and setting the Entity Limit to the minimum value (1), since the default JDK value is 64000 (so it gets override to 1).
- _XML Entity Reference_ and _XML External Entity DOS_ attacks are being mitigated by disabling external entities and parameters. Previous point set the entity processing to one (but is still one) so we need to make sure external entities are not processed.
# WHO
- Any dev


src/main/java/uk/gov/pay/connector/util/XMLUnmarshaller.java
```diff
@@ -40,11 +45,25 @@
...
     }
 
-    private static XMLReader buildXmlReaderWithDtdValidationDisabled() throws ParserConfigurationException, SAXException {
+    private static XMLReader buildXmlReader() throws ParserConfigurationException, SAXException {
         SAXParserFactory spf = SAXParserFactory.newInstance();
-        spf.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
-        spf.setFeature("http://xml.org/sax/features/validation", false);
-        spf.setNamespaceAware(true);
-        return spf.newSAXParser().getXMLReader();
+        spf.setFeature(FEATURE_SECURE_PROCESSING, true);
```
imranbohoran: Does this have any effect? Doesn't seem to affect the current tests. Or are we missing a test case?

rauligs: It doesn't because this particular implementation set it to true by default. Check the story comment where a review/defuzz/investigation is done and only we set it explicitly (so is visible to us).
As discussed still investigating what other advantages give us this being set to _true_ so then I can write more tests according, but at the moment is less harm setting it (as _OWASP_ recommends) rather than leaving the unmarshalling without it.

imranbohoran:
👍

PR #161, opened 2016-05-11
PP-669 Date formating cvs create date

imranbohoran:
## WHAT
- Formating to london time. Otherwise we'd be showing london time in the
  transaction listing page and UTC in the csv which will be inconsistent.

with @mcgoooo
## WHO

Any Dev other than @mcgoooo


src/test/java/uk/gov/pay/connector/util/DateTimeUtilsTest.java
```diff
@@ -65,4 +65,15 @@ public void shouldConvertAn_UTCOffsetDateStringToADateTime() throws Exception {
...
+    public void shouldConvertZoneDateTimeToLondonTimeZone() {
+        String aDate = "2016-07-07T17:24:48Z";
+        Optional<ZonedDateTime> zonedDateTime = DateTimeUtils.toUTCZonedDateTime(aDate);
+
+        String result = DateTimeUtils.toLondonZone(zonedDateTime.get());
+
+        assertThat(result, is("2016-07-07 18:24:48"));
+
+    }
 }
```
mrlumbu: For completeness it would be nice to have another test date in winter, to check that it is actually London time. But probably overkill to add in now

mrlumbu:
👍

src/test/java/uk/gov/pay/connector/util/DateTimeUtilsTest.java
mcgoooo: the test above does that for the moment, albeit implicitly
https://github.com/alphagov/pay-connector/blob/a32e2d01725509af3252259c894a57cd3b2899a8/src/test/java/uk/gov/pay/connector/util/DateTimeUtilsTest.java#L62

PR #162, opened 2016-05-12
Fixing inconsitent test runs on development machines

imranbohoran:
## WHAT
- The DAO integration tests were using the same dropwizard application
  as the other tests and as a result was interfering with the results. The
  DAO integration tests only need a persistent module and therefore this
  has been extracted to a base class which they can inherit from. This
  makes the test run more consistently on development environments.
## HOW TO REVIEW

Run all tests on your IDE. They should all pass.
## WHO

Any Dev other than @kakumara


imranbohoran:
rebuild

rauligs:
As discussed I would do the postgres rule a bit more intelligent (to pass a parameter to trigger or not the migration scripts optionally). But this fix the testing and is good!

PR #164, opened 2016-05-13
PP-582 Add logging filter

imranbohoran:
## WHAT
- Introducing a Logging filter that would log all incoming requests with a prefix. Once the request is processed by the relevant resource it would also log the end of the chain with a prefix.
- Any exception thrown will also be logged with a prefix, exception message and the stack trace. A catch on this filter was opted against a ExceptionMapper so we can get a nice trace of start to finish (with
  erros)
- This gives us a nice trace of the request in the form of;
  start->error(if any)->end
## HOW TO REVIEW

Run through a payment journey and you should be able to see start->end logging for each resource hit on the connector.
## WHO

Any Dev


imranbohoran:
rebuild

imranbohoran:
rebuild

imranbohoran:
rebuild

src/main/java/uk/gov/pay/connector/filters/LoggingFilter.java
```diff
@@ -0,0 +1,34 @@
...
+        } catch (Throwable throwable){
+            logger.error("Exception - connector request - "+ requestURL + " - exception - "+ throwable.getMessage(), throwable);
+        }
+        finally {
+            logger.info("End - connector request - " + requestURL);
+        }
+    }
+
+    @Override
+    public void destroy() {}
```
kakumara: Can this ever happen? How about a log message to indicate that in case if it happens ? (big/bold)

src/test/java/uk/gov/pay/connector/filters/LoggingFilterTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Captor
+    ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor;
+
+    @Before
+    public void setup() {
+        loggingFilter = new LoggingFilter();
+        Logger root = (Logger) LoggerFactory.getLogger(LoggingFilter.class);
+        mockAppender = mockAppender();
+        root.addAppender(mockAppender);
```
kakumara: nice 👍

src/main/java/uk/gov/pay/connector/filters/LoggingFilter.java
imranbohoran: Not sure if this will happen, but as discussed I'll add a warn in-case it happens as the container would call this. Will do this as part of other PRs as we are changing logging.

PR #166, opened 2016-05-16
PP-541 Remove status from API results

ijmad:
## WHAT

Now that state has been added everywhere, and is what publicapi now consumes,
we can remove the old-style external status field from the charges API.
## HOW

Remove LegacyChargeState and everything associated with it.

However, for now, we leave it in frontend, because frontend consumes the internal
statuses from the connector, so these must still be present for frontend to
function, and frontend does not yet read states. We can fix this later.

As a result of this, I've split ChargeResourceBuilder in to separate api/frontend
versions that produce appropriate fields.
## WHO

@kakumara because this story overlaps with PP-547 extensively.


kakumara:
Are you going to do the required frontend changes as part of this story? Or will there be a separate story?

kakumara:
other than that 👍

PR #169, opened 2016-05-17
PP 335 Pagination and Navigation hal links in charges search [Review Only, Do Not Merge]

rahul-vinayak:
- HalResponseBuilder class is added using which you can create a HAL
  representation for your HATEOS resource. (prev/next links working conditionally)
- Page defaults to 1 and display_size defaults to 500


rahul-vinayak:
rebuild

rahul-vinayak:
rebuild

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -74,14 +76,15 @@
             AUTHORISATION_SUBMITTED,
             AUTHORISATION_SUCCESS);
 
-    private static final Logger logger = LoggerFactory.getLogger(ChargesResource.class);
+    private static final Logger logger = LoggerFactory.getLogger(ChargesApiResource.class);
 
     @Inject
-    public ChargesResource(ChargeDao chargeDao, GatewayAccountDao gatewayAccountDao, ChargeService chargeService, CardCancelService cardCancelService) {
+    public ChargesApiResource(ChargeDao chargeDao, GatewayAccountDao gatewayAccountDao, ChargeService chargeService, CardCancelService cardCancelService, ConnectorConfiguration configuration) {
```
kakumara: 👍  renaming this class is a good idea

kakumara: Why not just get the `TransactionsPaginationServiceConfig` object here instead of the whole connector config here. We only need it.

src/main/java/uk/gov/pay/connector/app/TransactionsPaginationServiceConfig.java
```diff
@@ -0,0 +1,12 @@
+package uk.gov.pay.connector.app;
+
+import io.dropwizard.Configuration;
+
+public class TransactionsPaginationServiceConfig extends Configuration {
```
kakumara: Do we expect this config to grow in future? Coz there is only 1 config attrib here and can just keep it flat in the main configuration rather than wrapping in a complex object.

ijmad:
Can we squash these commits down in to sensible chunks before merge please, and get rid of the WIP commits.

e640bde36c9816dc0ab4166a324373135128e7db and 5ac1a70bc2319d9b0a97b30e597426a384de876f are master merges, instead we should rebase on master to bring the branch up to date before merge. You can rewrite history to fix this.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -114,7 +125,7 @@ public ChargeEntity mergeAndNotifyStatusHasChanged(ChargeEntity chargeEntity) {
         if (params.getGatewayAccountId() != null)
             predicates.add(cb.equal(charge.get(GATEWAY_ACCOUNT).get("id"), params.getGatewayAccountId()));
         if (params.getReference() != null)
-            predicates.add(cb.like(charge.get(REFERENCE), '%'+params.getReference()+'%'));
+            predicates.add(cb.like(charge.get(REFERENCE), '%' + params.getReference() + '%'));
```
kakumara: Might not be part of this PR, but is this `SQL Injection` safe?

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -83,13 +84,32 @@ public Long getDisplaySize() {
...
+            builder.append("&from_date=" + fromDate);
+        if (toDate != null)
+            builder.append("&to_date=" + toDate);
+        if (page != null)
+            builder.append("&page=" + page);
+        if (displaySize != null)
+            builder.append("&display_size=" + displaySize);
+        if (externalChargeStates != null && !externalChargeStates.isEmpty())
+            for (ExternalChargeState extState : externalChargeStates) {
+                builder.append("&state=" + extState.getStatus());
```
kakumara: This looks like this can have a url param like `&state=STATE_1&state=STATE_2&state=STATE3`. Repeating the same query param is not a good idea. If we need multiple states you can follow either `,` separated style or array style `[,,,]` for multiple values. See followings 
https://github.com/gocardless/http-api-design
http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -13,9 +14,10 @@
     private String reference;
     private ZonedDateTime fromDate;
     private ZonedDateTime toDate;
-    private Long page = 1L;
-    private Long displaySize = 100L;
-    private List<ChargeStatus> chargeStatuses = new LinkedList<>();
+    private Long page;
+    private Long displaySize;
+    private List<ChargeStatus> chargeStatuses = new ArrayList<>();
```
kakumara: minor: It might be safer to have a `Set` here instead of a `List` given that will avoid duplicates as its used in a `IN` query. (Although we currently may not have a duplicate mappings)

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -101,16 +104,27 @@ public Response getChargesJson(@PathParam(ACCOUNT_ID) Long accountId,
                                    @QueryParam(STATE_KEY) String state,
```
kakumara: This assumes that you are getting only 1 state? but somewhere there was an append of `&status=` multiple times. May be something is not in sync?

src/main/java/uk/gov/pay/connector/resources/ChargesPaginationResponseBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+    public ChargesPaginationResponseBuilder withTotalCount(Long total) {
+        this.totalCount = total;
+        return this;
+    }
+
+    public Response buildResponse() {
+        if (totalCount > 0) {
+            double lastPage = Math.ceil(new Double(totalCount) / searchParams.getDisplaySize());
+            if (invalidPageRequest(lastPage)) {
+                return notFoundResponse("the requested page is not found");
```
kakumara: minor: perhaps drop 'is' from the message.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -101,16 +104,27 @@ public Response getChargesJson(@PathParam(ACCOUNT_ID) Long accountId,
...
                 .orElseGet(() -> reduce(validateGatewayAccountReference(gatewayAccountDao, accountId)
-                        .bimap(handleError, listCharges(accountId, reference, state, fromDate, toDate, pageNumber, displaySize, jsonResponse()))));
+                        .bimap(handleError,
+                                listCharges(new ChargeSearchParams()
+                                        .withGatewayAccountId(accountId)
+                                        .withReferenceLike(reference)
+                                        .withExternalChargeState(parseState(state))
+                                        .withFromDate(parseDate(fromDate))
+                                        .withToDate(parseDate(toDate))
+                                        .withDisplaySize(displaySize != null && displaySize >= 1 ?
```
kakumara: Is there a condition to error here? In case if caller sends a <0 should it be a HTTP:400 instead of default page size?

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -122,41 +136,50 @@ public Response getChargesCsv(@PathParam(ACCOUNT_ID) Long accountId,
...
-                            .withFromDate(parseDate(fromDate))
-                            .withToDate(parseDate(toDate))
-                            .withDisplaySize(displaySize)
-                            .withPage(page)
-            );
-            return responseFunction.apply(charges);
-        };
+    private F<Boolean, Response> listCharges(ChargeSearchParams searchParams, UriInfo uriInfo) {
+        List<ChargeEntity> charges = chargeDao.findAllBy(searchParams);
+        Long totalCount = chargeDao.getTotalFor(searchParams);
```
kakumara: Possible Optimisation here? If you swap these calls to get total first, then we can validate if the requested page can be achieved using total, display size and requested page number. In case requested page is wrong or way too off we can simply error instead of querying database to load the charges.

src/main/java/uk/gov/pay/connector/resources/HalResourceBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+import black.door.hate.HalResource;
+import com.fasterxml.jackson.core.JsonProcessingException;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class HalResourceBuilder implements HalResource {
+    public static final String SELF_LINK_KEY = "self";
```
kakumara: 👍

src/test/java/uk/gov/pay/connector/dao/ChargeSearchParamsTest.java
```diff
@@ -0,0 +1,51 @@
...
+                .withGatewayAccountId(111L)
+                .withPage(-1L)
+                .withReferenceLike("ref")
+                .withFromDate(ZonedDateTime.parse(FROM_DATE))
+                .withToDate(ZonedDateTime.parse(TO_DATE));
+
+        assertEquals("query params string mismatch", format(EXPECTED_QUERY_STRING, 1, 500), params.buildQueryParams());
+
+    }
+}
```
kakumara: new line

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -69,12 +71,11 @@ public ChargeSearchParams withToDate(ZonedDateTime toDate) {
...
 
     public Long getPage() {
-        return page - 1; // 1 based page and 0 based offset
+        return page;
     }
 
     public ChargeSearchParams withPage(Long page) {
-        if (page != null && page >= 0)
-            this.page = page;
+        this.page = (page != null && page >= 1) ? page : 1; // always show first page if its an invalid page request
```
kakumara: Can this ever happen? If we validate page number to be not negative at resource level, we don't have to this assignment. If this happens it should be an Exception.

src/test/java/uk/gov/pay/connector/dao/ChargeSearchParamsTest.java
```diff
@@ -0,0 +1,51 @@
...
+                .withExternalChargeState(singletonList(EXTERNAL_CREATED))
+                .withGatewayAccountId(111L)
+                .withPage(-1L)
+                .withReferenceLike("ref")
+                .withFromDate(ZonedDateTime.parse(FROM_DATE))
+                .withToDate(ZonedDateTime.parse(TO_DATE));
+
+        assertEquals("query params string mismatch", format(EXPECTED_QUERY_STRING, 1, 500), params.buildQueryParams());
+
+    }
```
kakumara: You sure there's no more cases to test here? `ChargeSearchParams` seems like a complex builder and some more stuff going on there (e.g. statuses) that we can test.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -305,6 +306,142 @@ public void shouldFilterTransactionsBasedOnFromAndToDates() throws Exception {
...
+        addCharge(CAPTURED, "ref-4", now().minusDays(3));
+        addCharge(CAPTURED, "ref-5", now().minusDays(4));
+        String chargesPath = CHARGES_API_PATH.replace("{accountId}", accountId);
+
+        assertResultsAndJustSelfLinkWhenJustOneResult(chargesPath);
+        assertResultsAndNoPrevLinkWhenOnFirstPage(chargesPath);
+        assertResultsAndAllLinksWhenOnMiddlePage(chargesPath);
+        assertResultsAndNoNextLinksWhenOnLastPage(chargesPath);
+        assert404WhenRequestingInvalidPage(chargesPath);
+    }
```
kakumara: 1. How about a test case for <0 page number?
2. Can we request for a larger display size than the total count? In that case how will the result look like?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -305,6 +306,142 @@ public void shouldFilterTransactionsBasedOnFromAndToDates() throws Exception {
...
+    private void assert404WhenRequestingInvalidPage(String chargesPath) {
+        // when 5 charges are there, page is 10, display-size is 2
+        ValidatableResponse response = getChargeApi
+                .withAccountId(accountId)
+                .withQueryParam("reference", "ref")
+                .withQueryParam("page", "10")
+                .withQueryParam("display_size", "2")
+                .withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON)
+                .getTransactions()
+                .statusCode(NOT_FOUND.getStatusCode());
```
kakumara: minor: in addition to states code, probably to good to check the body is empty.

src/test/java/uk/gov/pay/connector/resources/HalResourceBuilderTest.java
```diff
@@ -0,0 +1,46 @@
...
+                        "}" +
+                "}",
+                result);
+    }
+
+    private URI uriOf(String link) throws URISyntaxException {
+        return new URI(link);
+    }
+
+}
```
kakumara: new line

src/main/java/uk/gov/pay/connector/app/TransactionsPaginationServiceConfig.java
rahul-vinayak: I thought that was the pattern we were following with other configs (with just 1 attrib) to have a wrapper to make it more specific, but surely can look at moving it in main.

kakumara: You may be right here. Can see another config with just one value (`links`). May be we can leave it. But opportunity to discuss at a tech huddle. Do we need a structure if we have only 1 config value?

rahul-vinayak:
@ijmad I have done a squash -> cherry-pick -> rebase of the branch, have a look at the new PR for the new branch at https://github.com/alphagov/pay-connector/pull/173 and see if it suffices

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
rahul-vinayak: I cant make out how it would allow someone to put in a parameter that would make it vulnerable to security. If you are pointing towards %like predicate then yes you cant stop anyone from putting an 'a' or 'b' and get all references matching it but thats whats the requirement is, and thats not referred to as sql injection. 
Can you elaborate a bit

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
rahul-vinayak: Yes I am a bit unable to see the realistic scenario how it would end up with duplicate states, and even in theory if it does, then how passing a duplicates in an 'IN' query be of a concern

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
rahul-vinayak: Can this ever happen? - Yes
This is build up at resource level, a missing page param is set as null, and if its missing the page=1 should be shown by default. 
There will be slight change to handle -ve numbers now (less than one page or display_size results in 400 now)

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
kakumara: this is coz we are not param binding, it just looks like a string concatenation. Perhaps predicates or this `cb` variable internally doing it? It's worthwhile to know.

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
kakumara: It just to avoid the possibility of a SQL like `SELECT blah FROM blah WHERE blan IN(V1, V1, V1)` which normally wouldn't write. a `Set` would simply avoid it (even in future).

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
rahul-vinayak: there are different school of thoughts on that one, rest as we all know in www follows &state=STATE_2&state=STATE3 pattern, and hence few people suggest that as it automatically gets assembled into an array. Having said that this wrong that its accepting multiple states, as it crept up while the new lifecycle story was merged and modified this piece. I have now fixed it so that it accepts only one state and hence builds params with one state only

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
rahul-vinayak: Thats a fair point. Yes this was wrong that its accepting multiple states, as it crept up while the new lifecycle story was merged and modified this piece. I have now fixed it so that it accepts only one state and hence builds params with one state only.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
rahul-vinayak: The way I feel it is that a negative value would be either a typo or an error and I have tried to handle it the same way a negative page number is handled, i.e. fall back to defaults. The response will say the display_size=500 which just saves a round trip. Can discuss with @ijmad what he thinks on this.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
rahul-vinayak: Fair point.
Will try and swap the logic

src/test/java/uk/gov/pay/connector/dao/ChargeSearchParamsTest.java
rahul-vinayak: As per the above changes there will only be one state be in the query param  (not an array any more) at a time, hence I think the test covers the scenario now.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
rahul-vinayak: So after a discussion, the nulls will be handled same way as to it will fall back to defaults, however -ve values will lead to bad_request, will be changing the code in the new branch PP-335-cherrypicker

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
rahul-vinayak: Added tests for page <1 and display_size<1.
Also modified code to throw bad_request in both of the above cases, and modified validator.
Can we request for a larger display size than the total count? In that case how will the result look like? - Yes, its just like not setting display_size, as it will be set to 500 by default, and you will get upto 500 results per page

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
rahul-vinayak: done

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
rahul-vinayak: Now it does a get total before findAll and returns 404 if requested page cant be served..

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
rahul-vinayak: done

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
rahul-vinayak: What I meant was IN(V1,V2,V1,V3) would still work even if duplicates would crop up. I have made it a Set if that if that would make a difference.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
rahul-vinayak: As discussed its not the String concatenation that poses the SQL injection issue but the user input going directly into a sql String that can make a query vulnerable to SQL Injection. As the criteria builder with JPA takes care of escaping all the vulnerable characters this is not SQL injection vulnerable. I have written a test which explores this vulnerability a test called aBasicTestAgainstSqlInjection() in ChargesDaoITest.java which backs up my claim.

rahul-vinayak:
All the review comments have been addressed in the branch PP-335-cherrypicker and is part of this PR 
https://github.com/alphagov/pay-connector/pull/173

PR #172, opened 2016-05-18
PP-712 Investigate connector db connection pool and entity manager problems

PauloPortugal:
## WHAT

_Incident report - https://docs.google.com/document/d/1i0LYBAkaQ98Yzti9f6O0T61mCurIh4vLkl-uPD_wC3A/edit_
- **Problem 1:**
  In the integration we were getting inconsistent results for charge status and charge events. This was happening while the database healthcheck was running.
  The database healthcheck has been disabled which seem to have resolved the problem. 
  We need to do investigate what the root cause is.
- **Problem 2:**
  The test environment became non-responsive and when a thread dump was enforced it was noticed that we were exhausting our connection pool. 
  We need to investigate why/what exhausted the connection pool and come up with more healthchecks to enable monitoring similar issues.
## HOW

_Steps to test or reproduce:_
- There is a problem on how Guice Persist manages units of work.
- Requesting an entity manager outside of a unit of work, like we
  were doing with the HealthCheck, Guice Persist will implicitly
  start the unit of work for you, but because we are not telling it
  to terminate the unit of work or the method is not annotated with
  `@Transactional`, which will enforce a termination of the unit of
  work once existing the method, the entityManager that is responsible
  for dealing with that request at that moment falls into a "limbo".
  state and can never flush any cached entities so they can be reloaded again.
- As a side note, the JPAPersistenceService manages a collection of
  EntityManagers (making them available to both  Jersey and the Dropwizard
  Healthcheck Registry) so depending on the entity Manager that was
  responsible to handle the query we would see different results.
- We will provide some additional information about the current
  state of the connection pool. There are quite a few more things
  that can be added (distinguish between `idle` and `active`
  connections, what type of monitoring system will consume these
  stats, etc. This can be done a later stage once the work has been
  prioritised.
- Hit the `/healthcheck` endpoint to check the additional DB stats
## WHO

- [x] Developers
- [ ] WebOps


src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
```diff
@@ -1,27 +1,52 @@
...
-//        entityManager.get().createNativeQuery(validationQuery).getSingleResult();
-        return Result.healthy();
+        ResultSet resultSet = null;
+        Connection connection = null;
+        try {
+            connection = DriverManager.getConnection(
+                configuration.getDataSourceFactory().getUrl(),
+                configuration.getDataSourceFactory().getUser(),
+                configuration.getDataSourceFactory().getPassword());
+            resultSet = connection.createStatement().executeQuery("SELECT count(*) FROM pg_stat_activity");
```
kakumara: My gut feel says this should be a really light weight query (e.g. `SELECT 1`) to have minimum/no overhead on database simply because we are running this against production DB and at every `few` seconds. All we want I thought was to see if the DB connectivity available from connector. 
For PG monitoring, i'm pretty sure there must be more native tools (or amazon RDS should have some tools already we can hook in to) rather than doing via java/connector layers. 

I strongly feel we need a tech discussion of `what monitoring we need in this area and how`, including @joshmyers, @bazbremner @ijmad before doing too many fancy things. I'm not too hung up merging this as it is if this has already been discussed several times.

src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
```diff
@@ -1,27 +1,52 @@
...
+            resultSet.next();
+        } catch (Exception e) {
+            return Result.unhealthy(e.getMessage());
+        } finally {
+            if (connection !=null) {
+                connection.close();
+            }
+        }
+        return Result.healthy(new ObjectMapper().writeValueAsString(getStats(resultSet)));
```
kakumara: Are you sure this is working? By this time you've already closed the connection and any associated objects. I thought both Statement/ResultsSet will close when the parent connection is closed. Suggest to put this inside try block.

see http://stackoverflow.com/questions/25493837/java-cant-use-resultset-after-connection-close

kakumara:
other than that 👍

src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
PauloPortugal: @kakumara I agree with you. Initially I used the validationQuery which is `SELECT 1`, but after speaking with @imranbohoran I ended up providing a bit more info. @imranbohoran plan was to have another endpoint with the connector connection-pool stats, that would only be accessible from within the box, or that would be consumed by a monitoring tool (nagios, sensu) but since we have not had a proper discussion with architects, business and infrastruture it seemed premature to move into that direction without prioritizing it.

Since @bazbremner is not here I would have that discussion with the people mentioned above plus @imranbohoran and come with a conclusion.

src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
PauloPortugal: It is working :), but you are right I forgot to close the resultSet.

PauloPortugal:
rebuild

PR #174, opened 2016-05-19
PP-712 Configuring eclipselink session

imranbohoran:
## WHAT

Configures the eclipselink session so that we don't retry for longer periods. This gives a bit more quicker feedback to the client code about any connection failures.
- Adding a eclipselink session customiser that configures the number of
  retrys and the duration between retries.
- The number of retries is 0 based so for 1 retry we have to configure it
  to be 0.
- The duration between retries is reduced to 2 seconds. This is
  5 seconds by default.
- Also reducing the validation query timeout.
## HOW

Run the connector locally and and kill the connector db. Then list some transactions. It would take few seconds and report 3 log entries about retrying (Internal Exception: org.postgresql.util.PSQLException: The connection attempt failed.) It's 3 because it logs for the 2 retries and as part of printing the stacktrace.
## WHO

- [x] Developers


PauloPortugal:
:+1:

PR #180, opened 2016-05-20
PP-335 Improvement while building search params for pagination

rahul-vinayak:
## WHAT
- Removes the leading ‘&’ from the searchParams while building it
- plus test changes


imranbohoran:
👍

PR #182, opened 2016-05-23
PP-702 Enforce/Fix CANCEL/EXPIRE status flow

kakumara:
## WHAT

This PR fixes the Charge state Transition flow around `USER_CANCEL`,  `SYSTEM_CANCEL` and `EXPIRE`  journeys. As per PP-702 state transition should enforce

```
[Allowed User Cancellable Statuses] -->  (USER_CANCELLED / USER_CANCEL_ERROR)) or additionally USER_CANCEL_READY if gateway cancellation required 

[Allowed System Cancellable Statuses] --> (SYSTEM_CANCELLED / SYSTEM_CANCEL_ERROR) or additionally SYSTEM_CANCEL_READY if gateway cancellation required 

[Allowed Expirable Statuses] --> (EXPIRED / EXPIRE_CANCEL_FAILED) or additionally EXPIRE_CANCEL_READY if gateway cancellation required 
```
## HOW

This refactors away the previous `Transaction Gateway Operation` approach. This introduces a dynamically composable  `TransactionFlow` with `n` number of `Transactional` and `NonTransactional` operations.

**Things to look:**
- `TransactionFlow.java`, `TransactionalOperation.java`, `NonTransactionalOperation.java`
- Usages in `ChargeCancelService.java`, `ChargeExpiryService.java`
  -  Introduced IntegrationTests which checks for side effects of all state transition transactions. (Events Changes)
- Removed previous cancel/expire related code.
- **All other code changes are around code that had to be touched upon due the above key changes**
## WHO
- [X] Developers
- [] WebOps


rahul-vinayak:
[Allowed Expirable Statuses] --> EXPIRE_CANCEL_PENDING --> ( EXPIRED / EXPIRE_CANCEL_FAILED)

Above does not stand true always. From [allowed expirable states] -> EXPIRE_CANCEL_PENDING happens only if the charge status to be expired was Authorised, else it goes straight to [Expired/Expiry-failed] states, worth making it clear here.

src/main/java/uk/gov/pay/connector/service/transaction/TransactionFlow.java
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * executes the given block of code in a Transactional boundary.
+     * @param op block of code to be executed
+     * @param <R> result to be persisted
+     * @return
+     */
+    @Transactional
+    public <R> TransactionFlow executeNext(TransactionalOperation<TransactionContext, R> op) {
+        Objects.requireNonNull(op);
+        R result = op.execute(context);
```
rahul-vinayak: Objects.requireNonNull just throws a NPE if param is null, was that what was expected out of it or is there something a miss here, cause NPE is going to be thrown by the next line even without requireNull check

kakumara: Yes. There is nothing can be done other than throwing the exception. The only thing `Objects.requireNonNull` gives you is that it reads nicely when debugging coz it clearly says this NPE is due to a expect input param being null.

src/main/java/uk/gov/pay/connector/resources/CardResource.java
```diff
@@ -75,13 +77,26 @@ private Response handleError(ErrorResponse error) {
             case GATEWAY_CONNECTION_SOCKET_ERROR:
                 return serviceErrorResponse(error.getMessage());
         }
+        /* FIXME: This state doesn't sound right here. Perticulalry when cancelling and
```
ckalista: Typo: particularly

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
```diff
@@ -0,0 +1,207 @@
...
+
+    private Function<GatewayResponse, ChargeStatus> determineTerminalState(ChargeStatus onSuccess, ChargeStatus onFail) {
+        return gatewayResponse -> gatewayResponse.isSuccessful() ? onSuccess : onFail;
+    }
+
+    private String getLegalStatusNames(ChargeStatus[] legalStatuses) {
+        return Stream.of(legalStatuses).map(ChargeStatus::toString).collect(Collectors.joining(", "));
+    }
+
+    static class CancellationStatusFlow {
```
ckalista: This is much cleaner 👍

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
```diff
@@ -0,0 +1,207 @@
...
+                .executeNext(finishCancel(determineTerminalState(statusFlow.getSuccessTerminalState(), statusFlow.getFailureTerminalState())))
+                .complete()
+                .get(CancelGatewayResponse.class));
+
+    }
+
+    private GatewayResponse nonGatewayCancel(ChargeEntity chargeEntity, CancellationStatusFlow statusFlow) {
+        ChargeStatus completeStatus = statusFlow.getSuccessTerminalState();
+        ChargeEntity processedCharge = transactionFlowProvider.get()
+                .executeNext((TransactionalOperation<TransactionContext, ChargeEntity>) context -> {
```
ckalista: Agree with the comment you made yourself during the quick tech huddle - would be a bit cleaner to extract this lambda out into its own Function.

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
```diff
@@ -0,0 +1,207 @@
...
+                    logger.info("charge status to update - from: " + chargeEntity.getStatus() + ", to: " + completeStatus + " for Charge ID: " + chargeEntity.getId());
+                    chargeEntity.setStatus(completeStatus);
+                    return chargeDao.mergeAndNotifyStatusHasChanged(chargeEntity);
+                }).complete()
+                .get(ChargeEntity.class);
+
+        if (completeStatus.getValue().equals(processedCharge.getStatus())) {
```
ckalista: Nothing you have introduced here but at some point we should refactor ChargeEntity.getStatus to return a ChargeStatus.

src/main/java/uk/gov/pay/connector/service/ChargeExpiryService.java
```diff
@@ -0,0 +1,174 @@
...
+
+        int expiredSuccess = expireChargesWithCancellationNotRequired(chargesToProcessExpiry.get(Boolean.FALSE));
+        Pair<Integer, Integer> expireWithCancellationResult = expireChargesWithCancellation(chargesToProcessExpiry.get(Boolean.TRUE));
+
+        return ImmutableMap.of(EXPIRY_SUCCESS, expiredSuccess + expireWithCancellationResult.getLeft(), EXPIRY_FAILED, expireWithCancellationResult.getRight());
+    }
+
+    private int expireChargesWithCancellationNotRequired(List<ChargeEntity> nonAuthSuccessCharges) {
+        List<ChargeEntity> processedEntities = nonAuthSuccessCharges.stream().map(chargeEntity -> transactionFlowProvider.get()
+                .executeNext((TransactionalOperation<TransactionContext, ChargeEntity>) (context -> {
```
ckalista: I  wonder if there is value to use the TransactionFlow at all here? Basically all we need is a function annotated @Transactional

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
```diff
@@ -0,0 +1,207 @@
...
+            String errorMsg = format("Could not update chargeId [%s] to status [%s]. Current state [%s]",
+                    processedCharge.getExternalId(),
+                    completeStatus,
+                    processedCharge.getStatus());
+            logger.error(errorMsg);
+            return new CancelGatewayResponse(FAILED, baseError(errorMsg), statusFlow.getFailureTerminalState());
+        }
+    }
+
+    private TransactionalOperation<TransactionContext, ChargeEntity> prepareForCancel(ChargeEntity chargeEntity, CancellationStatusFlow statusFlow) {
```
ckalista: We need to be really careful not to introduce regression issues hard to spot resulting in heretic behaviour in production. My concern is that not all transactional edge case scenarios are covered by automation tests. I would suggest to run through all the manual tests that were initially carried out when the transactional aspect was introduced.

For instance here if an OptimisticLockException is thrown during the prepareForCancel, we are not handling it specifically hence should it occur we would end up with an Internal Server Error while we would expect a 409 (CONFLICT). This is one of the advantage of the old approach with the TransactionalGatewayOperation enforcing behaviour and handling the OptimisticLockException accordingly in this scenario. 

Maybe we would need to reinstate some degree of control with the TransactionFlowProvider...Maybe a PreOperation functional interface extending TransactionalOperation which would be required during the composition?

Manual tests to conduct would be:

|  | Connector failure | Gateway failure | Database failure | Outcome |
| --- | --- | --- | --- | --- |
| prepareForCancel | RuntimeException |  |  | 500, transaction rollback |
| prepareForCancel | OptimisticLockEx |  |  | 409, transaction rollback |
| prepareForCancel |  |  | DB Down | 500, transaction rollback |
| doGatewayCancel | RuntimeException |  |  | 500, no transaction |
| doGatewayCancel |  | Cannot reach |  | 200, SYSTEM_ERROR |
| finishCancel | RuntimeException |  |  | 500, transaction rollback |
| finishCancel | OptimisticLockEx |  |  | 500, transaction rollback |
| finishCancel |  |  | DB Down | 500, transaction rollback |

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
```diff
@@ -0,0 +1,207 @@
...
+                .executeNext(doGatewayCancel())
+                .executeNext(finishCancel(determineTerminalState(statusFlow.getSuccessTerminalState(), statusFlow.getFailureTerminalState())))
+                .complete()
+                .get(CancelGatewayResponse.class));
+
+    }
+
+    private GatewayResponse nonGatewayCancel(ChargeEntity chargeEntity, CancellationStatusFlow statusFlow) {
+        ChargeStatus completeStatus = statusFlow.getSuccessTerminalState();
+        ChargeEntity processedCharge = transactionFlowProvider.get()
```
ckalista: I would question the need to use the TransactionFlow at all here - a simple method annotated @Transactional would do.

src/main/java/uk/gov/pay/connector/service/ChargeExpiryService.java
```diff
@@ -0,0 +1,174 @@
...
+                logger.error("ChargeEntity with id {} returned with unexpected status {} during expiry", chargeEntity.getExternalId(), chargeEntity.getStatus())
+        );
+
+        return Pair.of(
+                expireCancelled.size(),
+                expireCancelFailed.size()
+        );
+    }
+
+    private TransactionalOperation<TransactionContext, ChargeEntity> prepareForExpireCancel(ChargeEntity chargeEntity) {
```
ckalista: Same comment as above, careful with OptimisticLockExceptions - unless mistaken I don't think they are being handled properly.

ckalista:
Code review completed. A lot of good stuff in there 👍

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
kakumara: Moved all functions to a reusable class

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
kakumara: I haven't changed the core logic of how it was at all in this story. It was merely fixing the state transitions and transaction refactoring. The remains exactly how it was. But I do agree that `ChargeEntity.getStatus` probably should return a `ChargeStatus`

src/main/java/uk/gov/pay/connector/service/ChargeExpiryService.java
kakumara: Its entirely possible. But in this case I prefer to keep the transactionFlow for
1. Uniformity. (This means we possibly can avoid having @Transactional in all our resources/service classes (in theory) and a transactional function now simply a temporary prototypal object available on demand.)
2. No need to make private methods public in order to get it @Transactional.
3. Transaction flow is a really lightweight object. Pretty much immutable too. And it doesn't clutter the code too much as far as I'm concern. 
4. (Futuristic) this is purely hypothetical. Imagine if we have a requirement if we need to do something special around transactional operations, some common aspect (e.g. extra logging, notifications, audits, etc...)   we will have one place instead of 10 different places and this will be really handy.

Happy to discuss if you feel strongly otherwise.

src/main/java/uk/gov/pay/connector/service/ChargeExpiryService.java
```diff
@@ -0,0 +1,174 @@
...
+        return context -> {
+            ChargeEntity reloadedCharge = chargeDao.merge(chargeEntity);
```
rahul-vinayak: I know this is a pattern we are following at a few places, however there is no point in doing a merge even before making any change, it would just make an unnecessary DB call, if we want to check for DB version change that would happen anyways when we call mergeAndNotifyStatusHasChanged at the end.

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
kakumara: Good point on OptimisticLockException. As discussed I'll add a new marker for operations that require handling conflicts

kakumara: I will also do some manual tests around the scenarios given

src/main/java/uk/gov/pay/connector/service/ChargeCancelService.java
kakumara: Similar to above.

src/main/java/uk/gov/pay/connector/service/ChargeExpiryService.java
rahul-vinayak: I see prepareForXXX method being used in CancelService as well as ExpiryService, that looks identical, and I believe it would be identical for most of the operations if not all, cant the idea of this framework be to move out most of the common functionalities be pulled out, so you just set the flow like you did with CancellationStatusFlow (where you define allowed and locking states), and rest your framework handles and manages common set of code for prepare, do, finish like we do in cancel and expire atm. This common functionality approach was what was happening previously and we should use that aspect here as well.

kakumara: 👍  as discussed handling with a `PreTransactionalOperation` marker

kakumara: @rahul-vinayak yes, could be done. I'm moving all reusalble functions to a common class.

src/main/java/uk/gov/pay/connector/service/transaction/TransactionFlow.java
rahul-vinayak: No it wont, unless you pass in the message as an argument it wont show you a friendly readable message.

src/test/java/uk/gov/pay/connector/util/DatabaseTestHelper.java
```diff
@@ -197,4 +198,13 @@ public void addEvent(Long chargeId, String chargeStatus) {
                         chargeId, chargeStatus)
         );
     }
+
+    public List<String> getInternalEvents(String externalChargeId) {
+        return jdbi.withHandle(h ->
+                h.createQuery("SELECT status from charge_events WHERE charge_id = (SELECT id from charges WHERE external_id=:external_id)")
+                        .bind("external_id", externalChargeId)
```
rahul-vinayak: Any specific reason its a subquery and not a join

kakumara: No real reason. Either will do the job. Plus this is for test assertions only. I'm not too specific, but can change if you really think its important.

kakumara:
rebuild

kakumara:
rebuild

PR #183, opened 2016-05-23
PP 335 display size defaults fix

rahul-vinayak:
- display_size and page should be set to default values page=1, display_size=500(coming from config)
- added test for this scenario in connector
- was not able to read the default configuration from config, now fixed


imranbohoran:
👍

PR #184, opened 2016-05-26
PP-673 GET events - returns 404 for non existent charge-id

rahul-vinayak:
- GET events now returns 404 for a non existent charge-id as opposed to
  a 200 with empty list of events, which was getting passed as is to PublicApi
- and a test to cover that

with @MahmudH


rahul-vinayak:
Can be moved to story review/WFSO once connector and e2e gets merged

imranbohoran:
👍

PR #186, opened 2016-05-27
PP-702 migrate legacy status to new status

kakumara:
## WHAT

One of the legacy statuses which is already in integration environment causing problems in paginations.
This is the fix for that
## HOW

Migrated  
- `EXPIRE CANCEL PENDING` --> `EXPIRE CANCEL READY`
- `CANCEL READY` --> `SYSTEM CANCEL READY`
- `CANCEL ERROR` --> `SYSTEM CANCEL ERROR`
## WHO
- [X] Developers
- [] WebOps


mcgoooo:
:+1:

PR #188, opened 2016-06-01
PP 694 ammends the 

rahul-vinayak:
- this changes the default time before the spinner auth awaiting is shown to user to 1 sec
- this figure can be overridden using AUTH_READ_TIMEOUT_SECONDS env var
- updated test configs to be inline with the real config
- tests around AUTH read timeout
- SocketTimeoutException: Read Timeout increased to 90 seconds for call to provider for AUTH


src/main/java/uk/gov/pay/connector/app/CustomJerseyClientConfiguration.java
```diff
@@ -7,7 +7,7 @@
 import javax.validation.Valid;
 
 public class CustomJerseyClientConfiguration extends Configuration {
-    private Duration readTimeout = Duration.milliseconds(20000L);
+    private Duration readTimeout = Duration.milliseconds(90000L);
```
PauloPortugal: So the yaml config variable is not being read? We use an hardcoded value?
Could we not make use of yaml one?

src/main/java/uk/gov/pay/connector/service/ClientFactory.java
```diff
@@ -43,8 +43,7 @@ public Client createWithDropwizardClient(String name) {
                 .using(new ApacheConnectorProvider())
                 .using(clientConfiguration)
                 .withProperty(ClientProperties.READ_TIMEOUT, readTimeoutInMillis)
-                .withProperty(ApacheClientProperties.CONNECTION_MANAGER, createConnectionManager())
-        ;
+                .withProperty(ApacheClientProperties.CONNECTION_MANAGER, createConnectionManager());
```
PauloPortugal: :+1:

PauloPortugal:
Seems like we are not using the jersey read timeout config variable, but instead using a hardcoded one. If we could change that it would be great. Other than that seems :+1:

src/main/java/uk/gov/pay/connector/app/CustomJerseyClientConfiguration.java
rahul-vinayak: Actually it reads from config, and the value from config takes precedence over this. This was there although I think its not needed but might have been done in the first place as a fall back if config is not provided or something.

rahul-vinayak:
It does actually read it from the config and the value from config takes precedence over this hard coded thingy. This was there although I think its not needed but might have been done in the first place as a fall back if config is not provided or something.

rahul-vinayak:
I am gonna remove this hard coded duration as its not needed, and will end up in confusions.

PauloPortugal:
:+1:

PR #190, opened 2016-06-06
PP-447 Adds optional email field to charge entity

mrlumbu:
## WHAT

This enables us to store an email against a charge id. This has been done by actually adding email to charge object, as for each charge there will be one and only one email. This work:
- Adds email column to charge table
- Adds email field to charge entity
- Validates email if present when charge created

Email is currently optional, to ensure backwards compatibilty.

This work does _not_ yet enable patching an email on to existing charge
## HOW

_Steps to test or reproduce:_
Check tests cover all scenarios where email is/is not present on charge, is/is not valid
## WHO

- [x] Developers
- [] WebOps


src/main/java/uk/gov/pay/connector/resources/ApiValidators.java
```diff
@@ -54,4 +81,17 @@
         }
         return right(true);
     }
+
+    public static Optional<List<String>> validateChargeParams(Map<String, Object> inputData) {
```
imranbohoran: 👍

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -44,12 +44,14 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ConnectorConfigurat
 
     @Transactional
     public ChargeResponse create(Map<String, Object> chargeRequest, GatewayAccountEntity gatewayAccount, UriInfo uriInfo) {
-
+        String email = chargeRequest.get("email") != null ? chargeRequest.get("email").toString() : null;
```
imranbohoran: This is not necessarily a comment on this PR. But was thinking that this would be a good opportunity to refactor. The Map objects passed in are typed as Map<String, Object>, although the values are really all String. This results in using toString() all over the code where this map is used.
If we refactor this to be Map<String, String> all the ugly toString() can be removed. And also this line could be replaced as Optional.ofNullable(chargeRequest.get("email")).orElseGet(() -> null);
A quick check of making this Map<String, String> and fixing compilation failures makes all tests go green. 
Like I said, its not necessarily part of this PR but an opportunity to make it better. Up to you if you want to address it as part of this PR, a different PR or not as part of the work for this story.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -269,6 +324,8 @@ public void shouldFilterTransactionsBasedOnFromAndToDates() throws Exception {
         datesFrom(createdDateStrings).forEach(createdDate ->
                 assertThat(createdDate, is(within(1, DAYS, now())))
         );
+        List<String> emails = collect(results, "email");
+        assertThat(emails, contains(email, email));
```
imranbohoran: 👍

imranbohoran:
LGTM. Pending comment/refactor based on the comment on the properly typed Map, this is good to merge.

src/main/java/uk/gov/pay/connector/service/ChargeService.java
mrlumbu: Yep good point. There is another PR associated with this story, so I might do it in that instead of here?

imranbohoran: 👍  I'll merge this one then.

PR #191, opened 2016-06-07
Pp 447 patch email

mrlumbu:
## WHAT

2 bits to this:
- A user may wish to add an email to a charge after the charge is created.
  This work adds a patch method to the charges/{chargeId} endpoint to allow this.
  At the moment the only patch operation allowed is "replace" and the only
  patchable field on the Charge entity is the email field.
- Refactor all Map request objects to be of type String, String, and remove a bunch of toString()'s
## HOW

_Steps to test or reproduce:_
Check tests cover all patch scenarios.
## WHO

- [x] Developers
- [] WebOps


src/main/java/uk/gov/pay/connector/model/ChargePatchRequest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+
+import static uk.gov.pay.connector.model.PatchRequest.PatchField.*;
+
+public class ChargePatchRequest extends PatchRequest {
+    private static final List<String> patchableFields = Arrays.asList(ChargesApiResource.EMAIL_KEY);
+    private static final  List<String> validOps = Arrays.asList("replace");
```
imranbohoran: I feel these is best provided as a configuration or initialisation of the patch request object. If you have a PatchRequestBuilder and build it with the necessary patch fields and valid operations, we wouldn't need the `isValid(String)` and `isValidPath(String) overrides as well.
We could even get rid of the whole ChargePatchRequest class. What do you think?

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -130,7 +129,7 @@ public Response getChargesJson(@PathParam(ACCOUNT_ID) Long accountId,
     @POST
     @Path(CHARGES_API_PATH)
     @Produces(APPLICATION_JSON)
-    public Response createNewCharge(@PathParam(ACCOUNT_ID) Long accountId, Map<String, Object> chargeRequest, @Context UriInfo uriInfo) {
+    public Response createNewCharge(@PathParam(ACCOUNT_ID) Long accountId, Map<String, String> chargeRequest, @Context UriInfo uriInfo) {
```
imranbohoran: 👍  properly typed Maps FTW

src/main/java/uk/gov/pay/connector/model/ChargePatchRequest.java
mrlumbu: hmm ok, sounds good. Might need to check in with you about best way of doing that

imranbohoran:
👍 For the builder. One last minor comment. There's an unused variable at line 40 of ChargeFrontendResource.

imranbohoran:
👍

PR #195, opened 2016-07-05
PP-822 Changed email size from 50 to 254 characters

rauligs:
## WHAT
- As per definition in several RFCs. See https://en.wikipedia.org/wiki/Email_address for more information.
- Migrated column email in charges table from 50 characters to 254 characters.
## WHO
- [X] Developers
- [X] WebOps


ckalista:
Code review completed.

PR #197, opened 2016-07-10
PP-448 Send payment confirmation emails using GOV.UK notify

rahul-vinayak:
- This relies on the pay-notify java client jar which is not published
  yet, so is built locally and loaded from the source folder as a maven dependency instead till we have it
  published.
- Passes in the configuration from the environment variables to set up
  the client to make a call to notify to send email confirmation message
- These emails are atm restricted to be internal domain emails only by
  pay-notify
- Email notifications can be disabled using the environment variable
  “NOTIFY_EMAIL_ENABLED”, set it to true to enable sending emails,
  defaults to false.
- These emails are atm restricted to be internal domain emails only by
  pay-notify
- allows you to pass  personalisation of the template that you choose.

To test sending an email post confirmation, you would need to set following env vars:
NOTIFY_EMAIL_TEMPLATE=1a1df741-41ca-4fa4-b74a-4cea9e97cb60
NOTIFY_SERVICE_ID=7ef48fbb-55cf-4410-87bc-fc583b5c52a2
NOTIFY_SECRET=aaca7269-e3e9-443b-adde-4ad3b3601659
NOTIFY_BASE_URL=https://api.notifications.service.gov.uk
NOTIFY_EMAIL_ENABLED=true

above will send email using the dummy template I have created.


davidillsley:
Do we have a ticket open with Notify to get the jar published?

rahul-vinayak:
@davidillsley I believe Dominic had a word with Rebecca Law and as per her it is on their plans to have the client Jar published at some point. Not sure if there is a ticket raised though

rauligs:
I don't see any story regarding the jar to be published. I'll create one just in order to make the relevant changes for when this is published (Notify is already aware of this).
[PP-871](https://payments-platform.atlassian.net/browse/PP-871)

src/main/java/uk/gov/pay/connector/service/NotifyClientProvider.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public GovNotifyApiClient get() {
+        ClientConfiguration config = new ClientConfiguration.Builder()
+                .serviceId(readServiceIDConfig())
+                .baseUrl(readBaseURLConfig())
+                .secret(readSecretConfig())
+                .build();
+
+//        DropwizardGovUkHttpClient client = new DropwizardGovUkHttpClient(clientFactory);
```
rauligs: Remove comments

src/main/java/uk/gov/pay/connector/service/NotifyClientProvider.java
```diff
@@ -0,0 +1,54 @@
...
+        ApacheGovNotifyHttpClient client = new ApacheGovNotifyHttpClient();
+        return new GovNotifyApiClient(config, client);
+    }
+
+    private String readServiceIDConfig() {
+        if (StringUtils.isBlank(configuration.getNotifyConfiguration().getServiceId())) {
```
rauligs: This conditions on loading configuration itself is better and fail-fast. This class must assume the configuration is correct

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Inject
+    public UserNotificationService(NotifyClientProvider notifyClientProvider, ConnectorConfiguration configuration) {
+        this.notifyClientProvider = notifyClientProvider;
+        readEmailConfig(configuration);
+    }
+
+    public Optional<String> notifyPaymentSuccessEmail(String emailAddress) {
+        if (!emailNotifyEnabled) {
+            logger.warn("Email notifications is disabled by configuration");
```
rauligs: info is more appropriated since they are variables controlled by us. If we expect this to be a warning it sounds like we don't expect at all to be false. 
Where is email enabling in the AC? Is this a backward compatibility thing?

src/main/java/uk/gov/pay/connector/service/userNotifications/DropwizardGovUkHttpClient.java
```diff
@@ -0,0 +1,44 @@
...
+
+import javax.inject.Inject;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.Response;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+public class DropwizardGovUkHttpClient implements GovNotifyHttpClient {
```
rauligs: Minor: I wouldn't name classes after framework used

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
```diff
@@ -50,14 +55,19 @@ public GatewayResponse operation(ChargeEntity chargeEntity) {
...
-
         logger.info(format("Card captured response received - status = %s, charge_external_id = %s", captureResponse.getStatus(), chargeEntity.getExternalId()));
 
         ChargeEntity reloadedCharge = chargeDao.merge(chargeEntity);
         reloadedCharge.setStatus(captureResponse.getStatus());
-
         chargeDao.mergeAndNotifyStatusHasChanged(reloadedCharge);
 
+        if (operationResponse.isSuccessful()) {
+            sendEmailNotification(reloadedCharge);
```
rauligs: minor: Probably the method extract is too much? is just one line extracted

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -0,0 +1,77 @@
...
+            return Optional.empty();
+        }
+        if (StringUtils.isNotBlank(emailAddress)) {
```
rauligs: If we expect notifications, email is a mandatory attribute, if we don't send an email at least we should logging as a warn, my approach would be throwing an exception since we definitely don't expect this. Also I would change the condition to a positive is email enabled with then the logic to send the email (and throwing exception if email is blank, despite I wouldn't even check for that, this validation should be captured way before)

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -0,0 +1,77 @@
...
+        if (!emailPersonalisation.isEmpty()) {
+            emailRequestBuilder = emailRequestBuilder
+                    .personalisation(Personalisation.fromJsonString(new Gson().toJson(emailPersonalisation)));
+        }
+        return emailRequestBuilder.build();
+    }
+
+    private void readEmailConfig(ConnectorConfiguration configuration) {
+        if (StringUtils.isBlank(configuration.getNotifyConfiguration().getEmailTemplateId())) {
+            throw new RuntimeException("config property 'emailTemplateId' is missing or not set, which needs to point to the email template on the notify");
```
rauligs: Same here, configuration responsibility.

src/test/java/uk/gov/pay/connector/unit/service/CardCaptureServiceTest.java
```diff
@@ -26,12 +31,20 @@
...
 
-    private final CardCaptureService cardCaptureService = new CardCaptureService(mockedChargeDao, mockedProviders);
+    @Mock
+    private UserNotificationService mockUserNotificationService;
+    private CardCaptureService cardCaptureService;
+
+    @Before
+    public void beforeTest(){
+        cardCaptureService = new CardCaptureService(mockedChargeDao, mockedProviders, mockUserNotificationService);
+    }
```
rauligs: I would add zero interactions for the other scenarios

src/test/java/uk/gov/pay/connector/unit/service/UserNotificationServiceTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+import static junit.framework.TestCase.fail;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class UserNotificationServiceTest {
```
rauligs: Missing scenarios: 
- when email notification is not enabled
- exception sending an email

rauligs:
I don't see any integration test. I would expect a capture with mock server email service and verifying the call with expected parameters being sent.

rauligs:
Have we explore the possibility to have end to end tests for notifications (I believe Notify provides an API to query the emails sent). I think in a show and tell they explained it.

rauligs:
We can't forget to add these values in pay-infra when email notifications are being enabled.

rahul-vinayak:
rebuild

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
rahul-vinayak: Yes probably, there was much more than that initially when I extracted it. Will inline it

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
rahul-vinayak: I chose it to be a warning as normally the practice on Prod will be to expect confirmation emails to be sent through, and if you don't highlight the fact that the email enablement is turned off it can easily be overlooked at and people will wondering why emails are not going through. 
However currently it will keep warning on every time its trying to send an email, rather I have changed it now so that it warns just once when the app is coming up. 

Email enabling is just a preventive measure as we don't have an account or templates that we are going to use with Notify concluded as yet, setting it to true needs all the other configs to be set as well which should be finalised in coming stories.

src/main/java/uk/gov/pay/connector/service/userNotifications/DropwizardGovUkHttpClient.java
rahul-vinayak: This class is not even in use anymore, will get rid of it.

src/test/java/uk/gov/pay/connector/unit/service/CardCaptureServiceTest.java
rahul-vinayak: zero interactions on? can you elaborate..

If you meant adding zero interactions on the UserNotificationService, I have added them to make it more apparent.

src/test/java/uk/gov/pay/connector/unit/service/UserNotificationServiceTest.java
rahul-vinayak: Good point, added the test when notification is disabled.
Although the current story does not do any email delivery status checks, only thing when error occurs sending email we just log that, as sending email is a side effect before redirection we just log error and don't stop the the user being redirected to the service. Only thing we can test is if the appropriate logging is happening, however as we are not injecting logs it would be tricky to test that behaviour.

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
rahul-vinayak: I have refactored this method now.

src/main/java/uk/gov/pay/connector/service/NotifyClientProvider.java
rahul-vinayak: I prefer not to put any business logics in the configuration POJOs. Config classes should still read the values irrespective of the business rules and any exceptions or error should be handled at caller.

I have refactored the config reading part a bit to make it a bit cleaner now.

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
rauligs: 👍

rahul-vinayak:
We thought about writing an e2e test but because there is no stubbing env that notify provide that would have meant probably sending arbitrary emails every time tests are run. And also there was not a need to write a Stub for notify ourselves as we don't do anything in turn to an expected output from notify. So writing tests on a mock and verifying invocations was sufficient for the story. 
Future stories might need to check delivery status of emails/text messages and handle failed deliveries and it would make more sense to write stubs for notify and then e2e tests can be explored and relied upon.
As of now the various mock tests should be sufficient I believe, unless someone has some other ideas.

rahul-vinayak:
rebuild

rahul-vinayak:
As discussed about integration and e2e tests, we are going to write some stubs to write tests against. These will be followed in coming PRs.

PR #198, opened 2016-07-11
PP-770 Card number size

rauligs:
## WHAT
- Max size for card numbers is 19 characters instead of 16 characters. So changes according and added some validation tests.
## WHO
- [X] Developers
- [] WebOps
## Related PRs

[pay-frontend: PP-770 correct the size of the card numbers](https://github.com/alphagov/pay-frontend/pull/139)


PauloPortugal:
:+1:

PR #199, opened 2016-07-11
Updated README.md

rahul-vinayak:
| branch | PR |
| --- | --- |
| other_branch | [link]() |

The email notify story relies on the following env variables to be set in the connector:
- NOTIFY_EMAIL_ENABLED=
  {true/false} -- enable sending of emails through notify
- NOTIFY_EMAIL_TEMPLATE=
  {template-id}
- NOTIFY_SERVICE_ID=
  {service-id}
- NOTIFY_SECRET=
  {secret} -- possibly to be coming from cred-stash
- NOTIFY_BASE_URL=
  {base-url} -- typically https://api.notifications.service.gov.uk if we are using notify


rahul-vinayak:
rebuild

PR #202, opened 2016-07-18
PP-448 Added Integration tests for UserNotificationService running against stubs created

rahul-vinayak:
Notify email integration tests which run against wiremock stub for notify.


ijmad:
SSL issues, I can help with, but SSL needs to work end to end before we should merge this.

rahul-vinayak:
@ijmad now that we have changed the strategy of the notify integration tests in this PR from running against the Stubs to running against wiremocks, the above SSL issue is not relevant anymore. The E2E tests that will be running against the Notify Stubs though will still have the SSL issue to look into but thats a different PR anyways.

src/test/java/uk/gov/pay/connector/service/UserNotificationServiceIntTest.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Test
+    public void notifyPaymentFailedEmailRequest() throws Exception {
+        notifyEmailMock.responseWithEmailRequestResponse(400, BAD_REQUEST_RESPONSE, -1);
+        Optional<String> idOptional = userNotificationService.notifyPaymentSuccessEmail("mail@stubbed.com");
+        assertFalse(idOptional.isPresent());
+    }
+
+    @Test(expected = GovNotifyClientException.class)
```
imranbohoran: Minor: We seem to be using the old style of expected here. We have been using `ExpectedException` rule in other places. The expected rule is handy if you want to check for a specific message in the exception. If you think its useful here, lets put it. If not I'm happy with this too.

rahul-vinayak: I wont say its an old style, depends what you want your test to do, could make use of that but its thrown by the notify client library and is not with any message even, so just this basic level exception check (with fewer lines) serves the purpose I believe.

imranbohoran:
👍

PR #205, opened 2016-07-19
PP-449 Enable Notifications for existing accounts

PauloPortugal:
## WHAT

- This is a fix for a regression issue introduced by the bootstrapping
  of email notifications on account creation.
- This will allow email notifications for existing accounts
## HOW

_Steps to test or reproduce:_

```
cd $WORKSPACE/pay-scripts
msl reset && msl -a up && ./run-endtoend.sh
```
## WHO

- [X] Developers
- [X] WebOps


src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -44,8 +44,11 @@ public EmailNotificationResource(EmailNotificationsDao emailNotificationsDao) {
     public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAccountId) {
         logger.info("Getting email notification text for account id {}", gatewayAccountId);
 
-        return emailNotificationsDao.findByAccountId(gatewayAccountId)
-                .map(emailNotificationEntity -> Response.ok().entity(emailNotificationEntity).build())
+        return gatewayDao.findById(gatewayAccountId)
+                .map(gatewayAccount ->
+                        emailNotificationsDao.findByAccountId(gatewayAccount.getId())
+                                .map(emailNotificationEntity -> Response.ok().entity(emailNotificationEntity).build())
+                                .orElseGet(() -> Response.ok().build()))
```
rahul-vinayak: I can see what you are trying to do as in returning an empty response if notificationEntity is missing, however I think from REST point of view, is it ok to return an OK if the specific resource I am looking for doesn't exist?

rahul-vinayak:
good to go overall

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
PauloPortugal: @rahul-vinayak In this case it is to support accounts that do exist but do not have email notifications setup in the DB, we still want to return a 200 and then allow them to setup a new email notification.

PR #208, opened 2016-07-21
PP-446 Fix null email

mrlumbu:
If service had no email associated with it, got a null pointer exception. This case is now handled.


src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -86,15 +87,19 @@ private void readEmailConfig(ConnectorConfiguration configuration) {
 
     private Map<String, String> buildEmailPersonalisationFromCharge(ChargeEntity charge) {
         GatewayAccountEntity gatewayAccount = charge.getGatewayAccount();
+        EmailNotificationEntity emailNotification = gatewayAccount
```
PauloPortugal: The tests should be able to create an account with a bootstrapped email notification details, althought this is a good fix for our integration environment.

PauloPortugal:
:+1:

PR #209, opened 2016-07-21
PP-446 add integration test for email personalisation

mrlumbu:
Add test to check that email personalisation params are being correctly passed through.


src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -47,6 +47,7 @@ public UserNotificationService(NotifyClientProvider notifyClientProvider, Connec
             EmailRequest emailRequest = buildRequest(emailAddress, this.emailTemplateId, buildEmailPersonalisationFromCharge(chargeEntity));
             try {
                 NotificationCreatedResponse notificationCreatedResponse = govNotifyApiClient.sendEmail(emailRequest);
+                System.out.println(emailRequest.getPersonalisation().asMap());
```
imranbohoran: Is this System.out required?

mrlumbu: no 😞

imranbohoran:
The change intended for this PR LGTM. However a comment on the entire set, which, in hindsight I should have mentioned earlier when the test was introduced. 
We are maintaining a String for the expected response at https://github.com/alphagov/pay-connector/blob/84412b2d547857f0be0eb44c1fdad1be83cea302/src/test/java/uk/gov/pay/connector/service/UserNotificationServiceIntTest.java#L54 and https://github.com/alphagov/pay-connector/blob/84412b2d547857f0be0eb44c1fdad1be83cea302/src/test/java/uk/gov/pay/connector/service/UserNotificationServiceIntTest.java#L39. 
Bad experiences of maintaining json as a string just came rushing to my head and thought I'd point it out. Up to you though if you want to fix this at all or if you want to include it as part of this PR.

mrlumbu:
Yep good point, I will try to do it as a separate PR if/when I have time today.

imranbohoran:
👍

PR #213, opened 2016-07-29
PP 734 Create refunds for Worldpay

ckalista:
## WHAT
- Updated Liquidbase migration to create a refunds table linked to the charges table
  via foreign key. Note the presence of an external_id field used to store a uniquely
  generated public id for the refund similarly to what we have done already for
  the charge. The field is indexed and unique.
- RefundStatus models the 4 status a refund can transition to: CREATED, SUBMITTED,
  REFUNDED or ERROR. Each RefundStatus contains a mapping to its external equivalents
  modeled by ExternalRefundStatus.
- RefundEntity model the refund JPA entity. Note the second constructor allows to create
  a new instance referencing the charge the refund is for, generating a unique public id
  and setting the status to CREATED by default.
- ChargeEntity updated to reference a collection of refunds.
- RefundDao offers to query a refund by ID or by EXTERNAL ID + Unit tests
- Renamed AuthorisationRequest, AuthorisationResponse, CancelRequest, CancelResponse
  CaptureRequest, CaptureResponse.
- Removed the need to pass down the logger to many of the errorRespone methods of
  GatewayResponse subclasses.
- ChargeRefundService carry out the heavy lifting of the refund operation with the
  gateway. It itentifies the charge the refund is for, performs the preliminary checks
  to ensure the a refund is available, send the refund request to the provider and
  update the outcome of the request. The method estabishChargeRefundAvailability
  takes care of checking the charge status is at CAPTURED and the refund doesn't
  exceed the charge amount left to refunded. Should a refund not be available the
  service throws a RefundNotAvailableRuntimeException which rolls back the
  transaction and interrupts the operation prematurely.
- ChargeRefundsResource exposes a new api to create and query for refunds.
- Updated Worldpay payload with refunds for create and get charge.
- Extended refund status field to take a varchar(25) instead of varchar(15) which was too
  short to store some of our statuses such REFUND SUBMITTED.
- Fixed RefundEntity which was persisting the status enum name instead of persisting
  the status enum value.
- Helper method in RefundStatus to return an Enum based on its value.
- Regenerated equals, hashCode and toString methods of ChargeResponse.
- Refactored ChargeRefundService doRefund method to return a response object containing
  the refund gateway response and the refund entity which are both needed in the resource.
- Created a new ChargeRefundService createHalResourceBuilderFor method to simplify
  the creation of an HAL resource builder used to serialise a refund.
- Refactored ChargesApiResourse listCharges method to use ChargeService.buildChargeResponse
  to build the charge response.
- ChargeRefundsResource exposes a new endpoint to create a new refund relying on the refund service
  for the actual operation. The endpoint returns the new refund following HAL standard.
- ChargeRefundsResource exposes a new endpoint to retrieve an existing refund for a given account id,
  charge id and refund id.
- Unit tests + Integration tests.


ckalista:
rebuild

src/main/java/uk/gov/pay/connector/service/ChargeRefundService.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private PreTransactionalOperation<TransactionContext, RefundEntity> prepareForRefund(ChargeEntity chargeEntity, Long amount) {
+        return context -> {
+            ChargeEntity reloadedCharge = chargeDao.merge(chargeEntity);
+
+            RefundEntity refundEntity = new RefundEntity(reloadedCharge, amount);
+            reloadedCharge.getRefunds().add(refundEntity);
+
+            ExternalChargeRefundAvailability refundAvailablitiy = estabishChargeRefundAvailability(reloadedCharge);
```
simad: Very important comment: 
Typo, refundAvailablitiy -> refundAvailability

src/main/java/uk/gov/pay/connector/resources/ChargeRefundsResource.java
```diff
@@ -0,0 +1,76 @@
...
+    public ChargeRefundsResource(ChargeRefundService refundService, ChargeDao chargeDao) {
+        this.refundService = refundService;
+        this.chargeDao = chargeDao;
+    }
+
+    @POST
+    @Path(REFUNDS_API_PATH)
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    public Response submitRefund(@PathParam("accountId") String accountId, @PathParam("chargeId") String chargeId, RefundRequest refund, @Context UriInfo uriInfo) {
```
simad: String accountId could be Long accountId, like on line 58 of this same file.

src/main/java/uk/gov/pay/connector/service/OrderRefundRequestBuilder.java
```diff
@@ -0,0 +1,53 @@
...
+        return new OrderRefundRequestBuilder("/smartpay/SmartpayOrderRefundTemplate.xml");
+    }
+
+    public OrderRefundRequestBuilder(String templatePath) {
+        this.templateStringBuilder = new TemplateStringBuilder(templatePath);
+    }
+
+    public OrderRefundRequestBuilder withMerchantCode(String merchantCode) {
+        this.merchantCode = merchantCode;
+
```
simad: Remove space

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
```diff
@@ -72,7 +139,10 @@ public String getCreatedDate() {
         return DateTimeUtils.toUTCDateString(createdDate);
     }
 
-    protected ChargeResponse(String chargeId, Long amount, ExternalChargeState state, String gatewayTransactionId, String returnUrl, String email, String description, String reference, String providerName, ZonedDateTime createdDate, List<Map<String, Object>> dataLinks) {
+    @JsonProperty("refunds")
+    public Refund refunds;
```
simad: Reading the code, the variable name `refunds` is a bit ambiguous, it would be more clear if we had `refundSummary` or something along those lines.

Do we have an opportunity to change the serialised name that gets sent to the client?

src/main/java/uk/gov/pay/connector/model/builder/AbstractChargeResponseBuilder.java
```diff
@@ -95,5 +97,10 @@ public T withProviderName(String providerName) {
         return thisObject();
     }
 
+    public T withRefunds(ChargeResponse.Refund refunds) {
```
simad: same as above, could we call the class RefundSummary instead of Refund?

rauligs:
Hold merge. Get refunds for a charge is missing.

src/main/java/uk/gov/pay/connector/service/ChargeRefundService.java
ckalista: Fixed.

src/main/java/uk/gov/pay/connector/resources/ChargeRefundsResource.java
ckalista: Indeed! Fixed.

src/main/java/uk/gov/pay/connector/service/OrderRefundRequestBuilder.java
ckalista: Fixed.

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
ckalista: Agreed! Fixed. API design document also updated.

src/main/java/uk/gov/pay/connector/model/builder/AbstractChargeResponseBuilder.java
ckalista: Fixed.

ckalista:
Code review comments all addressed.

ckalista:
rebuild

ckalista:
rebuild

simad:
👍

PR #215, opened 2016-08-04
Pp 914 Add /ignore of Worldpay notification statuses

simad:
## WHAT
- The mapping function now produces three possible results:
  - The status is recognised and handled, returning a Optional.of(ChargeStatus)
  - The status is recognised and not handled, returning a Optional.empty()
  - The status is not recognised at all
- The provider takes the mapping result and applies it in two places:
  - parseNotification, where we apply a ChargeStatus (or nothing) to the notification
  - confirmStatus, where we take a decision as to whether to fail or acknowledge the notification
- In order to add a new status, we need to add a new enum in WorldPayStatusMapping
  - With a ChargeStatus, if we want to handle it
  - Without a ChargeStatus, if we want not to handle it
- SmartpayStatusesMapper now uses a similar pattern to the Worldpay one (see previous commit)
- The actual logic implemented in the PaymentProvider has not changed, as it is out of the
      scope of this story
  ## HOW 
  _Steps to test or reproduce:_


src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayStatusesMapper.java
```diff
@@ -1,36 +1,64 @@
...
+        } catch (IllegalArgumentException e) {
+            logger.debug(format("Unrecognised worldpay status %s", worldpayStatus));
+            return Either.left(worldpayStatus);
         }
-        return chargeStatus;
-
     }
 
+    private enum WorldpayStatusMapping {
+        AUTHORISED("AUTHORISED", AUTHORISATION_SUCCESS),
```
ckalista: I would be tempted running through the logs in search for occurrences we have seen but not handled so far. This document seems to give us some clue about available statuses, not sure if it is exhaustive though but we can ask Worldpay! http://support.worldpay.com/support/kb/gg/pdf/psd.pdf

src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayStatusesMapper.java
```diff
@@ -0,0 +1,59 @@
...
+        } catch (IllegalArgumentException e) {
+            logger.debug(format("Unrecognised smartpay status %s", smartpayStatus));
+            return Optional.empty();
+        }
+    }
+
+    private enum SmartpayStatusMapping {
+        AUTHORISATION("AUTHORISATION", ImmutableMap.of(true, AUTHORISATION_SUCCESS, false, AUTHORISATION_REJECTED)),
+        CAPTURE("CAPTURE", ImmutableMap.of(true, CAPTURED)),
+        CANCELLATION("REQUEST_FOR_INFORMATION"),
```
ckalista: Copy/paste mistake here? Should it not be CANCELLATION?

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -170,19 +176,25 @@ private StatusUpdates confirmStatus(GatewayAccountEntity gatewayAccount, String
...
                     logger.error(format("Could not map worldpay status %s to our internal status.", worldpayStatus));
-                    return FAILED;
-                });
+                    return NO_UPDATE;
+                },
+                chargeStatusMaybe ->
+                        chargeStatusMaybe
+                                .map(chargeStatus -> {
+                                    Pair<String, ChargeStatus> update = Pair.of(inquiryGatewayResponse.getTransactionId(), chargeStatus);
+                                    return StatusUpdates.withUpdate(NOTIFICATION_ACKNOWLEDGED, ImmutableList.of(update));
```
ckalista: :+1:

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -170,19 +176,25 @@ private StatusUpdates confirmStatus(GatewayAccountEntity gatewayAccount, String
...
-        return WorldpayStatusesMapper
-                .mapToChargeStatus(worldpayStatus)
-                .map(chargeStatus -> {
-                    Pair<String, ChargeStatus> update = Pair.of(inquiryGatewayResponse.getTransactionId(), chargeStatus);
-                    return StatusUpdates.withUpdate(NOTIFICATION_ACKNOWLEDGED, ImmutableList.of(update));
-                })
-                .orElseGet(() -> {
+        return WorldpayStatusesMapper.mapToChargeStatus(worldpayStatus).either(
+                status -> {
                     logger.error(format("Could not map worldpay status %s to our internal status.", worldpayStatus));
```
ckalista: This should probably be a WARNING - we are simply ignoring returning an NOTIFICATION_ACKNOWLEDGED hence it should probably not be at ERROR level.

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -135,7 +136,12 @@ public StatusUpdates handleNotification(String notificationPayload,
 
             return Optional.ofNullable(chargeNotification)
                     .map(notification -> {
-                        notification.setChargeStatus(WorldpayStatusesMapper.mapToChargeStatus(notification.getStatus()));
+                        Optional<ChargeStatus> chargeStatusMaybe = WorldpayStatusesMapper.mapToChargeStatus(
+                                notification.getStatus())
+                                .either(
+                                        unrecognisedStatus -> Optional.empty(),
```
ckalista: Here I think we should have a WARNING similar to below "Could not map worldpay status %s to our internal status."

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -170,19 +176,25 @@ private StatusUpdates confirmStatus(GatewayAccountEntity gatewayAccount, String
...
+                    return NO_UPDATE;
+                },
+                chargeStatusMaybe ->
+                        chargeStatusMaybe
+                                .map(chargeStatus -> {
+                                    Pair<String, ChargeStatus> update = Pair.of(inquiryGatewayResponse.getTransactionId(), chargeStatus);
+                                    return StatusUpdates.withUpdate(NOTIFICATION_ACKNOWLEDGED, ImmutableList.of(update));
+                                })
+                                .orElseGet(() -> {
+                                    logger.info(format("Worldpay status %s is not yet handled.", worldpayStatus));
```
ckalista: Message should probably be along the lines of "Worldpay notification with status %s ignored".

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.service.worldpay;
 
 import com.google.common.collect.ImmutableList;
+import fj.data.Either;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.joda.time.DateTime;
```
ckalista: Line 130 below, we should add an INFO  "Worldpay notification with status %s ignored". Also we should not return a FAILED anymore but an ACKNOLEDGE.

ckalista:
Looking really good :+1: Just needs to tidy up logging messages a bit I think, we can chat about that if you want.

Have you considered removing the Blacklist as it becomes irrelevant now with your changes?

ckalista:
As discussed and as agreed, notification handling should be such as:

| status type | log level | log message | return code |
| --- | --- | --- | --- |
| unhandled | INFO | Worldpay notification with status %s ignored. | return 200 [OK] |
| unrecodnised | WARNING | Could not map worldpay status %s to our internal status. | return 200 [OK] |

rauligs:
rebuild

PR #216, opened 2016-08-04
PP-921 Fix Optional.empty() status in logs

simad:
## WHAT
- ChargeStatus is always empty when inserted in that particular log entry.
  What appears now in the logs is the notification status, which is the original
  string coming from Worldpay
## HOW

_Steps to test or reproduce:_


rauligs:
rebuild

rauligs:
👍

simad:
rebuild

PR #217, opened 2016-08-04
PP-857 Add max length validation for an email

simad:
## WHAT
- Max email size is 254 characters. The email length is checked:
  - In the createNewCharge path (checkInvalidSizeFields)
  - In the patchCharge path (validateChargePatchParams)
- In both cases, if the email is too long, a BadRequest status is returned.
## HOW

_Steps to test or reproduce:_


src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -466,19 +468,20 @@ public void cannotMakeChargeForMissingGatewayAccount() throws Exception {
...
+                        .put(JSON_GATEWAY_ACC_KEY, accountId)
+                        .put(JSON_RETURN_URL_KEY, returnUrl).build());
 
         createChargeApi.postCreateCharge(postBody)
                 .statusCode(BAD_REQUEST.getStatusCode())
                 .contentType(JSON)
                 .header("Location", is(nullValue()))
                 .body(JSON_CHARGE_KEY, is(nullValue()))
-                .body(JSON_MESSAGE_KEY, is("Field(s) are too big: [description, reference]"));
+                .body(JSON_MESSAGE_KEY, is("Field(s) are too big: [description, reference, email]"));
```
rahul-vinayak: Its fine, although I prefer to return what is the maximum length allowed rather than "too big" in the response as well so that user knows it upfront, for a better user experience.

simad: While I was implementing this, I realised we have that generic response in different places (ie. GatewayAccountResource). 
Adding info specific to the charge resources in that file would therefore require a bigger refactoring; if it's ok for you, I would rather leave it as it is.

Consider that there is going to be already client-side validation with a more user-friendly error - if somebody reaches connector with an invalid payload, I think it can be safe to assume it is not the average user :)

PR #218, opened 2016-08-04
PP-942 Fix refunds payload on search payments

rauligs:
## WHAT
- Refunds summary serialization in search payments (for each payment) wasn't properly serialised since JSON annotations were ignored.


rauligs:
rebuild

simad:
👍

PR #220, opened 2016-08-08
PP-928  Call to Notify made non  blocking

mrlumbu:
In order to mitigate against any service failure/slowness from Notify, this puts the service call
in its own thread.

Also renames CardExecutorServiceConfig to ExecutorServiceConfig as it seems generic.

with @simad


src/test/java/uk/gov/pay/connector/service/UserNotificationServiceIntTest.java
```diff
@@ -133,8 +137,8 @@ public void notifyPaymentSuccessEmailWithPersonalisation() throws Exception {
     @Ignore
```
rauligs: Accidentally ignored?

rauligs:
After a chat with @mrlumbu, there are pending changes for Notify Java client that makes de code testable (since it makes it very difficult, so Ignored tests won't be removed but needs to be revisited when Notify Java client makes the suggested changes)

PR #224, opened 2016-08-16
PP-940 Create InquiryGatewayRequest and refactor inquire()

simad:
## WHAT
- `inquire()` now follows the same pattern as the other PaymentProvider methods
- transactionId and gatewayAccount are embedded in the InquiryGatewayRequest
  ## HOW 
  _Steps to test or reproduce:_


src/main/java/uk/gov/pay/connector/model/InquiryGatewayRequest.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.pay.connector.model;
+
+import uk.gov.pay.connector.model.domain.ChargeEntity;
+import uk.gov.pay.connector.model.domain.GatewayAccountEntity;
+
+public class InquiryGatewayRequest implements GatewayRequest {
+
+    private final GatewayAccountEntity gatewayAccountEntity;
```
ckalista: Not so sure we should tie the whole entity with the Request. Can we no just keep what we need from the entity instead?

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -74,32 +74,14 @@ public GatewayResponse cancel(CancelGatewayRequest request) {
...
-    private GatewayResponse mapToInquiryResponse(GatewayClient.Response response) {
-        return reduce(
-                client.unmarshallResponse(response, WorldpayOrderStatusResponse.class)
-                        .bimap(
-                                GatewayResponse::with,
-                                GatewayResponse::with
-
-                        )
-        );
+    public GatewayResponse inquire(InquiryGatewayRequest request) {
```
ckalista: yeah 👍

ckalista:
Code review completed - just one comment.

src/main/java/uk/gov/pay/connector/model/InquiryGatewayRequest.java
simad: In theory we don't need the whole object, true, but we have to keep it to conform to the GatewayRequest interface, which has `GatewayAccountEntity getGatewayAccount();`

PR #226, opened 2016-08-17
PP-941 Adding live provider configurations

rahul-vinayak:
- New column in gateway_accounts table called ‘type’ which basically
  tells whether the account type is a ‘Test’ one or a ‘Live’ one.
- New configurations in the config.yaml specifying the live or test
  provider url configurations. Which url is used depends on which account
  is used to generate the API key, which is based on the ‘type’ column in
  the gateway_accounts table above.
- Additions in migration.xml for liquibase changes
- Modified and fixed failing tests to suffice with the above two
  changes.
- Ansible and pay-script changes to follow
- TBM after scripts and ansible


src/main/resources/migrations.xml
```diff
@@ -349,4 +349,13 @@
         </createIndex>
     </changeSet>
 
+    <changeSet id="add type(TEST/LIVE) column to gateway_accounts table" author="shakespeare">
```
ckalista: Author Shakespeare ? :-)

rahul-vinayak: that was my easter egg, and to check who is paying attention :)

ckalista:
Code review completed. No further comments.

src/main/resources/migrations.xml
```diff
@@ -349,4 +349,13 @@
...
+        <addColumn tableName="gateway_accounts">
+            <column name="type" type="varchar(10)" defaultValue="TEST" remarks="possible values (TEST/LIVE)">
+                <constraints nullable="false" unique="false"/>
+            </column>
+        </addColumn>
+        <addUniqueConstraint tableName="gateway_accounts" columnNames="payment_provider, type"/>
```
ijmad: This constraint is wrong - we will have many live/test worldpay accounts, many sandbox accounts, etc. This constraint will prevent that and will not apply even on the existing contents of the gateway_accounts table in integration. This must be removed.

ijmad:
This must NOT be merged until the issue with the constraints is fixed.

src/main/resources/migrations.xml
ijmad: I appreciate a cute easter egg as much as the next dev, but please don't do this - undocumented features are an issue in PCI-DSS, and while this is harmless we should set good principles please.

src/main/resources/migrations.xml
rahul-vinayak: Constraint removed now

rahul-vinayak:
As discussed, removed the Unique key constraint now and merge shouldn't be stopped for that reason anymore.

PR #227, opened 2016-08-17
[REVIEW_ONLY] PP-696 Create card_details table in db

simad:
## WHAT
- We need to temporary persist a subset of card details in the database, so that
  we can retrieve them when resuming an incomplete payment. Right now they are
  stored in a cookie directly in front end
- The Card object now is a CardEntity, and directly models the card_details table
  in the db

with @PauloPortugal
## HOW

Please double check if card_details column names and types make sense


src/main/resources/migrations.xml
```diff
@@ -349,4 +349,39 @@
         </createIndex>
     </changeSet>
 
+    <changeSet id="build card_details table" author="">
+        <createTable tableName="card_details">
+            <column name="id" type="bigserial" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="card_number" type="char(19)">
```
whpearson: Surely this should only be 4 digits as we are only storing the last 4 card numbers.

PR #228, opened 2016-08-17
PP-736 Fix External refund status

rauligs:
## WHAT
- Pre-operation for refunds were adding the refunds to calculate refund availability after the refund and this was conflicting with calculating the refund status for a charge with the current refunds made so receiving wrong results for full refunds as available for refund.


rahul-vinayak:
happy to merge when its ready to be

rauligs:
rebuild

rauligs:
rebuild

PR #230, opened 2016-08-19
PP-914 Improved notification status mapping, ignoring unknown and unhandled notification

ckalista:
## WHAT
- Generic StatusMapper allowing to map provider statuses to internal statuses. Once the mapping
  is configured correctly, it is used to determine whether a status is unknown, ignored or mapped.
- Notification handling logic was updated to using the new StatusMapper. We now ignore notifications
  that have statuses that are either 'unknown' or 'ignored' according to the StatusMapper.
- Changed the notification handler to acknowledge at all times with a 200 [OK] even when an
  error occurs.
- Superseded Blacklist functionally was removed.


ckalista:
rebuild

ckalista:
rebuild

PauloPortugal:
:+1:

PR #231, opened 2016-08-22
PP-696 store and remove confirmation details

simad:
## WHAT
- We need to temporary persist a subset of card details in the database, so that
    we can retrieve them when resuming an incomplete payment. These are:
  - Last four digits of Card Number
  - Card Holder
  - Expiry Date
  - Billing Address
  - The details are stored in the ConfirmationDetailsEntity object, together with
    the Charge (1:1 mapping) they are related to.
- The details are stored when a charge is in AUTHORISATION_SUCCESS state and removed once the state changes to anything else
- The details are serialised in frontend but not in publicApi where they should never appear
## HOW

_Steps to test or reproduce:_

with @PauloPortugal


src/main/java/uk/gov/pay/connector/model/domain/ConfirmationDetailsEntity.java
```diff
@@ -0,0 +1,102 @@
...
+    public Address getBillingAddress() {
+        return billingAddress;
+    }
+
+    public void setBillingAddress(Address billingAddress) {
+        this.billingAddress = billingAddress;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
ckalista: JPA entities don't need equals and hash to be defined. The persistence engine takes care of that.

src/test/java/uk/gov/pay/connector/it/dao/ConfirmationDetailsDaoITest.java
```diff
@@ -0,0 +1,157 @@
...
+
+        assertThat(confirmationDetailsEntityMaybe.isPresent(), is(true));
+
+        ConfirmationDetailsEntity confirmationDetailsEntity = confirmationDetailsEntityMaybe.get();
+
+        assertNotNull(confirmationDetailsEntity.getId());
+        assertThat(confirmationDetailsEntity.getCardHolderName(), is(confirmationDetailsTestRecord.getCardHolderName()));
+        assertThat(confirmationDetailsEntity.getLastDigitsCardNumber(), is(confirmationDetailsTestRecord.getLastDigitsCardNumber()));
+        assertThat(confirmationDetailsEntity.getExpiryDate(), is(confirmationDetailsTestRecord.getExpiryDate()));
+        assertNotNull(confirmationDetailsEntity.getBillingAddress());
```
ckalista: It looks like we could benefit from a Hamcrest Javabean Matcher or something like this here: 
http://www.unitils.org/tutorial-reflectionassert.html

src/test/java/uk/gov/pay/connector/model/domain/AddressFixture.java
```diff
@@ -0,0 +1,55 @@
+package uk.gov.pay.connector.model.domain;
```
ckalista: 👍

src/test/java/uk/gov/pay/connector/model/domain/CardFixture.java
```diff
@@ -0,0 +1,55 @@
+package uk.gov.pay.connector.model.domain;
```
ckalista: 👍

ckalista:
Code review completed.

src/main/java/uk/gov/pay/connector/model/domain/ConfirmationDetailsEntity.java
simad: Oh cool. I didn't know; will remove them. Thanks!

src/test/java/uk/gov/pay/connector/it/dao/ConfirmationDetailsDaoITest.java
simad: Yes, this would be best. Moreover, I just discovered there is also an hamcrest matcher that does that, called refEq, so no need to import an external library.

The problem is that we are using test doubles (see DatabaseFixtures); this messes up any assertion, as basically we end up comparing two different classes.

Considering that we also have already ChargeFixtures, AddressFixtures (etc.) builders, which correctly return the original object and not a test double, I would suggest leaving this alone for now.
I will refactor the codebase to get rid of test doubles and use the fixtures correctly, and that will solve the problem in a successive PR :)

bazbremner:
rebuild

bazbremner:
rebuild

ijmad:
Merged in @ckalista 's stead, as on holiday.

PR #232, opened 2016-08-23
PP-951 Fixes bug, search email and reference with underscores

rahul-vinayak:
- underscores and % sign had special meaning in the like postgres
  queries, which results in unexpected behaviour, solution being escaping
  these symbols in filter charges by email and references.
- as an improvement made the references and email searches
  case-insensitive now.
- expanded tests to cover them.


rauligs:
👍

PR #237, opened 2016-08-30
Pp 940 Notification service capable of handling refunds

ckalista:
## WHAT

This is a complete rewrite of how notifications are handled in the connector to offer greater extendibility and level of assurance it is doing the right thing! This line of work initiated as we realised that the changes required to support refunds would have been too fundamental to justify patching up the codebase in this area any further. The refactoring introduces a new NotificationService taking care of all notification handling purposes in a generic way across providers (namely Worldpay, Smartpay and Sandbox) leaving the respective PaymentProvider classes uncluttered implementing specifics only.


rauligs:
rebuild

mrlumbu:
Ok, I have gone through this in some detail - reviewed tests, coverage, gone through all the commits. I think it looks fine, but I did not have much (any) experience of Notification part of the codebase before, so I would be happier if someone else also reviewed this before signing it off - preferably someone with more experience of Notifications (@PauloPortugal ?) But it's a 👍  from me

heathd:
@rahul-vinayak nitpick .. should fd1a79a say "reading the reference **from** notification"?

rahul-vinayak:
@heathd good spot .. reworded.

heathd:
only other comment would be to maybe squash together the commits relating to refund notifications 
- fd1a79a
- 69d8fec
- a81f823
- d0c7d63
- 314cab0

at present there's functionality added in fd1a79a and tests added subsequently in 69d8fec.

Alternatively split into two commits:
- a commit which inserts the refund reference into the payload sent to worldpay (currently most of the code in fd1a79a does this, but the title of the commit is). This would need to include the tests of this functionality as well
- a commit which unmarshalls the refund reference

heathd:
I also feel as though cd9c834 needs an explanation of the rationale for this change, as it's not obvious to me

rahul-vinayak:
@heathd addressed your last two comments

heathd:
👍  looks good

PR #239, opened 2016-08-31
Pp 720 fix return codes

mrlumbu:
Two bits to this:

**Fix returning 400 on cancelling charge in AUTH_READY**

Previously we returned 400 if auth ready charge was cancelled. We now return 409 (conflict), as auth ready is not a terminal state, it is a lock state.

**Fix returning 400 on cancellation of charge failing with gateway**

 Previously we returned a 400 when gateway cancel failed. From the user/service point of view it is better to return a success code, and then actually cancel the charge at a later date. Even though the cancel has failed with the gateway, the key thing s that the user can no longer interact with the charge.


rauligs:
I'm surprised this change made all the e2e / accept tests to pass (is there enough coverage for this or we just ignore whatever the outcome from connector in PublicAPI, frontend?)

rauligs:
👍

PR #241, opened 2016-09-01
PP-940 Increase logging level of some Notification messages from WARN to ERROR

ckalista:
## WHAT

All logs informing that a notification has been ignored is now logged as an ERROR to raise attention and encourage investigating the occurrence in production environment.


mrlumbu:
should this line https://github.com/alphagov/pay-connector/blob/82af4acdc5acc82cf9346965d86d8072f4f0a852/src/main/java/uk/gov/pay/connector/service/NotificationService.java#L95 not be error as well?

ckalista:
@mrlumbu The notification status is mapped before that. If for whatever reason the status cannot be mapped an ERROR is logged line 183 which results in an Empty internal status. Now after that all that is left to do is ignore the notification in line 95, hence the ignore is an INFO as the ERROR has already been logged.

mrlumbu:
Ok. 👍

PR #242, opened 2016-09-02
PP-1014 Sandbox refunds

rauligs:
## WHAT
- Enable refunds for _sandbox_ accounts.
- ExternalRefundAvailability updated as initially available for this gateway account.
- Sandbox Refunds through the API.
- Parse reference when receiving a notification for _sandbox_ gateway.
- Removed 'sandbox' notifications so statuses for capture and refund are success for 'sandbox' straight away (without awaiting for a notification). Hence the differentiation in state transitions.


src/main/java/uk/gov/pay/connector/model/domain/SandboxStateTransitions.java
```diff
@@ -0,0 +1,29 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import java.util.List;
+import java.util.Map;
+
+import static uk.gov.pay.connector.model.domain.ChargeStatus.*;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.USER_CANCELLED;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.USER_CANCEL_ERROR;
+
+final class SandboxStateTransitions extends StateTransitions {
```
rahul-vinayak: Can you explain briefly in the PR or better in commit why does it need a SandboxStateTransitions different from the DefaultTransition.

rauligs: Is in the commit message:
- Removed 'sandbox' notifications. **Statuses for capture and refund
  are success for 'sandbox'.**

The tests documents all the different states for default and sandbox

Updated the PR too.

PR #243, opened 2016-09-02
PP-718 Secure worldpay notifications' endpoint

simad:
## WHAT

- Any incoming notification hitting `v1/api/notifications/worldpay` triggers an additional check. We do a reverse DNS lookup on the notification's ip address, and if the host is unexpected (aka, it doesn't end with 'worldpay.com'), then connector returns a `403 forbidden` response.
  - The 403 is returned before any xml parsing / access to database, mitigating ddos risks.
- The reason for adding all the the dns lookup logic is that, apparently, `InetAddress` is not enough, see http://archive.oreilly.com/pub/post/reverse_dns_lookup_and_java.html.
- The secure endpoint is disabled for local dev and tested in IT (by setting the IP of our jenkins box as the "incoming notification" IP, and asserting that it belongs to the domain `amazonaws.com`)
## HOW

_Steps to test or reproduce:_


simad:
- [x] Create an environment variable to quickly enable and disable the secure endpoint system
- [x] Add the domain (ie. `worldpay.com`) as an environment variable
- [x] Check what happens with stubs / testing
- [x] Add env variable in `pay-ansible` to enable the secure endpoint
- [x] Check with Ops that `X-Real-IP` is the right header to obtain the original ip of the request
- [x] Switch to nginx header `X-Forwarded-For`

src/main/java/uk/gov/pay/connector/service/NotificationService.java
```diff
@@ -40,17 +41,32 @@
...
     }
 
-    public void acceptNotificationFor(PaymentGatewayName paymentGatewayName, String payload) {
-        new Handler(paymentProviders.byName(paymentGatewayName)).execute(payload);
+
+    public boolean handleNotificationFor(String ipAddress, PaymentGatewayName paymentGatewayName, String payload) {
+        PaymentProvider paymentProvider = paymentProviders.byName(paymentGatewayName);
+        Handler handler = new Handler(paymentProvider);
+        if (handler.hasSecuredEndpoint() && !handler.matchesIpWithDomain(ipAddress)) {
+            logger.error(format("received notification from domain different than '%s'", paymentProvider.getNotificationDomain()));
```
bazbremner: It'd be useful to log what the domain that the code _did_ receive is - useful for debugging.

simad: 👍 Done. In case of mismatch, we now log `Reverse DNS lookup on ip <address> - resolved domain: <resolved domain> does not match <expected domain>`

src/main/java/uk/gov/pay/connector/util/DnsUtils.java
```diff
@@ -1,19 +1,34 @@
...
     public boolean ipMatchesDomain(String ipAddress, String domain) {
         try {
             Optional<String> host = reverseDnsLookup(ipAddress);
-            return host.isPresent() && host.get().endsWith(domain + ".");
+            if (!host.isPresent()) {
+                throw new Exception("Host not found");
+            }
+            if (!host.get().endsWith(domain + ".")) {
+                logger.error(
+                        format("Reverse DNS lookup on ip %s - resolved domain: %s does not match %s", ipAddress, host.get(), domain));
```
bazbremner: 👍

src/main/java/uk/gov/pay/connector/resources/ApiPaths.java
```diff
@@ -31,4 +31,8 @@
     String REFUND_API_PATH = CHARGE_API_PATH + "/refunds/{refundId}";
 
     String CARD_TYPES_API_PATH = API_VERSION_PATH + "/api/card-types";
+
+    String NOTIFICATIONS_WORLDPAY_API_PATH = API_VERSION_PATH + "/api/notifications/worldpay";
+    String NOTIFICATIONS_SMARTPAY_API_PATH = API_VERSION_PATH + "/api/notifications/smartpay";
+    String NOTIFICATIONS_SANDBOX_API_PATH = API_VERSION_PATH + "/api/notifications/sandbox";
```
PauloPortugal: :+1:

src/main/java/uk/gov/pay/connector/resources/NotificationResource.java
```diff
@@ -55,7 +55,7 @@ private Response handleNotification(String ipAddress, String name, String notifi
         logger.info("Received notification from provider={}, notification={}", name, notification);
         PaymentGatewayName paymentGatewayName = PaymentGatewayName.valueFrom(name);
         if (!notificationService.handleNotificationFor(ipAddress, paymentGatewayName, notification)) {
-            return forbiddenErrorResponse("forbidden");
+            return forbiddenErrorResponse("Unknown Domain");
```
PauloPortugal: Do we want to tell whoever might try to exploit this endpoint that it is an unknown enpoint. Now they will know what to tackle.
It would be good that we returned an internal code. So that any genuine attempts would query us, but again, I guess we would have to make that error code available, in our documentation somewhere, defeating the purpose.

src/main/java/uk/gov/pay/connector/resources/NotificationResource.java
```diff
@@ -47,7 +47,7 @@ public Response authoriseSandboxNotifications(String notification) throws IOExce
     @Consumes(TEXT_XML)
     @Path("v1/api/notifications/worldpay")
     @Produces({TEXT_XML, APPLICATION_JSON})
-    public Response authoriseWorldpayNotifications(String notification, @HeaderParam("X-Real-IP") String ipAddress) throws IOException {
+    public Response authoriseWorldpayNotifications(String notification, @HeaderParam("X-Forwarded-For") String ipAddress) throws IOException {
```
PauloPortugal: @simad Why replace `X-Real-IP` with `X-Forwarded-For`? Could you add that do the commit message, to provide a bit more of context? Thanks :)

bazbremner: @PauloPortugal because https://payments-platform.atlassian.net/browse/PP-1075

src/main/java/uk/gov/pay/connector/resources/NotificationResource.java
simad: Yes, agreed. I couldn't find any good wording of that error without giving away details to a possible attacker. We can also not display any error message at all, as a 403 might be sufficient, after all.

src/main/java/uk/gov/pay/connector/resources/NotificationResource.java
simad: Thanks Baz! In the next commit I will address @PauloPortugal's comments, and I will amend the current commit, adding PP-1075's link, to give more context as to why we changed header

PauloPortugal:
Good to merge :+1: @bazbremner & @simad I will wait for the green light to merge this PR

PauloPortugal:
:+1:

PR #244, opened 2016-09-06
PP-719 Notification credentials

imranbohoran:
In order to secure smartpay notifications we will be using the basic auth feature from smartpay.
This PR introduces a table and its entities for us to hold the credentials for a particular gateway account. Password hashing etc will be in a separate PR.

with @mrlumbu


imranbohoran:
rebuild

heathd:
are the credentials also applicable to worldpay? if not, maybe this should be called `SmartpayNotificationCredentials`?

heathd:
LGTM 👍

PR #246, opened 2016-09-08
Pp 863 [REVIEW ONLY] Validate and associate card type to the charge during authorisation

ckalista:
- Changed type of CardTypeEntity Id to BigInt to avoid problems encountered with JPA when the type was UUID.  According to the documentation PostgreSQL JDBC driver has chosen an unfortunate way to represent non-JDBC-standard type codes hence all the difficulties faced when the type is UUID such as "id is of type uuid but expression is of type character varying" while persisting.
- Added an externalId to each card type used to expose a resource to the external world in very much the same way it was previously done for charge and refund.
- Added a CardBrandEntity one to many relationship to the ChargeEntity.
- Validate that `card type id` field is present in the payload of authorisation request, check if a card type exists for the given `card type id` and associate the card type to the charge being authorised.


ckalista:
rebuild

ckalista:
rebuild

ckalista:
build

src/main/resources/seed-data/card_types_with_external_ids.csv
```diff
@@ -0,0 +1,10 @@
+brand,type,label,external_id
+visa,DEBIT,Visa,ia4l3nln2eq3mgll5m1j4giipq
```
kakumara: Does this external id has a meaning to declare and load from outside? If not, why not have a not nullable column having default some sort of UUID format.

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -686,6 +688,11 @@ public void shouldCreateANewCharge() {
                 .withGatewayAccountEntity(gatewayAccount)
                 .build();
 
+        insertTestCardType();
+//        CardTypeEntity cardTypeEntity = new CardTypeEntity();
```
kakumara: ???

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -316,15 +320,37 @@ public void shouldFilterChargeStatusToReturnInProgressIfInternalStatusIsAuthoris
     }
 
     @Test
+    public void shouldReturnCardBrandWhenChargeIsStatusAuthorised() throws Exception {
+
+        long chargeId = RandomUtils.nextInt();
```
kakumara: Isn't there a GET charge test already that we can add this assertion in to?

src/main/resources/migrations.xml
```diff
@@ -420,4 +420,65 @@
             </column>
         </createTable>
     </changeSet>
+
+    <changeSet id="modify data type of card_types primary key from uuid to bigint" author="">
```
kakumara: You are going for a drop create migration here. Are you sure there is no live data already?

kakumara:
Other than that 👍

src/main/resources/seed-data/card_types_with_external_ids.csv
ckalista: As discussed there is no counter-argument to either approaches. The original line of thinking was that it could be advantageous to have predictable IDs for reference data in case in the future we want to pre-populate referring records.

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
ckalista: Removed those lines. 👍

src/main/resources/migrations.xml
ckalista: Yes, we discussed this with Till and Ian. The plan is to fix the data manually for our existing accounts as opposed as writing a migration script at this occasion.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
ckalista: We couldn't find an existing test that we could have just updated with the assertion so it felt cleaner to just create that new one.

ckalista:
Closing as the implementation has changed since the discovery that card types can not be detected for certain BIN ranges.

PR #247, opened 2016-09-09
PP-863 Prepare connector to receive card_type_id in authorisation payload request

ckalista:
## WHAT

This prepares the connector to receive the additional `card_type_id` field in the payload of authorisation request so that the frontend changes can be merged subsequently.


ckalista:
rebuild

PR #248, opened 2016-09-12
PP-1117 confirmation details fix

rauligs:
## WHAT
- Data were mixed up with concurrent users in authorise a charge due to CardAuthoriseService keeping cardDetails state. **These changes are the minimum possible to fix the issue.**
- A new story [**PP-1118**](https://payments-platform.atlassian.net/browse/PP-1118) has been created to refactor the service to use another approach. Already done in other services as _CardCancelService_


ijmad:
Agreed this looks like the right fix for now, and technical debt has been recorded in PP-1118.
Good tests to check we don't regress.

PR #249, opened 2016-09-12
PP-863 Persist card brand during authorisation and include card brand label in charge resources

ckalista:
## WHAT
- Updated all resources returning a charge to include the card brand label if it can be matched with a known card brand.
- `CardDetailsValidator` has been modified to ensure that the card brand is supplied during authorisation. The card brand is then extracted from the payload of authorisation requests and saved as it is against the charge.
- `card_brand` column added to the charge table.


ijmad:
rebuild

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -137,12 +141,26 @@ private boolean isValidStateTransition(ChargeStatus newChargeStatus) {
         return newChargeStatus.equals(ENTERING_CARD_DETAILS);
     }
 
+    private String findCardBrandLabel(String cardBrand) {
+        if (cardBrand == null) {
+            return "";
```
kakumara: I find `Optional<String>` returns much nicer in these type of finders (rather than "" or nulls)

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -112,6 +112,19 @@ private ChargeResponseBuilder chargeResponseBuilder(UriInfo uriInfo, ChargeEntit
                 }});
     }
 
+    private String findCardBrandLabel(String cardBrand) {
+        if (cardBrand == null) {
+            return "";
```
kakumara: Same as above. Wouldn't it be nicer to have an Optional<String> return here?

src/main/resources/migrations.xml
```diff
@@ -420,4 +420,13 @@
             </column>
         </createTable>
     </changeSet>
+
+    <changeSet id="add card_brand columns to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="card_brand" type="text">
+                <constraints nullable="true"/>
```
kakumara: Should this be 'nullable' ever given that pretty much all cards will have a brand? If its a problem with cardid algorithems, perhaps we should default to `UNKNOWN` or something?

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
ckalista: 👍 Done.

src/main/java/uk/gov/pay/connector/service/ChargeService.java
ckalista: Done.

src/main/resources/migrations.xml
ckalista: The reason why it has to be nullable is because we create the charge at much earlier stage than authorisation where we capture the card brand.

src/main/java/uk/gov/pay/connector/dao/CardTypeDao.java
```diff
@@ -29,4 +29,14 @@ public CardTypeDao(final Provider<EntityManager> entityManager) {
                 .createQuery(query, CardTypeEntity.class)
                 .getResultList();
     }
+
+    public List<CardTypeEntity> findByBrand(String brand) {
```
kakumara: Why list here? Doesn't one brand can have only one CardTypeEntity ?

src/main/resources/migrations.xml
kakumara: ah yes 👍

src/main/java/uk/gov/pay/connector/dao/CardTypeDao.java
ckalista: Well for the same brand you could have one CardTypeEntity for CREDIT and one for DEBIT.

ckalista:
rebuild

ckalista:
rebuild

kakumara:
👍

PR #253, opened 2016-09-15
PP-1145 Fixing flaky integration tests

imranbohoran:
## WHAT

- When fetching data through jdbi it appears that we are getting data
  for fields padded with white-spaces to match the column size. The external_id
  column in refunds table is char(26). So when jdbi fetches the data and transforms
  it and put it into a map the value in the map is padded with white-spaces to make up
  26 characters. This is very strange and bizzare. We are using jdbi only in tests so
  changing the matcher to trim the external id.


rauligs:
👍

PR #255, opened 2016-09-16
PP-1086 Improve logging

rauligs:
- This PR is not about the original story (thought it was a BUG) but having the logs we found out we need more information and also this information needs to be aligned with agreed format.
  ## WHAT
- Align logging messages.
- Added provider name for some relevant logging messages.


rauligs:
rebuild

rauligs:
rebuild

PR #257, opened 2016-09-16
PP-719 New endpoint for setting demo notification creds

imranbohoran:
- Add new endpoint /accounts/{accountId}/set-test-notification-credentials
- This endpoint is introduced so that we can use this to set up test accounts with existing
  credentials that are stored securely
- Doing it this way means we do not have to hold the password in plain text anywhere, and can reuse
  the existing credential management infrastructure

with @mrlumbu


imranbohoran:
It turns out that the script to setup demo accounts already look up credentials from credstash. So this end point doesn't add any value.

PR #261, opened 2016-09-20
PP-719 Logging when authentication fails

imranbohoran:
## WHAT

- We don't have useful logging to indicate if smartpay notification fails
  due to authentication failure. This change is to capture some logs.

with @mrlumbu


src/main/java/uk/gov/pay/connector/auth/SmartpayAuthenticator.java
```diff
@@ -4,19 +4,27 @@
...
     @Override
     public Optional<BasicAuthUser> authenticate(BasicCredentials basicCredentials) throws AuthenticationException {
-        return creds.asBasicCredentials().equals(basicCredentials)
-               ? Optional.of( new BasicAuthUser(creds.getUsername()) )
-               : Optional.absent();
+        if (creds.asBasicCredentials().equals(basicCredentials)) {
+            return Optional.of(new BasicAuthUser(creds.getUsername()));
+        }
+
+        logger.error(format("Authentication failure: failed for smartpay username %s", basicCredentials));
```
PauloPortugal: Do we want to print all the credentials here? The maximum allowed should be the username.

src/main/java/uk/gov/pay/connector/auth/SmartpayAccountSpecificAuthenticator.java
```diff
@@ -25,9 +30,11 @@ public SmartpayAccountSpecificAuthenticator(GatewayAccountDao gatewayAccountDao,
 
         return gatewayAccountDao.findByNotificationCredentialsUsername(basicCredentials.getUsername())
                 .filter((gatewayAccountEntity) -> matchCredentials(basicCredentials, gatewayAccountEntity))
-                .map(gatewayAccountEntity -> gatewayAccountEntity.getNotificationCredentials().toBasicAuthUser())
-                .map(Optional::fromNullable)
-                .orElseGet(Optional::absent);
+                .map(gatewayAccountEntity -> Optional.fromNullable(gatewayAccountEntity.getNotificationCredentials().toBasicAuthUser()))
+                .orElseGet(() -> {
+                    logger.error(format("Authentication failure: failed for smartpay username %s", basicCredentials));
```
PauloPortugal: Do we want to print all the credentials here? The maximum allowed should be the username.

PauloPortugal (commented):
I cannot see this in _PP-719_ the acceptance criteria to log the username/credentials.
If these are really required then we need to update the ticket in JIRA and we should not print all the credentials.
The question begs to be asked? Should we be logging usernames? If we get hacked and someone downloads all are logs or `grep` them then we will have a few accounts that could be hacked in a brute force manner.

src/main/java/uk/gov/pay/connector/auth/SmartpayAuthenticator.java
imranbohoran: The toString() of the BasicCredentials class masks the password. It only shows the username.

src/main/java/uk/gov/pay/connector/auth/SmartpayAccountSpecificAuthenticator.java
imranbohoran: The toString() of the BasicCredentials class masks the password. It only shows the username.

src/main/java/uk/gov/pay/connector/auth/SmartpayAuthenticator.java
PauloPortugal: :+1:

src/main/java/uk/gov/pay/connector/auth/SmartpayAccountSpecificAuthenticator.java
PauloPortugal: :+1:

PauloPortugal:
:+1: Spoke with Imran about logging the `username` and at the moment we do not have another option bu to do this. There is a plan to fix this in the near future.

PR #264, opened 2016-09-20
PP-946 log elapsedTime in outbound requests

kakumara:
## WHAT

This PR adds time elapsed logging message to outbound calls from connector

Uses Jersey client filters and MDC (similar to publicapi)

with @simad


imranbohoran:
As per our discussion lets test this out with multiple requests served on multiple threads to make sure the threadlocal works as we expect from the jersey client.

PR #265, opened 2016-09-22
PP-946 Substitution of unnecessary optional with defaultString

simad:
## WHAT

This is for consistency with the code we have in PublicApi and CardId
## HOW

_Steps to test or reproduce:_


PauloPortugal:
:+1:

PR #266, opened 2016-09-22
Stop using float points to calculate page number limits.

ijmad:
## WHAT

Stop using float points to calculate page number limits.
http://floating-point-gui.de/errors/comparison/

(Plus, small change to fix type hinting that upset my IntelliJ configuration)


src/main/java/uk/gov/pay/connector/resources/ChargesPaginationResponseBuilder.java
```diff
@@ -44,7 +44,9 @@ public ChargesPaginationResponseBuilder withTotalCount(Long total) {
     }
 
     public Response buildResponse() {
-        double lastPage = Math.ceil(new Double(totalCount) / searchParams.getDisplaySize());
+        long lastPage = totalCount / searchParams.getDisplaySize() +
+                (totalCount % searchParams.getDisplaySize() != 0 ? 1 : 0);
+
```
rahul-vinayak: I was fixing it via this, which I thought was shorter and simpler, still uses Double to calculate but its long after that all way along

> long lastPage = (long) Math.ceil(new Double(totalCount) / searchParams.getDisplaySize());

simad: We are not checking that searchParams.getDisplaySize() is != 0. It shouldn't happen, but I think it's worth making sure of that.

ijmad: The issue is with the operations on floats, not the conversion.

As a rule, we should never perform mathematical operations on floats when there is an obvious integer alternative available. Have a look at the referenced blog post.

ijmad: @simad good point. Will fix.

ijmad: @simad on further investigation it seems displaySize is passed through ApiValidators.validateQueryParams at https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java#L116 so it will be non-zero and positive. 

I have added an additional IllegalArgumentException just in case something leaks through.

Good enough?

simad: :+1:

PR #268, opened 2016-09-23
PP-1171 fixes bug for malformed navigation links

rahul-vinayak:
-The navigation links produced were malformed if you filter out
payments which don’t produce any results and next_link was produced
which was incorrect. This PR fixes it, plus modified tests to confirm
to this bug fix.
- If a search produces zero results only the self, first and last page
  links are shown.
- Plus modified the way of how we calculate the last_page.
- Getting rid of a few doubles


src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -200,10 +200,9 @@ private ZonedDateTime parseDate(String date) {
 
     private F<Boolean, Response> listCharges(ChargeSearchParams searchParams, UriInfo uriInfo) {
         Long totalCount = chargeDao.getTotalFor(searchParams);
-        if (totalCount > 0) {
-            long lastPage = totalCount / searchParams.getDisplaySize() +
-                (totalCount % searchParams.getDisplaySize() != 0 ? 1 : 0);
-
+        Long size = searchParams.getDisplaySize();
+        if (totalCount > 0 && size > 0) {
+            long lastPage = (totalCount + size - 1)/ size;
```
ijmad: Are you sure this produces the right result all the time?

rahul-vinayak: @ijmad I think it fits the bill (requirement) and is a fix for this bug as well, hope this addresses your earlier issues with the doubles as well.

ijmad: Yes, the issues with doubles was already fixed. I have done a manual comparison of the two calculations with data sizes between 1 and 1000 and page sizes between 1 and 100 and they seem to produce the same results, however, I must say I find your version of the calculation less clear. Is there a reason we're changing it again?

rahul-vinayak: Yes the issue with doubles was already fixed. I had to change the behaviour a bit in the file ChargesPaginationResponseBuilder:48 for the bug fix which was becoming too verbose on the same line and I found this to be shorter hence used the same logic in this file.
Its the difference between following two lines and I didn't find the first one too complex to read. 

> long lastPage = totalCount > 0 ? (totalCount + size - 1) / size : 1;
> 
> long lastPage = totalCount > 0 ? totalCount / searchParams.getDisplaySize() + (totalCount % searchParams.getDisplaySize() != 0 ? 1 : 0): 1;

PauloPortugal:
:+1:

PR #269, opened 2016-09-23
PP-1130 Add description and analytics_id to gateway accounts

simad:
## WHAT
- Adding `description` and `analytics_id` to `GatewayAccounts`. They can be both set in the POST request to create new accounts; if not set, for backwards compatibility, these fields **can be null**.
- Adding a new endpoint to PATCH either/both `description` and `analytics_id` for existing accounts.
- `description` and `analytics_id` will be serialised in all `GatewayResource`'s
  endpoints, potentially shown as null in the json response.
## HOW

_Steps to test or reproduce:_


rauligs:
It looks good. I will merge asap. 👍

PR #273, opened 2016-09-29
PP-1224 Ignore unknown properties for refund JSON request

rahul-vinayak:
- added to make connector tolerant to accept an additional fields
  coming from Public API, to be removed at a later point

with @PauloPortugal


imranbohoran:
👍

PR #274, opened 2016-09-30
PP-1224 [look at the last comment before merging] Connector now mandates "refund_amount_available" for refund request

rahul-vinayak:
- this extra field refund_amount_available is used to see if the refund
  request is legit or not
- rejects the refund request with 412 (precondition failed) if the
  refund_amount_available mismatches
- unignored the mistakenly ignored tests.
- removed the @JsonIgnoreProperties from refund request as that was
  done for backward comp.

with @PauloPortugal


rahul-vinayak:
rebuild

src/main/java/uk/gov/pay/connector/resources/ChargeRefundsResource.java
```diff
@@ -47,9 +47,9 @@ public ChargeRefundsResource(ChargeRefundService refundService, ChargeDao charge
     @Path(REFUNDS_API_PATH)
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
-    public Response submitRefund(@PathParam("accountId") Long accountId, @PathParam("chargeId") String chargeId, RefundRequest refund, @Context UriInfo uriInfo) {
-        validateAmount(refund.getAmount());
-        return refundService.doRefund(accountId, chargeId, refund.getAmount())
+    public Response submitRefund(@PathParam("accountId") Long accountId, @PathParam("chargeId") String chargeId, RefundRequest refundRequest, @Context UriInfo uriInfo) {
+        validateRefundRequest(refundRequest.getAmount());
```
kakumara: Do we need a `null check` here? What happens if `refundRequest` or `refundRequest.getAmount()` is null? Is there a test for it?

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
```diff
@@ -56,6 +56,11 @@ public static Response badRequestResponse(List message) {
         return responseWithMessageMap(BAD_REQUEST, message);
     }
 
+    public static Response preconditionFailedResponse(String message) {
+        logger.warn(message.toString());
```
kakumara: minor: Is this really a `warn` ? if this can happen its legitimate business scenario right? There's nothing Ops can do by seeing this warning message?

src/main/java/uk/gov/pay/connector/resources/ChargeRefundsResource.java
rahul-vinayak: This was an existing code I just renamed the method, however to answer your query, the DropWizard will throw an exception if there is no Entity in the request body or if the two properties defined in RefundRequest dont match up, else dropwizard handles it and returns an appropriate error.
If refundRequest.getAmount() is null that will make it 0, and a RefundsException will be thrown, covered by SandboxRefundITest and WorldpayRefundITest

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
rahul-vinayak: It shouldn't, changed it to info now.

rahul-vinayak:
This one needs to be merged first for this to go greeeeeen https://github.com/alphagov/pay-selfservice/pull/212

ijmad:
Not to be merged until https://github.com/alphagov/pay-selfservice/pull/212 is in production please

rahul-vinayak:
Can be merged now that 
https://github.com/alphagov/pay-selfservice/pull/212 is in prod

PR #276, opened 2016-10-04
Add description and analytics_id to GET requests in readme

simad:
## WHAT

- Adding `description` and `analytics_id` to responses to
  `GET /v1/frontend/{accountId}` and `GET /v1/api/{accountId}`
- This change should've been part of PP-1121.
## HOW

_Steps to test or reproduce:_


README.md
```diff
@@ -201,7 +203,8 @@ Content-Type: application/json
 | `gateway_account_id`                 | X | The account Id        |
 | `type`                 | X | Account type for this provider (test/live)|
 | `payment_provider`                 | X | The payment provider for which this account is created.       |
-
+| `description`            | X | An internal description to identify the gateway account. The default value is `null`.      |
```
imranbohoran: Is the default value `null` or just empty from a documentation point of view. Also if this is not always present we should probably remove the `X` in the always present column

README.md
```diff
@@ -201,7 +203,8 @@ Content-Type: application/json
...
+| `analytics_id`           | X | An identifier used to identify the service in Google Analytics. The default value is `null`.      |
```
imranbohoran: Ditto

README.md
```diff
@@ -734,7 +739,8 @@ Content-Type: application/json
 | `gateway_account_id`     | X | The account Id        |
 | `payment_provider`       | X | The payment provider for which this account is created.       |
 | `credentials`            | X | The payment provider credentials. Password is not returned. The default value is the empty JSON document {}      |
-
+| `description`            | X | An internal description to identify the gateway account. The default value is `null`.      |
```
imranbohoran: Is the default value null or just empty from a documentation point of view. Also if this is not always present we should probably remove the X in the always present column

README.md
```diff
@@ -734,7 +739,8 @@ Content-Type: application/json
...
+| `analytics_id`           | X | An identifier used to identify the service in Google Analytics. The default value is `null`.      |
```
imranbohoran: Ditto

PR #279, opened 2016-10-06
PP-719 Remove unused smartpay notification credentials code

imranbohoran:
## WHAT

- Since smartpay notification credentials are now account specific, there's no need
  for the use of environment variables.

with @mrlumbu


PauloPortugal:
:+1:

PR #280, opened 2016-10-06
PP-1094 Ensuring event history is in reverse chronological order

PauloPortugal:
## WHAT

 This attempts to fix two known issues:
- Prevent duplicate events (previously we were only achieving this
  if and only if the duplicate events were in successive order.
- Ensure that the event history appears in reverse chronological order
## HOW

_Steps to test or reproduce:_

```
cd $WORKSPACE/pay-scripts
msl reset && msl -a up && ./run-endtoend.sh
```


ijmad:
Would it be better to ask for them sorted from the database as well?

PauloPortugal:
rebuild

PauloPortugal:
rebuild

rahul-vinayak:
As far as I remember removing only successive repeating events, "and not removing the events that are getting repeated if there was another event that happened in between" was a specific requirement for this, @ijmad can recall this why was this specified, best to confirm with him this change.

ijmad:
If there are two internal state changes that map to the same external event we only want to show that event once in selfservice where you process that is entirely up to you. Certainly the /frontend API deals with internal states primarily, so should probably return all events including those that map to the same English, as the domain model for mapping events to English is in selfservice not connector, it should be doing the de-duping if it's not already, I can't remember how this stuff knits together without delving deeper.

ijmad:
Please also check the API contract for .equals and sorting makes sense for .distinct(), if there are two different timestamps the objects should not be .equals for a sorting operation, which may break .distinct if you just want to de-dupe on status.

PauloPortugal:
> _"If there are two internal state changes that map to the same external event we only want to show that event once in selfservice where you process that is entirely up to you."_ 

Connector seems the best place to do this fix as it will fix for any client consuming that endpoint.

> _"Certainly the /frontend API deals with internal states primarily, so should probably return all events including those that map to the same English, as the domain model for mapping events to English is in selfservice not connector, it should be doing the de-duping if it's not already, I can't remember how this stuff knits together without delving deeper."_ 

This is not an issue as FE does not consume this endpoint.

PauloPortugal:
> _"Please also check the API contract for .equals and sorting makes sense for .distinct(), if there are two different timestamps the objects should not be .equals for a sorting operation, which may break .distinct if you just want to de-dupe on status."_

There is no problem with the API contract. We should not use the `.equals()` method for sorting operation. the sorting operation is done on the `timestamp/updated` field

ijmad:
Apologies, forgot selfservice uses /api and not /frontend....
As for the sorting stuff, I still have some concerns but I can't delve until I'm back in the office.

ijmad:
Having discussed this verbally, I'm happy to sign off on my comments.

PR #283, opened 2016-10-11
PP-1226 use maintained Alpine-based OpenJDK baselayer

bazbremner:
# WHAT

Docker's "java" image has been deprecated in favour of the "openjdk"
image and will not receive any updates past 2016-12-31.

Given that the vanilla '8-jre' and '8-jdk' images are Debian Jessie
based and hence bring with them 150MB+ of packages (and hence security
updates to worry about), I figured it'd be worth experimenting with the
far smaller Alpine image, now that it's making it into the offical image
builds.


bazbremner:
rebuild

PR #286, opened 2016-10-27
PP-1302 Fix hardcoded merchantcode in SmartPay communications

kakumara:
## WHAT

We have been using a hardcoded MerchantCode in all smartpay communications until now. This must the username of the corresponsding account's SmartPay account credentials.

Fixed using the USERNAME field from the SmartPay account credentails for the `merchantCode` as per the documentation


kakumara:
rebuild

PR #287, opened 2016-10-31
PP-1302 SmartPay payment provider need `Merchant_id`

kakumara:
## WHAT
Add merchant_id support for SmartPay payment provider. 
This PR contains some backward compatible changes.
     - For now keeping it optional until all environment accounts are migrated
     - If a `merchant_id` exists it will use during the integration or will use the default one
     - Updated tests to have a `merchant_id` by default for both worldpay and smartpay


imranbohoran:
LGTM 👍

PR #288, opened 2016-11-01
PP-1302 remove backward compatible changes

kakumara:
## WHAT
Currently smartpay integrations sends a default (hard-coded) `merchant_id` in case if its not found in the current credentials. Once we migrate all the accounts in AWS, this needs to be removed as it must be taken from the credentials always.

This PR need to wait till we confirm all accounts in AWS are migrated correctly.


mrlumbu:
rebuild

rauligs:
Reviewed 👍

PR #289, opened 2016-11-09
PP-1282 Preserve Card Details

kakumara:
## WHAT
This PR preserves the card details that is used in every transaction, irrespective of if transaction is successful or not.
 - Payment card data is preserved in `charges` table as discussed. They are all nullable columns and are updated only after a transaction authorisation attempt.
- Removed ConfirmationDetailsService and related DAO as they no longer makes sense. CardDetailsEntity is now a Embeddable JPA entity in ChargeEntity.
- Leaving `confirmation_details` JSON property to support backward compatibility. However a new property `card_details` is also added to provide the same data which will be migrated in to.


PauloPortugal:
I think the decision to embed the card details inside the charge table was a good one. Thanks for doing it. :+1:

PR #290, opened 2016-11-11
PP-1282 backward compatible fix for potential break in payment confirmation view during deployment

kakumara:
## WHAT
Note: the scenario can only happen during production deployment where a card authorisation has happened in the connector node with `release 76` version while confirmation details page data load is happening with a `release 77` version of the connector node. 

 - Added a @Deprecated `ConfirmationDetailsEntity` object which be loaded if a matching record exists from `confirmation_details` table and will be used to populate Frontend charge response if its not null.


kakumara:
rebuild

kakumara:
rebuild

ckalista:
Code review completed - looking good!

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -151,27 +151,40 @@ private boolean isValidStateTransition(ChargeStatus newChargeStatus) {
...
+    private PersistedCard resolvePersistedCard(ChargeEntity charge) {
+        CardDetailsEntity cardDetails = charge.getCardDetails();
+        if (charge.getConfirmationDetailsEntity() != null) {
+            return charge.getConfirmationDetailsEntity().toCard(cardDetails != null ? cardDetails.getCardBrand() : "");
+        } else if (cardDetails != null) {
+            return cardDetails.toCard();
+        }
+        return null;
+    }
+
```
imranbohoran: As discussed, for the scenario we are trying to handle (one node having the new code and the confirmation landing on a node that still has the old code), this change will not make any difference as this is the new code. As you suggested the saving to the new and old structure seems the safest way to get this out in a backward compatible way.

imranbohoran (changes requested):

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
kakumara: Good catch! 👍  
Fixed by duplicating card details in old `confirmation_details` table.

imranbohoran (approved):

imranbohoran:
👍

PR #291, opened 2016-11-11
PP-1356 Posts to sandbox notification endpoint return 200

imranbohoran:
## WHAT
* We were returning a 500 for notifications reaching the sandbox endpoint
  As we are adding monitoring for notifications reachability as a smoke test,
  and we are going to hit the sandbox endpoint, we want to return a 200 to
  check that notification endpoints are reachable.

with @simad and @whpearson


bazbremner:
Should there be a unit test to validate this behaviour?

imranbohoran:
Didn't see the need for a unit test on this one as it wasn't doing much. But more than that I think we didn't get to writing a unit test because there isn't any for sandbox at all. That's not to say we shouldn't have any. So we will write a quick one to express the intent.

imranbohoran:
Created a test (similar to worldpay and smartpay) for sandbox notifications.

bazbremner:
Grand, thanks. I think a test is warranted, as if the behaviour changes, alerting will light up and it could be rather disruptive.

bazbremner:
👍

PR #292, opened 2016-11-14
PP-1282 Backward compatibility for in flight transactions with old code

kakumara:
## WHAT
To support transactions that are in progress with old code, to continue to display card details from the data captured in confirmation_details table (rather than card columns in charges table)


kakumara:
rebuild

ijmad:
:+1: Looks good, waiting for build.

PR #293, opened 2016-11-14
PP-1079 - Enabling Smartpay Refund

kakumara:
## WHAT
- Pushed the `getTransactionId` method down the `BaseResponse` hierarchy as not
       all gateway responses need this method such as for instance refund responses.
    - `getTransactionId` method added in `BaseAuthoriseResponse` as needed by all
       authorisation responses.
    - Introduced `BaseRefundResponse` used as base class model for all refund responses.
       Note that not all gateway providers return a refund reference. For instance for Worldpay
       since we generates the reference upfront and submit alongside the refund request, the
       provider doesn't return it back to us as part of the response hence `getReference` has
       been modeled as an Optional.
    - Added correct mappings for the XML(Un)marshaller to be able to (de-)serialise refund
       request/responses from Smartpay.
    - New reference field in `RefundEntity` which becomes the reference used to reconsile
       refund notifications with the original refund request orders.
    - Database mapping updated to add a reference field to the refund database table and also
       migrated existing refunds that have the reference field prepolulated with external refund id.
    - The refund lookup during notification handling is now based on the reference instead
       of the refund external id.
    - Introduced a new `generateRefundReference` method which all payment provider must
       supply their own implementation depending wether or not the provider is responsible
       to generate the reference upfront.
    - `ChargeRefundService` slightly changed the way refund requests are being issued to
       accomodate Smarpay which generates its own reference for the refunds that we want
       to capture for later on during reconsiliation when we receive notifications.
    - `NotificationService` was updated to lookup refunds based on their reference rather than
       external refund id.
    - Removed restriction that was put in place for MVP that prevented from issuing refunds for
       Smartpay.


kakumara:
rebuild

kakumara:
rebuild

kakumara:
rebuild

src/main/java/uk/gov/pay/connector/service/ChargeRefundService.java
```diff
@@ -142,37 +144,54 @@ public ChargeRefundService(ChargeDao chargeDao, RefundDao refundDao, PaymentProv
...
+                    return refundFinishSuccessStatusOf(chargeEntity.getPaymentGatewayName());
+                }
+                return RefundStatus.REFUND_ERROR;
+            };
 
-            logger.info("Card refund response received - charge_external_id={}, transaction_id={}, status={}",
-                    chargeEntity.getExternalId(), chargeEntity.getGatewayTransactionId(), status);
-            logger.info("Refund status to update - charge_external_id={}, status={}, to_status={} for charge_id={}, refund_id={}, refund_external_id={}, amount={}",
-                    chargeEntity.getExternalId(), refundEntity.getStatus(), status, chargeEntity.getId(), refundEntity.getId(), refundEntity.getExternalId(), refundEntity.getAmount());
+            Supplier<String> referenceF = () -> {
```
rauligs: Can these be created as two methods (one per supplier). They are suppliers which are immediately consumed.

heathd:
What is the plan for `externalId` on `Refund`? Will that continue to be used as well as `reference`? or will it be removed?

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundITest.java
```diff
@@ -72,7 +71,7 @@ public void shouldBeAbleToRequestARefund_partialAmount() {
 
         List<Map<String, Object>> refundsFoundByChargeId = databaseTestHelper.getRefundsByChargeId(defaultTestCharge.getChargeId());
         assertThat(refundsFoundByChargeId.size(), is(1));
-        assertThat(refundsFoundByChargeId, hasItems(aRefundMatching(refundId, defaultTestCharge.getChargeId(), refundAmount, "REFUND SUBMITTED")));
+        assertThat(refundsFoundByChargeId, hasItems(aRefundMatching(refundId, is(notNullValue()), defaultTestCharge.getChargeId(), refundAmount, "REFUND SUBMITTED")));
```
rauligs: At least in the happy path scenario, is it possible to assert that refund externalId is same as the reference?

rauligs (changes requested):
👍  LGTM. Only two minor suggestions.

src/main/java/uk/gov/pay/connector/service/ChargeRefundService.java
kakumara: 👍  fair point

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundITest.java
kakumara: Its already there is you see the the matcher. (see RefundMatcher.java)

kakumara:
@heathd re: comment.
AFAIK there is no plan. This field can be used in anyway we want. But I think it better to have the 2 fields, given that it is meant for `merchant's transaction reference` which could be interpreted as a reference from a government department that integrate with us. If they send us one we can pass it to payment gateway if it helps reconciliation or similar.

Generally, `external_id` is something we introduced globally (beyond refund) as an id from GOV.UK Pay for our clients. They were meant to hide the primary keys of our tables. It was just convenient to send it as merchant-reference in the refunds case .

heathd:
👍  looks good to me

We should make sure that Till is aware that although this is safe to ship to production and that technically it enables smartpay refunds to work, we're not ready to really do smartpay refunds in production until we have made the changes to properly record PSP references throughout the lifecycle of the refund.

PR #295, opened 2016-11-17
PP-1282 Fix card brand label on charges api resource

kakumara:
## WHAT
- It should be returning labled cardbrand (not the one persisted in database). Fixing it similar to the same way used in  ChargesFrontendResource.java


PauloPortugal:
:+1:

PR #296, opened 2016-11-17
PP-1377 As per ticket, the minimum length for bank card numbers is 12…

ijmad:
PP-1377 As per ticket, the minimum length for bank card numbers is 12, not 14.
This is correct in frontend but not in connector.
Fix connector to match.


bazbremner:
I'd suggest that there's value in retaining a negative test for 12 digit card numbers, rather than just updating the extant postive tests from 12 to 14 digits, but otherwise it looks good.

heathd:
Looks good. Would be nice to have a test that 12 digits are considered valid as well as considering 11 digits invalid (I **think** that's what @bazbremner was saying?)

ijmad:
rebuild

ijmad:
rebuild

PR #298, opened 2016-11-18
PP-1079 store transactionIds in charge_events table

kakumara:
## WHAT
Smartpay issue a new `pspReference` (maps transactionId in GOV.UK Pay) for every request. Given we have only a single transactionId column in charges, we loose the `pspReference` issued for AUTHORISED when a CAPTURE is completed as we are overriding the value.This PR introduces a new column in charge_events to store transaction ID for every state change so there will be separate rows for each state change.


PauloPortugal:
@kakumara this is more a general comment, not specific to this PR.
In our team we tend to do a lot of cosmetic refactoring, which should be captured in its own commit to avoid "polluting" the PR/story requirements.
This makes the life of the code reviewer a lot easier, history is cleaner and possibly roll backs (God forbid we have one!) are more straightforward.
What is your opinion on the matter?

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventEntity.java
```diff
@@ -18,6 +18,9 @@
     @Convert(converter = ChargeStatusConverter.class)
     private ChargeStatus status;
 
+    @Column(name = "gateway_transaction_id")
+    private String gatewayTransctionId;
```
PauloPortugal: Could you please fix the typo **"gatewayTransctionId"** to **"gatewayTransactionId"**?

kakumara:
I completely agree 👍 .  I do find it easier as well. 
I have this habit of spot refactoring when I touch a bit of code and forget to mark them separately. - 

- Can I make few changes more prominent? (One i can see is the renaming of EvenDao to ChargeEventDao, which actually was a feedback from @heathd) 
- anything else you see significant?

src/test/resources/templates/smartpay/notification-refund.json
```diff
@@ -9,11 +9,11 @@
         },
         "eventCode": "REFUND",
         "eventDate": "2015-10-08T13:48:30+02:00",
-        "merchantAccountCode": "DCOEcomm",
-        "merchantReference": "{{transactionId}}-merchant",
-        "originalReference": "{{transactionId}}",
+        "merchantAccountCode": "DCOTest",
```
PauloPortugal: :+1:

PauloPortugal (commented):
Looks OK

PauloPortugal:
@kakumara you do not need to put the changes in another commit for this story, unless you really want. This is more a future action that we should keep in the back of our minds, possibly we should discuss it in a TechRospective (we used to have those) or in a Retrospective. :+1: but apart from the typo, I am happy to approve the PR

src/main/java/uk/gov/pay/connector/model/domain/ChargeEventEntity.java
kakumara: 👍  addressed

PR #299, opened 2016-11-21
PP-1079 Fix smartpay namespace issue

kakumara:
## WHAT
1. Barclays (unfortunately) has 2 API name spaces (or have had in past). We must use the newer API names space `http://payment.services.adyen.com` everywhere. We were accidentally using a mixture of deprecated and new namespaces.
2. We have been using `charge_id` which the primary key of our charges table as merchant reference. Fixing it to use externalId.


src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayPaymentProvider.java
```diff
@@ -103,7 +103,7 @@ public StatusMapper getStatusMapper() {
     private Function<AuthorisationGatewayRequest, String> buildSubmitOrderFor() {
         return request -> aSmartpayOrderSubmitRequest()
                 .withMerchantCode(getMerchantCode(request))
-                .withPaymentPlatformReference(request.getChargeId())
```
mrlumbu: Confused why this change doesn't break things?

kakumara: It doesn't. Smartpay (or payment gateways in general) do not care about what we use as merchant reference. They just record it and playback. 
It is for us (and Government department particularly) make it useful when they query about a particular transaction directly with Gateways. Currently GOV departments only aware of our externalID's and it'll be best they could use the same reference everywhere.

mrlumbu: Ok, got it 👍

mrlumbu:
LGTM 👍

PR #302, opened 2016-11-29
PP-1386 Adding dropwizard metrics

imranbohoran:
## WHAT

 * Adds and initialised dropwizard metrics for connector. There
   aren't any custom metrics introduced in this commit. Default
   jvm metrics would be captured and published based on this commit.
 * Expects a METRICS_HOST environment variable to be set and
   in any environment apart from running it locally this would
   be set to the hostname of the host as seen by the docker container.
   Else it uses localhost.


imranbohoran:
rebuild

rauligs:
LGTM

PR #303, opened 2016-11-29
PP-1079 removing gateway_transaction_id from ChargeEvents

kakumara:
## WHAT
 - Upon manual testing it was discovered that our understanding of how smartpay `pspReference` worked was wrong
   The correct understanding should be:
     * Upon AUTH smartpay issues a `pspReference` which is the main transaction-ID for the payment
     * we should be using that pspReference(transactionID) for all subsequent modification requests (such as CAPTUREs, REFUNDs)
     * This is very similar to Worlpay except that we generate the transactionID
     * Therefore we no longer need to override the gateway_transaction_id in charges table or we don't need to keep a track of pspReferences of subsequent requests
     * this removes the persistence and usage of gateway_trasaction_id from charge_events table. The table column will be removed later


kakumara:
rebuild

rauligs:
LGTM 👍

PR #305, opened 2016-11-30
PP-1079 fix notifications to use correct transaction ID

kakumara:
## WHAT
 - As per the previous commit smartpay originalReference always refers to `pspReference` received as part of AUTHORISE response. (transaction-id) in our context
 - We have to make sure we lookup the charge by that originalRefence/transactionId  when we lookup for a charge when we receive a notification for charge (except for a refund notification)
 - For a refund notification, (as we can have mutiple refunds) we have to lookup refund record by the `pspReference` recieved for `[refund-received]` response. This is stored as `reference` column in our refunds table.
 - The problem was we were using the refund `pspReference` for both reference and transactionId. Fixed to use the originalReference during notification parsing.


kakumara:
rebuild

src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayPaymentProvider.java
```diff
@@ -87,7 +87,7 @@ public String getNotificationDomain() {
             objectMapper.readValue(payload, SmartpayNotificationList.class)
                     .getNotifications()
                     .forEach(notification -> builder.addNotificationFor(
-                            notification.getPspReference(), notification.getPspReference(), notification.getStatus()));
+                            notification.getOriginalReference(), notification.getPspReference(), notification.getStatus()));
```
PauloPortugal: :+1:

PauloPortugal (commented):
Looks good!

PR #306, opened 2016-12-05
PP-1457 (BUG) External status events date-time fix

rauligs:
## WHAT
- From a client point of view, external statuses are changing the Date time displayed, therefor is actually a misleading information.
- The problem is the distinct method used to discard external statuses uses the current order (newest to oldest status), so just changing the order before distinct method solves the situation and respect oldest external status mapping.


rauligs:
Closing. Still pending requirements for capturing times.

PR #309, opened 2016-12-07
PP-1386 Use gauges instead of histograms for CardExecutor metrics

imranbohoran:
## WHAT
- We are capturing counts  regarding the card executor, and they should
  be measured as gauges, not histograms

with @simad


imranbohoran:
rebuild

src/main/java/uk/gov/pay/connector/healthcheck/CardExecutorServiceHealthCheck.java
```diff
@@ -11,29 +12,27 @@
...
-    private MetricRegistry metricRegistry;
 
     @Inject
     public CardExecutorServiceHealthCheck(CardExecutorService cardExecutorService) {
         this.threadPoolExecutor = (ThreadPoolExecutor)cardExecutorService.getExecutor();
-        this.metricRegistry = cardExecutorService.getMetricRegistry();
+        initialiseMetrics(cardExecutorService.getMetricRegistry());
+    }
+
+    private void initialiseMetrics(MetricRegistry metricRegistry) {
```
PauloPortugal: Won't we have the same problem as a couple of days ago where the `.register(..., ...)` function was trying to insert again a new `<key,value>` pair. What happens if I call this `/healthcheck` endpoint in succession, will it update the values?

simad: We had that problem because we were previously calling `.register(...)` in the `check()` function itself. Metrics was, therefore, trying to register a gauge for each call to `/healthcheck`, throwing an exception because, after the first check, the metric was already registered.

We are now registering the gauges only once, in the constructor - these will be periodically polled and their result bundled with the rest of the metrics. You can see now that there's nothing metrics-specific in `check()`.

PauloPortugal: So you registered them once. How do they get updated?

simad: See my previous answer :) `these will be periodically polled and their result bundled with the rest of the metrics`
http://metrics.dropwizard.io/3.1.0/apidocs/com/codahale/metrics/graphite/GraphiteReporter.html

PR #310, opened 2016-12-12
PP-1433 [M&M] Add metrics for gateway response times and failures

simad:
## WHAT
- We want to have metrics regarding response times and failures for
  our payment providers. Instrumenting the `GatewayClient` is the
  quickest way to collect the information we need
- Response times will be recorded as histograms, failures as a counter.
  This number will be incremented for each failure, and we will be able
  to get the current failure rate using a function (ie.
  a derivative)
- We introduced a new class, `GatewayOrder`, containing request payload
  and type, used to produce a meaningful metrics' prefix.
- Our builders' structure is unnecessarily complex - we
  will refactor it in the next PR

## HOW 
_Steps to test or reproduce:_


kakumara:
General comment: 
As we will be getting more and more matrices in codebase it would be nicer to look out for options to collect them in a bit less intrusive way. As a first pass this could be fine, but we should try to find a pattern to do these without cluttering the application logic.  
Suggest looking out for AOP style options and this looks like it would be a nice place to use a around advice. Guice might have some support.

imranbohoran:
@kakumara This change set needed a lot of changes due to the various builders that is used. The actual part of capturing the metrics is really isolated to the GatewayClient. We have considered  the AOP approach for other things (i.e. resources etc) but need fleshing out.

kakumara:
@imranbohoran @simad agree, introducing AOP as it is would cause a lot of refactoring. So its fine for this one. 
But going forward this won't be the only thing we will be capturing matrix and possibly many more to come. So its better to come up with an approach (reusable) that can be used not only in here but all other microservices if/when we need matrix added.

PR #311, opened 2016-12-13
PP-1489 Fix special character issue with payment gateway request xmls

kakumara:
## WHAT
 - We are currently using a bit old version of freemarker (2.3.23). This is the best we could do with this version in order to fix this issue.

 - When we upgrade to dropwizard 1.0.x we should consider upgrading freemarker library as well. Where we could do this escaping when setting up the freemarker configuration (in one place). see setting up output formats in http://freemarker.org/docs/pgui_config_outputformatsautoesc.html


ijmad:
Looks like the right approach for now. We should upgrade dropwizard and freemarker soon though.

PR #314, opened 2016-12-15
PP-1435 [M&M] Add request type to gateway failures metrics

simad:
## WHAT
- We are currently recording metrics for gateway failures as
  `<gateway>.<gateway_type>.failures`. This allows us to get
  a general view of an issue ie. Worldpay live had X number of 
  failures
- By adding the request type we can break down failures further
  and understand when the problem actually occurred - during
  capture or authorisation, for example.
- The new shape of the metrics will be `<gateway>.<gateway_type>.
 <request_type>.failures`, ie. `worldpay.live.authorisation.failures`

## HOW 
_Steps to test or reproduce:_


ckalista:
LGTM

PR #315, opened 2016-12-16
PP-1457 Add capture timing information to the charges API

ijmad:
As per the API suggestion in the ticket.

To enable this work I have changed the sandbox provider so it produces CAPTURE_SUBMITTED as well as CAPTURED events as per other providers that support notifications. This was not the case before.

This has enabled some simplifications of the StateTransitions classes, as the state transitions are now identical between real and sandbox providers - I think this is of benefit.


simad:
rebuild

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
```diff
@@ -84,6 +84,75 @@ public String toString() {
...
+    @JsonFormat(shape = JsonFormat.Shape.OBJECT)
+    public static class SettlementSummary {
+        private ZonedDateTime captureSubmitTime, capturedTime;
+
+        public void setCaptureSubmitTime(ZonedDateTime captureSubmitTime) {
+            this.captureSubmitTime = captureSubmitTime;
+        }
+
+        @JsonProperty("capture_submit_time")
+        public String getCaptureSubmitTime() {
```
imranbohoran: Do we really want to do formatting here? If we must, returning null is probably the best choice if there's no submit date-time. Empty string or optionals perhaps

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
```diff
@@ -84,6 +84,75 @@ public String toString() {
...
+            return (captureSubmitTime != null) ? DateTimeUtils.toUTCDateString(captureSubmitTime) : null;
+        }
+
+        public void setCapturedTime(ZonedDateTime capturedTime) {
+            this.capturedTime = capturedTime;
+        }
+
+        @JsonProperty("captured_time")
+        public String getCapturedTime() {
```
imranbohoran: Ditto

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
```diff
@@ -84,6 +84,75 @@ public String toString() {
...
+            return (capturedTime != null) ? DateTimeUtils.toUTCDateString(capturedTime) : null;
+        }
+
+        @Override
+        public boolean equals(Object o) {
```
imranbohoran: We can derive a more concise implementation from the IDE provided implementation perhaps.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -183,6 +182,23 @@ public long getRefundedAmount() {
                 .sum();
     }
 
+    public ZonedDateTime getCaptureSubmitTime() {
```
imranbohoran: Can this not be an optional instead of returning null

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -183,6 +182,23 @@ public long getRefundedAmount() {
...
+        return this.events.stream()
+                .filter(e -> e.getStatus().equals(CAPTURE_SUBMITTED))
+                .findFirst()
+                .map(ChargeEventEntity::getUpdated)
+                .orElse(null);
+    }
+
+    public ZonedDateTime getCapturedTime() {
```
imranbohoran: Ditto

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
```diff
@@ -71,10 +87,13 @@ public void shouldCaptureAChargeForANonSandboxAccount() throws Exception {
...
-        ArgumentCaptor<ChargeEntity> argumentCaptor = ArgumentCaptor.forClass(ChargeEntity.class);
-        verify(mockedChargeDao).mergeAndNotifyStatusHasChanged(argumentCaptor.capture());
+        ArgumentCaptor<ChargeEntity> chargeEntityCaptor = ArgumentCaptor.forClass(ChargeEntity.class);
+        ArgumentCaptor<Optional> bookingDateCaptor = ArgumentCaptor.forClass(Optional.class);
+
+        verify(mockedChargeDao).mergeAndNotifyStatusHasChanged(chargeEntityCaptor.capture(), bookingDateCaptor.capture());
 
-        assertThat(argumentCaptor.getValue().getStatus(), is(CAPTURE_SUBMITTED.getValue()));
+        assertThat(chargeEntityCaptor.getValue().getStatus(), is(CAPTURE_SUBMITTED.getValue()));
+        assertFalse(bookingDateCaptor.getValue().isPresent());
```
imranbohoran: Do we really need a captor for this. If we are only checking that there's an value, it can be any non-empty optional date(?)

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
simad: I guess the reason the formatting right there is that the settlement is automagically serialised by Jackson, and thanks to the `@JsonInclude(Include.NON_NULL)` annotation, those fields will be completely missing from the json (as designed) if they are null

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
simad: I would indeed treat it as an optional, but again, ChargeEntity is what gets serialised by Jackson and I am not sure it can handle optionals' serialisation correctly. If you know that that will work, I am more than happy to change that :)

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
simad: Nice catch - we can just check that it's `Optional.empty()`

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
simad: That is the IntelliJ default implementation, unfortunately - just checked. Yes, it looks very ugly! A more concise one would come from using Guava or Apache's hashcode/equals builder, but I don't think it's worth the effort.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
simad: See above.

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
imranbohoran: Hmmm. Ugly it is, but I guess treating this as a purely jackson parsable and being a bit jackson specific make this ok.

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
imranbohoran: I got a different more concise one. If it's easy enough to change, I'd still do it as I don't think we'll come back to this at any point.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
imranbohoran: I'm sure Jackson can handle java 8 optionals. If we have it configured properly is not something I'm certain of. From a cursory look, it doesn't appear to be. I leave it to you If you want to do that as part of this work - https://github.com/FasterXML/jackson-modules-java8

imranbohoran:
👍

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
simad: Created https://payments-platform.atlassian.net/browse/PP-1578 to record tech debt about this

PR #316, opened 2016-12-28
PP-1434 [M&M] Add metrics for Notify's response times and failures

simad:
## WHAT
- Adding `notify-operations.response_time` and
  `notify-operations.failures` metrics.
- Response time is updated no matter if a call to notify fails
  or succeeds
- Failures are reported as a counter (ever-incrementing) so in order
  to capture a sensible metrics for graphing/alerting, it's necessary
  to calculate the failure rate (ie. NonNegativeDerivative, see the
  payment gateway dashboard on HostedGraphite)
## HOW 
_Steps to test or reproduce:_


src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -56,7 +56,11 @@ public UserNotificationService(NotifyClientProvider notifyClientProvider, Connec
                     return Optional.of(response.getNotificationId());
                 } catch (NotificationClientException e) {
                     logger.error("Failed to send confirmation email - charge_external_id=" + chargeEntity.getExternalId(), e);
+                    metricRegistry.counter("notify-operations.failures").inc();
```
imranbohoran: Do we want the notificationId included in the metric namespace? we might get false negatives if the there are failures in tests.

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -56,7 +56,11 @@ public UserNotificationService(NotifyClientProvider notifyClientProvider, Connec
...
                     return Optional.empty();
+                } finally {
+                    responseTimeStopwatch.stop();
+                    metricRegistry.histogram("notify-operations.response_time").update(responseTimeStopwatch.elapsed(TimeUnit.MILLISECONDS));
```
imranbohoran: Same comment as above about metric name spaces to include the serviceId, which would be useful to separate test vs production.

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
imranbohoran: Having gone through the code and how serviceIDs are used, I have convinced my self the additional namespacing is not necessary. We are using the same serviceid and plus the metrics are going to grouped in the relevant environment's node as well.

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
imranbohoran: Same as above.

imranbohoran:
rebuild

imranbohoran:
👍

PR #317, opened 2017-01-11
PP-1434 Using the correct MetricRegistry

imranbohoran:
* The MetricRegistry in the dropwizard environment is configured with
   the correct graphite host, namespacing and timings. Therefore we must
   use the MetricRegistry from the dropwizard environment. If not guice
   would create a new MetricRegistry which is not configured and we wont
   get any metrics.


PauloPortugal:
:+1:

PR #318, opened 2017-01-16
PP-1471 Refactoring of OrderRequestBuilders and GatewayClient to enable future support of ePDQ requests

ckalista:
## WHAT
_A brief description of the pull request:

### `GatewayClient` refactoring
   - Order requests in ePDQ require credentials to be submitted in the body of the request rather than using HTTP Authentication. No HTTP Authentication headers should be sent.
   - Order requests in ePDQ must be sent in an application/x-www-form-urlencoded body (with the correct Content-Type request header).
   - Unlike other providers which submit all operations to a simple endpoint, ePDQ expects authorisation request orders to be submitted to a different end-point (orderdirect.asp) than for other operations which are submitted to the maintenance end-point (maintenancedirect.asp).

### `OrderRequestBuilder`s refactoring
The request builders must be refactored to promote maximum code reuse and reduce the following limitations. OrderRequestBuilders are used to build the request payloads of all order operations that we currently support and for all payment providers (Worldpay and SmartPay). Typically those builders are organised by type of operation, that is for instance OrderCaptureRequestBuilder is the builder of choice for all capture requests for all providers. This choice for organising the code implies a different builder is required per type of operation which quickly leads to a lot of code duplication especially since most request payloads share commonalities. Another limitation of this choice of code slicing is the lack of flexibility we have in terms of implementation. It has to be the same implementation for all providers where we often need a different implementation per provider. For instance, ePDQ mandates the request payloads to be `application/x-www-form-urlencoded` as opposed as XML which would lead to a totally different implementation than for the other providers.
   - `OrderRequestBuilder` is the base abstract class laying the foundations for all request builders.  It encapsulates most parameters that all payment providers have in common regardless of the type of operation. Its constructor requires the type of operation to be specified i.e. one of the supported operations defined by `OrderRequestType`. Its constructor expects a concrete implementation of `PayloadStringBuilder` which takes care of the actual payload generation when the builder's build method is called.
   - We currently support two different implementations for payload generation:
      - `FormUrlEncodedStringBuilder` used for ePDQ request orders. The implementation is based on an URL encoding library producing a query string like URL encoded payload.
      - `TemplateStringBuilder` used for Worldpay and Smartpay request orders. The implemtation is based on an XML templating engine producing the XML payload.
   - Each payment provider has its own builder extending `OrderRequestBuilder` with custom parameters in the payload generation: `WorldpayOrderRequestBuilder`, `SmartpayOrderRequestBuilder`, `EpdqOrderRequestBuilder`.


PauloPortugal:
As mentioned with @ckalista this PR should have been broken down into several commits, although the approach to sit down with the reviewer and drive through the code has smoothed the review process.
Strong usage of design patterns and SOLID principles were used. :+1: 
Happy to approve once it passes the E2E tests

ckalista:
rebuild

ckalista:
rebuild

PR #322, opened 2017-01-20
PP-1588 Add indices on charge_events, refunds and card_types

simad:
## WHAT
- When making a call to `/charges` connector endpoint (loading
  the transactions' list, for example) we are making these calls
  in the background, for each charge we retrieve:

SELECT ID, gateway_event_date, STATUS, UPDATED, version, charge_id
FROM charge_events WHERE (charge_id = ?) ORDER BY UPDATED DESC
SELECT ID, AMOUNT, created_date, external_id, reference, STATUS,
version, charge_id FROM refunds WHERE (charge_id = ?)
SELECT ID, BRAND, LABEL, TYPE, version FROM card_types WHERE (BRAND =
?)

- These are super slow without an index, so we are adding indices to
  those tables


simad:
rebuild

PR #323, opened 2017-01-23
PP-1588 Fixing flaky tests

imranbohoran:
- These tests have been flaky due to the timings. Therefore
  not checking for created date values to be within time ranges.
  We have other tests that assert created date values more precisely.
  Just asserting that there is a created date.


imranbohoran:
rebuild

simad:
:+1:

PR #324, opened 2017-01-25
PP-1582 Ignore unknown json properties from payment details

imranbohoran:
* As part of the 3DS work, we have to pass in the accept header
   and user agent header. To pass this from the frontend, we have to
   first let connector accept unknown properties.
 * This is temporary. Once frontend is passing the values, this code will be
   changed to remove the ignore and accept the specific headers.

with @alexbishop1


imranbohoran:
rebuild

imranbohoran:
rebuild

imranbohoran:
rebuild

rauligs:
👍

PR #325, opened 2017-01-25
PP-1620 [REVIEW-ONLY] Force 'https' schema on links pagination (charges)

rauligs:
## WHAT
- A support ticket was raised because download CSV wasn't working for paginated search. In the deployed app, 'next_page' links are coming with 'tcp://' schema coming from connector, this is due to load balancers making the calls using this schema rather than the original _https_ schema.
- A story has been opened https://payments-platform.atlassian.net/browse/PP-1622 to deal with this situation globally for all our Dropwizard apps.


rauligs:
Closed since PP-1622 is the approach suggested from the beginning and simple enough

PR #326, opened 2017-01-25
Upgrade dropwizard to 0.9.3

ijmad:
Upgrade to DropWizard 0.9.3. This contains some newer dependencies.
Remove old/broken/unused newrelic JARs.


ijmad:
rebuild

src/main/java/uk/gov/pay/connector/model/domain/NotificationCredentials.java
```diff
@@ -12,9 +12,8 @@
 public class NotificationCredentials extends AbstractEntity {
 
     @OneToOne
-    @JsonIgnore
     @JoinColumn(name = "account_id", nullable = false)
-    @JsonManagedReference
```
alexbishop1: Why this change?

ijmad: As in the commit message:

@JsonManagedDependency not matched with @JsonBackreference (as it should be) seems to upset Jackson 2.6.7 more than it upset previous versions. Remove this as the intention is not to return these fields, as @JsonIgnore is present anyway.

alexbishop1: Sorry, missed that.

alexbishop1 (approved):
Looks good 👍

PR #327, opened 2017-01-26
PP-1622 Rewrite request scheme to 'https'

rauligs:
## WHAT
- Some requests are not guarantee they are 'https' (likely to come as 'tcp').
- Add filter to rewrite scheme to 'https' so any usage of _UriInfo_ will return the correct URI builder (used for creating links).
- This solves issues as PP-1620 where unexpected links with 'tcp' scheme are generated and are having some issues with certain REST clients.
- See [PP-1620](https://payments-platform.atlassian.net/browse/PP-1620)


mrlumbu (approved):

mrlumbu:
👍

PR #329, opened 2017-01-30
PP-1582 Pass through 3DS specific headers for Worldpay Authorisation

imranbohoran:
* Authorisation request from frontend contains the Accept and User-Agent
   headers that are required for 3DS orders. This change is for the authorise
   endpoint to accept the new values in the post data.
 * Given that we are passing header information along with the card data for
   authorisation, this is not just card data anymore. Therefore many files have been
   refactored to be renamed as AutorisationDetails.
 * Including these fields does not have any impact on orders created for merchant codes
   without 3DS enabled.

with @alexbishop1


imranbohoran:
rebuild

imranbohoran:
rebuild

imranbohoran:
rebuild

imranbohoran:
rebuild

simad:
👍

PR #332, opened 2017-02-06
PP-1645 Add authorisation response status metrics

mrlumbu:
In order to better diagnose issues with payment gateways, this adds two metrics:
- per gateway account counter of authorisations success, failure, error
- per service name counter authorisations success, failure, error

with @markyjones


src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
```diff
@@ -59,6 +63,15 @@ public void setupPaymentProviderMock(String transactionId, boolean isAuthorised,
         when(mockedPaymentProvider.authorise(any())).thenReturn(authorisationResponse);
     }
 
+    @BeforeClass
```
alexbishop1: I don’t understand why this isn’t just a regular `@Before` (with `mockMetricRegistry` being non-static).

alexbishop1 (approved):
👍

PR #333, opened 2017-02-06
PP-1582 Handle 3DS for Worldpay

imranbohoran:
## WHAT

- Handles 3DS required responses from Worldpay and sends appropriate response to the frontend.
- The 3DS details are returned as part of the charge response.
- See commits for details.

with @alexbishop1


alexbishop1:
Rebuild

alexbishop1:
Rebuild

src/main/java/uk/gov/pay/connector/model/api/ExternalChargeState.java
```diff
@@ -14,6 +14,7 @@
     EXTERNAL_CREATED("created", false),
 
     EXTERNAL_STARTED("started", false),
+    EXTERNAL_3DS_STARTED("3DS started", false),
```
ckalista: Are we sure we want to introduce a new state at this stage? I am just wondering if the impact has been considered on public API and self service, i.e. filtering works, displaying of new status is correct etc...?

src/main/java/uk/gov/pay/connector/model/domain/DefaultStateTransitions.java
```diff
@@ -13,7 +13,7 @@
 
             .put(CREATED,               validTransitions(ENTERING_CARD_DETAILS, SYSTEM_CANCELLED, EXPIRED))
             .put(ENTERING_CARD_DETAILS, validTransitions(AUTHORISATION_READY, EXPIRED, USER_CANCELLED, SYSTEM_CANCELLED))
-            .put(AUTHORISATION_READY,   validTransitions(AUTHORISATION_SUCCESS, AUTHORISATION_REJECTED, AUTHORISATION_ERROR))
+            .put(AUTHORISATION_READY,   validTransitions(AUTHORISATION_SUCCESS, AUTHORISATION_REJECTED, AUTHORISATION_ERROR, AUTHORISATION_3DS_REQUIRED))
```
ckalista: Maybe ahead of myself here but should we not also include new transitions from `AUTHORISATION_3DS_REQUIRED` or is this part of the next PR?

ckalista (commented):
Code review completed. Looks good.

src/main/java/uk/gov/pay/connector/model/api/ExternalChargeState.java
imranbohoran: We discussed this and we are now removing this status.

src/main/java/uk/gov/pay/connector/model/domain/DefaultStateTransitions.java
imranbohoran: Yes, this will be in the next story.

PR #336, opened 2017-02-10
PP-1597 Send 3DS response to gateway

alexbishop1:
Adds an endpoint (`/v1/frontend/charges/{chargeId}/3ds`) to receive the result of the payer’s 3D Secure interaction with the card issuer from frontend (essentially, just a `pa_response`). When receiving data on this endpoint, the state of the charge is changed from `AUTHORISATION_3DS_REQUIRED` to `AUTHORISATION_3DS_READY` before the data is sent to Worldpay (the “second order”), which replies with an authorised/rejected/cancel/error response (just like for a non-3DS authorisation), which we then handle (changing the state as appropriate).

with @imranbohoran


src/main/java/uk/gov/pay/connector/resources/CardResource.java
```diff
@@ -114,4 +122,10 @@ private Response handleGatewayResponse(GatewayResponse<? extends BaseResponse> r
                 .map(this::handleError)
                 .orElseGet(ResponseUtil::noContentResponse);
     }
+
+    private static boolean transactionDeclined(GatewayResponse<BaseAuthoriseResponse> response) {
+        return response.getBaseResponse()
+                .map(baseResponse -> (baseResponse.authoriseStatus() == AuthoriseStatus.REJECTED || baseResponse.authoriseStatus() == AuthoriseStatus.ERROR))
```
simad: Being quite picky here, this would be more expressive using a `filter` rather than a `map`

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
```diff
@@ -0,0 +1,81 @@
...
+        return operationResponse;
+    }
+
+    @Override
+    protected ChargeStatus[] getLegalStates() {
+        return new ChargeStatus[]{
+                AUTHORISATION_3DS_REQUIRED
+        };
+    }
+}
```
simad: Newline :)

src/main/java/uk/gov/pay/connector/service/CardAuthoriseBaseService.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public CardAuthoriseBaseService(ChargeDao chargeDao, PaymentProviders providers, CardExecutorService cardExecutorService, MetricRegistry metricRegistry) {
+        super(chargeDao, providers, metricRegistry);
+        this.cardExecutorService = cardExecutorService;
+    }
+
+
+    public GatewayResponse doAuthorise(String chargeId, T gatewayAuthRequest) {
+
+        Optional<ChargeEntity> chargeEntityOptional = chargeDao.findByExternalId(chargeId);
```
simad: We followed the convention of expressing optionals as 'maybe', ie. `chargeEntityMaybe`, in other parts of the codebase; we could keep that consistent, here.

src/main/java/uk/gov/pay/connector/service/CardAuthoriseBaseService.java
```diff
@@ -0,0 +1,75 @@
...
+
+        if (chargeEntityOptional.isPresent()) {
```
simad: any particular reason to use the `ifPresent` `else`  constructs instead of just mapping on the optional? I think `map` `orElseThrow` would read better

src/test/resources/templates/worldpay/valid-3ds-response-auth-worldpay-request.xml
```diff
@@ -0,0 +1,13 @@
...
+<paymentService version="1.4" merchantCode="GBSRECABOFFGDSGBP">
+    <submit>
+        <order orderCode="MyUniqueTransactionId!">
+            <info3DSecure>
+                <paResponse>I am an opaque 3D Secure PA response from the card issuer</paResponse>
+            </info3DSecure>
+            <session id="uniqueSessionId"/>
+        </order>
+    </submit>
+</paymentService>
```
simad: Newline :)

src/main/java/uk/gov/pay/connector/model/domain/DefaultStateTransitions.java
```diff
@@ -11,15 +11,17 @@
...
-            .put(AUTHORISATION_SUCCESS, validTransitions(CAPTURE_READY, SYSTEM_CANCEL_READY, USER_CANCEL_READY, EXPIRE_CANCEL_READY))
-            .put(CAPTURE_READY,         validTransitions(CAPTURE_SUBMITTED, CAPTURE_ERROR))
-            .put(CAPTURE_SUBMITTED,     validTransitions(CAPTURED)) // can this ever be a capture error?
-            .put(EXPIRE_CANCEL_READY,   validTransitions(EXPIRE_CANCEL_FAILED, EXPIRED))
-            .put(SYSTEM_CANCEL_READY,   validTransitions(SYSTEM_CANCEL_ERROR, SYSTEM_CANCELLED))
-            .put(USER_CANCEL_READY,     validTransitions(USER_CANCEL_ERROR, USER_CANCELLED))
+            .put(CREATED,                       validTransitions(ENTERING_CARD_DETAILS, SYSTEM_CANCELLED, EXPIRED))
+            .put(ENTERING_CARD_DETAILS,         validTransitions(AUTHORISATION_READY, EXPIRED, USER_CANCELLED, SYSTEM_CANCELLED))
+            .put(AUTHORISATION_READY,           validTransitions(AUTHORISATION_SUCCESS, AUTHORISATION_REJECTED, AUTHORISATION_ERROR, AUTHORISATION_3DS_REQUIRED))
+            .put(AUTHORISATION_3DS_REQUIRED,    validTransitions(AUTHORISATION_3DS_READY))
```
simad: I guess there is no valid transition to expire or cancel because at this point there hasn't been any authorisation from worldpay, so if a user abandons the journey here, there's no need to do anything. Is this a correct assumption?

alexbishop1: Yes, from our perspective there will be no payment in Worldpay yet.

src/main/java/uk/gov/pay/connector/resources/CardResource.java
alexbishop1: Changed to use `filter` instead of `map`

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
alexbishop1: Newline appended!

src/main/java/uk/gov/pay/connector/service/CardAuthoriseBaseService.java
alexbishop1: After addressing your other comment for this file, this temporary variable became redundant and we removed it

src/main/java/uk/gov/pay/connector/service/CardAuthoriseBaseService.java
alexbishop1: Actually had it using `map`/`orElseThrow` earlier but it got lost in an aggressive bout of copy-pasting. Reinstated now with `ifPresent` banished.

src/test/resources/templates/worldpay/valid-3ds-response-auth-worldpay-request.xml
alexbishop1: Newline appended and IntelliJ _Ensure line feed at file end on Save_ setting enabled!

alexbishop1:
Rebuild

simad:
👍

PR #337, opened 2017-02-13
PP-1669 Map Worldpay cancel to correct states

alexbishop1:
Map Worldpay `CANCELLED` status to appropriate internal and external states, rather than error states

- **Before**: internal state was `AUTHORISATION_ERROR`, external state was `EXTERNAL_GATEWAY_ERROR`
- **After**: internal state is `AUTHORISATION_CANCELLED`, external state is `EXTERNAL_FAILED_REJECTED`

Status transition diagram updated at - https://docs.google.com/drawings/d/1q8mkwGuL8roJjpIMspQWRMb71hLIoEoDSF81c0tyWik/edit


markyjones (approved):
ChargeSeachParamsTest has a fairly random // value = comment . not sure if that was intended or useful?

imranbohoran:
@markyjones Changed the method name as discussed.

PR #339, opened 2017-02-14
PP-1645 Use the MatricRegistry already configured for the Environment

imranbohoran:
## WHAT

 * We should be using the MetricRegistry that is already configured
   against the active Dropwizard environment. When using @inject annotation
   if MetricRegistry is passed, Guice would instentiate a new MetricRegistry
   as we don't register a singleton of a MetricRegistry in guice. What we have
   is an Environment singleton that has a MetricRegistry configured within it.


markyjones:
@heathd as an FYI, does look like a DI thing.

imranbohoran:
rebuild

markyjones:
Looks good. Will raise a Ticket re: the failed rebuilds. https://payments-platform.atlassian.net/browse/PP-1690

PR #341, opened 2017-02-16
PP-1669 Allowing state transitions to AUTHORISED_CANCELLED

imranbohoran:
## WHAT
 * The state transition blocks moving to AUTHORISATION_CANCELLED if
   we don't specify it in the state transition model.
 * Added tests to verify this.

with @alexbishop1


ckalista:
Code review completed.

PR #345, opened 2017-02-22
PP-1713 Extracting machine cookie name for worldpay.

imranbohoran:
## WHAT

This is the first part of the change to support extracting the session cookie as a response identifier.
Still not storing it though

 * The GatewayClient now captures the response cookies.
 * The GatewayResponse has a concept of a responseIdentifier. And a new builder
   is responsible in creating a GatewayResponse.
 * When mapping the response from the gateway into a  GatewayResponse object,
   the responsibility of how it's parsed is handed over to the individual
 * PaymentProvider implementations. The function to work out the responseIdentifier
   is passed in as a function to the sendReceive method.


src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
```diff
@@ -21,8 +21,11 @@
 
     protected Either<GatewayError, T> response;
 
-    private GatewayResponse(T baseResponse) {
+    private String responseIdentifier;
```
alexbishop1: I like that this is abstracted from the specifics of the Worldpay `machine` cookie but I’m not convinced about the name… `responseIdentifier` makes me think of something that’s unique to each response. Perhaps `sessionIdentifier` might be a better name (as it’s being used for session management, essentially).

I spent a bit of time pondering if perhaps this belongs inside implementations of `BaseResponse` but decided it’s fine here (since it’s metadata rather than part of the response itself).

But this should be `final` 😀

src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
```diff
@@ -66,23 +73,55 @@ public boolean isFailed() {
...
+                return new GatewayResponse<>(gatewayError);
+            }
+
+            Optional<String> errorCode = getErrorCode(response);
+            Optional<String> errorMessage = getErrorMessage(response);
+
+            if (errorCode.isPresent() || errorMessage.isPresent()) {
+                StringBuilder sb = new StringBuilder();
+                sb.append(errorCode.map(e -> format("[%s] ", e)).orElse(""));
+                sb.append(errorMessage.orElse(""));
```
alexbishop1: I know you’ve just moved this code but all this `orElse("")` could be avoided:

```
errorCode.ifPresent(e -> sb.append(format("[%s] ", e)));
errorMessage.ifPresent(sb::append);
```

src/main/java/uk/gov/pay/connector/service/BasePaymentProvider.java
```diff
@@ -23,28 +25,43 @@ public BasePaymentProvider(GatewayClient client, boolean isNotificationEndpointS
         this.notificationDomain = notificationDomain;
     }
 
-    protected <U extends GatewayRequest> GatewayResponse sendReceive(U request, Function<U, GatewayOrder> order, Class<? extends BaseResponse> clazz) {
-        return sendReceive(null, request, order, clazz);
+    protected <U extends GatewayRequest> GatewayResponse sendReceive(U request, Function<U, GatewayOrder> order,
+                                                                     Class<? extends BaseResponse> clazz,
+                                                                     Function<GatewayClient.Response, Optional<String>> responseIdentifier) {
```
alexbishop1: Some people frown on using `Optional` for method parameters (but I’m not going to insist on changing this).

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
```diff
@@ -128,10 +129,14 @@ private void incrementFailureCounter(MetricRegistry metricRegistry, String metri
...
         private final int status;
         private final String entity;
+        private final Map<String, String> responseCookies = new HashMap<>();
 
         protected Response(final javax.ws.rs.core.Response delegate) {
             this.status = delegate.getStatus();
             this.entity = delegate.readEntity(String.class);
+            delegate.getCookies().entrySet().forEach(cookie -> {
+                responseCookies.put(cookie.getKey(), cookie.getValue().getValue());
+            });
```
alexbishop1: You can do this without the intermediate `entrySet()`:

```
delegate.getCookies().forEach((name, cookie) -> responseCookies.put(name, cookie.getValue()));
```

src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayPaymentProvider.java
```diff
@@ -137,4 +137,11 @@ public StatusMapper getStatusMapper() {
     private String getMerchantCode(GatewayRequest request) {
         return request.getGatewayAccount().getCredentials().get(CREDENTIALS_MERCHANT_ID);
     }
+
+    private Function<GatewayClient.Response, Optional<String>> extractResponseIdentifier() {
+        return response -> {
+            Optional<String> emptyResponseIdentifierForSmartpay = Optional.empty();
+            return emptyResponseIdentifierForSmartpay;
+        };
```
alexbishop1: This could just be:

```
return response -> { return Optional.empty(); };
```

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -147,4 +149,13 @@ public StatusMapper getStatusMapper() {
...
                 .build();
     }
+
+    private Function<GatewayClient.Response, Optional<String>> extractResponseIdentifier() {
+        return response -> {
+            if(response.getResponseCookies().containsKey(WORLDPAY_MACHINE_COOKIE_NAME)) {
+                return Optional.ofNullable(response.getResponseCookies().get(WORLDPAY_MACHINE_COOKIE_NAME));
+            }
+            return Optional.empty();
+        };
```
alexbishop1: You don’t need to do the `containsKey(WORLDPAY_MACHINE_COOKIE_NAME)` check: `get(WORLDPAY_MACHINE_COOKIE_NAME)` will return `null` if the key isn’t present and you’re already using `Optional.ofNullable(…)`!

Just do:

```
return Optional.ofNullable(response.getResponseCookies().get(WORLDPAY_MACHINE_COOKIE_NAME));
```

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
```diff
@@ -26,6 +26,8 @@
 
 public class WorldpayPaymentProvider extends BasePaymentProvider<BaseResponse> {
 
+    private static final String WORLDPAY_MACHINE_COOKIE_NAME = "machine";
```
alexbishop1: Maybe `WORLDPAY_SESSION_COOKIE_NAME`? (Their “machine” terminology is a bit opaque to us.)

src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
```diff
@@ -110,6 +110,7 @@ public void shouldBeAbleToSendAuthorisationRequestForMerchantUsing3ds() throws E
         GatewayResponse<WorldpayOrderStatusResponse> response = successfulWorldpayCardAuthFor3ds(connector);
 
         assertTrue(response.getBaseResponse().isPresent());
+        assertTrue(response.getResponseIdentifier().isPresent());
```
alexbishop1: Maybe have the inverse in the `SmartpayPaymentProviderTest`?

alexbishop1 (commented):
General approach is good (makes me a little sad that the other payment providers get mildly tainted by the existence of the Worldpay-specific stuff but you’ve managed to keep it generic and at least theoretically abstract).

Not going to refuse approval if any of the other comments aren’t addressed as they’re all fairly minor.

src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
imranbohoran: Making it final means that this then needs be assigned to something when constructing a GatewayResponse with a gateway error. I didn't want to make a null assignment. And initially had the field declared as Optional. Then again not a fan of making fields optional, so instead made the getter return Optional. 

With regard to the name, I didn't want to put any "session" term in here because that is based on our knowledge of worldpay. Hence went with something that corresponds to the response. I don't feel too strongly on this though. So if you think calling it sessionIdentifier is still worth it after my reasoning, I can do it.

src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
imranbohoran: 👍 Will do this.

src/main/java/uk/gov/pay/connector/service/BasePaymentProvider.java
imranbohoran: Yeah, since this is the return type of the function and I wanted the function to return an Optional I went this.

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
imranbohoran: 👍 Will address this

src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayPaymentProvider.java
imranbohoran: Yes it could be. I used the variable to explicitly say it's empty as oppose to leaving it to be decided if this was missed in implementation. The compiler will inline anyway, I like seeing named things when/if they make sense.

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
imranbohoran: 👍 Yes - can change this. Since this is worldpay specific and since it the worldpay implementation it makes sense.

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
imranbohoran: 👍 True. And can even use a lambda expression as oppose to the return statement.

src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
imranbohoran: I didn't bother because, SmartPaymentProviderTest is not executed locally nor is it executed in Jenkins. I only used the Worldpay one to  get some assurance. Which then reminds me that I didn't modify the unit test for this. So I'll do that as part of this PR.

src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
alexbishop1: I missed that the field isn’t assigned when there’s a `GatewayError` so forget what I said about `final`.

I’m still conflicted on the name… I’d like something that indicates the intent but you’re right that “session” assumes something about what it means to someone else. Would `reponseMetadata` be too generic? That’s essentially what it is: metadata about the response we want to keep around.

src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayPaymentProvider.java
alexbishop1: Fair enough.

src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
imranbohoran: I'll put you out of your misery. It's now sessionIdentifier as agreed.

src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
imranbohoran: Done

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
imranbohoran: Done

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
imranbohoran: Done

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProvider.java
imranbohoran: Done

src/test/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -349,9 +355,13 @@ private void mockWorldpayResponse(int httpStatus, String responsePayload) {
         when(mockTarget.request()).thenReturn(mockBuilder);
         when(mockBuilder.header(anyString(), anyObject())).thenReturn(mockBuilder);
 
+        Map<String, NewCookie> responseCookies = new HashMap<>();
+        responseCookies.put(WORLDPAY_MACHINE_COOKIE_NAME, NewCookie.valueOf("value-from-worldpay"));
```
alexbishop1: Really minor, but you could just do `Map<String, NewCookie> responseCookies = Collections.singletonMap(WORLDPAY_MACHINE_COOKIE_NAME, NewCookie.valueOf("value-from-worldpay"))` here.

alexbishop1 (approved):
Probably won’t bring about the apocalypse

imranbohoran:
rebuild

PR #346, opened 2017-02-23
Copyright should be Crown Copyright

annashipman:
It's permissible to put the organisation in brackets, but the copyright belongs to the Crown. Otherwise, this licence is correct.

@markyjones As discussed, you can copy this to the other licences.


markyjones:
Thanks!

markyjones (approved):

PR #347, opened 2017-02-23
PP-1713 Store "machine" cookie and send it in 3DS orders

imranbohoran:
## WHAT
This PR takes care of storing the worldpay "machine" cookie as a sessionIdentifier and sends it back in the second order to worldpay in case of a 3DS order.
For smartpay, this will not be included as there isn't any session identifiers known to us at the moment.

See the commits for more details.


src/main/java/uk/gov/pay/connector/model/gateway/Auth3dsResponseGatewayRequest.java
```diff
@@ -30,6 +30,10 @@ public String getChargeExternalId() {
         return String.valueOf(charge.getExternalId());
     }
 
+    public Optional<String> getProviderSessionid() {
```
alexbishop1: `getProviderSessionId()` please.

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
```diff
@@ -66,12 +71,17 @@ public static GatewayClient createGatewayClient(Client client, Map<String, Strin
...
+            Builder requestBuilder = client.target(gatewayUrl)
                     .request()
                     .header(AUTHORIZATION, encode(
                             account.getCredentials().get(CREDENTIALS_USERNAME),
-                            account.getCredentials().get(CREDENTIALS_PASSWORD)))
-                    .post(Entity.entity(request.getPayload(), mediaType));
+                            account.getCredentials().get(CREDENTIALS_PASSWORD)));
+
+                    response = request.getProviderSessionId()
+                            .map(sessionId -> requestBuilder.cookie(sessionIdentifierName, sessionId))
```
alexbishop1: Ugh. Is it feasible to do something like have a function or a decorator or something passed in from the `WorldpayPaymentProvider` that modifies the response as appropriate to keep this very Worldpay-specific logic out of our generic `GatewayClient`?

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -822,6 +822,27 @@ public void shouldCreateANewChargeWith3dsDetails() {
...
+        gatewayAccount.setId(defaultTestAccount.getAccountId());
+
+        String providerSessionId = "provider-session-id-value";
+        ChargeEntity chargeEntity = aValidChargeEntity()
+                .withId(null)
+                .withGatewayAccountEntity(gatewayAccount)
+                .withProviderSessionId(providerSessionId)
+                .build();
+
+        assertThat(chargeEntity.getId(), is(nullValue()));
```
alexbishop1: Do we really care about that for this specific test?

joshmyers:
rebuild

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
```diff
@@ -227,6 +230,31 @@ public void shouldStoreCardDetailsEvenIfInAuthorisationError() {
     }
 
     @Test
+    public void shouldStoreProviderSessionIdIfAuthorisationSuccess() {
+        String transactionId = "transaction-id";
+        AuthCardDetails authCardDetails = AuthUtils.aValidAuthorisationDetails();
+        anAuthorisationSuccessResponse(charge, reloadedCharge, transactionId, authCardDetails);
+
+        assertThat(reloadedCharge.getProviderSessionId(), is(SESSION_IDENTIFIER));
```
alexbishop1: It feels a bit odd that this test spends a lot of time configuring transaction IDs (which we don’t care about here) but all the session ID configuration (which we do care about) happens elsewhere.

Might be nice if we had variants of `anAuthorisationSuccessResponse` and `setupPaymentProviderMock` that take the provider session ID but generate the transaction ID so we only have to specify what we care about (but that might be a bit much for this).

src/test/java/uk/gov/pay/connector/service/GatewayClientTest.java
```diff
@@ -0,0 +1,210 @@
...
+        when(mockGatewayAccountEntity.getCredentials()).thenReturn(credentialMap);
+        when(mockGatewayOrder.getOrderRequestType()).thenReturn(OrderRequestType.AUTHORISE);
+        when(mockGatewayOrder.getPayload()).thenReturn(orderPayload);
+        when(mockGatewayOrder.getProviderSessionId()).thenReturn(Optional.empty());
+
+        when(mockClient.target(WORLDPAY_API_ENDPOINT)).thenReturn(mockWebTarget);
+        when(mockWebTarget.request()).thenReturn(mockBuilder).thenReturn(mockBuilder);
+        when(mockBuilder.header(AUTHORIZATION, encode("user", "password"))).thenReturn(mockBuilder);
+        when(mockBuilder.cookie(any(Cookie.class))).thenReturn(mockBuilder);
+        when(mockBuilder.post(Entity.entity(orderPayload, MediaType.APPLICATION_XML_TYPE))).thenReturn(mockResponse);
```
alexbishop1: As best I can tell, this is the first line that differs from previous test. Can we move some of the common mock creation, stubbing etc. into the `@Before` so the tests are shorter and easier to understand?

src/test/java/uk/gov/pay/connector/service/GatewayClientTest.java
```diff
@@ -0,0 +1,210 @@
...
+        when(mockResponse.getStatus()).thenReturn(200);
+
+        gatewayClient.postRequestFor(null, mockGatewayAccountEntity, mockGatewayOrder);
+
+        verify(mockBuilder).cookie(WorldpayPaymentProvider.WORLDPAY_MACHINE_COOKIE_NAME, providerSessionid);
```
alexbishop1: Rather than using `verify`, you can check this using more sophistocated stubbing:

```
Cookie sessionCookie = // create a Cookie that’s equal to what we expect or use an argument captor etc.
when(mockBuilder.cookie(sessionCookie)).thenReturn(mockBuilderWithSessionCookie);
when(mockBuilderWithSessionCookie.post(Entity.entity(orderPayload, MediaType.APPLICATION_XML_TYPE))).thenReturn(mockResponse);
```

This would give us a more thorough test because it ensures the desired cookie is the one that’s actually set and makes sure the cookie is added before the request is sent.

joshmyers:
rebuild

src/test/java/uk/gov/pay/connector/service/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -355,8 +418,8 @@ private void mockWorldpayResponse(int httpStatus, String responsePayload) {
         when(mockTarget.request()).thenReturn(mockBuilder);
         when(mockBuilder.header(anyString(), anyObject())).thenReturn(mockBuilder);
 
-        Map<String, NewCookie> responseCookies = new HashMap<>();
-        responseCookies.put(WORLDPAY_MACHINE_COOKIE_NAME, NewCookie.valueOf("value-from-worldpay"));
+        Map<String, NewCookie> responseCookies =
+                Collections.singletonMap(WORLDPAY_MACHINE_COOKIE_NAME, NewCookie.valueOf("value-from-worldpay"));
```
alexbishop1: 👍

src/main/java/uk/gov/pay/connector/service/GatewayClient.java
imranbohoran: As discussed, I completely agree on this. But we'll do this change in a separate PR

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
imranbohoran: This was the only one that let me quickly verify the migrations were fine. So as much as it looks very primitive, yes.

src/main/java/uk/gov/pay/connector/model/gateway/Auth3dsResponseGatewayRequest.java
imranbohoran: Done

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
imranbohoran: Introduced another variant of anAuthorisationSuccessResponse

src/test/java/uk/gov/pay/connector/service/GatewayClientTest.java
imranbohoran: Agreed. Moved most of it to the `@Before`

src/test/java/uk/gov/pay/connector/service/GatewayClientTest.java
imranbohoran: Instead of the stubbing, InOrder verification is added.

alexbishop1 (approved):
LGTM with the understanding we will try to reduce the Worldpay-specificness in a future PR

PR #350, opened 2017-02-24
PP-1713 Setting the provider session id on the reloaded charge

imranbohoran:


alexbishop1 (approved):
This is the sort of thing that would really benefit from a test and it’s frustrating that the current set-up of CardAuthoriseServiceTest makes this basically impossible to verify.

But the fix looks good.

PR #351, opened 2017-02-27
PP-1713 Refactor to separate out the implementation of session identifier

imranbohoran:
## WHAT

 * Worldpay requires the session identifier as a cookie. However having the cookie
   been set in the GatewayClient makes it very Worldpay specific. Therefore
   pulling the logic on how a session identifier is set to the individual payment
   provider implementations and passing it in.


src/main/java/uk/gov/pay/connector/service/GatewayClient.java
```diff
@@ -39,20 +40,21 @@
...
 
     private GatewayClient(Client client, Map<String, String> gatewayUrlMap, MediaType mediaType,
-                          String sessionIdentifierName, MetricRegistry metricRegistry) {
+                          BiFunction<GatewayOrder, Builder, Builder> sessionIdentifier, MetricRegistry metricRegistry) {
         this.gatewayUrlMap = gatewayUrlMap;
         this.client = client;
         this.metricRegistry = metricRegistry;
         this.mediaType = mediaType;
-        this.sessionIdentifierName = sessionIdentifierName;
+        this.sessionIdentifierName = sessionIdentifier;
```
alexbishop1: This needs renaming because it doesn’t represent a session identifier name any more.

Theoretically, this could do a lot of gateway-specific functionality, not just setting a session identifier, so maybe the name should be more specific. Not quite sure what is should be though … `gatewaySpecificRequestBuilderModifier`?

imranbohoran: As discussed leaving this as sessionIdentifier given that we have introduced the concept of provider sessionID, naming it as a `sessionIdentifier`, I think is fine. 
I probably could do other functionality, but I think other functionality should have their own function.

alexbishop1 (approved):
🆔

PR #352, opened 2017-03-01
PP-1755 Allowing user cancellations when payment in 3DS statuses

imranbohoran:
## WHAT

 * When a payment is aborted or stuck at 3DS Required and the user
   hits the back button we let them start again. When this happens
   we have to cancel the payment so we abide to the API contract with
   our integrators (not returning to them from our platform in a non-terminal)
   state.
 * Allowing the cancellation process to cancel 3DS required payments.


simad:
👍 LGTM

PR #353, opened 2017-03-14
PP-1753 Upgraded project dependencies

ckalista:
## WHAT
- Upgraded the following project dependencies to their latest stable version:
  - `Dropwizard` from version `0.9.3` to `1.0.6`
  - `wiremock` from version `1.57` to `2.5.1`
  - `jpa` from version `2.6.2` to `2.6.4`
  - `guice` from version `4.0` to `4.1.0`
  - `jackson-datatype-jsr310` from version `2.7.2` to `2.8.7`
  - `guava` from version `18.0` to `21.0`
  - `functionaljava-java8` from version `4.4` to `4.6`
  - `commons-validator` from version `1.5.0` to `1.6`
  - `metrics-core` from version `3.1.2` to `3.2.0`
  - `metrics-graphite` from version `3.1.2` to `3.2.0`
  - `commons-csv` from version `1.2` to `1.4`
  - `hamcrest-date` from version `2.0.1` to `2.0.4`
  - `rest-assured` from version `2.4.0` to `2.9.0`
  - `docker-client` from version `3.1.1` to `8.1.1`                        
  - `gson` from version `2.3.1` to `2.8.0`                        

- Upgrading `Dropwizard` to version `1.0.6` pulled down version `2.23.2` of Jersey Client which introduced an undesirable side effect when attempting to explicitly close the response message instance of the HTTP client. As per the documentation, the purpose of the `close` method is to CLOSE the underlying message entity input stream (if available and open) as well as releasing any other resources associated with the response. The operation is supposed to be safe on all instances hence it is being invoked in a finally block in our `GatewayClient` to ensure all resources associated with the response instance are properly cleaned-up and also to prevent potential memory leaks. Since the upgrade though it was discovered that closing an un-consumed entity input stream throws a `javax.ws.rs.ProcessingException` exception `Fatal exception: Error closing message content input stream`. After further investigation this turned out to be a regression issue in Jersey introduced since version `2.22.2` that hasn't been addressed yet unfortunately https://java.net/jira/browse/JERSEY-3214. The problem has also been identified here with a possible fix: https://github.com/jersey/jersey/commit/0d1326dee69d9c693014184bea3e4a44b323daee To work around the problem for the time being and until an official fix has been released, we ensured the response message is consumed entirely before closing.

- Excluded `Jersey` from `wiremock` explicitly to avoid a conflict resulting in the wrong version being pulled.

- The implementation of `Dropwizard Authentication` has undergone a major overhaul since version `0.9.x` of `Dropwizard` introducing a new interface for `Authenticators` and the use of a `Principal`. The code was refactored to conform to the new Auth implementation.

- Rewrote custom `ArgumentMatcher`s asserting a refund entity and a
  refund request.

The SourceClear report still shows a medium risk vulnerabilites in
`jersey-common` and `jersey-media-jaxb` libraries even in the latest
version.

https://app.sourceclear.com/teams/100t11M/scans/1090060

### Note
The current `liquibase` version of the project has been pinned down to `3.3.5` for now. Attempting to upgrade to a more recent version leads to a strange condition whereby the previously applied changesets are not being recognised during the database migration:

```
Validation Error:
      1 change sets have changed since they were ran against the database
           migrations.xml::Seed database table card_types:: was: 7:d296959e27ac0e2400d2472a318394d2 but is now: 7:6293475e6d940c3cc7294331fdcdaf45
 Exception in thread "main" liquibase.exception.ValidationFailedException: Validation Failed:
      1 change sets check sum
           migrations.xml::Seed database table card_types:: was: 7:d296959e27ac0e2400d2472a318394d2 but is now: 7:6293475e6d940c3cc7294331fdcdaf45
```


ghost:

heathd:
I see that liquibase is pinned to `3.3.5` as per 9f47809f8f8b4ef5596 (~1 year ago), did you try updating that?

[update] after discussion with Chris in person

the upgrade was attempted, there appears to be some change in the way that liquibase determines whether migrations are needed. This results in liquibase attempting to re-apply all migrations even in an existing database. Needs further investigation, but don't need to address in this story.

ghost:

ckalista:
Code review comment addressed.

heathd:
Nice work Chris, I've investigated the liquibase issue a little more and raised a Jira ticket for it https://payments-platform.atlassian.net/browse/PP-1801

PR #357, opened 2017-03-21
PP-1828 Altering ChargeStatus arrays to immutable lists

markyjones:
## WHAT
Responding to Findbugs report of returning mutable arrays being a security concern. Also removing merchant code references from our unit test code.

## HOW 
Altering the code to use Immutable lists to remove the issue.


ghost:

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -178,8 +178,8 @@ public boolean isAssociatedTo(Long accountId) {
         return this.getGatewayAccount().getId().equals(accountId);
     }
 
-    public boolean hasStatus(ChargeStatus... status) {
-        return Arrays.stream(status).anyMatch(s -> equalsIgnoreCase(s.getValue(), getStatus()));
+    public boolean hasStatus(List<ChargeStatus> status) {
```
alexbishop1: I actually think this method should continue using varargs, even though that’s an array under the covers.

If this method takes a list, invoking it becomes rather clunky (because the arguments have to be explicitly wrapped into a list). I just don’t think it’s worth it, even though it would gain us a little extra safety.

The main change to make `CardAuthoriseBaseService.getLegalStates()` return an immutable list rather than a mutable array is righteous. It eliminates a situation where an object returns one of its internal data structures and allows callers to modify it.

However, I don’t think we need to go as far as making this method take a list. The array passed in is essentially a one-shot thing that’s used and immediately disposed. In most cases, the method will be invoked using the varargs syntax, so there will never actually be a usable array (in the sense that there is a reference to an actual array object) at the call site.

Sure, it’s possible to conceive of situations where continuing to pass an array could lead to bad things happening but that would require changes to both this method (it would either have to modify the array or maintain a reference to it after it’s called) and the call site (it would have to maintain a reference to the array passed in). And all this would also be possible if this method takes a list because nothing actually guarantees that it’s always going to be an immutable list.

So, I think here (but here only) we should continue using varargs.

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
```diff
@@ -73,9 +75,9 @@ public ChargeEntity preOperation(ChargeEntity chargeEntity) {
     }
 
     @Override
-    protected ChargeStatus[] getLegalStates() {
-        return new ChargeStatus[]{
+    protected List<ChargeStatus> getLegalStates() {
+        return ImmutableList.of(
                 AUTHORISATION_3DS_REQUIRED
-        };
+        );
```
alexbishop1: You could actually do `Collections.singletonList(AUTHORISATION_3DS_REQUIRED)` here (the returned list is immutable) but I understand if you want to keep it explicit and consistent with the other subclasses of `CardAuthoriseBaseService`.

src/main/java/uk/gov/pay/connector/service/CardAuthoriseBaseService.java
```diff
@@ -61,6 +62,6 @@ public GatewayResponse doAuthorise(String chargeId, T gatewayAuthRequest) {
 
     protected abstract GatewayResponse<BaseAuthoriseResponse> operation(ChargeEntity preOperationResponse, T gatewayAuthRequest);
 
-    protected abstract ChargeStatus[] getLegalStates();
+    protected abstract List<ChargeStatus> getLegalStates();
```
alexbishop1: :1st_place_medal:

src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
```diff
@@ -82,9 +82,9 @@ public void setShaPassphrase(String shaPassphrase) {
 
     private static EpdqSignedPayloadDefinitionFactory signedPayloadDefinitionFactory = anEpdqSignedPayloadDefinitionFactory(new EpdqSha512SignatureGenerator());
 
-    public static PayloadBuilder AUTHORISE_ORDER_TEMPLATE_BUILDER = createPayloadBuilderForNewOrder();
-    public static PayloadBuilder CAPTURE_ORDER_TEMPLATE_BUILDER = createPayloadBuilderForMaintenanceOrder();
-    public static PayloadBuilder CANCEL_ORDER_TEMPLATE_BUILDER = createPayloadBuilderForMaintenanceOrder();
+    final public static PayloadBuilder AUTHORISE_ORDER_TEMPLATE_BUILDER = createPayloadBuilderForNewOrder();
```
alexbishop1: Why was this changed? The Java Language Specification recommends (“it  is customary, though not required”) `public static final`.

alexbishop1 (commented):
Overall, this makes the world a better place but I would recommend not going so far as to change `ChargeEntity.hasStatus(ChargeStatus...)` to take a list (see longer comment there for more discussion) and I’m not keen on the modifier order changes.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
alexbishop1: Oh, I see. The `final` wasn’t there before. Sorry, I missed that. Still think it should be `public static final` though. 😀

ghost:

ghost:

ghost:

alexbishop1 (approved):

ghost:

ghost:

PR #358, opened 2017-03-22
PP-1776 Background process to perform charge captures

heathd:
Introduce background process to capture payments.

It will capture any charges in the `CAPTURE_APPROVED` status. 

We won't have any charges in this status at present.

In [PP-1775](https://payments-platform.atlassian.net/browse/PP-1775) we will make connector set a charge to `CAPTURE_APPROVED` status to indicate the user's intent to capture the payment, instead of synchronously performing the capture.

However, this change does introduce the possibility that charges will enter `CAPTURE_APPROVED` state. This would happen if the synchronous capture request times out. We see this very rarely (<5 times in the past 3 months) and if it does happen the process will repeatedly retry the capture until it succeeds. We have already [verified](https://payments-platform.atlassian.net/browse/PP-1485) that repeatedly attempting to capture a payment is safe.

Therefore we believe it is safe to ship this change to production.

## Prompts for reviewer

- the background process is run as a dropwizard managed thread. We've chosen to set a random interval between invocations of the background process as an attempt to avoid clashes between processes running concurrently on multiple nodes. There is locking protection around the capture calls (via writing a locking status to the charge).


ghost:

ghost:

ghost:

ghost:

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -0,0 +1,59 @@
...
+            logger.info("Capturing : "+ chargesToCapture.size() + " charges");
+            metricRegistry.counter("gateway-operations.capture-process.count").inc();
+
+            chargesToCapture
+                .forEach((charge) ->  captureService.doCapture(charge.getExternalId()));
+        } catch (Exception e) {
+            logger.error("Exception when running capture", e);
+        } finally {
+            responseTimeStopwatch.stop();
+            metricRegistry.histogram("gateway-operations.capture-process.running_time").update(responseTimeStopwatch.elapsed(TimeUnit.MILLISECONDS));
```
ckalista: If any execution of the given method throws an exception, the scheduler stops executing the task. May be worth to also safeguarding all code in the finally block from throwing an exception. In general and even though I don't think it will be a problem in a practical situation, we could also safeguard the instanciation of the `Stopwatch`.

ckalista (commented):
LGTM 👍

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
imranbohoran: Moved the exception handling to the Runnable implementation that calls this method. That we feel we are more clear on our intent as to why we are being defensive against unexpected exceptions.

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import static uk.gov.pay.connector.model.domain.ChargeStatus.CAPTURE_APPROVED;
+
+public class CardCaptureProcess {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+    static final long BATCH_SIZE = 10;
```
bazbremner: As discussed in meatspace, I would like to see batch size as something that can be overridden via an environment variable. A sane default is fine, but we don't know how a small number of nodes will get through spikes of work and I'd rather not have to go through a 45 minute build process plus deploys to change a value.

imranbohoran: 👍 This is been addressed as part of the PP-1777 (the retry story).

ghost:

ckalista (approved):

PR #361, opened 2017-03-23
PP-1793 Instrument gateway clients to monitor their connection pool

simad:
## WHAT
Ensure PaymentProviders is a Singleton
- We were assuming `PaymentProviders` to be a singleton, instantiated
  when connector starts up. Instead, guice was automatically
  creating a new istance every time `PaymentProviders` was requested.
  I counted 6+ instances getting created at startup. Not so single
  after all :)
- As this class is responsible for the creation of gateway clients,
  we were therefore creating unnecessary instances of clients. 
  So, we are now explicitly binding `PaymentPrivider` to the singleton class.

Instrument gateway clients
- We are creating a gateway client for each operation. Instead of
  using the default connection manager, we can use the instrumented one
  brought by dropwizard-metrics.
  See
https://github.com/dropwizard/metrics/blob/3.2-development/metrics-httpclient/src/main/java/com/codahale/metrics/httpclient/InstrumentedHttpClientConnectionManager.java
- This instrumented client keeps track of 4 metrics, `available-connections`,
  `leased-connections`, `pending-connections`, `max-connections`.
  As we have a client for each operation, we will have these metrics
  per operation, too. The metrics path will look like:
  `<payment_provider>.<operation>.<metric>`. For example, for the
  worldpay auth client: `worldpay.auth.max-connections`.
- All of these new metrics are gauges.

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/ClientFactory.java
+++ b/src/main/java/uk/gov/pay/connector/service/ClientFactory.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.service;
 
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.httpclient.InstrumentedHttpClientConnectionManager;
 import io.dropwizard.client.JerseyClientBuilder;
 import io.dropwizard.client.JerseyClientConfiguration;
 import io.dropwizard.client.proxy.ProxyConfiguration;
@@ -11,6 +13,7 @@ import org.apache.http.conn.socket.PlainConnectionSocketFactory;
 import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
 import org.apache.http.impl.conn.ManagedHttpClientConnectionFactory;
 import org.apache.http.impl.conn.PoolingHttpClientConnectionManager;
+import org.apache.http.impl.conn.SystemDefaultDnsResolver;
 import org.glassfish.jersey.SslConfigurator;
 import org.glassfish.jersey.apache.connector.ApacheClientProperties;
 import org.glassfish.jersey.apache.connector.ApacheConnectorProvider;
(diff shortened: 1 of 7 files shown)


ghost:

ghost:

simad:
rebuild

ghost:

ghost:

src/main/java/uk/gov/pay/connector/service/ClientFactory.java
```diff
@@ -23,6 +26,9 @@
 import javax.inject.Inject;
 import javax.net.ssl.HostnameVerifier;
 import javax.ws.rs.client.Client;
+import java.util.concurrent.TimeUnit;
+
+import static java.lang.String.*;
```
daniele-occhipinti, line 31: Minor point - isn't it a bit unusual to import `java.lang.String`? I know you do it for conveniency so that you invoke `format` rather than `String.format` twice in your code. However, devs maybe more accustomed to see `String.format` - if they see just `format` they may wonder what kind of formatting that is.
However I don't feel strong about it.

daniele-occhipinti (approved):
Just one minor thing - up to you whether to change it or not.

src/main/java/uk/gov/pay/connector/service/ClientFactory.java
simad: Yes, we usually call the plain `format` with a static import of String - it's such a common method that it's pretty unusual to wonder about where it comes from. Same with the mockito/junit static imports. But agreed, with more obscure libraries it's better to be a bit more verbose.

daniele-occhipinti: OK, makes sense.
I am going to merge.

PR #363, opened 2017-03-24
PP-1777 capture background process retries

heathd:
With @imranbohoran 

In #358 we added the background process and a retrying mechanism, however it would continue to retry captures indefinitely.

This PR adds support for retrying capture attempts in a sensible way. Key changes:

- 6d61fef tuned the timeouts of both capture and auth
- a7ea9f0 set status to error if maximum retries exceeded
- 5e80a32 configuration of maximum retries

## Notes for reviewer

We tried using a `guage` to record the queue length in hosted graphite. However this didn't seem to work. We've changed to a `meter`, however this doesn't seem to be quite the right thing either. We suggest just shipping this change and we can improve the metrics later.

Please review the setting of timeout values (see screenshots of actual timings attached)

screen shot 2017-03-22 at 13 54 42
screen shot 2017-03-22 at 13 55 07
screen shot 2017-03-22 at 13 56 32
screen shot 2017-03-22 at 13 57 26
screen shot 2017-03-22 at 13 57 49
screen shot 2017-03-22 at 14 05 51
screen shot 2017-03-22 at 14 06 59


ghost:

ghost:

src/main/java/uk/gov/pay/connector/app/CaptureProcessConfig.java
```diff
@@ -0,0 +1,26 @@
...
+
+    public int getBatchSize() {
+        return batchSize;
+    }
+
+    public Duration getRetryFailuresEvery() {
+        return retryFailuresEvery;
+    }
+
+    public java.time.Duration getRetryFailuresEveryAsJavaDuration() {
```
simad, line 19: An API for this is to encapsulate the conversion in a small class which returns a dropwizard duration or a java duration, using field methods like `duration.asDropwizard()` or `duration.asJava()`. This is usually a good way to keep it nice and tight, but as this config class is already quite small and self contained, I don't feel so strongly about making this change.

src/test/java/uk/gov/pay/connector/service/CardCaptureProcessTest.java
```diff
@@ -1,31 +1,32 @@
...
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertEquals;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Mockito.*;
-import static uk.gov.pay.connector.model.domain.ChargeStatus.CAPTURE_APPROVED;
 
 @RunWith(MockitoJUnitRunner.class)
 public class CardCaptureProcessTest {
 
+    public static final int MAXIMUM_RERTRIES = 10;
```
simad: Typo in `MAXIMUM_RERTRIES`

src/test/resources/config/client-factory-test-config-with-smartpay-timeout-override.yaml
```diff
@@ -66,6 +66,11 @@ executorServiceConfig:
   timeoutInSeconds: ${AUTH_READ_TIMEOUT_SECONDS:-1}
   threadsPerCpu: ${NUMBER_OF_THREADS_PER_CPU:-100}
 
+captureProcessConfig:
```
simad, line 69: Why are we adding the config here? This file is used only in the client factory tests, so we could avoid editing it for all the other tests. In truth this config duplication is a bit useless, we could raise a jira ticket as tech debt

simad (changes requested):

src/test/resources/config/client-factory-test-config-with-smartpay-timeout-override.yaml
heathd: as far as I could make out, Dropwizard requires this element to be present due to the way it deserializes the YAML using jax-rs. We couldn't figure out how to make an optional config section.

heathd:
@simad rebased and addressed comments

ghost:

simad (approved):

PR #365, opened 2017-03-27
PP-1776 Allow capture approved charges to be captured

heathd:
The `CardCaptureService` has a precondition guard on the status of charges
which are allowed to be captured. This is useful for preventing calls to the
(private) `/frontend/charges/{chargeId}/capture` endpoint from attempting to
capture a charge which should not be captured.

We overlooked this guard in the original work because the
`CardCaptureProcessTest` mocks out the `CardCaptureService`, and there is no
integrating test.


simad (approved):
That's one of the examples when mocking can hide problems. :+1:

ghost:

PR #366, opened 2017-03-28
Pp 1789 correlation id improved logging

daniele-occhipinti:
## WHAT
This PR aims at improving logging for connector by solving 2 problems:

1. not all the logging entries had a requestId
2. even when the requestId was included in the log line, it was not clearly marked as the request id

We solved issue 1) by moving the inclusion of the requestId from the LoggingFilter to the logging configuration (in config/config.yaml) - this makes sure that every time Logger::info() (or the other levels) are called in the code, they are affected by the new format in the configuration. In fact, in the previous implementation the requestId was included only when the LoggingFilter::filter method was invoked.

We solved issue 2) by adding the string `requestID=` just before the request id.

We had to introduce default values in the config.yaml because we are using this configuration for the new tests.

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -13,7 +13,7 @@ logging:
         threshold: ALL
         timeZone: UTC
         target: stdout
-        logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n"
+        logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%thread] %highlight(%-5level) %cyan(%logger{15}) [requestID=%X{X-Request-Id}] - %msg %n"
 
 links:
   frontendUrl: ${FRONTEND_URL}
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -1,10 +1,10 @@
 server:
   applicationConnectors:
     - type: http
-      port: ${PORT}
+      port: ${PORT:-0}
   adminConnectors:
     - type: http
-      port: ${ADMIN_PORT}
+      port: ${ADMIN_PORT:-0}
 
 logging:
     level: INFO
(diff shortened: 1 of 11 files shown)


ghost:

bazbremner:
Please prefix all commit messages with the JIRA ticket number, i.e. `PP-1789 log request ID everywhere`

ghost:

ghost:

simad:
rebuild

src/main/resources/config/config.yaml
```diff
@@ -78,9 +78,9 @@ jerseyClient:
   userAgent: connector
   gzipEnabledForRequests: false
   proxy:
-      host: ${HTTP_PROXY_HOST}
-      port: ${HTTP_PROXY_PORT}
-      scheme : ${HTTP_PROXY_SCHEME}
+      host: ${HTTP_PROXY_HOST:-0}
+      port: ${HTTP_PROXY_PORT:-0}
+      scheme : ${HTTP_PROXY_SCHEME:-https}
```
bazbremner, line 83: This change was lost in your commits - squashing different functional changes can surprise people.

ghost:

ghost:

src/main/resources/config/config.yaml
daniele-occhipinti: I have fixed this by creating 2 separates commits rather than squashing all of them together

src/main/resources/config/config.yaml
```diff
@@ -13,7 +13,7 @@ logging:
         threshold: ALL
         timeZone: UTC
         target: stdout
-        logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%thread] %highlight(%-5level) %cyan(%logger{15}) - %msg %n"
+        logFormat: "[%d{yyyy-MM-dd HH:mm:ss.SSS}] [%thread] %highlight(%-5level) %cyan(%logger{15}) [requestID=%X{X-Request-Id}] - %msg %n"
```
ckalista, line 16: 👍

src/test/java/uk/gov/pay/connector/it/logging/GeneralLoggingTest.java
```diff
@@ -0,0 +1,53 @@
+package uk.gov.pay.connector.it.logging;
```
ckalista, line 1: +1 for testing as well.

ckalista (approved):
I like the global approach that has been introduced here for logging the request id. Much more consistent and less error prone.

PR #367, opened 2017-03-28
[REVIEW ONLY] PP 1775/persist user capture intent

ckalista:
## WHAT

- Allowed the new transition from `AUTHORISATION_SUCCESS` to `CAPTURE_APPROVED`.
- Added a new service method `markChargeAsCaptureApproved` which as its name implies marks the charge by setting its status to `CAPTURE_APPROVED`. The method loads the charge referred to by its external id, throwing an `ChargeNotFoundRuntimeException` if no charge can be found. A pre-condition checks that the status is `AUTHORISATION_SUCCESS` before updating the charge and throwing an `IllegalStateRuntimeException` otherwise.
- Remapped internal status `CAPTURE_APPROVED` status to the appropriate external status of `EXTERNAL_SUCCESS` to revert temporary compatibility fix introduced by [PP-1776](https://payments-platform.atlassian.net/browse/PP-1776).
- Updated the capture endpoint to merely delegate the operation of marking the charge using the service method `markChargeAsCaptureApproved`, returning a 204 [NO-CONTENT], when the operation is successful.


ghost:

PR #369, opened 2017-04-04
PP-1897 Fixed a bug in the metric to track the size of the queue of c…

daniele-occhipinti:
## WHAT

A Meter was used for tracking the size of the queue, but a Meter is to measure the rate of events over time.
We rather need to just log the size of the queue overtime, which is just an integer.
Among the Metrics available in Dropwizard, probably the Gauge is the more appropriate,
but that is not very efficient as that would need to calculate the size of the queue (which changes
very slowly) at regular interval. A CachedGauge is then a better option, but that complicates
things a bit and we still don't have control on when the metric collection is triggered.
That's why I have decided to use a Counter, which still makes sense, but we need to use it
a bit cleverly, as it just provide methods to increment or decrement the counter. Ideally,
to consider in the futre, is to create a new Metric (probably inheriting from Counter) which
would allow us to set a value directly with a set method.

For more information, you can see Dropwizard Metrics guide.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
+++ b/src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
@@ -1,6 +1,6 @@
 package uk.gov.pay.connector.service;
 
-import com.codahale.metrics.Meter;
+import com.codahale.metrics.Counter;
 import com.codahale.metrics.MetricRegistry;
 import com.google.common.base.Stopwatch;
 import io.dropwizard.setup.Environment;
@@ -23,7 +23,7 @@ public class CardCaptureProcess {
     private final MetricRegistry metricRegistry;
     private final CaptureProcessConfig captureConfig;
     private volatile long queueSize;
-    private final Meter captureQueue;
+    private final Counter queueSizeMetric;
 
     @Inject
     public CardCaptureProcess(Environment environment, ChargeDao chargeDao, CardCaptureService cardCaptureService, ConnectorConfiguration connectorConfiguration) {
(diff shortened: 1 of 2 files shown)


ghost:

ghost:

ghost:

ghost:

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -56,14 +54,29 @@ public void runCapture() {
...
         return chargeDao.countCaptureRetriesForCharge(charge.getId()) < captureConfig.getMaximumRetries();
     }
 
+    private void updateQueueSizeMetric(long newQueueSize) {
+        // Unfortunately a Counter does not provide a set method to record a spot value, thus
+        // we need this workaround.
+        // TODO Potentially create a new metric that inherits from Counter and provide a set method
+        long currentQueueSizeCounter = queueSizeMetric.getCount();
+
+        if (newQueueSize > currentQueueSizeCounter) {
```
ckalista: Instead of decreasing, we can increase with a negative value which eliminates the need for a conditional check.

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -56,14 +54,29 @@ public void runCapture() {
             logger.error("Exception when running capture", e);
         } finally {
             responseTimeStopwatch.stop();
+
             metricRegistry.histogram("gateway-operations.capture-process.running_time").update(responseTimeStopwatch.elapsed(TimeUnit.MILLISECONDS));
+            updateQueueSizeMetric(queueSize);
```
ckalista: Not sure this should be in the finally - if anything wrong happens in chargeDao.countChargesForCapture() we don't want to update the metric as we don't know what to update it with.

ckalista (changes requested):

daniele-occhipinti:
Chris, I have pushed both the changes you recommended.

ckalista (approved):

ckalista:
LGTM

ghost:

ghost:

PR #370, opened 2017-04-06
PP-1775 Extracted scheduling parameters out of `CaptureProcessScheduler` into the config file

ckalista:
## WHAT

- Extracted scheduling parameters out of `CaptureProcessScheduler` into the config file to gain the ability to customise the scheduling time interval per environment. It became apparent that it would be useful to run the capture process more frequently in end to end tests to capture charges rapidly and avoid unnecessary delays in the tests relying on the capture process.
- Suppressed logging showing how charges being processed if no charges needs processing in the current batch. This is to avoid littering the logs with useless lines: `INFO  u.g.p.c.s.CardCaptureProcess - Capturing : 0 of 0 charges`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
+++ b/src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
@@ -44,10 +44,12 @@ public class CardCaptureProcess {
 
             List<ChargeEntity> chargesToCapture = chargeDao.findChargesForCapture(captureConfig.getBatchSize(), captureConfig.getRetryFailuresEveryAsJavaDuration());
 
-            logger.info("Capturing : "+ chargesToCapture.size() + " of " + queueSize + " charges");
+            if (chargesToCapture.size() > 0) {
+                logger.info("Capturing : " + chargesToCapture.size() + " of " + queueSize + " charges");
+            }
 
             chargesToCapture.forEach((charge) -> {
-                if(shouldRetry(charge)) {
+                if (shouldRetry(charge)) {
                     captureService.doCapture(charge.getExternalId());
                 } else {
                     captureService.markChargeAsCaptureError(charge);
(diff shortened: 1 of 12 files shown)


ghost:

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -44,10 +44,12 @@ public void runCapture() {
 
             List<ChargeEntity> chargesToCapture = chargeDao.findChargesForCapture(captureConfig.getBatchSize(), captureConfig.getRetryFailuresEveryAsJavaDuration());
 
-            logger.info("Capturing : "+ chargesToCapture.size() + " of " + queueSize + " charges");
+            if (chargesToCapture.size() > 0) {
```
heathd, line 47: hmm, not sure I agree that logging `Capturing : 0 of 0 charges` is useless. It could be useful to help diagnose, e.g. confirming for sure precisely when the polling is happening on each machine etc.

But I can see why you would want to suppress it for testing.

We can revisit this when we play PP-1906

src/test/java/uk/gov/pay/connector/service/CaptureProcessSchedulerTest.java
```diff
@@ -76,9 +80,56 @@ public void shouldScheduleCaptureProcess() {
...
+
+    @Test
     public void shouldShutdownScheduledExecutorServiceWhenStopped() {
+        CaptureProcessScheduler captureProcessScheduler = new CaptureProcessScheduler(null, environment, cardCaptureProcess);
         captureProcessScheduler.stop();
 
         verify(scheduledExecutorService).shutdown();
     }
+
+    private ConnectorConfiguration mockConnectorConfigurationWith(long initialDelayInSeconds, long randomIntervalMinimumInSeconds, long randomIntervalMaximumInSeconds) {
```
heathd, line 124: couldn't we use the real ones? They're just POJO value objects.

heathd (approved):
couple of minor suggestions but not blockers to merging

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
ckalista: If we are after this kind of information for diagnostic purposes we should use the metrics which are still enabled at all times.

src/test/java/uk/gov/pay/connector/service/CaptureProcessSchedulerTest.java
ckalista: yep indeed we should :-)

PR #374, opened 2017-04-20
PP-1948 Feature flag allowing the deployment of the new capture mechanism

ckalista:
## WHAT

- Feature flag allowing the deployment of the new capture mechanism (PP-1775) to production in a better controlled way. The feature flag is turned off by default which means capture remains synchronous until the environment variable `ASYNCHRONOUS_CAPTURE` is set to `true`.

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -54,6 +54,8 @@ executorServiceConfig:
   timeoutInSeconds: ${AUTH_READ_TIMEOUT_SECONDS:-1}
   threadsPerCpu: ${NUMBER_OF_THREADS_PER_CPU:-100}
 
+asynchronousCapture: ${ASYNCHRONOUS_CAPTURE:-false}
+
 captureProcessConfig:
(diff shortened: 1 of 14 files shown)


ghost:

ghost:

src/main/resources/config/config.yaml
```diff
@@ -54,6 +54,8 @@ executorServiceConfig:
   timeoutInSeconds: ${AUTH_READ_TIMEOUT_SECONDS:-1}
   threadsPerCpu: ${NUMBER_OF_THREADS_PER_CPU:-100}
 
+asynchronousCapture: ${ASYNCHRONOUS_CAPTURE:-false}
```
daniele-occhipinti, line 57: Is this the line that guarantee us the async capture will be off by default?

ckalista: Yes, if we don't specify a `ASYNCHRONOUS_CAPTURE` environment variable the default is `false` which means capture operations are carried out in a synchronous fashion.

daniele-occhipinti (approved):

PR #378, opened 2017-04-25
PP 1494 ePDQ auth support

ckalista:
## WHAT

- Built out the class hierarchy for ePDQ authorisation responses, mapping all supported status codes to the following internal statuses:
   - Status `5` to `AUTHORISATION SUCCESS`
   - Status `50`, `51` to `AUTHORISATION SUBMITTED`
   - Status `2` to `AUTHORISATION REJECTED`
   - Any other status to `AUTHORISATION ERROR`
- Introduced a new transient waiting state `AUTHORISATION_SUBMITTED`, which we should not see many occurrences of as authorisation are almost never deferred in real life. The rare instances of non-immediate authorisation (ePDQ status `50` and `51`) are treated as errors, hence `AUTHORISATION_SUBMITTED` is mapped to `EXTERNAL_ERROR_GATEWAY`.
- The state engine has been updated to handle the new `AUTHORISATION_SUBMITTED` status that was introduced to cater for situations when ePDQ defers authorisation which we then treat as an error. The precondition for a charge to transition to `AUTHORISATION_SUBMITTED` is to be in a state of `AUTHORISATION_READY` although it is not mandatory for a charge to transition through `AUTHORISATION_SUBMITTED` in order to proceed to `AUTHORISATION_SUCCESS`, `AUTHORISATION_REJECTED`, `AUTHORISATION_CANCELLED` or `AUTHORISATION_ERROR`. In other words `AUTHORISATION_SUBMITTED` is optional and transient.
 - Transaction that are deferred during authorisation return a 400 bad response with an error message: `This transaction was deferred.`
- The initial implementation of `EpdqPaymentProvider` offers support for authorisation operation only at this stage. When building the authorisation request, the `ORDER ID` submitted to the payment provider in this implementation is going to be the `external charge id` for now until we are provided with more insight by HMCTS as to what sort of reference would be preferable instead in order to aid reconciliation.
- Credentials in `GatewayAccount` has been extended with a SHA passphrase which is required for any request payload signature.
- The configuration has a new section under `epdq` for `urls` and `jerseyClientOverrides` settings. Note that the Jersey Client settings are provisional and subject to change once we have collected real metrics and in a position to set sensible default values.
- New utility class `TestTemplateResourceLoader` used in tests to facilitate loading all request/response payloads that are made available as resources under `src/test/resources/templates`.
- Added argument check ensuring the passphrase is not blank for EPDQ message payment signature.


src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @Override
+    public Either<String, Notifications<Pair<String, Boolean>>> parseNotification(String payload) {
+        return null;
+    }
+
+    @Override
+    public StatusMapper getStatusMapper() {
+        return SmartpayStatusMapper.get();
```
mrlumbu: is this right? At least this could do with a comment to explain

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -0,0 +1,129 @@
...
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static uk.gov.pay.connector.model.domain.ChargeEntityFixture.aValidChargeEntity;
+import static uk.gov.pay.connector.model.domain.GatewayAccountEntity.Type.TEST;
+import static uk.gov.pay.connector.util.AuthUtils.buildAuthCardDetails;
+import static uk.gov.pay.connector.util.SystemUtils.envOrThrow;
+
+@Ignore("Ignoring as this test is failing in Jenkins because it's failing to locate the certificates - PP-1707")
```
mrlumbu, line 44: should this be ignored?

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
```diff
@@ -0,0 +1,77 @@
+package uk.gov.pay.connector.it.resources.epdq;
+
+import org.hamcrest.Matchers;
+import org.junit.Test;
+import uk.gov.pay.connector.it.base.ChargingITestBase;
+
+import static uk.gov.pay.connector.model.domain.ChargeStatus.*;
+
+public class EpdqCardResourceITest extends ChargingITestBase {
```
mrlumbu, line 9: This is a bit unfair as you are following patterns set down across connector codebase, but do you think it would be better to push some of these tests down to PaymentProvider unit test level. It feels wrong to me that there are 5 integration tests and 2 provider unit tests

src/test/java/uk/gov/pay/connector/it/resources/smartpay/SmartpayNotificationResourceWithAccountSpecificAuthITest.java
```diff
@@ -24,6 +25,8 @@
 import static uk.gov.pay.connector.model.domain.ChargeStatus.*;
 import static uk.gov.pay.connector.model.domain.RefundStatus.REFUND_SUBMITTED;
 import static uk.gov.pay.connector.util.JsonEncoder.toJson;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.SMARTPAY_MULTIPLE_NOTIFICATIONS;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.SMARTPAY_NOTIFICATION_CAPTURE;
```
mrlumbu: unused import

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
```diff
@@ -0,0 +1,77 @@
...
+    @Test
+    public void shouldNotAuthorise_whenTransactionIsRefused() throws Exception {
+        epdq.mockAuthorisationFailure();
+
+        String expectedErrorMessage = "This transaction was declined.";
+        String expectedChargeStatus = AUTHORISATION_REJECTED.getValue();
+        shouldReturnErrorForAuthorisationDetailsWithMessage(authorisationDetails, expectedErrorMessage, expectedChargeStatus);
+    }
+
+    @Test
```
mrlumbu, line 42: Just a question really: do we always just pass on the payment provider error message upstream - I would've thought it would make more sense to send up a generic error message?

mrlumbu (commented):
Generally really good. All my comments are just comments rather than things I definitely think need to be changed, but would be good to discuss.

src/main/java/uk/gov/pay/connector/service/BasePaymentProvider.java
```diff
@@ -29,10 +29,17 @@ public BasePaymentProvider(EnumMap<GatewayOperation, GatewayClient> operationCli
     protected <U extends GatewayRequest> GatewayResponse sendReceive(U request, Function<U, GatewayOrder> order,
                                                                      Class<? extends BaseResponse> clazz,
                                                                      Function<GatewayClient.Response, Optional<String>> responseIdentifier) {
+
+        return sendReceive(null, request, order, clazz, responseIdentifier);
```
daniele-occhipinti, line 33: Can we try to rename this? True is is not a public method, but still it is a bit puzzling when you read it.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    private Function<AuthorisationGatewayRequest, GatewayOrder> buildAuthoriseOrderFor() {
+        return request -> anEpdqAuthoriseOrderRequestBuilder()
```
daniele-occhipinti: nice!

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
```diff
@@ -0,0 +1,129 @@
...
+                .withUserId(request.getGatewayAccount().getCredentials().get(CREDENTIALS_USERNAME))
+                .withPassword(request.getGatewayAccount().getCredentials().get(CREDENTIALS_PASSWORD))
+                .withShaPassphrase(request.getGatewayAccount().getCredentials().get(CREDENTIALS_SHA_PASSPHRASE))
+                .withPayId(request.getTransactionId())
+                .build();
+    }
+
+    private Function<CancelGatewayRequest, GatewayOrder> buildCancelOrderFor() {
+        return request -> anEpdqCancelOrderRequestBuilder()
+                .withPspId("gov6nh")
```
daniele-occhipinti: is this some dummy data because we implement the feature?
Are we merging to master with this?

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
ckalista: Agreed, this shouldn't have been there. I have changed it in favour of an exception until we need an implementation most likely when we will look at supporting ePDQ notifications.

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
ckalista: Yes, there test should not be exercised as part the unit test suite by CI. However the code in here is extremely during development to spike new integration features with ePDQ.

ckalista: Yes, these tests should not be exercised as part the unit test suite by CI. However the code in here is useful during development for instance to spike new integration features with ePDQ.

src/test/java/uk/gov/pay/connector/it/resources/smartpay/SmartpayNotificationResourceWithAccountSpecificAuthITest.java
ckalista: Fixed. 👍

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
ckalista: You are absolutely right. The underlying error message from the payment provider is propagated all the way through. This is something I am not happy with either however for now I have just embraced what has been done for the other Gateways.

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
ckalista: Agreed, probably something worth looking but as a separate piece of work as it would require refactoring the tests of other providers.

ghost:

mrlumbu (approved):

daniele-occhipinti (changes requested):

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
ckalista: Well spotted, this shouldn't have been committed in the first place. I have force-pushed removing those two private methods.

src/main/java/uk/gov/pay/connector/service/BasePaymentProvider.java
ckalista: I am not keen to doing that now. I think in general this area require improvement but as it is shared with other gateways it should be a separate piece of work.

daniele-occhipinti: OK, that makes sense - I will create a separate ticket for it and general refactoring.

ghost:

daniele-occhipinti (approved):

PR #386, opened 2017-05-04
PP-1924 fix test clean up

mrlumbu:
For unknown reasons (presumably different test order but not sure) unit
tests started failing on Jenkins. The error was a foreign key constraint violation
on card_types and accepted_card_types during the `deleteAllCardTypes` utility
method, so have modified to clear accepted_card_types before card_types.


alexbishop1 (approved):
Makes sense and we have a green build for the branch

PR #387, opened 2017-05-04
PP-1940 Don’t log payment descriptions because they may contain PII

alexbishop1:
Some services put personally-identifiable information into payment descriptions. In order to avoid this ending up in the logs, exclude the description when logging payment details.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
+++ b/src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
@@ -337,6 +337,7 @@ public class ChargeResponse {
 
     @Override
     public String toString() {
+        // Some services put PII in the description, so don’t include it in the stringification
         return "ChargeResponse{" +
                 "dataLinks=" + dataLinks +
                 ", chargeId='" + chargeId + '\'' +
@@ -345,7 +346,6 @@ public class ChargeResponse {
                 ", cardBrand='" + cardBrand + '\'' +
                 ", gatewayTransactionId='" + gatewayTransactionId + '\'' +
                 ", returnUrl='" + returnUrl + '\'' +
-                ", description='" + description + '\'' +
                 ", reference='" + reference + '\'' +
                 ", providerName='" + providerName + '\'' +
                 ", createdDate=" + createdDate +
--- a/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
@@ -2,8 +2,6 @@ package uk.gov.pay.connector.resources;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Lists;
 import fj.F;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
@@ -15,7 +13,6 @@ import uk.gov.pay.connector.dao.ChargeSearchParams;
 import uk.gov.pay.connector.dao.GatewayAccountDao;
 import uk.gov.pay.connector.model.ChargeResponse;
 import uk.gov.pay.connector.model.domain.ChargeEntity;
-import uk.gov.pay.connector.model.domain.ChargeStatus;
 import uk.gov.pay.connector.service.ChargeExpiryService;
 import uk.gov.pay.connector.service.ChargeService;
 import uk.gov.pay.connector.util.ResponseUtil;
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
```diff
@@ -345,7 +346,6 @@ public String toString() {
                 ", cardBrand='" + cardBrand + '\'' +
                 ", gatewayTransactionId='" + gatewayTransactionId + '\'' +
                 ", returnUrl='" + returnUrl + '\'' +
```
daniele-occhipinti, line 348: Removing the key completely may puzzle people as to why the toString() does not print it - they may think it is a bug somewhere.
What about keep the key and set as its content to something like "[masked for security reasons]".
That's way we avoid a code comment, which is a good practise when we can avoid them by making the code more expressive.
What do you think?

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -234,6 +237,12 @@ private boolean isFieldSizeValid(Map<String, String> chargeRequest, String field
         return !value.isPresent() || value.get().length() <= fieldSize; //already checked that mandatory fields are already there
     }
 
+    private static Map<String, String> sanitizedMap(Map<String, String> map, Set<String> keysToRemove) {
```
daniele-occhipinti: I think this method does not sanitize a Map - it just remove keys from a Map.

I would suggest to rename this to better reflect that (e.g. `removeKeysFromMap`), and to have a method like `removeSensitiveInfoFromCharge`, which will call the `removeKeysFromMap`.

How does that sound?

daniele-occhipinti (changes requested):

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
bazbremner: I'm not a fan of explicitly logging anything overly verbose where we're clearly redacting data.

We're already doing this for card numbers and it feels vaguely ridiculous, as well as being a not-insignificant portion of the message being logged.

daniele-occhipinti: I understand. So what about just using `[removed]` or any minimal string? That way, for sure the description would be shorter than what we have now.
That will also allow us to remove the comment at the top of the block and have it closer to where it belongs, like this:
`"description": "[removed]"  // for security reasons`

alexbishop1: I’m inclined to agree with Baz. While we do it a lot, there’s no requirement for `toString()` to exhaustively include every field: it just needs enough to be useful. Explicitly outputting that we’re not outputting something(!) would clutter the logs when we really want our logs to be as concise as practically possible. If someone gets confused, there’s a comment in the `toString()` method explaining why we’re excluding the description.

bazbremner: ...and on top of the comment @alexbishop1 has introduced,  there's all that fine training every member of the team goes through about secure coding to fall back on :)

daniele-occhipinti: OK, fine - I am convinced. Let's move on :-)

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
alexbishop1: It’s not even actually removing keys from a map: it’s returning a copy of the map with some keys removed! 😃

I take your point though and the method names could be more expressive to better describe both the process and the intent. I’m having a think and I’ll commit a more nuanced proposal.

daniele-occhipinti: Good point :-).
Sounds goods - looking forward to the new proposal.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -150,7 +149,7 @@ public Response createNewCharge(@PathParam(ACCOUNT_ID) Long accountId, Map<Strin
 
         return gatewayAccountDao.findById(accountId).map(
                 gatewayAccountEntity -> {
-                    logger.info("Creating new charge - {}", chargeRequest);
+                    logger.info("Creating new charge - {}", stringifyChargeRequestWithoutPii(chargeRequest));
```
daniele-occhipinti, line 152: Nice solution!

daniele-occhipinti (approved):

PR #392, opened 2017-05-17
PP-2022 upgrading to notify client v3.x

kakumara:
The upgrade needed to be done in order to consistent with adminusers and to be compatible with the latest API from notify. The primary change is that they have moved in favour of a API Key style auth mechanism over the existing serviceId + secret approach.

Currently not happy with the Asynchronous implementation in UserNotificationService.java. This uses a fixed unmanaged threadpool + uses a Future<T> style to make it asynchronous . This could be improved to a more promise style approach (see adminusers notify integration style) and using a Dropwizard managed execution service.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
@@ -11,7 +11,6 @@ import uk.gov.pay.connector.model.builder.EntityBuilder;
 import uk.gov.pay.connector.service.CardExecutorService;
 import uk.gov.pay.connector.service.NotifyClientProvider;
 import uk.gov.pay.connector.service.PaymentProviders;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -11,6 +11,7 @@
 import uk.gov.pay.connector.service.CardExecutorService;
 import uk.gov.pay.connector.service.NotifyClientProvider;
 import uk.gov.pay.connector.service.PaymentProviders;
+import uk.gov.pay.connector.service.UserNotificationService;
```
SandorArpa: Unused import?

SandorArpa (changes requested):
README update

SandorArpa (approved):
LGTM

PR #394, opened 2017-06-01
PP-2120 Removed the logging of the entire notification payload but only relevant notification data

ckalista:
## WHAT

- Removed the line that was responsible for logging all notification payloads verbatim which sometimes included personally identifiable information which was undesirable from a security point of view.
- This has been replaces in favour of logging the parsed notification transpiring in the logs only the following relevant information: `reference`, `transactionId`, `status`, `gatewayEventDate`.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/BaseNotification.java
+++ b/src/main/java/uk/gov/pay/connector/model/BaseNotification.java
@@ -40,10 +40,14 @@ public class BaseNotification<T> implements Notification {
     public String toString() {
         return new StringBuilder()
                 .append("Notification [")
-                .append("reference=" + reference)
-                .append(", transactionId=" + transactionId)
-                .append(", status=" + status)
-                .append(", gatewayEventDate=" + gatewayEventDate)
+                .append("reference=")
+                .append(reference)
+                .append(", transactionId=")
+                .append(transactionId)
+                .append(", status=")
+                .append(status)
+                .append(", gatewayEventDate=")
+                .append(gatewayEventDate)
                 .append("]")
                 .toString();
     }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/model/Notifications.java
```diff
@@ -36,5 +37,8 @@ private Notifications(ImmutableList<Notification<T>> notifications) {
         return notifications;
     }
 
-
+    @Override
+    public String toString() {
+        return String.format("Notifications [notifications=[%s]]", StringUtils.join(get(), ", "));
```
alexbishop1: You can do this just using what’s built in to Java:

```
return String.format("Notifications [notifications=[%s]]", notifications.stream().collect(Collectors.joining(",")));
```

But `ImmutableList`s `toString()` pretty nicely, so you should just be able to do this:

```
return String.format("Notifications [notifications=%s]", notifications);
```

src/main/java/uk/gov/pay/connector/model/BaseNotification.java
```diff
@@ -35,4 +35,16 @@ public T getStatus() {
     public ZonedDateTime getGatewayEventDate() {
         return gatewayEventDate;
     }
+
+    @Override
+    public String toString() {
+        return new StringBuilder()
+                .append("Notification [")
+                .append("reference=" + reference)
```
alexbishop1: Don’t use `+` string concatenation if you already have a `StringBuilder`!

Do `append("reference=").append(reference)` instead.

alexbishop1 (changes requested):
Basically fine but we can be a little smarter with the stringification

src/main/java/uk/gov/pay/connector/model/Notifications.java
ckalista: I am all for conciseness 👍

src/main/java/uk/gov/pay/connector/model/BaseNotification.java
ckalista: Arguably reads a lot less well but fine!

ckalista (commented):
Addressed all review comments.

alexbishop1 (approved):
👍

PR #396, opened 2017-06-13
Pp 1500/epqd notifications status mapper

daniele-occhipinti:
## WHAT

    Added support for deferred payment status changes in response
    to ePDQ notifications. This is because, for this first time,
    we have to make status changes that require information from
    the payment that is not available in the notification itself.
    Thus, handling notifications is no longer always a case of simply
    mapping something in the notification to an internal status.
    We sometimes have to look up the payment to get contextual information
    on what its status should become.
    Specifically, this is for cancellations: an ePDQ
    notification will just say a payment is cancelled but we need to
    look at the payment in the database to know if it is a system
    cancel, user cancel or expiry. This we defer until later, after
    the initial mapping is done, introducing the concept of an
    "deferred" status (one that needs further contextual
    information from the payment to resolve).
    The new method BaseStatusMapper.Builder.mapDeferred() and the
    new class CancelStatusResolver take care of that. 

    We also created the Dropwizard resource (endpoint) for ePDQ notifications.

    We also have introduced a new common interface InternalExternalStatus
    for both ChargeStatus and RefundStatus. Because of that, we can
    be more specific about their type when passed as argument - we were
    using just Enum before whilst now we can use the new interface as the type.
    As a result of that, we removed a test which was not longer applicable: shouldIgnoreNotificationWhenNeitherOfTypeChargeOrRefund

    with @ebiere
    with @alexbishop1


src/main/java/uk/gov/pay/connector/model/domain/InternalExternalStatus.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.pay.connector.model.domain;
+
+
+public interface InternalExternalStatus {
```
mrlumbu, line 4: mmm not sure I understand what an internal external status is. Is there a clearer name? Or could you at least explain in a comment what an InternalExternalStatus is?

src/main/java/uk/gov/pay/connector/service/BaseStatusMapper.java
```diff
@@ -1,6 +1,8 @@
 package uk.gov.pay.connector.service;
```
mrlumbu, line 1: It might just be my tired brain but I'm struggling to wrap my head round this class. I feel like if we have time on our side (to check with @roryjhs) we should try to refactor this into an enum somehow. I have discussed this with @alexbishop1 and he thinks it's doable. My fear is this will start to get out of control in same way NotificationService has.

mrlumbu (changes requested):

ckalista:
Discussed the overall refactoring needed in this PR with @alexbishop1. Closing for now as major rework is underway.

PR #397, opened 2017-06-22
Pp 1500/epqd notifications status mapper

daniele-occhipinti:
## WHAT
Added support for deferred payment status changes in response
to ePDQ notifications. This is because, for this first time,
we have to make status changes that require information from
the payment that is not available in the notification itself.
Thus, handling notifications is no longer always a case of simply
mapping something in the notification to an internal status.
We sometimes have to look up the payment to get contextual information
on what its status should become.
Specifically, this is for cancellations: an ePDQ
notification will just say a payment is cancelled but we need to
look at the payment in the database to know if it is a system
cancel, user cancel or expiry. This we defer until later, after
the initial mapping is done, introducing the concept of an
"deferred" status (one that needs further contextual
information from the payment to resolve).
The new method BaseStatusMapper.Builder.mapDeferred() and the
new class CancelStatusResolver take care of that. 

We also created the Dropwizard resource (endpoint) for ePDQ notifications.

We also have introduced a new common interface InternalExternalStatus
for both ChargeStatus and RefundStatus. Because of that, we can
be more specific about their type when passed as argument - we were
using just Enum before whilst now we can use the new interface as the type.
As a result of that, we removed a test which was not longer applicable: shouldIgnoreNotificationWhenNeitherOfTypeChargeOrRefund

with @ebiere
with @alexbishop1


heathd:
Quick one that we just discussed in person. The name `InternalExternalStatus` is not very clear. We noted that when this code was written the name `Status` was used by something else, but that code has been refactored and now `Status` is unused.

So let's rename `InternalExternalStatus` to `Status`

heathd:
rebuild

daniele-occhipinti:
rebuild

src/main/java/uk/gov/pay/connector/service/epdq/EpdqStatusMapper.java
```diff
@@ -0,0 +1,25 @@
...
+import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.CAPTURED;
+
+public class EpdqStatusMapper {
+
+    private static final BaseStatusMapper<String> STATUS_MAPPER =
+            BaseStatusMapper
+                    .<String>builder()
+                    .map("2", AUTHORISATION_REJECTED)
```
heathd: maybe local constants for these numbers? eg. something like

```
EPDQ_AUTHORISATION_REFUSED = "2"
EPDQ_AUTHORISED = "5"
EPDQ_PAYMENT_REQUESTED = "9"
...
```

src/test/java/uk/gov/pay/connector/service/BaseStatusMapperTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.Optional;
+import org.junit.Test;
+import uk.gov.pay.connector.model.domain.ChargeEntity;
+import uk.gov.pay.connector.model.domain.ChargeStatus;
+
+public class BaseStatusMapperTest {
+
+  @Test
+  public void shouldStatusBeDeferredIfMappedAsDeferred()
```
heathd: this reads like a question to me

should it be deferred? I don't know.. I was hoping you could tell me that :-)

better would be:

`statusShouldBeDeferredIfMappedAsDeferred`

or

`shouldMapStatusAsDeferred_whenMappedAsDeferred`

or 

`mapDeferred_shouldMapStatusAsDeferred`

src/test/java/uk/gov/pay/connector/service/BaseStatusMapperTest.java
```diff
@@ -0,0 +1,46 @@
...
+            .mapDeferred("DEFERRED_STATUS", mockedCancelStatusResolver)
+            .build();
+
+    InterpretedStatus deferredStatus = statusMapper.from("DEFERRED_STATUS");
+
+    assertThat(deferredStatus.isDeferred(), is(true));
+  }
+
+  @Test
+  public void shouldStatusBeIgnoredIfMappedAsIgnore()
```
heathd, line 34: `ignore_shouldMapStatusAsIgnored`?

src/test/java/uk/gov/pay/connector/service/CancelStatusResolverTest.java
```diff
@@ -0,0 +1,84 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class CancelStatusResolverTest {
+
+    CancelStatusResolver cancelStatusResolver;
+
+    @Test
+    public void shouldUserCancelSubmittedStateBeResolvedToUserCancelledState() {
```
heathd, line 26: same here, the test name reads like a question rather than a statement

src/main/java/uk/gov/pay/connector/model/domain/ChargeStatus.java
```diff
@@ -9,7 +9,7 @@
 import static java.util.Arrays.stream;
 import static uk.gov.pay.connector.model.api.ExternalChargeState.*;
 
-public enum ChargeStatus implements InternalExternalStatus {
+public enum ChargeStatus implements Status {
```
heathd: I would consider squashing this commit into d8d4dca3a988506fc506c097624e6eaef5388410 

you'd also need to re-word the commit message to make it congruent

heathd (dismissed):
A few comments about wording especially around tests, but given that this is going to be significantly rewritten by @alexbishop1's refactoring I'd suggest getting this merged and addressing those after the refactoring.

One thing maybe worth doing before merging, however, is squashing the rename of `InternalExternalStatus`->`Status`, as otherwise it's just noise in the commit history

src/test/java/uk/gov/pay/connector/service/BaseStatusMapperTest.java
alexbishop1: So, funny story, turns out there was a `BaseStatusMapperTest` and an older `StatusMapperTest`! The refactoring on the other branch will merge them into one has less quizzical test names.

src/test/java/uk/gov/pay/connector/service/BaseStatusMapperTest.java
alexbishop1: This will become the compendious `shouldIgnore()`.

src/test/java/uk/gov/pay/connector/service/CancelStatusResolverTest.java
alexbishop1: This class will go away altogether.

daniele-occhipinti:
Thanks David for giving those renaming suggestions (and thanks Alex also for pointing out what is going to go).
We are going to do the renaming of what is going to stay and the squashing of those commits.

src/test/java/uk/gov/pay/connector/service/BaseStatusMapperTest.java
daniele-occhipinti: We are going to rename this as it is going to stay, right? Thanks.

ckalista:
This branch was superseded  by PP-1500_notification_handling_refactoring. Closing this pull request.

PR #400, opened 2017-06-27
PP 1500 Support ePDQ notifications

alexbishop1:
PP-1500 Support ePDQ notifications

NotificationService now has four well-defined steps: parse, verify, evaluate,
update:
* Parse: parse payload into one or more notifications.
* Verify: check notification has necessary information and, in the case of
  ePDQ, check the SHA signature is correct.
* Evaluate: decide what to do with the notification (if it maps to a status we
  wish to apply to the payment). Unlike before, this happens after the payment
  is retrieved from the database so can take account of its current status.
* Update: modify the payment in the database.

StatusMapper is expanded beyond doing simple mappings from values in
notifications to a ChargeStatus or RefundStatus. While this is still supported,
it is now also possible to decide what to map to based on the value in the
notification and the current status of the payment. This is supported by new
StatusMapFromStatus objects to represent the two types of keys:
ProviderStatusOnly (key is just the value in the notification) and
ProviderStatusWithCurrentStatus (key is the value in the notification in
combination with the current status of the payment). When doing lookups, the
latter is always used. For this reason, a ProviderStatusOnly and a
ProviderStatusWithCurrentStatus are considered equal if they have the same
provider status (value in the notification).

This allows us to support ePDQ cancellation notifications, where the status
they provide us in the notification is "6" but we need to know the current
status of the payment in order to decide whether it should transition to
USER_CANCELLED, SYSTEM_CANCELLED or EXPIRED.

MappedStatus, UnknownStatus and IgnoredStatus moved out of BaseStatusMapper
into top-level classes. Additionally, MappedStatus split into two:
MappedChargeStatus and MappedRefundStatus. 

ExtendedNotification renamed to EvaluatedNotification, which is an interface
with two implementations: EvaluatedChargeStatusNotification and
EvaluatedRefundStatusNotification (no longer used to represent notifications
that do not map to a payment status modification i.e. notifications that are
ignored or unknown).

NotificationService no longer uses TransactionFlowProvider.

with @daniele-occhipinti
with @ebiere
with @ckalista


src/main/java/uk/gov/pay/connector/service/BaseStatusMapper.java
```diff
@@ -1,119 +1,112 @@
...
+import uk.gov.pay.connector.model.domain.Status;
+import uk.gov.pay.connector.model.domain.RefundStatus;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
-public class BaseStatusMapper<V> implements StatusMapper<V> {
+public class BaseStatusMapper<T> implements StatusMapper<T> {
```
heathd: Could be the moment to rename this class, it's not `Base` since its not a base class for anything. 

these are also a bit hand-wavey but how about `GenericStatusMapper` or `StatusMapperImplementation`?

src/main/java/uk/gov/pay/connector/model/ProviderStatusOnly.java
```diff
@@ -0,0 +1,46 @@
+package uk.gov.pay.connector.model;
+
+import java.util.Objects;
+
+public class ProviderStatusOnly<T> implements StatusMapFromStatus<T> {
```
heathd: `Provider` is a little ambiguous out of context.

`Gateway` or `PaymentGateway` might be clearer

src/test/java/uk/gov/pay/connector/service/epdq/EpdqStatusMapperTest.java
```diff
@@ -0,0 +1,95 @@
...
+import static uk.gov.pay.connector.service.StatusFlow.USER_CANCELLATION_FLOW;
+
+public class EpdqStatusMapperTest {
+
+    @Test
+    public void shouldReturnAuthorisationRejectedStatusFromValue2() throws Exception {
+        InterpretedStatus status = EpdqStatusMapper.get().from("2", AUTHORISATION_SUBMITTED);
+
+        assertThat(status.getType(), is(InterpretedStatus.Type.CHARGE_STATUS));
+        assertThat(((MappedChargeStatus) status).getChargeStatus(), is(AUTHORISATION_REJECTED));
```
heathd: don't think you need the `(MappedChargeStatus)` cast

heathd (commented):
Looks really good.

Great improvements to the clarity of `NotificationService` 🎉 👍 .

It would be good to squash 9d19d5f into d30f197 to avoid polluting the commit history.

I made a few suggestions about naming improvements but don't consider those blockers to merging this.

I wonder whether `NotificationService.Handler` is actually taking on responsibilities which would be better taken on by `PaymentProvider` and its implementations. Again, something to ponder for a future refactoring.

alexbishop1:
Squashed 9d19d5f and df7bd51 into d30f197 (now 3552d33), moving 15ea225 to be after them to facilitate this.

src/main/java/uk/gov/pay/connector/service/BaseStatusMapper.java
alexbishop1: Removed the `StatusMapper` interface (`BaseStatusMapper` was its only implementation) and renamed `BaseStatusMapper` to `StatusMapper`.

src/main/java/uk/gov/pay/connector/model/ProviderStatusOnly.java
alexbishop1: Renamed `ProviderStatusOnly` to `GatewayStatusOnly` and `ProviderStatusWithCurrentStatus` to `GatewayStatusWithCurrentStatus`.

https://www.youtube.com/watch?v=JRjEfTLiS_M

src/test/java/uk/gov/pay/connector/service/epdq/EpdqStatusMapperTest.java
alexbishop1: Removed this cast and similar.

heathd (approved):
Looks good, merge when tests passed

PR #404, opened 2017-07-03
PP-2032: Connector sending database metrics 

whpearson:
This PR makes the DatabaseHealthChecker have another function.

It now also does a query to pg_stats_database and spits them out to
the metrics.

Positives: It doesn't need a separate thread to update the db stats.
Negatives: It only updates the db stats when the healthcheck end point is hit.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
+++ b/src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
@@ -1,19 +1,65 @@
 package uk.gov.pay.connector.healthcheck;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.health.HealthCheck;
+import io.dropwizard.setup.Environment;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 
 import javax.inject.Inject;
 import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
 import java.sql.DriverManager;
+import java.util.HashMap;
+import java.util.Map;
 
 public class DatabaseHealthCheck extends HealthCheck {
 
     private ConnectorConfiguration configuration;
+    private static final Map<String, Long> longDatabaseStatsMap;
+    private static final Map<String, Double> doubleDatabaseStatsMap;
+    private Integer statsHealthy = 0;
+
+    static {
+        longDatabaseStatsMap = new HashMap<String, Long>();
+        longDatabaseStatsMap.put("numbackends", 0l);
+        longDatabaseStatsMap.put("numbackends", 0l);
+        longDatabaseStatsMap.put("xact_commit", 0l);
+        longDatabaseStatsMap.put("xact_rollback", 0l);
+        longDatabaseStatsMap.put("blks_read", 0l);
+        longDatabaseStatsMap.put("blks_hit", 0l);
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
```diff
@@ -1,21 +1,64 @@
...
+import java.sql.ResultSet;
+import java.sql.Statement;
+
 import java.sql.DriverManager;
 
 public class DatabaseHealthCheck extends HealthCheck {
 
     private ConnectorConfiguration configuration;
 
+    private Integer numBackends = 0;
```
kakumara: I would probably opt for a `Map<String, Number>` for this or may be a separate custom class specifically for this (`DatabaseStatusMetric` or something).

src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
```diff
@@ -1,21 +1,64 @@
...
+    private Double  blkWriteTime = 0.0;
+
     @Inject
-    public DatabaseHealthCheck(ConnectorConfiguration configuration) {
+    public DatabaseHealthCheck(ConnectorConfiguration configuration, Environment environment ) {
         this.configuration = configuration;
+        initialiseMetrics(environment.metrics());
     }
+    private void initialiseMetrics(MetricRegistry metricRegistry) {
+        metricRegistry.<Gauge<Integer>>register("connectordb.numbackends", () -> numBackends);
```
kakumara: not sure how the metricRegistry works fully. But does this mean we are only reporting at initialise time or does it get send everytime we `check()` ?

src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
```diff
@@ -25,6 +68,12 @@ protected Result check() throws Exception {
                 configuration.getDataSourceFactory().getUser(),
                 configuration.getDataSourceFactory().getPassword());
             connection.setReadOnly(true);
+            Statement statement = connection.createStatement();
```
kakumara: This particular bit of loading data can be moved to a separate method, perhaps.

src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
```diff
@@ -35,4 +84,19 @@ protected Result check() throws Exception {
...
+        statement.execute("select * from pg_stat_database where datname='connector';");
+        ResultSet resultSet = statement.getResultSet();
+        resultSet.next();
+        for (String key: integerDatabaseStatsMap.keySet() ) {
+            integerDatabaseStatsMap.put(key, resultSet.getInt(key));
+        }
+        for (String key: doubleDatabaseStatsMap.keySet() ) {
+            doubleDatabaseStatsMap.put(key, resultSet.getDouble(key));
+        }
+        resultSet.close();
```
markyjones: Could we just check whether these are managed resources or not and see whether we should be explicitly closing in case of an exception etc.

markyjones (approved):
As per the other changes, looks good. We should consider pulling this out into a common binary at some point if every DB enabled service is going to have this boiler plate.

whpearson:
If we had a place to store libraries I would have done that.

markyjones:
Can probably presume to just have an open source repo and push to a maven central... Not for now.

PR #410, opened 2017-07-22
PP-2245 Abort ignored notification fast

ckalista:
## WHAT

- Notifications of a certain type and for certain providers (i.e. prior authorisation for Worldpay) don't have a transaction id in their payload per design and hence there would be no way to reconcile with the charge that originated the transaction feedback. However since we genuinely don't care about these notifications specifically, they are generally to be ignored gracefully by Connector. This change is to identify notifications that are to be ignored at the early stage of processing during verification phase and abort gracefully and immediately with appropriate logging at INFO level.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/NotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/service/NotificationService.java
@@ -96,6 +96,14 @@ public class NotificationService {
         private <T> boolean verify(Notification<T> notification) {
             logger.info("Verifying {} notification {}", paymentProvider.getPaymentGatewayName(), notification);
 
+            // If the notification is intended to be ignored, we abort immediately to avoid further complications.
+            // Depending on the payment provider and the type of the notification we may not always for instance have a
+            // transaction id, hence if we intend to ignore the notification it is safer to do it here.
+            if (paymentProvider.getStatusMapper().from(notification.getStatus()).getType() == InterpretedStatus.Type.IGNORED) {
+                logger.info("{} notification {} ignored", paymentProvider.getPaymentGatewayName(), notification);
+                return false;
+            }
+
             if (isBlank(notification.getTransactionId())) {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/service/StatusMapper.java
```diff
@@ -79,11 +80,11 @@ private StatusMapper(List<StatusMap<T>> validStatuses) {
         return new Builder<>();
     }
 
-    public InterpretedStatus from(T gatewayStatus, ChargeStatus currentStatus) {
-        GatewayStatusWithCurrentStatus gatewayStatusWithCurrentStatus = GatewayStatusWithCurrentStatus.of(gatewayStatus, currentStatus);
-        Optional<StatusMap<T>> statusMap = validStatuses
+    private InterpretedStatus from(StatusMapFromStatus<T> gatewayStatus, List<StatusMap<T>> wantedValidStatuses) {
```
daniele-occhipinti: I would move this private method at the bottom of the file, so that we have the 2 public from() methods before.
I think that is more standard and make things a bit clearer.

src/main/java/uk/gov/pay/connector/service/NotificationService.java
```diff
@@ -96,6 +96,14 @@ public boolean matchesIpWithDomain(String ipAddress) {
         private <T> boolean verify(Notification<T> notification) {
             logger.info("Verifying {} notification {}", paymentProvider.getPaymentGatewayName(), notification);
 
+            // If the notification is intended to be ignored, we abort immediately to avoid further complications.
```
daniele-occhipinti: As discussed, I would create a new step for this operation.

daniele-occhipinti (changes requested):

daniele-occhipinti (dismissed):

daniele-occhipinti (approved):

PR #413, opened 2017-07-26
PP-2017 remove jstl dependency

kakumara:
This was pulled by liquibase, although they themselves doesn't need it


ckalista (commented):
Does that mean that liquibase have dropped their dependency with jstl?

kakumara:
Nope. 
it was a transitive dependency for liquibase. They are pulling some springframework library (which what they want) and jstl was pulled by that actually. Liquibase doesn't really use it

ckalista (approved):
OK - got it. Makes sense.

PR #414, opened 2017-07-28
PP-1965 Remove async capture feature flag

alexbishop1:
Remove async capture feature flag from configuration etc. and remove checks for it in the code (always use async capture, never sync).

Remove tiny bit of code that only runs if the flag is disabled.

Remove tests that test scenarios where the async capture flag is disabled.

with @juanjoqmelian


whpearson (dismissed):
Looks good

whpearson (approved):
Looks fine.

PR #415, opened 2017-07-31
PP-2410 - Added dynamic adding of ^ to custom paragraph if set

maxcbc:
# PP-2410 - Added dynamic adding of ^ to custom paragraph if set

## WHAT
* If custom paragraph is set for a confirmation email; it is now prefixed with "^ ", else "" is sent

## WHY
* Previously, hardcoding a ^ in our GOV.UK Notify templates prior to an optional paragraph meant that it would be correctly rendered IF the optional paragraph was set, but would simply render ^ if it was NOT set

## NOTES
* This commit necessitates the removal of the hardcoded ^ from our GOV.UK Notify templates


ckalista (approved):
LGTM

PR #417, opened 2017-08-01
PP-2417 Accepting Maestro

rauligs:
## WHAT
-  Add by default 'non 3ds required' card types as *false.* So Created a new column to identify cards requiring _3ds_ to be enabled in order to be used.
- Add 'Maestro' as a new card type.
- Add 'requires_3ds' in card type _Json_ response.
- Handle update gateway account accepted card types 3ds mismatch configuration (409 Conflict).
- Handle _toggle 3ds_ mismatch configuration (having card types with 3ds required), returns 409 Conflict.
- Check 3ds required on Card Auth. So if gateway account has _3ds disabled_ but the card type requires 3ds will throw an exception and set the charge to a final status as `AUTHORISATION ABORTED`.


src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
```diff
@@ -214,4 +215,10 @@ public void setType(Type type) {
     public boolean hasEmailNotificationsEnabled() {
         return emailNotification != null && emailNotification.isEnabled();
     }
+
+    public boolean hasAnyAcceptedCardType3dsRequired() {
+        return cardTypes.stream()
+                .filter(CardTypeEntity::isRequires3ds)
+                .count() > 0;
```
alexbishop1: `return cardTypes.stream().anyMatch(CardTypeEntity::isRequires3ds);`

rauligs: indeed!

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -12,33 +13,55 @@
...
     @Transactional
-    public ChargeEntity preOperation(ChargeEntity chargeEntity) {
-        chargeEntity = preOperation(chargeEntity, OperationType.AUTHORISATION, getLegalStates(), AUTHORISATION_READY);
-        getPaymentProviderFor(chargeEntity).generateTransactionId().ifPresent(chargeEntity::setGatewayTransactionId);
+    public ChargeEntity preOperation(ChargeEntity chargeEntity, AuthCardDetails authCardDetails) {
+
+        String cardBrand = authCardDetails.getCardBrand();
+        ChargeEntity reloadedCharge = chargeDao.merge(chargeEntity);
+
+        if (!reloadedCharge.getGatewayAccount().isRequires3ds() && cardBrandRequires3ds(cardBrand)) {
```
ckalista, line 45: :+1:

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -12,33 +13,55 @@
...
+            chargeEntity = chargeDao.mergeAndNotifyStatusHasChanged(reloadedCharge, Optional.empty());
+        } else {
+            chargeEntity = preOperation(chargeEntity, OperationType.AUTHORISATION, getLegalStates(), AUTHORISATION_READY);
+            getPaymentProviderFor(chargeEntity).generateTransactionId().ifPresent(chargeEntity::setGatewayTransactionId);
+        }
+
         return chargeEntity;
     }
 
+    private boolean cardBrandRequires3ds(String cardBrand) {
```
ckalista, line 58: We had a chat about weather we should also check that the card type is one of the accepted ones on the account and decided that for now it should be sufficient to only check that the account has 3DS on should the card require 3DS.

ckalista (dismissed):
Looks good and ready to go as soon as we get a green build.

ckalista (approved):
Re-approving after rebase.

PR #418, opened 2017-08-01
PP-2422 Send ePDQ refund requests and handle immediate responses

SandorArpa:
Sends `RFD` (partial refund) requests to ePDQ. Expects either status `81` (refund pending) or `71` (deletion pending) responses (usually be 81 but can be 71 if the capture has not yet been processed), which become `REFUND SUBMITTED`. Other responses become `REFUND ERROR`.

Refund references are constructed by concatenating the `PAYID` (gateway transaction ID) with a solidus followed by the `PAYIDSUB` (number that increments with each operation) from the response. The same `PAYID` and `PAYIDSUB` will be included in the subsequent notification, which will allow us to do the lookup by reference.

Pseudo feature flag (hardcoded boolean set to true for tests and false for real) used to stop refunds from working because we haven’t added support for handling the subsequent notifications yet. This means that pressing the refund button in selfservice or using publicapi will continue to display an error as it currently does.

with @juanjoqmelian
with @alexbishop1


SandorArpa:
rebuild

src/main/java/uk/gov/pay/connector/service/epdq/EpdqRefundResponse.java
```diff
@@ -0,0 +1,48 @@
...
+            return super.getErrorMessage();
+        return null;
+    }
+
+    @Override
+    public Optional<String> getReference() {
+        if (isNullOrEmpty(transactionId) && isNullOrEmpty(payIdSub)) {
+            return Optional.empty();
+        }
+        return Optional.of(transactionId + "/" + payIdSub);
```
ckalista: Careful here - if for whatever reason either `payIdSub` OR `transactionId` is empty or null we should probably assume we don't have a valid reference. For instance consider a situation where we are not able to extract the `payIdSub` and hence build a reference which is always the same `trasactionId` + `/` for all refunds for that particular payment, - although very unlikely I am sure it could lead to some strange mismatches.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForMaintenanceOrder.java
```diff
@@ -19,7 +20,12 @@
     public ImmutableList<NameValuePair> extract(EpdqTemplateData templateData) {
 
         // Keep this list in alphabetical order
-        return newParameterBuilder().add(OPERATION_KEY, templateData.getOperationType())
+        EpdqPayloadDefinition.ParameterBuilder parameterBuilder = newParameterBuilder();
+        String amount = templateData.getAmount();
```
ckalista, line 24: Is it mandated by Barclay to have the amount as first parameter? If not maybe better to push it down a bit for readability, possibly after `PSW_KEY`?

ckalista (changes requested):
Looks good. 

Just wondering since we have decided to handle status `71` which according to the comment could only happen if the payment hasn't been captured:
- is it even possible to have such an occurrence considering we only allow refunds post full capture of a charge.
- if it is a valid genuine scenario, what do we do with these refunds?

Also would you be able to add a little bit more context in the commit message. There are a few changes worth documenting in the commit message.

alexbishop1:
> Just wondering since we have decided to handle status 71 which according to the comment could only happen if the payment hasn't been captured:
> • is it even possible to have such an occurrence considering we only allow refunds post full capture of a charge.
> • if it is a valid genuine scenario, what do we do with these refunds?

According to ePDQ, we can actually get a `71` if the payment is in status `9` (“the time scale for the deletion option showing instead of refund can be closer to 24-36 hours, and so it is still possible to delete a transaction in status `9`”). From our perspective, we treat it just like any other refund: whether ePDQ treat it as a true refund or a deletion doesn’t matter to us; we just need to make sure we handle all the possible statuses.

> Also would you be able to add a little bit more context in the commit message. There are a few changes worth documenting in the commit message.

Yeah, was thinking that. Will rewrite the commit message to make it more like the pull request description.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqRefundResponse.java
alexbishop1: Yes, that `&&` should be a `||`.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForMaintenanceOrder.java
alexbishop1: It’s not actually required that the parameters are in alphabetical order (that comment is a bit outdated) but we do have to sort the parameters alphabetically in order to calculate the SHA signature. So I’m inclined to keep them in alphabetical order always.

ckalista: Sure, I don't really have a preference so long as there is some rational let's leave as it is.

alexbishop1:
Addressed comments, rewrote original commit message, added new integration test, changed pseudo feature flag to real feature flag (to facilitate integration test). Can you take another look please, @ckalista?

src/main/java/uk/gov/pay/connector/service/epdq/EpdqRefundResponse.java
alexbishop1: Fixed (by amending original commit).

ckalista (approved):

PR #421, opened 2017-08-09
PP-2436 Make ePDQ payments refundable if in CAPTURE SUBMITTED

alexbishop1:
Change `ExternalChargeRefundAvailability` enum so that it no longer contains the logic to calculate whether a charge is refundable. Instead, this logic is in `ExternalRefundAvailabilityCalculator`, which has per-gateway implementations, allowing for ePDQ to have different rules.

with @juanjoqmelian
with @SandorArpa


kakumara (approved):
Approving.
I probably would have opted for a slightly different design choice, where the calculator to be a function that takes in collection of statuses and outputs the availability thing. 
So that we could construct providers with different flavour of the same function just by inputting a slightly different flavour of `STATUSES_THAT_MAP_TO_EXTERNAL_AVAILABLE_OR_EXTERNAL_FULL`

Optimisation there is that we don't need to come up with different extended classes if we have to extend this for more payment providers.

But for now this works! so ✅

PR #426, opened 2017-08-21
PP-2279 Block PANs in other fields when processing auth requests

alexbishop1:
Make connector reject requests to /v1/frontend/charges/_chargeId_/cards if any fields apart from the card number field contain more than eleven digits (not necessarily contiguous: we want to reject things like “1234-5678 90.12” or even “12345678 Mr. J Smith 9012”)

We don’t actually need to check every field as some of them (e.g expiry date) have strict validation that will filter out PAN-like strings at an earlier stage

with @SandorArpa
with @juanjoqmelian

### DIFF
--- a/src/main/java/uk/gov/pay/connector/resources/AuthCardDetailsValidator.java
+++ b/src/main/java/uk/gov/pay/connector/resources/AuthCardDetailsValidator.java
@@ -3,17 +3,34 @@ package uk.gov.pay.connector.resources;
 import uk.gov.pay.connector.model.domain.Address;
 import uk.gov.pay.connector.model.domain.AuthCardDetails;
 
+import java.util.regex.Pattern;
+
+import static java.util.regex.Pattern.compile;
 import static org.apache.commons.lang3.StringUtils.isNoneBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class AuthCardDetailsValidator {
+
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/resources/AuthCardDetailsValidator.java
```diff
@@ -3,17 +3,34 @@
...
+    private static final Pattern CONTAINS_MORE_THAN_11_NOT_NECESSARILY_CONTIGUOUS_DIGITS = compile(".*([0-9].*){12,}");
+
     public static boolean isWellFormatted(AuthCardDetails authCardDetails) {
         return isValidCardNumberLength(authCardDetails.getCardNo()) &&
                 isBetween3To4Digits(authCardDetails.getCvc()) &&
                 hasExpiryDateFormat(authCardDetails.getEndDate()) &&
                 hasAddress(authCardDetails.getAddress()) &&
-                hasCardBrand(authCardDetails.getCardBrand());
+                hasCardBrand(authCardDetails.getCardBrand()) &&
+                unlikelyToContainACardNumber(authCardDetails.getCardHolder()) &&
```
maxcbc, line 22: I wonder whether it might be worth moving all these `unlikelyToContainCardNumber` checks into a method that just takes `authCardDetails`.

I'm conscious that if we want to add more validators as we go on, this `isWellFormatted` method could end up being a really long list which would potentially make it less readable.

maxcbc (dismissed):
1 little thing, if you disagree happy to approve otherwise.

maxcbc (approved):

PR #427, opened 2017-08-21
PP-2536 Added missing index on charges.gateway_account_id

ckalista:
## WHAT

There is a FK constraint on the `charges.gateway_account_id` referencing the PK of `gateway_account` table. An index on `charges.gateway_account_id` will improve performance on joins between those to tables/columns.


rauligs (approved):
LGTM 👍

PR #428, opened 2017-08-22
PP-2538 Added missing index on refunds.created_date

ckalista:
## WHAT

Underlying query used to produce the transaction list and the csv report orders records by `refunds.created_date`, hence an index is beneficial for performances.


rauligs (dismissed):
LGTM 👍

maxcbc (approved):
L G T M

PR #429, opened 2017-08-22
PP-2279 Block CVC in cardholder when processing auth requests

alexbishop1:
Make connector reject requests to /v1/frontend/charges/_chargeId_/cards if the cardholder name consists of exactly three or four characters (possibly surrounded by whitespace)

Compile other validation regular expressions just once and remove some duplication

* By “CVC”, we mean “CVC/CVV2/CSC/CID/security code/those three digits on the back/except if it’s AmEx when it’s four digits on the front”

with @SandorArpa
with @juanjoqmelian


maxcbc (approved):
- L - G - T - M -

PR #430, opened 2017-08-23
PP-2279 Block CVC with no surrounding whitespace in cardholder

alexbishop1:
Make connector reject requests to /v1/frontend/charges/_chargeId_/cards if the cardholder name consists of exactly three or four digits (with no surrounding whitespace)

The original regex was wrong (used \s+ instead of \s*)

with @juanjoqmelian


SandorArpa (approved):
LGTM 👍

PR #431, opened 2017-08-23
PP 2126 differentiate between authorisation errors

SandorArpa:
## WHAT
Split AUTHORISATION_ERROR different errors into three different categories

## HOW 
- Introducing `AUTHORISATION_TIMEOUT`. Mapping it to its own status instead of `AUTHORISATION_ERROR`.
- Introducing `AUTHORISATION_UNEXPECTED_ERROR`. In case we are unable to map an error to either an error status or a timeout error it's going to be mapped to `AUTHORISATION_UNEXPECTED_ERROR` rather than `AUTHORISATION_ERROR`. Eg http statuscode 500, malformed response, DNS lookup error

@alexbishop1 
@juanjoqmelian


heathd (approved):
Looks good

I struggled to understand the tests in `CardAuthoriseServiceTest` and so I've proposed some changes #432 to try to make them easier to understand.

PR #432, opened 2017-08-29
Refactor card authorise service test for clarity

heathd:
Interested in feedback on this

this is partly a [comprehension refactoring](https://martinfowler.com/articles/workflowsOfRefactoring/#comprehension), but also an attempt to make
the tests more readable/easy to understand by moving towards a more consistent
given/when/then pattern for each test.

In my opinion, this structure makes it easier to understand a unit test by clearly separating out these elements:

1. setting up a stable and repeatable context within which the subject of the
   test will be exercised. This could involve setting up stubs or other test
   doubles, injecting these doubles or otherwise constructing objects which
   comprise the fixed and repeatable environment of the test (aka 'given')
2. explicitly invoking the subject of the test to trigger the tested behaviour (aka 'when')
3. inspecting return values and/or side effects of probing the test subject
   typically by making assertions or verifying mock interactions (aka 'then')

I have deliberately refactored helper methods which contained a combination of 'given' and 'when'.

For example the `authorisationResponse` method both configured the mock payment
provider to return a particular response ('given') and invoked the
`doAuthorise` method of `cardAuthorisationService` ('when'). I've now separated
all of these helper methods so that they only do mock setup (ie. purely
'given') and do not invoke the test subject.

Every test now contains an explicit invocation of the test subject `cardAuthorisationService.doAuthorise()`.

I've also renamed and refactored some of the fixture/mock setup methods.
Instead of `mockXXX` I've followed a pattern of naming `providerWillX`.

## Feedback requested

Do you think the tests are more readable this way?

What do you think of the given/when/then pattern?

### DIFF
--- a/src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
@@ -130,7 +130,6 @@ public class CardAuthoriseServiceTest extends CardServiceTest {
     @Test
     public void shouldRespondAuthorisationSuccess() throws Exception {
         AuthCardDetails authCardDetails = AuthUtils.aValidAuthorisationDetails();
-
         providerWillAuthorise();
 
         GatewayResponse response = cardAuthorisationService.doAuthorise(charge.getExternalId(), authCardDetails);
@@ -142,7 +141,6 @@ public class CardAuthoriseServiceTest extends CardServiceTest {
 
     @Test(expected = ConflictRuntimeException.class)
     public void shouldRespondAuthorisationFailedWhen3dsRequiredConflictingConfigurationOfCardTypeWithGatewayAccount() throws Exception {
-
         AuthCardDetails authCardDetails = AuthUtils.aValidAuthorisationDetails();
 
         GatewayAccountEntity gatewayAccountEntity = new GatewayAccountEntity();
@@ -209,12 +207,9 @@ public class CardAuthoriseServiceTest extends CardServiceTest {
     @Test
     public void shouldRetainGeneratedTransactionIdIfAuthorisationAborted() throws Exception {
         String generatedTransactionId = "generated-transaction-id";
-
         when(mockedProviders.byName(charge.getPaymentGatewayName())).thenReturn(mockedPaymentProvider);
         when(mockedPaymentProvider.generateTransactionId()).thenReturn(Optional.of(generatedTransactionId));
-
         mockExecutorServiceWillReturnCompletedResultWithSupplierReturnValue();
-
         when(mockedPaymentProvider.authorise(any())).thenThrow(RuntimeException.class);
 
         try {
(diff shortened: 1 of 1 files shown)


alexbishop1:
I am very in favour of the ‘given’/‘when’/‘then’ style to improve readability. Especially when blank lines are used between each section. I can’t believe we had tests that had the ‘when’ part buried in an external method!

One annoyance is that the name of the ‘when’ step conflicts with Mockito’s `when` method (which conceptually belongs to ‘given’). Fortunately, they thought of that and gave us `org.mockito.BDDMockito.given`. Just change `when(mockSomething.getSomething()).thenReturn(mockSomethingElse)` to `given(mockSomething.getSomething()).willReturn(mockSomethingElse)`. There’s possibly a wider discussion about if we want to standardise on `given` rather than `when`.

As for the helper methods, while I’m all in favour of sensible separation and clear naming, my preference would be to try to avoid having helper methods at all if possible. It makes it hard to look at a test and figure out at a glance what it’s doing without looking elsewhere. We have a lot of test fixture stuff that needs to be done for each test (possibly a sign that the class under test is trying to do too much) but avoiding helper methods doesn’t mean we have to have two dozen lines of setup in each test if we’re smart. Take, for example, the following test:

```
@Test
public void shouldRespondAuthorisationSuccess() throws Exception {
    AuthCardDetails authCardDetails = AuthUtils.aValidAuthorisationDetails();
    providerWillAuthorise();

    GatewayResponse response = cardAuthorisationService.doAuthorise(charge.getExternalId(), authCardDetails);

    assertThat(response.isSuccessful(), is(true));
    assertThat(reloadedCharge.getStatus(), is(AUTHORISATION_SUCCESS.toString()));
    assertThat(reloadedCharge.getGatewayTransactionId(), is(providerTransactionId));
}    
```

The `providerWillAuthorise()` method actually calls three other methods. It looks deceptively simple but there’s quite a lot going on behind the scenes.

First, it calls the `mockExecutorServiceWillReturnCompletedResultWithSupplierReturnValue()` method (which is all the way at the other end of the file).

Then it calls the `mockAuthResponse(String providerTransactionId, AuthoriseStatus authoriseStatus, String errorCode)` method. This mocks a `WorldpayOrderStatusResponse` and stubs it to return an appropriate transaction ID, status and error code. We could instead have `@Mock` fields like `mockAuthorisationSuccessWorldpayOrderStatusResponse` and `mockAuthorisationRejectedWorldpayOrderStatusResponse` where each one is stubbed in the `@Before` method to return the appropriate things.

Finally, the `providerWillRespondToAuthoriseWith(GatewayResponse value)` method is called, which is the only one that actually stubs what the payment provider returns. This contains the `when(mockedPaymentProvider.authorise(any())).thenReturn(value);` line that looks like it does the real work. Its other two lines — `when(mockedProviders.byName(charge.getPaymentGatewayName())).thenReturn(mockedPaymentProvider);` and `when(mockedPaymentProvider.generateTransactionId()).thenReturn(Optional.empty());` — look like they’re common enough that they could just be lifted into the `@Before` method.

Leveraging changes like this, I see no real reason why we couldn’t make the test as simple as this:

```
@Test
public void shouldRespondAuthorisationSuccess() throws Exception {
    given(mockedPaymentProvider.authorise(any())).willReturn(mockAuthorisationSuccessWorldpayOrderStatusResponse);

    GatewayResponse response = cardAuthorisationService.doAuthorise(charge.getExternalId(), AuthUtils.aValidAuthorisationDetails());

    assertThat(response.isSuccessful(), is(true));
    assertThat(reloadedCharge.getStatus(), is(AUTHORISATION_SUCCESS.toString()));
    assertThat(reloadedCharge.getGatewayTransactionId(), is(providerTransactionId));
}
```

This would be nice because it makes it obvious that the standout thing that makes this test different to the others is that the payment provider returns an authorisation success response. There’s no need to drill down through the `providerWillAuthorise()` method and its dependencies to figure out what’s actually changing (this is another reason I’m generally cynical about helper methods: they encourage us to be a little bit lazy and repeat lots of boilerplate in each test without putting it in the designated boilerplate location i.e. the `@Before` method and making the mocks and other collaborators reusable).

heathd:
Thanks for the detailed feedback Alex, that's really helpful.

I agree as a general rule that individual tests should be self-contained so that they can be more easily understood without needing to dig through lots of helper methods.

Let's try to get this merged as a step an the right direction and look at what further improvements can be made.

alexbishop1 (approved):
Moves us in the right direction 💯

PR #433, opened 2017-08-30
PP-2398 Add migrations script for refunds history

rauligs:
## WHAT
- Add migrations scripts for refunds history (preparation for next PR). This is needed so we can make end to end tests backward compatible.


ebiere (approved):
LGTM

PR #434, opened 2017-08-30
PP-2398 Fix refunds events history

rauligs:
## WHAT
- Refunds with different refund references and same amount and state were being missing in the transaction events list. This was due to the fact that references for these refunds weren't being compared in `equals` method for transaction events.


ckalista (dismissed):
Yes, makes sense 👍

ebiere (approved):
LGTM

PR #436, opened 2017-08-31
PP 2540 refunds history feature flag only

ebiere:
## WHAT
This routes the Get Charges API call to the backend DB depending on the request header settings
## HOW 
Make a call to the API with the header flag set and the response should be different (when the DAO is finally done)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
@@ -16,6 +16,7 @@ import uk.gov.pay.connector.model.domain.ChargeEntity;
 import uk.gov.pay.connector.service.ChargeExpiryService;
 import uk.gov.pay.connector.service.ChargeService;
 import uk.gov.pay.connector.util.ResponseUtil;
+import com.google.common.base.Splitter;
 
 import javax.inject.Inject;
 import javax.ws.rs.*;
@@ -38,6 +39,10 @@ import static uk.gov.pay.connector.util.ResponseUtil.*;
 
 @Path("/")
 public class ChargesApiResource {
+    public static final String FEATURES_HEADER = "GOVUK-Pay-Features";
+    public static final String FEATURE_REFUNDS_IN_TX_LIST = "REFUNDS_IN_TX_LIST";
+    private static final Splitter FEATURES_STRING_SPLITTER = Splitter.on(',').trimResults().omitEmptyStrings();
+
     static final String AMOUNT_KEY = "amount";
     private static final String DESCRIPTION_KEY = "description";
     private static final String RETURN_URL_KEY = "return_url";
@@ -60,6 +65,7 @@ public class ChargesApiResource {
     private static final String ACCOUNT_ID = "accountId";
     private static final String PAGE = "page";
     private static final String DISPLAY_SIZE = "display_size";
+
 
     private static final Set<String> CHARGE_REQUEST_KEYS_THAT_MAY_HAVE_PII = Collections.singleton("description");
 
@@ -106,17 +112,19 @@ public class ChargesApiResource {
                                    @QueryParam(TO_DATE_KEY) String toDate,
                                    @QueryParam(PAGE) Long pageNumber,
                                    @QueryParam(DISPLAY_SIZE) Long displaySize,
-                                   @Context UriInfo uriInfo) {
+                                   @Context UriInfo uriInfo, @HeaderParam(FEATURES_HEADER) String features) {
 
         List<Pair<String, String>> inputDatePairMap = ImmutableList.of(Pair.of(FROM_DATE_KEY, fromDate), Pair.of(TO_DATE_KEY, toDate));
         List<Pair<String, Long>> nonNegativePairMap = ImmutableList.of(Pair.of(PAGE, pageNumber), Pair.of(DISPLAY_SIZE, displaySize));
+        List<String> userFeatures = Optional.ofNullable(features).map(FEATURES_STRING_SPLITTER::splitToList).orElseGet(Collections::emptyList);
+
 
         return ApiValidators
                 .validateQueryParams(inputDatePairMap, nonNegativePairMap) //TODO - improvement, get the entire searchparam object into the validateQueryParams
                 .map(ResponseUtil::badRequestResponse)
                 .orElseGet(() -> reduce(validateGatewayAccountReference(gatewayAccountDao, accountId)
                         .bimap(handleError,
-                                listCharges(new ChargeSearchParams()
+                                listTransactions(new ChargeSearchParams()
                                         .withGatewayAccountId(accountId)
                                         .withEmailLike(email)
                                         .withReferenceLike(reference)
@@ -125,7 +133,7 @@ public class ChargesApiResource {
                                         .withFromDate(parseDate(fromDate))
                                         .withToDate(parseDate(toDate))
                                         .withDisplaySize(displaySize != null ? displaySize : configuration.getTransactionsPaginationConfig().getDisplayPageSize())
-                                        .withPage(pageNumber != null ? pageNumber : 1), uriInfo)))); // always the first page if its missing
+                                        .withPage(pageNumber != null ? pageNumber : 1), uriInfo, userFeatures)))); // always the first page if its missing
     }
 
     @POST
@@ -194,6 +202,14 @@ public class ChargesApiResource {
                 : Optional.of(missing);
     }
 
+    private F<Boolean, Response> listTransactions(ChargeSearchParams searchParams, UriInfo uriInfo, List<String> userFeatures) {
+        if(userFeatures.contains(FEATURE_REFUNDS_IN_TX_LIST)){
+            return listChargesAndRefunds(searchParams, uriInfo);
+        }
+
+        return listCharges(searchParams, uriInfo);
+    }
+
     private F<Boolean, Response> listCharges(ChargeSearchParams searchParams, UriInfo uriInfo) {
         Long totalCount = chargeDao.getTotalFor(searchParams);
         Long size = searchParams.getDisplaySize();
@@ -215,6 +231,11 @@ public class ChargesApiResource {
                         .withChargeResponses(chargesResponse)
                         .withTotalCount(totalCount)
                         .buildResponse();
+    }
+
+    //this needs to hook into the new DAOs/Services
+    private F<Boolean,Response> listChargesAndRefunds(ChargeSearchParams searchParams, UriInfo uriInfo) {
+        return listCharges(searchParams, uriInfo);
     }
 
     private Optional<List<String>> checkInvalidSizeFields(Map<String, String> inputData) {
--- a/src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
@@ -4,6 +4,8 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.jayway.restassured.response.ValidatableResponse;
 import org.apache.commons.lang.math.RandomUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.junit.Ignore;
 import org.junit.Test;
 import uk.gov.pay.connector.it.base.ChargingITestBase;
 import uk.gov.pay.connector.it.dao.DatabaseFixtures;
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -38,6 +38,8 @@
 
 @Path("/")
 public class ChargesApiResource {
+    public static final String FEATURES_HEADER = "features";
```
alexbishop1: Do we pseudo-namespace our header names in any way (e.g. give them names like `GOVUK-Pay-Features` or whatever)? Might be good to follow the convention, if we have one. I’m worried `features` is generic enough that it might be used as the official name of an HTTP header in the future.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -125,7 +128,7 @@ public Response getChargesJson(@PathParam(ACCOUNT_ID) Long accountId,
                                         .withFromDate(parseDate(fromDate))
                                         .withToDate(parseDate(toDate))
                                         .withDisplaySize(displaySize != null ? displaySize : configuration.getTransactionsPaginationConfig().getDisplayPageSize())
-                                        .withPage(pageNumber != null ? pageNumber : 1), uriInfo)))); // always the first page if its missing
+                                        .withPage(pageNumber != null ? pageNumber : 1), uriInfo, Optional.ofNullable(features))))); // always the first page if its missing
```
alexbishop1: The `features` string is a comma-separated list, right? Rather than passing it all the way down, I think it might be desirable to tokenise it into a (possibly empty) list of strings here and pass that down.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -217,6 +224,38 @@ private ZonedDateTime parseDate(String date) {
...
+        List<ChargeResponse> chargesResponse =
+                charges.stream()
+                        .map(charge -> chargeService.buildChargeResponse(uriInfo, charge)
+                        ).collect(Collectors.toList());
+
+        return success ->
+                new ChargesPaginationResponseBuilder(searchParams, uriInfo)
+                        .withChargeResponses(chargesResponse)
+                        .withTotalCount(totalCount)
+                        .buildResponse();
```
alexbishop1: I understand that this duplication is temporary these twenty lines are literally identical to the last part of the `listCharges` method. Could you remove the duplication by moving the common bits in a private method?

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -217,6 +224,38 @@ private ZonedDateTime parseDate(String date) {
...
+    }
+
+    private boolean isRefundsInResultListFeatureEnabledForRequest(Optional<String> features) {
+        if(features.isPresent() && features.get().contains(FEATURE_REFUNDS_IN_TX_LIST)){
+            return true;
+        }
+
+        return false;
```
alexbishop1: If you’re going to take an `Optional<String>` (and the opinion of the designers of Java is that methods should not take `Optional` parameters but that’s a whole other discussion!), a better way to do this is:

```
return features.map(f -> f.contains(FEATURE_REFUNDS_IN_TX_LIST)).orElse(false);
```

However, doing a substring search in a comma-separated list is a bit dangerous, so I would recommend doing the tokenisation early as stated in a previous comment and having this method receive a `List<String>`.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -443,6 +447,66 @@ public void shouldGetChargeLegacyTransactions() throws Exception {
                 .body("results[0].card_details.expiry_date", nullValue());
     }
 
+    //ignore till new DAO (from PP-2515) is ready
+    //@Test
```
alexbishop1: Probably preferable to leave `@Test` uncommented and add `@Ignore` so the test still appears in lists etc. even if it isn’t run

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -443,6 +447,66 @@ public void shouldGetChargeLegacyTransactions() throws Exception {
...
+    public void shouldFilterTransactionsBasedOnRefundsHeaderAndFromAndToDates() throws Exception {
```
alexbishop1: `shouldInterleaveChargesAndRefundsIfFeatureEnabled` perhaps?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -443,6 +447,66 @@ public void shouldGetChargeLegacyTransactions() throws Exception {
...
+
+        addChargeAndCardDetails(CREATED, "ref-1", now());
+        addChargeAndCardDetails(AUTHORISATION_READY, "ref-2", now());
+        String externalChargeId = addChargeAndCardDetails(CAPTURED, "ref-3", now().minusDays(2));
+
+        Long chargeId = Long.parseLong(StringUtils.removeStart(externalChargeId, "charge"));
+        String externalRefundId = createNewRefundWith(REFUND_SUBMITTED, 10L, chargeId, "refund_ref");
```
alexbishop1: The `externalRefundId` variable is never used

alexbishop1: Using the `createNewRefundWith` has the unfortunate effect of hiding the time the refund was made

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -443,6 +447,66 @@ public void shouldGetChargeLegacyTransactions() throws Exception {
...
+                .body("results[0].refund_summary.amount_submitted", is(0))
+                .body("results[0].refund_summary.amount_available", isNumber(AMOUNT))
+                .body("results[0].refund_summary.status", is("pending"))
+                .body("results[1].settlement_summary.capture_submit_time", nullValue())
+                .body("results[1].settlement_summary.captured_time", nullValue())
+                .body("results[1].refund_summary.amount_submitted", is(0))
+                .body("results[1].refund_summary.amount_available", isNumber(AMOUNT))
+                .body("results[1].transaction_type", is("charge"))
+                .body("results[1].refund_summary.status", is("pending"))
+                .body("results[2].transaction_type", is("refund")
```
alexbishop1: I think what’s being tested here gets lost amongst all the assertions. This can probably be pared down to just distinguihsing between charges and refunds, so it’s little more than:
```
.body("results[0].transaction_type", is("charge"))
.body("results[1].transaction_type", is("charge"))
.body("results[2].transaction_type", is("refund"))
```

It might be a bit more thorough to have the refund be the middle event chronologically (to ensure the code under test isn’t just being lazy and putting all the refunds at the end).

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -443,6 +447,66 @@ public void shouldGetChargeLegacyTransactions() throws Exception {
...
+                );
+
+
+
+        List<Map<String, Object>> results = response.extract().body().jsonPath().getList("results");
+        List<String> references = collect(results, "reference");
+        assertThat(references, containsInAnyOrder("ref-1", "ref-2"));
+        assertThat(references, not(contains("ref-3")));
```
alexbishop1: We’re not testing that a refund that happened outside the time window isn’t included, which would be desirable.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -443,6 +447,66 @@ public void shouldGetChargeLegacyTransactions() throws Exception {
...
+        assertThat(statuses, containsInAnyOrder("created", "started"));
+        assertThat(statuses, not(contains("confirmed")));
+
+        List<String> createdDateStrings = collect(results, "created_date");
+        datesFrom(createdDateStrings).forEach(createdDate ->
+                assertThat(createdDate, is(within(1, DAYS, now())))
+        );
+        List<String> emails = collect(results, "email");
+        assertThat(emails, contains(email, email));
+    }
```
alexbishop1: I think this test is trying to cover too much. I think it might be clearer if we had several tests: one to check charges and refunds appear together in the list in order, one to check that refunds have all the necessary properties, one to check that refunds that are outside of the time window aren’t included, etc.

alexbishop1 (changes requested):

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
alexbishop1: Actually, considering this is an integration test, maybe we don’t need several tests. But I think this one could be made more concise and understandable.

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -38,6 +38,9 @@
 
 @Path("/")
 public class ChargesApiResource {
+    public static final String FEATURES_HEADER = "GOVUK-Pay-Features";
+    public static final String FEATURE_REFUNDS_IN_TX_LIST = "REFUNDS_IN_TX_LIST";
+    public static final String FEATURES_DELIM = ",";
```
alexbishop1: Don’t think you really need a constant for this

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -106,10 +110,11 @@ public Response getChargesJson(@PathParam(ACCOUNT_ID) Long accountId,
                                    @QueryParam(TO_DATE_KEY) String toDate,
                                    @QueryParam(PAGE) Long pageNumber,
                                    @QueryParam(DISPLAY_SIZE) Long displaySize,
-                                   @Context UriInfo uriInfo) {
+                                   @Context UriInfo uriInfo, @HeaderParam(FEATURES_HEADER) String features) {
 
         List<Pair<String, String>> inputDatePairMap = ImmutableList.of(Pair.of(FROM_DATE_KEY, fromDate), Pair.of(TO_DATE_KEY, toDate));
         List<Pair<String, Long>> nonNegativePairMap = ImmutableList.of(Pair.of(PAGE, pageNumber), Pair.of(DISPLAY_SIZE, displaySize));
+        List<String> userFeatures = Optional.ofNullable(features).isPresent() ? Arrays.asList(features.split(FEATURES_DELIM)) : Collections.EMPTY_LIST;
```
alexbishop1: Don’t think wrapping this in an `Optional` is gaining us much here. And `Collections.emptyList()` is generally preferable to `Collections.EMPTY_LIST` because it returns a generic list. Also, we have Guava, so you can use Guava’s `Splitter`, which can do nice things like trim leading and trailing whitespace on results and omit emit results (not sure how much of a concern this is in practice); it also may have better performance than the regex-based `String.split`.

I would probably do this by having a Guava splitter in a constant:

```
private static final Splitter FEATURES_STRING_SPLITTER = Splitter.on(',').trimResults().omitEmptyStrings();
```

… then use it in conjunction with a null check:

```
List<String> userFeatures = features != null ? FEATURES_STRING_SPLITTER.splitToList(features) : Collections.emptyList();
```

If you’re really keen on using an `Optional` in preference to a null check, you can do it like this:

```
List<String> userFeatures = Optional.ofNullable(features).map(FEATURES_STRING_SPLITTER::splitToList).orElseGet(() -> Collections.emptyList());
```

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -194,7 +199,15 @@ private ZonedDateTime parseDate(String date) {
...
-    private F<Boolean, Response> listCharges(ChargeSearchParams searchParams, UriInfo uriInfo) {
+    private F<Boolean, Response> listCharges(ChargeSearchParams searchParams, UriInfo uriInfo, List<String> userFeatures) {
+        if(userFeatures.contains(FEATURE_REFUNDS_IN_TX_LIST)){
+            return listChargesAndRefunds(searchParams, uriInfo);
+        }
+
+        return getBooleanResponseF(searchParams, uriInfo);
+    }
+
+    private F<Boolean, Response> getBooleanResponseF(ChargeSearchParams searchParams, UriInfo uriInfo) {
```
alexbishop1: Can we give this a nicer name? Perhaps you could rename the existing `listCharges` method to `listTransactions` (since it produces the transactions list and we seem to be coalescing on the idea that “transactions” refers to both charges and refunds). That would then free up the name `listCharges` for this method, which strictly returns just charges.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -443,6 +448,33 @@ public void shouldGetChargeLegacyTransactions() throws Exception {
...
+                .withQueryParam("to_date", DateTimeUtils.toUTCDateTimeString(now().plusDays(1)))
+                .withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON)
+                .withHeader(FEATURES_HEADER, FEATURE_REFUNDS_IN_TX_LIST)
+                .getTransactions()
+                .statusCode(OK.getStatusCode())
+                .contentType(JSON)
+                .body("results.size()", is(3))
+                .body("results[0].transaction_type", is("charge"))
+                .body("results[1].transaction_type", is("charge"))
+                .body("results[2].transaction_type", is("refund")
```
alexbishop1: This is much easier to understand!

I still think the tests we have for this should be more comprehensive in terms of properly checking the properties on refunds, that refunds and charges sort by date and not by type and that refunds outside the date range are not included. However, it’s fine if these aren’t implemented until the new DAO is actually integrated.

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
alexbishop1: Actually, the `Optional` version could be made even shorter:

```
List<String> userFeatures = Optional.ofNullable(features).map(FEATURES_STRING_SPLITTER::splitToList).orElseGet(Collections::emptyList);
```

alexbishop1 (approved):

PR #439, opened 2017-09-06
PP-2515 Get payments including refunds

rauligs:
## WHAT
- Introduced mechanism to switch between fetching transactions:
    - `TransactionSearchStrategy` fetches both charges and refunds
    - `ChargeSearchStrategy` fetches charges only, using the original implementation
    - The strategy is selected according to the `REFUNDS_IN_TX_LIST` feature flag.

- The strategy and switching via feature flag are temporary. We intend to replace the `TransactionSearchStrategy` implementation after refactoring the data model to store charges and refunds in the same database table.

- The implementation of `TransactionDao` is meant to be used a as drop-in replacement for `ChargeDao` when we want to take refunds into consideration while searching.

- A transaction can be of two types `charge` or `refund`. Depending on whether we want to search for only charges or refunds, an optional payment type can be specified in the filtering criteria. The absence of a payment type in the search criteria implies all types are considered during the search.

- The filtering criteria has been augmented to allow searching by refund status or charge status or any combination of both.


kakumara:
I would questions the selection of `jooq` for its value that is brining. Doesn't seem to make it any more readable or simpler. In fact the code looks a bit more cluttered than to how it would have been with native queries.

However, given the temporary nature of this Dao and the possibility of this being removed later on I'll probably be okay to use it for now.

src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
```diff
@@ -0,0 +1,241 @@
...
+
+        return (long) result.getSingleResult();
+    }
+
+    private SelectOrderByStep buildQueryFor(Long gatewayAccountId, QueryType queryType, ChargeSearchParams params) {
+        Condition queryFilters = field("c.gateway_account_id").eq(gatewayAccountId);
+
+        if (isNotBlank(params.getEmail())) {
+            queryFilters = queryFilters.and(
+                    field("c.email").lower().like('%' + escape(params.getEmail().toLowerCase()) + '%'));
```
kakumara, line 106: Why did you have to do this escaping? i believe jooq support bind params

kakumara (approved):
Bit of a YOLO approval, to get moving. Too many changes once again in the PR. Overall looks okay, apart from the comments added

PR #440, opened 2017-09-11
Clarify escaping of like clause

heathd:
having a function called `escape` immediately sets off alarm bells when
building SQL code. Why are we needing to explicitly escape a string? In fact we
are quoting the special symbols `%` and `_` which can appear in a like clause.


rauligs (approved):
LGTM 👍  
It seems it got lost after some refactoring?

PR #441, opened 2017-09-11
PP-2576 Add refund created in sandbox

maxcbc:
# PP-2576 Add refund created in sandbox
with @SandorArpa

## WHY
Refund submitted isn't showing in the transaction events (in Sandbox)

## WHAT
* Made sandbox refund lifecycle mirror gateway refund lifecycle
i.e. CREATED -> REFUND SUBMITTED -> REFUNDED

## HOW
* Added additional step when creating a refund if creation is successful and using a sandbox account where the refund is automatically moved to a REFUNDED status

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/ChargeRefundService.java
+++ b/src/main/java/uk/gov/pay/connector/service/ChargeRefundService.java
@@ -66,11 +66,27 @@ public class ChargeRefundService {
     }
 
     public Optional<Response> doRefund(Long accountId, String chargeId, RefundRequest refundRequest) {
+
         return Optional.ofNullable(transactionFlowProvider.get()
                 .executeNext(prepareForRefund(providers, accountId, chargeId, refundRequest))
                 .executeNext(doGatewayRefund(providers))
-                .executeNext(finishRefund())
+                .executeNext(setAsSubmitted())
+                .executeNext(setSandboxAsRefunded())
                 .complete().get(Response.class));
+    }
+
+    private TransactionalOperation<TransactionContext, Response> setSandboxAsRefunded() {
+        return context -> {
+            Response response = context.get(Response.class);
+            RefundEntity refund = response.getRefundEntity();
+            if (refund.getChargeEntity().getPaymentGatewayName() == PaymentGatewayName.SANDBOX
+                    && refund.hasStatus(RefundStatus.REFUND_SUBMITTED)) {
+                RefundEntity refundEntity = refundDao.findById(refund.getId()).get();
+                refundEntity.setStatus(RefundStatus.REFUNDED);
+                response = new Response(response.getRefundGatewayResponse(), refundEntity);
+            }
+            return response;
+        };
     }
 
     private PreTransactionalOperation<TransactionContext, RefundEntity> prepareForRefund(PaymentProviders providers, Long accountId, String chargeId, RefundRequest refundRequest) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/service/ChargeRefundService.java
```diff
@@ -67,7 +67,16 @@ public ChargeRefundService(ChargeDao chargeDao, RefundDao refundDao, PaymentProv
 
     public Optional<Response> doRefund(Long accountId, String chargeId, RefundRequest refundRequest) {
         return chargeDao.findByExternalIdAndGatewayAccount(chargeId, accountId)
-                .map(chargeEntity -> refundWithGateway(chargeEntity, refundRequest))
+                .map(chargeEntity -> {
+                    Optional<Response> refundResult = refundWithGateway(chargeEntity, refundRequest);
```
rauligs: - I would add an executeNext() step keeping all a bit separated filtering out those Responses not Sandbox
- I would make it a bit cleaner by doing something the filter using.

```
 private Optional<Response> refundWithGateway(ChargeEntity charge, RefundRequest refundRequest) {
        return Optional.ofNullable(transactionFlowProvider.get()
                .executeNext(createRefund(providers, charge, refundRequest))
                .executeNext(doGatewayRefund(providers))
                .executeNext(setRefundAsSubmitted())
                .executeNext(setSandboxRefundAsRefunded())
                .complete().get(Response.class));
    }
```
end in that step, receiving the result response:

```
refundResult.isSandbox()
 - then set to refunded and update, return
 - otherwise just return the given response
```

It is just I would be very explicit on each step. We can discuss it

src/test/java/uk/gov/pay/connector/service/ChargeRefundServiceTest.java
```diff
@@ -295,7 +289,7 @@ public void shouldRefundSuccessfullyForSandbox() {
         verify(mockChargeDao).merge(charge);
         verify(mockRefundDao).persist(argThat(aRefundEntity(amount, charge)));
         verify(mockProvider).refund(argThat(aRefundRequestWith(charge, amount)));
-        verify(mockRefundDao).merge(any(RefundEntity.class));
+        verify(mockRefundDao, times(2)).merge(any(RefundEntity.class));
```
rauligs: - Rather than set number of times, I would check for one time set as SUBMITTED and a second time as REFUNDED.

rauligs (changes requested):
I have made some suggestions. We can discuss them later

rauligs:
Also needs rebasing

src/test/java/uk/gov/pay/connector/service/ChargeRefundServiceTest.java
```diff
@@ -32,13 +32,7 @@
...
 import static org.junit.Assert.assertThat;
 import static org.junit.Assert.fail;
-import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.argThat;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
juanjoqmelian, line 35: I'd change your IDE settings to avoid using wildcard imports. No need to change it, just to avoid it in the future.

juanjoqmelian (approved):
LGTM

rauligs (approved):

PR #446, opened 2017-09-18
PP-2626 Authorisation refactoring. Transactions.

rauligs:
## WHAT
Authorisation transactions refactoring
- Transaction operations should be defined having the minimum steps necessary while leaving the database in a consistent state

- A typical Authorisation request:
    1. Transaction. PreOperation -> Read charge and update its status to AUTHORISATION_READY (from ENTERING_CARD_DETAILS).

    2. No Transactional. Gateway operation

    3. Transaction.PostOperation -> Read data from Operation (Gateway response) and update charge accordingly + Change status event.

- What has been done:

    1. PreOperation (transactional annotated method) accepting external charge id rather than entity since reads must be within the transaction. Retrieving data must be part of the transaction in order to guarantee consistency.

    2. Still with doubts on OptimisticLockException catching for Authorisation PreOperation. A clearer message will be added in order to analyse when this occurs.

    3. Amended other operations as we go along (PreOperation for capture will execute the merge before to keep backward compatibility).

    4. PostOperation. Removed`merge` to read the entity as done in the PreOperation.

    5. Avoid `merge` charge when creating events, committing the transaction should take care of the necessary queries to update the data according to the entity updates made.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
@@ -29,7 +29,7 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
     private static final String CARD_DETAILS = "cardDetails";
     private static final String REFERENCE = "reference";
     private static final String EMAIL = "email";
-    public static final String SQL_ESCAPE_SEQ = "\\\\";
+    private static final String SQL_ESCAPE_SEQ = "\\\\";
 
     private ChargeEventDao chargeEventDao;
 
@@ -66,7 +66,16 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
     }
 
     public Optional<ChargeEntity> findByExternalIdAndGatewayAccount(String externalId, Long accountId) {
-        return findByExternalId(externalId).filter(charge -> charge.isAssociatedTo(accountId));
+
+        String query = "SELECT c FROM ChargeEntity c " +
+                "WHERE c.externalId = :externalId " +
+                "AND c.gatewayAccount.id = :accountId";
+
+        return entityManager.get()
+                .createQuery(query, ChargeEntity.class)
+                .setParameter("externalId", externalId)
+                .setParameter("accountId", accountId)
+                .getResultList().stream().findFirst();
     }
 
     public Optional<ChargeEntity> findByProviderAndTransactionId(String provider, String transactionId) {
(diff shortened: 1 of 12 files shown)


rauligs:
Next PR will do Capture. Also pending some other clean up

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -129,6 +138,10 @@ public ChargeEntity mergeAndNotifyStatusHasChanged(ChargeEntity chargeEntity, Op
         return mergedCharge;
     }
 
+    public void notifyStatusHasChanged(ChargeEntity chargeEntity, Optional<ZonedDateTime> gatewayEventDate) {
```
mrlumbu, line 141: I know you're following a pattern but I really don't like this method name - wouldn't persistChargeEvent be more informative? Unless I am missing something

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -66,7 +66,16 @@ public ChargeDao(final Provider<EntityManager> entityManager, ChargeEventDao cha
...
+
+        String query = "SELECT c FROM ChargeEntity c " +
+                "WHERE c.externalId = :externalId " +
+                "AND c.gatewayAccount.id = :accountId";
+
+        return entityManager.get()
+                .createQuery(query, ChargeEntity.class)
+                .setParameter("externalId", externalId)
+                .setParameter("accountId", accountId)
+                .getResultList().stream().findFirst();
```
mrlumbu, line 78: This change doesn't seem relevant to purpose of PR. Would be good if you explained why you are doing it

mrlumbu (approved):
Looks good! Can you link in PR description to documentation of issue you are trying to solve in this PR (ie. why do we thing reads need to be in transaction). I have left one comment it would be good if you could address in future PR

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
rauligs: Yes, this is coming in a next PR since for capture will happen the same, in the clean up in the end of PP-2626 will be done.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
rauligs: Very true, forgot mentioning about this, just refactored since we were getting all and then filtering by gateway account.

PR #447, opened 2017-09-19
PP-2626 Capture refactoring. Transactions.

rauligs:
## WHAT
- Capture. Transactions refactoring.
- PreOperation and PostOperation (transactional annotated methods) accepting external charge id rather than entity since reads must be within the transaction. Retrieving data must be part of the transaction in order to guarantee consistency.
- A transaction for an update must be opened before read the entity and close after the update (JPA using entity set method), reading the object and then executing merge can cause undesirable results (very unlikely in our case but still could happen).
- See: 
    - http://blog.xebia.com/jpa-implementation-patterns-saving-detached-entities/
    - https://stackoverflow.com/questions/18101488/does-guice-persist-provide-transaction-scoped-or-application-managed-entitymanag


mrlumbu (approved):
Looks good 👍

PR #448, opened 2017-09-19
PP-2627 Add note about retryFailuresEvery and maximumRetries

alexbishop1:
There’s a reason these values combine to cause retries for two days, so document it.


daniele-occhipinti (approved):
Looks good to me.

I am not terribly happy we use comments, rather than making the code "bullet-proof". However, I understand we need to balance dreams and reality.

This has been approved.

PR #449, opened 2017-09-20
PP-2626 Cancel. Transactions refactoring

rauligs:
## WHAT
- Refactor Cancel services to read charges within the transaction rather than before performing the operation. More info: https://stackoverflow.com/questions/18101488/does-guice-persist-provide-transaction-scoped-or-application-managed-entitymanag

- Change Expire charges as result of cancel methods signature changes, accepting charge external id rather than charge entity.

- CardResource clean-up

### DIFF
--- a/src/test/java/uk/gov/pay/connector/service/ChargeCancelServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/ChargeCancelServiceTest.java
@@ -1,0 +1,219 @@
+package uk.gov.pay.connector.service;
+
+import org.hamcrest.Description;
+import org.hamcrest.TypeSafeMatcher;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.internal.hamcrest.HamcrestArgumentMatcher;
+import org.mockito.runners.MockitoJUnitRunner;
+import uk.gov.pay.connector.dao.ChargeDao;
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/connector/service/ChargeCancelServiceTest.java
```diff
@@ -0,0 +1,219 @@
...
+    public void doUserCancel_shouldFail_whenChargeNotFound() throws Exception {
+
+        String externalChargeId = "external-charge-id";
+
+        when(mockChargeDao.findByExternalId(externalChargeId)).thenReturn(Optional.empty());
+
+        chargeCancelService.doUserCancel(externalChargeId);
+    }
+
+    private HamcrestArgumentMatcher<ChargeEntity> chargeEntityHasStatus(ChargeStatus expectedStatus) {
```
heathd, line 189: maybe rename to `isChargeEntityWithStatus`

src/test/java/uk/gov/pay/connector/service/ChargeCancelServiceTest.java
```diff
@@ -0,0 +1,219 @@
...
+            }
+
+            @Override
+            public void describeTo(Description description) {
+
+            }
+        });
+    }
+
+    private HamcrestArgumentMatcher<CancelGatewayRequest> aCancelGatewayRequestMatching(ChargeEntity chargeEntity) {
```
heathd, line 203: maybe rename to `isCancelGatewayRequestForCharge`

heathd (approved):
ok looks good

two minor changes suggested but we can do that later

PR #450, opened 2017-09-20
PP-2626 Capture approval and error

rauligs:
## WHAT
- Perform read operations within the transactional methods. More info: https://stackoverflow.com/questions/18101488/does-guice-persist-provide-transaction-scoped-or-application-managed-entitymanag

- Ignore charge not found in mark charge as error. Ids are coming from a previous retrieved list.


heathd (approved):
looks good

PR #451, opened 2017-09-21
PP-2626 Refunds. Transactions refactoring

rauligs:
## WHAT
- Refactor Refunds service to read charges within the transaction rather than before performing the operation. More info: https://stackoverflow.com/questions/18101488/does-guice-persist-provide-transaction-scoped-or-application-managed-entitymanag

- Note that there is no null check in finish method for a refund. Intentional I would expect a _NullpointerException_ if this ever happens.

- https://github.com/alphagov/pay-connector/pull/450 to be merged first


heathd (approved):
looks good

PR #452, opened 2017-09-22
PP 2626 Transactions refactoring

rauligs:
## WHAT
- Wrap read of the charges following operation within a transaction.
    - Charges update status. 
    - Email notifications.
    - Patch charges.
    - Create charges.

PR #453, opened 2017-09-22
PP-2626 Transactions refactoring. Clean up

rauligs:
## WHAT

- Use _ChargeEventDao_ instead of _ChargeDao_ to persist events of a charge.
- Removed unused Transactional method update statuses
- Merge after https://github.com/alphagov/pay-connector/pull/452


ckalista (dismissed):
Definitely a safer approach. In the future we will probably get rid of having to explicitly record charge event entries by leveraging EclipseLink History Policy the same way we keep track of history records for refund.

ckalista (dismissed):

ckalista (approved):

PR #456, opened 2017-09-25
Improve gateway operation response logging

alexbishop1:
Recently, there have been a few occassions where I’ve had to look in the logs and found the information logged when we receive a response to a gateway operation request (authorisation, capture, cancel, refund, etc.) to be somewhat lacking.

Currently, when we receive a response to an authorisation request, we log, at INFO level, the following (with the placeholders replaced by actual values, of course, and some line breaks have been added):

```
AuthCardDetails authorisation response received - charge_external_id={},
operation_type={}, transaction_id={}, status={}
```

Note that `status` is the new status that the payment will be moved to as a result of the response. Meanwhile, `operation_type` is always “Authorisation”, so is basically redundant.

For 3D Secure authorisations, we log this at INFO:

```
AuthCardDetails authorisation response received - charge_external_id={},
operation_type={}, transaction_id={}, status={}
```

There’s a mistake here: it should be `Auth3dsDetails` rather than `AuthCardDetails`. The operation type is “3D Secure Response Authorisation” but otherwise it’s the same as regular authorisation.

Capture logs a line at INFO like this:

```
Card capture response received - charge_external_id={}, operation_type={},
transaction_id={}, status={}
```

Once again, `status` is the new status. The `operation_type` is always “Capture”.

For cancel, we split the relevant details across two lines at INFO:

```
Card cancel response received - charge_external_id={}, transaction_id={},
status={}
```

```
Charge status to update - charge_external_id={}, status={}, to_status={}
```

In both, `status` is the old status.

Refunds also get two lines at INFO:

```
Card refund response received -  transaction_id={}, charge_id={},
charge_external_id={}, refund_id={}, refund_external_id={},
refund_reference={}, refund_status={}, refund_amount={}
```

```
Refund status to update - status={}, to_status={} for transaction_id={},
charge_id={}, charge_external_id={}, refund_id={}, refund_external_id={},
refund_reference={}, refund_status={}, refund_amount={}
```

The `refund_id` is an entity database ID, which we never seem to log for anything else. Both `refund_status` and `status` are the old status of the refund, so the second line includes it twice!

What none of these log lines contain is any real indication of what was in the gateway response (or if some problem meant there wasn’t a gateway response): the best we get is the status the payment (or refund) will become. Sometimes more details about the content of the response or errors encountered in receiving it will be logged elsewhere but they are usually devoid of information about which payment they relate to. This can mean having to look at adjacent log lines or compare timestamps to fully investigate a problem.

Furthermore, there is no indication of the payment gateway or service in question.

This pull request makes the log lines look like this (with some example values):

```
Authorisation for abcdefghijklmnopqrstuvwxyz (epdq 1111111) for MOF-TLA (42) —
ePDQ authorisation response (PAYID: 1111111, STATUS: 5) ∴ AUTHORISATION READY →
AUTHORISED
```

```
3DS response authorisation for abcdefghijklmnopqrstuvwxyz (worldpay
fddebf25-960e-4fe5-aee6-ce2c8b29ed5c) for MOF-TLA (42) — Worldpay authorisation
response (orderCode: fddebf25-960e-4fe5-aee6-ce2c8b29ed5c, lastEvent:
AUTHORISED) ∴ AUTHORISATION 3DS READY → AUTHORISED
```

```
Capture for abcdefghijklmnopqrstuvwxyz (worldpay
fddebf25-960e-4fe5-aee6-ce2c8b29ed5c) for MOF-TLA (42) — Gateway error:
Worldpay capture response (error code: 1, error: Internal error) ∴ CAPTURE
READY → CAPTURE APPROVED RETRY
```

```
Cancel for abcdefghijklmnopqrstuvwxyz (smartpay 8313547924770610) for MOF-TLA
(42) — Gateway error: Gateway connection timeout error ∴ USER CANCEL READY
→ USER CANCEL ERROR
```

```
Refund zyxwvutsrqponmlkjihgfedcba (ePDQ 1111111/3) for
abcdefghijklmnopqrstuvwxyz (ePDQ 1111111) for MOF-TLA (42) — ePDQ refund
response (PAYID: 1111111, PAYIDSUB: 3, STATUS: 81) ∴ CREATED → REFUND
SUBMITTED
```

While not perfect, the intention is to fit in more information and rely on context rather than snake case labels that are neither the exact names of fields nor true structured logging.

(Note that the refund log lines no longer include the entity database IDs nor the amount, which are probably redundant.)

Each log line has three parts: the bit before the em dash, the bit between the em dash and the “∴” (“therefore” symbol) and the bit after the “∴”.

With the exception of refunds, the part before the hyphen includes the type of operation, the external ID of the charge, the payment gateway and its transaction ID (grouped together in parentheses), and the analytics ID of the service (seems a reasonable alternative to the verbose full name) with the gateway account ID in brackets (because many gateway accounts may share the same service name). With the exception of authorisation, the transaction ID is always what the charge already had, not what’s in the response (which will either be the same or not present).

For refunds, we include the operation type, the external ID of the refund, the payment gateway and refund reference (grouped together in parentheses), and the charge details as described above.

After the hyphen is simply what `toString()` returns for the `GatewayResponse`. A `GatewayResponse` is really just a wrapper around an `Either` of something that extends `BaseResponse` or a `GatewayError`. So the `toString()` of `GatewayResponse` just defers to these.

All classes that extend `BaseResponse` have had a `toString()` method added that outputs the fields we extract from the response (omitting any that are blank), using their original names (or close to them) to make cross-referencing with the gateways’ documentation etc. easier.

The `GatewayError` has had its `toString()` rewritten to be “Gateway Error: ” followed by its `message`. This message may be something along the lines of “Gateway connection timeout error” or it may come from a failed `BaseResponse`.

When constructing the `message` for a `GatewayError` from a failed `BaseResponse`, we now just `toString()` the `BaseResponse` (see above) rather than extracting the error code and message and converting to the form “[1] Internal error” (a couple of tests that tested this behaviour have been removed).

We no longer include the value of the `ErrorType` enum in the `toString()` of `GatewayError` as it usually just repeats what the message said or is simply `GENERIC_GATEWAY_ERROR`.

Finally, after the “∴” (“therefore” symbol), we display the old and new status of the charge or refund.

For now, no other log lines have been removed, even though they may now contain duplicated information (albeit at higher levels like ERROR than INFO). These may be examined at a later date.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/GatewayError.java
+++ b/src/main/java/uk/gov/pay/connector/model/GatewayError.java
@@ -1,6 +1,11 @@
 package uk.gov.pay.connector.model;
 
-import static uk.gov.pay.connector.model.ErrorType.*;
+import static uk.gov.pay.connector.model.ErrorType.GATEWAY_CONNECTION_SOCKET_ERROR;
+import static uk.gov.pay.connector.model.ErrorType.GATEWAY_CONNECTION_TIMEOUT_ERROR;
+import static uk.gov.pay.connector.model.ErrorType.GATEWAY_URL_DNS_ERROR;
+import static uk.gov.pay.connector.model.ErrorType.GENERIC_GATEWAY_ERROR;
+import static uk.gov.pay.connector.model.ErrorType.MALFORMED_RESPONSE_RECEIVED_FROM_GATEWAY;
+import static uk.gov.pay.connector.model.ErrorType.UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY;
 
 public class GatewayError {
     private String message;
@@ -45,9 +50,6 @@ public class GatewayError {
 
     @Override
     public String toString() {
-        return "GatewayError{" +
-                "message='" + message + '\'' +
-                ", errorType=" + errorType +
-                '}';
+        return "Gateway error: " + message;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
+++ b/src/main/java/uk/gov/pay/connector/model/gateway/GatewayResponse.java
@@ -12,7 +12,6 @@ import java.util.Optional;
 import static fj.data.Either.left;
 import static fj.data.Either.right;
 import static java.lang.String.format;
-import static org.apache.commons.lang3.StringUtils.trim;
 import static uk.gov.pay.connector.model.GatewayError.baseError;
 
 public class GatewayResponse<T extends BaseResponse> {
@@ -73,6 +72,11 @@ public class GatewayResponse<T extends BaseResponse> {
         );
     }
 
+    @Override
+    public String toString() {
+        return response.either(GatewayError::toString, T::toString);
+    }
+
     static public <T extends BaseResponse> GatewayResponse<T> with(GatewayError gatewayError) {
         logger.error(format("Error received from gateway: %s", gatewayError));
         return new GatewayResponse<>(gatewayError);
@@ -114,10 +118,7 @@ public class GatewayResponse<T extends BaseResponse> {
             Optional<String> errorMessage = getErrorMessage(response);
 
             if (errorCode.isPresent() || errorMessage.isPresent()) {
-                StringBuilder sb = new StringBuilder();
-                errorCode.ifPresent(e -> sb.append(format("[%s] ", e)));
-                errorMessage.ifPresent(sb::append);
-                return new GatewayResponse<>(baseError(trim(sb.toString())));
+                return new GatewayResponse<>(baseError(response.toString()));
             }
 
             return new GatewayResponse<>(response, sessionIdentifier);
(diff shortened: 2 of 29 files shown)


mrlumbu:
This all looks marvellous. A couple of things:
- need to check what effect if any this would have on alerts etc. 
- you have as part of this converted all wildcard imports to fully qualified ones - could you mention this and why in your PR description (if nothing else to educate others). It would perhaps have been nicer if this had been done in a separate commit

alexbishop1:
> need to check what effect if any this would have on alerts etc.

There’s a Sumo Logic search for “Gateway error” or “Internal error”. That’s going to start finding more stuff: we’ll log “Internal error” in more places if we get it in a gateway response and we’re also start logging “Gateway error”. There is already PP-2665 to reassess that alert, as it gets a lot of false positives.

> you have as part of this converted all wildcard imports to fully qualified ones - could you mention this and why in your PR description (if nothing else to educate others). It would perhaps have been nicer if this had been done in a separate commit

That wasn’t really intentional: I just optimised the imports (to make sure there were no unused ones) and my IDE settings did the rest. The proposed Java style guide at https://github.com/alphagov/gds-tech/pull/43/files does suggest not using wildcard imports. See http://javadude.com/articles/importondemandisevil.html for some reasoning I mostly agree with.

mrlumbu:
Ok cool. LGTM  👍

mrlumbu (dismissed):

mrlumbu (approved):

heathd:
Nice!

PR #460, opened 2017-09-27
PP-2626 Add extra logging in caught OptimisticLockExceptions

rauligs:
## WHAT

- For those ones that are caught and charge `external_id` is available, Log `INFO` with the exception with `external_id` of the charge affected.

- For _TransactionFlow_ instances the exception thrown is modified to add some extra information (`ConflictRuntimeException`).


ckalista (dismissed):
That'll definitely help!

ckalista (approved):

PR #463, opened 2017-10-04
Add toString() methods to sandbox gateway responses

alexbishop1:
We now output stringifications of `BaseResponse` implementations in the logs. Appropriate `toString()` methods were added for the other payment gateways but sandbox was missed because its responses are anonymous classes within `SandboxPaymentProvider`.

Also make it so that `getTransactionId()` doesn’t return a different (random) identifier each time, which is surely wrong.


rauligs (approved):
LGTM 👍

PR #465, opened 2017-10-09
PP-2714 Create payment_requests table

georgeracu:
- Created PaymentRequestEntity;
- Created PaymentRequestDao;
- Created test for Dao;
- Created PaymentRequestEntityFixture;
- Changed ChargeService#create to write to payment_requests table after writing into charges, inside the same transaction;
- Added test for the new insert for the ChargeService;
- Added migration for the new table;

with @alexbishop1 

_WHAT_
The purpose of this change is to enable database refactoring for the table `charges`. This table will be split into 3 other tables, and the first one is `payment_charges`.
Initially the `create` charge operation into the `ChargeService` will start writing into both tables `charges` and `payment_requests` but will keep reading only from the `charges`. The `id` column in `payment_charges` will be the same as in `charges` because later on the missing data from the `payment_charges` will be backfilled.

Edit: added context on what this change is and enables for future work.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
@@ -1,0 +1,106 @@
+package uk.gov.pay.connector.model.domain;
+
+import javax.persistence.*;
+import java.time.ZonedDateTime;
+
+@Entity
+@Table(name = "payment_requests")
+@Access(AccessType.FIELD)
+public class PaymentRequestEntity extends AbstractEntity {
+
+    @Column(name = "amount")
+    private Long amount;
+
+    @Column(name = "return_url")
+    private String returnUrl;
+
+    @ManyToOne
+    @JoinColumn(name = "gateway_account_id", updatable = false)
+    private GatewayAccountEntity gatewayAccount;
+
+    @Column(name = "description")
+    private String description;
+
+    @Column(name = "reference")
+    private String reference;
+
+    @Column(name = "created_date")
+    @Convert(converter = UTCDateTimeConverter.class)
+    private ZonedDateTime createdDate;
+
+    @Column(name = "external_id")
+    private String externalId;
+
+    public PaymentRequestEntity() {
+        // enjoy it JPA
+    }
+
+    public Long getAmount() {
+        return amount;
+    }
+
+    public void setAmount(Long amount) {
+        this.amount = amount;
+    }
+
+    public String getReturnUrl() {
+        return returnUrl;
+    }
+
+    public void setReturnUrl(String returnUrl) {
+        this.returnUrl = returnUrl;
+    }
+
+    public GatewayAccountEntity getGatewayAccount() {
+        return gatewayAccount;
+    }
+
+    public void setGatewayAccount(GatewayAccountEntity gatewayAccount) {
+        this.gatewayAccount = gatewayAccount;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    public String getReference() {
+        return reference;
+    }
+
+    public void setReference(String reference) {
+        this.reference = reference;
+    }
+
+    public ZonedDateTime getCreatedDate() {
+        return createdDate;
+    }
+
+    public void setCreatedDate(ZonedDateTime createdDate) {
+        this.createdDate = createdDate;
+    }
+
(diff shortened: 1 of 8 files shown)


mrlumbu:
Can I request you put a bit more about why you're doing this/context into the PR  and commit message? Good practice, allows us to reason about changes later, and allows a greater range of people to confidently review changes

src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
```diff
@@ -0,0 +1,107 @@
...
+        return externalId;
+    }
+
+    public void setExternalId(String externalId) {
+        this.externalId = externalId;
+    }
+
+    public static PaymentRequestEntity from(ChargeEntity chargeEntity) {
+        PaymentRequestEntity paymentEntity = new PaymentRequestEntity();
+        paymentEntity.setId(chargeEntity.getId());
```
heathd: on one hand, I can see some sense in keeping the IDs the same. On the other hand, you'll risk overlapping with IDs generated by the database sequence. I'm not totally sure whether liquibase creates a sequence automatically for primary key fields? 

Another option would be to just allow the paymentrequest to use the natural PK generated from the sequence and rely on the fact that the external IDs are the same between charge and PaymentRequest

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestIDaoTest.java
```diff
@@ -0,0 +1,57 @@
...
+import uk.gov.pay.connector.model.domain.PaymentRequestEntity;
+
+import java.util.HashMap;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.*;
+import static uk.gov.pay.connector.model.domain.GatewayAccountEntity.Type.TEST;
+import static uk.gov.pay.connector.model.domain.PaymentRequestEntityFixture.aValidPaymentRequestEntity;
+
+public class PaymentRequestIDaoTest extends DaoITestBase {
```
heathd: rename to `PaymentRequestDaoITest`

src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
alexbishop1: After much discussion, we have decided that while it has some appeal, we’re not going to try to keep the same IDs across both tables. We will rely on the default autogenerated sequences and the fact that the external IDs will match.

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestIDaoTest.java
alexbishop1: ✅

src/test/java/uk/gov/pay/connector/model/domain/PaymentRequestEntityFixture.java
```diff
@@ -0,0 +1,48 @@
...
+    private GatewayAccountEntity gatewayAccountEntity = defaultGatewayAccountEntity();
+    private ZonedDateTime createdDate = ZonedDateTime.now(ZoneId.of("UTC"));
+
+    public static PaymentRequestEntityFixture aValidPaymentRequestEntity() {
+        return new PaymentRequestEntityFixture();
+    }
+
+    public PaymentRequestEntity build() {
+        PaymentRequestEntity entity = new PaymentRequestEntity();
+        entity.setId(this.id);
```
heathd: do we really want to set this in the fixture rather than using the auto-generated value?

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Test
+    public void shouldCreateAPaymentRequest() throws Exception {
+        GatewayAccountEntity gatewayAccount = new GatewayAccountEntity(
+                defaultTestAccount.getPaymentProvider(), new HashMap<>(), TEST);
+        gatewayAccount.setId(defaultTestAccount.getAccountId());
+
+        PaymentRequestEntity paymentRequestEntity = aValidPaymentRequestEntity()
+                .withId(null)
```
heathd: can we just remove the `withId()` method, since we're auto-incrementing this now?

heathd (approved):
Looks good 👍

PR #466, opened 2017-10-11
PP-2744 Create cards table

georgeracu:
- Added migration script;
- Added entity `CardEntity`;
- Added tests and fixtures;

with @alexbishop1

## WHAT
* Another step into splitting the `charges` table into smaller pieces. This table `cards` holds details regarding address, name and card details for a payment request. 
* It follows the pattern where we write in parallel in `charges` and `cards` but we read only from `charges`.
* It will be backfilled later on with data and we will switch to read from it as well.


src/main/java/uk/gov/pay/connector/model/domain/CardEntity.java
```diff
@@ -0,0 +1,136 @@
...
+@Table(name = "cards")
+public class CardEntity extends AbstractEntity {
+    @Column(name = "last_digits_card_number")
+    private String lastDigitsCardNumber;
+
+    @Column(name = "cardholder_name")
+    private String cardHolderName;
+
+    @Column(name = "expiry_date")
+    private String expiryDate;
```
chrisgrimble, line 18: Should be a ZonedDateTime and has a UTCDateTimeConverter on it.

src/main/resources/migrations.xml
```diff
@@ -695,4 +695,54 @@
...
+            <column name="email" type="varchar(50)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="last_digits_card_number" type="char(4)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="cardholder_name" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="expiry_date" type="varchar(6)">
```
chrisgrimble, line 721: Should be a timestamp without timezone not a varchar

chrisgrimble (changes requested):
Please see comments in code about dates.

src/main/java/uk/gov/pay/connector/model/domain/CardEntity.java
alexbishop1: Expiry date is just a month and a year (it’s exactly like the “09/22” on the card). We store these as strings (we never treat them as dates, just send them to the payment gateway and show them in selfservice). So I think it’s fine as it is.

chrisgrimble (approved):
Comments explain why we would not want to save this as a date.

PR #467, opened 2017-10-12
PP-2714 Add not null constraint on payment_requests.external_id

georgeracu:
with @alexbishop1

## WHAT
Create `not null constraint` on `payment_requests.external_id`

## HOW 
Using Liquibase migration


whpearson (approved):
Looks good.

PR #468, opened 2017-10-12
PP-2760 Send cardholder name and address to ePDQ with authorisation requests

alexbishop1:
When sending authorisation `RES` requests to ePDQ, include the cardholder name and address, so that they can be used for fraud checks etc.

We now send the following additional parameters (documented at https://support.epdq.co.uk/en/guides/integration%20guides/directlink/request-a-new%20order#requestparameters):

`CN` = cardholder name
`OWNERADDRESS` = street address
`OWNERCTY` = country code (e.g. GB)
`OWNERTOWN` = city
`OWNERZIP` = postcode

While we have two fields for the street address (a mandatory first line and an optional second line), ePDQ just have one. When there are two lines, we combine them by concatenating the first line, then a comma, then a space, then the second line (for example, if the first line is “The White Chapel Building” and the second line is “10 Whitechapel High Street”, we combine them as “The White Chapel Building, 10 Whitechapel High Street”), which seems to be the only sensible thing to do.

Testing against ePDQ’s test environment shows that the data seems to be processed appropriately. In the ePDQ back office, the cardholder name appears against the “Delivery customer”, “Cardholder's name” and “Invoicing customer”. The street address, postcode, city and country code appear against the “Address”. The country code and mapped country name appear against the “Buyer's invoicing country”.

ePDQ have length limits for the values of some of the parameters that are smaller than ours. However, testing shows that if we send something that is too long, they silently truncate at their end, so this should not be a concern.

We already send this information to other payment gateways, so there are no new privacy concerns.


rauligs (approved):
LGTM 👍

PR #469, opened 2017-10-16
PP-2759 Create table card_3ds

georgeracu:
- Updated migration script
- Added new entity `Card3dsEntity`
- Added new Dao `Card3dsDao`
- Added creation of `Card3dsEntity` when the auth response has necessary
fields;
- Added tests and updated existing to test new code;

with @alexbishop1
with @chrisgrimble

## WHAT
Create a new table `card_3ds` and populate it at the same time when populating records
 in table `charges`.
The table will be backfilled at a later stage.


heathd (approved):
looks good

PR #470, opened 2017-10-16
PP-2744 Make email column in cards table be varchar(254)

alexbishop1:
Amusingly, we repeated the same mistake made when the `email` column was first added to the `charges` table by making it a `varchar(50)`, which is too short.

with @georgeracu
with @chrisgrimble


daniele-occhipinti (approved):
LGTM

PR #471, opened 2017-10-19
PP-2768 Backfill cards table from charges

georgeracu:
- Added index on column cards.charge_id
- Added migration to insert into cards missing records from charges

with @chrisgrimble

## WHAT
Add to the `cards` table missing rows from `charges`.

## HOW 
Use Liquibase migrations to copy data from `charges` to `cards`.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -779,7 +779,7 @@
     <changeSet id="createIndex cards.charge_id" author="">
         <createIndex indexName="idx_cards_charge_id"
                      tableName="cards"
-                     unique="false">
+                     unique="true">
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -776,4 +776,47 @@
                         tableName="cards"/>
     </changeSet>
 
+    <changeSet id="createIndex cards.charge_id" author="">
+        <createIndex indexName="idx_cards_charge_id"
+                     tableName="cards"
+                     unique="false">
```
simad: Any reason this shouldn't be unique? `charge_id` was the id generated by JPA so it _should_ have been unique

simad (approved):

PR #474, opened 2017-10-26
PP-2825 Remove email from cards table

chrisgrimble:
The email is not written at the same time as the other card data and we
are thinking of moving it to transactions. For now remove it from cards
table and just leave it on charges. Have already stopped the email
column being used.

- Dropped email from the cards table.


alexbishop1 (approved):
🙃

PR #475, opened 2017-10-26
PP-2768 Backfill cards table

chrisgrimble:
Backfill the data from charges in to cards. We have some data missing as
originally we did not store all the card details. We also need to check
the user has entered card details as they could be in flight or have
abandoned their payment. In which case we will have a row in charges but
do not want one in cards.

- Add index on charge_id in cards.
- Made chargeId a bigint not bigserial as it is the forign key to
charges and so should not be set by a sequence.
- Make columns nullable as some are null in the real data. This is
because some rows were created before the columns were added to the
charges table.
- Create backfill sql. Do not backfill the rows in charges where all the
card details are null. This is because the payment is in progress and we
don't create the entry in cards until after the user has authorised
their card details.

with @alexbishop1 @georgeracu

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -783,4 +783,66 @@
     <changeSet id="drop email from cards table" author="">
         <dropColumn tableName="cards" columnName="email"/>
     </changeSet>
+
+    <changeSet id="createIndex cards.charge_id" author="">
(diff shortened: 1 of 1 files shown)


ckalista (approved):
It seems like a sensible query to run given the context. 

👍

PR #476, opened 2017-10-26
PP-2743 Backfill payment_requests table from charges

chrisgrimble:
- Added index on column payment_requests.external_id
- Added migration to insert into payment_requests
missing records from charges
- Changed the type of gateway_account_id as we do not want the default
to be set from a sequence. If is a foreign key on the gateway-accounts
table.

with @georgeracu

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -819,30 +819,67 @@
             charge_id
             )
             SELECT
-                ch.cardholder_name,
-                ch.card_brand,
-                ch.last_digits_card_number,
-                ch.expiry_date,
-                ch.address_line1,
-                ch.address_line2,
-                ch.address_postcode,
-                ch.address_city,
-                ch.address_county,
-                ch.address_country,
-                ch.id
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -819,30 +819,69 @@
...
+    <changeSet id="backfill payment_requests from charges" author="">
+        <sql>
+            INSERT INTO payment_requests (
+                amount,
+                return_url,
+                gateway_account_id,
+                description,
+                reference,
+                created_date,
+                version,
```
alexbishop1: Not sure about this. Conceptually, the `version` should be `0` because it’s the first version of this row in the `payment_requests` table. It probably doesn’t matter if we just copy across the `version` from `charges`. It might be nicer to keep them all at `0` though, since we expect rows in `payment_requests` to never change after the initial write (and if any do become higher than `0`, that’s an indication we’ve screwed up somewhere!).

alexbishop1 (approved):
This probably won’t bring about the apocalypse

PR #478, opened 2017-10-27
PP-2777 Create transactions table

chrisgrimble:
- Added liquibase changeset to create the transactions table
- Mapped Transactions table to Transactions entity which links to the
PaymentRequestEntity.
- Made TransactionEntity reference PaymentRequestEntity. This creates a
bidirectional relationship which means the transaction can be inserted
with a single sql statement. Otherwise JPA will try to insert the
Transaction with a null in the payment_request_id column, then create
another statement to update the the payment_request_id.
- Moved chargeStatusUpdate to ChargeService from ChargeFrontendResource.
- Update Transactions when change status.
- Setting gatewayTransactionId on the chargeTransaction in the same
places as it is set on the charge.

with @georgeracu

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/PaymentRequestDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/PaymentRequestDao.java
@@ -3,9 +3,11 @@ package uk.gov.pay.connector.dao;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
+import uk.gov.pay.connector.model.domain.ChargeStatus;
 import uk.gov.pay.connector.model.domain.PaymentRequestEntity;
-
 import javax.persistence.EntityManager;
+import java.util.Optional;
+
 
 @Transactional
 public class PaymentRequestDao extends JpaDao<PaymentRequestEntity> {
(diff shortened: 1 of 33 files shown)


mrlumbu:
Not gonna do a full review now, but just an initial comment - I think I would prefer the migration and code changes to be in separate PR's, unless there is some compelling reason not to. That way it is easier to make clear explicit dependence of code changes on migration, and everything is less confusing.

chrisgrimble:
Still need to do the migration of the data. This just creates the new table and uses it for any new payment transactions created. Have removed 

- Added liquibase migration to populate Transactions table.

from the PR description as that was not in there no idea why that got included.

mrlumbu:
I still think given migration and code changes will be deployed at two separate times (or am I wrong about this - I see we still have the migrate step in docker-startup.sh?), it would be less confusing to have this as two separate PRs

chrisgrimble:
Not yet, but we have just prioritised getting that work out. So will split these two up as it will be needed once we deploy database changes separately to startup.

chrisgrimble:
Closing this PR as we are going to take creating the table out and this has been done on a new branch. Will raise an new PR on this branch when that is done.

PR #479, opened 2017-10-30
PP-2769 Rollback backfill for card_3ds

georgeracu:
- Rollback backfill until we implement a new way of doing through code;
- Left changing column type and adding index

@with chrisgrimble

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -897,25 +897,4 @@
         </createIndex>
     </changeSet>
 
-    <changeSet id="backfill card_3ds table" author="">
-        <sql>
-            INSERT INTO card_3ds (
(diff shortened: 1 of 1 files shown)


kakumara:
Is this already rolled out to production/AWS? if so, rollback this way won't do any good. you probably will need to have a separate migration to rollback rather than removing a existing migration. Coz liquibase maintains a history and as far as I know you cannot revert history rather you can only add on top of it

_PRs runs on a fresh DB, so it might become green. you need to test this against a DB which already in the same state_

georgeracu:
@kakumara is not in production nor staging. It passed testing and was waiting promotion to staging but it didn't happen.

kakumara (approved):
LGTM

PR #480, opened 2017-10-30
PP-2777 Create transactions table

chrisgrimble:
Create the transactions table. Table will not be used for now but as we
are splitting up db releases and migrating the database the table needs
to be there before we can start using it.

- Add create table to migrations scripts

with @georgeracu


SandorArpa (approved):
LGTM 👍

PR #482, opened 2017-11-01
PP-2733 Revert disabled automatic migrations

chrisgrimble:
Disbaling the automatic migration at startup cause issues with the tests
in CI and for other devs as they have not built their dbs.

- Revert change to docker-startup.sh so we are running the migration as
part of the startup again.

with @georgeracu


SandorArpa (approved):
LGTM 👍

PR #483, opened 2017-11-01
PP-2733 Make connector database migrations not be run during deployment

chrisgrimble:
Disable the automatic db migration at startup. As some environments
were depending on this to create their db's at startup added a new
script that will still run the db migration at startup. Then override
the docker command in docker-compose to use this script.

- Added new docker-local-startup.sh file that still does the db
migration this is what the docker-start.sh file used to to do.
- Add the new file to the docker image that is created.
- Removed db migrate from docker-start.sh.

with @georgeracu


juanjoqmelian (approved):
LGTM

PR #484, opened 2017-11-06
PP-2777 create table transaction

chrisgrimble:
Start writing and updating the Transaction table as well as the Charges 
table.

Check that a PaymentRequest has a charge transaction before 
modifying it. This is needed until we back fill the transactions table with
anything from charges. Once that has been done this commit can be
reverted. Have done it this way as it is a relatively small change which
will be easy to revert. Thought about creating a new ChargeTransaction
when you call getChargeTransaction if it did not exist. However this
required a ChargeEntity to be passed to getChargeTransaction() and the
changes were becoming considerable. Want this to be easy to remove once
the transactions table has been back filled.

- Mapped Transactions table to Transactions entity which links to the
PaymentRequestEntity.
- Made TransactionEntity reference PaymentRequestEntity. This creates a
bidirectional relationship which means the transaction can be inserted
with a single sql statement. Otherwise JPA will try to insert the
Transaction with a null in the payment_request_id column, then create
another statement to update the the payment_request_id.
- Moved chargeStatusUpdate to ChargeService from ChargeFrontendResource.
- Update Transactions when change status.
- Setting gatewayTransactionId on the chargeTransaction in the same
places as it is set on the charge.
- Add a hasChargeTransaction() to PaymentRequestEntity.
- Check hasChargeTransaction() before making a change to the charge
transaction.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/PaymentRequestDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/PaymentRequestDao.java
@@ -3,9 +3,11 @@ package uk.gov.pay.connector.dao;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
+import uk.gov.pay.connector.model.domain.ChargeStatus;
 import uk.gov.pay.connector.model.domain.PaymentRequestEntity;
-
 import javax.persistence.EntityManager;
+import java.util.Optional;
+
 
 @Transactional
 public class PaymentRequestDao extends JpaDao<PaymentRequestEntity> {
@@ -13,5 +15,15 @@ public class PaymentRequestDao extends JpaDao<PaymentRequestEntity> {
     @Inject
     public PaymentRequestDao(final Provider<EntityManager> entityManager) {
         super(entityManager);
+    }
+
+    public Optional<PaymentRequestEntity> findByExternalId(String externalId) {
+        String query = "SELECT p FROM PaymentRequestEntity p " +
+                "WHERE p.externalId = :externalId";
+
+        return entityManager.get()
+                .createQuery(query, PaymentRequestEntity.class)
+                .setParameter("externalId", externalId)
+                .getResultList().stream().findFirst();
     }
 }
--- a/src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
@@ -14,7 +14,7 @@ import java.util.List;
 
 import static org.apache.commons.lang3.StringUtils.equalsIgnoreCase;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.*;
-import static uk.gov.pay.connector.model.domain.PaymentGatewayStateTransitions.stateTransitionsFor;
+import static uk.gov.pay.connector.model.domain.PaymentGatewayStateTransitions.defaultTransitions;
 
 @Entity
 @Table(name = "charges")
@@ -150,7 +150,7 @@ public class ChargeEntity extends AbstractEntity {
     }
 
     public void setStatus(ChargeStatus targetStatus) throws InvalidStateTransitionException {
-        if (stateTransitionsFor(getPaymentGatewayName()).isValidTransition(fromString(this.status), targetStatus)) {
+        if (defaultTransitions().isValidTransition(fromString(this.status), targetStatus)) {
             this.status = targetStatus.getValue();
         } else {
             throw new InvalidStateTransitionException(this.status, targetStatus.getValue());
--- a/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
@@ -1,7 +1,15 @@
 package uk.gov.pay.connector.model.domain;
 
+import uk.gov.pay.connector.model.domain.transaction.ChargeTransactionEntity;
+import uk.gov.pay.connector.model.domain.transaction.TransactionEntity;
+import uk.gov.pay.connector.model.domain.transaction.TransactionOperation;
+
 import javax.persistence.*;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Predicate;
 
 @Entity
 @Table(name = "payment_requests")
@@ -30,6 +38,9 @@ public class PaymentRequestEntity extends AbstractEntity {
 
     @Column(name = "external_id")
     private String externalId;
+
+    @OneToMany(mappedBy = "paymentRequest", cascade = CascadeType.ALL)
+    private List<TransactionEntity> transactions = new ArrayList<>();
 
     public PaymentRequestEntity() {
         // enjoy it JPA
@@ -91,7 +102,31 @@ public class PaymentRequestEntity extends AbstractEntity {
         this.externalId = externalId;
     }
 
-    public static PaymentRequestEntity from(ChargeEntity chargeEntity) {
+    public List<TransactionEntity> getTransactions() {
+        return Collections.unmodifiableList(transactions);
+    }
+
+    public void addTransaction(TransactionEntity transactionEntity) {
+        this.transactions.add(transactionEntity);
+        transactionEntity.setPaymentRequest(this);
+    }
+
+    public void setTransactions(List<TransactionEntity> transactions) {
+        this.transactions = transactions;
+        transactions.forEach(transactionEntity -> transactionEntity.setPaymentRequest(this));
+    }
+
+    public ChargeTransactionEntity getChargeTransaction() {
+        return (ChargeTransactionEntity)transactions.stream().filter(byChargeTransactions()).findFirst().orElseThrow(
+                () -> new IllegalStateException("Payment Request has been initialised without a charge transaction")
+        );
+    }
+
+    private Predicate<TransactionEntity> byChargeTransactions() {
+        return transactionEntity -> transactionEntity.getOperation().equals(TransactionOperation.CHARGE);
+    }
+
+    public static PaymentRequestEntity from(ChargeEntity chargeEntity, ChargeTransactionEntity transactionEntity) {
         PaymentRequestEntity paymentEntity = new PaymentRequestEntity();
         paymentEntity.setAmount(chargeEntity.getAmount());
         paymentEntity.setCreatedDate(chargeEntity.getCreatedDate());
@@ -100,6 +135,7 @@ public class PaymentRequestEntity extends AbstractEntity {
         paymentEntity.setGatewayAccount(chargeEntity.getGatewayAccount());
         paymentEntity.setReference(chargeEntity.getReference());
         paymentEntity.setReturnUrl(chargeEntity.getReturnUrl());
+        paymentEntity.addTransaction(transactionEntity);
 
         return paymentEntity;
     }
--- a/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
@@ -122,6 +122,11 @@ public class PaymentRequestEntity extends AbstractEntity {
         );
     }
 
+    //Remove this once transactions table has been backfilled and we will no longer need to set this.
+    public boolean hasChargeTransaction() {
+        return transactions.stream().anyMatch(byChargeTransactions());
+    }
+
     private Predicate<TransactionEntity> byChargeTransactions() {
         return transactionEntity -> transactionEntity.getOperation().equals(TransactionOperation.CHARGE);
     }
--- a/src/main/java/uk/gov/pay/connector/model/domain/transaction/ChargeTransactionEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/transaction/ChargeTransactionEntity.java
@@ -1,0 +1,44 @@
+package uk.gov.pay.connector.model.domain.transaction;
+
+import uk.gov.pay.connector.exception.InvalidStateTransitionException;
+import uk.gov.pay.connector.model.domain.*;
+
+import javax.persistence.*;
+
+import static uk.gov.pay.connector.model.domain.PaymentGatewayStateTransitions.defaultTransitions;
+
+@Entity
+@DiscriminatorValue(value = "CHARGE")
+public class ChargeTransactionEntity extends TransactionEntity<ChargeStatus> {
+
+    @Column(name = "gateway_transaction_id")
+    private String gatewayTransactionId;
+
+    public ChargeTransactionEntity() {
+    }
+
+    public String getGatewayTransactionId() {
+        return gatewayTransactionId;
+    }
+
+    public void setGatewayTransactionId(String gatewayTransactionId) {
+        this.gatewayTransactionId = gatewayTransactionId;
+    }
+
+    public void setStatus(ChargeStatus status) {
+        if (this.status != null && !defaultTransitions().isValidTransition(this.status, status)) {
+            throw new InvalidStateTransitionException(this.status.getValue(), status.getValue());
+        }
+        this.status = status;
+    }
+
+    public static ChargeTransactionEntity from(ChargeEntity chargeEntity) {
+        ChargeTransactionEntity transactionEntity = new ChargeTransactionEntity();
+        transactionEntity.setGatewayTransactionId(chargeEntity.getGatewayTransactionId());
+        transactionEntity.setAmount(chargeEntity.getAmount());
+        transactionEntity.setStatus(ChargeStatus.fromString(chargeEntity.getStatus()));
+        transactionEntity.setOperation(TransactionOperation.CHARGE);
+
+        return transactionEntity;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
@@ -1,0 +1,64 @@
+package uk.gov.pay.connector.model.domain.transaction;
+
+import uk.gov.pay.connector.model.domain.AbstractEntity;
+import uk.gov.pay.connector.model.domain.PaymentRequestEntity;
+import uk.gov.pay.connector.model.domain.Status;
+
+import javax.persistence.Column;
+import javax.persistence.DiscriminatorColumn;
+import javax.persistence.Entity;
+import javax.persistence.EnumType;
+import javax.persistence.Enumerated;
+import javax.persistence.Inheritance;
+import javax.persistence.InheritanceType;
+import javax.persistence.JoinColumn;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "transactions")
+@Inheritance(strategy = InheritanceType.SINGLE_TABLE)
+@DiscriminatorColumn(name = "operation")
+public abstract class TransactionEntity<E extends Status> extends AbstractEntity {
+    @Column(name = "status")
+    @Enumerated(EnumType.STRING)
+    protected E status;
+    @Column(name = "amount")
+    private Long amount;
+    @Column(name = "operation")
+    @Enumerated(EnumType.STRING)
+    private TransactionOperation operation;
+    @ManyToOne
+    @JoinColumn(name = "payment_request_id", referencedColumnName = "id", updatable = false)
+    private PaymentRequestEntity paymentRequest;
+
+    public Long getAmount() {
+        return amount;
+    }
+
+    public void setAmount(Long amount) {
+        this.amount = amount;
+    }
+
+    public PaymentRequestEntity getPaymentRequest() {
+        return paymentRequest;
+    }
+
+    public void setPaymentRequest(PaymentRequestEntity paymentRequest) {
+        this.paymentRequest = paymentRequest;
+    }
+
+    public TransactionOperation getOperation() {
+        return operation;
+    }
+
+    public void setOperation(TransactionOperation operation) {
+        this.operation = operation;
+    }
+
+    public E getStatus() {
+        return status;
+    }
+
+    public abstract void setStatus(E status);
+}
--- a/src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
+++ b/src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
@@ -2,13 +2,11 @@ package uk.gov.pay.connector.resources;
 
 import com.fasterxml.jackson.annotation.JsonView;
 import com.google.common.collect.ImmutableList;
-import com.google.inject.persist.Transactional;
 import io.dropwizard.jersey.PATCH;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.dao.CardTypeDao;
 import uk.gov.pay.connector.dao.ChargeDao;
-import uk.gov.pay.connector.dao.ChargeEventDao;
 import uk.gov.pay.connector.model.ChargeResponse;
 import uk.gov.pay.connector.model.builder.PatchRequestBuilder;
 import uk.gov.pay.connector.model.domain.*;
@@ -21,19 +19,15 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 import java.net.URI;
-import java.time.ZonedDateTime;
 import java.util.Collections;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
-import static com.google.common.collect.Lists.newArrayList;
 import static javax.ws.rs.HttpMethod.GET;
 import static javax.ws.rs.HttpMethod.POST;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static uk.gov.pay.connector.model.FrontendChargeResponse.aFrontendChargeResponse;
 import static uk.gov.pay.connector.model.builder.PatchRequestBuilder.aPatchRequestBuilder;
-import static uk.gov.pay.connector.model.domain.ChargeStatus.CREATED;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
 import static uk.gov.pay.connector.resources.ApiPaths.*;
 import static uk.gov.pay.connector.resources.ApiValidators.validateChargePatchParams;
@@ -44,18 +38,16 @@ import static uk.gov.pay.connector.util.ResponseUtil.*;
 public class ChargesFrontendResource {
 
     private static final Logger logger = LoggerFactory.getLogger(ChargesFrontendResource.class);
-    private static final List<ChargeStatus> CURRENT_STATUSES_ALLOWING_UPDATE_TO_NEW_STATUS = newArrayList(CREATED, ENTERING_CARD_DETAILS);
+    private static final String NEW_STATUS = "new_status";
     private final ChargeDao chargeDao;
     private final ChargeService chargeService;
     private final CardTypeDao cardTypeDao;
-    private final ChargeEventDao chargeEventDao;
 
     @Inject
-    public ChargesFrontendResource(ChargeDao chargeDao, ChargeService chargeService, CardTypeDao cardTypeDao, ChargeEventDao chargeEventDao) {
+    public ChargesFrontendResource(ChargeDao chargeDao, ChargeService chargeService, CardTypeDao cardTypeDao) {
         this.chargeDao = chargeDao;
         this.chargeService = chargeService;
         this.cardTypeDao = cardTypeDao;
-        this.chargeEventDao = chargeEventDao;
     }
 
     @GET
@@ -102,36 +94,35 @@ public class ChargesFrontendResource {
     @Path(FRONTEND_CHARGE_STATUS_API_PATH)
     @Produces(APPLICATION_JSON)
     @JsonView(GatewayAccountEntity.Views.FrontendView.class)
-    @Transactional
     public Response updateChargeStatus(@PathParam("chargeId") String chargeId, Map newStatusMap) {
         if (invalidInput(newStatusMap)) {
-            return fieldsMissingResponse(ImmutableList.of("new_status"));
+            return fieldsMissingResponse(ImmutableList.of(NEW_STATUS));
         }
+
+        ChargeStatus newChargeStatus;
         try {
-            return updateStatus(chargeId, ChargeStatus.fromString(newStatusMap.get("new_status").toString()), Optional.empty());
+            newChargeStatus = ChargeStatus.fromString(newStatusMap.get(NEW_STATUS).toString());
         } catch (IllegalArgumentException e) {
             logger.error(e.getMessage(), e);
             return badRequestResponse(e.getMessage());
         }
+
+        if (!isValidStateTransition(newChargeStatus)) {
+            return getInvalidStatusResponse(chargeId, newChargeStatus);
+        }
+
+        return chargeService.updateFromInitialStatus(chargeId, newChargeStatus)
+                .map(chargeEntity -> Response.noContent().build())
+                .orElseGet(() -> getInvalidStatusResponse(chargeId, newChargeStatus));
+    }
+
+    private Response getInvalidStatusResponse(String chargeId, ChargeStatus newChargeStatus) {
+        return badRequestResponse("charge with id: " + chargeId +
+                " cannot be updated to the new status: " + newChargeStatus.getValue());
     }
 
     private boolean invalidInput(Map newStatusMap) {
-        return newStatusMap == null || newStatusMap.get("new_status") == null;
-    }
-
-    private Response updateStatus(String chargeId, ChargeStatus newChargeStatus, Optional<ZonedDateTime> generatedTime) {
-        if (!isValidStateTransition(newChargeStatus)) {
-            return badRequestResponse("charge with id: " + chargeId + " cant be updated to the new state: " + newChargeStatus.getValue());
-        }
-        return chargeDao.findByExternalId(chargeId)
-                .map(chargeEntity -> {
-                    if (CURRENT_STATUSES_ALLOWING_UPDATE_TO_NEW_STATUS.contains(ChargeStatus.fromString(chargeEntity.getStatus()))) {
-                        chargeEntity.setStatus(newChargeStatus);
-                        chargeEventDao.persistChargeEventOf(chargeEntity, generatedTime);
-                        return noContentResponse();
-                    }
-                    return badRequestResponse("charge with id: " + chargeId + " cant be updated to the new state: " + newChargeStatus.getValue());
-                }).get();
+        return newStatusMap == null || newStatusMap.get(NEW_STATUS) == null;
     }
 
     private boolean isValidStateTransition(ChargeStatus newChargeStatus) {
--- a/src/main/java/uk/gov/pay/connector/service/CancelServiceFunctions.java
+++ b/src/main/java/uk/gov/pay/connector/service/CancelServiceFunctions.java
@@ -4,6 +4,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.dao.ChargeDao;
 import uk.gov.pay.connector.dao.ChargeEventDao;
+import uk.gov.pay.connector.dao.PaymentRequestDao;
 import uk.gov.pay.connector.exception.ChargeNotFoundRuntimeException;
 import uk.gov.pay.connector.exception.ConflictRuntimeException;
 import uk.gov.pay.connector.exception.IllegalStateRuntimeException;
@@ -34,13 +35,14 @@ class CancelServiceFunctions {
 
     private static final Logger logger = LoggerFactory.getLogger(CancelServiceFunctions.class);
 
-    static TransactionalOperation<TransactionContext, ChargeEntity> changeStatusTo(ChargeDao chargeDao, ChargeEventDao chargeEventDao, String chargeId, ChargeStatus targetStatus, Optional<ZonedDateTime> generationTime) {
+    static TransactionalOperation<TransactionContext, ChargeEntity> changeStatusTo(ChargeDao chargeDao, ChargeEventDao chargeEventDao, String chargeId, ChargeStatus targetStatus, Optional<ZonedDateTime> generationTime, StatusUpdater statusUpdater) {
         return context -> chargeDao.findByExternalId(chargeId)
                 .map(chargeEntity -> {
                     logger.info("Charge status to update - charge_external_id={}, status={}, to_status={}",
                             chargeEntity.getExternalId(), chargeEntity.getStatus(), targetStatus);
                     chargeEntity.setStatus(targetStatus);
                     chargeEventDao.persistChargeEventOf(chargeEntity, generationTime);
+                    statusUpdater.updateChargeTransactionStatus(chargeEntity.getExternalId(), targetStatus);
                     return chargeEntity;
                 })
                 .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
@@ -49,10 +51,12 @@ class CancelServiceFunctions {
     static PreTransactionalOperation<TransactionContext, ChargeEntity> prepareForTerminate(ChargeDao chargeDao,
                                                                                            ChargeEventDao chargeEventDao,
                                                                                            String chargeId,
-                                                                                           StatusFlow statusFlow) {
+                                                                                           StatusFlow statusFlow,
+                                                                                           StatusUpdater statusUpdater) {
         return context -> chargeDao.findByExternalId(chargeId).map(chargeEntity -> {
+            ChargeStatus newStatus = statusFlow.getLockState();
             if (!chargeEntity.hasStatus(statusFlow.getTerminatableStatuses())) {
-                if (chargeEntity.hasStatus(statusFlow.getLockState())) {
+                if (chargeEntity.hasStatus(newStatus)) {
                     throw new OperationAlreadyInProgressRuntimeException(statusFlow.getName(), chargeId);
                 } else if (chargeEntity.hasStatus(ChargeStatus.AUTHORISATION_READY, AUTHORISATION_3DS_READY)) {
                     throw new ConflictRuntimeException(chargeEntity.getExternalId());
@@ -63,7 +67,8 @@ class CancelServiceFunctions {
 
                 throw new IllegalStateRuntimeException(chargeId);
             }
-            chargeEntity.setStatus(statusFlow.getLockState());
+            chargeEntity.setStatus(newStatus);
+            statusUpdater.updateChargeTransactionStatus(chargeEntity.getExternalId(), newStatus);
             GatewayAccountEntity gatewayAccount = chargeEntity.getGatewayAccount();
 
             logger.info("Card cancel request sent - charge_external_id={}, charge_status={}, account_id={}, transaction_id={}, amount={}, operation_type={}, provider={}, provider_type={}, locking_status={}",
@@ -75,9 +80,10 @@ class CancelServiceFunctions {
                     OperationType.CANCELLATION.getValue(),
                     gatewayAccount.getGatewayName(),
                     gatewayAccount.getType(),
-                    statusFlow.getLockState());
+                    newStatus);
 
             chargeEventDao.persistChargeEventOf(chargeEntity, Optional.empty());
+
             return chargeEntity;
         }).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
     }
--- a/src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
+++ b/src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
@@ -6,11 +6,13 @@ import io.dropwizard.setup.Environment;
 import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.connector.dao.ChargeDao;
 import uk.gov.pay.connector.dao.ChargeEventDao;
+import uk.gov.pay.connector.dao.PaymentRequestDao;
 import uk.gov.pay.connector.exception.ChargeNotFoundRuntimeException;
 import uk.gov.pay.connector.model.domain.Auth3dsDetails;
 import uk.gov.pay.connector.model.domain.ChargeEntity;
 import uk.gov.pay.connector.model.domain.ChargeStatus;
 import uk.gov.pay.connector.model.domain.GatewayAccountEntity;
+import uk.gov.pay.connector.model.domain.PaymentRequestEntity;
 import uk.gov.pay.connector.model.gateway.Auth3dsResponseGatewayRequest;
 import uk.gov.pay.connector.model.gateway.GatewayResponse;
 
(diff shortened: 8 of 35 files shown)


src/main/java/uk/gov/pay/connector/dao/PaymentRequestDao.java
```diff
@@ -14,4 +16,24 @@
...
+        String query = "SELECT p FROM PaymentRequestEntity p " +
+                "WHERE p.externalId = :externalId";
+
+        return entityManager.get()
+                .createQuery(query, PaymentRequestEntity.class)
+                .setParameter("externalId", externalId)
+                .getResultList().stream().findFirst();
+    }
+
+    public Optional<PaymentRequestEntity> updateChargeTransactionStatus(String externalId, ChargeStatus newChargeStatus) {
```
alexbishop1: Not sure this class is the right place for this method. It feels like updating the status of a transaction should be something that’s done via a `TransactionDao` (yes, I know that name is already used: it can be renamed) or something. I realise the fact that right now not everything in the `charges` table will have something equivalent in the `transactions` table complicates things.

We may perhaps want to go further and try centralising how we do status updates. Right now, it’s usually done by directly modifying the `ChargeEntity` and then persisting it. If we instead called a centralised method (maybe on `ChargeService`), we could have that do the updating of the `ChargeEntity` and, if applicable, the `TransactionEntity` in the same place. Later, we could add code to insert into the `TransactionEvents` table. Then, we want to stop writing to the `Charges` table, we’d only have to do the change in one place (more or less). And in eighteen months time when we’re changing how the database stores data again, we wouldn’t have to go find everywhere we change statuses again.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -150,7 +150,7 @@ public void setExternalId(String externalId) {
     }
 
     public void setStatus(ChargeStatus targetStatus) throws InvalidStateTransitionException {
-        if (stateTransitionsFor(getPaymentGatewayName()).isValidTransition(fromString(this.status), targetStatus)) {
+        if (defaultTransitions().isValidTransition(fromString(this.status), targetStatus)) {
```
alexbishop1, line 153: Why was this changed?

src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
```diff
@@ -91,7 +101,36 @@ public void setExternalId(String externalId) {
...
+        this.transactions.add(transactionEntity);
+        transactionEntity.setPaymentRequest(this);
+    }
+
+    public void setTransactions(List<TransactionEntity> transactions) {
+        this.transactions = transactions;
+        transactions.forEach(transactionEntity -> transactionEntity.setPaymentRequest(this));
+    }
+
+    public TransactionEntity getChargeTransaction() {
```
alexbishop1: Ideally this would be called something simple like `getCharge()` but maybe we should address this after we don’t have two things called “charge” in the system

src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
```diff
@@ -91,7 +101,36 @@ public void setExternalId(String externalId) {
...
+        return transactions.stream().filter(byChargeTransactions()).findFirst().orElseThrow(
```
alexbishop1: This scares me a little bit. It’s expected that people should do this, right?

```
if (paymentRequestEntity.hasChargeTransaction()) {
    paymentRequestEntity.getChargeTransaction().setStatus(newChargeStatus);
}
```

We could instead make this method return an `Optional` (with an `Optional.empty()` instead of an exception) and expect people to use it like this:

```
paymentRequestEntity.getChargeTransaction().ifPresent(transaction -> transaction.setStatus(newChargeStatus));
```

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
```diff
@@ -0,0 +1,118 @@
...
+import uk.gov.pay.connector.exception.InvalidStateTransitionException;
+import uk.gov.pay.connector.model.domain.*;
+
+import javax.persistence.*;
+
+import static uk.gov.pay.connector.model.domain.PaymentGatewayStateTransitions.defaultTransitions;
+
+@Entity
+@Table(name = "transactions")
+public class TransactionEntity extends AbstractEntity {
```
alexbishop1: I may have missed some of the discussion but I thought there was a general consensus that we would use single table inheritance here and have `TransactionEntity` be abstract (probably) with a concrete `ChargeTransactionEntity` (or whatever we decide to call it) and eventually a `RefundTransactionEntity`, each exposing just the fields they care about.

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
```diff
@@ -0,0 +1,118 @@
...
+        TransactionEntity transactionEntity = new TransactionEntity();
+        transactionEntity.setGatewayTransactionId(chargeEntity.getGatewayTransactionId());
+        transactionEntity.setAmount(chargeEntity.getAmount());
+        transactionEntity.setStatus(ChargeStatus.fromString(chargeEntity.getStatus()));
+        transactionEntity.setOperation(TransactionOperation.CHARGE);
+
+        return transactionEntity;
+    }
+
+    private void checkValidStateTransition(ChargeStatus newChargeStatus) {
```
alexbishop1: I’d be quite tempted to roll this in to the `setStatus(ChargeStatus)` method

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -102,36 +94,35 @@ public Response patchCharge(@PathParam("chargeId") String chargeId, Map<String,
...
+                " cannot be updated to the new status: " + newChargeStatus.getValue());
     }
 
-    private Response updateStatus(String chargeId, ChargeStatus newChargeStatus, Optional<ZonedDateTime> generatedTime) {
-        if (!isValidStateTransition(newChargeStatus)) {
-            return badRequestResponse("charge with id: " + chargeId + " cant be updated to the new state: " + newChargeStatus.getValue());
-        }
-        return chargeDao.findByExternalId(chargeId)
-                .map(chargeEntity -> {
-                    if (CURRENT_STATUSES_ALLOWING_UPDATE_TO_NEW_STATUS.contains(ChargeStatus.fromString(chargeEntity.getStatus()))) {
```
alexbishop1, line 128: Why has this check been removed?

src/main/java/uk/gov/pay/connector/service/CancelServiceFunctions.java
```diff
@@ -78,6 +81,8 @@
                     statusFlow.getLockState());
 
             chargeEventDao.persistChargeEventOf(chargeEntity, Optional.empty());
+            paymentRequestDao.updateChargeTransactionStatus(chargeEntity.getExternalId(), statusFlow.getLockState());
```
alexbishop1: I think it would be clearer if the `statusFlow.getLockState()` was pulled out into a local variable (called `newStatus`, say) declared just before the `chargeEntity.setStatus(ChargeStatus)` call (with the call to that method changed to pass in `newStatus`). And maybe this line should be moved up. It’s not immediately clear that this is setting the same value on the `ChargeEntity` and the `TransactionEntity`.

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
```diff
@@ -6,11 +6,9 @@
 import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.connector.dao.ChargeDao;
 import uk.gov.pay.connector.dao.ChargeEventDao;
+import uk.gov.pay.connector.dao.PaymentRequestDao;
 import uk.gov.pay.connector.exception.ChargeNotFoundRuntimeException;
-import uk.gov.pay.connector.model.domain.Auth3dsDetails;
-import uk.gov.pay.connector.model.domain.ChargeEntity;
-import uk.gov.pay.connector.model.domain.ChargeStatus;
-import uk.gov.pay.connector.model.domain.GatewayAccountEntity;
+import uk.gov.pay.connector.model.domain.*;
```
alexbishop1: You’re going against the proposed GDS Java style guide with wildcard imports

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
```diff
@@ -68,11 +66,17 @@ public ChargeEntity preOperation(String chargeId, Auth3dsDetails auth3DsDetails)
             metricRegistry.counter(String.format("gateway-operations.%s.%s.%s.authorise-3ds.result.%s", account.getGatewayName(), account.getType(), account.getId(), status.toString())).inc();
 
             chargeEntity.setStatus(status);
+            Optional<PaymentRequestEntity> paymentRequestEntity = paymentRequestDao.updateChargeTransactionStatus(chargeEntity.getExternalId(), status);
```
alexbishop1: Feels a little funky that this method both changes something and returns something else

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -39,33 +36,46 @@ public CardAuthoriseService(ChargeDao chargeDao,
...
         this.auth3dsDetailsFactory = auth3dsDetailsFactory;
         this.card3dsDao = card3dsDao;
+        this.paymentRequestDao = paymentRequestDao;
     }
 
     @Transactional
     public ChargeEntity preOperation(String chargeId, AuthCardDetails authCardDetails) {
 
-        return chargeDao.findByExternalId(chargeId).map(chargeEntity -> {
+        return chargeDao.findByExternalId(chargeId).map((ChargeEntity chargeEntity) -> {
```
alexbishop1: Unnecessary cast?

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -39,33 +36,46 @@ public CardAuthoriseService(ChargeDao chargeDao,
...
+                Optional<PaymentRequestEntity> paymentRequestEntity = paymentRequestDao.findByExternalId(chargeEntity.getExternalId());
+
+                getPaymentProviderFor(chargeEntity).generateTransactionId().ifPresent(transactionIdValue -> {
+                    chargeEntity.setGatewayTransactionId(transactionIdValue);
+                    paymentRequestEntity.ifPresent(paymentRequest -> {
+                        if (paymentRequest.hasChargeTransaction()) {
+                            paymentRequest.getChargeTransaction().setGatewayTransactionId(transactionIdValue);
+                        }
+                    });
+                });
```
alexbishop1: This is basically a duplicate of the new code in `Card3dsResponseAuthoriseService` and below. I know it’s going to be fairly short-lived but it might be worth separating it out.

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -18,19 +19,25 @@
...
 import static uk.gov.pay.connector.model.ChargeResponse.aChargeResponseBuilder;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.CREATED;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
 import static uk.gov.pay.connector.model.domain.NumbersInStringsSanitizer.sanitize;
 import static uk.gov.pay.connector.resources.ApiPaths.CHARGE_API_PATH;
 import static uk.gov.pay.connector.resources.ApiPaths.REFUNDS_API_PATH;
 
 public class ChargeService {
 
+    private static final List<ChargeStatus> CURRENT_STATUSES_ALLOWING_UPDATE_TO_NEW_STATUS = newArrayList(CREATED, ENTERING_CARD_DETAILS);
```
alexbishop1, line 39: So that’s where the code from `ChargeFrontendResource` went! I’m not sure it really belongs here: as I understand it, the restrictions are supposed to only be enforced if processing a status update request from frontend, so `ChargeFrontendResource` seems like the right place to do this.

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -92,6 +105,21 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
                 .orElseGet(Optional::empty);
     }
 
+    @Transactional
+    public Optional<ChargeEntity> updateStatus(String externalId, ChargeStatus newChargeStatus) {
```
alexbishop1: If this method is going to stay here, we should probably rename it to make clear the restricted circumstances in which it’s appropriate to call it

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
```diff
@@ -1,35 +1,101 @@
...
     public void shouldCreateAPaymentRequest() throws Exception {
+        PaymentRequestEntity paymentRequestEntity = aValidPaymentRequestEntity()
+                .withGatewayAccountEntity(gatewayAccount)
+                .build();
+
+        paymentRequestDao.persist(paymentRequestEntity);
+
+        assertThat(paymentRequestEntity.getId(), is(notNullValue()));
+        // Ensure always max precision is being millis
+        assertThat(paymentRequestEntity.getCreatedDate().getNano() % 1000000, is(0));
```
alexbishop1, line 52: Why do we care about this?

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
```diff
@@ -1,35 +1,101 @@
...
+        paymentRequestDao.updateChargeTransactionStatus(paymentRequestEntity.getExternalId(),
+                ChargeStatus.ENTERING_CARD_DETAILS);
+        final Optional<PaymentRequestEntity> byExternalId =
+                paymentRequestDao.findByExternalId(paymentRequestEntity.getExternalId());
+
+        assertThat(byExternalId.get().getTransactions().get(0).getStatus(), is(ChargeStatus.ENTERING_CARD_DETAILS));
+    }
+
+    @Test(expected = InvalidStateTransitionException.class)
+    public void shouldThrowException_whenUpdatingToInvalidState() throws Exception {
```
alexbishop1: It would be nice if the initial state was explicitly, er, stated in this test

src/test/java/uk/gov/pay/connector/model/domain/PaymentRequestEntityFixture.java
```diff
@@ -38,4 +43,9 @@ public PaymentRequestEntityFixture withGatewayAccountEntity(GatewayAccountEntity
         this.gatewayAccountEntity = gatewayAccount;
         return this;
     }
+
+    public PaymentRequestEntityFixture withTransactions(List<TransactionEntity> transactions) {
```
alexbishop1: If you made this varargs instead (`withTransactions(TransactionEntity... transactions)`), it would make the calling code look nicer

src/test/java/uk/gov/pay/connector/model/domain/PaymentRequestEntityTest.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Test
+    public void shouldReturnChargeTransactionWhenThereIsOnlyAChargeTransaction() throws Exception {
+        TransactionEntity chargeTransaction = paymentRequestEntity.getChargeTransaction();
+        assertThat(chargeTransaction, is(expectedChargeTransaction));
+    }
+
+    @Test
+    public void shouldReturnChargeTransactionWhenThereIsManyTransactions() throws Exception {
```
alexbishop1, line 39: Nice test

src/test/java/uk/gov/pay/connector/service/Card3dsResponseAuthServiceTest.java
```diff
@@ -92,12 +90,17 @@ public void doAuthorise_shouldRespondAuthorisationSuccess() throws Exception {
         setupPaymentProviderMock(charge.getGatewayTransactionId(), AuthoriseStatus.AUTHORISED, null, argumentCaptor);
 
         when(mockedProviders.byName(charge.getPaymentGatewayName())).thenReturn(mockedPaymentProvider);
+        PaymentRequestEntity paymentRequestEntity = PaymentRequestEntityFixture.aValidPaymentRequestEntity().build();
```
alexbishop1: Since there’s another test that uses the result of `PaymentRequestEntityFixture.aValidPaymentRequestEntity().build()`, we could move this to be a field

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
```diff
@@ -129,6 +135,10 @@ public void doAuthorise_shouldRespondAuthorisationSuccess() throws Exception {
         verify(mockedChargeEventDao).persistChargeEventOf(charge, Optional.empty());
         assertThat(charge.get3dsDetails(), is(nullValue()));
         assertThat(charge.getCardDetails(), is(notNullValue()));
+        assertThat(paymentRequest.getChargeTransaction().getGatewayTransactionId(), is(TRANSACTION_ID));
+
+        verify(mockPaymentRequestDao).updateChargeTransactionStatus(charge.getExternalId(), AUTHORISATION_READY);
+        verify(mockPaymentRequestDao).updateChargeTransactionStatus(charge.getExternalId(), AUTHORISATION_SUCCESS);
```
alexbishop1: Use in-order verification here?

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
```diff
@@ -126,6 +126,8 @@ public void doCapture_shouldCaptureAChargeForANonSandboxAccount() throws Excepti
         ArgumentCaptor<ChargeEntity> chargeEntityCaptor = ArgumentCaptor.forClass(ChargeEntity.class);
 
         verify(mockedChargeEventDao).persistChargeEventOf(chargeEntityCaptor.capture(), eq(Optional.empty()));
+        verify(mockPaymentRequestDao).updateChargeTransactionStatus(charge.getExternalId(), CAPTURE_READY);
+        verify(mockPaymentRequestDao).updateChargeTransactionStatus(charge.getExternalId(), CAPTURE_SUBMITTED);
```
alexbishop1: Another possible in-order verification opportunity

src/test/java/uk/gov/pay/connector/service/ChargeCancelServiceTest.java
```diff
@@ -57,9 +57,12 @@
     @Mock
     private PaymentProvider mockedPaymentProvider;
 
+    @Mock
+    private PaymentRequestDao mockedPaymentRequestDao;
```
alexbishop1: The prefix `mock` (as opposed to `mocked` seems to be more prevalent in this file)

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -320,6 +324,47 @@ public void shouldNotFindAChargeWhenNoChargeForChargeIdAndAccountId() {
...
+    public void shouldUpdateTransactionStatus_whenUpdatingChargeStatus()  throws Exception {
+
+        service.create(CHARGE_REQUEST, GATEWAY_ACCOUNT_ID, mockedUriInfo);
+
+        ArgumentCaptor<ChargeEntity> chargeEntityArgumentCaptor = forClass(ChargeEntity.class);
+        verify(mockedChargeDao).persist(chargeEntityArgumentCaptor.capture());
+
+        ChargeEntity createdChargeEntity = chargeEntityArgumentCaptor.getValue();
+
+        when(mockedChargeDao.findByExternalId(anyString()))
```
alexbishop1: You know the specific external ID this will be called with

alexbishop1 (changes requested):
Looks like it should do the job but I haven’t exhaustively checked for everywhere we update the status of a payment and formally proved the use of transactions (in the database sense).

I do think it may be worth looking at centralising status updates so that they all go through one place and we can handle the updating in both places (and maybe even appending to the event history tables) there. It would make future changes easier (not just when we remove the old `charges` table but also if we decide to do consistent logging or fire metrics at Graphite for every status change). I suggested adding this to `ChargeService` but I wouldn’t be opposed to having a new `StatusUpdater` that deals solely with updating statuses: it’s a major part of what connector does, after all (whether or not connector should do this is a whole other discussion!).

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
chrisgrimble: This gets the allowed transaction transitions. We used to need to pass in the gateway name to find this out but it was not being used so we removed the argument and changed the method name. It is statically imported from PaymentGatewayStateTransitions.

src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
chrisgrimble: Yep think this will cause confusion naming it that now.

src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
chrisgrimble: This will be reverted once the transactions table has been backfilled. Did look at returning an optional but the change set was getting massive so decided this would be easier to revert in a few days time once the backfill has been done.

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
chrisgrimble: Ok

src/main/java/uk/gov/pay/connector/service/CancelServiceFunctions.java
chrisgrimble: Ok, except pulled out newStatus at the start of the method, as it is used earlier on and does make the code easier to read as newStatus.

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
chrisgrimble: Ok

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
chrisgrimble: Ok split into two methods

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
chrisgrimble: Yep

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
chrisgrimble: ok

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
chrisgrimble: Ok

src/test/java/uk/gov/pay/connector/model/domain/PaymentRequestEntityFixture.java
chrisgrimble: Ok

src/test/java/uk/gov/pay/connector/model/domain/PaymentRequestEntityTest.java
chrisgrimble: Thanks

src/test/java/uk/gov/pay/connector/service/Card3dsResponseAuthServiceTest.java
chrisgrimble: Ok

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
chrisgrimble: Ok didn't know you could do that on a single mock but guess there is no reason not too.

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
chrisgrimble: Ok

src/test/java/uk/gov/pay/connector/service/ChargeCancelServiceTest.java
chrisgrimble: Ok

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
chrisgrimble: Ok

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
chrisgrimble: Done

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
chrisgrimble: Done

src/main/java/uk/gov/pay/connector/service/ChargeService.java
chrisgrimble: Decide we would leave this here but the method where this is used is now updateFromInitialStatus not just updateStatus

src/main/java/uk/gov/pay/connector/service/ChargeService.java
chrisgrimble: Decide we would leave this here but the method where this is used is now updateFromInitialStatus not just updateStatus

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
chrisgrimble: Was already like that. Not too sure why this was done but don't want to be changing other things at the moment.

src/main/java/uk/gov/pay/connector/dao/PaymentRequestDao.java
chrisgrimble: Have moved the status update logic to StatusUpdater object. Looked at also updating the ChargeEntity in there too but was causing too many changes to our current code which we were trying not to change. Also would like to keep the StatusUpdater just dealing with the new objects (PaymentRequestEntity, TransactionEntity etc) not the old ones (ChargesEntity).

alexbishop1: Looks like this method still exists in `PaymentRequestDao` even though there is now a `StatusUpdater`

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
alexbishop1: At some point (not now), we should probably stop pretending that different payment gateways have different state transitions and merge `StateTransitions` and `DefaultStateTransitions`

src/main/java/uk/gov/pay/connector/model/domain/transaction/ChargeTransactionEntity.java
```diff
@@ -0,0 +1,44 @@
...
+import uk.gov.pay.connector.exception.InvalidStateTransitionException;
+import uk.gov.pay.connector.model.domain.*;
+
+import javax.persistence.*;
+
+import static uk.gov.pay.connector.model.domain.PaymentGatewayStateTransitions.defaultTransitions;
+
+@Entity
+@DiscriminatorValue(value = "CHARGE")
+public class ChargeTransactionEntity extends TransactionEntity<ChargeStatus> {
```
alexbishop1, line 12: This is much more concise now. Good job!

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -43,19 +40,19 @@
 @Path("/")
 public class ChargesFrontendResource {
 
-    private static final Logger logger = LoggerFactory.getLogger(ChargesFrontendResource.class);
     private static final List<ChargeStatus> CURRENT_STATUSES_ALLOWING_UPDATE_TO_NEW_STATUS = newArrayList(CREATED, ENTERING_CARD_DETAILS);
```
alexbishop1: This can go now

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -102,36 +99,35 @@ public Response patchCharge(@PathParam("chargeId") String chargeId, Map<String,
...
-            return updateStatus(chargeId, ChargeStatus.fromString(newStatusMap.get("new_status").toString()), Optional.empty());
-        } catch (IllegalArgumentException e) {
-            logger.error(e.getMessage(), e);
-            return badRequestResponse(e.getMessage());
+
+        String chargeStatusString = newStatusMap.get(NEW_STATUS).toString();
+        if (!ChargeStatus.isValid(chargeStatusString)) {
+            String msg = "charge status not recognized: " + chargeStatusString;
+            logger.error(msg);
+            return badRequestResponse(msg);
```
alexbishop1: We’re repeating the error message here (it’s the `IllegalArgumentException` message from `ChargeStatus.fromString(String)`): I think wrapping the `ChargeStatus.fromString(String)` in a try/catch and `return`ing `badRequestResponse(e.getMessage())` (closer to how it behaved before would be fine)

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
```diff
@@ -68,11 +72,18 @@ public ChargeEntity preOperation(String chargeId, Auth3dsDetails auth3DsDetails)
...
 
             if (StringUtils.isBlank(transactionId)) {
                 logger.warn("Auth3DSDetails authorisation response received with no transaction id. -  charge_external_id={}", chargeId);
             } else {
                 chargeEntity.setGatewayTransactionId(transactionId);
+                paymentRequestEntity.ifPresent(paymentRequest -> {
+                    if (paymentRequest.hasChargeTransaction()) {
+                        paymentRequest.getChargeTransaction().setGatewayTransactionId(transactionId);
+                    }
+                });
```
alexbishop1: Can we call the `setGatewayTransactionId(ChargeEntity, String, Optional<PaymentRequestEntity>)` superclass method to reduce duplication?

src/main/java/uk/gov/pay/connector/service/CardAuthoriseBaseService.java
```diff
@@ -60,6 +63,15 @@ public GatewayResponse doAuthorise(String chargeId, T gatewayAuthRequest) {
...
     }
 
+    protected void setGatewayTransactionId(ChargeEntity chargeEntity, String transactionId, Optional<PaymentRequestEntity> paymentRequestEntity) {
+        chargeEntity.setGatewayTransactionId(transactionId);
+        paymentRequestEntity.ifPresent(paymentRequest -> {
+            if (paymentRequest.hasChargeTransaction()) {
+                paymentRequest.getChargeTransaction().setGatewayTransactionId(transactionId);
+            }
+        });
+    }
```
alexbishop1, line 73: Again, can we leverage the superduper class here?

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
alexbishop1: So… going to do that? 😃

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -0,0 +1,26 @@
+package uk.gov.pay.connector.service;
+
+import com.google.inject.persist.Transactional;
+import uk.gov.pay.connector.dao.PaymentRequestDao;
+import uk.gov.pay.connector.model.domain.ChargeStatus;
+
+import javax.inject.Inject;
+
+ @Transactional
```
alexbishop1: Spurious leading space here

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -0,0 +1,26 @@
...
+public class StatusUpdater {
+
+    private PaymentRequestDao paymentRequestDao;
```
alexbishop1: This can be `final`

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -0,0 +1,26 @@
...
+    public StatusUpdater(PaymentRequestDao paymentRequestDao) {
+        this.paymentRequestDao = paymentRequestDao;
+    }
+
+    public  void updateChargeTransactionStatus(String externalId, ChargeStatus newChargeStatus) {
+        paymentRequestDao.findByExternalId(externalId).ifPresent(paymentRequestEntity -> {
+            if (paymentRequestEntity.hasChargeTransaction()) {
+                paymentRequestEntity.getChargeTransaction().setStatus(newChargeStatus);
+            }
+        });
```
alexbishop1, line 24: Could do the below but maybe it’s taking things a bit far 😃

```
paymentRequestDao.findByExternalId(externalId)
    .filter(PaymentRequestEntity::hasChargeTransaction)
    .ifPresent(paymentRequestEntity -> paymentRequestEntity.getChargeTransaction().setStatus(newChargeStatus));
```

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
```diff
@@ -1,35 +1,103 @@
...
+        assertThat(paymentRequestEntity.getId(), is(notNullValue()));
+        // Ensure always max precision is being millis
+        assertThat(paymentRequestEntity.getCreatedDate().getNano() % 1000000, is(0));
+        paymentRequestEntity.getTransactions().forEach(
+                transactionEntity -> assertThat(transactionEntity.getId(), is(notNullValue()))
+        );
+    }
+
+    @Test
+    public void shouldUpdateTransactionStatus() throws Exception {
```
alexbishop1, line 59: I wonder if this test (and the two below) belong here now: in the future, we may be able to do some status changes without going via a `PaymentRequest` (e.g. if we get a notification and use the gateway transaction ID to retrieve the charge). Probably worth revisiting when that’s the case.

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
```diff
@@ -40,9 +108,7 @@ public void shouldCreateAPaymentRequest() throws Exception {
 
         paymentRequestDao.persist(paymentRequestEntity);
 
-        assertThat(paymentRequestEntity.getId(), is(notNullValue()));
-        // Ensure always max precision is being millis
-        assertThat(paymentRequestEntity.getCreatedDate().getNano() % 1000000, is(0));
+        assertThat(paymentRequestEntity.getChargeTransaction().getId(), is(IsNull.notNullValue()));
```
alexbishop1: You can static import `notNullValue()` for a readability win

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
```diff
@@ -162,6 +166,10 @@ public void doCapture_shouldCaptureAChargeForASandboxAccount() throws Exception
         verify(mockedChargeEventDao, times(2)).persistChargeEventOf(chargeEntityCaptor.capture(), bookingDateCaptor.capture());
         assertThat(chargeEntityCaptor.getValue().getStatus(), is(CAPTURED.getValue()));
 
+        verify(mockStatusUpdater).updateChargeTransactionStatus(charge.getExternalId(), CAPTURE_READY);
```
alexbishop1: More in-order verification opportunities

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
```diff
@@ -263,6 +271,8 @@ public void doCapture_shouldSetChargeStatusToCaptureApprovedRetryOnError_whenPro
         inOrder.verify(chargeSpy).setStatus(CAPTURE_APPROVED_RETRY);
 
         verify(mockedChargeEventDao).persistChargeEventOf(chargeSpy, Optional.empty());
+        verify(mockStatusUpdater).updateChargeTransactionStatus(charge.getExternalId(), CAPTURE_READY);
+        verify(mockStatusUpdater).updateChargeTransactionStatus(charge.getExternalId(), CAPTURE_APPROVED_RETRY);
```
alexbishop1: Further in-order verification opportunities

src/test/java/uk/gov/pay/connector/service/ChargeExpiryServiceTest.java
```diff
@@ -53,9 +54,15 @@
     @Mock
     private WorldpayCancelResponse mockWorldpayCancelResponse;
 
+    @Mock
+    private PaymentRequestDao mockPaymentRequestDao;
```
alexbishop1: This is never used

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -320,6 +327,44 @@ public void shouldNotFindAChargeWhenNoChargeForChargeIdAndAccountId() {
...
+        PaymentRequestEntity paymentRequestEntity = argumentCaptor.getValue();
+        assertThat(paymentRequestEntity.getTransactions().size(), is(1));
+        TransactionEntity transactionEntity = paymentRequestEntity.getTransactions().get(0);
+        assertThat(transactionEntity.getAmount(), is(100L));
+        assertThat(transactionEntity.getStatus(), is(ChargeStatus.CREATED));
+        assertThat(transactionEntity.getOperation(), is(TransactionOperation.CHARGE));
+    }
+
+    @Test
+    public void shouldUpdateTransactionStatus_whenUpdatingChargeStatus()  throws Exception {
```
alexbishop1: Please rename this to make it clear it’s using the narrow `ChargeService.updateFromInitialStatus(String, ChargeStatus)` method

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+@RunWith(MockitoJUnitRunner.class)
+public class StatusUpdaterTest {
+    @Mock
+    private PaymentRequestDao mockPaymentRequestDao;
+    private String someExternalId;
+
+    @Before
+    public void setUp() throws Exception {
+        someExternalId = "someExternalId";
```
alexbishop1: Just inline this into the field declaration (which you can make into a fine SCREAMING_SNAKE_CASE constant)

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @Test
+    public void updatesChangeTransactionStatus() throws Exception {
+        PaymentRequestEntity paymentRequest = PaymentRequestEntityFixture.aValidPaymentRequestEntity()
+                .withTransactions(aChargeTransactionEntity().withStatus(ChargeStatus.CREATED).build())
+                .build();
+        when(mockPaymentRequestDao.findByExternalId(someExternalId)).thenReturn(Optional.of(paymentRequest));
+
+        ChargeStatus newChargeStatus = ChargeStatus.ENTERING_CARD_DETAILS;
```
alexbishop1: I’m not convinced having this variable is really buying us anything; I think it will be clear what’s going on if we just repeat `ENTERING_CARD_DETAILS` each time

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
```diff
@@ -0,0 +1,62 @@
...
+        new StatusUpdater(mockPaymentRequestDao).updateChargeTransactionStatus(someExternalId, newChargeStatus);
```
alexbishop1: I’m not going to insist on this but I usually like to have the class under test as a field

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void doesNotUpdatesChangeTransactionStatusIfChargeTransactionDoesNotExist() throws Exception {
+        PaymentRequestEntity paymentRequest = mock(PaymentRequestEntity.class);
+        when(mockPaymentRequestDao.findByExternalId(someExternalId)).thenReturn(Optional.of(paymentRequest));
+        when(paymentRequest.hasChargeTransaction()).thenReturn(false);
+
+        ChargeStatus newChargeStatus = ChargeStatus.ENTERING_CARD_DETAILS;
+        new StatusUpdater(mockPaymentRequestDao).updateChargeTransactionStatus(someExternalId, newChargeStatus);
+
+        verify(paymentRequest).hasChargeTransaction();
+        verifyNoMoreInteractions(paymentRequest);
```
alexbishop1: I’m generally in agreement with the notion that stubbing doesn’t usually need to be verified, so I’d remove these two lines

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
```
alexbishop1, line 53: New line at the end of the file, please

alexbishop1 (changes requested):
Almost there — just a handful of fairly minor things and then I think it should be good to go (if you look me in the eye and promise you’ve tracked down everywhere we change charge statuses and are convinced the database transactions are sound!)

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
chrisgrimble: Yep added the method and then don't appear to have used it.

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
chrisgrimble: Done

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
chrisgrimble: Done

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
chrisgrimble: Codes going to go as soon as we have back filled so won't do but yep probably should have done it that way.

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
chrisgrimble: Ok

src/test/java/uk/gov/pay/connector/it/dao/PaymentRequestDaoITest.java
chrisgrimble: Ok its notNullValue imported by Matchers but don't see why that can't be used here.

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
chrisgrimble: Done

src/test/java/uk/gov/pay/connector/service/CardCaptureServiceTest.java
chrisgrimble: Done

src/test/java/uk/gov/pay/connector/service/ChargeExpiryServiceTest.java
chrisgrimble: Done

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
chrisgrimble: Done

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
chrisgrimble: Have done as it can be a constant. 

At some time we should chat about the way we initialise things for tests. Generally much clearer if we just do it in one place and the only place you can initialise everything is in a setup method.

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
chrisgrimble: Done

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
chrisgrimble: Ok haven't as I'm not sure I would here

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
chrisgrimble: If we don't do this then we being explicit about the paymentRequets' charge transaction being updated

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
chrisgrimble: Done

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
chrisgrimble: done

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
chrisgrimble: Done

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
chrisgrimble: Done

src/main/java/uk/gov/pay/connector/service/CardAuthoriseBaseService.java
chrisgrimble: done

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -122,10 +125,15 @@ private boolean cardBrandRequires3ds(String cardBrand) {
...
 
+            statusUpdater.updateChargeTransactionStatus(chargeEntity.getExternalId(), status);
+            Optional<PaymentRequestEntity> paymentRequestEntity = paymentRequestDao.findByExternalId(chargeEntity.getExternalId());
             if (StringUtils.isBlank(transactionId)) {
                 logger.warn("AuthCardDetails authorisation response received with no transaction id. -  charge_external_id={}", chargeEntity.getExternalId());
             } else {
                 chargeEntity.setGatewayTransactionId(transactionId);
+                paymentRequestEntity.ifPresent(paymentRequest -> {
+                    paymentRequest.getChargeTransaction().setGatewayTransactionId(transactionId);
+                });
```
alexbishop1: I think we can use the superclass `setGatewayTransactionId` method here

alexbishop1 (changes requested):
One final thing…

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
chrisgrimble: Done

alexbishop1 (approved):
We made it in the end!

This is a very important step in our data restructuring, good work!

mrlumbu:
🎉

PR #487, opened 2017-11-14
PP-2858 Migrate In batches

chrisgrimble:
Split the migrate payment_requests in batches of 10000

Do not merge for now


src/main/java/uk/gov/pay/connector/healthcheck/DatabaseHealthCheck.java
```diff
@@ -84,21 +84,21 @@ protected Result check() throws Exception {
...
+//                    longDatabaseStatsMap.put(key, resultSet.getLong(key));
+//                }
+//                for (String key : doubleDatabaseStatsMap.keySet()) {
+//                    doubleDatabaseStatsMap.put(key, resultSet.getDouble(key));
+//                }
+//            }
+//            statsHealthy = 1;
+//        } catch (SQLException e) {
+//            statsHealthy = 0;
+//        }
```
bazbremner, line 101: Rather than commenting out code, you should delete it - quite obviously we have source control if we want to retrieve removed code in future.

bazbremner (changes requested):

chrisgrimble:
No need to split this like this. Was only a branch for testing things out. Created PR to get a build on Jenkins. Closing.

PR #488, opened 2017-11-14
PP-1797 Add Java EditorConfig from the GDS Way

alexbishop1:
The GDS Way now specifies [EditorConfig](http://editorconfig.org/) properties to make compatible tooling (such as IntelliJ IDEA) automatically adhere to some of the guidelines in the Java style guide. This pull request adds a `.editorconfig` file containing those properties.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root: true
+
+[*.java]
+indent_style: space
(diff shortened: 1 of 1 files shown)


DanailMinchev (approved):
AWESOME! 👍

PR #491, opened 2017-11-16
Revert "PP-2777 create table transaction"

chrisgrimble:
Reverts alphagov/pay-connector#484

Transition of states on Transactions is causing exceptions to be thrown reverting until we have worked out why.

Also reverts the Docker base image upgrade, superseding alphagov/pay-connector#490, because otherwise the pipeline is broken.


chrisgrimble:
Closing as we now have a fix for this rather than rolling back. Will raise a separate request for 

Also reverts the Docker base image upgrade, superseding alphagov/pay-connector#490, because otherwise the pipeline is broken.

PR #492, opened 2017-11-16
PP-2777 Create transaction table, fix exception when changing status

chrisgrimble:
Were getting exceptions with invalid state transitions on transactions.
As we are dual writing the state to charges and transactions and the
change to charges was successful, only log an warning if the transition
on transactions is invalid. We think these were due to some instances
being upgraded before others so transaction status were being updated
on some instances but not others. If your requests came into a
number of instances your transaction status may not get updated.

- Change throwing an exception when doing an invalid transition between
states on transactions to logging a warning.
- Add logging whenever we try and change a transaction status.
- Improve logging when we throw an exception in the CardCaptureProcess
so we knoe the external id of the charge causing the exception.
- Removed test from PaymentRequestDaoITest that checked changing
between invalid states threw an exception. This should probably should
have been moved to TransactionEntityTest as the code had moved to
TransactionEntity, but now is being removed from there.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.model.domain;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


georgeracu (approved):
Looks good

PR #493, opened 2017-11-16
PP-2777 Create transaction table, drop transactions

chrisgrimble:
We rolled back a release where some transaction status did not stay in
sync with charge status. Delete everything in transactions to clear
these out. This is safe to do as we are dual writing to charges and
transactions but only using charges at the moment. They will then get
populated later when we back fill transactions.


chrisgrimble:
Not going to do this like this now so closing PR.

PR #494, opened 2017-11-17
PP-2756 add notify_settings column to gateway_accounts table

SandorArpa:
Add a new, json type column to store the api key and template id in a json format for custom branded notifications.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -930,4 +930,12 @@
             </column>
         </createTable>
     </changeSet>
+
+    <changeSet id="add notify settings column to gateway_accounts table" author="">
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -930,4 +930,12 @@
             </column>
         </createTable>
     </changeSet>
+
+    <changeSet id="add notify settings column to gateway_accounts table" author="">
+        <addColumn tableName="gateway_accounts">
+            <column name="notify_settings" type="json" defaultValue="{}">
+                <constraints nullable="false" unique="false"/>
```
kakumara: it should be `nullable=true`

kakumara (approved):
LGTM

PR #496, opened 2017-11-21
PP-2777 Log status changes

chrisgrimble:
We add logging to show the changes between statuses on Transactions and
Charges but were logging the status after we had updated it. This
corrects that so it is not showing the status transition in the logs.


alexbishop1 (approved):
:man_facepalming:

PR #497, opened 2017-11-21
PP-2779 Create transaction_events table

chrisgrimble:
Added new transaction_events table which will need to be deployed
before we start using it.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -938,4 +938,42 @@
             </column>
         </addColumn>
     </changeSet>
+
+    <changeSet id="create table transaction events" author="">
+        <createTable tableName="transaction_events">
+            <column name="id" type="bigserial" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="transaction_id" type="bigint">
+                <constraints foreignKeyName="fk__transactions_id"
+                             referencedTableName="transactions"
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -938,4 +938,39 @@
...
+            </column>
+            <column name="transaction_id" type="bigint">
+                <constraints foreignKeyName="fk__transactions_id"
+                             referencedTableName="transactions"
+                             referencedColumnNames="id" nullable="false"/>
+            </column>
+            <column name="status" type="text">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created_date" type="timestamp without timezone"
```
alexbishop1: To avoid confusion, I think it’s best to stick to `updated` as the column name (that’s what we use for `charge_events`)

alexbishop1 (changes requested):
What’s there looks OK but I think we also want a `gateway_event_date` column like in `charge_events`.

This is because for capture, the notifications sent by some gateways contains a timestamp that dates to before the receipt of the notification.

I recall we did discuss calling this `capture_date` but there’s the possibility we may do something similar for settlement and I think some manual database fix-ups in the past may have left us with a handful of non-capture events that have `gateway_event_date`s (i.e. we set the `updated` date to when it actually got inserted to the database and then put the date it should have been inserted into the `gateway_event_date`).

alexbishop1 (approved):
👍

PR #499, opened 2017-11-23
PP-2769 Back fill 3D Secure data from charges table to cards_3ds table

chrisgrimble:
Back fill the data into the cards_3ds table that was created a number
of weeks ago. This was reverted once before going totest, staging or
prod as the load it caused was too high for the database and resulted
in service outage. Database servers have now been upgraded and so
adding this back in.


alexbishop1 (approved):
Looks good

PR #500, opened 2017-11-23
PP-2756 refactoring notify client provider

kakumara:
## WHAT
Along with PP-2756 we are required to support custom email branding, which can only be supported using a separate notify service account. This requires being able to instantiate NotifyClient's depending on the gateway account settings rather than the default GOVUK Pay notify account settings
This enables that possibility by being able to construct NotifyClients based on a given apiKey. (Although, this refactoring still uses the default apiKey from the env variables)


SandorArpa (dismissed):
LGTM 👍

SandorArpa (approved):

PR #503, opened 2017-11-23
PP-2756 send payment confirmation email based on custom branding settings

kakumara:
## WHAT

If the gateway account contains a custom notify account api key and its relevant template it should use that to send the email instead of the default govuk.pay email. Otherwise the default Pay email will be sent as usual.

Removed the unused `checkDeliveryStatus` method and its corresponding test


SandorArpa (dismissed):
LGTM 👍

SandorArpa (approved):

PR #504, opened 2017-11-24
Pp 2555 replace identity with sequence

georgeracu:
## WHAT
Currently our entities in connector inherit from AbstractEntity which has an id who's generation strategy is IDENTITY. When we declare an id column of type `*serial` in PostgreSQL, the database will automatically generate a `sequence` with `allocationSize` of 1. EclipseLink doesn't recommend using generation strategy IDENTITY, but SEQUENCE. Given that we already have sequence generators, let's tell EclipseLink to use them.

## HOW 
When using generation type SEQUENCE, then you need to tell EclipseLink which sequence generator to use, otherwise will default to a non existing one. For all entities in connector to have ids with strategy SEQUENCE is not possible to inherit from a base class as the sequence generators need to be unique and cannot be overridden by annotations so I had to pull out the id from the base class into each entity. 

This changeset adds an id to each entity class, leaves the base class only with field version and specifies to EclipseLink which sequence generator to use.


src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -1,30 +1,55 @@
...
+import static uk.gov.pay.connector.model.domain.ChargeStatus.fromString;
 import static uk.gov.pay.connector.model.domain.PaymentGatewayStateTransitions.defaultTransitions;
 
 @Entity
 @Table(name = "charges")
-@SequenceGenerator(name = "charges_charge_id_seq", sequenceName = "charges_charge_id_seq", allocationSize = 1)
 @Access(AccessType.FIELD)
-public class ChargeEntity extends AbstractEntity {
+@SequenceGenerator(name = "charges_charge_id_seq",
+        sequenceName = "charges_charge_id_seq", allocationSize = 1)
```
chrisgrimble, line 44: The previous sequence was cards_id_seq so should this be charges_id_seq

src/main/java/uk/gov/pay/connector/model/domain/NotificationCredentials.java
```diff
@@ -26,6 +40,14 @@
...
     }
 
+    public Long getId() {
+        return id;
+    }
+
+    public void setId(Long id) {
+        this.id = id;
+    }
+
```
chrisgrimble, line 54: Can we move these down below the other constructors please. Missed the fact there was more than the default constructor when scanning the code.

chrisgrimble (changes requested):
Some very tiny things to change.

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
georgeracu: Hibernate generates a sequence using a naming convention like `table_name_column_name_seq` so when this sequence was created the `id` column was named `charge_id` and it got renamed later to `id` but the sequence didn't get renamed. Now we have to use the initial sequence generated.

src/main/java/uk/gov/pay/connector/model/domain/NotificationCredentials.java
georgeracu: Done

georgeracu:
Did a `squash` on commits to prepare it for merge with master

chrisgrimble (approved):

PR #508, opened 2017-11-27
PP-2876 Expose transactions summary data for dashboard from connector

alexbishop1:
Adds an internal connector API endpoint to handle requests like this:

`GET /v1/api/accounts/123/transactions-summary?from_date=2017-11-03T00:00:00Z&to_date=2017-11-10T00:00:00Z`

… with responses like this:

```
HTTP/1.1 200 OK
Content-Type: application/json

{
    "successful_payments" : {
        "count": 10,
        "total_in_pence": 55000
    },
    "refunded_payments" : {
        "count": 2,
        "total_in_pence": 11000
    },
    "net_income" : {
        "total_in_pence": 44000
    }
}
```

A payment is included if it has an external success status and its created date (but not necessarily the date it went into a success state) falls within the range . A refund is included if it has an external success status and its created date (but not necessarily the date it went into a success state) falls within the range. For refunds, only the created date of the refund matters: the created date of the payment the refund belongs to need not fall within the range for the refund to be included.

Net income is simply the successful payments total minus the refunded payments total.

This pull request is a little on the lengthy side but the commits are separated fairly sensibly:

1. a005ad4  — Adds a method to the `ChargeDao` to get all charges within a date range matching one of the supplied statuses
2. 9fbeccc   — Adds a method to the `RefundDao` to get all refunds within a date range matching one of the supplied statuses
3. 7234fe4 — Adds the API endpoint (with input validation), logic to call the above new methods, arithmetic, and Java classes to represent the output JSON
4. 67981e9 — Integration test
5. 5a14c63 — Removes “(to be implemented)” from the internal API documentation

Manual testing has been conducted to verify the endpoint does actually return the appropriate JSON.


georgeracu (approved):
Looks good.

PR #509, opened 2017-11-28
Pp 2778 write refunds to transactions

chrisgrimble:
## WHAT
Start writing to the transactions table whenever we do a refund.
Have also started writing refund events to the 
transaction_events table which had been broken out to a 
different story. This is because this code is shared with 
ChargeTransactions and was easier to leave in than disable.

## HOW 
- Created RefundTransactionEntity and 
RefundTransactionEventEntity.
- Created RefundStatusUpdater to update status on 
RefundTransactionEntity.
- Call method on RefundStatusUpdater whenever we we changing
the status on RefundEntity.
- Set refundReference on RefundTransactions once we have a response 
from the gateway.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
@@ -2,8 +2,11 @@ package uk.gov.pay.connector.model.domain;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import uk.gov.pay.connector.model.domain.transaction.ChargeTransactionEntity;
+import uk.gov.pay.connector.model.domain.transaction.ChargeTransactionEventEntity;
+import uk.gov.pay.connector.model.domain.transaction.RefundTransactionEntity;
 import uk.gov.pay.connector.model.domain.transaction.TransactionEntity;
 import uk.gov.pay.connector.model.domain.transaction.TransactionOperation;
+import uk.gov.pay.connector.service.CardService;
 
 import javax.persistence.Access;
 import javax.persistence.AccessType;
@@ -24,6 +27,10 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import static uk.gov.pay.connector.model.domain.transaction.TransactionOperation.CHARGE;
+import static uk.gov.pay.connector.model.domain.transaction.TransactionOperation.REFUND;
 
 @Entity
 @Table(name = "payment_requests")
@@ -141,23 +148,29 @@ public class PaymentRequestEntity extends AbstractVersionedEntity {
     }
 
     public void setTransactions(List<TransactionEntity> transactions) {
-        this.transactions = transactions;
-        transactions.forEach(transactionEntity -> transactionEntity.setPaymentRequest(this));
+        transactions.forEach(this::addTransaction);
     }
 
     public ChargeTransactionEntity getChargeTransaction() {
-        return (ChargeTransactionEntity)transactions.stream().filter(byChargeTransactions()).findFirst().orElseThrow(
+        return (ChargeTransactionEntity)transactions.stream().filter(byOperation(CHARGE)).findFirst().orElseThrow(
                 () -> new IllegalStateException("Payment Request has been initialised without a charge transaction")
         );
     }
 
-    //Remove this once transactions table has been backfilled and we will no longer need to set this.
-    public boolean hasChargeTransaction() {
-        return transactions.stream().anyMatch(byChargeTransactions());
+    public List<RefundTransactionEntity> getRefundTransactions() {
+        return transactions.stream()
+                .filter(byOperation(REFUND))
+                .map(RefundTransactionEntity.class::cast)
+                .collect(Collectors.toList());
     }
 
-    private Predicate<TransactionEntity> byChargeTransactions() {
-        return transactionEntity -> transactionEntity.getOperation().equals(TransactionOperation.CHARGE);
+    //Remove this once transactions table has been backfilled and we will no longer need to set this.
+    public boolean hasChargeTransaction() {
+        return transactions.stream().anyMatch(byOperation(CHARGE));
+    }
+
+    private Predicate<TransactionEntity> byOperation(TransactionOperation transactionOperation) {
+        return transactionEntity -> transactionEntity.getOperation().equals(transactionOperation);
     }
 
     public static PaymentRequestEntity from(ChargeEntity chargeEntity, ChargeTransactionEntity transactionEntity) {
--- a/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
@@ -18,8 +18,8 @@ import javax.persistence.Inheritance;
 import javax.persistence.InheritanceType;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
-import javax.persistence.SequenceGenerator;
 import javax.persistence.OneToMany;
+import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
@@ -33,16 +33,11 @@ import java.util.List;
 @SequenceGenerator(name = "transactions_id_seq",
         sequenceName = "transactions_id_seq",
         allocationSize = 1)
-public abstract class TransactionEntity<E extends Status> extends AbstractVersionedEntity {
-
+public abstract class TransactionEntity<S extends Status, T extends TransactionEventEntity<S, T>> extends AbstractVersionedEntity {
     @Id
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "transactions_id_seq")
     @JsonIgnore
     private Long id;
-
-    @Column(name = "status")
-    @Enumerated(EnumType.STRING)
-    protected E status;
     @Column(name = "amount")
     private Long amount;
     @Column(name = "operation")
@@ -52,7 +47,11 @@ public abstract class TransactionEntity<E extends Status> extends AbstractVersio
     @JoinColumn(name = "payment_request_id", referencedColumnName = "id", updatable = false)
     private PaymentRequestEntity paymentRequest;
     @OneToMany(mappedBy = "transaction", cascade = CascadeType.ALL)
-    private List<TransactionEventEntity> transactionEvents = new ArrayList<>();
+    private List<T> transactionEvents = new ArrayList<>();
+
+    TransactionEntity(TransactionOperation operation) {
+        this.operation = operation;
+    }
 
     public Long getId() {
         return id;
@@ -82,25 +81,28 @@ public abstract class TransactionEntity<E extends Status> extends AbstractVersio
         return operation;
     }
 
-    public void setOperation(TransactionOperation operation) {
-        this.operation = operation;
-    }
+    public abstract S getStatus();
+    abstract void setStatus(S status);
 
-    public E getStatus() {
-        return status;
-    }
-
-    public List<TransactionEventEntity> getTransactionEvents() {
+    public List<T> getTransactionEvents() {
         return transactionEvents;
     }
 
-    protected void addTransactionEvent(TransactionEventEntity<E> transactionEvent, E newStatus) {
+    public final void updateStatus(S newStatus) {
+        updateStatus(newStatus, createNewTransactionEvent());
+    }
+
+    void updateStatus(S newStatus, T transactionEvent) {
+        setStatus(newStatus);
+        addTransactionEvent(transactionEvent, newStatus);
+    }
+
+    private void addTransactionEvent(T transactionEvent, S newStatus) {
         transactionEvent.setTransaction(this);
         transactionEvent.setStatus(newStatus);
         transactionEvent.setUpdated(ZonedDateTime.now());
         getTransactionEvents().add(transactionEvent);
     }
 
-    public abstract void updateStatus(E newStatus);
-    public abstract void updateStatus(E newStatus, ZonedDateTime gatewayEventTime);
+    protected abstract T createNewTransactionEvent();
 }
--- a/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEventEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEventEntity.java
@@ -27,7 +27,7 @@ import java.time.ZonedDateTime;
 @SequenceGenerator(name = "transaction_events_id_seq",
         sequenceName = "transaction_events_id_seq",
         allocationSize = 1)
-public abstract class TransactionEventEntity<E extends Status> extends AbstractVersionedEntity {
+public abstract class TransactionEventEntity<S extends Status, T extends TransactionEventEntity<S, T>> extends AbstractVersionedEntity {
 
     @Id
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "transaction_events_id_seq")
@@ -36,15 +36,11 @@ public abstract class TransactionEventEntity<E extends Status> extends AbstractV
 
     @ManyToOne
     @JoinColumn(name = "transaction_id", referencedColumnName = "id", updatable = false)
-    private TransactionEntity<E> transaction;
+    private TransactionEntity<S, T> transaction;
 
     @Column(name = "updated")
     @Convert(converter = UTCDateTimeConverter.class)
     private ZonedDateTime updated;
-
-    @Column(name = "gateway_event_date")
-    @Convert(converter = UTCDateTimeConverter.class)
-    private ZonedDateTime gatewayEventDate;
 
     public Long getId() {
         return id;
@@ -54,11 +50,7 @@ public abstract class TransactionEventEntity<E extends Status> extends AbstractV
         this.id = id;
     }
 
-    public TransactionEntity<E> getTransaction() {
-        return transaction;
-    }
-
-    public void setTransaction(TransactionEntity<E> transaction) {
+    public void setTransaction(TransactionEntity<S, T> transaction) {
         this.transaction = transaction;
     }
 
@@ -70,14 +62,6 @@ public abstract class TransactionEventEntity<E extends Status> extends AbstractV
         this.updated = updated;
     }
 
-    public ZonedDateTime getGatewayEventDate() {
-        return gatewayEventDate;
-    }
-
-    public void setGatewayEventDate(ZonedDateTime gatewayEventDate) {
-        this.gatewayEventDate = gatewayEventDate;
-    }
-
-    public abstract void setStatus(E status);
-    public abstract E getStatus();
+    public abstract void setStatus(S status);
+    public abstract S getStatus();
 }
--- a/src/main/java/uk/gov/pay/connector/service/PaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/service/PaymentProvider.java
@@ -17,7 +17,7 @@ import java.util.Optional;
 
 public interface PaymentProvider<T extends BaseResponse, R> {
 
-    String getPaymentGatewayName();
+    PaymentGatewayName getPaymentGatewayName();
 
     StatusMapper getStatusMapper();
 
--- a/src/main/java/uk/gov/pay/connector/service/RefundStatusUpdater.java
+++ b/src/main/java/uk/gov/pay/connector/service/RefundStatusUpdater.java
@@ -1,0 +1,62 @@
+package uk.gov.pay.connector.service;
+
+import com.google.inject.persist.Transactional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.dao.RefundTransactionDao;
+import uk.gov.pay.connector.model.domain.RefundStatus;
+import uk.gov.pay.connector.model.domain.transaction.RefundTransactionEntity;
+
+import javax.inject.Inject;
+import java.util.Optional;
+
+import static java.lang.String.format;
+
+@Transactional
+public class RefundStatusUpdater {
+    private static final Logger logger = LoggerFactory.getLogger(RefundStatusUpdater.class);
+
+    private final RefundTransactionDao refundTransactionDao;
+
+    @Inject
+    public RefundStatusUpdater(RefundTransactionDao refundTransactionDao) {
+        this.refundTransactionDao = refundTransactionDao;
+    }
+
+    public void updateRefundTransactionStatus(PaymentGatewayName provider, String refundReference, RefundStatus newRefundStatus) {
+        Optional<RefundTransactionEntity> refundTransaction = refundTransactionDao.findByProviderAndReference(provider, refundReference);
+        if (refundTransaction.isPresent()) {
+            updateRefundTransactionStatus(newRefundStatus, refundTransaction.get());
+        } else {
+            logger.warn(format("Not updating refund transaction status for refundReference [%s] to [%s] refund transaction not found",
+                    refundReference,
+                    newRefundStatus.getValue()
+            ));
+        }
+    }
+
+    public void setReferenceAndUpdateTransactionStatus(String refundExternalId, String refundReference, RefundStatus newRefundStatus) {
+        Optional<RefundTransactionEntity> refundTransactionOptional = refundTransactionDao.findByExternalId(refundExternalId);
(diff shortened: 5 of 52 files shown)


src/main/java/uk/gov/pay/connector/model/domain/PaymentRequestEntity.java
```diff
@@ -141,23 +148,29 @@ public void addTransaction(TransactionEntity transactionEntity) {
...
     }
 
     public ChargeTransactionEntity getChargeTransaction() {
-        return (ChargeTransactionEntity)transactions.stream().filter(byChargeTransactions()).findFirst().orElseThrow(
+        return (ChargeTransactionEntity)transactions.stream().filter(byOperation(CHARGE)).findFirst().orElseThrow(
                 () -> new IllegalStateException("Payment Request has been initialised without a charge transaction")
         );
     }
 
+    public Optional<RefundTransactionEntity> getRefundTransaction() {
```
alexbishop1: This doesn’t feel quite right: a payment request (or charge if you have an old-fashioned view of our domain) can have multiple refunds associated with it, so I would expect this method to return a list.

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
```diff
@@ -33,16 +33,11 @@
 @SequenceGenerator(name = "transactions_id_seq",
         sequenceName = "transactions_id_seq",
         allocationSize = 1)
-public abstract class TransactionEntity<E extends Status> extends AbstractVersionedEntity {
-
+public abstract class TransactionEntity<E extends Status, T extends TransactionEventEntity<E>> extends AbstractVersionedEntity {
```
alexbishop1: The letters are a bit non-obvious here: maybe `E` could be `S` for `Status` (OK with `TransactionEventEntity` being either `T` for `TransactionEventEntity` or `E` for `Event`)

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
```diff
@@ -82,25 +81,28 @@ public TransactionOperation getOperation() {
...
-        return status;
+    public final void updateStatus(E newStatus) {
+        updateStatus(newStatus, createNewTransactionEvent());
     }
 
-    public List<TransactionEventEntity> getTransactionEvents() {
-        return transactionEvents;
+    void updateStatus(E newStatus, T transactionEvent) {
+        setStatus(newStatus);
+        addTransactionEvent(transactionEvent, newStatus);
```
alexbishop1: I like the coupling of setting the status and adding a transaction event here

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEventEntity.java
```diff
@@ -36,16 +36,12 @@
 
     @ManyToOne
     @JoinColumn(name = "transaction_id", referencedColumnName = "id", updatable = false)
-    private TransactionEntity<E> transaction;
+    private TransactionEntity<E, ? extends TransactionEventEntity> transaction;
```
alexbishop1: This probably shouldn’t be a wildcard because for any concrete implementation, it’s always going to be the same type at every point in this class where the wildcard is used. Fortunately, we can eliminate the wildcard.

In this class, change the class declaration from `TransactionEventEntity<E extends Status>` to the self-referential `TransactionEventEntity<E extends Status, T extends TransactionEventEntity<E, T>>`.

Next, go to the `TransactionEntity` class and change its declaration from `TransactionEntity<E extends Status, T extends TransactionEventEntity<E>>` to `TransactionEntity<E extends Status, T extends TransactionEventEntity<E, T>>`.

Then go to the `ChargeTransactionEventEntity` class make it so that it `extends TransactionEventEntity<ChargeStatus, ChargeTransactionEventEntity>`. After that, go to the `RefundTransactionEventEntity` and make it so that it `extends TransactionEventEntity<RefundStatus, RefundTransactionEventEntity>`.

Then you can go back to this class and change every `? extends TransactionEventEntity` to `T`.

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -29,7 +32,7 @@ public  void updateChargeTransactionStatus(String externalId, ChargeStatus newCh
         );
     }
 
-    public  void updateChargeTransactionStatus(String externalId, ChargeStatus newChargeStatus) {
+    public void updateChargeTransactionStatus(String externalId, ChargeStatus newChargeStatus) {
```
alexbishop1, line 32: Nice extraneous space removal

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -57,4 +60,28 @@ private void updateChargeTransactionStatus(String externalId, ChargeStatus newCh
             }
         });
     }
+
+    public void updateRefundTransactionStatus(String externalId, RefundStatus newRefundStatus) {
```
alexbishop1: Might want to name `externalId` so it’s clear that it’s the external ID of the payment request, not the refund (which has its own different external ID)

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -57,4 +60,28 @@ private void updateChargeTransactionStatus(String externalId, ChargeStatus newCh
...
+        paymentRequestDao.findByExternalId(externalId).ifPresent(paymentRequestEntity -> {
+            Optional<RefundTransactionEntity> optionalRefundTransaction = paymentRequestEntity.getRefundTransaction();
+
+            if (optionalRefundTransaction.isPresent()) {
```
alexbishop1: I can’t wait for Java 9’s `Optional.ifPresentOrElse()`

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -57,4 +60,28 @@ private void updateChargeTransactionStatus(String externalId, ChargeStatus newCh
             }
         });
     }
+
+    public void updateRefundTransactionStatus(String externalId, RefundStatus newRefundStatus) {
+        paymentRequestDao.findByExternalId(externalId).ifPresent(paymentRequestEntity -> {
+            Optional<RefundTransactionEntity> optionalRefundTransaction = paymentRequestEntity.getRefundTransaction();
```
alexbishop1: This means we’re always going to update the first refund we find for a payment request, which may not be the one we want

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -57,4 +60,28 @@ private void updateChargeTransactionStatus(String externalId, ChargeStatus newCh
...
+
+            if (optionalRefundTransaction.isPresent()) {
+                RefundTransactionEntity refundTransaction = optionalRefundTransaction.get();
+                logger.info(String.format("Changing refund transaction status for externalId [%s] transactionId [%s] [%s]->[%s]",
```
alexbishop1: This is a bit ambiguous: `externalId` relates to the payment request but `transactionId` relates to the refund (and is the database ID — generally we don’t log these in code, preferring to use external IDs or similar).

Also, with SLF4J, we don’t need to use `String.format`: it directly understands `{}` placeholders, for example:

```
logger.info("After eating {}, I like to eat {} and {}", "chips", "cake", "ice-cream");
```

The SLF4J folks also claim it’s much faster than `String.format`.

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -57,4 +60,28 @@ private void updateChargeTransactionStatus(String externalId, ChargeStatus newCh
...
+                        externalId,
+                        refundTransaction.getId(),
+                        refundTransaction.getStatus().getValue(),
+                        newRefundStatus.getValue())
+                );
+                refundTransaction.updateStatus(newRefundStatus);
+            } else {
+                logger.info(
+                        String.format("Not updating refund transaction status for externalId [%s] to [%s] charge transaction not found",
```
alexbishop1: It’s the refund that’s not found, not the charge

src/test/java/uk/gov/pay/connector/model/domain/PaymentRequestEntityTest.java
```diff
@@ -37,11 +34,10 @@ public void shouldReturnChargeTransactionWhenThereIsOnlyAChargeTransaction() thr
...
-        refundTransaction.setOperation(TransactionOperation.REFUND);
+        RefundTransactionEntity refundTransaction = new RefundTransactionEntity();
         paymentRequestEntity.addTransaction(refundTransaction);
 
         ChargeTransactionEntity chargeTransaction = paymentRequestEntity.getChargeTransaction();
         assertThat(chargeTransaction, is(expectedChargeTransaction));
     }
-}
\ No newline at end of file
+}
```
alexbishop1: We would probably benefit from having tests that check a payment request returns the right refund when it has a charge and refunds

src/test/java/uk/gov/pay/connector/model/domain/transaction/RefundTransactionEntityTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Test;
+import uk.gov.pay.connector.model.domain.ChargeStatus;
+import uk.gov.pay.connector.model.domain.RefundEntity;
+import uk.gov.pay.connector.model.domain.RefundEntityFixture;
+import uk.gov.pay.connector.model.domain.RefundStatus;
+
+import java.util.List;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.*;
```
alexbishop1: Our Java style guide says we should not use wildcard imports

src/test/java/uk/gov/pay/connector/model/domain/transaction/RefundTransactionEntityTest.java
```diff
@@ -0,0 +1,60 @@
...
+                .withAmount(amount)
+                .withExternalId(externalId)
+                .withStatus(RefundStatus.REFUND_ERROR)
+                .build();
+
+        RefundTransactionEntity transactionEntity = RefundTransactionEntity.from(refundEntity);
+
+        assertThat(transactionEntity.getAmount(), is(amount));
+        assertThat(transactionEntity.getRefundExternalId(), is(externalId));
+        assertThat(transactionEntity.getUserExternalId(), is(refundEntity.getUserExternalId()));
```
alexbishop1: I think it would be nicer if the user external ID were explicitly specified (in the say way that `externalId` and `amount` are)

src/test/java/uk/gov/pay/connector/model/domain/transaction/RefundTransactionEntityTest.java
```diff
@@ -0,0 +1,60 @@
...
+    public void setStatusAddsTransactionEvent() throws Exception {
+        RefundTransactionEntity refundTransactionEntity = new RefundTransactionEntity();
+        RefundStatus expectedStatus = RefundStatus.CREATED;
+
+        assertThat(refundTransactionEntity.getTransactionEvents().size(), is(0));
+        refundTransactionEntity.updateStatus(expectedStatus);
+
+        List<RefundTransactionEventEntity> transactionEvents = refundTransactionEntity.getTransactionEvents();
+        assertThat(transactionEvents.size(), is(1));
+        assertThat(transactionEvents.get(0).getStatus(), is(expectedStatus));
```
alexbishop1, line 45: Might be nice to check the updated timestamp gets set correctly here but the fact that we don’t control the clock will make it a pain

src/test/java/uk/gov/pay/connector/service/StatusUpdaterTest.java
```diff
@@ -84,16 +86,40 @@ public void canHandleGatewayEventDateBeingNullAndDoesNotSetAValue() throws Excep
...
+        PaymentRequestEntity paymentRequest = mock(PaymentRequestEntity.class);
+        when(mockPaymentRequestDao.findByExternalId(SOME_EXTERNAL_ID)).thenReturn(Optional.of(paymentRequest));
+        when(paymentRequest.getRefundTransaction()).thenReturn(Optional.empty());
+
+        new StatusUpdater(mockPaymentRequestDao).updateRefundTransactionStatus(SOME_EXTERNAL_ID, REFUND_SUBMITTED);
+
+        verify(paymentRequest).getRefundTransaction();
+        verifyNoMoreInteractions(paymentRequest);
+    }
+
```
alexbishop1: We need tests to check that we handle updating a status correctly when mulitple refunds exist for a single payment request

alexbishop1 (changes requested):
Unfortunately, these changes assume that there will only ever be one refund for a particular payment request, which isn’t the case (there can be many). We need to support multiple refunds, including finding the the right refund to update.

src/test/java/uk/gov/pay/connector/model/domain/transaction/RefundTransactionEntityTest.java
chrisgrimble: Would be nice to control the date from a date provider which is injected into the object. However this is created by eclipse link so is not as simple as just injecting it when the object is created.

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
chrisgrimble: Removed

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEventEntity.java
```diff
@@ -36,16 +36,12 @@
 
     @ManyToOne
     @JoinColumn(name = "transaction_id", referencedColumnName = "id", updatable = false)
-    private TransactionEntity<E> transaction;
+    private TransactionEntity<S, ? extends TransactionEventEntity> transaction;
```
alexbishop1: See my previous https://github.com/alphagov/pay-connector/pull/509#discussion_r153751551

src/main/java/uk/gov/pay/connector/service/PaymentProvider.java
```diff
@@ -17,7 +17,11 @@
 
 public interface PaymentProvider<T extends BaseResponse, R> {
 
-    String getPaymentGatewayName();
+    default String getPaymentGatewayNameAsString() {
+        return getPaymentGatewayName().getName();
+    }
```
alexbishop1: I’m not entirely convinced this reads any better than`getPaymentGatewayName().toString()` would, which we could do by implementing `toString()` on `PaymentGatewayName`

src/main/java/uk/gov/pay/connector/service/RefundStatusUpdater.java
```diff
@@ -0,0 +1,64 @@
...
+import uk.gov.pay.connector.model.domain.transaction.RefundTransactionEntity;
+
+import javax.inject.Inject;
+import java.util.Optional;
+
+import static java.lang.String.format;
+
+@Transactional
+public class RefundStatusUpdater {
+    private static final Logger logger = LoggerFactory.getLogger(StatusUpdater.class);
```
alexbishop1: Wrong class!

src/main/java/uk/gov/pay/connector/service/RefundStatusUpdater.java
```diff
@@ -0,0 +1,64 @@
...
+    private final RefundTransactionDao refundTransactionDao;
+
+    @Inject
+    public RefundStatusUpdater(RefundTransactionDao refundTransactionDao) {
+        this.refundTransactionDao = refundTransactionDao;
+    }
+
+    public void updateRefundTransactionStatus(PaymentGatewayName provider, String refundReference, RefundStatus newRefundStatus) {
+        Optional<RefundTransactionEntity> refundTransaction = refundTransactionDao.findByProviderAndReference(provider, refundReference);
+        String errorMessage = format("Not updating refund transaction status for refundReference [%s] to [%s] charge transaction not found",
```
alexbishop1: It’s the refund transaction that’s not found, not the charge transaction

alexbishop1: I can see why we’re doing this create-an-error-message-in-advance-just-in-case malarky but it gives undue prominence to the errors and makes the flow less obvious.

I think it might be better to go with something like this instead:

```
    public void updateRefundTransactionStatus(PaymentGatewayName provider, String refundReference, RefundStatus newRefundStatus) {
        Optional<RefundTransactionEntity> refundTransactionOptional = refundTransactionDao.findByProviderAndReference(provider, refundReference);
        if (refundTransactionOptional.isPresent()) {
            updateRefundTransactionStatus(refundTransactionOptional.get(), newRefundStatus);
        } else {
            logger.info("Not updating refund transaction status for refundReference [%s] to [%s] refund transaction not found",
                    refundReference,
                    newRefundStatus.getValue());
        }
    }

    public void setReferenceAndUpdateTransactionStatus(String refundExternalId, String refundReference, RefundStatus newRefundStatus) {
        Optional<RefundTransactionEntity> refundTransactionOptional = refundTransactionDao.findByExternalId(refundExternalId);
        refundTransactionOptional.ifPresent(refundTransaction -> refundTransaction.setRefundReference(refundReference));
        if (refundTransactionOptional.isPresent()) {
            updateRefundTransactionStatus(refundTransactionOptional.get(), newRefundStatus);
        } else {
            logger.info("Not updating refund transaction status for externalId [%s] to [%s] refund transaction not found",
                    refundExternalId,
                    newRefundStatus.getValue());
        }
    }

    private static void updateRefundTransactionStatus(RefundTransactionEntity refundTransaction, RefundStatus newRefundStatus) {
        logger.info("Changing refund transaction status for refundReference [{}] transactionId [{}] [{}]->[{}]",
                refundTransaction.getRefundReference(),
                refundTransaction.getId(),
                refundTransaction.getStatus().getValue(),
                newRefundStatus.getValue()
        );
        refundTransaction.updateStatus(newRefundStatus);
    }
```

We can replace the `isPresent`s with `ifPresentOrElse` when we upgrade to Java 9

src/main/java/uk/gov/pay/connector/service/RefundStatusUpdater.java
```diff
@@ -0,0 +1,64 @@
...
+        updateRefundTransactionStatus(refundTransaction, newRefundStatus, errorMessage);
+    }
+
+    public void setReferenceAndUpdateTransactionStatus(String refundExternalId, String refundReference, RefundStatus newRefundStatus) {
+        Optional<RefundTransactionEntity> refundTransactionOptional = refundTransactionDao.findByExternalId(refundExternalId);
+        refundTransactionOptional.ifPresent(refundTransaction -> refundTransaction.setRefundReference(refundReference));
+        updateRefundTransactionStatus(
+                refundTransactionOptional,
+                newRefundStatus,
+                format("Not updating refund transaction status for externalId [%s] to [%s] charge transaction not found",
```
alexbishop1: It’s the refund transaction that’s not found, not the charge transaction

src/main/java/uk/gov/pay/connector/service/RefundStatusUpdater.java
```diff
@@ -0,0 +1,64 @@
...
+                )
+        );
+    }
+
+    private void updateRefundTransactionStatus(Optional<RefundTransactionEntity> optionalRefundTransaction, RefundStatus newRefundStatus, String errorMessage) {
+        if (optionalRefundTransaction.isPresent()) {
+            RefundTransactionEntity refundTransaction = optionalRefundTransaction.get();
+            logger.info("Changing refund transaction status for refundReference [{}] transactionId [{}] [{}]->[{}]",
+                    refundTransaction.getRefundReference(),
+                    refundTransaction.getId(),
```
alexbishop1: `refundTransaction.getId()` is the database ID of the refund. I assume this shoud have been the refund’s external ID or maybe something relating to the associated charge or payment request

src/main/java/uk/gov/pay/connector/service/RefundStatusUpdater.java
```diff
@@ -0,0 +1,64 @@
...
+                    refundTransaction.getStatus().getValue(),
+                    newRefundStatus.getValue()
+            );
+
+            refundTransaction.updateStatus(newRefundStatus);
+        } else {
+            logger.info(errorMessage);
```
alexbishop1: I wonder if this should be a higher level than info…

src/main/java/uk/gov/pay/connector/service/StatusUpdater.java
```diff
@@ -40,19 +40,17 @@ private void updateChargeTransactionStatus(String externalId, ChargeStatus newCh
         paymentRequestDao.findByExternalId(externalId).ifPresent(paymentRequestEntity -> {
             if (paymentRequestEntity.hasChargeTransaction()) {
                 ChargeTransactionEntity chargeTransaction = paymentRequestEntity.getChargeTransaction();
-                logger.info(String.format("Changing transaction status for externalId [%s] transactionId [%s] [%s]->[%s]",
+                logger.info("Changing transaction status for externalId [{}] transactionId [{}] [{}]->[{}]",
                         externalId,
                         chargeTransaction.getId(),
```
alexbishop1: This should probably be `chargeTransaction.getGatewayTransactionId()`

src/test/java/uk/gov/pay/connector/model/domain/PaymentRequestEntityTest.java
```diff
@@ -36,12 +35,47 @@ public void shouldReturnChargeTransactionWhenThereIsOnlyAChargeTransaction() thr
...
+    public void shouldNotReturnRefundTransactionWhenThereIsOnlyAChargeTransaction() throws Exception {
+        List<RefundTransactionEntity> refundTransactions = paymentRequestEntity.getRefundTransactions();
+        assertThat(refundTransactions.isEmpty(), is(true));
+    }
+
+    @Test
+    public void shouldReturnRefundTransactionWhenThereAreChargeAndRefundTransactions() throws Exception {
+        RefundTransactionEntity refundTransaction = new RefundTransactionEntity();
+        String refundReference = "someRefundReference";
+        refundTransaction.setRefundExternalId(refundReference);
```
alexbishop1: The local variable is called `refundReference` but it’s used as the external ID

src/test/java/uk/gov/pay/connector/model/domain/transaction/RefundTransactionEntityBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+        this.userExternalId = userExternalId;
+        return this;
+    }
+
+    public RefundTransactionEntityBuilder withAmount(Long amount) {
+        this.amount = amount;
+        return this;
+    }
+
+    public RefundTransactionEntityBuilder withRefindReference(String refundReference) {
```
alexbishop1: `withRefundReference`

src/test/java/uk/gov/pay/connector/service/ChargeRefundServiceTest.java
```diff
@@ -369,10 +431,14 @@ public void shouldUpdateRefundRecordToFailWhenRefundFails() {
         verify(mockProvider).refund(argThat(aRefundRequestWith(capturedCharge, amount)));
         verify(mockRefundDao).findById(any(Long.class));
         verify(spiedRefundEntity).setStatus(RefundStatus.REFUND_ERROR);
+        verify(mockRefundStatusUpdater).setReferenceAndUpdateTransactionStatus(
+                refundExternalId,
+                "",
+                RefundStatus.REFUND_ERROR
+        );
```
alexbishop1, line 457: I think `shouldOverrideGeneratedReferenceIfProviderReturnAReference` and `shouldStoreEmptyGatewayReferenceIfGatewayReturnsAnError` might want updating too

alexbishop1 (changes requested):
Assuming you’ve caught everywhere we change refund statuses, the logic looks sound now so no major changes requested.

We should probably rename `StatusUpdater` to `ChargeStatusUpdater` now that we have a `RefundStatusUpdater`.

src/test/java/uk/gov/pay/connector/service/ChargeRefundServiceTest.java
chrisgrimble: Done

alexbishop1 (changes requested):
Looks basically done. Please rebase and then add another commit to rename `StatusUpdater` to `ChargeStatusUpdater`.

alexbishop1 (changes requested):
Looks great except that `RefundStatusUpdaterTest` seems to have been accidentally renamed to `RefundChargeStatusUpdaterTest`

alexbishop1 (approved):
🎁

PR #511, opened 2017-11-29
PP-2790 Backfill transactions table

georgeracu:
Backfill table transactions with records from table charges and payment_requests

- Added Liquibase migration script

with @chrisgrimble

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1005,4 +1005,41 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="backfill into transactions all charges that are in terminal states" author="">
+        <sql>
+            INSERT INTO transactions (
+                payment_request_id,
+                gateway_transaction_id,
+                amount,
+                status,
+                operation
+                )
+            SELECT
+                pr.id,
+                ch.gateway_transaction_id,
+                ch.amount,
+                ch.status,
+                'CHARGE'
+            FROM payment_requests pr
+            LEFT JOIN transactions t ON pr.id = t.payment_request_id AND t.operation='CHARGE'
+            JOIN charges ch ON pr.external_id = ch.external_id
+            WHERE t.id IS NULL AND ch.status IN (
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -996,6 +996,42 @@
                 AND ch.issuer_url_3ds IS NOT NULL;
         </sql>
     </changeSet>
+    <changeSet id="backfill transactions from charges" author="">
```
alexbishop1: Can you give this a more specific description? Something like “backfill into transactions all charges that are in terminal states”

alexbishop1 (changes requested):
OK, this is what I think this SQL will do:

This will select all payment requests (due to previous backfills, all existing charges have a payment request) and join them to the transactions, keeping all payment requests (there will be many payment requests without joined transactions). It will then join to the charges (every payment request has a charge, so this will be all of them). It will then filter out everything except those rows that have a null transaction ID, which will leave us with payment requests (joined to their charges) that do not already have a transaction. It will then filter out everything except those rows that have a terminal charge status. For every row that’s left, it will insert a row into transactions with the relevant data from charges and a reference to the correct payment request.

The end result will be that all charges in a terminal state will get a matching row inserted to the transactions table if there is not already one. Once this migration is run, only old (pre-dual–writing) charges in non-terminal states will not have a row in the transactions table.

This seems like what we want to do. Only selecting charges that are in terminal states (which should never change again in the future) should ensure nothing changes underneath us while doing the backfill.

But please make the `changeSet id` more descriptive.

alexbishop1 (changes requested):
I double-checked the terminal states and they’re correct.

However, in the `transactions` table, we use the `ChargeStatus` enum `name()` (which uses underscores instead of spaces). In the `charges` table, we use the `ChargeStatus` enum `toString()` (which uses spaces). We need to replace the spaces with underscores when doing the insert.

alexbishop1 (approved):

PR #512, opened 2017-11-29
PP-2982 Publish metric to hosted graphite with connector build time [DO NOT MERGE]

juanjoqmelian:
## WHAT
We want to introduce some metrics to measure the time that it takes to build connector and run its tests. In that way we will be able to measure the improvement after the work is done to only spin postgres docker container once (before all tests start).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,12 @@
 #!/usr/bin/env groovy
 
+node {
+  withCredentials([string(credentialsId: 'graphite_account_id', variable: 'HOSTED_GRAPHITE_ACCOUNT_ID'),
+                    string(credentialsId: 'graphite_api_key', variable: 'HOSTED_GRAPHITE_API_KEY')]) {
+                        sh 'echo $HOSTED_GRAPHITE_ACCOUNT_ID'
+                    }
+}
+
 pipeline {
   agent any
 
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -28,9 +28,34 @@ pipeline {
...
+                    exit 1
+                fi
+
+                if [ -z "${HOSTED_GRAPHITE_ACCOUNT_ID}" ] ; then
+                    echo "You must supply an account id for the carbon prefix"
+                    exit 1
+                fi
+
+                if [ "${HOSTED_GRAPHITE_API_KEY}" ]; then
+                  METRIC_NAME="${HOSTED_GRAPHITE_API_KEY}.$1"
```
mrlumbu: should this be `HOSTED_GRAPHITE_ACCOUNT_ID ` not `HOSTED_GRAPHITE_API_KEY`?

Jenkinsfile
```diff
@@ -28,9 +28,34 @@ pipeline {
...
+      sh (
+        script: """
+            function postMetricToGraphite() {
+
+                if [ $# -ne 2 ]; then
+                    echo "You must supply metric name and metric value"
+                    exit 1
+                fi
+
+                if [ -z "${HOSTED_GRAPHITE_ACCOUNT_ID}" ] ; then
```
mrlumbu: is this set on jenkins then?

juanjoqmelian: yes, merged PR is here:

https://github.com/alphagov/pay-chef/pull/368

Jenkinsfile
juanjoqmelian: I haven't checked on hosted graphite docs, but that's what we have in other places where we send metrics. So I assume that is the api key what has to be sent.

Jenkinsfile
mrlumbu: 👍

Jenkinsfile
mrlumbu: ok

PR #515, opened 2017-12-04
PP-2790 Backfill into transactions

georgeracu:
## WHAT
Backfill into Transactions all Charges that are not in terminal
state and before 2017-12-01
 🎅

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1048,4 +1048,32 @@
         </sql>
     </changeSet>
 
+    <changeSet id="backfill into transactions all charges that are NOT IN terminal states and that are before 2017-12-01" author="">
+        <sql stripComments="true">
+            INSERT INTO transactions (
+                payment_request_id,
+                gateway_transaction_id,
+                amount,
+                status,
+                operation
+            )
+            SELECT
+                pr.id,
+                ch.gateway_transaction_id,
+                ch.amount,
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1048,4 +1048,32 @@
...
+                -- use all PaymentRequest
+                FROM payment_requests pr
+                -- use all ChargeTransaction and fill missing one from PaymentRequest with null
+                LEFT JOIN transactions t ON pr.id = t.payment_request_id AND t.operation='CHARGE'
+                -- use only Charge that has a PaymentRequest
+                JOIN charges ch ON pr.external_id = ch.external_id
+                -- use only PaymentRequest that don't have a ChargeTransaction
+                WHERE t.id IS NULL
+                -- use only Charge before 2017-11-13 when there was production incident
+                AND ch.created_date &lt; '2017-11-13 00:00:00';
```
alexbishop1: The production incident was on September 13th, not November. November 13th (or something like that) was the created date of the most recent payment with a `CAPTURE_SUBMITTED` status (the couple of payments with other non-terminal statuses were from before then). We should be fine with anything before December really: any payments from November should be in a terminal state by now so are unlikely to change underneath us while backfilling.

alexbishop1 (changes requested):
Looks fine apart from the date

alexbishop1 (approved):
❄️

PR #516, opened 2017-12-05
PP-2790 Backfill ChargeTransactions with charges from 17-11-2017

georgeracu:
- Update transactions status during window of 3-4pm on 17-11-2017

🎅

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1075,6 +1075,32 @@
                 AND ch.created_date &lt; '2017-12-01 00:00:00';
         </sql>
     </changeSet>
+    <changeSet
+            id="backfill into transactions all charges that resulted on 15-Nov-2017 during migration"
+            author="">
+        <sql stripComments="true">
+            UPDATE transactions t2 SET
+                status = a.status
+            FROM (
+                -- do an inner select to get only the ChargeTransaction from the required time frame
+                SELECT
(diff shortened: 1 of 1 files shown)


alexbishop1 (changes requested):
This doesn’t look like it’s what we want: there is already a transaction for each of the affected payments, it just has the wrong status. We want to be doing updates, not inserts.

src/main/resources/migrations.xml
```diff
@@ -1075,5 +1075,35 @@
...
+                REPLACE(ch.status, ' ', '_'),
+                'CHARGE'
+            -- use all PaymentRequest
+            FROM payment_requests pr
+            -- use all ChargeTransaction and fill missing one from PaymentRequest with null
+            LEFT JOIN transactions t ON pr.id = t.payment_request_id AND t.operation='CHARGE'
+            -- use only Charge that has a PaymentRequest
+            JOIN charges ch ON pr.external_id = ch.external_id
+            -- use only PaymentRequest that don't have a ChargeTransaction
+            WHERE t.id IS NULL
```
chrisgrimble: Don't these payments have a charge transaction thought?

alexbishop1 (dismissed):
Not thrilled with the name `a` but it looks like it will work 🎄

alexbishop1 (approved):
☃️

PR #517, opened 2017-12-05
PP-2783 Add transaction_id column to cards table

alexbishop1:
Add a `transaction_id` column to the `cards` table (with a foreign key constraint referencing the `id` column in the `transactions` table). The column is `NULLABLE` for now but we will backfill it in the near future.

Tested by running connector migrations locally.


chrisgrimble (approved):
This constraint will need to be dropped and one that is not nullable added once we have populated the column and are writing to it.

PR #519, opened 2017-12-06
PP-3027 Remove color output from logs in jenkins

simad:
## WHAT
- According to https://issues.jenkins-ci.org/browse/JENKINS-11752 the ansicolor
  plugin exaggerates log files's size by an order of magnitude.
  As an example, Connector log files go from ~160MB to ~16MB with colours disabled.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,6 @@ pipeline {
   }
 
   options {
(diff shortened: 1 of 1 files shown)


rauligs (approved):
LGTM 👍

PR #520, opened 2017-12-08
PP-2791 Backfill RefundTransaction

georgeracu:
## WHAT
- Backfill table transactions with RefundTransaction that have been created before start writing to it
- Insert only refunds that are in terminal state to avoid those in transit
🎅 🎄 🎁 ☃️

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1145,5 +1145,33 @@
             AND transactions.operation = 'CHARGE';
         </sql>
     </changeSet>
+    
+    <changeSet id="Backfill completed refunds into refund transactions" author="">
+        <sql>
+            INSERT INTO TRANSACTIONS (
+                payment_request_id,
+                amount,
+                refund_external_id,
+                status,
+                user_external_id,
+                operation,
+                refund_reference
+            )
+            SELECT
+                pr.id,
+                r.amount,
+                r.external_id,
+                REPLACE(r.status, ' ', '_'),
+                r.user_external_id,
+                'REFUND',
+                r.reference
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1113,4 +1113,44 @@
...
+                amount,
+                refund_external_id,
+                status,
+                user_external_id,
+                operation,
+                refund_reference
+            )
+            SELECT
+                pr.id,
+                ch.gateway_transaction_id,
```
alexbishop1: A refund shouldn’t have a `gateway_transaction_id` (it will always be `NULL`) so we don’t want to do this

alexbishop1 (changes requested):
I’m not sure this is going to do the right thing because a payment request (or charge) can have multiple refunds.

Each refund will have a unique `external_id` however: we should probably use this to determine if a refund is already in the `transactions` table.

alexbishop1 (changes requested):
I’m not sure this needs to be so complicated — I think we might be able to just do something like this:

```
INSERT INTO TRANSACTIONS (…)
SELECT … FROM refunds r
JOIN charges c ON r.charge_id=c.id
JOIN payment_requests pr ON c.external_id=pr.external_id
LEFT JOIN transactions t ON r.external_id=t.refund_external_id
WHERE r.status IN ('REFUND ERROR','REFUNDED')
AND t.refund_external_id IS NULL;
```

Possibly I’ve missed something but I think we can simplify a bit if we exploit the fact that refund external IDs are unique.

alexbishop1 (approved):
☃️

PR #521, opened 2017-12-08
PP-2783 Replace CardEntity chargeId Long with chargeTransactionEntity object

alexbishop1:
In CardEntity, replace the `Long` field `chargeId` (mapped to the `charge_id` column in the `charges` table) with a `ChargeTransactionEntity` field `chargeTransactionEntity` (using the `transaction_id` column in the `cards` table to join to the `transactions` table on its `id` column).

Currently, the `CardEntity` is created and persisted but then never used again. In addition, we never read in existing `CardEntity` entities. Nothing is relying on the `chargeId` field so we’re OK removing it.

Also remove the `equals` and `hashCode` methods from `CardEntity`. Again, nothing is relying on them and most of our other entities don’t have these methods, relying instead on reference equality. In addition, the `CardEntity`’s `equals` method does actually rely on `AddressEntity`’s `equals` method, which uses reference equality, which could give someone a nasty surprise. We can add the methods back if we find a genuine need for them other than making unit tests slightly easier (if you mind the `AddressEntity` caveat), which appears to be why they were added in the first place.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/CardEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/CardEntity.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.model.domain;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import uk.gov.pay.connector.model.domain.transaction.ChargeTransactionEntity;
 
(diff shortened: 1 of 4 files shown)


georgeracu (approved):
LGTM

PR #524, opened 2017-12-11
Revert "PP-2783 Replace CardEntity chargeId Long with chargeTransactionEntity object"

DanailMinchev:
Reverts alphagov/pay-connector#521


alexbishop1 (approved):
This one actually reverts the change!

PR #526, opened 2017-12-14
State machine implementation using immutable directed value graph

heathd:
This data structure is a better fit with a state machine.

It allows us to do all of the various checks we would want to do with a state machine.

The 'value' part also allows us to store an arbitrary value for each 'edge' in
the graph (an edge corresponds to a state transition).

In the future we might store a richer data structure such as an object
representing an event.

Some light weight tests. We don't need to exhaustively test every state transition.

Removed the previous implementations

states dot


src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -8,34 +8,16 @@
...
-import javax.persistence.GeneratedValue;
-import javax.persistence.GenerationType;
-import javax.persistence.Id;
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
-import javax.persistence.OneToMany;
-import javax.persistence.OrderBy;
-import javax.persistence.SequenceGenerator;
-import javax.persistence.Table;
+import javax.persistence.*;
```
rauligs: Using explicit imports rather than using wildcards.
IntelliJ -> Preferences > Editor > Code Style > Java > Imports tab set Class count to use import with '*' and Names count to use static import with '*' to 999 or something.

src/main/java/uk/gov/pay/connector/model/domain/PaymentGatewayStateTransitions.java
```diff
@@ -1,9 +1,120 @@
...
+                .directed()
+                .build();
+
+        graph.putEdgeValue(CREATED, EXPIRED, "ChargeExpiryService");
+        graph.putEdgeValue(ENTERING_CARD_DETAILS, EXPIRED, "ChargeExpiryService");
+        graph.putEdgeValue(AUTHORISATION_3DS_REQUIRED, EXPIRED, "ChargeExpiryService");
+        graph.putEdgeValue(EXPIRE_CANCEL_READY, EXPIRED, "ChargeExpiryService");
+        graph.putEdgeValue(EXPIRE_CANCEL_SUBMITTED, EXPIRED, "ChargeExpiryService");
+
+        graph.putEdgeValue(CREATED, ENTERING_CARD_DETAILS, "frontend: POST /frontend/charges/{chargeId}/status");
```
rauligs, line 70: What is representing the edge value? I though values represent an input (such as an event) to transition to the next state, but this seems different (unless is a wip thing)

rauligs (commented):
Added some comments

src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
heathd: 👍

src/main/java/uk/gov/pay/connector/model/domain/PaymentGatewayStateTransitions.java
heathd: for now it's just a human-readable string to annotate the edge. 

So yes it's a 'wip' thing, not used by the code, but used when rendering the diagram.

Hopefully we can gradually add more annotations until we have a better documented state machine and then we can look to introduce a machine processable edge value

rauligs (dismissed):
LGTM 👍

rauligs (approved):
Re-approved after rebase

PR #528, opened 2017-12-18
PP-2783 Backfill transaction_id column in cards table

alexbishop1:
Currently, some rows in the `cards` table have the `charge_id` column populated (referencing the `id` column of the `charges` table) while others have the `transaction_id` column populated (referencing the `id` column of the `transactions` table).

Backfill the `transaction_id` column so it is always populated, in preparation for the `charge_id` column being dropped.


chrisgrimble (approved):
Looks good to me

PR #529, opened 2017-12-18
Remove backfills for cards transaction_id and completed refunds

alexbishop1:
Remove “backfill transaction_id column in cards table” migration and “Backfill completed refunds into refund transactions” migration (f553a722e4ecd6d0361ccefe0d68b71b0d960c42 and 20e031f19653366ede9664111b232526fed2949d, respectively) until we can solve the probable performance problems.

(These were never run on staging or production.)


alexbishop1:
Changed our minds: just want to remove “backfill transaction_id column in cards table”

PR #535, opened 2017-12-21
PP-2783 add relation between cards and transactions

chrisgrimble:
Currently the cards table references the charge_id. As we will be
deleting the charges tables we want to reference the transaction_id of
the ChargeTransaction. The column has already been added and for new
entires in cards is being populated. This will back fill the entries in
cards that are missing the transaction_id with a Task which can be run
from the admin port of the app.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/CardEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/CardEntity.java
@@ -39,8 +39,12 @@ public class CardEntity extends AbstractVersionedEntity {
     @Embedded
     private AddressEntity billingAddress;
 
-    @JoinColumn(name = "transaction_id", referencedColumnName = "id", updatable = false)
+    @JoinColumn(name = "transaction_id", referencedColumnName = "id")
     private ChargeTransactionEntity chargeTransactionEntity;
+
+    //Only needed for loading old data from database never set this in Java code
+    @Column(name = "charge_id")
+    private Long chargeId;
 
     public CardEntity() {
     }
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/model/domain/CardEntity.java
```diff
@@ -39,9 +39,13 @@
     @Embedded
     private AddressEntity billingAddress;
 
-    @JoinColumn(name = "transaction_id", referencedColumnName = "id", updatable = false)
+    @JoinColumn(name = "transaction_id", referencedColumnName = "id", updatable = true)
```
georgeracu: `updatable = true` is default so no need to specify it

src/test/java/uk/gov/pay/connector/tasks/AddTransactionIdToCardsWorkerITest.java
```diff
@@ -0,0 +1,128 @@
...
+    private AddTransactionIdToCardsWorker worker;
+
+    @Before
+    public void setUp() {
+        worker = env.getInstance(AddTransactionIdToCardsWorker.class);
+        insertTestAccount();
+    }
+
+    @Test
+    public void shouldAddTransactionIdToCard() throws Exception {
```
georgeracu: This test methods don't throw exceptions, so no need to declare it

georgeracu (dismissed):
Looks good, just a couple of comments on unnecessary code

georgeracu (approved):
LGTM

PR #536, opened 2017-12-22
PP-2784 add relationship cards3ds to transactions

chrisgrimble:
Currently the card_3ds table references the charge_id. As we will be
deleting the charges tables we want to reference the transaction_id of
the ChargeTransaction. This is the first step where we will add the new
column to the database. In the next release we will back fill and start
writing to the column.


georgeracu (dismissed):
LGTM

georgeracu (approved):
LGTM

PR #542, opened 2018-01-09
PP-2085 Drop index from pa_request column on card_3ds table

alexbishop1:
Adding this index was a mistake. We never look up rows in the `card_3ds` table using the `pa_request`; we use the `charge_id`.


georgeracu (dismissed):
LGTM

georgeracu (dismissed):
LGTM

georgeracu (approved):
LGTM

PR #544, opened 2018-01-11
PP-2783 Add constraint and remove backfill task

georgeracu:
- Add NOT NULL constraint on column transaction_id from table cards
- Remove task for backfilling transcation_id as it has been completed


kakumara (approved):
LOKTM

src/main/resources/migrations.xml
```diff
@@ -1204,11 +1204,15 @@
...
-
     <changeSet id="add index to transaction_id column on card_3ds table" runInTransaction="false" author="">
         <sql>
             CREATE INDEX CONCURRENTLY idx_card_3ds_transaction_id ON card_3ds (transaction_id);
         </sql>
     </changeSet>
 
+    <changeSet id="add NotNullConstraint to cards.transaction_id" author="">
+        <addNotNullConstraint columnName="transaction_id"
+                              tableName="cards"/>
```
alexbishop1, line 1215: Are we confident this isn’t going to lock the table for too long?

alexbishop1 (approved):
Looks fine as long we’re confident adding the `NOT NULL` constraint won’t cause things to back up for too long

src/main/resources/migrations.xml
georgeracu: We will test it and see the result 😸

PR #545, opened 2018-01-11
PP-2784 Add NOT NULL constraint on transaction_id card_3ds

georgeracu:
- Removed field charge_id from Card3dsEntity. This column is not needed anymore as the linking is done via `transaction_id`
- Removed DropWizard task to backfill column `transaction_id`. Has been completed in a previous release
- Added NOT NULL constraint on `card_3ds.transaction_id`
- Dropped default value constraint on `card_3ds.charge_id`
- Dropped sequence `card_3ds_charge_id_seq sequence` as not in use by `card_3ds.charge_id`


alexbishop1 (dismissed):
Looks good, with the same performance testing caveat as #544

src/main/resources/migrations.xml
```diff
@@ -1215,4 +1215,16 @@
...
+        <addNotNullConstraint tableName="card_3ds" columnName="transaction_id"/>
+    </changeSet>
+
+    <changeSet id="drop not null constraint on charge_id column in card_3ds table" author="">
+        <dropNotNullConstraint tableName="card_3ds" columnName="charge_id"/>
+    </changeSet>
+
+    <changeSet id="drop card_3ds_charge_id_seq sequence" author="">
+        <dropSequence sequenceName="card_3ds_charge_id_seq"/>
+    </changeSet>
```
alexbishop1, line 1240: I think we’re going to want to drop the default value that uses this sequence before doing this:

```
<changeSet id="drop default value from charge_id column in card_3ds table" author="">
    <dropDefaultValue tableName="card_3ds" columnName="charge_id"/>
</changeSet>
```

alexbishop1 (changes requested):
Please update the description to reflect the additional database changes this is making now

src/main/resources/migrations.xml
georgeracu: Has been updated and amended commit

alexbishop1 (dismissed):
Looks good, with the same mild fear as #544 (test that adding the `NOT NULL` constraint isn’t too slow)

src/main/resources/migrations.xml
chrisgrimble: Don't we just want to drop the charge_id column on card_3ds?

alexbishop1: I wouldn’t want to drop the column until we’re certain the backfilling was successful

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #548, opened 2018-01-15
PP-3154 Rename column to be consistent

georgeracu:
- Renamed column to `created_date` to be consistent with table
payment_requests


alexbishop1 (approved):
👍
It’s fine to just rename this because the column hasn’t yet been created in any environments

PR #549, opened 2018-01-17
Pp 3154 backfill created date on transactions

georgeracu:
A DropWizard Task to backfill `createdDate` on Charge and Refund Transactions

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
@@ -1,0 +1,80 @@
+package uk.gov.pay.connector.tasks;
+
+import com.google.inject.persist.Transactional;
+import org.apache.commons.lang3.RandomUtils;
+import org.jboss.logging.MDC;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.dao.ChargeDao;
+import uk.gov.pay.connector.dao.PaymentRequestDao;
+import uk.gov.pay.connector.dao.RefundDao;
+import uk.gov.pay.connector.model.domain.ChargeEntity;
+import uk.gov.pay.connector.model.domain.PaymentRequestEntity;
+import uk.gov.pay.connector.model.domain.transaction.RefundTransactionEntity;
+
+import javax.inject.Inject;
+
+import java.util.Optional;
+
+import static uk.gov.pay.connector.filters.LoggingFilter.HEADER_REQUEST_ID;
+
+public class BackfillCreatedDateOnTransactionsWorker {
+    private final ChargeDao chargeDao;
+    private final PaymentRequestDao paymentRequestDao;
+    private final RefundDao refundDao;
+
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Inject
+    public BackfillCreatedDateOnTransactionsWorker(ChargeDao chargeDao,
+                                                   PaymentRequestDao paymentRequestDao,
+                                                   RefundDao refundDao) {
+        this.chargeDao = chargeDao;
+        this.paymentRequestDao = paymentRequestDao;
+        this.refundDao = refundDao;
+    }
+
+    public void execute(Long startId) {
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
```diff
@@ -0,0 +1,80 @@
...
+
+    @Transactional
+    public void setCreatedDateOnTransactions(long paymentRequestId) {
+        logger.info("Backfilling transactions for payment request [" + paymentRequestId + "]");
+
+        paymentRequestDao.findById(PaymentRequestEntity.class, paymentRequestId)
+                .ifPresent(paymentRequestEntity -> {
+                    final Optional<ChargeEntity> chargeEntity = chargeDao.findByExternalId(paymentRequestEntity.getExternalId());
+                    chargeEntity.ifPresent(charge ->
+                            paymentRequestEntity.getChargeTransaction().setCreatedDate(charge.getCreatedDate())
```
kakumara, line 68: Don't you have to `paymentRequestDao.merge(paymentRequestEntity)` here ? It probably might issue implicit merge, but i normally do it so that no surprises

src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
```diff
@@ -0,0 +1,80 @@
...
+                    );
+                    paymentRequestEntity.getRefundTransactions().forEach(
+                            this::updateRefundTransactionEntity);
+                });
+    }
+
+    private void updateRefundTransactionEntity(RefundTransactionEntity refundTransactionEntity) {
+        final String refundExternalId = refundTransactionEntity.getRefundExternalId();
+        refundDao.findByExternalId(refundExternalId).ifPresent(
+                refundEntity -> refundTransactionEntity.setCreatedDate(refundEntity.getCreatedDate()));
```
kakumara, line 78: same, isn't it safe to issue a `merge` for the `refundTransactionEntity` here?

src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
```diff
@@ -0,0 +1,80 @@
...
+            updateTransactionsWithRetry(paymentRequestId, retries);
+        }
+    }
+
+    private void updateTransactionsWithRetry(long paymentRequestId, long retries) {
+        try {
+            setCreatedDateOnTransactions(paymentRequestId);
+        } catch (Exception exc) {
+            if (retries < 3) {
+                logger.error("Problem migrating [" + paymentRequestId + "] " + exc.getMessage() + " retry count [" + retries + "]");
```
kakumara: How about putting adding the whole `exc` to the logger (2nd argument) rather than just getting the message? So that we'll have the full stacktrace

kakumara (commented):
Other than that 👍

src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
georgeracu: This method is contained into the scope of `@Transactional` from `setCreatedDateOnTransactions` so same as above, `EntityManager` does the `merge` before commit. I don't think we should be explicit about it.

src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
georgeracu: `@Transactional` will make sure to do the merge before committing. We made an explicit `merge` in another task, but is not necessary.

src/main/java/uk/gov/pay/connector/tasks/BackfillCreatedDateOnTransactionsWorker.java
georgeracu: 🍰

kakumara (approved):
LOKTM

PR #550, opened 2018-01-17
PP-2781 Back fill CAPTURE_APPROVED_RETRY events

chrisgrimble:
When we originally migrated charge_events to transaction_events we
thought they were all unique for a charge. However you can have
multiple CAPTURE_APPROVED_RETRY events for a charge. This meant we had
only migrated across the last one. This will migrate across the missing
events.


georgeracu (approved):
LGTM

PR #552, opened 2018-01-19
PP-3181 upgrade jackson-databind

heathd:


chrisgrimble (approved):
LGTM. Need to remember to remove this when we update Dropwizard.

PR #553, opened 2018-01-22
PP-2784 Drop column charge_id from table card_3ds

georgeracu:
- This column has been used for backfilling, which is complete,
so this column can go


chrisgrimble (approved):
LGTM

PR #554, opened 2018-01-23
PP-2783 Drop column charge_id from table cards

georgeracu:
- This column has been used for backfilling, which is complete,
so this column can go


daniele-occhipinti (approved):
LGTM

PR #555, opened 2018-01-23
PP-3079 Add column email to table transactions

georgeracu:
- This is the first step from migrating field `email` from `charges` to
`transactions`

with @sandorarpa


daniele-occhipinti (approved):
LGTM

PR #556, opened 2018-01-23
PP-3176 External ids getting padded with spaces

chrisgrimble:
Some tests are failing as they have an external id with a space at the
end. This appears to be when the external id we generated has less than
26 characters in it. The database stores them in a fixed width column
of 26 chars. JPA trims off any trailing padding but jDBI does not.

- Added padding when we generate the ids so they are always 26 chars
long.


heathd:
looks good 👍

heathd (approved):

PR #557, opened 2018-01-24
PP-3079 Add field `email` to entity ChargeTransaction

georgeracu:
## WHAT
Store `email` into one of the new entities

## HOW 
Add field `email` on `ChargeTransaction` and set this field when the email is set on 
`Charge` (dual-write to both places)

with @sandorarpa

### DIFF
--- a/src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.connector.service;
 
+import com.google.common.collect.ImmutableMap;
 import org.exparity.hamcrest.date.ZonedDateTimeMatchers;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -179,7 +182,47 @@ public void shouldCreateAPaymentRequest() throws Exception{
     }
 
     @Test
-    public void shouldCreateAToken() throws Exception{
+    public void shouldUpdateEmailToChargeTransaction() {
+        service.create(CHARGE_REQUEST, GATEWAY_ACCOUNT_ID, mockedUriInfo);
+
+        ArgumentCaptor<ChargeEntity> chargeEntityArgumentCaptor = forClass(ChargeEntity.class);
+        verify(mockedChargeDao).persist(chargeEntityArgumentCaptor.capture());
```
chrisgrimble: Do we need to use an ArgumentCaptor here? service.create returns an Optional<ChargeResponse> which has the externalId on it.

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -179,7 +182,47 @@ public void shouldCreateAPaymentRequest() throws Exception{
...
+                .withValidPaths(singletonList("email"))
+                .build();
+
+        assertThat(paymentRequestEntity.getChargeTransaction().getEmail(), is(nullValue()));
+
+        service.updateCharge(chargeEntityExternalId, patchRequest);
+
+        verify(mockedPaymentRequestDao).findByExternalId(chargeEntityExternalId);
+        Optional<PaymentRequestEntity> maybePaymentRequest = mockedPaymentRequestDao.findByExternalId(chargeEntityExternalId);
+        assertThat(maybePaymentRequest.isPresent(), is(true));
```
chrisgrimble: Don't think we need to do this as we already have the PaymentRequest above because we created it.

chrisgrimble (changes requested):
Some small changes on the test.

chrisgrimble (approved):
LGTM

PR #559, opened 2018-01-24
PP-3081 Back fill the email column from the charges table to Charge Transactions

chrisgrimble:
Added admin task that can be run under url /migrate-email that will
migrate the email address on the charges table to the email address on
the transactions table for charge transactions.


georgeracu (approved):
LGTM

PR #560, opened 2018-01-25
WIP PP-3173 read payment request when updating email

chrisgrimble:
Change the code that passes round a ChargeEntity to use a
PaymentRequestEntity instead. This is because we are now using the new
PaymentRequest tables and want to start reading from them. Eventually
we will stop updating the Charges table and delete the code. Still need
to keep updating Charges until everything is updated but after this
change that should be trivial.

-Read payment request when updating email.
-Read payment request when getting a charge.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/PersistedCard.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/PersistedCard.java
@@ -62,4 +62,16 @@ public class PersistedCard {
     public void setCardBrand(String cardBrand) {
         this.cardBrand = cardBrand;
     }
+
+    public static PersistedCard from(CardEntity cardEntity, String cardBrand) {
+            PersistedCard card = new PersistedCard();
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/model/domain/PersistedCard.java
```diff
@@ -62,4 +62,16 @@ public String getCardBrand() {
     public void setCardBrand(String cardBrand) {
         this.cardBrand = cardBrand;
     }
+
+    public static PersistedCard from(CardEntity cardEntity, String cardBrand) {
+            PersistedCard card = new PersistedCard();
+            card.setLastDigitsCardNumber(cardEntity.getLastDigitsCardNumber());
+            card.setCardBrand(cardEntity.getCardBrand());
```
chrisgrimble, line 69: Card brand is being set later to the label which is probably what we want no point setting it here too.

PR #561, opened 2018-01-26
PP-2705 Allow transaction searches for multiple card brands

chrisgrimble:
Updated the getChargesJson endpoint so it can now take a list of
cardBrands in the format card_brand=first&card_brand=second.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
@@ -3,7 +3,11 @@ package uk.gov.pay.connector.dao;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
-import org.jooq.*;
+import org.jooq.Condition;
+import org.jooq.SelectConditionStep;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
```diff
@@ -18,7 +22,10 @@
 import java.util.stream.Collectors;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
-import static org.jooq.impl.DSL.*;
+import static org.jooq.impl.DSL.count;
+import static org.jooq.impl.DSL.field;
+import static org.jooq.impl.DSL.selectDistinct;
+import static org.jooq.impl.DSL.table;
```
kakumara, line 28: Are we still using jooq 😖

kakumara (approved):
LOKTM

src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
chrisgrimble: When we swap to using the new transactions table we have to rewrite this so be a good time to get rid of it then.

PR #562, opened 2018-01-29
PP-2705 Allow transaction searches for multiple card brands

chrisgrimble:
Found a bug where you did not select a card-brand in self service. This
submitted the transaction search with a blank card_brand argument and
resulted in a List of length 1 containing "". Search was then looking
for card_brand in ("").

- Filter these blanks in the controller.
- Add a test for card_brand=


georgeracu (approved):
LGTM

PR #566, opened 2018-02-06
PP-3239 Get ready to use new data structure in search

chrisgrimble:
The new search is a bit slower than we would like. One of the biggest
reasons for this is joining payment_requests to transactions when we
have a large amount of data. This is the case if you just load
transactions for a given gateway_account_id (if you search by reference
also on payment_requests you are reducing the amount of rows by enough
for this not to be inefficient). Adding the gatewayAccountId to
TransactionEntities when we add the Transactions to the
PaymentRequestEntity. We probably do not want expose it in the Java, we
will just use it when searching transactions.

- Added the gatewayAccountId to TransactionEntity and set during
creation.
- Added back fill to set gatewayAccountId on TransactionEntities from
PaymentRequestEntity.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
@@ -43,6 +43,10 @@ public abstract class TransactionEntity<S extends Status, T extends TransactionE
     @ManyToOne
     @JoinColumn(name = "payment_request_id", referencedColumnName = "id", updatable = false)
     private PaymentRequestEntity paymentRequest;
+    @Column(name = "gateway_account_id")
+    // This just needs to be set when we save the transaction. It is then used to optimise
+    // transaction search don't access in Java code.
+    private Long gatewayAccountId;
 
     TransactionEntity(TransactionOperation operation) {
         this.operation = operation;
@@ -70,6 +74,12 @@ public abstract class TransactionEntity<S extends Status, T extends TransactionE
 
     public void setPaymentRequest(PaymentRequestEntity paymentRequest) {
         this.paymentRequest = paymentRequest;
+        this.gatewayAccountId = paymentRequest.getGatewayAccount().getId();
+    }
+
+    //todo remove this once we have back filled the gatewayAccountId.
+    public void setGatewayAccountId(Long gatewayAccountId) {
+        this.gatewayAccountId = gatewayAccountId;
     }
 
     public TransactionOperation getOperation() {
--- a/src/main/java/uk/gov/pay/connector/tasks/BackfillGatewayAccountIdOnTransactionsWorker.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/BackfillGatewayAccountIdOnTransactionsWorker.java
@@ -1,0 +1,67 @@
+package uk.gov.pay.connector.tasks;
+
+import com.google.inject.persist.Transactional;
+import org.apache.commons.lang3.RandomUtils;
+import org.jboss.logging.MDC;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.dao.ChargeDao;
+import uk.gov.pay.connector.dao.PaymentRequestDao;
+import uk.gov.pay.connector.dao.RefundDao;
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/tasks/BackfillGatewayAccountIdOnTransactionsWorker.java
```diff
@@ -0,0 +1,67 @@
...
+            updateTransactionsWithRetry(paymentRequestId, retries);
+        }
+    }
+
+    private void updateTransactionsWithRetry(long paymentRequestId, long retries) {
+        try {
+            setGatewayAccountOnTransactions(paymentRequestId);
+        } catch (Exception exc) {
+            if (retries < 3) {
+                logger.error("Problem migrating [" + paymentRequestId + "] " + " retry count [" + retries + "]", exc );
```
alexbishop1, line 48: Since this is basically one-shot code, I’m not going to withhold approval on this, but SLF4J has in-built support for formatting strings and can do smart stuff like avoid concatenating strings if the log line isn’t actually going to be output:

```
logger.error("Problem migrating [{}] retry count [{}]", paymentRequestId, retries, exc);
```

src/test/java/uk/gov/pay/connector/tasks/BackfillGatewayAccountIdOnTransactionsWorkerITest.java
```diff
@@ -0,0 +1,165 @@
...
+        addRefundTransactions(ids.getLeft(), testRefund);
+        worker.execute(1L);
+        final List<Map<String, Object>> refundTransactions = databaseTestHelper.getRefundTransactions(ids.getLeft());
+        refundTransactions.forEach(refundTransaction ->
+                assertThat(refundTransaction.get("gateway_account_id"), is(expectedGatewayAccountId))
+        );
+    }
+
+    @Test
+    public void shouldAddDateToMultipleRefundTransactions() {
```
alexbishop1: `Date`?

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
```diff
@@ -70,6 +74,12 @@ public PaymentRequestEntity getPaymentRequest() {
 
     public void setPaymentRequest(PaymentRequestEntity paymentRequest) {
         this.paymentRequest = paymentRequest;
+        this.gatewayAccountId = paymentRequest.getGatewayAccount().getId();
```
alexbishop1, line 77: Might be nice to have a test to check this behaviour (though I appreciate it will be hard with it being a private field)

alexbishop1 (changes requested):
Looks good (when the backfill is run, it will overwrite the new transactions that already have a `gatewayAccountId` but this is OK because they’ll be the same)

The commit and PR descriptions could do with tweaking (we’re not using the new data structure for search yet, just adding the field and backfill task) and there’s one place where I think we could maybe add a test

src/main/java/uk/gov/pay/connector/model/domain/transaction/TransactionEntity.java
chrisgrimble: Yep not sure unless we expose it. I did write the test then had to remove it as I think it is important it is not exposed.

src/main/java/uk/gov/pay/connector/tasks/BackfillGatewayAccountIdOnTransactionsWorker.java
chrisgrimble: Yep one shot code also as we do have error enabled at the moment. According to the docs if you do have error enabled it's slightly more expensive to do the var args way.

"However, this variant incurs the hidden (and relatively small) cost of creating an <code>Object[]</code> before invoking the method"

src/test/java/uk/gov/pay/connector/tasks/BackfillGatewayAccountIdOnTransactionsWorkerITest.java
chrisgrimble: Oops

alexbishop1 (approved):
Update the PR title and this is good to go

PR #568, opened 2018-02-08
PP-3239 use new data structure in search

chrisgrimble:
Add a new end point /api/accounts/{accountId}/transactions that will
use the new data structure to produce a list of transactions.

- Wrote dao that will search new data structure takes all the old
parameters from the previous search.
- As part of another branch added gateway_account_id on transactions.
The used this gateway_account_id in the search as it is much quicker
than joining to transactions to payment_requests.
- Only join payment_reqyests, cards and cards_3ds tables when needed.
The sort of searches that need these tables will only return a few
results. If we have a large number of results its expensive to join
on these tables and the data is not needed for the search results.
- Split search into two a count to get the total results with limit
and offset to give one page of data.
- Added new end point.
- Duplicated tests that were calling the old endpoint to test the new
one. This resulted in a number of small bug fixes.
- Load CardTypes once for a search result and use this information to
create the objects returned by connector.


kakumara (approved):
LOKTM

PR #569, opened 2018-02-08
PP-3239 Rename TransactionDao and TransactionSearchStrategy

chrisgrimble:
As we now have a table called Transactions that contains both charges
and refunds, we created a Java class called TransactionEntity. When we
went to create the DAO and search strategy for this entity we found
there was a TransactionDao and TransactionsSearchStrategy. This was
used by the old search which combined items from the Charges and
Refunds tables. Renamed these to OldTransactionDAO and
OldTransactionSearchStrategy.


kakumara (approved):
It seems there's no unit tests for the TransactionSearchStrategy 😞  (coz I cannot see any test changes). But i guess its tested transitively.

PR #570, opened 2018-02-09
PP-3239 Fix logging of searches

chrisgrimble:
Used the varargs version of logger.info. Made an incorrect assumption
that this used the same replacement patterns as String.format but
instead it used {}. Corrected this so we are now logging our searches.


kakumara (approved):
LGTM

PR #571, opened 2018-02-12
PP-3239 Add index to transactions.created_date

chrisgrimble:
Searching by created_date so adding an index for it.


kakumara:
can this cause a prod DB slowdown? perhaps we should deploy this during a quiet window? (just asking, coz I haven't done this type of thing)

kakumara (approved):
LOKTM

PR #573, opened 2018-02-13
PP-3239 Add index to transactions.operation

chrisgrimble:
Searching by operation on transactions so adding an index for it.


georgeracu (approved):
LGTM

PR #574, opened 2018-02-13
PP-3239 Experimental search api was not mapping its Java objects correctly

chrisgrimble:
- Fixed mapping.
- Added test that uses the api.


georgeracu (approved):
LGTM

PR #575, opened 2018-02-14
PP-3239 Experimental search api was not mapping its Java object

chrisgrimble:
correctly

- Hard coded limit and offset as this does not appear to be being set.


georgeracu (approved):
LGTM

PR #577, opened 2018-02-15
PP-3239 Merge Experimental search and TransactionSearch

chrisgrimble:
- Updated TransactionDao to use a combination of the Transaction search
which could handle search parameters and the Experimental Search which
maps to a flat Java object.
- Used ILIKE instead of LIKE in reference and email search.
- Updated test and removed checks for returnUrl and provider name as
they are no longer returned and not need by self-service.


georgeracu (dismissed):
Looks good

georgeracu (approved):
LGTM

PR #580, opened 2018-02-19
Inline all resource path constants

heathd:
The API paths form a primary part of the external API of the service.
Using constants to declare those paths makes it harder to read the
resource classes and understand what the actual endpoints are. I have
often found myself jumping back and forth between the API paths file and
the resource file to try to figure out which URL is used to trigger a
particular resource.

The only place apart from the resources where the paths were used is in
the tests. This effectively makes the checks tautologous -- in other
words you could change the path constant and the tests would still pass,
but in fact you would have broken dependent apps. This feels like a
disadvantage to me.

So, I think having paths inline in the resources (and explicitly stated
in the tests) makes things clearer, easier to read and simpler.

What do you think?


juanjoqmelian:
I totally agree. I've been having the same problem every time that I wanted to find an endpoint URL. 👍

rhowe-gds (approved):
Completely agree with the sentiment of this PR; haven't checked the details

alexbishop1 (approved):
I think I agree. It certainly makes `@Path` annotations easier to read. I was initially slightly concerned about the URLs included in JSON responses not sharing a constant with the `@Path` ones (because one could change one but forget to change the other) but all these URLs are public anyway (in the sense that other external or internal services rely on them), so we’ve already got to be really careful about changing them.

ckalista:
YES! Not just that, but it also became convoluted to come up with variable names that made sense. Thanks you

PR #582, opened 2018-02-23
PP-3382 adding `v2` transactions list resource

kakumara:
## WHAT
v1 resource currently (temporarily) supporting two variations of transactions to be retrieved based on featureFlag availability. This takes the feature flag enabling bit out in to a new resources so that we can remove the whole feature flag enablement from v1.
For now `selfservice` is the only intended client of it. And by default selfservice is required use the v2 version of the result set.


georgeracu (approved):
One small comment is on avoiding wildcard imports

kakumara:
I'll fix my intelliJ. and that should be fixed  in the next PR

PR #583, opened 2018-02-28
Add @Timed annotations to search endpoints for metric goodness

alexbishop1:
Searches can be kind of slow so it would be nice to get metrics on how long they take


juanjoqmelian (approved):
LGTM

PR #585, opened 2018-03-07
PP-3381 Instrument all stages of a master build and deployment to Test

belindac:
Use new groovy scripts with metric integration

solo @belindac

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,8 +28,19 @@ pipeline {
         }
       }
       steps {
-        sh 'docker pull govukpay/postgres:9.4.4'
-        sh 'mvn clean package'
+        script {
+          def long stepBuildTime = System.currentTimeMillis()
+
+          sh 'docker pull govukpay/postgres:9.4.4'
+          sh 'mvn clean package'
+
+          postSuccessfulMetrics("connector.maven-build", stepBuildTime)
+        }
+      }
+      post {
+        failure {
+          postMetric("connector.maven-build.failure", 1, "new")
+        }
       }
     }
     stage('Maven Build Without Tests') {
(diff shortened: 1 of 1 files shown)


tlwr (dismissed):

Jenkinsfile
```diff
@@ -18,6 +18,8 @@ pipeline {
   environment {
     DOCKER_HOST = "unix:///var/run/docker.sock"
     RUN_END_TO_END_ON_PR = "${params.runEndToEndOnPR}"
+    HOSTED_GRAPHITE_ACCOUNT_ID = credentials('graphite_account_id')
+    HOSTED_GRAPHITE_API_KEY = credentials('graphite_api_key')
```
mrlumbu: why do we need to add these here rather than in jenkins library?

Jenkinsfile
```diff
@@ -31,6 +33,14 @@ pipeline {
         sh 'docker pull govukpay/postgres:9.4.4'
         sh 'mvn clean package'
       }
+      post {
+        failure {
+          postMetric("connector.maven-build.failure", 1, "new")
+        }
+        success {
+          postSuccessfulMetrics("connector.maven-build")
+        }
```
mrlumbu: this is fine, but relies on fact that this step comes first. If we were to change order, or add an extra step this would no longer be correct, but it wouldn't be obvious. I think it might be better to send the time value explicitly.

Jenkinsfile
```diff
@@ -28,8 +28,18 @@ pipeline {
         }
       }
       steps {
+        script {
+          def long stepBuildTime = System.currentTimeMillis()
+        }
         sh 'docker pull govukpay/postgres:9.4.4'
         sh 'mvn clean package'
+        postSuccessfulMetrics("connector.maven-build", stepBuildTime)
+        exit 1
```
mrlumbu: exit 1?

mrlumbu: ah you're just testing it

mrlumbu (approved):

PR #586, opened 2018-03-08
PP-3450 EPDQ 3Ds Store request data

chrisgrimble:
When we make an authorise request to EPDQ and 3Ds is enabled we will
not get a pa_request or issure_url. Instead we get a block of html
which we will need to send to the user and will redirect them to EPDQ.

- Add a html_out column to charges and card_3ds to store the html.
- Make pa_request and issurer_url nullable on card_3ds as we will have
them or html_out.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1281,4 +1281,25 @@
         </sql>
     </changeSet>
 
+    <changeSet id="add html_out_3ds to charges table" author="">
+        <addColumn tableName="charges">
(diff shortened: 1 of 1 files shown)


rhowe-gds:
What proportion of charges will have this data? If it's not many, a separate table might be better in order to keep `charges` lean

chrisgrimble:
Very few will have this data, but the separate table is one of the things the database refactor brought in and that work is currently on ice.

kakumara (approved):
LGTM

PR #587, opened 2018-03-08
Pp 3450 epdq 3ds auth request

chrisgrimble:
Create and handle an EPDQ 3ds authorise request. The request needs a extra fields on it to tell EPDQ the request is a 3ds authorise request. This includes some callback url that EPDQ will send the user to once 3ds has been completed. For now these are not confirmed and may need to change when we do the frontend logic for this. EPDQ will then return a response with a status of 46 and a new element containing some html. The html needs to be save. Then later on frontend will make a new request for the charge data and this needs to include the html.

- Mapped the correct status code.
- created a separated request builder for authorise 3ds (putting non-existent frontend URLs for now).
- Map the response from EPDQ and add the new html to the entities above.
- Add html to the response from connector.


src/main/java/uk/gov/pay/connector/model/Auth3dsDetailsDTO.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.pay.connector.model;
+
+public class Auth3dsDetailsDTO {
```
alexbishop1: This name is a bit confusing because we also have `Auth3dsDetails`, which is only used for 3D Secure for Worldpay (but could potentially be used by other payment gateways that have a Worldpayesque 3D Secure flow). Perhaps we could clarify things by renaming `Auth3dsDetails` (and its friends`Auth3dsResponseGatewayRequest`, which `Card3dsResponseAuthService`) to something that indicates they’re basically Worldpay-specific.

But I digress. This `Auth3dsDetailsDTO` class will only ever have `issuerUrl` and `paRequest` set, or `htmlOut` set. Could we perhaps make this into an interface, called, say, `GatewayParamsFor3DSecure` (since that’s what it is: the parameters we need from the gateway to set up the 3D Secure authentication process for the user) and then have concrete `WorldpayParamsFor3DSecure` and `EpdqParamsFor3DSecure` implementations?

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
```diff
@@ -144,6 +145,9 @@ public String toString() {
         @JsonProperty("issuerUrl")
         private String issuerUrl;
 
+        @JsonProperty("htmlOut")
```
alexbishop1, line 148: Devoid of its 3D Secure context, the name `htmlOut` is a bit vague (to be fair, `issuerUrl` has the same problem)

src/main/java/uk/gov/pay/connector/model/domain/Card3dsEntity.java
```diff
@@ -95,17 +106,26 @@ public static Card3dsEntity from(ChargeEntity chargeEntity) {
...
+
+        Card3dsEntity that = (Card3dsEntity) o;
+
+        if (id != null ? !id.equals(that.id) : that.id != null) return false;
+        if (paRequest != null ? !paRequest.equals(that.paRequest) : that.paRequest != null) return false;
+        if (issuerUrl != null ? !issuerUrl.equals(that.issuerUrl) : that.issuerUrl != null) return false;
+        if (worldpayMachineCookie != null ? !worldpayMachineCookie.equals(that.worldpayMachineCookie) : that.worldpayMachineCookie != null)
+            return false;
+        if (htmlOut != null ? !htmlOut.equals(that.htmlOut) : that.htmlOut != null) return false;
+        return chargeTransactionEntity != null ? chargeTransactionEntity.equals(that.chargeTransactionEntity) : that.chargeTransactionEntity == null;
```
alexbishop1, line 118: What was wrong with `Objects.equal`?

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetailsEntity.java
```diff
@@ -12,6 +14,18 @@
     @Column(name = "issuer_url_3ds")
     private String issuerUrl;
 
+    @Column(name = "html_out_3ds")
+    private String htmlOut;
+
+    // needed for JPA
+    public Auth3dsDetailsEntity(){}
+
+    public Auth3dsDetailsEntity(Auth3dsDetailsDTO auth3dsDetails) {
```
alexbishop1: Having this variable called `auth3dsDetails` when it’s an `Auth3dsDetailsDTO` object and not an `Auth3dsDetails` object is going to confuse and confound people

src/main/java/uk/gov/pay/connector/service/BaseAuthoriseResponse.java
```diff
@@ -1,16 +1,17 @@
...
 public interface BaseAuthoriseResponse extends BaseResponse {
 
     String getTransactionId();
 
     AuthoriseStatus authoriseStatus();
 
-    String get3dsPaRequest();
-
-    String get3dsIssuerUrl();
+    Optional<Auth3dsDetailsDTO> getAuth3dsDetails();
```
alexbishop1: This is nicer. Yay for encapsulation!

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
```diff
@@ -36,17 +43,18 @@ public AuthoriseStatus authoriseStatus() {
             return AuthoriseStatus.REJECTED;
         }
 
+        if(WAITING_3DS.equals(status)) {
```
alexbishop1: I think we’ve reached the point where this should become a `switch` statement

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
```diff
@@ -36,17 +43,18 @@ public AuthoriseStatus authoriseStatus() {
...
-    public String get3dsPaRequest() {
-        return null;
-    }
-
-    @Override
-    public String get3dsIssuerUrl() {
-        return null;
+    public Optional<Auth3dsDetailsDTO> getAuth3dsDetails() {
+        if (htmlAnswer != null) {
+            return Optional.of(new Auth3dsDetailsDTO(null, null, htmlAnswer));
```
alexbishop1: Overloaded constructor to avoid this null-ugliness? Or maybe that subclassing idea I suggested in an earlier comment

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
```diff
@@ -72,6 +84,10 @@ public String getErrorMessage() {
         return null;
     }
 
+    public String getStatus() {
```
alexbishop1: This seems to only be used in tests and equivalent information can be retrieved using the `authoriseStatus` method

src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
```diff
@@ -20,6 +20,7 @@
         private String userId;
         private String shaInPassphrase;
         private String amount;
+        private String frontendUrl;
```
alexbishop1, line 23: `frontendUrl` for that?

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
```diff
@@ -0,0 +1,68 @@
...
+    final static String OWNER_TOWN_KEY = "OWNERTOWN";
+    final static String OWNER_ZIP_KEY = "OWNERZIP";
+    final static String PSPID_KEY = "PSPID";
+    final static String PSWD_KEY = "PSWD";
+    final static String USERID_KEY = "USERID";
+
+    @Override
+    public ImmutableList<NameValuePair> extract(EpdqTemplateData templateData) {
+
+        String epdqCallbackUrl = String.format("%s/card_details/%s/3ds_required_in", templateData.getFrontendUrl(), templateData.getOrderId());
```
alexbishop1: This variable exists in a class with `Epdq` in the name: it doesn’t need to be prefixed with `epdq`. Also, it’s a redirect URL, rather than a callback URL.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
```diff
@@ -0,0 +1,68 @@
...
+        return newParameterBuilder()
+                .add("ACCEPTURL", epdqCallbackUrl)
+                .add(AMOUNT_KEY, templateData.getAmount())
+                .add(CARD_NO_KEY, templateData.getAuthCardDetails().getCardNo())
+                .add(CARDHOLDER_NAME_KEY, templateData.getAuthCardDetails().getCardHolder())
+                .add("COMPLUS", "")
+                .add(CURRENCY_KEY, "GBP")
+                .add(CVC_KEY, templateData.getAuthCardDetails().getCvc())
+                .add("DECLINEURL", epdqCallbackUrl + "/decline")
+                .add("EXCEPTIONURL", epdqCallbackUrl + "/exception")
```
alexbishop1: Not sure we want to put a term like `exception` in our URLs — `error`, maybe?

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
```diff
@@ -0,0 +1,68 @@
...
+                .add(USERID_KEY, templateData.getUserId())
+                .add("WIN3DS", "MAINW")
+                .build();
+    }
+
+    private static String concatAddressLines(String addressLine1, String addressLine2) {
+        return StringUtils.isBlank(addressLine2) ? addressLine1 : addressLine1 + ", " + addressLine2;
+    }
+
+}
```
alexbishop1, line 63: This class is very similar to `EpdqPayloadDefinitionForNewOrder` and it’s likely changes to that will require changes to this class. Can we somehow extend what we already have?

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
```diff
@@ -52,11 +53,13 @@
     public static final String ROUTE_FOR_MAINTENANCE_ORDER = "maintenancedirect.asp";
 
     private final SignatureGenerator signatureGenerator;
+    private final String frontendUrl;
```
alexbishop1, line 56: Can you name this field such that its purpose of this field is clearer?

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
```diff
@@ -148,26 +151,33 @@ public ExternalChargeRefundAvailability getExternalChargeRefundAvailability(Char
...
-                .withMerchantCode(request.getGatewayAccount().getCredentials().get(CREDENTIALS_MERCHANT_ID))
-                .withDescription(request.getDescription())
-                .withAmount(request.getAmount())
-                .withAuthorisationDetails(request.getAuthCardDetails())
-                .build();
+    private Function<AuthorisationGatewayRequest, GatewayOrder> buildAuthoriseOrderFor(String frontendUrl) {
+        return request -> {
+            EpdqOrderRequestBuilder epdqOrderRequestBuilder = anEpdqAuthoriseOrderRequestBuilder();
+            if (request.getGatewayAccount().isRequires3ds()) {
+                epdqOrderRequestBuilder = anEpdq3DsAuthoriseOrderRequestBuilder(frontendUrl);
```
alexbishop1: Ternary operator maybe?

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -145,8 +144,11 @@ public String toString() {
         if (StringUtils.isNotBlank(getRefusedReturnCodeDescription())) {
             joiner.add("ISO8583ReturnCode description: " + getRefusedReturnCodeDescription());
         }
-        if (StringUtils.isNotBlank(get3dsIssuerUrl())) {
-            joiner.add("issuerURL: " + get3dsIssuerUrl());
+        if (StringUtils.isNotBlank(issuerUrl)) {
+            joiner.add("issuerURL: " + issuerUrl);
+        }
+        if (StringUtils.isNotBlank(paRequest)) {
+            joiner.add("issuerURL: " + paRequest);
```
alexbishop1: The label’s wrong here! Also, it’s just an opaque string — is there any value to outputting this?

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -66,8 +65,7 @@
     private Histogram mockHistogram;
     private Counter mockCounter;
 
-    @Before
-    public void setUpAndCheckThatEpdqIsUp() {
+    private void setUpAndCheckThatEpdqIsUp(boolean require3ds) {
```
alexbishop1: Not fond of using an `boolean` for this: how about differently-named methods (that mostly defer to a common third method)?

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -208,7 +222,7 @@ private static RefundGatewayRequest buildRefundRequest(ChargeEntity chargeEntity
     }
 
     private static AuthCardDetails aValidEpdqCard() {
-        String validSandboxCard = "5555444433331111";
+        String validSandboxCard = "4000000000000002";
```
alexbishop1, line 234: Why was this changed? Also, `validSandboxCard` should probably be `validEpdqCard` (I appreciate this was like this before you got there)

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
```diff
@@ -1,14 +1,11 @@
...
 import org.hamcrest.Matchers;
 import org.junit.Test;
 import uk.gov.pay.connector.it.base.ChargingITestBase;
 
-import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_ERROR;
-import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
-import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_SUBMITTED;
-import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
-import static uk.gov.pay.connector.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.*;
```
alexbishop1: We’ve standardised on non-wildcard imports

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
```diff
@@ -205,6 +210,25 @@ public void doAuthorise_shouldRespondWith3dsResponseFor3dsOrders() {
...
+
+        assertThat(response.isSuccessful(), is(true));
+
+        assertThat(charge.getStatus(), is(AUTHORISATION_3DS_REQUIRED.getValue()));
+        verify(mockedChargeEventDao).persistChargeEventOf(charge, Optional.empty());
+        assertThat(charge.get3dsDetails().getHtmlOut(), is(notNullValue()));
+        Card3dsEntity card3ds = paymentRequest.getChargeTransaction().getCard3ds();
+        assertThat(card3ds.getIssuerUrl(), is(nullValue()));
+        assertThat(card3ds.getPaRequest(), is(nullValue()));
+        assertThat(card3ds.getHtmlOut(), is(notNullValue()));
```
alexbishop1: Can’t we test for an actual value?

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
```diff
@@ -542,6 +566,21 @@ private void providerWillRequire3ds(String sessionIdentifier) {
         when(mockedPaymentProvider.generateTransactionId()).thenReturn(Optional.of(TRANSACTION_ID));
     }
 
+    private void providerWillRequireEpdq3ds() throws XMLUnmarshallerException {
```
alexbishop1: This could do with a better name like `epdqProviderWillRequire3ds` — also, `providerWillRequire3ds` should become something like `worldpayProviderWillRequire3ds`

src/test/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProviderTest.java
```diff
@@ -73,18 +73,7 @@
...
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_CANCEL_REQUEST;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_CANCEL_SUCCESS_RESPONSE;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_CAPTURE_ERROR_RESPONSE;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_CAPTURE_SUCCESS_RESPONSE;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_DELETE_SUCCESS_RESPONSE;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_NOTIFICATION_TEMPLATE;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_REFUND_ERROR_RESPONSE;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_REFUND_REQUEST;
-import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_REFUND_SUCCESS_RESPONSE;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.*;
```
alexbishop1: No wildcard imports!

src/test/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProviderTest.java
```diff
@@ -174,6 +163,16 @@ public void shouldAuthorise() {
         assertThat(response.getBaseResponse().get().getTransactionId(), is("3014644340"));
     }
 
+    @Test
+    public void shouldAuthorise3DRequest() {
```
alexbishop1: `3ds`

src/test/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProviderTest.java
```diff
@@ -174,6 +163,16 @@ public void shouldAuthorise() {
...
+        when(mockGatewayAccountEntity.isRequires3ds()).thenReturn(true);
+        mockPaymentProviderResponse(200, successAuth3dResponse());
+        GatewayResponse<EpdqAuthorisationResponse> response = provider.authorise(buildTestAuthorisationRequest());
+        verifyPaymentProviderRequest(successAuthRequest());
+        assertTrue(response.isSuccessful());
+        assertThat(response.getBaseResponse().get().getStatus(),is("46"));
```
alexbishop1: Assert on `authoriseStatus()` here please

alexbishop1 (changes requested):

src/test/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProviderTest.java
kakumara: 👍

src/test/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProviderTest.java
kakumara: 👍

src/test/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProviderTest.java
kakumara: :|

kakumara: 😑

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
kakumara: 👍

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
kakumara: could do: But will pollute the code a bit having to add 2 extra setters (just for testing) 😑

src/main/java/uk/gov/pay/connector/service/BaseAuthoriseResponse.java
kakumara: 🎉

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
kakumara: well, its inside a `Auth3dsData` context. So clients will read as `auth_3ds_data.htmlOut`. We tried to comply to the `out` and `in` concept of 3ds handover in frontend so this being the html form for `out` going. We couldn't think of a better name to be honest !

src/main/java/uk/gov/pay/connector/model/domain/Card3dsEntity.java
kakumara: This is intellij auto-generated. Do we care? 🤔

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
kakumara: yes, its needed in order to tell epdq, what are the redirection points to send the user back to

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
kakumara: renamed to `frontend3dsIncomingUrl`

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
kakumara: 👍 (_although i hate inheritence_)

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
kakumara: This is the same of config's `frontendUrl` variable. I guess its global and known to everyone everywhere. the only thing I can think of is `frontendBaseUrl` but not sure if that serves any extra purpose.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
kakumara: 😑  really? okay 👍

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayOrderStatusResponse.java
kakumara: probably not. But it has always been there. So I'm going to leave this refactoring for another day 
Fixed the label!

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
kakumara: 👍

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
kakumara: This was needed. This card works for both 3ds and non-3ds scenarios.

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
alexbishop1: Missed that it’s in an inner class: in that case, the name is fine

src/main/java/uk/gov/pay/connector/model/EpdqParamsFor3DSecure.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.pay.connector.model;
+
+import uk.gov.pay.connector.model.domain.Auth3dsDetailsEntity;
+
+public class EpdqParamsFor3DSecure implements GatewayParamsFor3DSecure {
```
alexbishop1, line 5: OK, I know I suggested this name but it seems our convention is to use `3ds` rather than `3DSecure` in class names, so please do that

src/main/java/uk/gov/pay/connector/model/EpdqParamsFor3DSecure.java
```diff
@@ -0,0 +1,19 @@
...
+
+    public final String htmlOut;
```
alexbishop1, line 7: `public`?

src/main/java/uk/gov/pay/connector/model/WorldpayParamsFor3DSecure.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.pay.connector.model;
+
+import uk.gov.pay.connector.model.domain.Auth3dsDetailsEntity;
+
+public class WorldpayParamsFor3DSecure implements GatewayParamsFor3DSecure {
+
+    public final String issuerUrl;
+    public final String paRequest;
```
alexbishop1, line 8: Stop airing our dirty laundry in `public`!

src/main/java/uk/gov/pay/connector/model/domain/Card3dsEntity.java
alexbishop1: I actually have lots of opinions about the perils of autogenerated `equals`, `hashCode` and `toString` methods but that’s perhaps something for the pub.

In this case, I’m OK with it.

src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
```diff
@@ -160,7 +160,7 @@ public NotificationCredentials getNotificationCredentials() {
         return notificationCredentials;
     }
 
-    public boolean isRequires3ds() {
+    public boolean requires3ds() {
```
alexbishop1, line 163: I like this renaming

src/main/java/uk/gov/pay/connector/service/BaseAuthoriseResponse.java
```diff
@@ -1,16 +1,17 @@
...
 public interface BaseAuthoriseResponse extends BaseResponse {
 
     String getTransactionId();
 
     AuthoriseStatus authoriseStatus();
 
-    String get3dsPaRequest();
-
-    String get3dsIssuerUrl();
+    Optional<? extends GatewayParamsFor3DSecure> getAuth3dsDetails();
```
alexbishop1, line 14: Rename this method please.

I would really to not use the wildcard `?` type here (so we’d make the class be `BaseAuthoriseResponse<T>` where `T extends GatewayParamsFor3DSecure`, `getAuth3dsDetails()` would return `T` and we’d have `EpdqAuthorisationResponse implements  BaseAuthoriseResponse<EpdqParamsFor3DSecure>` etc.).

But I know adding generics like this can be like tugging a loose thread, so I’m not going to insist on it. But maybe try? You have my permission to revert after ten minutes if it’s turning into a giant mess.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
```diff
@@ -36,17 +40,18 @@ public AuthoriseStatus authoriseStatus() {
             return AuthoriseStatus.REJECTED;
         }
 
+        if (WAITING_3DS.equals(status)) {
```
alexbishop1, line 43: I still think this should be a `switch` statement

src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
alexbishop1: I’m still not too clear on what it actually means. What actual value would this have? I think we can tease out a less generic name.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
```diff
@@ -0,0 +1,63 @@
...
+    static final String HTTPACCEPT_URL = "HTTP_ACCEPT";
+    static final String HTTPUSER_AGENT_URL = "HTTP_USER_AGENT";
+    static final String LANGUAGE_URL = "LANGUAGE";
+    static final String PARAMPLUS_URL = "PARAMPLUS";
+    static final String WIN3DS_URL = "WIN3DS";
+
+    @Override
+    public ImmutableList<NameValuePair> extract(EpdqTemplateData templateData) {
+
+        String frontend3dsIncomingUrl = String.format("%s/card_details/%s/3ds_required_in", templateData.getFrontendUrl(), templateData.getOrderId());
```
alexbishop1, line 27: `frontendUrlToRedirectUserToAfterTheyComplete3dsAuthentication`, maybe? It’s long but it makes it really clear where this URL is used in the 3D Secure flow (I get confused by all this in, out, shake-it-all-about stuff).

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
alexbishop1: So are you going to do that then? 😃

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
alexbishop1: If this never changes, I wonder if we could somehow use a constant to make it clearer…

src/main/java/uk/gov/pay/connector/service/worldpay/WorldpayOrderStatusResponse.java
alexbishop1: What I meant was that the `paRequest` was not included in the `toString()` previously and I don’t think there’s much reason to add it

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
alexbishop1: Fair enough on the value. Variable name still needs renaming though!

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
```diff
@@ -34,6 +36,24 @@ public void shouldAuthorise_whenTransactionIsSuccessful() throws Exception {
...
+        String chargeId = createNewChargeWithNoTransactionId(ENTERING_CARD_DETAILS);
+        epdq.mockAuthorisation3dsSuccess();
+
+        ValidatableResponse response = givenSetup()
+                .body(authorisationDetails)
+                .post(authoriseChargeUrlFor(chargeId))
+                .then();
+
+        response
+                .body("status", Matchers.is(AUTHORISATION_3DS_REQUIRED.toString()))
```
alexbishop1, line 51: Static import `is` please (more readable)

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
alexbishop1: But is this a Base64-encoded value? Maybe the name could indicate that.

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
alexbishop1: But you did it anyway! 😃

alexbishop1 (changes requested):
Looks better! Thanks for making the changes. Just a few things remaining.

alexbishop1 (approved):
As discussed, since none of my requests are show-stoppers, I will approve this in order that you can get this merged and unblock other things, on the condition you address the remaining comments in a subsequent PR.

kakumara:
I'll address the remaining in the next PR

src/main/java/uk/gov/pay/connector/model/EpdqParamsFor3DSecure.java
kakumara: 🤕 👍

src/main/java/uk/gov/pay/connector/model/EpdqParamsFor3DSecure.java
kakumara: it was intentional. (reasoning below) 👍

src/main/java/uk/gov/pay/connector/model/WorldpayParamsFor3DSecure.java
kakumara: It was deliberate . Note it is `final` and it is a DTO. This avoid the clutters of getters(). 

Although, in this case I've refactored the uses of them to use other means for testing (so removed public). But this particular aspect is something I would vote over having to  generate clutters of getters and setters on DTO classes.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
kakumara: So on this case, this name is consistent with the existing 3ds implementation flow. There is a concept of `3ds_out` and `3ds_in` already spanned across connector and frontend. This refers to the same `in` url. we believe whoever going to work on 3ds should be aware of these 2 aspects of 3ds? Happy to discuss.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
kakumara: This is a relative opinion. I've changed it to a `switch` but in this case we feel if statements are more readable than switch.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
kakumara: I've renamed to `frontendBaseUrl`.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
kakumara: done it

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPaymentProvider.java
kakumara: This is not a constant unfortunately. Its set by dependency injection at startup

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
kakumara: of course 😖 👍

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/service/BaseAuthoriseResponse.java
kakumara: Yes. I went through the exact same thought and got tangled my self. It all fine until SandboxAuthoriseResponse and SmartpayAuthoriseResponse cannot extend with a type because they don't exist (at least yet). so given that this seems to be the closest i can get to.

src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
kakumara: but that broke the build 😢  🔴

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
kakumara: Changed my mind (after chatting with couple of others) , the switch more unreadable (relatively) compared to the if's here. Particularly when there's a `or`/ `and` involved in somewhere.

src/main/java/uk/gov/pay/connector/model/WorldpayParamsFor3DSecure.java
alexbishop1: Fair enough; thanks for explaining. `public` fields are less of a concern if they’re immutable.

src/main/java/uk/gov/pay/connector/service/BaseAuthoriseResponse.java
alexbishop1: Well, thanks for trying!

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
alexbishop1: I think the _out_ and _in_ stuff has mostly been restricted to frontend so far (and I never found those terms that descriptive anyway). But you’re right that it is a convention, so I’m happy to leave this as it is.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForNew3dsOrder.java
alexbishop1: Sorry for not noticing you’d started extending `EpdqPayloadDefinitionForNewOrder`. There’s still quite a bit of duplication though, which it would be nice to address (in particular things like the `concatAddressLines` method being in both classes).

src/main/java/uk/gov/pay/connector/model/EpdqParamsFor3DSecure.java
alexbishop1: This has been done in https://github.com/alphagov/pay-connector/pull/588/files

src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
alexbishop1: This has been done in https://github.com/alphagov/pay-connector/pull/588/files

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
alexbishop1: This has been done in https://github.com/alphagov/pay-connector/pull/588/files

src/main/java/uk/gov/pay/connector/model/WorldpayParamsFor3DSecure.java
alexbishop1: This has been done in https://github.com/alphagov/pay-connector/pull/588/files

PR #588, opened 2018-03-09
PP-3450 handle epdq 3ds result

kakumara:
This needs to wait until https://github.com/alphagov/pay-connector/pull/587
## WHAT
Using the same transaction flow (pre, op, post) so that we stick to the same state transitions. The main difference is that we construct a success authorisation response in the `operation` (and not invoking any epdq gateway operation as there i no need) so that we can fit in to the same 3ds flow.

Introduced a `auth_3d_result` payload entry which will only be used by epds 3ds flow.

with @chrisgrimble

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
@@ -4,7 +4,13 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class Auth3dsDetails implements AuthorisationDetails {
 
+    public enum Auth3DResult {
+        AUTHORISED, DECLINED, ERROR
+    }
+
     private String paResponse;
+
+    private Auth3DResult auth3DResult;
 
     @JsonProperty("pa_response")
     public void setPaResponse(String paResponse) {
@@ -15,4 +21,12 @@ public class Auth3dsDetails implements AuthorisationDetails {
             return paResponse;
         }
 
+    public Auth3DResult getAuth3DResult() {
+        return auth3DResult;
+    }
+
+    @JsonProperty("auth_3d_result")
+    public void setAuth3DResult(String auth3DResult) {
+        this.auth3DResult = auth3DResult == null ? null : Auth3DResult.valueOf(auth3DResult);
+    }
 }
--- a/src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
@@ -4,13 +4,13 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 public class Auth3dsDetails implements AuthorisationDetails {
 
-    public enum Auth3DResult {
+    public enum Auth3dsResult {
         AUTHORISED, DECLINED, ERROR
     }
 
     private String paResponse;
 
-    private Auth3DResult auth3DResult;
+    private Auth3dsResult auth3DsResult;
 
     @JsonProperty("pa_response")
     public void setPaResponse(String paResponse) {
@@ -21,12 +21,12 @@ public class Auth3dsDetails implements AuthorisationDetails {
             return paResponse;
         }
 
-    public Auth3DResult getAuth3DResult() {
-        return auth3DResult;
+    public Auth3dsResult getAuth3DsResult() {
+        return auth3DsResult;
     }
 
-    @JsonProperty("auth_3d_result")
-    public void setAuth3DResult(String auth3DResult) {
-        this.auth3DResult = auth3DResult == null ? null : Auth3DResult.valueOf(auth3DResult);
+    @JsonProperty("auth_3ds_result")
+    public void setAuth3dsResult(String auth3dsResult) {
+        this.auth3DsResult = auth3dsResult == null ? Auth3dsResult.ERROR : Auth3dsResult.valueOf(auth3dsResult);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
+++ b/src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
@@ -20,6 +20,7 @@ import javax.inject.Inject;
 import java.util.List;
 import java.util.Optional;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_3DS_READY;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
 
@@ -75,12 +76,9 @@ public class Card3dsResponseAuthService extends CardAuthoriseBaseService<Auth3ds
             chargeStatusUpdater.updateChargeTransactionStatus(chargeEntity.getExternalId(), status);
             Optional<PaymentRequestEntity> paymentRequestEntity = paymentRequestDao.findByExternalId(chargeEntity.getExternalId());
 
-            if (StringUtils.isBlank(transactionId)) {
-                logger.warn("Auth3DSDetails authorisation response received with no transaction id. -  charge_external_id={}", chargeId);
-            } else {
+            if (!isBlank(transactionId)) {
                 setGatewayTransactionId(chargeEntity, transactionId, paymentRequestEntity);
             }
-
             chargeEventDao.persistChargeEventOf(chargeEntity, Optional.empty());
             return operationResponse;
 
--- a/src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
+++ b/src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
@@ -2,6 +2,7 @@ package uk.gov.pay.connector.service.epdq;
 
 import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.connector.model.EpdqParamsFor3DSecure;
+import uk.gov.pay.connector.model.domain.Auth3dsDetails;
 import uk.gov.pay.connector.service.BaseAuthoriseResponse;
 
 import javax.xml.bind.annotation.XmlAttribute;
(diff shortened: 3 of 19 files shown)


src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
```diff
@@ -4,8 +4,14 @@
 
 public class Auth3dsDetails implements AuthorisationDetails {
 
+    public enum Auth3DResult {
```
alexbishop1: `Auth3dsResult` for consistency, please

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
```diff
@@ -4,8 +4,14 @@
...
+        AUTHORISED, DECLINED, ERROR
```
alexbishop1: `SUCCESS, REJECTED, ERROR` would align better with the corresponding internal statuses

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
```diff
@@ -15,4 +21,12 @@ public String getPaResponse() {
             return paResponse;
         }
 
+    public Auth3DResult getAuth3DResult() {
+        return auth3DResult;
+    }
+
+    @JsonProperty("auth_3d_result")
```
alexbishop1: `3ds` rather than just `3d`, please

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
```diff
@@ -15,4 +21,12 @@ public String getPaResponse() {
...
+    public void setAuth3DResult(String auth3DResult) {
+        this.auth3DResult = auth3DResult == null ? null : Auth3DResult.valueOf(auth3DResult);
```
alexbishop1: The null check is a bit _ugh_ — I wonder if perhaps there should be a `NOT_APPLICABLE` variant or something in the enum.

This will throw an exception if the value is invalid. Are we OK with that or do we want a more graceful failure?

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
```diff
@@ -15,4 +21,12 @@ public String getPaResponse() {
...
+    }
```
alexbishop1: I’m not too thrilled by what the class has become.

It encapsulates the result of the 3D Secure authentication process, which is OK — I actually think it would be nice to rename this `Auth3dsDetails` class to `Auth3dsResult` and rename the inner `Auth3DResult` enum to `Status` or something.

However, the way it has _either_ the `paResponse` field non-null _or_ the `auth3DResult` non-null is a bit icky. We’re hamstrung by this class being the deserialisation of the data posted to the 3D Secure endpoint. I wonder if we could have two endpoints — and hence two classes — for 3D Secure: one for when there’s a PA response and one for when there’s a status.

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
```diff
@@ -75,12 +76,9 @@ public ChargeEntity preOperation(String chargeId, Auth3dsDetails auth3DsDetails)
             chargeStatusUpdater.updateChargeTransactionStatus(chargeEntity.getExternalId(), status);
             Optional<PaymentRequestEntity> paymentRequestEntity = paymentRequestDao.findByExternalId(chargeEntity.getExternalId());
 
-            if (StringUtils.isBlank(transactionId)) {
-                logger.warn("Auth3DSDetails authorisation response received with no transaction id. -  charge_external_id={}", chargeId);
```
alexbishop1, line 79: Why was this logging removed?

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
```diff
@@ -19,6 +20,18 @@
     private static final String WAITING = "51";
     private static final String REJECTED = "2";
 
+    public static EpdqAuthorisationResponse createPost3dsResponseFor(Auth3dsDetails.Auth3DResult threeDsResult) {
```
alexbishop1: I’m not convinced this belongs here — for the post-3DS case, this wouldn’t actually represent a deserialised response from ePDQ.

I’m skeptical about the approach of wrapping this in a `GatewayResponse` and sending it down the same path as all the other gateway responses because it isn’t a gateway response in the same way that others are: it’s data from frontend, not ePDQ itself. It feels a bit like we’re shoehorning it.

At the very least, I think we should have a separate class from `EpdqAuthorisationResponse` to encapsulate this post-3DS data. But I also think it might be worth taking a step back and thinking a bit more broadly about how we handle this.

src/main/java/uk/gov/pay/connector/service/sandbox/SandboxPaymentProvider.java
```diff
@@ -152,7 +152,7 @@ public String getErrorMessage() {
             }
 
             @Override
-            public Optional<GatewayParamsFor3DSecure> getAuth3dsDetails() {
+            public Optional<GatewayParamsFor3ds> getAuth3dsDetails() {
```
alexbishop1: Please rename this method to match the return type

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
```diff
@@ -48,12 +53,28 @@ public void shouldSuccessfully_authorise3ds() throws Exception {
...
         response
-                .body("status", Matchers.is(AUTHORISATION_3DS_REQUIRED.toString()))
+                .body("status", is(AUTHORISATION_3DS_REQUIRED.toString()))
                 .statusCode(200);
 
         assertFrontendChargeStatusIs(chargeId, AUTHORISATION_3DS_REQUIRED.toString());
     }
 
+    @Test
+    public void shouldSuccessfully_authorise3ds() throws Exception {
```
alexbishop1, line 64: Can we name this in a way that makes the distinction with `shouldSuccessfully_authoriseForAChargeRequiring3ds` clearer? `shouldAuthorise_whenRequires3dsAnd3dsAuthenticationSuccessful` or something?

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
```diff
@@ -48,12 +53,28 @@ public void shouldSuccessfully_authorise3ds() throws Exception {
...
+        app.getDatabaseTestHelper().enable3dsForGatewayAccount(Long.parseLong(accountId));
+        String chargeId = createNewChargeWith(AUTHORISATION_3DS_REQUIRED, RandomIdGenerator.newId());
+
+        Map<String, String> payload = ImmutableMap.of("auth_3d_result", "AUTHORISED");
```
alexbishop1: I’m sure there’s a constant you can use here

src/test/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponseTest.java
```diff
@@ -0,0 +1,35 @@
+package uk.gov.pay.connector.service.epdq;
+
+import org.junit.Test;
+import uk.gov.pay.connector.model.domain.Auth3dsDetails;
+import uk.gov.pay.connector.service.BaseAuthoriseResponse;
+
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.*;
```
alexbishop1: Wildcard imports!

alexbishop1 (dismissed):
While it’s easy to implement, I’m uneasy with the approach of essentially synthesising a gateway response to handle the result of the 3D Secure authentication (there’s a comment below about that).

Thanks for addressing the remaining comments from https://github.com/alphagov/pay-connector/pull/587/files. There’s a couple of things still left over from there:
* The addition of `paRequest` to `WorldpayOrderStatusResponse`: still not persuaded it’s adding much
* I still think we could do a bit more to reduce the duplication between `EpdqPayloadDefinitionForNewOrder` and `EpdqPayloadDefinitionForNew3dsOrder`

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
kakumara: This is the epdq result coming from frontend. States are more or less in sync with the other states from `EpdqAuthorisationResponse` except REJECTED. (AUTHORISED, ERROR are already in sync). I believe its better to keep this representing the origin rather than Charge Internal states are they are two different aspects.

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
kakumara: I changed it to ERROR. as we treat anything other than AUTHORISED or REJECTED as errors.

src/main/java/uk/gov/pay/connector/service/Card3dsResponseAuthService.java
kakumara: Because it serves no purpose. 
Few line above we have 
```
String transactionId = operationResponse.getBaseResponse()
                    .map(BaseAuthoriseResponse::getTransactionId).orElse("");
```
where we defaulting to "" which will cause this log line.

src/main/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponse.java
kakumara: This will be removed in the next PR. We discussed we will have a verification with EPDQ after the authorisation so no need of this static constructor anymore.

src/main/java/uk/gov/pay/connector/service/sandbox/SandboxPaymentProvider.java
kakumara: 👍

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
kakumara: Fully agree 👍  😖

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceITest.java
kakumara: 👍

src/test/java/uk/gov/pay/connector/service/epdq/EpdqAuthorisationResponseTest.java
kakumara: 👍

src/main/java/uk/gov/pay/connector/model/domain/Auth3dsDetails.java
kakumara: So we had a discussion about this point separately. It felt like having 2 endpoints looks like a way we should probably attempt, but can be done as a separate story. Created a story specifically for this
https://payments-platform.atlassian.net/browse/PP-3532

rauligs (approved):
Already addressed comments and had an explanation from @kakumara.

PR #589, opened 2018-03-12
PP-3497 CSV download including incorrect payment states

chrisgrimble:
The problem is because the next link is being generated for the old
search end point not the new one. For the csv download we follow the
next link recursively to download all the transactions. We were do the
search correctly for the first 500 result then using the old search
which has a different parameter name for status for the next page. We
kept the parameter names the same so we used the new parameters with
the old search.

To fix use the path from the UriInfo object which we are already using
to get the base uri.

with @kakumara


georgeracu (approved):
LGTM

PR #590, opened 2018-03-12
BAU Run e2e on PR by default

mrlumbu:
There is a perception that disabling e2e by default is causing
too many problems in master. This flips the deafult behaviour to run
e2e on PRs, but it is still possible to disable them.


ckalista (approved):
Looks good.

PR #591, opened 2018-03-12
BAU differentiate PR master metrics

mrlumbu:
Different things happen on master and PR builds so we should
differentiate times


belindac:
Did you want all the metrics differentiating between master and PR, or just the total build ones ?

belindac:
This has been implemented

PR #592, opened 2018-03-12
PP-3498 Results from 'State' dropdown on transactions filter should

chrisgrimble:
only relate to the selected payment state

When you searched in self service for any of the error states it would
bring back all the transactions in any of the states. This is because
they were all being mapped to failed. Changed the external states so we
map to declined, timeout and cancelled.

with @kakumara


daniele-occhipinti (approved):
LGTM

PR #593, opened 2018-03-13
PP-3498 Results from 'State' dropdown on transactions filter should

chrisgrimble:
only relate to the selected payment state

When you searched in self service for any of the error states it would
bring back all the transactions in any of the states. This is because
they were all being mapped to failed. We had changed so we searched by
declined, timeout and cancelled but the results were all mapped to
failed. This will also change the reponses to also be declined, timeout
and cancelled. Had to do this is a separate commit as self service
needed to be updated to hand these before we could send them.


daniele-occhipinti (approved):
LGTM

PR #595, opened 2018-03-15
PP-3450 Fix broken e2e.

kakumara:
## WHAT
Somehow GatewayAccountEntity.isRequire3ds is getting JSONised during get gateway account. For now putting the name of the method to how it was to fix the e2e build.


chrisgrimble (approved):
LGTM

PR #596, opened 2018-03-16
PP-3450 Adding 3ds frontend result verification with epdq

kakumara:
need to wait till https://github.com/alphagov/pay-connector/pull/588 goes in 
## WHAT
Making sure the EPDQ status is inline with frontend user authorised status by verifying using a payment status query.
In the case of a discrepancy, the charge is not authorised irrespective of what ever the status from EPDQ where the charge status will be marked to non success state from frontend or (followed by) non success error from EPDQ.

Also changed the EPDQ url's to match with the frontend changes of PP-3453

Refactored the EPDQ provider unit tests in to base and 2 subclass (one for 3ds tests specifically) as the class was growing too much.


chrisgrimble (approved):
LGTM

PR #597, opened 2018-03-20
PP-3436 Use separate card payments e2e job

maxcbc:


Jenkinsfile
```diff
@@ -73,15 +73,15 @@ pipeline {
...
+        stage('Card Payment End-to-End Tests') {
             when {
                 anyOf {
                   branch 'master'
                   environment name: 'RUN_END_TO_END_ON_PR', value: 'true'
                 }
             }
             steps {
-                runE2E("connector")
+                runCardPaymentsE2E("connector")
```
juanjoqmelian, line 84: Am I right saying that products uses connector? If so, I think we should run products end to end tests too unless that there's no risk in not running them.

maxcbc: I did think that, but I concluded its pretty unlikely that a connector change will break products e2e tests while not breaking card payments e2e tests. This isn't true of selfservice which could break one and not the other. If it starts happening its not difficult to add, but I think we should chance it for now.

juanjoqmelian: Ok, fair enough. I'll approve it then.

juanjoqmelian (approved):

PR #599, opened 2018-03-21
PP-3438 Remove products smoke test from connector build

mrlumbu:
Connector is not affected by products over and above
functionality covered in card smoke test, so removing this.


juanjoqmelian (dismissed):

kakumara (approved):
LOKTM

PR #602, opened 2018-03-23
PP-3526 Add `md` (smartpay 3ds response) column

kakumara:
## WHAT
`SmartPay` 3ds flow requires collecting and re-posting the value in `md` as part of the 3ds handover. This is in addition to `issuerUrl` and `paRequest`. (which we already have)

MD = A payment session identifier returned by the card issuer.


georgeracu (approved):
LGTM

PR #603, opened 2018-03-23
PP-3526 smartpay 3ds required request handling

kakumara:
This must wait for https://github.com/alphagov/pay-connector/pull/602 and until we get a success response for the contract test. 
## WHAT

Includes:
- 3ds require elements added to the smartpay order template based on Gateway account 3ds require flag.
- Card3dsEntity and Auth3dsDetailsEntity extended to use the new field `md` specific to smartpay.
- extended SmartpayAuthorisationResponse to extract the 3ds required elements out of the XML response. (issuerUrl, paRequest, md). Also determining the authorisationStatus() based on `RedirectShopper` status.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
@@ -129,7 +129,7 @@ public class EpdqOrderRequestBuilder extends OrderRequestBuilder {
     public static EpdqOrderRequestBuilder anEpdq3DsAuthoriseOrderRequestBuilder(String frontendUrl) {
         EpdqTemplateData epdqTemplateData = new EpdqTemplateData();
         epdqTemplateData.setFrontendUrl(frontendUrl);
-        return new EpdqOrderRequestBuilder(epdqTemplateData, AUTHORISE_3DS_ORDER_TEMPLATE_BUILDER, OrderRequestType.AUTHORISE, AUTHORISE_OPERATION_TYPE);
+        return new EpdqOrderRequestBuilder(epdqTemplateData, AUTHORISE_3DS_ORDER_TEMPLATE_BUILDER, OrderRequestType.AUTHORISE_3DS, AUTHORISE_OPERATION_TYPE);
     }
 
     public static EpdqOrderRequestBuilder anEpdqCaptureOrderRequestBuilder() {
--- a/src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayAuthorisationResponse.java
+++ b/src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayAuthorisationResponse.java
@@ -15,6 +15,7 @@ public class SmartpayAuthorisationResponse extends SmartpayBaseResponse implemen
 
     private static final String AUTHORISED = "Authorised";
     private static final String REDIRECT_SHOPPER = "RedirectShopper";
+    private static final String REFUSED = "Refused";
 
     @XmlPath("soap:Body/ns1:authoriseResponse/ns1:paymentResult/ns1:resultCode/text()")
     private String result;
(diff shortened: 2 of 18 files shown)


src/main/java/uk/gov/pay/connector/service/epdq/EpdqOrderRequestBuilder.java
```diff
@@ -129,7 +129,7 @@ public static EpdqOrderRequestBuilder anEpdqQueryOrderRequestBuilder() {
     public static EpdqOrderRequestBuilder anEpdq3DsAuthoriseOrderRequestBuilder(String frontendUrl) {
         EpdqTemplateData epdqTemplateData = new EpdqTemplateData();
         epdqTemplateData.setFrontendUrl(frontendUrl);
-        return new EpdqOrderRequestBuilder(epdqTemplateData, AUTHORISE_3DS_ORDER_TEMPLATE_BUILDER, OrderRequestType.AUTHORISE, AUTHORISE_OPERATION_TYPE);
+        return new EpdqOrderRequestBuilder(epdqTemplateData, AUTHORISE_3DS_ORDER_TEMPLATE_BUILDER, OrderRequestType.AUTHORISE_3DS, AUTHORISE_OPERATION_TYPE);
```
chrisgrimble: Why did this change for smartpay

src/main/resources/migrations.xml
```diff
@@ -1302,4 +1302,19 @@
...
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="add md to card_3ds table" author="">
+        <addColumn tableName="card_3ds">
+            <column name="md" type="text">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
```
chrisgrimble: Don't we need to deploy this first as one branch then start using it in the Java in a later release? Maybe we can get away with this as no one will do a 3ds payment until we turn it on.

src/test/java/uk/gov/pay/connector/it/contract/SmartpayPaymentProviderTest.java
```diff
@@ -109,6 +111,20 @@ public void shouldSendSuccessfullyAnOrderForMerchant() throws Exception {
         testCardAuthorisation(paymentProvider, chargeEntity);
     }
 
+    @Ignore
+    @Test
+    //TODO: fix this before merging
```
chrisgrimble: You said it yourself

src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayAuthorisationResponse.java
```diff
@@ -13,32 +14,67 @@
...
+        switch (result) {
+            case AUTHORISED:
+                return AuthoriseStatus.AUTHORISED;
+            case REDIRECT_SHOPPER:
+                return AuthoriseStatus.REQUIRES_3DS;
+            case REFUSED:
+                return AuthoriseStatus.REJECTED;
+            default:
+                return AuthoriseStatus.ERROR;
+        }
```
chrisgrimble, line 53: More a question but before we defaulted to REJECTED rather than error. I guess this isn't really a problem but will end up with a different page in frontend when things go wrong?

chrisgrimble (changes requested):

src/main/resources/migrations.xml
kakumara: yes. This is because this branch is still based on the DB change branch. but once it goes in, and this is rebased this will disappear from here

src/main/java/uk/gov/pay/connector/service/smartpay/SmartpayAuthorisationResponse.java
kakumara: Yes. I realised out smartpay response mapping was not inline with other payment providers as I did this. even a network error (soap fault) we are marking as AUTH REJECTED which is wrong. We are handling these nicely in worldpay and epdq, so I changed the default to ERROR while REJECTED is only if smartpay tells us REJECTED.

src/test/java/uk/gov/pay/connector/it/contract/SmartpayPaymentProviderTest.java
kakumara: Yep, I still get a error 500 from smartpay. So we have to wait till we get some response from smartpay support before we push this to master. Will be fixing it!

chrisgrimble (dismissed):
LGTM

chrisgrimble (approved):
Rebased so LGTM

PR #606, opened 2018-04-04
Ensure java is always run with $JAVA_OPTS

rhowe-gds:
This ensures our memory limits are adhered to etc

### DIFF
--- a/docker-startup-with-db-migration.sh
+++ b/docker-startup-with-db-migration.sh
@@ -1,4 +1,0 @@
-#!/usr/bin/env bash
-java -jar *-allinone.jar waitOnDependencies *.yaml && \
-  java -jar *-allinone.jar db migrate *.yaml && \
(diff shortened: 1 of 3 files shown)


docker-startup-with-db-migration.sh
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env bash
-java -jar *-allinone.jar waitOnDependencies *.yaml && \
-  java -jar *-allinone.jar db migrate *.yaml && \
-  java $JAVA_OPTS -jar *-allinone.jar server *.yaml
+
```
tlwr: can you delete this file, it shouldn't be here (on my todo list to delete it)

rhowe-gds: With pleasure

tlwr (dismissed):

tlwr:
Need to remove the db-migration script from Dockerfile

tlwr (approved):

PR #607, opened 2018-04-07
BAU: Add performance report resource

tlwr:
We need a better way to get performance platform data, I propose we
query the database, to do this we need an API endpoint which returns to
us the relevant data.

This creates a resource at `/v1/api/reports/performance-report`.

We can then query this from an ad-hoc task within our infrastructure which will ensure this gets to the relevant people and to the performance platform.

I fully expect to get roasted with this PR as I haven't written serious
java in about 4 or 5 years. Especially because I used vim and so there are probably a ton of unused imports. Also no tests...

@alexbishop1 :eyes: pls

solo @tlwr

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
@@ -37,6 +37,7 @@ import uk.gov.pay.connector.resources.EmailNotificationResource;
 import uk.gov.pay.connector.resources.GatewayAccountResource;
 import uk.gov.pay.connector.resources.HealthCheckResource;
 import uk.gov.pay.connector.resources.NotificationResource;
+import uk.gov.pay.connector.resources.PerformanceReportResource;
 import uk.gov.pay.connector.resources.SecurityTokensResource;
 import uk.gov.pay.connector.resources.TransactionsSummaryResource;
 import uk.gov.pay.connector.service.CaptureProcessScheduler;
@@ -98,6 +99,7 @@ public class ConnectorApp extends Application<ConnectorConfiguration> {
         environment.jersey().register(injector.getInstance(HealthCheckResource.class));
         environment.jersey().register(injector.getInstance(EmailNotificationResource.class));
         environment.jersey().register(injector.getInstance(SchemeRewriteFilter.class));
+        environment.jersey().register(injector.getInstance(PerformanceReportResource.class));
 
         setupSchedulers(configuration, environment, injector);
 
--- a/src/main/java/uk/gov/pay/connector/dao/PerformanceReportDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/PerformanceReportDao.java
@@ -1,0 +1,37 @@
+package uk.gov.pay.connector.dao;
+
+import com.google.inject.Provider;
+import com.google.inject.persist.Transactional;
+import org.apache.commons.lang3.StringUtils;
+import uk.gov.pay.connector.model.domain.report.PerformanceReportEntity;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.CAPTURED;
+
+import javax.inject.Inject;
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+import javax.persistence.criteria.CriteriaBuilder;
+import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Path;
+import javax.persistence.criteria.Predicate;
+import javax.persistence.criteria.Root;
+import java.time.Duration;
+import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+
+@Transactional
+public class PerformanceReportDao extends JpaDao<PerformanceReportEntity> {
+    @Inject
+    public PerformanceReportDao(final Provider<EntityManager> entityManager) {
+        super(entityManager);
+    }
+
+    public PerformanceReportEntity aggregateNumberAndValueOfPayments() {
+            return (PerformanceReportEntity) entityManager.get()
+                    .createQuery("SELECT NEW uk.gov.pay.connector.model.domain.report.PerformanceReportEntity(COUNT(c.amount), SUM(c.amount), AVG(c.amount)) FROM ChargeEntity c WHERE c.status = :status")
+                    .setParameter("status", CAPTURED.toString())
+                    .setMaxResults(1)
+                    .getSingleResult();
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/model/domain/report/PerformanceReportEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/report/PerformanceReportEntity.java
@@ -1,0 +1,25 @@
+package uk.gov.pay.connector.model.domain.report;
+
+public class PerformanceReportEntity {
+    private final long totalVolume;
+    private final long totalAmount;
+    private final double averageAmount;
+
+    public PerformanceReportEntity(long totalVolume, long totalAmount, double averageAmount) {
+        this.totalVolume = totalVolume;
+        this.totalAmount = totalAmount;
+        this.averageAmount = averageAmount;
+    }
+
+    public long getTotalVolume() {
+        return totalVolume;
+    }
+
+    public long getTotalAmount() {
+        return totalAmount;
+    }
+
+    public double getAverageAmount() {
+        return averageAmount;
+    }
+}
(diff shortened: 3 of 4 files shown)


src/main/java/uk/gov/pay/connector/model/domain/report/PerformanceReportEntity.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.pay.connector.model.domain.report;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+
+@Entity
+public class PerformanceReportEntity {
```
alexbishop1: As you’ve discovered, this isn’t going to work because there’s no primary key. Since this class doesn’t represent anything real in the database, I don’t think we need to actually use an `@Entity` class here. We could just have a POJO and initialise it directly by interrogating the query result rather than relying on JPA’s entity mapping.

src/main/java/uk/gov/pay/connector/model/domain/report/PerformanceReportEntity.java
```diff
@@ -0,0 +1,31 @@
...
+    private final static Logger logger = LoggerFactory.getLogger(PerformanceReportEntity.class);
```
alexbishop1: Don’t need a logger if you’re not doing any logging!

src/main/java/uk/gov/pay/connector/resources/PerformanceReportResource.java
```diff
@@ -0,0 +1,40 @@
...
+public class PerformanceReportResource {
+    private PerformanceReportDao performanceReportDao;
+
+    @Inject
+    public PerformanceReportResource(PerformanceReportDao performanceReportDao) {
+        this.performanceReportDao = performanceReportDao;
+    }
+
+    @GET
+    @Path("/v1/api/reports/performanceReport")
```
alexbishop1: We prefer to have URLs like `performance-report`

alexbishop1 (changes requested):
Good start!

I note that when we query, we just get the totals for everything (i.e. from the beginning of time until now). I presume if we want daily reports or whatever, we might want to restrict by date. This could be tricky though because charges only have a created date, which might not match the date it was captured, so we could end up missing some.

We might instead want to query the `charge_events` table instead (which does have timestamps for captures), find all the charges that were captured on a particular day, then join back on to the `charges` table to get the amounts.

src/main/java/uk/gov/pay/connector/dao/PerformanceReportDao.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Transactional
+public class PerformanceReportDao extends JpaDao<PerformanceReportEntity> {
+    @Inject
+    public PerformanceReportDao(final Provider<EntityManager> entityManager) {
+        super(entityManager);
+    }
+
+    public PerformanceReportEntity aggregateNumberAndValueOfPayments() {
+            long[] results = (long[]) entityManager.get()
```
alexbishop1: Unfortunately, this cast doesn’t work. Even if it did, not all the returned columns are longs (the `AVG` ends up as a double).

From my fairly rudimentary knowledge of JPA, mapping the results of a query to a POJO without involving an entity or a bunch of XML is a bit of a pain. Using a native query makes it more so.

However, your query is not actually beyond the capabilities of JPQL, so we can use plain `createQuery`. Something like this should allow us to get a `PerformanceReportEntity` object out:

```
SELECT NEW uk.gov.pay.connector.model.domain.report.PerformanceReportEntity(COUNT(c.amount), SUM(c.amount), AVG(c.amount)) FROM ChargeEntity c WHERE c.status = :status
```

But perhaps someone with a better grasp of JPA than me may be able to offer a better suggestion.

src/main/java/uk/gov/pay/connector/dao/PerformanceReportDao.java
```diff
@@ -0,0 +1,47 @@
...
+                    .createNativeQuery("SELECT COUNT(c.amount) as total_volume, SUM(c.amount) as total_amount, AVG(c.amount) as average_amount FROM ChargeEntity c WHERE c.status = :status")
+                    .setParameter("status", CAPTURED)
```
alexbishop1: We need to convert the enum variant to a string: `CAPTURED.toString()`

src/main/java/uk/gov/pay/connector/model/domain/report/PerformanceReportEntity.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.pay.connector.model.domain.report;
+
+public class PerformanceReportEntity {
+    private long totalVolume;
+    private long totalAmount;
+    private long averageAmount;
```
alexbishop1: This needs to be a double because it’s a floating point number

alexbishop1: All of these fields can be `final`

src/main/java/uk/gov/pay/connector/resources/PerformanceReportResource.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.ws.rs.core.Response;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+import static javax.ws.rs.core.Response.ok;
+
+@Path("/")
+public class PerformanceReportResource {
```
alexbishop1, line 21: To make this work, you also need to add `PerformanceReportResource.class` to `ConnectorApp`

alexbishop1 (dismissed):
Assuming we don’t need to filter by date yet, this is getting there

tlwr:
Rebased this onto master and have added Alex's request

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -37,6 +37,7 @@
 import uk.gov.pay.connector.resources.GatewayAccountResource;
 import uk.gov.pay.connector.resources.HealthCheckResource;
 import uk.gov.pay.connector.resources.NotificationResource;
+import uk.gov.pay.connector.resources.PerformanceReportResource;
```
alexbishop1, line 40: You also need a `environment.jersey().register(injector.getInstance(PerformanceReportResource.class));` line in the `run(ConnectorConfiguration, Environment)` method in this class.

alexbishop1 (changes requested):
Needs more glue code to hook everything up

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
tlwr: done

src/main/java/uk/gov/pay/connector/dao/PerformanceReportDao.java
```diff
@@ -0,0 +1,37 @@
...
+@Transactional
+public class PerformanceReportDao extends JpaDao<PerformanceReportEntity> {
+    @Inject
+    public PerformanceReportDao(final Provider<EntityManager> entityManager) {
+        super(entityManager);
+    }
+
+    public PerformanceReportEntity aggregateNumberAndValueOfPayments() {
+            return (PerformanceReportEntity) entityManager.get()
+                    .createNativeQuery("SELECT NEW uk.gov.pay.connector.model.domain.report.PerformanceReportEntity(COUNT(c.amount), SUM(c.amount), AVG(c.amount)) FROM ChargeEntity c WHERE c.status = :status")
```
alexbishop1: As discussed in an outdated comment, this needs to be `createQuery` now

src/main/java/uk/gov/pay/connector/dao/PerformanceReportDao.java
```diff
@@ -0,0 +1,37 @@
...
+                    .setParameter("status", CAPTURED.toString())
+                    .setMaxResults(1)
```
alexbishop1: This will fail hard if there aren’t any payments. Since that’s not going to happen much in the real world, not sure how much of a concern that is.

src/main/java/uk/gov/pay/connector/model/domain/report/PerformanceReportEntity.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.pay.connector.model.domain.report;
+
+public class PerformanceReportEntity {
+    private final long totalVolume;
+    private final long totalAmount;
+    private final float averageAmount;
```
alexbishop1: The query will give us a `double` rather than a float (also update the constructor and getter)

alexbishop1 (changes requested):
Getting there!

alexbishop1 (approved):
I’m still not entirely convinced that getting a count of all payments ever made using the platform is especially useful, but the code works (and should be fast due to the index on the `status` column) so I trust you know what you’re doing

PR #608, opened 2018-04-10
PP-3512 Extend expiry window time to 90 minutes

simad:
## WHAT
 - In frontend we have a 90 min window of validity for the session cookie,
   while in connector we sweep-expire all charges which have been inactive for > 60 minutes.
 - This PR increase the expiry window to 90 minutes, to match what we have in frontend. Update to our tech docs will follow.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
@@ -76,7 +76,7 @@ public class ChargesApiResource {
     private static final String PAGE = "page";
     private static final String DISPLAY_SIZE = "display_size";
     private static final Set<String> CHARGE_REQUEST_KEYS_THAT_MAY_HAVE_PII = Collections.singleton("description");
-    private static final int ONE_HOUR = 3600;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
```diff
@@ -76,7 +76,7 @@
     private static final String PAGE = "page";
     private static final String DISPLAY_SIZE = "display_size";
     private static final Set<String> CHARGE_REQUEST_KEYS_THAT_MAY_HAVE_PII = Collections.singleton("description");
-    private static final int ONE_HOUR = 3600;
+    private static final int ONE_HOUR_AND_A_HALF = 5400;
```
alexbishop1, line 79: Minor but I’d be tempted to have this as `NINETY_MINUTES_IN_SECONDS = 90 * 60` — units in names are often useful and why do complex maths ourselves when we have computers to do that? (If I’m remembering correctly, the calculation will be done once, at compile time)

alexbishop1 (approved):
Looks good! Suggested a change to the constant declaration but I’m not going to withhold approval on it

PR #609, opened 2018-04-30
Replace EditorConfig with latest from GDS Way

alexbishop1:
In addition to actually being valid, the GDS Way EditorConfig has everything we already had plus some extra goodies, such as settings for YAML files


jonheslop (approved):
👍

PR #611, opened 2018-05-16
Remove PA Request from WorldpayOrderStatusResponse toString()

alexbishop1:
The PA request is opaque to us so there’s no point including it


simad (approved):
I approved pls

PR #613, opened 2018-05-21
PP-3650 Warn if return_url is non-https on live account

mrlumbu:
We are planning to relax restriction that the return_url set on a payment
be https. As part of this, we want some way of monitoring if partners
ever do set http return urls on live accounts, in order to guauge if this
is a problem, and whether we need to build more fine grained control
of this sort of thing.
To monitor this, this commit just logs an info
whenever a payment is created against a live account with a non-https url.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
@@ -242,4 +242,8 @@ public class GatewayAccountEntity extends AbstractVersionedEntity {
     public void setId(Long id) {
         this.id = id;
     }
+    
+    public boolean isLive() {
+        return Type.LIVE.equals(type);
+    }
 }
--- a/src/main/java/uk/gov/pay/connector/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/service/ChargeService.java
@@ -1,6 +1,8 @@
 package uk.gov.pay.connector.service;
 
 import com.google.inject.persist.Transactional;
+import org.slf4j.Logger;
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
```diff
@@ -242,4 +242,8 @@ public boolean hasAnyAcceptedCardType3dsRequired() {
     public void setId(Long id) {
         this.id = id;
     }
+    
+    public boolean isLive() {
+        return Type.LIVE.equals(type);
```
whpearson, line 247: Maybe should be this.type for clarity?

whpearson (changes requested):

src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
mrlumbu: mmm, not really how we've done things - see eg. https://github.com/alphagov/pay-connector/blob/4354aa994486462b6959acab017218c1a6c5d34c/src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java#L238 (and more generally in Java, not usually done unless required e.g. when initialising)

whpearson: Back in my day 👴

mrlumbu: ah, well I stand corrected. But I think it makes sense to follow the convention we have set

whpearson (approved):

PR #616, opened 2018-05-31
PP-3864 Increase width of notification_credentials username, password

mrlumbu:
These were unnecessarily tight, and pg is clever enoough not to
allocate unused space, so giving them both 255.

See https://www.postgresql.org/docs/9.3/static/datatype-character.html
```
If the string to be stored is shorter than the declared length, values of
type character will be space-padded; values of type character varying will
simply store the shorter string.
```

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1317,4 +1317,15 @@
             </column>
         </addColumn>
     </changeSet>
+
+    <changeSet id="increase width of notification_credentials username, password columns" author="">
+        <modifyDataType
+            tableName="notification_credentials"
+            columnName="username"
+            newDataType="varchar(255)"/>
+        <modifyDataType
+            tableName="notification_credentials"
+            columnName="password"
+            newDataType="varchar(255)"/>
+    </changeSet>
 </databaseChangeLog>


oswaldquek (dismissed):

alexbishop1 (dismissed):
This reassures me:

“Increasing the length limit for a varchar or varbit column, or removing the limit altogether, no longer requires a table rewrite. Similarly, increasing the allowable precision of a numeric column, or changing a column from constrained numeric to unconstrained numeric, no longer requires a table rewrite. Table rewrites are also avoided in similar cases involving the interval, timestamp, and timestamptz types.”
— from https://www.postgresql.org/docs/9.2/static/release-9-2.html

However, I’m slightly worried about the fact that it will still require an access exclusive lock. Please check how long this takes to run in staging and reassess if it’s more than a few dozen milliseconds.

mrlumbu:
Yes, should have mentioned that it does not do a table rewrite, thanks @alexbishop1 
Also, we do need to be aware that whilst the table is locked we will not be able to receive smartpay notifications. However it will not affect in flight payments. I believe (I will check) that smartpay resends failed notifications, so not overly concerned about table lock. I will check size of table, but I think it must be pretty tiny.

alexbishop1 (approved):

PR #620, opened 2018-06-13
Change error message when gateway responds with unexpected HTTP status code

alexbishop1:
Change the error message returned when a payment gateway responds to a request with a non-200 HTTP status code from:

“Unexpected Response Code From Gateway”

… to:

“Unexpected HTTP status code _XXX_ from gateway”

… where _XXX_ is the actual HTTP status code received e.g. 401, 500 etc.

The status code is really useful for diagnosing problems (it already gets logged in a different error logging line, which is unfortunately devoid of most other context about the request so unlikely to be found in a log search).

We add “HTTP” to make it explicit that it’s an HTTP status code and not a code internal to the gateway. We use “status code” rather than “Response Code” because that’s what it’s called in HTTP.


simad (approved):
💯 moar logging

PR #622, opened 2018-06-15
Run provider contract tests

mrlumbu:


oswaldquek:
would also be good to run the provider contract test on the publicapi side, for e.g. see https://github.com/alphagov/pay-selfservice/blob/master/Jenkinsfile#L48 which runs provider contract tests on the consumer (selfservice) side.

oswaldquek (approved):

whpearson:
Can this be closed?

PR #627, opened 2018-06-22
PP-3059: Redacted email in connector logging.

whpearson:
This makes a change to ChargeSearchParams so that
buildQueryParams has an option to redactPII.

The redactPII option does not add the email address to the params
but a constant "email=*****".

This is then exposed as a new function that is used whilst
building the query params for logging.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
+++ b/src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.dao;
 
 import uk.gov.pay.connector.model.TransactionType;
+
 import uk.gov.pay.connector.model.api.ExternalChargeState;
 import uk.gov.pay.connector.model.api.ExternalRefundStatus;
 import uk.gov.pay.connector.model.domain.ChargeStatus;
@@ -210,14 +211,27 @@ public class ChargeSearchParams {
     }
 
     public String buildQueryParams() {
+        return this.buildQueryParams(false);
+    }
+
+    public String buildQueryParamsWithPiiRedaction () {
+        return this.buildQueryParams(true);
+    }
+
+    private String buildQueryParams(boolean redactPii) {
         StringBuilder builder = new StringBuilder();
         if (transactionType != null) {
             builder.append("&transaction_type=").append(transactionType.getValue());
         }
         if (isNotBlank(reference))
             builder.append("&reference=").append(reference);
-        if (email != null)
-            builder.append("&email=").append(email);
+        if (email != null) {
+            if (redactPii) {
+                builder.append("&email=*****");
+            } else {
+                builder.append("&email=").append(email);
+            }
+        }
         if (fromDate != null)
             builder.append("&from_date=").append(fromDate);
         if (toDate != null)
--- a/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/resources/ChargesApiResource.java
@@ -342,7 +342,7 @@ public class ChargesApiResource {
             logger.info("Charge Search - is feature transactions enabled [{}] took [{}] params [{}]",
                     isFeatureTransactionsEnabled,
                     (endTime - startTime) / 1000000000.0,
-                    searchParams.buildQueryParams());
+                    searchParams.buildQueryParamsWithPiiRedaction());
         }
     }
 
@@ -354,7 +354,7 @@ public class ChargesApiResource {
             long endTime = System.nanoTime();
             logger.info("Transaction Search - took [{}] params [{}]",
                     (endTime - startTime) / 1000000000.0,
-                    searchParams.buildQueryParams());
+                    searchParams.buildQueryParamsWithPiiRedaction());
         }
     }
 
--- a/src/test/java/uk/gov/pay/connector/dao/ChargeSearchParamsTest.java
+++ b/src/test/java/uk/gov/pay/connector/dao/ChargeSearchParamsTest.java
@@ -67,6 +67,34 @@ public class ChargeSearchParamsTest {
     }
 
     @Test
+    public void buildQueryParamsWithPiiRedaction_chargeSearch_withAllParameters() throws Exception {
+
+        String expectedQueryString =
+                "reference=ref" +
+                        "&email=*****" +
+                        "&from_date=2012-06-30T12:30:40Z[UTC]" +
+                        "&to_date=2012-07-30T12:30:40Z[UTC]" +
+                        "&page=2" +
+                        "&display_size=5" +
+                        "&state=created" +
+                        "&card_brand=visa";
+
+        ChargeSearchParams params = new ChargeSearchParams()
+                .withDisplaySize(5L)
+                .withExternalState(EXTERNAL_CREATED.getStatus())
+                .withCardBrand("visa")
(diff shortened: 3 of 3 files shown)


src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -250,6 +252,11 @@ public String buildQueryParams() {
         }
         return builder.toString().replaceFirst("&", "");
     }
+    public String buildFilteredQueryParams() throws PatternSyntaxException {
+        String queryParams = this.buildQueryParams();
+        return queryParams.toString().replaceAll("email=[^&+]", "email=*****" );
```
alexbishop1: My regex foo is a little weak but I’m not sure if this regex will actually match what we want

alexbishop1: Also, `queryParams` is already a string: you don’t need to `toString()` it!

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -250,6 +252,11 @@ public String buildQueryParams() {
         }
         return builder.toString().replaceFirst("&", "");
     }
+    public String buildFilteredQueryParams() throws PatternSyntaxException {
```
alexbishop1: We can probably make this method name more specific: `buildQueryParamsRedactingPii` (“PII” = “personally-identifiable information”) maybe?

Rather than building the query params and then doing a search-and-replaces on the resulting string, I think it might be a bit cleaner to exclude the email when building the query params. This could be done in a minimally invasive way by doing something like this:

```java
public String buildQueryParams() {
    return buildQueryParams(false);
}

public String buildQueryParamsRedactingPii() {
    return buildQueryParams(true);
}

private String buildQueryParams(boolean redactPii) {
    StringBuilder builder = new StringBuilder();
    if (transactionType != null) {
        builder.append("&transaction_type=").append(transactionType.getValue());
    }
    if (isNotBlank(reference))
        builder.append("&reference=").append(reference);
    if (email != null)
        if (redactPii) {
            builder.append("&email=****");
        } else {
            builder.append("&email=").append(email);
        }
    if (fromDate != null)
        builder.append("&from_date=").append(fromDate);
    if (toDate != null)
        builder.append("&to_date=").append(toDate);
    if (page != null)
        builder.append("&page=").append(page);
    if (displaySize != null)
        builder.append("&display_size=").append(displaySize);

    getExternalStates().stream()
            .findFirst()
            .ifPresent(state -> builder.append("&state=").append(state));

    String paymentStates = this.externalChargeStates.stream()
            .collect(Collectors.joining(","));

    String refundStates = this.externalRefundStates.stream()
            .collect(Collectors.joining(","));

    if (isNotEmpty(paymentStates)) {
        builder.append("&payment_states=").append(paymentStates);
    }

    if (isNotEmpty(refundStates)) {
        builder.append("&refund_states=").append(refundStates);
    }

    if (!cardBrands.isEmpty()) {
        cardBrands.forEach(cardBrand -> builder.append("&card_brand=").append(cardBrand));
    }
    return builder.toString().replaceFirst("&", "");
}
```

(We could also avoid the need to remove the first ampersand by using a `StringJoiner` — or maybe something designed to create URL query strings — but let’s not try to fix all the world’s problems here.)

alexbishop1 (changes requested):
I’ve made a comment about a way of doing this I think might be a bit neater

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
whpearson: Sounds like a better approach.

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -210,14 +212,26 @@ public ChargeSearchParams withDisplaySize(Long displaySize) {
...
 
     public String buildQueryParams() {
+        return (this.buildQueryParams(false));
+    }
+
+    public String buildQueryParamsWithPiiRedaction () {
+        return (this.buildQueryParams(true));
+    }
+
+    public String buildQueryParams(boolean redactPii) {
```
alexbishop1: This should be `private` (as we only want people to call the two `public` methods)

alexbishop1 (changes requested):
Looking good.

This could probably do with at least one test in `ChargeSearchParamsTest`. You can just copy the first one, make it call the `buildQueryParamsWithPiiRedaction()` method and change the expected result to include `&email=xxxx`.

tlwr (dismissed):

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -210,14 +212,26 @@ public ChargeSearchParams withDisplaySize(Long displaySize) {
...
         if (isNotBlank(reference))
             builder.append("&reference=").append(reference);
-        if (email != null)
-            builder.append("&email=").append(email);
+        if (email != null) {
+            if (redactPii) 
+                builder.append("&email=*****");
+            else
+                builder.append("&email=").append(email);
+        }
```
alexbishop1: Can you always use curly braces even for single-line `if`/`else` statements please?

Alternatively, I’d be OK with a ternary operator like this:

```java
if (email != null) {
    builder.append("&email=");
    redactPii ? builder.append("****") : builder.append(email);
}
```

… but some people really, really don’t like ternary operators!

alexbishop1 (changes requested):
Looks good: just a minor stylistic point about curly braces

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -12,6 +13,7 @@
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
+import java.util.regex.PatternSyntaxException;
```
alexbishop1: This import can be removed; sorry, didn’t notice this until now

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -210,14 +212,28 @@ public ChargeSearchParams withDisplaySize(Long displaySize) {
...
         }
         if (isNotBlank(reference))
             builder.append("&reference=").append(reference);
-        if (email != null)
-            builder.append("&email=").append(email);
+        if (email != null) {
+            if (redactPii) {
+                builder.append("&email=*****");
+            }
+            else {
```
alexbishop1: We prefer this style:

```java
} else {
```

src/main/java/uk/gov/pay/connector/dao/ChargeSearchParams.java
```diff
@@ -210,14 +212,28 @@ public ChargeSearchParams withDisplaySize(Long displaySize) {
     }
 
     public String buildQueryParams() {
+        return (this.buildQueryParams(false));
```
alexbishop1: Don’t need the outer parentheses here (applies to the method below too)

alexbishop1 (dismissed):
Great stuff!

Left a couple of comments about trivial things but I’m not going to hold approval on them

alexbishop1 (approved):
👌

PR #629, opened 2018-06-27
BAU add state for create charge pact

mrlumbu:
Creating a charge requires an existing garteway account.
This PR also upgrades common testing classes to latest version.


brid (approved):
Approved.

PR #630, opened 2018-06-29
PP-3932: Upgrade Dropwizard from 1.2.2 to 1.2.8

rhowe-gds:
## WHAT
This upgrades Jetty from 9.4.7 to 9.4.11 and upgrades Jackson from 2.9.1 to
2.9.6

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <version>0.1-SNAPSHOT</version>
     <artifactId>pay-connector</artifactId>
     <properties>
-        <dropwizard.version>1.2.7</dropwizard.version>
+        <dropwizard.version>1.2.8</dropwizard.version>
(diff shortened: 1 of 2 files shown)


tlwr (dismissed):

tlwr:
I've attempted to make a test less flakey, which previous failed the build 0.1% of the time

```
Expected: is <Result{isHealthy=true, timestamp=2018-07-02T10:55:21.438+0000}>
    but: was <Result{isHealthy=true, timestamp=2018-07-02T10:55:21.437+0000}>
```

This is a slightly annoying way of failing a 20 minute build, we only
care that it healthchecks properly not that the healthcheck occurred in
the same subsecond time period.

oswaldquek (approved):

PR #631, opened 2018-07-02
BAU: Debug logging and jackson 2.9.6

tlwr:
- warn logging level - reduces logging size per build about 60mb
- dropwizard/jetty upgrade uses 2.9.6 jackson so we update that too

solo @tlwr


heathd (approved):
Looks good, great to reduce logging noise in tests 👍

PR #632, opened 2018-07-02
PP-3942 Introduce ServicePaymentReference object to connector

alexbishop1:
The new `ServicePaymentReference` is a simple immutable object that wraps a `String` containing the `reference` specified by a service when they create a payment.

Replace uses of `String`s to represent the service payment reference with the new `ServicePaymentReference` type to improve type-safety and aid comprehension.

Introduce a `ServicePaymentReferenceConverter` to convert between the new `ServicePaymentReference` type and Strings (and vice-versa) when persisting/depersisting from the database.

Use Jackson’s `ToStringSerializer` to convert between the new `ServicePaymentReference` type and Strings when serialising JSON.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/ServicePaymentReference.java
+++ b/src/main/java/uk/gov/pay/connector/model/ServicePaymentReference.java
@@ -10,8 +10,8 @@ public class ServicePaymentReference {
         this.servicePaymentReference = Objects.requireNonNull(servicePaymentReference);
     }
 
-    public static ServicePaymentReference of(String providerMandateReference) {
-        return new ServicePaymentReference(providerMandateReference);
+    public static ServicePaymentReference of(String servicePaymentReference) {
+        return new ServicePaymentReference(servicePaymentReference);
     }
 
     @Override
--- a/src/main/java/uk/gov/pay/connector/model/ServicePaymentReference.java
+++ b/src/main/java/uk/gov/pay/connector/model/ServicePaymentReference.java
@@ -1,0 +1,36 @@
+package uk.gov.pay.connector.model;
+
+import java.util.Objects;
+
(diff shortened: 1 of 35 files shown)


mrlumbu:
https://www.youtube.com/watch?v=-6BsiVyC1kM :trollface:

src/main/java/uk/gov/pay/connector/model/ServicePaymentReference.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class ServicePaymentReference {
+
+    private final String servicePaymentReference;
+
+    private ServicePaymentReference(String servicePaymentReference) {
+        this.servicePaymentReference = Objects.requireNonNull(servicePaymentReference);
+    }
+
+    public static ServicePaymentReference of(String providerMandateReference) {
```
mrlumbu: don't think this should be `providerMandateReference`?

alexbishop1: Correct — that’s something for Direct Debit! Will fix.

alexbishop1: Fixed

mrlumbu:
I am slightly ambivalent on wrapping such simple things as this. But no massive objections. I think the benefits are more obvious when there is some validation encapsulated in the object. In this case, what we gain in type-safety, we lose at least a bit of in complexity.

mrlumbu (approved):

alexbishop1:
Yeah, I was slightly disappointed with this one. Service payment references don’t do much apart from get persisted and returned. I hope the benefits will show through more clearly with objects that get more deeply involved in the business logic. I agree with you on validation also.

PR #633, opened 2018-07-04
PP-3928: Account list endpoint filter

tlwr:
@tlwr @brid

We return all gateway accounts from the `/v1/api/accounts` endpoint, but this PR adds the `accountIds` argument which will only retrieve those from the database.

This also adds the `/v1/frontend/accounts` endpoint, which is the same as the `api` one. This is so it can be consumed in selfservice to speed up the loading of the services page.

The same work should be replicated in dd-connector

### DIFF
--- a/src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
+++ b/src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.connector.resources;
 
+import com.google.common.base.Splitter;
 import com.fasterxml.jackson.annotation.JsonView;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -24,10 +25,12 @@ import uk.gov.pay.connector.service.PaymentGatewayName;
 
 import javax.inject.Inject;
 import javax.ws.rs.Consumes;
+import javax.ws.rs.DefaultValue;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
+import javax.ws.rs.QueryParam;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
@@ -38,6 +41,7 @@ import java.net.URI;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.ArrayList;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
@@ -60,6 +64,7 @@ import static uk.gov.pay.connector.util.ResponseUtil.successResponseWithEntity;
 public class GatewayAccountResource {
 
     private static final Logger logger = LoggerFactory.getLogger(GatewayAccountResource.class);
+    private static final Splitter COMMA_SEPARATOR = Splitter.on(',').trimResults().omitEmptyStrings();
 
     private static final String DESCRIPTION_FIELD_NAME = "description";
     private static final String ANALYTICS_ID_FIELD_NAME = "analytics_id";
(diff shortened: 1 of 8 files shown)


tlwr:
This work was sponsored by
image

jonheslop (dismissed):
Seems legit

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -110,18 +113,65 @@ public Response getGatewayAccount(@PathParam("accountId") Long accountId) {
 
     }
 
+    // This private method, instead of using a regex Path is due to, as far as
+    // I can tell, the binding of @POST causing /v1/api/accounts to 405, thus
+    // we have to explicitly define /v1/api/accounts with @GET
+    // As such, split out the common functionality in a private method and call
+    // it twice for each explicitly defined resource
+    private Response dualGetGatewayAccounts(
```
mrlumbu: don't like this name. what's wrong with getGatewayAccounts?

mrlumbu (changes requested):

src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoITest.java
```diff
@@ -281,6 +281,43 @@ public void shouldListAllAccounts() throws Exception {
...
+        List<GatewayAccountResourceDTO> gatewayAccounts = gatewayAccountDao.list(accountIds);
+
+        assertEquals(1, gatewayAccounts.size());
+        assertThat(gatewayAccounts.get(0).getAccountId(), is(456L));
+        assertEquals("provider-2", gatewayAccounts.get(0).getPaymentProvider());
+        assertEquals("description-2", gatewayAccounts.get(0).getDescription());
+        assertEquals("service-name-2", gatewayAccounts.get(0).getServiceName());
+        assertEquals(TEST.toString(), gatewayAccounts.get(0).getType());
+        assertEquals("analytics-id-2", gatewayAccounts.get(0).getAnalyticsId());
+    }
```
mrlumbu, line 319: you should have a test where you return multiple accounts as a list

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -110,18 +113,65 @@ public Response getGatewayAccount(@PathParam("accountId") Long accountId) {
...
-        logger.debug("Getting all gateway accounts");
-        List<GatewayAccountResourceDTO> gatewayAccountResourceDTOList = gatewayDao.listAll();
-        gatewayAccountResourceDTOList.forEach(account ->
-                account.addLink("self", buildUri(uriInfo, account.getAccountId()))
-        );
-        return Response
-                .ok(ImmutableMap.of("accounts", gatewayAccountResourceDTOList))
-                .build();
+    public Response getApiGatewayAccounts(
+        @DefaultValue("") @QueryParam("accountIds") String accountIdsArg,
```
mrlumbu, line 161: Jackson can deserialise arrays. If you do it in the form `?accountIds[]=1&accountIds[]=5...` (i.e. standard way) you can just have `@Query("accountIds") String... accountIds)` and you don't have to do any of this deserialisation stuff

tlwr:
- [x] test added
- [x] naming change (this was historical and your change is 👌 )
- [x] tests passing

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
tlwr: I am doing this using Splitter from the google common package, so that we are consistent with adminusers:

https://github.com/alphagov/pay-adminusers/blob/master/src/main/java/uk/gov/pay/adminusers/resources/UserResource.java#L58

This means the query parameter will still look like: `accountIds=1,2,3` but the serialisation is not done by hand

mrlumbu (dismissed):

danworth (approved):
With regards to the updated regex to fix flakey tests it looks good. The rest looks sensible so I'm happy if Dom is satisfied that his changes have been incorporated (appear to be so e.g. inclusion of test for multiple accounts).

mrlumbu:
💚

PR #635, opened 2018-07-06
PP-3493 Add user_external_id to RefundSummary

simad:
## WHAT
 - We need to display the issuer of a refund in the transaction list (csv file) in selfservice
   This specific section of code in connector is in a bit of a flux; this is probably not the
   nicest way to get the job done, but it will be reworked pretty soon (tm).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/OldTransactionDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/OldTransactionDao.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.jooq.impl.DSL.count;
 import static org.jooq.impl.DSL.field;
+import static org.jooq.impl.DSL.inline;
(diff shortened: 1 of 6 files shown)


danworth:
In light of Alex's work to de-String ids, is this an opportune moment to make userExternalId an object rather than adding it as a String? Sounds like this is a temporary implementation, and I'm not as familiar with the code base and hence the extent of any additional refactoring that might be necessary, so happy to be told 'no' :)

simad:
Ideally yes, I think creating a type instead of using strings everywhere is a great practice. Alex is working on dd connector, though, and I'd prefer to have that piece of work done before starting to do the same in connector (which is a bit of a bigger beast)

alexbishop1 (approved):
Looks like it should do the job (even if it does show that we may have made some suboptimal decisions in our generalisations of how we model charges and refunds).

While I agree introducing a `UserExternalId` would be a good step forward, it’s not strictly necessary for this story.

If you feel like it, it might be good to have some tests to check how this behaves if a refund has no user associated with it (which is the case for refunds issued using the API).

PR #636, opened 2018-07-18
PP-4072 Xray integration

belindac:


pom.xml
```diff
@@ -11,7 +11,7 @@
         <guice.version>4.1.0</guice.version>
         <docker-client.version>8.9.2</docker-client.version>
         <jackson.version>2.9.6</jackson.version>
-        <pay-java-commons.version>1.0.0-7</pay-java-commons.version>
+        <pay-java-commons.version>1.0.0-7bc143b9ea2caf1b5d6f39a281691d7a02d57bb4</pay-java-commons.version>
```
simad, line 14: git commit sha as build number ftw

src/main/java/uk/gov/pay/connector/service/CaptureProcessScheduler.java
```diff
@@ -51,9 +57,12 @@ public void start() {
 
         scheduledExecutorService.scheduleAtFixedRate(() -> {
             try {
+                xrayUtils.beginSegment();
```
simad, line 60: I am a bit concerned about this. I am thinking that if there's an exception in the `beginSegment()` method we don't run capture at all - which is not ideal. 
I am no xray expert, is this even possible?

If `beginSegment()` can indeed blow up we might think about moving the capture outside of this try catch so that it would still run

simad (commented):
Looks good, just one question

pom.xml
oswaldquek: yeah the number used to be the jenkins number but when jenkins got upgraded the build numbers got reset to 1, which is a problem as the build 1.0.0-1 already exists. agreed with @alexbishop1 that we should use the git commit sha.

src/main/java/uk/gov/pay/connector/service/CaptureProcessScheduler.java
oswaldquek: yeah good point. we don't want the application to stop working because of an x-ray error. i'll add logic to prevent this.

oswaldquek: this is now dealt with, see https://github.com/alphagov/pay-connector/pull/636/commits/d598b15cfbbac94ae402e45397fb41573f553d9a

simad: 👍

simad (approved):
Looks good. There was a concern regarding the application not running capture in case of a xray error, but this has been addressed. 🚀

PR #637, opened 2018-07-20
PP-4041 Improve capture queue processing

heathd:


src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -187,7 +187,7 @@ public int countChargesForCapture() {
                 "    ce.status = :eventStatus AND " +
                 "    ce.updated >= :cutoffDate " +
                 ") " +
-                "ORDER BY c.createdDate ASC";
+                "ORDER BY random()";
```
DanailMinchev: I think `RANDOM()` is not supported in JPA

DanailMinchev (approved):
`Collections.shuffle` looks reasonable, not sure if we have other option

PR #643, opened 2018-07-24
PP-4054: Add index for gateway_transaction_id to charges

whpearson:
We currently look it up when we get notifications. This is
consuming a fair amount of compute time on our db.

We don't have an index. So add one.

https://github.com/alphagov/pay-connector/blob/4096a80aeae533dcd59f75aaf9bda28a807a401c/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java

Also add an index for looking up tokens by the secure_redirect_token. We do that a lot.


simad (approved):
🚀

PR #652, opened 2018-08-10
PP-4094 Add language column to charges table

DanailMinchev:
## WHAT

- Add language column to charges table with `char(2)`, because it contains only `ISO-639-1`
- Add default value and update the existing rows

Don’t add the default in the same statement as adding the column because this will lock the table while updating all the existing rows.

with @alexbishop1


georgeracu (dismissed):
LGTM

georgeracu (approved):
👍

PR #656, opened 2018-08-14
PP-4119 Improve capture queue reporting

danworth:
Currently we only capture the total size of the charges waiting to be sent to the provider to be processed. However it is more useful to split this into two measures:
- The number of charges eligible to be processed now.
- The number of charges that are within the waiting period to be retried and so are not eligible at that moment.

Therefore the existing `gateway-operations.capture-process.queue-size` metric has been modified to count only charges which can be retried (whilst not applying the configurable batch size limit used by `findChargesForCapture`). A new metric named `gateway-operations.capture-process.waiting-queue-size` has been added to show the charges which are waiting to be retried because they were last attempted within the delay period (this is configurable but currently set to an hour).

As part of this work we have also:
- modified the metric type from `Counter` to `Gauge` which seems more appropriate.
- modified the log message at the end of the capture run to report the number of successful retries and those that failed. This introduces a check on the `GatewayResponse` and a new category in the log message `failed_capture`.

Tests have been added to confirm the registration of the gauges and that they return the correct value.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
@@ -166,15 +166,18 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
 
         return cb.like(cb.lower(expression), '%' + escapedReference.toLowerCase() + '%');
     }
-
-    public int countChargesForCapture(Duration notAttemptedWithin) {
-        String query = "SELECT count(c) FROM ChargeEntity c WHERE (c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus)"+
-                "AND NOT EXISTS (" +
-                "  SELECT ce FROM ChargeEventEntity ce WHERE " +
-                "    ce.chargeEntity = c AND " +
-                "    ce.status = :eventStatus AND " +
-                "    ce.updated >= :cutoffDate " +
-                ") ";
+    
+    private static final String FIND_CAPTURE_CHARGES_WHERE_CLAUSE = 
+            "WHERE (c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus)"+
+            "AND NOT EXISTS (" +
+            "  SELECT ce FROM ChargeEventEntity ce WHERE " +
+            "    ce.chargeEntity = c AND " +
+            "    ce.status = :eventStatus AND " +
+            "    ce.updated >= :cutoffDate " +
+            ") ";
+    
+    public int countChargesForImmediateCapture(Duration notAttemptedWithin) {
+        String query = "SELECT count(c) FROM ChargeEntity c " + FIND_CAPTURE_CHARGES_WHERE_CLAUSE;
 
         Number count = (Number) entityManager.get()
                 .createQuery(query)
@@ -185,36 +188,9 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
                 .getSingleResult();
         return count.intValue();
     }
-    
-    public int countChargesAwaitingCaptureRetry(Duration notAttemptedWithin) {
-        String query = "SELECT count(c) FROM ChargeEntity c WHERE (c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus)"+
-                "AND EXISTS (" +
-                "  SELECT ce FROM ChargeEventEntity ce WHERE " +
-                "    ce.chargeEntity = c AND " +
-                "    ce.status = :eventStatus AND " +
-                "    ce.updated >= :cutoffDate " +
-                ") ";
 
-        Number count = (Number) entityManager.get()
-                .createQuery(query)
-                .setParameter("captureApprovedStatus", CAPTURE_APPROVED.getValue())
-                .setParameter("captureApprovedRetryStatus", CAPTURE_APPROVED_RETRY.getValue())
-                .setParameter("eventStatus", CAPTURE_APPROVED_RETRY)
-                .setParameter("cutoffDate", ZonedDateTime.now().minus(notAttemptedWithin))
-                .getSingleResult();
-        return count.intValue();
-    }
-    
     public List<ChargeEntity> findChargesForCapture(int maxNumberOfCharges, Duration notAttemptedWithin) {
-        String query = "SELECT c FROM ChargeEntity c WHERE " +
-                "(c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus) " +
-                "AND NOT EXISTS (" +
-                "  SELECT ce FROM ChargeEventEntity ce WHERE " +
-                "    ce.chargeEntity = c AND " +
-                "    ce.status = :eventStatus AND " +
-                "    ce.updated >= :cutoffDate " +
-                ") " +
-                "ORDER BY c.createdDate ASC";
+        String query = "SELECT c FROM ChargeEntity c " + FIND_CAPTURE_CHARGES_WHERE_CLAUSE + "ORDER BY c.createdDate ASC";
 
         return entityManager.get()
                 .createQuery(query, ChargeEntity.class)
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -167,17 +167,44 @@ private Predicate likePredicate(CriteriaBuilder cb, Path<String> expression, Str
...
-    public int countChargesForCapture() {
-        String query = "SELECT count(c) FROM ChargeEntity c WHERE c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus";
+    public int countChargesForCapture(Duration notAttemptedWithin) {
+        String query = "SELECT count(c) FROM ChargeEntity c WHERE (c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus)"+
+                "AND NOT EXISTS (" +
+                "  SELECT ce FROM ChargeEventEntity ce WHERE " +
+                "    ce.chargeEntity = c AND " +
+                "    ce.status = :eventStatus AND " +
+                "    ce.updated >= :cutoffDate " +
+                ") ";
```
alexbishop1, line 212: So this query is intended to find all payments that are in the (conceptual) capture queue (those with statuses of `CAPTURE APPROVED` and `CAPTURE APPROVED RETRY`) except those that have had a failed capture attempt (indicated by a transition to `CAPTURE APPROVED RETRY`) within the last _x_ time units (`notAttemptedWithin`)?

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -167,17 +167,44 @@ private Predicate likePredicate(CriteriaBuilder cb, Path<String> expression, Str
...
 
         Number count = (Number) entityManager.get()
                 .createQuery(query)
                 .setParameter("captureApprovedStatus", CAPTURE_APPROVED.getValue())
                 .setParameter("captureApprovedRetryStatus", CAPTURE_APPROVED_RETRY.getValue())
+                .setParameter("eventStatus", CAPTURE_APPROVED_RETRY)
```
alexbishop1, line 186: Because this is only `CAPTURE_APPROVED_RETRY` (and not `CAPTURE_APPROVED`), if one calls this method with a duration of, say, 3 minutes, they will not get payments that have been scheduled for a retry within the last 3 minutes but they will get payments that have been scheduled for a first capture attempt within the last 3 minutes. Is this what you want?

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -167,17 +167,44 @@ private Predicate likePredicate(CriteriaBuilder cb, Path<String> expression, Str
...
     }
+    
+    public int countChargesAwaitingCaptureRetry(Duration notAttemptedWithin) {
+        String query = "SELECT count(c) FROM ChargeEntity c WHERE (c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus)"+
+                "AND EXISTS (" +
+                "  SELECT ce FROM ChargeEventEntity ce WHERE " +
+                "    ce.chargeEntity = c AND " +
+                "    ce.status = :eventStatus AND " +
+                "    ce.updated >= :cutoffDate " +
+                ") ";
```
alexbishop1, line 212: Am I right in thinking this finds all payments that are in the (conceptual) capture queue and have had a failed capture attempt within the last _x_ time units?

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -167,17 +167,44 @@ private Predicate likePredicate(CriteriaBuilder cb, Path<String> expression, Str
...
 
+        Number count = (Number) entityManager.get()
+                .createQuery(query)
+                .setParameter("captureApprovedStatus", CAPTURE_APPROVED.getValue())
```
alexbishop1: I don’t think you need this: if a payment has a `CAPTURE APPROVED RETRY` event, its status cannot be `CAPTURE APPROVED`

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -167,17 +167,44 @@ private Predicate likePredicate(CriteriaBuilder cb, Path<String> expression, Str
         return cb.like(cb.lower(expression), '%' + escapedReference.toLowerCase() + '%');
     }
 
-    public int countChargesForCapture() {
-        String query = "SELECT count(c) FROM ChargeEntity c WHERE c.status=:captureApprovedStatus OR c.status=:captureApprovedRetryStatus";
+    public int countChargesForCapture(Duration notAttemptedWithin) {
```
alexbishop1: It looks like this method is very similar to the existing `findChargesForCapture` except it does a count rather returning a list — potential to avoid code duplication?

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -29,44 +29,49 @@
...
 
         Stopwatch responseTimeStopwatch = Stopwatch.createStarted();
-        long captured = 0, skipped = 0, error = 0, total = 0;
+        long captured = 0, skipped = 0, error = 0, failedCapture = 0, total = 0;
 
         try {
-            queueSize = chargeDao.countChargesForCapture();
-
-            updateQueueSizeMetric(queueSize);
+            queueSize = chargeDao.countChargesForCapture(captureConfig.getRetryFailuresEveryAsJavaDuration());
```
alexbishop1: Due to the similarity between `ChargeDao.countChargesForCapture(Duration)` and `ChargeDao.findChargesForCapture(int, Duration)`, if the number of payments to capture is below `captureConfig.getBatchSize()` then there’s the possibility to wait until we have after `chargesToCapture` has been calculated and use the `size()` of that (obviously this may have performance implications if `captureConfig.getBatchSize()` is significantly smaller).

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -29,44 +29,49 @@
...
+            waitingQueueSize = chargeDao.countChargesAwaitingCaptureRetry(captureConfig.getRetryFailuresEveryAsJavaDuration());
 
             List<ChargeEntity> chargesToCapture = chargeDao.findChargesForCapture(captureConfig.getBatchSize(), captureConfig.getRetryFailuresEveryAsJavaDuration());
 
             if (chargesToCapture.size() > 0) {
-                logger.info("Capturing : " + chargesToCapture.size() + " of " + queueSize + " charges");
+                logger.info("Capturing : " + chargesToCapture.size() + " of " + waitingQueueSize + " charges");
```
alexbishop1: Not sure this going to say what you want: `waitingQueueSize` isn’t the total number of payments that are in the capture queue period, it’s just those that have had a failed capture attempt within the last `captureConfig.getRetryFailuresEveryAsJavaDuration()`. If I’m understanding it right, the total will be `queueSize + waitingQueueSize`.

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
```diff
@@ -29,44 +29,49 @@
...
             }
 
             Collections.shuffle(chargesToCapture);
```
alexbishop1: We shuffle the payments before capturing them (presumably to stop one that causes a catastrophic error spoiling everyone’s fun every time)? I never knew that!

alexbishop1 (changes requested):
Left a few comments — there’s only a couple that I think might be actual problems; the rest are really just questions to help me understand what exactly we’re trying to measure here.

I think some of the names (like `countChargesForCapture`, `queueSize` and `waitingQueueSize`) could be renamed to make it clearer what they represent.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
danworth: Correct.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
danworth: Yes, if we have already attempted to capture a charge within the `notAttemptedWithin` period then we do not want to include it in this count. If it's new, `CAPTURE APPROVED` then we should include it, even if it moved to that state within the `notAttemptedWithin`.  Is that what you mean?

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
danworth: Yes, that is the intention.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
danworth: Yes I think you're right. I've removed it and the tests still pass...

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
danworth: Great idea, I've implemented that logic.

src/main/java/uk/gov/pay/connector/service/CardCaptureProcess.java
danworth: Yes you are right, thank you! Updated to use the sum of the two counts.

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
danworth: Ok good point, I've factored out the common where clause.

heathd:
can you squash the 'pseudo code' commit as well?

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
alexbishop1: Yeah, I mostly just want to make sure we both agree on what the queries are supposed to be doing 😃

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #660, opened 2018-08-22
PP-4119 Improve the naming of capture queue size metrics.

danworth:
- Place both metrics within existing capture-process/queue-size folder.
- Rename `immediateCaptureQueueSize` to `readyCaptureQueueSize` which seems
more appropriate.
- Rename the metric that represents the number of charges which can be retried
on the next pass from `queue-size` to `readyCaptureQueueSize` for clarity since
there are now two metrics which describe the capture queue.


alexbishop1 (approved):
I think I like `ready` more than `immediate` anyway

PR #662, opened 2018-08-22
PP-4082: Refresh database

oswaldquek:
Because we need to run the contract tests multiple times (using pacts tagged
with 'master', 'test', 'staging' and 'production') we need to refresh the
database by deleting data.

Also took the opportunity to remove unused code.

@oswaldquek


brid (approved):
LGTM

PR #664, opened 2018-08-24
PP-3177 Escape backlashes in like statements

simad:
## WHAT
 - When fields (eg. reference) contain a backslash we were unable to search
   properly because the backslash is an escape character in postgres.
   For example, for a reference of `abc\def` we were searching for `abcdef`,
   returning the wrong charge.
 - Escaping backslashes (ie. substituting `\` with `\\`) fixes this

Also, because `\` is an escape char in java, too, in the code we actually have to replace `\\\\` with `\\\\\\\\` #justjavathings 💯


alexbishop1 (dismissed):
`\\\\\\\\`

alexbishop1 (approved):

PR #665, opened 2018-08-28
PP-3840 Add GIN index on cardholder name

simad:
## WHAT
We want to be able to do a partial search on cardholder name. Adding a GIN index will help with performance.


brid (approved):
LTGM. Approved!

PR #666, opened 2018-08-29
PP-3840 Allow searching by cardholder name and last digits of card number

simad:
## WHAT
- Adding both query params for searching payments via publicapi (old `/v1/charges` endpoint)
- Adding just the cardholder name query param for searching via selfservice (new `/v2/charges` endpoint)
- Cardholder name will match on partial search, eg. searching for `Jane` will match both `Jane Doe` and `Bob Somejane`, while last digit of card number must be provided in full, as in searching for `12` won't match `1234`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
@@ -140,6 +140,10 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
 
     private List<Predicate> buildParamPredicates(ChargeSearchParams params, CriteriaBuilder cb, Root<ChargeEntity> charge) {
         List<Predicate> predicates = new ArrayList<>();
+        if (params.getCardHolderName() != null && StringUtils.isNotBlank(params.getCardHolderName().toString()))
+            predicates.add(likePredicate(cb, charge.get(CARD_DETAILS).get("cardHolderName"), params.getCardHolderName().toString().toLowerCase()));
+        if (params.getLastDigitsCardNumber() != null && StringUtils.isNotBlank(params.getLastDigitsCardNumber().toString()))
+            predicates.add(cb.equal(charge.get(CARD_DETAILS).get("lastDigitsCardNumber"), params.getLastDigitsCardNumber().toString()));
         if (params.getGatewayAccountId() != null)
             predicates.add(cb.equal(charge.get(GATEWAY_ACCOUNT).get("id"), params.getGatewayAccountId()));
         if (params.getReference() != null && StringUtils.isNotBlank(params.getReference().toString()))
--- a/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
@@ -141,7 +141,7 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
     private List<Predicate> buildParamPredicates(ChargeSearchParams params, CriteriaBuilder cb, Root<ChargeEntity> charge) {
         List<Predicate> predicates = new ArrayList<>();
         if (params.getCardHolderName() != null && StringUtils.isNotBlank(params.getCardHolderName().toString()))
-            predicates.add(likePredicate(cb, charge.get(CARD_DETAILS).get("cardHolderName"), params.getCardHolderName().toString().toLowerCase()));
+            predicates.add(likePredicate(cb, charge.get(CARD_DETAILS).get("cardHolderName"), params.getCardHolderName().toString()));
         if (params.getLastDigitsCardNumber() != null && StringUtils.isNotBlank(params.getLastDigitsCardNumber().toString()))
             predicates.add(cb.equal(charge.get(CARD_DETAILS).get("lastDigitsCardNumber"), params.getLastDigitsCardNumber().toString()));
         if (params.getGatewayAccountId() != null)
--- a/src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
@@ -111,6 +111,11 @@ public class TransactionDao {
     private SelectOrderByStep buildQueryFor(Long gatewayAccountId, QueryType queryType, ChargeSearchParams params) {
         Condition queryFilters = field("c.gateway_account_id").eq(gatewayAccountId);
 
+        if (params.getCardHolderName() != null && isNotBlank(params.getCardHolderName().toString())) {
+            queryFilters = queryFilters.and(
+                    field("c.cardholder_name").lower().like(buildLikeClauseContaining(params.getCardHolderName().toString().toLowerCase())));
+        }
+        
         if (isNotBlank(params.getEmail())) {
             queryFilters = queryFilters.and(
                     field("c.email").lower().like(buildLikeClauseContaining(params.getEmail().toLowerCase())));
@@ -245,13 +250,13 @@ public class TransactionDao {
 
     private Set<String> mapChargeStatuses(Set<ChargeStatus> status) {
         return status.stream()
-                .map(s -> s.getValue())
+                .map(ChargeStatus::getValue)
                 .collect(Collectors.toSet());
     }
 
     private Set<String> mapRefundStatuses(Set<RefundStatus> status) {
         return status.stream()
-                .map(s -> s.getValue())
+                .map(RefundStatus::getValue)
                 .collect(Collectors.toSet());
     }
 }
--- a/src/main/java/uk/gov/pay/connector/model/CardHolderName.java
+++ b/src/main/java/uk/gov/pay/connector/model/CardHolderName.java
@@ -1,0 +1,36 @@
+package uk.gov.pay.connector.model;
+
+import java.util.Objects;
+
+public class CardHolderName {
+
+    private final String cardHolderName;
+
+    private CardHolderName(String cardHolderName) {
+        this.cardHolderName = Objects.requireNonNull(cardHolderName);
+    }
+
+    public static CardHolderName of(String cardHolderName) {
+        return new CardHolderName(cardHolderName);
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        if (other != null && other.getClass() == CardHolderName.class) {
+            CardHolderName that = (CardHolderName) other;
+            return this.cardHolderName.equals(that.cardHolderName);
+        }
+        return false;
+    }
+
+    @Override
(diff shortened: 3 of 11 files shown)


src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
```diff
@@ -140,6 +140,10 @@ public Long getTotalFor(ChargeSearchParams params) {
 
     private List<Predicate> buildParamPredicates(ChargeSearchParams params, CriteriaBuilder cb, Root<ChargeEntity> charge) {
         List<Predicate> predicates = new ArrayList<>();
+        if (params.getCardHolderName() != null && StringUtils.isNotBlank(params.getCardHolderName().toString()))
+            predicates.add(likePredicate(cb, charge.get(CARD_DETAILS).get("cardHolderName"), params.getCardHolderName().toString().toLowerCase()));
```
alexbishop1: I don’t think we need the `toLowerCase()` here: the `likePredicate(CriteriaBuilder, Path<String>, String)` method will take care of that for us.

Using `.toLowerCase()` is equivalent to`.toLowerCase(Locale.getDefault())`, which is probably equivalent to `.toLowerCase(Locale.ENGLISH)` or `.toLowerCase(Locale.UK)` for us. Due to different languages having different lower-casing rules, this may make the search do interesting things with non-English names. There’s also the issue of it probably being desirable to have “garcia” match “García”. Probably not something we need to address now (this is a complex problem) but worth bearing mind for the future.

src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
```diff
@@ -111,6 +111,11 @@ public Long getTotalFor(Long gatewayAccountId, ChargeSearchParams params) {
     private SelectOrderByStep buildQueryFor(Long gatewayAccountId, QueryType queryType, ChargeSearchParams params) {
         Condition queryFilters = field("c.gateway_account_id").eq(gatewayAccountId);
 
+        if (params.getCardHolderName() != null && isNotBlank(params.getCardHolderName().toString())) {
+            queryFilters = queryFilters.and(
+                    field("c.cardholder_name").lower().like(buildLikeClauseContaining(params.getCardHolderName().toString().toLowerCase())));
+        }
```
alexbishop1, line 117: Why do we not need something similar for the last four digits?

src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
```diff
@@ -245,13 +250,13 @@ private String buildLikeClauseContaining(String textToFind) {
 
     private Set<String> mapChargeStatuses(Set<ChargeStatus> status) {
         return status.stream()
-                .map(s -> s.getValue())
+                .map(ChargeStatus::getValue)
```
alexbishop1, line 253: 🥇

src/main/java/uk/gov/pay/connector/model/CardHolderName.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.pay.connector.model;
+
+import java.util.Objects;
+
+public class CardHolderName {
```
alexbishop1, line 5: 😍

src/main/java/uk/gov/pay/connector/model/LastDigitsCardNumber.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.pay.connector.model;
+
+import java.util.Objects;
+
+public class LastDigitsCardNumber {
+
+    private final String lastDigitsCardNumber;
+
+    private LastDigitsCardNumber(String lastDigitsCardNumber) {
+        this.lastDigitsCardNumber = Objects.requireNonNull(lastDigitsCardNumber);
```
alexbishop1, line 10: It would be nice to be able to guarantee that this string contains exactly four digits. Unfortunately, it looks like there’s no validation on the incoming search parameter so this can contain whatever string was entered.

So, open question: could we validate the incoming last four digits search parameter? Would that be a good idea?

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -131,7 +134,94 @@ public void searchChargesByPartialEmailMatch() throws Exception {
...
+                .insert();
+        ChargeSearchParams params = new ChargeSearchParams()
+                .withCardHolderNameLike(CardHolderName.of(testCharge.cardDetails.getCardHolderName()));
+
+        // when
+        List<ChargeEntity> charges = chargeDao.findAllBy(params);
+
+        // then
+        assertThat(charges.size(), is(1));
+        assertCharge("visa", testCharge, charges.get(0));
```
alexbishop1: Surely asserting on the cardholder name would be better?

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -131,7 +134,94 @@ public void searchChargesByPartialEmailMatch() throws Exception {
...
+        Long chargeId = 12L;
+        TestCharge testCharge = DatabaseFixtures
+                .withDatabaseTestHelper(databaseTestHelper)
+                .aTestCharge()
+                .withTestAccount(defaultTestAccount)
+                .withCardDetails(defaultTestCardDetails.withChargeId(chargeId))
+                .withChargeId(chargeId)
+                .insert();
+        ChargeSearchParams params = new ChargeSearchParams()
+                .withCardHolderNameLike(CardHolderName.of("pay"));
```
alexbishop1: It might clearer what’s going on if we explicitly set the cardholder name in the fixture, so the fact that the search parameter is a substring is more obvious

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceGetChargesJsonITest.java
```diff
@@ -367,6 +367,24 @@ public void shouldFilterTransactionsByEmail() throws Exception {
...
+        getChargeApi
+                .withAccountId(accountId)
+                .withQueryParam("cardholder_name", "McPayment")
+                .withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON)
+                .getTransactionsAPI()
+                .statusCode(OK.getStatusCode())
+                .contentType(JSON)
+                .body("results.size()", is(3))
+                .body("results[0].card_details.cardholder_name", is("Mr. McPayment"));
+    }
```
alexbishop1: Test for last four digits as well?

alexbishop1 (changes requested):
Looks good: just a few small things

src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
simad: Because we decided not to have that filtering in selfservice :/ it's the bit of work I had to remove

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
simad: 😳

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
simad: Nicely spotted, didn't think about the locale. It is true, we call the default `toLowerCase()` when escaping the like predicate, so it's something to think about. I have to say we are not doing much about l18n, in general, for obvious reasons.

src/main/java/uk/gov/pay/connector/model/LastDigitsCardNumber.java
simad: I originally had added that validation in this class, checking both that it's 4 digits and that it's fully numeric. I was throwing an exception in the `.of()` method if those criteria were not met.

Thing is, at resource level, we either return a bad response or we do something a bit nasty (ie. return a default) - considering we are not validating any query param and the default would probably simply not ever match, I've decided to leave it like this, as it won't match anything unless it's 4 valid digits anyway

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
simad: 👍

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceGetChargesJsonITest.java
simad: 👍

src/main/java/uk/gov/pay/connector/dao/ChargeDao.java
simad: Also if somebody is called “García”, that's his problem @rauligs

rauligs: I promise I won't use accents! 🤐

src/main/java/uk/gov/pay/connector/model/LastDigitsCardNumber.java
alexbishop1: In theory, we could validate it to check it’s four digits (and do we do validate things like date formats).

I’ll be sad if we ever up in the situation where we’re using the `LastDigitsCardNumber` to represent something that we _know_ is exactly four digits and we’re unable to enforce that, but I suppose we could always split it into two objects then (one representing the search parameter and the other a confirmed valid value).

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceGetChargesJsonITest.java
alexbishop1: Sorry, I meant here to have another test to check filtering by the last four digits

alexbishop1 (changes requested):
Thanks for making the changes.

I still think `ChargesApiResourceGetChargesJsonITest` should have a test to check filtering by the last four digits, but other than I think it’s good to go.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceGetChargesJsonITest.java
alexbishop1: Whoops! This tests the transactions endpoint, which is only used by selfservice so doesn’t need to support searching for the last four digits.

alexbishop1 (approved):
I was mistaken about `ChargesApiResourceGetChargesJsonITest`: it tests the transactions endpoint, which is only used by selfservice so doesn’t need to support searching for the last four digits

PR #667, opened 2018-08-30
PP-4105 Add delayed_capture column to charges table.

danworth:
- Create new boolean column called `delayed_capture` in charges table.
- Set default to false.
- Updated existing rows to false.

@alexbishop1


belindac (approved):

whpearson (approved):
Looks good.

PR #670, opened 2018-09-03
PP-4105 Add delayed capture flag to charge and persist to database

alexbishop1:
- Add delayed capture boolean flag to `ChargeEntity` such that it is persisted to the database

- When creating a payment, set the delayed capture boolean on the `ChargeEntity` based on what was in the create payment request, defaulting to false if none was supplied

with @danworth


simad (dismissed):
🐑 it

simad (approved):
💯

PR #671, opened 2018-09-03
PP-3875 Decode epdq notifications using Window-1252 charset

simad:
## WHAT
 - We were previously decoding epdq notifications using UTF-8, and  this was causing some problems with epdq. I could reproduce the issue in test, so after this pr is merged I'll be able to check if the problem is solved - if we still have validation failed errors, we'll probably need to do more work on this.


oswaldquek (approved):

oswaldquek:
LGTM

PR #673, opened 2018-09-04
PP-4105 Add delayed_capture to payment JSON responses

alexbishop1:
Add `delayed_capture` property to JSON returned when getting a charge from an API endpoint (including charges, transactions and frontend charges endpoints)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
+++ b/src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
@@ -249,7 +249,7 @@ public class ChargeResponse {
         public ChargeResponse build() {
             return new ChargeResponse(chargeId, amount, state, cardBrand, gatewayTransactionId, returnUrl, email,
                     description, reference, providerName, createdDate, links, refundSummary, settlementSummary,
-                    cardDetails, auth3dsData, language);
+                    cardDetails, auth3dsData, language, delayedCapture);
         }
     }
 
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
```diff
@@ -409,61 +418,32 @@ public URI getLink(String rel) {
...
-            return false;
-        if (settlementSummary != null ? !settlementSummary.equals(that.settlementSummary) : that.settlementSummary != null)
-            return false;
-        if (auth3dsData != null ? !auth3dsData.equals(that.auth3dsData) : that.auth3dsData != null)
-            return false;
-        if (cardDetails != null ? !cardDetails.equals(that.cardDetails) : that.cardDetails != null)
-            return false;
-        return language.equals(that.language);
+        return delayedCapture == that.delayedCapture &&
+                Objects.equals(dataLinks, that.dataLinks) &&
```
DanailMinchev, line 423: I think we are using "IntelliJ IDEA" style of `equals` in most of the places, but I don't mind if we use `Objects.equals` which is probably more standard way now

DanailMinchev (approved):

src/main/java/uk/gov/pay/connector/model/ChargeResponse.java
alexbishop1: This was generated by IntelliJ IDEA. Join the new world!

In general, I think `Objects.equals(…)` (which comes from the Java standard library) is more standard that whatever the IntelliJ decides to use this year, so I’d prefer to standardise on that if anything (usually caveats about not blindly including every field in `equals()` and being cautious about changing the `equals()` method of objects between releases apply).

PR #677, opened 2018-09-04
PP-4108 Add new AWAITING CAPTURE REQUEST charge state

danworth:
- Added new charge state `AWAITING_CAPTURE_REQUEST`
- Added new state to `PaymentGatewayStateTransitions` with edges to appropriate
states.
- Added new state to expirable states in `ChargeExpiryService`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/PaymentGatewayStateTransitions.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/PaymentGatewayStateTransitions.java
@@ -19,6 +19,7 @@ import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_SUBMI
 import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_TIMEOUT;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.AUTHORISATION_UNEXPECTED_ERROR;
+import static uk.gov.pay.connector.model.domain.ChargeStatus.AWAITING_CAPTURE_REQUEST;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.CAPTURED;
 import static uk.gov.pay.connector.model.domain.ChargeStatus.CAPTURE_APPROVED;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/model/domain/PaymentGatewayStateTransitions.java
```diff
@@ -96,6 +97,11 @@ private PaymentGatewayStateTransitions() {
         graph.putEdgeValue(AUTHORISATION_SUCCESS, SYSTEM_CANCEL_READY, "");
         graph.putEdgeValue(AUTHORISATION_SUCCESS, USER_CANCEL_READY, "");
         graph.putEdgeValue(AUTHORISATION_SUCCESS, EXPIRE_CANCEL_READY, "");
+        graph.putEdgeValue(AUTHORISATION_SUCCESS, AWAITING_CAPTURE_REQUEST, "");
+        graph.putEdgeValue(AWAITING_CAPTURE_REQUEST, CAPTURE_APPROVED, "");
+        graph.putEdgeValue(AWAITING_CAPTURE_REQUEST, SYSTEM_CANCEL_READY, "");
+        graph.putEdgeValue(AWAITING_CAPTURE_REQUEST, EXPIRE_CANCEL_READY, "");
+        graph.putEdgeValue(AWAITING_CAPTURE_REQUEST, EXPIRED, "");
```
alexbishop1: I don’t think we want this one

alexbishop1 (changes requested):
Looks good with the exception of one transition I don’t think we want

src/main/java/uk/gov/pay/connector/model/domain/PaymentGatewayStateTransitions.java
danworth: Ok, adding the new state to the expire process is a little more involved and requires the state to be added to more than just the `EXPIRABLE_STATUSES` within `ChargeExpiryService`. I've taken that step out so that this PR is simply adding the new state to `ChargeState` and the necessary edges as defined in the story to `PaymentGatewayStateTransitions`

alexbishop1: Yeah, good idea.

alexbishop1 (approved):

PR #679, opened 2018-09-06
Pp 4108 set delayed capture charges to awaiting capture retry

georgeracu:
## WHAT

When `connector` receives a request to capture a charge, then it checks if the charge has enabled `delayed capture` and then transitions that charge to either `CAPTURE_APPROVED` or `AWAITING_CAPTURE_REQUEST`


src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
```diff
@@ -53,7 +54,7 @@ public CardCaptureService(ChargeDao chargeDao, ChargeEventDao chargeEventDao, Pa
         try {
             charge = preOperation(externalId);
         } catch (OptimisticLockException e) {
-            LOG.info("OptimisticLockException in doCapture for charge external_id=" + externalId);
+            LOG.info("OptimisticLockException in doCapture for charge external_id={}", externalId);
```
DanailMinchev, line 57: Good optimisation 👍 
Not sure why we actually have `LOG`, can we use `logger` instead?

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
```diff
@@ -69,24 +70,28 @@ public ChargeEntity preOperation(String chargeId) {
...
+            ChargeStatus currentChargeStatus = ChargeStatus.fromString(charge.getStatus());
+            if (!isValidTransition(currentChargeStatus, targetStatus)) {
+                LOG.error("Charge with state " + currentChargeStatus + " cannot proceed to " + targetStatus +
+                        " [charge_external_id={}, charge_status={}]", charge.getExternalId(), currentChargeStatus);
                 throw new IllegalStateRuntimeException(charge.getExternalId());
             }
 
-            logger.info("CAPTURE_APPROVED for charge [charge_external_id={}]", externalId);
-            charge.setStatus(CAPTURE_APPROVED);
+            LOG.info("{} for charge [charge_external_id={}]", targetStatus, externalId);
```
DanailMinchev, line 85: now I see you change it to `LOG` everywhere, which is probably better and more explicit

DanailMinchev (approved):
LGTM 👍

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
georgeracu: `LOG` has been added by Raul when he was investigating some bugs so I used it, as it was there. Also makes the logs more explicit to which class they are issued from.

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
georgeracu: 👍

PR #682, opened 2018-09-07
PP-4106 Add get payment with delayed_capture true pact test

SandorArpa:
- add setup for new pact test 'createChargeWithDelayedCaptureTrue'

with @danailminchev

### DIFF
--- a/src/test/java/uk/gov/pay/connector/pact/TransactionsApiContractTest.java
+++ b/src/test/java/uk/gov/pay/connector/pact/TransactionsApiContractTest.java
@@ -56,14 +56,18 @@ public class TransactionsApiContractTest {
         }
     }
 
-    private void setUpCharges(int numberOfCharges, String accountID, ZonedDateTime createdDate){
+    private void setUpCharges(int numberOfCharges, String accountId, ZonedDateTime createdDate){
         for (int i = 0; i < numberOfCharges; i++) {
             Long chargeId = ThreadLocalRandom.current().nextLong(100, 100000);
-            dbHelper.addCharge(chargeId, Long.toString(chargeId), accountID, 100L, ChargeStatus.CREATED,"aReturnUrl",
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/pact/TransactionsApiContractTest.java
```diff
@@ -66,6 +66,16 @@ private void setUpCharges(int numberOfCharges, String accountID, ZonedDateTime c
         }
     }
 
+    private void setUpCharges(int numberOfCharges, String accountID, String chargeExternalId, ZonedDateTime createdDate, boolean delayedCapture){
```
alexbishop1: Can we make this share some code with the other method with the same name e.g. by having that one call this one with `delayedCapture` as `false`?

src/test/java/uk/gov/pay/connector/pact/TransactionsApiContractTest.java
```diff
@@ -66,6 +66,16 @@ private void setUpCharges(int numberOfCharges, String accountID, ZonedDateTime c
...
+        for (int i = 0; i < numberOfCharges; i++) {
+            Long chargeId = ThreadLocalRandom.current().nextLong(100, 100000);
+            dbHelper.addCharge(chargeId, chargeExternalId, accountID, 100L, ChargeStatus.CREATED,"aReturnUrl",
```
alexbishop1: This is going to add lots of charges with the identical external IDs, which isn’t a realistic reflection of the world

georgeracu (dismissed):
LGTM

georgeracu (dismissed):
👍

alexbishop1 (changes requested):
Looks good, though I think as we add more of these we’ll need to think carefully about how we organise and compose them. For now, we can go with this.

Just a couple of comments.

alexbishop1 (approved):
Looks fine.

We will probably have to make this code more reusable at some point in the future but there’s no point rushing to do that now.

PR #685, opened 2018-09-10
Use single Charset instance for all ePDQ form encoding needs

alexbishop1:
`Charset` objects are thread-safe. We already share a single instance for constructing `application/x-www-form-urlencoded` payloads to send to ePDQ. However, when decoding `application/x-www-form-urlencoded notifications` from ePDQ, we were asking (via the `Charset.forName(String)` method) for a `Charset` object every time (there’s a cache so it won’t have actually created a new object every time).

- Move the `Charset` constant from `EpdqOrderRequestBuilder` to `EpdqPaymentProvider`

- Add Javadoc to this constant to aid our future selves

- Use `Charset` constant in `EpdqNotification` for decoding

- Update outdated log line in `EpdqNotification`

### DIFF
--- a/src/test/java/uk/gov/pay/connector/service/epdq/EpdqNotificationTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/epdq/EpdqNotificationTest.java
@@ -91,13 +91,13 @@ public class EpdqNotificationTest {
 
     @Test
     public void shouldDecodeNotificationsAccordingToTheRightEpdqCharset() {
-        String cardHolderName = "mr payment%92"; //encoded value for ’ (single quote)
+        String cardHolderName = "Mr O%92Payment"; // %92 is encoded value for ’ (right single quotation mark)
         String payload = notificationPayloadForTransaction(cardHolderName, STATUS, PAY_ID, PAY_ID_SUB, SHA_SIGN);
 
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/service/epdq/EpdqNotificationTest.java
```diff
@@ -91,13 +91,13 @@ public void shouldReturnParams() {
...
     public void shouldDecodeNotificationsAccordingToTheRightEpdqCharset() {
-        String cardHolderName = "mr payment%92"; //encoded value for ’ (single quote)
+        String cardHolderName = "Mr O%92Payment"; // %92 is encoded value for ’ (right single quotation mark)
         String payload = notificationPayloadForTransaction(cardHolderName, STATUS, PAY_ID, PAY_ID_SUB, SHA_SIGN);
 
         EpdqNotification epdqNotification = new EpdqNotification(payload);
 
         assertThat(epdqNotification.getParams(), hasItem(
-                new BasicNameValuePair("CN", "mr payment’")));
+                new BasicNameValuePair("CN", "Mr O’Payment")));
```
DanailMinchev, line 100: `’` is more visible now! 👍

DanailMinchev (approved):
LGTM 👍

PR #686, opened 2018-09-10
PP-4106 Use ZonedDateTime.now() for Pact tests

DanailMinchev:
## WHAT

Use `ZonedDateTime.now()` for Pact tests instead of `ZonedDateTime.of()` because it fails with invalid parsing of milliseconds


SandorArpa (dismissed):
LOKTM

SandorArpa (approved):
LOKTM

PR #689, opened 2018-09-11
BAU - Small code improvements

georgeracu:
## HOW 

- Reordered modifiers on class/field declarations to be easier to read
- Renamed fields to avoid confusion with same name in lower case
- Added missing `@Override` annotations
- Changed field access modifiers to be less visible than necessary (private, package private vs public)
- Replaced lambda with method reference
- Removed unnecessary `format` from loggers
- Replaced `switch` construct with `if` as there was only one branch of `switch`
- Replaced map initialisation with new map to avoid possible memory leaks
- Removed some beloved white space

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForMaintenanceOrder.java
+++ b/src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForMaintenanceOrder.java
@@ -9,12 +9,12 @@ import static uk.gov.pay.connector.service.epdq.EpdqPayloadDefinition.newParamet
 
 public class EpdqPayloadDefinitionForMaintenanceOrder implements PayloadDefinition<EpdqTemplateData> {
 
-    final static String AMOUNT = "AMOUNT";
-    final static String OPERATION_KEY = "OPERATION";
-    final static String PAYID_KEY = "PAYID";
-    final static String PSPID_KEY = "PSPID";
-    final static String PSWD_KEY = "PSWD";
-    final static String USERID_KEY = "USERID";
+    static final String AMOUNT_KEY = "AMOUNT";
+    static final String OPERATION_KEY = "OPERATION";
+    static final String PAYID_KEY = "PAYID";
+    static final String PSPID_KEY = "PSPID";
+    static final String PSWD_KEY = "PSWD";
+    static final String USERID_KEY = "USERID";
 
     @Override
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForMaintenanceOrder.java
```diff
@@ -9,12 +9,12 @@
 
 public class EpdqPayloadDefinitionForMaintenanceOrder implements PayloadDefinition<EpdqTemplateData> {
 
-    final static String AMOUNT = "AMOUNT";
-    final static String OPERATION_KEY = "OPERATION";
-    final static String PAYID_KEY = "PAYID";
-    final static String PSPID_KEY = "PSPID";
-    final static String PSWD_KEY = "PSWD";
-    final static String USERID_KEY = "USERID";
+    static final String AMOUNT_KEY = "AMOUNT";
```
DanailMinchev, line 12: 👍

src/test/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java
```diff
@@ -49,10 +49,10 @@ public void testBaseValuePairsWithAmount(){
         BasicNameValuePair merchantCode = new BasicNameValuePair(PSPID_KEY, "Merchant-code");
         BasicNameValuePair password = new BasicNameValuePair(PSWD_KEY, "Password");
         BasicNameValuePair userId = new BasicNameValuePair(USERID_KEY, "User-id");
-        BasicNameValuePair amount = new BasicNameValuePair(AMOUNT, "400");
+        BasicNameValuePair amount = new BasicNameValuePair(AMOUNT_KEY, "400");
```
DanailMinchev, line 57: Could you also remove wildcard imports, like https://github.com/alphagov/pay-connector/blob/8cf4e7281ee1367d20927934e606c11eed28b644/src/test/java/uk/gov/pay/connector/service/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java#L10, it was not clear where `AMOUNT` was coming if you just review in GitHub, thanks

DanailMinchev (dismissed):
Cool! 👍

DanailMinchev (dismissed):

belindac (approved):

PR #693, opened 2018-09-11
Pp 4165 create capture resource

georgeracu:
## WHAT

* Added a new POST resource in `CardResource` that will be called on
`"/v1/api/accounts/{accountId}/charges/{chargeId}/capture"` when a service
will want to mark a delayed charge as `CAPTURE APPROVED`.

* When a charge is in status `AWAITING CAPTURE REQUEST` then we transition
to status `CAPTURE APPROVED`, we record an event and we return 204

* When a charge is in statuses `CAPTURED, CAPTURE APPROVED,
CAPTURE APPROVED RETRY, CAPTURE READY, CAPTURE SUBMITTED` then we
do not transition and we return 204

* When a charge is in a status that cannot transition to (none of the above)
then we return 409

* When a charge it doesn't exist then we return 404

* Updated documentation

with @danworth

### DIFF
--- a/src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
+++ b/src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
@@ -19,11 +19,8 @@ import uk.gov.pay.connector.model.gateway.GatewayResponse;
 import javax.inject.Inject;
 import javax.persistence.OptimisticLockException;
 import java.time.ZonedDateTime;
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
-import java.util.Set;
 
 import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.isBlank;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
```diff
@@ -39,6 +46,13 @@
             CAPTURE_APPROVED,
             CAPTURE_APPROVED_RETRY
     );
+    private static final Set<ChargeStatus> IGNORABLE_CAPTURE_STATES = new HashSet<>(Arrays.asList(
+            CAPTURE_APPROVED,
+            CAPTURE_APPROVED_RETRY,
+            CAPTURE_READY,
+            CAPTURE_SUBMITTED,
+            CAPTURED
+    ));
```
SandorArpa: Fully understand why we're using Set but for the sake of consistency within a class I suggest to change `List<ChargeStatus> legalStatuses` to a Set as well

SandorArpa (changes requested):

src/main/java/uk/gov/pay/connector/service/CardCaptureService.java
danworth: Great idea, will do!

SandorArpa (approved):
LOKTM

PR #694, opened 2018-09-12
PP-4108 Omit nextUrl for Awaiting Capture Request status

danworth:
- When returning charges the nextUrl should be omitted for charges with states
`AWAITING CAPTURE REQUEST` similiary to other states considered final.
- `AWAITING CAPTURE REQUEST` maps to `submitted` external state which is  not
a final state. Since the existing check only considers whether the external state
is final it has been extended to also explicitly exclude `AWAITING CAPTURE REQUEST`.
- Added tests to ChargeServiceTest and tidied up a lint duplicate code warning
by factoring out the duplicated code.


DanailMinchev (dismissed):
LGTM 👍

belindac (approved):

PR #696, opened 2018-09-13
Pp 4113 basic get refunds endpoint

cobainc0:
## WHAT
Adds new Refunds endpoint, which responds with all refunds for a Service if refund(s) exist
## HOW 
_Steps to test or reproduce:_
Pact tests: 
Return all refunds when refunds exist 
Return no refunds when no refunds exist

### DIFF
--- a/src/main/java/uk/gov/pay/connector/dao/RefundDao.java
+++ b/src/main/java/uk/gov/pay/connector/dao/RefundDao.java
@@ -8,12 +8,16 @@ import uk.gov.pay.connector.model.domain.RefundStatus;
 
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
+import javax.persistence.Query;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+
 
 @Transactional
 public class RefundDao extends JpaDao<RefundEntity> {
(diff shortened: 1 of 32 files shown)


src/main/java/uk/gov/pay/connector/dao/RefundDao.java
```diff
@@ -97,4 +101,47 @@ public RefundDao(final Provider<EntityManager> entityManager) {
...
+            predicates.add(cb.greaterThanOrEqualTo(refundEntityRoot.get(CREATED_DATE), params.getFromDate()));
+        if (params.getToDate() != null)
+            predicates.add(cb.lessThan(refundEntityRoot.get(CREATED_DATE), params.getToDate()));
+
+        return predicates;
+    }
+
+    public List<RefundEntity> findAllBy(ChargeSearchParams params) {
+        CriteriaBuilder cb = entityManager.get().getCriteriaBuilder();
+        CriteriaQuery<RefundEntity> cq = cb.createQuery(RefundEntity.class);
```
simad, line 135: I can see this DAO being generalised because it's basically the same as the ChargeDao, but marshalling to RefundEntity rather than ChargeEntity. But it's an improvement which can be done as part of another PR. 👍 for now

src/main/java/uk/gov/pay/connector/service/SearchRefundsService.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public Response getAllRefunds(UriInfo uriInfo, Long accountId, Long pageNumber, Long displaySize) {
+        List<Pair<String, Long>> queryParams = ImmutableList.of(
+                Pair.of("page", pageNumber),
+                Pair.of("display_size", displaySize));
+
+        return validateQueryParams(queryParams)
+                .map(ResponseUtil::badRequestResponse)
+                .orElseGet(() -> {
+                    ChargeSearchParams searchParams = new ChargeSearchParams()
```
simad: `ChargeSearchParams` is not really related to charge anymore. Could rename simply to `SearchParams` or something along those lines

src/main/java/uk/gov/pay/connector/resources/ChargesPaginationResponseBuilder.java
```diff
@@ -32,7 +30,7 @@ public ChargesPaginationResponseBuilder(ChargeSearchParams searchParams, UriInfo
         selfLink = uriWithParams(searchParams.buildQueryParams());
     }
 
-    public ChargesPaginationResponseBuilder withChargeResponses(List<ChargeResponse> chargeResponses) {
```
simad, line 35: We are generifying (which is probably not a real word, but 🤷‍♀️ ) the builder so that it's not Charge specific anymore. It should probably be renamed to `PaginationResponseBuilder` or something like that

simad: Same for `chargeResponses`

src/main/java/uk/gov/pay/connector/service/SearchRefundsService.java
```diff
@@ -0,0 +1,129 @@
...
+                            .withGatewayAccountId(accountId)
+                            .withDisplaySize(displaySize != null ? displaySize : 0)
+                            .withPage(pageNumber != null ? pageNumber : 1);
+                    return search(searchParams, uriInfo);
+                });
+    }
+    
+    private Response search(ChargeSearchParams searchParams, UriInfo uriInfo) {
```
simad: This is duplicating `AbstractSearchStrategy`. This is not really great because we introduce two places where the same logic is stored, so two places to change, we need to add tests etc. etc.


You can implement a RefundSearchStrategy which extends `AbstractSearchStrategy`. 
It seems you can get rid of `  private final CardTypeDao cardTypeDao;` and remove the card type from `abstract protected R buildResponse(UriInfo uriInfo, T transaction);`.
The abstract search strategy will have two types, T and R. R is `ChargeResponse` or `SearchRefundsResponse`, depending on the search strategy.

This is the first way that came to my mind for avoiding duplication. Let me know if you need help refactoring this.

src/main/java/uk/gov/pay/connector/service/SearchRefundsService.java
```diff
@@ -0,0 +1,129 @@
...
+    private ChargeResponse.RefundSummary buildRefundSummary(RefundEntity refundEntity) {
+        ChargeEntity chargeEntity = refundEntity.getChargeEntity();
+        ChargeResponse.RefundSummary refund = new ChargeResponse.RefundSummary();
+        refund.setStatus(providers.byName(chargeEntity.getPaymentGatewayName()).getExternalChargeRefundAvailability(chargeEntity).getStatus());
+        refund.setAmountSubmitted(chargeEntity.getRefundedAmount());
+        refund.setAmountAvailable(chargeEntity.getTotalAmountToBeRefunded());
+        return refund;
+    }
+
+    private static Optional<List> validateQueryParams(List<Pair<String, Long>> nonNegativePairMap) {
```
simad: In general, `Optional<List>` makes _never_ sense. If you return a List of errors, the list is either empty (meaning no error) or full (meaning there are errors). You don't need to represent that via an optional.

src/main/java/uk/gov/pay/connector/service/SearchRefundsService.java
```diff
@@ -0,0 +1,129 @@
...
+
+        nonNegativePairMap.forEach(param -> {
+            if (param.getRight() != null && param.getRight() < 1) {
+                invalidQueryParams.put(param.getLeft(), "query param '%s' should be a non zero positive integer");
+            }
+        });
+
+        if (!invalidQueryParams.isEmpty()) {
+            List<String> invalidResponse = newArrayList();
+            invalidResponse.addAll(invalidQueryParams.keySet()
```
simad: you don't need to addAll, you can just map directly. Ie. it should be possible to do
```
List<String> bla = invalidQueryParams.keySet().stream()
                    .map(param -> String.format(invalidQueryParams.get(param), param))
                    .collect(Collectors.toList())

```

simad (changes requested):
Good work. Quite a few comments to address, but looks promising!

src/main/java/uk/gov/pay/connector/service/SearchRefundsService.java
cobainc0: Thanks. This was very informative. Updated!

src/main/java/uk/gov/pay/connector/resources/ChargesPaginationResponseBuilder.java
cobainc0: Sounds like a word to me.

simad (approved):

simad:
There is a test which should be removed and moved to a new package. Happy to approve for now and do that later. 👍

PR #697, opened 2018-09-13
PP-4111 Add type to notifications and email_collection_mode to gateway accounts

simad:
## WHAT
 - Notifications can be, for now, of type `REFUND_ISSUED` or `PAYMENT_CONFIRMED`. We used to only
   have confirmation notifications, so we are going to backfill existing values.
 - `email_collection_mode` is at a Gateway Account level and needs to be backfilled as
   mandatory.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1482,4 +1482,36 @@
         </update>
     </changeSet>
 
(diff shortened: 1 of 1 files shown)


belindac (dismissed):

DanailMinchev (approved):
LGTM 👍

PR #698, opened 2018-09-13
PP-4110 Add a new Pact state for capture payment

DanailMinchev:
## WHAT

- Add `a charge with delayed capture true and awaiting capture request status exists` state which will be used in publicapi Pact tests for capturing a payment

with @SandorArpa


danworth (approved):
LGTM 👍

PR #702, opened 2018-09-17
PP-4122 add corporate surcharge columns to gateway accounts

SandorArpa:
## WHAT
add corporate surcharge columns

## HOW 

- add corporate_credit_card_surcharge_amount and corporate_debit_card_surcharge_amount columns to gateway_accounts table
- set default value to 0
- back fill existing records

with @danailminchev

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1442,4 +1442,44 @@
         </update>
     </changeSet>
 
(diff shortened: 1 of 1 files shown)


alexbishop1 (approved):
Assuming you’ve tested this locally, this looks good.

Modifying every gateway account is way less scary than modifying every charge but keep an eye on how this performs on staging all the same.

PR #703, opened 2018-09-18
PP-4111 Add support for refund issued emails

simad:
## WHAT
 - Adding type to email notifications, to distinguish between refund and confirmation
 - A gateway account can now have multiple notifications (at least two)
 - Made the patch endpoint modify both `template_body` and `enabled` of a notification, so that
   the post endpoint can be made redundant and removed once selfservice is merged
 - Email notifications is now serialised as part of the gateway account object, this removes
   the need to the email notifications dao which will be removed once the `GET` email notification
   endpoint will be removed
- Sending refund issued email when a refund is issued (duh)


src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -82,13 +83,13 @@ public DataSourceFactory getDataSourceFactory(ConnectorConfiguration configurati
...
     @Override
-    public void run(ConnectorConfiguration configuration, Environment environment) throws Exception {
+    public void run(ConnectorConfiguration configuration, Environment environment) {
         final Injector injector = createInjector(environment, new ConnectorModule(configuration, environment));
 
         injector.getInstance(PersistenceServiceInitialiser.class);
 
         initialiseMetrics(configuration, environment);
-
+        environment.jersey().register(new ValidationExceptionMapper());
```
alexbishop1, line 93: Is this our first use of exception mappers in connector? The inconsistency troubles me a bit but if we do generally want to move to using exception mappers (like we do in Direct Debit connector) then I guess we have to start somewhere…

src/main/java/uk/gov/pay/connector/model/domain/EmailCollectionMode.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.pay.connector.model.domain;
+
+public enum EmailCollectionMode {
+    MANDATORY, OPTIONAL, OFF;
+
+    public static EmailCollectionMode fromString(String mode) {
+        try {
+            return EmailCollectionMode.valueOf(mode.toUpperCase());
```
alexbishop1, line 8: I don’t think we need to bother supporting lower-case variants of the names — it’s just additional complexity that we may end up supporting forever for no real good reason

src/main/java/uk/gov/pay/connector/model/domain/GatewayAccount.java
```diff
@@ -16,6 +16,7 @@
     public static final String FIELD_VALUE = "value";
     public static final String FIELD_NOTIFY_API_TOKEN = "api_token";
     public static final String FIELD_NOTIFY_TEMPLATE_ID = "template_id";
+    public static final String FIELD_NOTIFY_REFUND_ISSUED_TEMPLATE_ID = "refund_issued_template_id";
```
alexbishop1, line 19: No we have `FIELD_NOTIFY_REFUND_ISSUED_TEMPLATE_ID`, is `FIELD_NOTIFY_TEMPLATE_ID` only applicable to payment confirmation emails? Should we rename it?

src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
```diff
@@ -208,14 +231,20 @@ public void setNotifySettings(Map<String, String> notifySettings){
...
+        this.emailCollectionMode = emailCollectionMode;
+    }
+
     public Map<String, String> getNotifySettings() {
         return notifySettings;
     }
 
-    public Map<String, String> withoutCredentials() {
-        Map<String, String> account = newHashMap();
+    public Map<String, Object> withoutCredentials() {
```
alexbishop1: The generic type parameter being changed to a wider type (`Object`) makes me sad, but the ultimate cause of my sadness is serialising JSON from a map, so let’s leave it for now and try to rationalise how we serialise gateway accounts (at least four different ways, by my count) later.

src/main/java/uk/gov/pay/connector/model/domain/EmailCollectionMode.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.pay.connector.model.domain;
+
+public enum EmailCollectionMode {
+    MANDATORY, OPTIONAL, OFF;
```
alexbishop1, line 4: I wonder about trying to make this names more explicit (at the expense of brevity). Something like `USER_MUST_ENTER_EMAIL`, `USER_MAY_ENTER_EMAIL_BUT_NOT_REQUIRED`, `USER_NOT_ASKED_FOR_EMAIL`.

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
         PatchRequestBuilder.PatchRequest emailPatchRequest;
-
         try {
             emailPatchRequest = aPatchRequestBuilder(emailPatchMap)
                     .withValidOps(Collections.singletonList("replace"))
-                    .withValidPaths(Collections.singletonList(EMAIL_NOTIFICATION_ENABLED))
+                    .withValidPaths(
+                            buildValidPaths(
+                                    Arrays.asList(EmailNotificationType.PAYMENT_CONFIRMED.toString(), EmailNotificationType.REFUND_ISSUED.toString()),
+                                    Arrays.asList(EMAIL_NOTIFICATION_ENABLED, EMAIL_NOTIFICATION_TEMPLATE_BODY)))
```
alexbishop1: It took me longer than I care admit to figure out what we list we end up with here. Since it’s just five static strings, it might be more straightforward to just whack them into a list and have it as a static field (though actually a set would probably be more appropriate).

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
-                .map(gatewayAccount -> emailNotificationsDao.findByAccountId(gatewayAccountId)
-                        .map(emailNotificationEntity -> {
-                            emailNotificationEntity.setEnabled(Boolean.parseBoolean(emailPatchRequest.getValue()));
-                            return Response.ok().build();
-                        }).orElseGet(() -> {
-                            gatewayAccount.setEmailNotification(new EmailNotificationEntity(gatewayAccount));
-                            return Response.ok().build();
-                        }))
+                .map(gatewayAccount -> {
+                    Pair<EmailNotificationType, String> patchInfo = getNotificationTypeFromPath(emailPatchRequest);
```
alexbishop1: Not generally a huge fan of `Pair`: I’d almost prefer to use an inner class so we can name the two fields within it

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+                    EmailNotificationType type = patchInfo.getLeft();
```
alexbishop1: Could you also assign `patchInfo.getRight()` to a local variable here please? I’ve forgotten what it is by the time I get to the `patch(…)` call

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+                    EmailNotificationEntity notificationEntity = Optional.ofNullable(gatewayAccount.getEmailNotifications().get(type))
+                            .orElseGet(() -> {
+                                //PP-4111 we are not going to backfill and add refund notifications for existing gateway accounts, so this is unfortunately needed
+                                EmailNotificationEntity emailNotificationEntity = new EmailNotificationEntity(gatewayAccount, null, false);
+                                gatewayAccount.addNotification(type, emailNotificationEntity);
+                                return  emailNotificationEntity;
+                            });
```
alexbishop1, line 129: Could you yank this block out into a private method called `newDisabledEmailNotificationEntityWithNoTemplate()` or something? It took a while to figure out the intent of this

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+                    patch(notificationEntity, patchInfo.getRight(), emailPatchRequest);
+                    return Response.ok().build();
+                })
                 .orElseGet(() -> notFoundResponse(format("The gateway account id '%s' does not exist", gatewayAccountId)));
     }
+
+    private Pair<EmailNotificationType, String> getNotificationTypeFromPath(PatchRequestBuilder.PatchRequest emailPatchRequest) {
+        String[] paths = emailPatchRequest.getPath().split("/");
```
alexbishop1: It feels like tokenising the path should be something the `PatchRequest` should be able to do

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+        // PP-4111 remove after selfservice is merged
+        if (paths.length < 2) {
```
alexbishop1: Doesn’t the slash at the beginning mean there will usually be three items?

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+            return Pair.of(EmailNotificationType.PAYMENT_CONFIRMED, "enabled");
```
alexbishop1: If this gets out of sync with the validation, we’ll end up assuming anything without a slash means enable payment confirmation emails, which is terrifying, but I’m OK if this is temporary

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+        }
+        return Pair.of(EmailNotificationType.fromString(paths[1]), paths[2]);
+
+    }
+    private void patch(EmailNotificationEntity emailNotificationEntity, String attribute, PatchRequestBuilder.PatchRequest emailPatchRequest) {
```
alexbishop1: It feels a bit wrong that this needs the `attribute` when it also has the `emailPatchRequest` that it’s parsed out of: I think either nothing should be parsed before calling this method, or everything this method needs should be parsed, not half and half

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+        switch (attribute) {
+            case EMAIL_NOTIFICATION_ENABLED:
+                emailNotificationEntity.setEnabled(Boolean.parseBoolean(emailPatchRequest.getValue()));
+                break;
+            case EMAIL_NOTIFICATION_TEMPLATE_BODY_OLD:
+            case EMAIL_NOTIFICATION_TEMPLATE_BODY:
+                emailNotificationEntity.setTemplateBody(emailPatchRequest.getValue());
+                break;
+        }
```
alexbishop1, line 154: It feels like the resource isn’t the right place to be doing this (though I appreciate that this resource is already very heavy). Eventually we probably want to have a generic `JsonPatchService` or something.

Overall, I found the changes in this class very hard to follow. I think it might be clearer if the body first gets parsed into a `PatchRequest`, then the `PatchRequest` is processed to produce an object whose type represents the change (e.g. `EnableEmailNotificationRequest`, `ReplaceEmailTemplateBodyRequest` etc.) and whose fields represent the appropriate values. This would then be passed to a service to apply the actual change.

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +68,98 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
-                                    new EmailNotificationEntity(gatewayAccount, payload.get(EMAIL_NOTIFICATION_TEMPLATE_BODY)));
+                            gatewayAccount.addNotification(EmailNotificationType.PAYMENT_CONFIRMED,
+                                    new EmailNotificationEntity(gatewayAccount, payload.get(EMAIL_NOTIFICATION_TEMPLATE_BODY_OLD)));
                             return Response.ok().build();
                         }))
                 .orElseGet(() -> notFoundResponse(format("The gateway account id '%s' does not exist", gatewayAccountId)));
     }
 
+
+    //PP-4111 backward compatibility, remove once selfservice is merged
```
alexbishop1: Why is there so much new functionality added to this method if the entire method is going to be removed soon?

src/main/java/uk/gov/pay/connector/resources/GatewayAccountRequestValidator.java
```diff
@@ -26,39 +27,46 @@
...
+        if(!VALID_PATHS.contains(operation)) {
+            throw new ValidationException(Collections.singletonList(format("Operation [%s] not supported for path [%s]",
                     FIELD_OPERATION,
                     payload.findValue(FIELD_OPERATION_PATH).asText())));
         }
-        return validateNotifySettingsRequest(payload).map(Errors::from);
-
+        if (operation.equalsIgnoreCase(FIELD_NOTIFY_SETTINGS)) {
+            validateNotifySettingsRequest(payload);
+        }
```
alexbishop1, line 52: So this means we are not doing any validation on requests with a `"path:" "email_collection_mode"`?

src/main/java/uk/gov/pay/connector/resources/GatewayAccountRequestValidator.java
```diff
@@ -26,39 +27,46 @@
...
         }
-        if(op.equals("remove")) {
-            return Optional.empty();
-        }
-        JsonNode valueNode = payload.get(FIELD_VALUE);
-        if(null == valueNode) {
-            return Optional.of(asList(format("Field [%s] is required", FIELD_VALUE)));
+        if (!op.equalsIgnoreCase("remove")) {
+            JsonNode valueNode = payload.get(FIELD_VALUE);
+            if(null == valueNode) {
```
alexbishop1: I know this was already like this but it’s possible this could be a `NullNode`

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -241,6 +242,9 @@ public Response createNewGatewayAccount(JsonNode node, @Context UriInfo uriInfo)
         if (node.has(ANALYTICS_ID_FIELD_NAME)) {
             entity.setAnalyticsId(node.get(ANALYTICS_ID_FIELD_NAME).textValue());
         }
+        
+        entity.addNotification(EmailNotificationType.PAYMENT_CONFIRMED, new EmailNotificationEntity(entity));
```
alexbishop1, line 246: Why wasn’t this needed before?

src/main/java/uk/gov/pay/connector/service/NotificationService.java
```diff
@@ -36,14 +36,16 @@
     private final RefundDao refundDao;
     private final PaymentProviders paymentProviders;
     private final DnsUtils dnsUtils;
+    private UserNotificationService userNotificationService;
```
alexbishop1: This can be `final`

src/main/java/uk/gov/pay/connector/service/NotificationService.java
```diff
@@ -236,6 +238,9 @@ public boolean matchesIpWithDomain(String ipAddress) {
 
             refundEntity.setStatus(newStatus);
             
+            if (newStatus.equals(RefundStatus.REFUNDED)) {
+                userNotificationService.sendRefundIssuedEmail(refundEntity);
```
alexbishop1, line 242: This means that the email will be sent only when the notification saying the payment has been refunded has received. I thought the requirement was to send it when the button was pressed (which may be hours or even days earlier).

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -53,15 +57,28 @@ public UserNotificationService(NotifyClientFactoryProvider notifyClientFactoryPr
...
+    public Future<Optional<String>> sendRefundIssuedEmail(RefundEntity refundEntity) {
+        return sendEmail(EmailNotificationType.REFUND_ISSUED, refundEntity.getChargeEntity(), buildRefundEmailPersonalisationFrom(refundEntity));
+    }
+
+    public Future<Optional<String>> sendPaymentConfirmedEmail(ChargeEntity chargeEntity) {
+        return sendEmail(EmailNotificationType.PAYMENT_CONFIRMED, chargeEntity, buildConfirmationEmailPersonalisationFrom(chargeEntity));
+    }
+
+    private Future<Optional<String>> sendEmail(EmailNotificationType emailNotificationType, ChargeEntity chargeEntity, HashMap<String, String> personalisation) {
+        Boolean isEmailEnabled = Optional.ofNullable(chargeEntity.getGatewayAccount().
```
alexbishop1: This `Boolean` can be a primitive; it doesn’t need to be boxed

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -53,15 +57,28 @@ public UserNotificationService(NotifyClientFactoryProvider notifyClientFactoryPr
...
+                getEmailNotifications()
+                .get(emailNotificationType))
+                .map(EmailNotificationEntity::isEnabled)
+                .orElse(false);
```
alexbishop1, line 70: It took me a little bit of time to figure out what the `Optional.ofNullable(…)` is wrapping. Perhaps the linebreaks could be altered to make it clearer:

```java
boolean isEmailEnabled = Optional.ofNullable(chargeEntity.getGatewayAccount().getEmailNotifications().get(emailNotificationType))
    .map(EmailNotificationEntity::isEnabled)
    .orElse(false);
```

Alternatively, this could be moved out into its own method, which could take the luxury of intermediate local variables etc. to show what’s going on.

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -76,37 +93,59 @@ public UserNotificationService(NotifyClientFactoryProvider notifyClientFactoryPr
...
+                return null;
+        }
+    }
+
+    private Pair<NotificationClient, String> getNotifyClientSettings(EmailNotificationType emailNotificationType, ChargeEntity chargeEntity) {
         Map<String, String> notifySettings = chargeEntity.getGatewayAccount().getNotifySettings();
         NotifyClientFactory notifyClientFactory = notifyClientFactoryProvider.clientFactory();
-        if (notifySettings != null
-                //TODO: replace with a constants when validator available
-                && !isBlank(notifySettings.get("api_token"))
```
alexbishop1, line 84: Are we not worried about this any more?

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
```diff
@@ -76,37 +93,59 @@ public UserNotificationService(NotifyClientFactoryProvider notifyClientFactoryPr
...
         }
-        if (emailNotifyGloballyEnabled && isBlank(emailTemplateId)) {
-            throw new RuntimeException("config property 'emailTemplateId' is missing or not set, which needs to point to the email template on the notify");
+        if (emailNotifyGloballyEnabled && (isBlank(confirmationEmailTemplateId) || isBlank(refundIssuedEmailTemplateId))) {
+            throw new RuntimeException("Check notify config, need to set 'emailTemplateId' (payment confirmation email) and 'refundIssuedEmailTemplateId' properties");
         }
     }
 
-    private HashMap<String, String> buildEmailPersonalisationFromCharge(ChargeEntity charge) {
+    private String getTemplateIdFor(EmailNotificationType type) {
```
alexbishop1: So we have `getCustomTemplate(…)` and `getTemplateIdFor(…)`, both of which switch on the `EmailNotificationType`. Could we somehow encapsulate the two strings into a single object? Or maybe just inline this last one into `getNotifyClientSettings(…)`:

```java
switch (emailNotificationType){
    case REFUND_ISSUED:
        Pair.of(notifyClientFactory.getInstance(), refundIssuedEmailTemplateId);
    case PAYMENT_CONFIRMED:
        return Pair.of(notifyClientFactory.getInstance(), confirmationEmailTemplateId);
}
```

src/test/java/uk/gov/pay/connector/it/resources/EmailNotificationResourceITest.java
```diff
@@ -93,48 +99,169 @@ public void updateEmailNotification_shouldNotUpdateIfAccountIdDoesNotExist() {
...
+        Map<String, Object> emailNotification = app.getDatabaseTestHelper().getEmailForAccountAndType(testAccount.getAccountId(), EmailNotificationType.PAYMENT_CONFIRMED);
         assertThat(emailNotification.get("enabled"), is(false));
     }
 
+    // PP-4111 Old test (using the old patch path)
     @Test
-    public void enableEmailNotification_shouldUpdateSuccessfully() {
-        String accountId = createAGatewayAccountFor("smartpay");
-        app.getDatabaseTestHelper().updateEmailNotification(Long.parseLong(accountId), "lorem ipsum", false);
+    public void patchEnableNotification_shouldUpdateSuccessfully() {
```
alexbishop1, line 118: This is now testing disabling email notifications now, right? (Which I assume is because the default is to have them enabled. We may want to make explicitly enable/disable email notifications when setting up the test account fixture.)

src/test/java/uk/gov/pay/connector/it/resources/EmailNotificationResourceITest.java
```diff
@@ -93,48 +99,169 @@ public void updateEmailNotification_shouldNotUpdateIfAccountIdDoesNotExist() {
...
+
+        givenSetup().accept(JSON)
+                .body(getPatchRequestBody("replace", "/refund_issued/enabled", true))
+                .patch(ACCOUNTS_API_URL + testAccount.getAccountId() + "/email-notification")
+                .then()
+                .statusCode(200);
+
+        Map<String, Object> confirmationEmail = app.getDatabaseTestHelper().getEmailForAccountAndType(testAccount.getAccountId(), EmailNotificationType.PAYMENT_CONFIRMED);
+        Map<String, Object> refundEmail = app.getDatabaseTestHelper().getEmailForAccountAndType(testAccount.getAccountId(), EmailNotificationType.REFUND_ISSUED);
+        assertThat(confirmationEmail.get("enabled"), is(true));
```
alexbishop1, line 147: Why are we also checking the payment confirmed emails are enabled here? Is it just to check that toggling the refund emails doesn’t have any effect on the payment confirmation emails?

src/test/java/uk/gov/pay/connector/it/resources/EmailNotificationResourceITest.java
```diff
@@ -93,48 +99,169 @@ public void updateEmailNotification_shouldNotUpdateIfAccountIdDoesNotExist() {
...
+
+        givenSetup().accept(JSON)
+                .body(getPatchRequestBody("replace", "/refund_issued/template_body", newTemplateBody))
+                .patch(ACCOUNTS_API_URL + testAccount.getAccountId() + "/email-notification")
+                .then()
+                .statusCode(200);
+
+        Map<String, Object> confirmationEmail = app.getDatabaseTestHelper().getEmailForAccountAndType(testAccount.getAccountId(), EmailNotificationType.PAYMENT_CONFIRMED);
+        Map<String, Object> refundEmail = app.getDatabaseTestHelper().getEmailForAccountAndType(testAccount.getAccountId(), EmailNotificationType.REFUND_ISSUED);
+        assertThat(confirmationEmail.get("template_body"), is("Lorem ipsum dolor sit amet, consectetur adipiscing elit."));
```
alexbishop1, line 184: It’s not immediately clear where this _lorem ipsum_ comes from. If we’re checking that it hasn’t changed, we might want to either explicitly set this test up in the fixture or grab the current value in a local variable before the request is made and then compare the new value to that one.

src/test/java/uk/gov/pay/connector/it/resources/EmailNotificationResourceITest.java
```diff
@@ -93,48 +99,169 @@ public void updateEmailNotification_shouldNotUpdateIfAccountIdDoesNotExist() {
...
+                .statusCode(200);
+
+        Map<String, Object> confirmationEmail = app.getDatabaseTestHelper().getEmailForAccountAndType(testAccount.getAccountId(), EmailNotificationType.PAYMENT_CONFIRMED);
+        Map<String, Object> refundEmail = app.getDatabaseTestHelper().getEmailForAccountAndType(testAccount.getAccountId(), EmailNotificationType.REFUND_ISSUED);
+        assertThat(confirmationEmail.get("template_body"), is(newTemplateBody));
+        assertThat(refundEmail.get("template_body"), is("Lorem ipsum dolor sit amet, consectetur adipiscing elit."));
+    }
+
+    @Test
+    public void patchTemplateBodyNotification_shouldUpdateSuccessfullyNotificationIfMissing() {
```
alexbishop1, line 209: Can this be named in such a way that it makes it clear it creates the notification configuration if it‘s “missing”?

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
```diff
@@ -463,7 +469,34 @@ public void shouldReturn400_whenNotifySettingsIsUpdated_withWrongOp() throws Exc
...
+    public void shouldReturn500_whenEmailCollectionModeIsUpdated_withWrongValue() throws Exception{
+        String gatewayAccountId = createAGatewayAccountFor("worldpay");
+        String payload = new ObjectMapper().writeValueAsString(ImmutableMap.of("op", "replace",
+                "path", "email_collection_mode",
+                "value", "nope"));
+        givenSetup()
+                .body(payload)
+                .patch("/v1/api/accounts/" + gatewayAccountId)
+                .then()
+                .statusCode(INTERNAL_SERVER_ERROR.getStatusCode());
```
alexbishop1: Hmm, feels like this should be a 4xx error because there was a problem with the request…

src/test/java/uk/gov/pay/connector/resources/GatewayAccountRequestValidatorTest.java
```diff
@@ -5,11 +5,10 @@
 import com.google.common.collect.ImmutableMap;
 import org.junit.Before;
 import org.junit.Test;
-import uk.gov.pay.connector.util.Errors;
+import uk.gov.pay.connector.exception.ValidationException;
 import uk.gov.pay.connector.validations.RequestValidator;
 
-import java.util.Optional;
-
+import static junit.framework.TestCase.fail;
```
alexbishop1: Please use `org.junit.Assert.fail`

src/test/java/uk/gov/pay/connector/service/GatewayAccountUpdaterTest.java
```diff
@@ -65,4 +64,21 @@ public void shouldUpdateNotifySettingsWhenRemove() throws Exception {
...
+                "path", "email_collection_mode",
+                "value", "off")));
+        GatewayAccountEntity entity = mock(GatewayAccountEntity.class);
+
+        when(gatewayAccountDao.findById(gatewayAccountId)).thenReturn(Optional.of(entity));
+
+        Optional<GatewayAccount> optionalGatewayAcc = updater.doPatch(gatewayAccountId, request);
+
+        assertThat(optionalGatewayAcc.isPresent(), is(true));
+        verify(entity, times(1)).setEmailCollectionMode(EmailCollectionMode.OFF);
```
alexbishop1: Don’t need `times(1)` (it’s the default)

src/test/java/uk/gov/pay/connector/service/GatewayAccountUpdaterTest.java
```diff
@@ -65,4 +64,21 @@ public void shouldUpdateNotifySettingsWhenRemove() throws Exception {
...
+        verify(gatewayAccountDao, times(1)).merge(entity);
```
alexbishop1: Verifying that the above two lines are called in order would be useful here

src/test/java/uk/gov/pay/connector/service/UserNotificationServiceTest.java
```diff
@@ -111,50 +122,117 @@ public void shouldSendEmailIfEmailNotifyIsEnabled() throws Exception {
     }
 
     @Test
-    public void testEmailSendingThrowsExceptionForMissingTemplate() throws Exception {
+    public void shouldSendRefundIssuedEmailIfEmailNotifyIsEnabled() throws Exception {
+        when(mockConfig.getNotifyConfiguration().isEmailNotifyEnabled()).thenReturn(true);
+        when(mockNotifyClientFactoryProvider.clientFactory()).thenReturn(mockNotifyClientFactory);
+        when(mockNotifyClientFactory.getInstance()).thenReturn(mockNotifyClient);
+        when(mockNotifyClient.sendEmail(any(), any(), any(), any())).thenReturn(mockNotificationCreatedResponse);
```
alexbishop1: I know this is just copied from an existing test but we can usually do better than a bunch of `any()`s

alexbishop1 (changes requested):
I’ve left a few comments. Also, I expect some connector internal API docs need to updated to reflect the changes here.

It’s a bit late for this now but in the future, could you please try to split features like this into a series of smaller PRs? For example, this could have been a PR to make it so that email notifications are serialised as part of the gateway account, then a PR to change the model so a gateway account can have a separate refund email associated with it, then a PR to make the changes to the email notification endpoint to allow it to be used to configure the refund email, and finally a PR to actually send the refund email.

Going further, even small changes with ripple effects (especially in tests) like renaming `GatewayAccountRequest` to `PatchRequest` and changing `RequestValidator` to use empty lists rather than empty optionals could have each been their own PRs.

I don’t need to tell you the benefits of small PRs but I think I would have found this much easier to review if it was actually several smaller things. Keeping track of all the different aspects being modified (especially when they’re by necessity interleaved across different files and it’s not always immediately clear which aspect a particular change relates to) can make my head hurt and I find it trickier to do things like ensure test coverage is being maintained when there’s so much going on. I wasn’t actually able to review this as well as I would have liked, whereas with small PRs I would have been able to focus on each individual bit more thoroughly.

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
simad: Yes it is. We should definitely move there and this was a good place to start, I think.

src/main/java/uk/gov/pay/connector/model/domain/EmailCollectionMode.java
simad: It's a pretty safe check we have almost in every enum. I think it solved a few headdaches so that people can write slightly better js code on the frontend side. If you don't feel strongly about it, I'd keep it, otherwise I'll kill it with fire

src/main/java/uk/gov/pay/connector/model/domain/GatewayAccount.java
simad: yes it is, I'll leave the variable "template_id" as it is but I'll rename the field 👍

src/main/java/uk/gov/pay/connector/model/domain/GatewayAccountEntity.java
simad: yes, indeed :) usually when there is Object somewhere I consider it a code smell, but as you said it's the expression of a broader problem (using plain maps to serialise json). We definitely have to refactor the GatewayAccount handling, it's a mess right now

src/main/java/uk/gov/pay/connector/model/domain/EmailCollectionMode.java
simad: I find the terse ones pretty clear - but maybe because I've discussed about them in ACs and at kickoff etc. Do you think we should go for the longer ones?

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: 👍

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: agreed! I didn't add to add more bulk to this. Will do it now

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: 👍 I added a comment but it's always better to do it through the code

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: done with the inner class

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: yes this will be removed as soon as selfservice is in

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: Agreed. I wanted to do a broader refactoring of this whole logic, ie. separate completely the handling of a Patch request (which should be generic). and the specific handling of a patch email notification. It would've taken too long and made the PR even bigger than what it already is, so I decided to make it work and do tiny bits of refactoring

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: Ideally, that would be the case. Resources in connector are super heavy unfortunately. See comment above

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: because it's a lie, it was meant for the `POST` endpoint above 😳

src/main/java/uk/gov/pay/connector/resources/GatewayAccountRequestValidator.java
simad: This whole validation is a bit twisted :( will try to make a small change to introduce email collection validation, but it needs a broader refactoring

simad: Done, the whole `GatewayAccountRequestValidator` needs to be refactored. It is mixing up gateway account specific logic (ie. notify settings and so on) and generic patch validation. We should definitely split these responsibilities.

src/main/java/uk/gov/pay/connector/resources/GatewayAccountRequestValidator.java
simad: this makes me sad

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
simad: because sadly it was done by overriding the persist method of the entity

src/main/java/uk/gov/pay/connector/service/NotificationService.java
simad: 👍

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
simad: oopsie

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
simad: but `map()`

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
simad: I want to refactor everything :(

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
simad: yes :) sorry to point out again that it needs a broader refactoring - I want to introduce an object for notify settings (instead of plain map) and that will contain some logic to discern whether to use custom branding or not etc. etc. will do it next

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
simad: Yes, added that in the validation

src/test/java/uk/gov/pay/connector/resources/GatewayAccountRequestValidatorTest.java
simad: oops, my bad

simad: didn't spot that

src/test/java/uk/gov/pay/connector/service/GatewayAccountUpdaterTest.java
simad: I just copied things \o/

src/test/java/uk/gov/pay/connector/service/GatewayAccountUpdaterTest.java
simad: 👍

src/test/java/uk/gov/pay/connector/service/UserNotificationServiceTest.java
simad: shit, I usually replace that but this time I didn't spot it, sorry

src/main/java/uk/gov/pay/connector/model/domain/EmailCollectionMode.java
alexbishop1: It’s not the hill I’m going to die on but when it comes to accepting computer-generated input (as opposed to user-generated input), I think it’s fine to be fairly restrictive. It’s shouldn’t be too onerous for someone working on another microservice to use the canonical case.

(Also, converting to upper-case has some interesting subtleties: `toUpperCase()` can actually give different results depending on the default locale of the system the JVM is running on, with interesting consequences like `str.toUpperCase().toLowerCase()` not necessarily returning something equal to the original string etc.)

src/main/java/uk/gov/pay/connector/model/domain/EmailCollectionMode.java
alexbishop1: I’m happy for you to decide. I don’t think the terse ones are bad by any means. The appeal to me of more explicit names is that they are more self-describing, particularly when encountered out of context (imagine I just came across `OFF` but didn’t know the name of the enum or the other variants). The disadvantage is that they tend to be longer but this may not be an issue if they’re only used in a few places.

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
alexbishop1: I don’t follow.

By moving it out to its own method, I meant there could be a method like this:

```java
private boolean isEmailEnabled(GatewayAccountEntity gatewayAccount, EmailNotificationType emailNotificationType) {
    EmailNotificationEntity emailNotification = gatewayAccount.getEmailNotifications().get(emailNotificationType);
    return Optional.ofNullable(emailNotification).map(EmailNotificationEntity::isEnabled).orElse(false); // or even just emailNotification != null ? emailNotification.isEnabled() : false;
```

… which could then be used like this:

```java
boolean isEmailEnabled = isEmailEnabled(chargeEntity.getGatewayAccount(), emailNotificationType);
```

src/main/java/uk/gov/pay/connector/model/builder/PatchRequestBuilder.java
```diff
@@ -93,16 +96,20 @@ public String getValue() {
...
 
-        private  boolean isValidPath(String path) {
+        private boolean isValidPath(String path) {
             return Optional.of(validPaths)
                     .map(validOps -> validOps.contains(path))
                     .orElse(true);
         }
+        
+        public List<String> getPathTokens() {
+            return asList(path.substring(1).split("/"));
```
alexbishop1: I assume the `substring(1)` is to skip the leading `/`. Can we be sure this will always be there? I think I’d feel more confident if it was something like `Arrays.stream(path.split("/")).skip(1).collect(Collectors.toList())`, which won’t blow up if it isn’t.

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +75,114 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+                                //PP-4111 we are not going to backfill and add refund notifications for existing gateway accounts, so this is unfortunately needed
+                                return newDisabledEmailNotificationEntityWithNoTemplate(gatewayAccount, type);
+                            });
+                    patch(notificationEntity, patchInfo);
+                    return Response.ok().build();
+                })
                 .orElseGet(() -> notFoundResponse(format("The gateway account id '%s' does not exist", gatewayAccountId)));
     }
+
+    private NotificationPatchInfo getNotificationTypeFromPath(PatchRequestBuilder.PatchRequest emailPatchRequest) {
```
alexbishop1: This method gets more than just the notification type so probably wants to be renamed

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
```diff
@@ -52,56 +75,114 @@ public Response getEmailNotificationText(@PathParam("accountId") Long gatewayAcc
...
+        EmailNotificationEntity emailNotificationEntity = new EmailNotificationEntity(gatewayAccount, null, false);
+        gatewayAccount.addNotification(type, emailNotificationEntity);
+        return  emailNotificationEntity;
+    }
+
+    private class NotificationPatchInfo {
+
+        private EmailNotificationType emailNotificationType;
+        private String path;
+        private String value;
```
alexbishop1: `final`!

src/main/java/uk/gov/pay/connector/resources/GatewayAccountRequestValidator.java
```diff
@@ -26,39 +28,65 @@
...
-        if(!payload.findValue(FIELD_OPERATION_PATH).asText().equals(FIELD_NOTIFY_SETTINGS)) {
-            return Optional.of(Errors.from(format("Operation [%s] not supported for path [%s]",
+        String path = payload.findValue(FIELD_OPERATION_PATH).asText();
+        if(!VALID_PATHS.contains(path)) {
+            throw new ValidationException(Collections.singletonList(format("Operation [%s] not supported for path [%s]",
                     FIELD_OPERATION,
-                    payload.findValue(FIELD_OPERATION_PATH).asText())));
+                    path)));
+        }
+        if (path.equalsIgnoreCase(FIELD_NOTIFY_SETTINGS)) {
```
alexbishop1, line 50: Not convinced by the need to do a case-insensitive comparison here but I’ll leave that up to you

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
alexbishop1: Refactoring is good but as it stands, some validation is being removed. I’m not sure we want to do that.

alexbishop1 (changes requested):
This is getting fairly close. I’ve left a few comments. There’s also still some unaddressed ones from my first review, which you might want to look at again (none of them are show-stoppers: I’ve left new comments on this round for the ones I’m more concerned about).

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: owwwww 👍 sorry

src/main/java/uk/gov/pay/connector/resources/EmailNotificationResource.java
simad: 👍

src/main/java/uk/gov/pay/connector/model/builder/PatchRequestBuilder.java
simad: ideally i'd like to throw an exception if the patch is malformed rather than making sure things don't blow up. I'll see if I can do that without making too many changes

simad: there is the `isValidPath()` function but I can't make it throw on malformed (as in missing /) patches because it wouldn't work with existing code. Will make the change you suggested and flag it for improvement

src/main/java/uk/gov/pay/connector/service/UserNotificationService.java
alexbishop1: Oh, I see, the validation has been just moved further down into the new `NotifyClientSettings` inner class

alexbishop1 (approved):
Most of my most pressing concerns have been addressed so I’m approving.

Hopefully, your forthcoming clean-up and the removal of the legacy endpoints when selfservice has been updated will make this even better.

PR #704, opened 2018-09-19
PP-4139 Add first_digits_card_number column to charges table

simad:
## WHAT
 - Also, indexing on both that column and on the last 4 digits, which seemed to have escaped the previous story

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1513,5 +1513,24 @@
             UPDATE gateway_accounts SET email_collection_mode = 'MANDATORY';
         </sql>
     </changeSet>
-    
+
+    <changeSet id="add first digits card number columns to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="first_digits_card_number" type="VARCHAR(6)">
+                <constraints nullable="true" unique="false"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="add trigram index to charges.last_digits_card_number" runInTransaction="false" author="">
+        <sql>
+            CREATE INDEX CONCURRENTLY idx_last_digits_card_number_idx ON charges(last_digits_card_number);
+        </sql>
+    </changeSet>
+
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1513,5 +1513,24 @@
...
+
+    <changeSet id="add first digits card number columns to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="first_digits_card_number" type="VARCHAR(6)">
+                <constraints nullable="true" unique="false"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="add trigram index to charges.last_digits_card_number" runInTransaction="false" author="">
```
alexbishop1: This isn’t a trigram index, is it?

src/main/resources/migrations.xml
```diff
@@ -1513,5 +1513,24 @@
...
+        <sql>
+            CREATE INDEX CONCURRENTLY idx_last_digits_card_number_idx ON charges(last_digits_card_number);
```
alexbishop1: Not sure we need `idx` at the beginning and the end!

alexbishop1 (changes requested):
SQL (or XML that will become SQL) looks great but I don’t think the names are quite right

src/main/resources/migrations.xml
simad: righto, I had it as trigram but removed it and forgot to rename the  rest xD

alexbishop1 (approved):

src/main/resources/migrations.xml
DanailMinchev: I guess this comment was not addressed, because we use `idx` at the beginning and the end as a convention for other indexes

alexbishop1: We actually seem to be inconsistent. It’s a little redundant but I guess it doesn’t matter in the grand scheme of things.

PR #705, opened 2018-09-19
PP-4139 Retrieve first six digits of credit card from the db

simad:
## WHAT
This PR simply makes sure we can retrieve the first six digits from the db, and serialise it as a field in the frontend charge response,  similarly to how's done for last 4 digits. Next PR(s) will expand this one, adding search and probably refactoring the domain objects in attempt to find a saner way through connector.


src/main/java/uk/gov/pay/connector/model/FirstDigitsCardNumber.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.pay.connector.model;
+
+import java.util.Objects;
+
+public class FirstDigitsCardNumber {
+
+    private final String firstDigitsCardNumber;
+
+    private FirstDigitsCardNumber(String firstDigitsCardNumber) {
+        this.firstDigitsCardNumber = Objects.requireNonNull(firstDigitsCardNumber);
```
alexbishop1, line 10: Can we guarantee if this object is instantiated, it will always be six digits? If so, can we validate it with a regex (like we do for `SortCode` in Direct Debit connector)? It would be really nice to enforce this from outset — if it gets too widely used, our future selves may be too scared to change it later.

src/main/java/uk/gov/pay/connector/model/FirstDigitsCardNumber.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return firstDigitsCardNumber.hashCode();
+    }
+
+    @Override
+    public String toString() {
+        return String.valueOf(firstDigitsCardNumber);
```
alexbishop1: `firstDigitsCardNumber` is already a `String`

src/test/java/uk/gov/pay/connector/it/dao/DatabaseFixtures.java
```diff
@@ -269,7 +279,7 @@ public TestAddress getBillingAddress() {
         }
 
         public TestCardDetails update() {
-            databaseTestHelper.updateChargeCardDetails(chargeId, cardBrand, lastDigitsCardNumber, cardHolderName, expiryDate, billingAddress.getLine1(), billingAddress.getLine2(), billingAddress.getPostcode(), billingAddress.getCity(), billingAddress.getCounty(), billingAddress.getCountry());
+            databaseTestHelper.updateChargeCardDetails(chargeId, cardBrand, lastDigitsCardNumber, firstDigitsCardNumber,cardHolderName, expiryDate, billingAddress.getLine1(), billingAddress.getLine2(), billingAddress.getPostcode(), billingAddress.getCity(), billingAddress.getCounty(), billingAddress.getCountry());
```
alexbishop1: Missing space after comma

alexbishop1 (changes requested):
Looks good! Just a few comments

src/main/java/uk/gov/pay/connector/model/FirstDigitsCardNumber.java
simad: I am doing that as part of the next PR which introduces searching for first digits. I will try to validate both last and first digits in a sane way.

src/main/java/uk/gov/pay/connector/model/FirstDigitsCardNumber.java
simad: rightO

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #709, opened 2018-09-21
PP-4140 Add first 6 digits as search param

simad:
## WHAT
We can now search by first 6 digits of card number.
- Added "type" object to strongly type LastDigits and FirstDigits instead of Strings. These classes should be moved to pay java commons, I've created a ticket for that.
- Added validation of length / numeric string inside the object themselves. I think it simplifies things, because we either have valid objects or nothing. This is open to discussion, of course.


src/main/java/uk/gov/pay/connector/model/LastDigitsCardNumber.java
```diff
@@ -1,16 +1,30 @@
...
-        this.lastDigitsCardNumber = Objects.requireNonNull(lastDigitsCardNumber);
+        this.lastDigitsCardNumber = lastDigitsCardNumber;
     }
 
+    private static boolean isValid(String lastDigitsCardNumber) {
+        return lastDigitsCardNumber != null && lastDigitsCardNumber.length() == 4 && StringUtils.isNumeric(lastDigitsCardNumber);
+    }
+    
     public static LastDigitsCardNumber of(String lastDigitsCardNumber) {
+        if (!(isValid(lastDigitsCardNumber))) {
```
DanailMinchev, line 18: can be `if (!isValid(lastDigitsCardNumber)` without extra `()` but I don't mind

src/test/java/uk/gov/pay/connector/dao/ChargeSearchParamsTest.java
```diff
@@ -68,7 +75,7 @@ public void buildQueryParams_chargeSearch_withAllParameters() throws Exception {
     }
 
     @Test
-    public void buildQueryParamsWithPiiRedaction_chargeSearch_withAllParameters() throws Exception {
```
DanailMinchev, line 71: 👍

src/test/java/uk/gov/pay/connector/model/FirstDigitsCardNumberTest.java
```diff
@@ -0,0 +1,38 @@
...
+        Assert.assertThat(FirstDigitsCardNumber.ofNullable("123456").toString(), is("123456"));
+    }
+
+    @Test
+    public void shouldReturnNullIfStringIsNull() {
+        Assert.assertThat(FirstDigitsCardNumber.ofNullable(null), is(nullValue()));
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void shouldThrowIfNonNumericFirstSixDigitsOfCard() {
```
DanailMinchev, line 24: Great test coverage! I was wondering should we have a test for whitespace like `1  234` which should throw a RuntimeException as well?

DanailMinchev (approved):
LGTM 👍 May be just few minor stuff which can be added later

src/test/java/uk/gov/pay/connector/model/FirstDigitsCardNumberTest.java
simad: I'll add that next :) thanks!

PR #710, opened 2018-09-24
PP-4126 Add corporate surcharge column to charges table

SandorArpa:
- add column corporate_surcharge (BIGINT) to charges table
- default null

with @georgeracu


DanailMinchev (approved):
LGTM 👍

PR #711, opened 2018-09-24
PP-4126 add corporate surcharge to charge entity

SandorArpa:
## WHAT
Add corporate surcharge to `ChargeEntity`

## HOW 
- add corporateSurchargeField to ChargeEntity
- refactor constructor
- update relevant tests
- refactor `GatewayResponse` (remove duplicate code)

with @georgeracu

### DIFF
--- a/src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
@@ -72,6 +72,9 @@ public class ChargeEntity extends AbstractVersionedEntity {
 
     @Column(name = "email")
     private String email;
+    
+    @Column(name = "corporate_surcharge")
+    private Long corporateSurcharge;
 
     @Embedded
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/model/domain/ChargeEntity.java
```diff
@@ -72,6 +72,9 @@
 
     @Column(name = "email")
     private String email;
+    
+    @Column(name = "corporate_surcharge")
+    private Long corporateSurcharge;
```
DanailMinchev, line 77: It's ok to use `Long` for `ChargeEntity` for amounts, because the actual `amount` is `Long` as well.
But we are using `long` primitive for `GatewayAccountEntity` for similar stuff .. Probably we should normalise our approach everywhere someday

DanailMinchev (dismissed):
LGTM 👍

DanailMinchev (approved):
👍

PR #713, opened 2018-09-26
PP-4126 Add corporate surcharge and total amount

georgeracu:
## WHAT

- Added fields `corporate_card` and `card_type` to request object `AuthCardDetails`
- `corporate_card` will map to `Boolean` and `card_type` to `CREDIT, DEBIT, CREDIT_OR_DEBIT`
- Added `CorporateSurchargeCalculator` as a holder for utility methods to calculate `totalAmount` and `corporateSurcharge` for charge authorisation
- Added `corporateSurcharge` and `totalAmount` to `ChargeResponse`
- Added `corporateSurcharge` and `totalAmount` to `FrontendChargeResponse`
- Added `corporateSurcharge` and `totalAmount` to `TransactionResponse`
- Updated `CardAuthoriseService` to add corporate surcharge when authorisation required for
corporate card. This will check the type of the card and will set a surcharge based on the
values set in the gateway account for that given charge
- Corporate surcharge and total amount will be added to the response object only if
corporate surcharge is present (greater than 0), otherwise none will be put in the response
- Extracted test string values to constants
- Renamed method from `getTransactions` to `getCharges` as we don't have a concept
of `Transaction` anymore
- Removed duplicate field declaration
- Added tests for new functionality

with @SandorArpa


src/main/java/uk/gov/pay/connector/model/domain/CardType.java
```diff
@@ -0,0 +1,17 @@
...
+ * The enum type that should be used to map from and to JSON when
+ * dealing with what type (CREDIT, DEBIT, CREDIT_OR_DEBIT) card is
+ * used to make a payment. This is also used to calculate corporate
+ * surcharges, based on other rules.
+ * <p>
+ * This should not be confused with {@link CardTypeEntity.Type}
+ * which is used to map values from the database to frontend
+ * labels. This is used to drive the frontend UI
+ */
+public enum CardType {
```
alexbishop1: I wonder if there’s a way we could name this to make the distinction with the “other” card type clearer. `IdentifiedCardType` or `CardTypeUsedToMakePayment` or something?

src/main/java/uk/gov/pay/connector/model/domain/CardTypeEntity.java
```diff
@@ -3,13 +3,24 @@
...
 
+    /**
+     * Internal entity used to drive the frontend UI based on the
+     * strings stored in the table {@code card_types}. This represents
+     * which card types are supported by the service, not what card types
+     * are used for payment by the paying user
+     * <p>
+     * This should be used only for driving the UI
+     */
     public enum Type {
```
alexbishop1: Again, maybe we could call this something like `SupportedCardType`

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -167,6 +168,12 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
                 .withLink("self", GET, selfUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeId))
                 .withLink("refunds", GET, refundsUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeEntity.getExternalId()));
 
+        if (chargeEntity.getCorporateSurcharge() != null) {
```
alexbishop1: How hard would it be to make `ChargeEntity.getCorporateSurcharge()` return an `Optional<Long>`? As well as allowing us to use a beautiful `ifPresent(…)` here, it would also make it more obvious that it’s not always set.

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+
+import uk.gov.pay.connector.model.domain.AuthCardDetails;
+import uk.gov.pay.connector.model.domain.CardType;
+import uk.gov.pay.connector.model.domain.ChargeEntity;
+import uk.gov.pay.connector.model.domain.GatewayAccountEntity;
+
+/**
+ * Holder for utility methods used to calculate values around corporate surcharge
+ */
+public interface CorporateSurchargeCalculator {
```
alexbishop1: I think I’d prefer it if this was just a regular class that’s injected where it’s needed. Being injected would mean it could be mocked in tests (probably not an issue here because it’s basically just adding two numbers).

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+     * {@link GatewayAccountEntity#corporateDebitCardSurchargeAmount} based on the
+     * {@link AuthCardDetails#cardType}
+     * <p>
+     * The logic here is that if these authorisation details are for a corporate card
+     * and the gateway account has specified a value other than zero or null then this
+     * charge will be updated
+     *
+     * @param authCardDetails An {@link AuthCardDetails} that is sent to be authorised
+     * @param chargeEntity    A {@link ChargeEntity} for which {@param authCardDetails} are sent
+     */
```
alexbishop1: I think it’s pretty clear from the code what this method does; I don’t think we really need the comment

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+    static void setCorporateSurchargeFor(AuthCardDetails authCardDetails, ChargeEntity chargeEntity) {
+        final Boolean corporateCard = authCardDetails.isCorporateCard();
+        if (corporateCard != null && corporateCard) {
```
alexbishop1: We could make the `authCardDetails.isCorporateCard()` method just return `false` if the `Boolean` is `null` so we don’t have to care about that possibility here. I don’t think we have any real reason to distinguish between frontend telling us the card isn’t corporate or not telling us anything at all.

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+            final CardType cardType = authCardDetails.getCardType();
+            if (cardType != null) {
```
alexbishop1: Again, there’s possibly an argument for defaulting this to `CREDIT_OR_DEBIT` rather than returning `null` (though I’m less sold on this than I am with having `corporateCard` default to `false`)

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+                final GatewayAccountEntity gatewayAccount = chargeEntity.getGatewayAccount();
+                if (cardType.equals(CardType.CREDIT) && gatewayAccount.getCorporateCreditCardSurchargeAmount() > 0) {
```
alexbishop1: I was pondering if we do something clever with the enum here. Imagine there was a method on `GatewayAccountEntity` called `getCorporateSurchargeAmount(CardType)`, taking a `CardType` as its argument and returning the appropriate surcharge (which could be a simple `switch` statement). This method would just need to do:

 ```java
if (gatewayAccount.getCorporateSurchargeAmount(cardType) > 0) {
   // …
} // No need for an else if block!
```

File: src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java [hunk 9]
```diff
@@ -0,0 +1,58 @@
...
+                    chargeEntity.setCorporateSurcharge(gatewayAccount.getCorporateCreditCardSurchargeAmount());
+                } else if (cardType.equals(CardType.DEBIT) && gatewayAccount.getCorporateDebitCardSurchargeAmount() > 0) {
+                    chargeEntity.setCorporateSurcharge(gatewayAccount.getCorporateDebitCardSurchargeAmount());
```
alexbishop1: Rather than modifying the entity directly, this method could return the result as either a `long` (with a possible zero value) or an `Optional<Long>`. Then the caller could be responsible for modifying the `ChargeEntity`. That would make this method pure.

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+     * Utility method to calculate total amount for a charge that has a corporate
+     * surcharge.
+     * <p>
+     * Output from this method is the sum between {@link ChargeEntity#corporateSurcharge}
+     * and {@link ChargeEntity#amount} when the corporate surcharge is greater than zero
+     * or amount only, when corporate surcharge is zero or {@link null}
+     *
+     * @param charge The {@link ChargeEntity} for which to get the total amount
+     * @return A {@link Long}
+     */
```
alexbishop1: Again, not sure the comment is really necessary

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,58 @@
...
+    static Long getTotalAmountFor(ChargeEntity charge) {
+        if (charge.getCorporateSurcharge() != null && charge.getCorporateSurcharge() > 0) {
```
alexbishop1: The part after the `&&` can be removed without affecting the result

src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
```diff
@@ -46,14 +47,25 @@
 
 public class ChargingITestBase {
 
+    private static final String ADDRESS_LINE_1 = "The Money Pool";
+    private static final String ADDRESS_LINE_2 = "Moneybags Avenue";
+    private static final String ADDRESS_CITY = "London";
+    private static final String ADDRESS_COUNTY = "Greater London";
+    private static final String ADDRESS_POSTCODE = "DO11 4RS";
+    private static final String ADDRESS_COUNTRY_GB = "GB";
+    private static final String CARD_HOLDER_NAME = "Scrooge McDuck";
```
alexbishop1, line 56: We should use this in all our tests!

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -84,8 +84,10 @@
     private static final String JSON_PROVIDER_KEY = "payment_provider";
     private static final String JSON_LANGUAGE_KEY = "language";
     private static final String JSON_DELAYED_CAPTURE_KEY = "delayed_capture";
+    private static final String JSON_CORPORATE_SURCHARGE_KEY = "corporate_surcharge";
+    private static final String JSON_TOTAL_AMOUNT_KEY = "total_amount";
     private static final String PROVIDER_NAME = "sandbox";
-    private static final long AMOUNT = 6234L;
+    private static final Long AMOUNT = 6234L;
```
alexbishop1, line 90: Unclear if this actually needed to be changed

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -309,7 +313,7 @@ public void shouldReturn404OnGetTransactionsWhenAccountIdIsNonNumeric() {
         getChargeApi
                 .withAccountId("invalidAccountId")
                 .withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON)
-                .getTransactions()
+                .getCharges()
```
alexbishop1: This makes the test call a different endpoint `/v1/api/accounts/{accountId}/charges` rather than `/v1/api/accounts/{accountId}/transactions`. I don’t think we want that.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -458,6 +462,29 @@ public void shouldReturnEmptyCardBrandLabelWhenChargeIsAuthorisedAndBrandUnknown
...
+
+        getChargeApi
+                .withAccountId(accountId)
+                .getCharges()
+                .statusCode(OK.getStatusCode())
+                .contentType(JSON)
+                .body("results[0]." + JSON_CHARGE_KEY, is(externalChargeId))
+                .body("results[0]." + JSON_CORPORATE_SURCHARGE_KEY, is(50))
+                .body("results[0]." + JSON_TOTAL_AMOUNT_KEY, is(AMOUNT.intValue() + 50));
+    }
```
alexbishop1, line 486: Do we need versions of this test that check the other endpoints in the `ChargesApiResource` class (for there is a plethora)?

src/test/java/uk/gov/pay/connector/it/resources/ChargesFrontendResourceITest.java
```diff
@@ -227,7 +227,7 @@ public void shouldReturnAllTransactionsForAGivenGatewayAccount() {
         setupLifeCycleEventsFor(app, chargeId2, statuses);
         setupLifeCycleEventsFor(app, chargeId3, statuses);
 
-        ValidatableResponse response = connectorRestApi.withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON).getTransactions();
+        ValidatableResponse response = connectorRestApi.withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON).getCharges();
```
alexbishop1: I’m not sure why this has been changed

src/test/java/uk/gov/pay/connector/it/resources/TransactionsApiResourceITest.java
```diff
@@ -92,7 +92,7 @@ public void shouldGetExpectedTransactionsByChargeReferenceAndStatusesForBothChar
                 .withQueryParam("refund_states", "submitted")
                 .withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON)
                 .withHeader("features", "REFUNDS_IN_TX_LIST")
-                .getTransactions()
+                .getCharges()
```
alexbishop1: Going by the name, I assume the `TransactionsApiResourceITest` is intended to test the `getTransactions()` endpoint, not the `getCharges()` one

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceITest.java
```diff
@@ -19,7 +24,7 @@ public WorldpayCardResourceITest() {
     }
 
     @Test
-    public void shouldAuthoriseCharge_ForValidAuthorisationDetails() throws Exception {
+    public void shouldAuthoriseChargeWithoutCorporateCard_ForValidAuthorisationDetails() {
```
alexbishop1, line 27: Once frontend is always sending `"corporate_card"` and friends, we can probably remove this test

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -264,6 +265,44 @@ public void shouldFindChargeForChargeIdAndAccountIdWithNextUrlWhenChargeStatusIs
         shouldFindChargeForChargeIdAndAccountIdWithNextUrlWhenChargeStatusIs(ENTERING_CARD_DETAILS);
     }
 
+    @Test
+    public void shouldFindChargeForChargeIdAndAccountIdWithNextUrlWhenChargeStatusIsAuthorisationReady_andNoCorporateSurcharge() throws Exception {
+        shouldFindChargeForChargeIdAndAccountIdWithNextUrlWhenChargeStatusIs(AUTHORISATION_READY);
+    }
+
+    @Test
+    public void shouldFindChargeForChargeIdWhenChargeStatusIsAuthorisationReady_andCorporateSurcharge() {
```
alexbishop1: I think `AuthorisationReady` can be removed from the name of this test: it’s not essential to what’s being tested

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -292,15 +331,19 @@ private void shouldFindChargeForChargeIdAndAccountIdWithNextUrlWhenChargeStatusI
         assertThat(tokenEntity.getChargeEntity().getId(), is(newCharge.getId()));
         assertThat(tokenEntity.getToken(), is(notNullValue()));
 
-        ChargeResponseBuilder expectedChargeResponse = chargeResponseBuilderOf(chargeEntity.get());
-        expectedChargeResponse.withLink("self", GET, new URI(SERVICE_HOST + "/v1/api/accounts/1/charges/" + externalId));
-        expectedChargeResponse.withLink("refunds", GET, new URI(SERVICE_HOST + "/v1/api/accounts/1/charges/" + externalId + "/refunds"));
-        expectedChargeResponse.withLink("next_url", GET, new URI("http://payments.com/secure/" + tokenEntity.getToken()));
-        expectedChargeResponse.withLink("next_url_post", POST, new URI("http://payments.com/secure"), "application/x-www-form-urlencoded", new HashMap<String, Object>() {{
+        ChargeResponseBuilder chargeResponseWithoutCorporateSurcharge = chargeResponseBuilderOf(chargeEntity.get());
```
alexbishop1: Not sure adding `WithoutCorporateSurcharge` is adding much. We can probably assume the no surcharge is the standard case and doesn’t need to be explicitly called out.

src/test/java/uk/gov/pay/connector/service/ChargeServiceTest.java
```diff
@@ -292,15 +331,19 @@ private void shouldFindChargeForChargeIdAndAccountIdWithNextUrlWhenChargeStatusI
...
+        chargeResponseWithoutCorporateSurcharge.withLink("next_url_post", POST, new URI("http://payments.com/secure"), "application/x-www-form-urlencoded", new HashMap<String, Object>() {{
             put("chargeTokenId", tokenEntity.getToken());
         }});
 
-        assertThat(chargeResponseForAccount.get(), is(expectedChargeResponse.build()));
+        assertThat(chargeResponseForAccount.isPresent(), is(true));
+        final ChargeResponse chargeResponse = chargeResponseForAccount.get();
+        assertThat(chargeResponse.getCorporateSurcharge(), is(nullValue()));
+        assertThat(chargeResponse.getTotalAmount(), is(nullValue()));
+        assertThat(chargeResponse, is(chargeResponseWithoutCorporateSurcharge.build()));
```
alexbishop1: I think this can just be changed back a single `assertThat(chargeResponseForAccount.get(), is(expectedChargeResponse.build())`: the `equals(…)` method should take care of ensuring the two new fields are `null`.

src/test/java/uk/gov/pay/connector/util/RestAssuredClient.java
```diff
@@ -129,14 +129,14 @@ public ValidatableResponse postChargeCancellation() {
                 .then();
     }
 
-    public ValidatableResponse getTransactions() {
+    public ValidatableResponse getCharges() {
```
alexbishop1: OK, this is solving some of my earlier confusion!

Please put renames like this with large ripple effects in a separate PR.

In any case, I’m not sure we actually want to rename this. If I recall correctly, these endpoints can return both charges _and_ refunds intermixed. Generally we use the term “transactions” when we mean both.

src/test/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculatorTest.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    @Test
+    public void shouldCalculateTotalAmountForCorporateSurchargeIsZero() {
+        ChargeEntity chargeEntity = ChargeEntityFixture.aValidChargeEntity().withCorporateSurcharge(0L).build();
+        final Long actualAmount = CorporateSurchargeCalculator.getTotalAmountFor(chargeEntity);
+        final Long expectedAmount = chargeEntity.getAmount();
+        assertThat(actualAmount, is(expectedAmount));
+    }
+}
```
alexbishop1, line 170: Nice comprehensive set of tests

alexbishop1 (changes requested):
Implementation looks good in general. Left a few suggestions.

I’m not convinced this will correctly include the corporate surcharge in payments returned from the `v1/api/accounts/ACCOUNT_ID/transactions` endpoint. I remember getting this working being quite involved. See https://github.com/alphagov/pay-connector/pull/673/files (where we did a similar thing for `"delayed_capture"` and look at the `Transaction` class and friends.

Not sure about renaming the methods in `RestAssured` client for reasons explained in the comment attached to that class.

Finally (too late for this now), this could have been done as two separate PRs: one to persist the surcharge to the charge entity and another to include it in the responses.

src/main/java/uk/gov/pay/connector/model/domain/CardType.java
SandorArpa: CardType -> PayersCardType
CardTypeEntity.Type -> CardEntity.AcceptedType

src/main/java/uk/gov/pay/connector/service/ChargeService.java
SandorArpa: I thought that using an Object instead of a primitive would make that obvious but I'll change it to Optional.

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
SandorArpa: agreed

src/main/java/uk/gov/pay/connector/model/domain/PayersCardType.java
```diff
@@ -0,0 +1,17 @@
...
+ * The enum type that should be used to map from and to JSON when
+ * dealing with what type (CREDIT, DEBIT, CREDIT_OR_DEBIT) card is
+ * used to make a payment. This is also used to calculate corporate
+ * surcharges, based on other rules.
+ * <p>
+ * This should not be confused with {@link CardTypeEntity.SupportedType}
+ * which is used to map values from the database to frontend
+ * labels. This is used to drive the frontend UI
+ */
+public enum PayersCardType {
```
alexbishop1, line 13: Love this name

src/main/java/uk/gov/pay/connector/resources/GatewayAccountResource.java
```diff
@@ -375,7 +375,7 @@ public Response updateGatewayAccountAcceptedCardTypes(@PathParam("accountId") Lo
                 .collect(Collectors.toList());
 
         if (cardTypeIds.size() != cardTypeEntities.size()) {
-            return badRequestResponse(format("CardType(s) referenced by id(s) '%s' not found", String.join(",", extractNotFoundCardTypeIds(cardTypeIds, cardTypeEntities))));
+            return badRequestResponse(format("AcceptedCardType(s) referenced by id(s) '%s' not found", String.join(",", extractNotFoundCardTypeIds(cardTypeIds, cardTypeEntities))));
```
alexbishop1: I wouldn’t camel-case `AcceptedCardType` if it doesn’t refer to an actual type

src/main/java/uk/gov/pay/connector/service/CardAuthoriseService.java
```diff
@@ -68,6 +69,11 @@ public ChargeEntity preOperation(String chargeId, AuthCardDetails authCardDetail
             } else {
                 preOperation(chargeEntity, OperationType.AUTHORISATION, getLegalStates(), AUTHORISATION_READY);
 
+                Optional<Long> corporateSurcharge = getCorporateSurchargeFor(authCardDetails, chargeEntity);
+                if (corporateSurcharge.isPresent()) {
+                    chargeEntity.setCorporateSurcharge(corporateSurcharge.get());
+                }
```
alexbishop1:
```java
getCorporateSurchargeFor(authCardDetails, chargeEntity).ifPresent(chargeEntity::setCorporateSurcharge);
```

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -167,6 +168,12 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
                 .withLink("self", GET, selfUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeId))
                 .withLink("refunds", GET, refundsUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeEntity.getExternalId()));
 
+        if (chargeEntity.getCorporateSurcharge().isPresent()) {
+            reponseBuilder
```
alexbishop1: Could you add another `s` to `reponseBuilder` please?

src/main/java/uk/gov/pay/connector/service/ChargeService.java
```diff
@@ -167,6 +168,12 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
...
+                    .withCorporateSurcharge(chargeEntity.getCorporateSurcharge().get())
+                    .withTotalAmount(CorporateSurchargeCalculator.getTotalAmountFor(chargeEntity));
+        }
```
alexbishop1:
```java
chargeEntity.getCorporateSurcharge().ifPresent(corporateSurcharge ->
    reponseBuilder.withCorporateSurcharge(corporateSurcharge)
        .withTotalAmount(CorporateSurchargeCalculator.getTotalAmountFor(chargeEntity)));
```

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,33 @@
...
+import uk.gov.pay.connector.model.domain.ChargeEntity;
+import uk.gov.pay.connector.model.domain.PayersCardType;
+
+import java.util.Optional;
+
+/**
+ * Holder for utility methods used to calculate values around corporate surcharge
+ */
+public class CorporateSurchargeCalculator {
+    public static Optional<Long> getCorporateSurchargeFor(AuthCardDetails authCardDetails, ChargeEntity chargeEntity) {
```
alexbishop1, line 13: I think this can just be a regular non-static method i.e. we should create an instance of `CorporateSurchargeCalculator` and inject it where needed using Guice

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -0,0 +1,33 @@
...
+        }
+        return Optional.empty();
+    }
+
+
+    public static Long getTotalAmountFor(ChargeEntity charge) {
+        if (charge.getCorporateSurcharge().isPresent()) {
+            return charge.getCorporateSurcharge().get() + charge.getAmount();
+        }
+        return charge.getAmount();
```
alexbishop1:
```java
return charge.getCorporateSurcharge()
    .map(surcharge -> surcharge + charge.getAmount())
    .orElseGet(charge::getAmount);
```

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
alexbishop1: You have explained that those endpoint changes will be coming in another PR

src/test/java/uk/gov/pay/connector/it/resources/TransactionsApiV2ResourceITest.java
```diff
@@ -96,7 +96,7 @@ public void shouldGetExpectedTransactionsByChargeReferenceAndStatusesForBothChar
                 .withQueryParam("payment_states", "success,timedout")
                 .withQueryParam("refund_states", "submitted")
                 .withHeader(HttpHeaders.ACCEPT, APPLICATION_JSON)
-                .getTransactionsV2()
+                .getChargesV2()
```
alexbishop1: I thought we decided not to do this method name change for now

src/test/java/uk/gov/pay/connector/service/CardAuthoriseServiceTest.java
```diff
@@ -144,7 +146,77 @@ public void doAuthorise_shouldRespondAuthorisationSuccess() {
         verify(mockedChargeEventDao).persistChargeEventOf(charge, Optional.empty());
         assertThat(charge.get3dsDetails(), is(nullValue()));
         assertThat(charge.getCardDetails(), is(notNullValue()));
+        assertThat(charge.getCorporateSurcharge().isPresent(), is(false));
+    }
+
+    @Test
+    public void doAuthoriseWithCorporateCard_shouldRespondAuthorisationSuccess() {
```
alexbishop1: Can this be named in such a way to make it clear that even though it’s a corporate card, we don’t expect a surcharge to be applied because the account doesn’t have a surcharge?

src/test/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculatorTest.java
```diff
@@ -0,0 +1,170 @@
...
+        assertThat(chargeEntity.getCorporateSurcharge().isPresent(), is(false));
+        assertThat(getTotalAmountFor(chargeEntity), is(chargeEntity.getAmount()));
+
+        Optional<Long> optionalSurcharge = getCorporateSurchargeFor(creditOrDebit, chargeEntity);
+
+        assertThat(optionalSurcharge.isPresent(), is(false));
+    }
+
+    @Test
+    public void shouldNotGetCorporateSurchargeForNull() {
```
alexbishop1: Could you specify what exactly it is that’s null here please?

alexbishop1 (changes requested):
Thanks for making the requested changes.

It’s clearer now that the method renames have been taken out (but I think one might have been missed).

Thanks also for explaining that this only does the charges endpoint and that the others will be done in a separate PR.

Left a few comments, all of them fairly minor.

alexbishop1 (dismissed):
I still think the `CorporateSurchargeCalculator` would be better as an injected instance rather than a class a static methods but that’s tiny on the grand scheme of things, so I’m approving

alexbishop1 (approved):

PR #714, opened 2018-09-27
PP-4113 Fix refund amount in response

cobainc0:
## WHAT
- fixes mistake where we were returning the Charge amount instead of the Refund amount

Author: @cobainc0


SandorArpa (approved):
LOKTM

SandorArpa (approved):

PR #721, opened 2018-10-03
:fire:break: Add allow_web_payments to gateway_accounts table

oswaldquek:
Add this flag to indicate whether Apple Pay can be used.

@oswaldquek @willpearson


whpearson (approved):
Looks good.

PR #724, opened 2018-10-04
PP-4248 Upgrade Mockito from 2.0.54-beta to 2.22.0

alexbishop1:
Upgrade Mockito from 2.0.54-beta to 2.22.0 for Java 11 compatibility.

Newer Mockito versions fail if they detect unnecessary stubbing, so also remove all the stubbing lines that Mockito indicates are unused.

When stubbing, newer Mockito releases seem to not consider `null` to match `any(Long.class)`. Some tests in `ChargeRefundServiceTest` stub returning a `RefundEntity` from the (mocked) `RefundDao` by an ID that turns out to actually be `null`, due to a `RefundEntity` only getting an ID when the `RefundDao` persists it (the mock `RefundDao` does not set an ID). Fix this by using a Mockito `Answer` to set the ID on the `RefundEntity` (and in one test, on its spy double) to an arbitrary (but non-`null`) value that can then be used in stubbing. This is not great and `ChargeRefundServiceTest` could do with some love but we want to minimise the number of changes we make at once.

(There is a version 2.22.1 but it’s not in Maven Central.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -280,7 +280,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
-            <version>2.0.54-beta</version>
+            <version>2.22.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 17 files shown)


DanailMinchev (approved):
🎉

PR #726, opened 2018-10-08
PP-4126 Add corporate surcharge and total amount

georgeracu:
## WHAT
PP-4126 Add `coporate_surcharge` and `total_amount` to transactions search response

## HOW 
- Added `corporate_surcharge` and `total_amount` to Transaction and
    response from ChargesFrontendResource and TransactionSearchStrategy
    - Updated CorporateSurchargeCalculator to calculate total amount
    for for transaction
    - Updated tests and added tests for these fields
    - Added helper methods on db test helper to find charge by external id
    and updated db fixture to add a charge with `corporate_surcharge` and
    `total_amount`


src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
```diff
@@ -9,7 +9,6 @@
 import org.jooq.SelectOrderByStep;
 import org.jooq.SelectSeekStep1;
 import org.jooq.impl.DSL;
-import uk.gov.pay.connector.model.LastDigitsCardNumber;
```
danworth, line 12: 👍 appears this isn't used and hence removed.

src/main/java/uk/gov/pay/connector/dao/TransactionDao.java
```diff
@@ -122,16 +122,16 @@ private SelectOrderByStep buildQueryFor(Long gatewayAccountId, QueryType queryTy
...
                     field("c.last_digits_card_number").eq(lastDigitsCardNumberConverter.convertToDatabaseColumn(params.getLastDigitsCardNumber())));
         }
 
-        if (params.getFirstDigitsCardNumber() != null && isNotBlank(params.getFirstDigitsCardNumber().toString())) { 
+        if (params.getFirstDigitsCardNumber() != null && isNotBlank(params.getFirstDigitsCardNumber().toString())) {
             queryFilters = queryFilters.and(
                     field("c.first_digits_card_number").eq(params.getFirstDigitsCardNumber().toString()));
         }
-        
+
```
danworth, line 134: 👍 tidying up :)

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -66,10 +68,7 @@ public ChargesFrontendResource(ChargeDao chargeDao, ChargeService chargeService,
     @JsonView(GatewayAccountEntity.Views.FrontendView.class)
     public Response getCharge(@PathParam("chargeId") String chargeId, @Context UriInfo uriInfo) {
 
-        Optional<ChargeEntity> maybeCharge = chargeDao.findByExternalId(chargeId);
-        logger.debug("charge from DB: " + maybeCharge);
-
-        return maybeCharge
+        return chargeDao.findByExternalId(chargeId)
```
danworth, line 71: Tidier 👍

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
```diff
@@ -183,7 +182,17 @@ private ChargeResponse buildChargeResponse(UriInfo uriInfo, ChargeEntity charge)
                 .withDelayedCapture(charge.isDelayedCapture())
                 .withLink("self", GET, locationUriFor("/v1/frontend/charges/{chargeId}", uriInfo, chargeId))
                 .withLink("cardAuth", POST, locationUriFor("/v1/frontend/charges/{chargeId}/cards", uriInfo, chargeId))
-                .withLink("cardCapture", POST, locationUriFor("/v1/frontend/charges/{chargeId}/capture", uriInfo, chargeId)).build();
+                .withLink("cardCapture", POST, locationUriFor("/v1/frontend/charges/{chargeId}/capture", uriInfo, chargeId));
+        charge.getCorporateSurcharge().ifPresent(surcharge -> {
+            if (surcharge > 0) {
```
danworth, line 187: It seems having this logic within the `.withCorporateSurcharge` method to also set the `totalAmount` would be safer in that it will ensure the `totalAmount` is always set when `corporateSurcharge` is and avoid the duplication in the TransactionSearchStrategy a little further down. You could then just pass in the value if present and put the `if (surcharge > 0)` logic in there too (since this may change and it'll be easy to find in the one file responsible for creating the response).

src/main/java/uk/gov/pay/connector/service/search/TransactionSearchStrategy.java
```diff
@@ -83,6 +84,15 @@ public ChargeResponse buildResponse(UriInfo uriInfo, Transaction transaction) {
         if (refundSummary != null) {
             transactionResponseBuilder.withRefunds(refundSummary);
         }
+
+        transaction.getCorporateSurcharge().ifPresent(surcharge -> {
```
danworth, line 88: Same as above, to avoid this duplication the logic and call to set the totalAmount if surcharge is > 0 could be within the `withCorporateSurchage`.

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
```diff
@@ -29,4 +34,21 @@ public static Long getTotalAmountFor(ChargeEntity charge) {
...
+     * surcharge.
+     * <p>
+     * Output from this method is the sum between {@link Transaction#corporateSurcharge}
+     * and {@link Transaction#amount} when the corporate surcharge exists
+     * or amount only, when corporate surcharge is zero or {@link null}
+     *
+     * @param transaction The {@link Transaction} for which to get the total amount
+     * @return A {@link Long}
+     */
+    public static Long getTotalAmountFor(Transaction transaction) {
```
danworth, line 49: Certainly one way to do it. Did you consider having `getTotalAmount()` as a method on the `Transaction` class, it seems a simple calculation to add its own `amount` to its own `corporateSurcharge` and return it? This avoids needing a utility method in a separate class. It also makes it clear to anyone in the future who looks at `Transaction` class that there is an `amount` and a `totalAmount` which might prompt them to consider which they really need should they not know about this utility method.

danworth (changes requested):
Hello, it looks good. I've made two suggestions, one regarding the response building to ensure that `totalAmount` is set whenever `corporateSurcharge` is (which seems appropriate) and avoid duplication of the logic. Another around wether `getTotalAmount` could be a method on the `Transaction` class (happy to hear your thoughts if you considered and dismissed such an idea).

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
georgeracu: I went on and off on having this method on the `Transaction` object and I believe is better to be in a utility method because:
* `Transaction` is holding data, should not be aware of the business logic used to calculate different amounts (same goes with `total`, a bit in our way at the moment, but we need to live with it)
* Having it in a separate utility method makes it easy to change the business logic and to add other values that get calculated on the fly. Might seem premature, but I think we should have our data objects pure, to keep only values from our database.

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
SandorArpa: As discussed, we came to the decision to use a utility class for the purpose of calculating total amount. `ResponseBuilder` classes are the few in `connector` that are pure in a sense that they do not contain business logic. `.withCorporateSurcharge` and `.withTotalAmount` are virtual values (meaning they are not stored and their only representation is in the outgoing Json response). That's why we decided to 'hide' them in plain sight and use the `CalculateCorporateSurcharge` utility class to the benefit of future generations to come. Adding logic (eg setting the `totalAmount` when setting the `corporateSurcharge`) would make (in our opinion) connector even more spaghetti bound. I agree it does not look too pretty but on the other hand it makes the whole process very visible.

src/main/java/uk/gov/pay/connector/service/search/TransactionSearchStrategy.java
SandorArpa: Either way the duplication would be present as both endpoints need the logic of setting `corporateSurcharge` and `totalAmount`.

src/main/java/uk/gov/pay/connector/util/charge/CorporateSurchargeCalculator.java
danworth: Ok, thanks for the explanation. I'm happy to go with your judgement 👍

src/main/java/uk/gov/pay/connector/resources/ChargesFrontendResource.java
danworth: Ok, thanks for the explanation. I'm happy to go with your judgement 👍

danworth (approved):

PR #727, opened 2018-10-08
[WIP] PP-4183 Multiple threads - CardCaptureProcess

kbottla:
### `Background`

The capture process (captures charges asynchronously in the background) in connector is a single threaded job that takes a number of charges (based on batch size) and captures them one by one. Potential issues with this approach:

1. If there are multiple connector nodes running, could potentially end up with overlap of charges being captured.
2. The average response time (live) to capture charges is ~100ms. So a batch size of 50 will take ~5 seconds to process a batch. So a single connector capture job scheduled every 10 seconds can only capture ~17k charges/day.
3. Increasing number of connector nodes doesn't necessarilty multiply the number of charges the can be captured as charges can overlap with other connector jobs

So given a high volume of transactions (~1 million/day), capture processes can take weeks to clear charges for a single day and may end up with expiring authorisations.

### `Solution`

Proposed solution is to introduce a queue (within connector node) and run multi-threaded capture process so that charges can be captured parallely and quickly. For a batch size of 50 charges and 10 capture threads we can clear the batch with in ~500ms. It can also likely to mitigate the issue of overlapping charges between connector nodes.

### `Changes`

- CaptureProcessScheduler now schedules multiple threads to capture charges
- Scheduler first loads shared queue (with charges awaiting capture)
- Each thread reads (removes charge from head of the queue - if queue is not empty) a charge from queue to further process it.
        - `ArrayBlockingQueue` is used as a queue which seemed most suitable to use as it is a fixed size queue that can hold charges equivalent to batch size, thread safe and size() method is constant-time operation. Although it is blocking queue, operations used (poll(), isEmpty()) are not blocking. So a thread can exit from operation without being blocked.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/CaptureProcessConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/CaptureProcessConfig.java
@@ -7,6 +7,7 @@ public class CaptureProcessConfig extends Configuration {
     private long schedulerInitialDelayInSeconds;
     private long schedulerRandomIntervalMinimumInSeconds;
     private long schedulerRandomIntervalMaximumInSeconds;
+    private int schedulerThreads;
 
     private int batchSize;
     private Duration retryFailuresEvery;
@@ -38,5 +39,9 @@ public class CaptureProcessConfig extends Configuration {
 
     public int getMaximumRetries() {
         return maximumRetries;
+    }
+
+    public int getSchedulerThreads() {
+        return schedulerThreads;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
@@ -9,6 +9,7 @@ import com.google.inject.persist.jpa.JpaPersistModule;
 import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.setup.Environment;
 import uk.gov.pay.connector.applepay.ApplePayDecrypter;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.validator.RequestValidator;
 import uk.gov.pay.connector.gateway.GatewayClientFactory;
 import uk.gov.pay.connector.gateway.PaymentProviders;
@@ -24,6 +25,8 @@ import uk.gov.pay.connector.util.HashUtil;
 import uk.gov.pay.connector.util.XrayUtils;
 
 import java.util.Properties;
+import java.util.Queue;
+import java.util.concurrent.ArrayBlockingQueue;
 
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.STRIPE;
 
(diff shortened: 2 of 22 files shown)


heathd:
I'd like some more explanation in the commit message about the rationale for the choice of queue type, and the logic around handling concurrent readers/writers and ensuring correct semantics.

oswaldquek:
@heathd @kbottla sorry to butt in here, but i think any solution here should warrant further discussion. i know i was quite up for using a real queue (e.g. SQS). though i'm not up for that anymore i recall @heathd saying we should consider using a real queue. and other people might possibly have more ideas for solutions.

granted, i haven't gone through this PR in detail but i'm not really sure about using the java.util.Queue.... if the connector app dies everything in the queue is gone.

i actually have a firebreak idea in mind to get rid of CardCaptureProcess. very briefly, the idea would be to perform the capture asynchronously as soon as the user clicks 'Confirm' on the payment. this is easily achieved using

```
CompletableFuture
.supplyAsync(() -> getPaymentProvider().capture(...))
.handle(
  //handle errors here, e.g. setting charge status to CAPTURE_RETRY
);
```
Using the `CompletableFuture` avoids the need for instantiating any sort of ExecutorService. very nice in my opinion. it's a bit like javascript!

in the event where the payment gateway is down, we'd have a cloudwatch (or some other thing) event to hit a connector endpoint which will retry the failed charges. this way only one instance of connector processes these and we have no conflict issues.

though i still think using a real queue would be great, if we can avoid adding infrastructure, we should do that.

heathd:
@oswaldquek I haven't had time to fully digest your comments, but just to say the main reason for the background process is to ensure that capture is retried. More background here https://payments-platform.atlassian.net/browse/PP-1633

kbottla:
@oswaldquek Thanks for the comment.
I agree a real queue would serve the purpose better. The current solution is enhancement to what is currently in place and is quick win for short term solution.

> granted, i haven't gone through this PR in detail but i'm not really sure about using the java.util.Queue.... if the connector app dies everything in the queue is gone.

Queue is used only to hold charges to capture. It doesn't hold real state. When connector dies we will end up with two scenarios
 1. Orderly shutdown : Scheduler thread is hooked to dropwizard lifecycle. So it will wait for all threads to complete before shutdown.
2. Forced shutdown/killed : As the state is maintained in database another connector can pickup charges not processed

> get rid of CardCaptureProcess.

As @heathd mentioned we need to retry captures (currently every 1hour for 48 hours). As per my current understanding this happens frequently for EPDQ. Background asnyc process sounds good though and combining with CardCaptureProcess to retry captures could lead to smaller queue size

oswaldquek:
@heathd yes, i've accounted for retrying the capture:
```
in the event where the payment gateway is down, we'd have a cloudwatch (or some other thing) event to hit a connector endpoint which will retry the failed charges. this way only one instance of connector processes these and we have no conflict issues.```

kbottla:
Closing this as there is new PR https://github.com/alphagov/pay-connector/pull/947

PR #732, opened 2018-10-10
PP-4111 Add service name to refund issued template

simad:
## WHAT
We need to show that in the refund issued email copy


brid (approved):
LGTM

PR #734, opened 2018-10-11
PP-4111 Use charge date, not refund date

simad:
## WHAT
 - The body of the email refers to the original payment date, not the
   refund date


brid (approved):
LGTM!

PR #736, opened 2018-10-11
PP-4113 Amend connector path for Search Refunds

cobainc0:
Author: @cobainc0

## WHAT
Amends connector path for Search Refunds Resource so that it matches the path specified in Public API.


simad (dismissed):
Shipit

simad (dismissed):
now you can shipit🥉

simad (approved):
nice work 👍

PR #737, opened 2018-10-11
PP-4276: Use direct debit connector's DropwizardJUnitRunner

oswaldquek:
This commit largely copies's dd-connector's DropwizardJUnitRunner which spins
up the dropwizard app and postgres container only once for all tests that use
it. This commit reduces the build time from ~20minutes to <10minutes.
    
In the ChargingITestBase we have had to make the
[worldpay|smartpay|epdq]MockClients static as some tests fail when all the
tests are being run. Probably some weird connection pooling issue.
    
@oswaldquek, @georgeracu


src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
```diff
@@ -63,48 +65,49 @@
...
+
     protected static final String RETURN_URL = "http://service.url/success-page/";
     protected static final String EMAIL = randomAlphabetic(242) + "@example.com";
-    protected RestAssuredClient connectorRestApiClient;
     protected static final long AMOUNT = 6234L;
-    protected WorldpayMockClient worldpay;
-    protected SmartpayMockClient smartpay;
-    protected EpdqMockClient epdq;
-    protected final String accountId;
+    protected static final WorldpayMockClient worldpayMockClient = new WorldpayMockClient();
```
IreneLau-GDS, line 72: Just wondering why we are making these static. If the plan is to make tests run in parallel in the future, this can cause unexpected test failure because the mocking behaviour might overwrite each other.

src/test/java/uk/gov/pay/connector/it/resources/CardAuthoriseResourceITest.java
```diff
@@ -262,14 +268,19 @@ public void shouldReturnErrorWithoutChangingChargeState_IfOriginalStateIsNotEnte
...
         timeoutInSeconds.setInt(conf, 0);
 
         String chargeId = createNewChargeWithNoTransactionId(ENTERING_CARD_DETAILS);
         String message = format("Authorisation for charge already in progress, %s", chargeId);
-        authoriseAndVerifyFor(chargeId, validCardDetails, message, 202);
+        try {
+            authoriseAndVerifyFor(chargeId, validCardDetails, message, 202);
+        } finally {
+            timeoutInSeconds.setInt(conf, 1);
+            timeoutInSeconds.setAccessible(false);
```
IreneLau-GDS: I understand from the PR description that the changes were largely originated from DD Connector so you probably didn't want to change much of the behaviour. However it scares me a bit when looking at `DropwizardTestApplications` where we keep multiple DW app test instances in memory and we are manipulating the config on the fly. 

I totally understand the changes in the try-no catch-just finally block is to fix the config not being reset but wouldn't it be safer if we actually have a separate config called `test-it-config-with-executor-service-override.yaml` for this test? (I know it probably sounds OTT to you but I think it would be safer.)

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -344,6 +351,8 @@ public void shouldGetSuccessResponseForExpiryChargeTaskFor3dsRequiredPayments()
 
     @Test
     public void shouldGetSuccessForExpiryChargeTask_withStatus_awaitingCaptureRequest() {
+        
+//        databaseTestHelper.deleteAllChargesOnAccount();
```
IreneLau-GDS: Might as well get rid of commented out code?

src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceITest.java
```diff
@@ -2,53 +2,68 @@
...
+    public void setup() {
+        databaseTestHelper = testContext.getDatabaseTestHelper();
 
         defaultTestAccount = DatabaseFixtures
-                .withDatabaseTestHelper(app.getDatabaseTestHelper())
+                .withDatabaseTestHelper(databaseTestHelper)
                 .aTestAccount()
                 .insert();
 
         this.defaultTestCharge = DatabaseFixtures
```
IreneLau-GDS: You don't need the `this.`. Just a minor thing. Sorry for being pedantic.

src/test/java/uk/gov/pay/connector/junit/DropwizardTestApplications.java
```diff
@@ -0,0 +1,59 @@
...
+        getRuntime().addShutdownHook(new Thread(() -> {
+            for (DropwizardTestSupport applicationsRunning : apps.values()) {
+                applicationsRunning.after();
+            }
+        }));
+    }
+
+    static Optional<DropwizardTestSupport> createIfNotRunning(Class<? extends Application> appClass, String configClasspathLocation, ConfigOverride... configOverrides) {
+        Pair<Class<? extends Application>, String> key = Pair.of(appClass, configClasspathLocation);
+        if (!apps.containsKey(key)) {
```
IreneLau-GDS, line 41: See my comments way above... the one about an individual test manipulating the config values using reflection... 😱

IreneLau-GDS (commented):
I think, the truncate with cascade works if we *always* have PK-FK defined across all our tables. Otherwise, strange test failure might happen because either it's trying to insert test data with the same values into the uniquely indexed columns or in the opposite case, the select will end up returning multiple things that we are not expecting (and if we are only picking the first one then good luck to us!)

src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
oswaldquek: Please see the PR description: `In the ChargingITestBase we have had to make the
[worldpay|smartpay|epdq]MockClients static as some tests fail when all the
tests are being run. Probably some weird connection pooling issue.`
In other words, I'm not sure why things fail if we don't make these static 😕

oswaldquek:
@IreneLau-GDS With regards to the truncate cascade comment, originally I didn't want to do this. In an ideal world I would re-create the database (using a template) for each test. That's what we do in dd-connector. However we can't re-create the database like we do in dd-connector as we use JPA (dd-connector uses JDBI) and JPA doesn't like the fact we re-create the database. To get around this we truncate the database here. I appreciate it's not ideal but applying the principle of YAGNI, the tests work fine now and if we encounter a problem in the future because of this we should deal with it then.

src/test/java/uk/gov/pay/connector/it/resources/CardAuthoriseResourceITest.java
oswaldquek: Yeah maybe better to extract this test out into its own test class with a different config yaml.

src/test/java/uk/gov/pay/connector/junit/DropwizardTestApplications.java
oswaldquek: sorry i don't see any reflection being used here?

IreneLau-GDS: > See my comments way above... the one about an individual test manipulating the config values using reflection...

is my comment in the one all the way up in `CardAuthoriseResourceITest` where we use reflection to pull the field out then set the field value.

oswaldquek: I think so. We have moved that test into its own test class now so we don't use reflection. See CardAuthorizeDelayedGatewayResponseTest.java: https://github.com/alphagov/pay-connector/pull/737/files#diff-637d50ea2bfd0c1fa9b226ffe168580e

IreneLau-GDS: Yay!

IreneLau-GDS (approved):
I think it looks fine (the java side of thing), with my limited knowledge of connector.

PR #738, opened 2018-10-11
PP-4280 Reorganised Gateway (payment gateway) related classes

kbottla:
## WHAT
First in the line of PRs to restructure classes (through out the connector) into clearly identifiable top level functions/domains.

This change includes
- Organised payment gateway (worldpay, epdq, smartpay, sandbox) specific classes into `gateway` package and sub packages
- Mostly includes changes to package name & corresponding imports
- Qualifier changes for few static variables

Doesn't include

@mrlumbu


src/test/java/uk/gov/pay/connector/service/UnknownStatusTest.java
```diff
@@ -1,13 +1,15 @@
...
+import uk.gov.pay.connector.paymentserviceprovider.model.status.InterpretedStatus;
+import uk.gov.pay.connector.paymentserviceprovider.model.status.UnknownStatus;
 
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
 
 public class UnknownStatusTest {
 
-    private final UnknownStatus UnknownStatus = new UnknownStatus();
+    private final uk.gov.pay.connector.paymentserviceprovider.model.status.UnknownStatus UnknownStatus = new UnknownStatus();
```
alexbishop1: 🤨

alexbishop1 (changes requested):
Not going to look through all of this but I trust you (thanks so much for not combining this with other changes!).

I did spot one place where it seems fully-qualified class name is being used unnecessarily.

Also — and this is just a thought, not a request — `gateway` might be shorter than `paymentserviceprovider` in package names (generally, “gateway” refers specifically to card payment processors, whereas “payment service provider” is more broad).

Now might be a good opportunity to remove all the wildcard imports from the modified classes (in IntelliJ, select all of the files in the _Version Control_ view and then open the _Code_ menu and choose _Optimize Imports_) since they’re changing anyway.

mrlumbu:
@alexbishop1  `gateway` is not a bad idea. I would like to standardise on one term (duh). Or maybe `paymentgateway`? Then use this everywhere - `PaymentProvider` makes no sense. We could just use `gateway` everywhere, but it feels a bit too generic, even in the bounded context of connector

alexbishop1:
I’m sure we could bikeshed on the names forever. The way connector alternates between “gateway” and “(payment) provider” is frustrating, though it seems to mainly use “gateway” (though often in the context of “gateway account”, which is a concept that we’ve allowed to expand to represent more than purely an account with a payment gateway).

I don’t think `gateway` is too generic for connector: it’s a fairly well-defined term within the domain of card payments.

The abbreviation `psp` is another option and has the appeal of being short.

mrlumbu:
Ok, let's go with `gateway`, if you're happy with that @kbottla?

kbottla:
That sounds good to me :)

mrlumbu:
In a future PR we'll use gateway everywhere that we currently use PaymentProvider and friends

kbottla:
Thanks @alexbishop1 . Changes made as follows:
- Renamed package to `gateway`
- Wild imports have been removed for organised packages. Thanks for the tip
- Removed fully classified name in `UnknownStatusTest`

alexbishop1 (approved):

PR #739, opened 2018-10-12
PP-4277 🔪connector tests

simad:
## WHAT
- Using `org.mockito.junit.MockitoJUnitRunner` because the old runner was deprecated
- Removing unused GET /transactions endpoint
- Using new mockito `ArgumentMatchers` instead of deprecated matchers
- Removing unused fixtures and generally unused code


mrlumbu:
re c3907d5 it is actually quite useful to have these in the codebase, esp. when doing a new integration. It means you can have a quick local way of figuring out it what you are doing works. They are not intended to be run as part of CI pipeline.

mrlumbu (approved):

PR #741, opened 2018-10-12
PP-4196 Fix PayersCardType enum

DanailMinchev:
## WHAT

- Fix PayersCardType enum to have `CREDIT_OR_DEBIT` instead of `DEBIT_OR_CREDIT` value as frontend is converting `CD` type to `CREDIT_OR_DEBIT`

with @stephencdaly


alexbishop1 (dismissed):
True story

alexbishop1 (approved):

PR #743, opened 2018-10-15
PP-4276 Code cleanup

georgeracu:
## WHAT

- All classes that inherit from `ChargingITestBase` call `super.setup()`
when they have a method annotated with `@Before`
- The methods annotated with `@Before` have been renamed to `setup` so it
will override the method in the parent class. In some cases, the parent
method annotated with `@Before` was not called and some of the clients
were not instantiated, throwing NPE. JUnit uses Reflection for methods
annotated with `@Before` and when overriding it in a child class, the
parent functionality was not called. In some cases it was working because
it was called `setUp` or something else (mind the upper case `U`)
- Optimised imports and code style
- Removed locally instantiated client objects, as there's an `connectorRestApi`
object that gets created every time when a test is run. As we only use this
client objects to call the resources, it seems perfectly fine to use the
object that gets created anyway
- This change might help in the future in adding all the generic setup
required in the base class and test behaviour in the child classes that
also have the tests

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelResourceITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelResourceITest.java
@@ -1,11 +1,9 @@
 package uk.gov.pay.connector.it.resources;
 
 import org.apache.commons.lang.StringUtils;
-import org.junit.Before;
 import org.junit.Test;
 import uk.gov.pay.connector.it.base.ChargingITestBase;
 import uk.gov.pay.connector.model.domain.ChargeStatus;
-import uk.gov.pay.connector.util.RestAssuredClient;
 
 import java.time.ZonedDateTime;
 import java.util.List;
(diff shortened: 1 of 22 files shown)


kbottla (dismissed):

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelResourceITest.java
```diff
@@ -192,7 +202,7 @@ public void respondWith404__IfChargeIdDoNotBelongToAccount() {
         String chargeId = createNewInPastChargeWithStatus(CREATED);
         String expectedMessage = format("Charge with id [%s] not found.", chargeId);
 
-        restApiCall
+        connectorRestApi
```
mrlumbu: since you have to rebase anyway, could you rename this to connectorRestApiClient

mrlumbu (changes requested):
Nice one, this looks good. Could definitely do with a bit more thinking here about what really belongs on the base class and what should just be a utility, but this is a good first step. I have made a small request to change a variable name

mrlumbu (approved):

PR #745, opened 2018-10-15
PP-4127 Refactor AuthorisationGatewayRequest to include corporate surcharge

SandorArpa:
## WHAT
- Refactor `AuthorisationGatewayRequest.getAmount()` to calculate total amount when corporate surcharge is present
- Add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
@@ -4,6 +4,7 @@ import uk.gov.pay.connector.gateway.GatewayOperation;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.model.domain.AuthCardDetails;
 import uk.gov.pay.connector.model.domain.ChargeEntity;
+import uk.gov.pay.connector.util.charge.CorporateSurchargeCalculator;
 
 import java.util.Optional;
 
@@ -25,7 +26,7 @@ public class AuthorisationGatewayRequest implements GatewayRequest {
     }
 
     public String getAmount() {
-        return String.valueOf(charge.getAmount());
+        return String.valueOf(CorporateSurchargeCalculator.getTotalAmountFor(charge));
     }
 
     public String getDescription() {
--- a/src/test/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequestTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequestTest.java
@@ -1,0 +1,31 @@
+package uk.gov.pay.connector.gateway.model.request;
+
+
+import org.junit.Test;
+import uk.gov.pay.connector.model.domain.ChargeEntity;
+import uk.gov.pay.connector.model.domain.ChargeEntityFixture;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class AuthorisationGatewayRequestTest {
+    
+    @Test
+    public void shouldReturnTotalAmount_whenThereIsACorporateSurcharge() {
+        ChargeEntity chargeEntity = ChargeEntityFixture.aValidChargeEntity()
+                .withAmount(2000L)
+                .withCorporateSurcharge(250L)
+                .build();
+        AuthorisationGatewayRequest gatewayRequest = new AuthorisationGatewayRequest(chargeEntity, null);
+        assertThat(gatewayRequest.getAmount(), is("2250"));
+    }
+    
+    @Test
+    public void shouldReturnBaseAmount_whenThereIsNoCorporateSurcharge() {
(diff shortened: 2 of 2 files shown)


DanailMinchev (dismissed):
LGTM 👍

DanailMinchev (dismissed):

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
```diff
@@ -48,4 +48,9 @@ public GatewayOperation getRequestType() {
     public static AuthorisationGatewayRequest valueOf(ChargeEntity charge, AuthCardDetails authCardDetails) {
         return new AuthorisationGatewayRequest(charge, authCardDetails);
     }
+    
+    private String calculateAmount() {
+        return charge.getCorporateSurcharge().isPresent() ? String.valueOf(charge.getAmount() + charge.getCorporateSurcharge().get()) :
+                String.valueOf(charge.getAmount());
+    }
```
mrlumbu: can we not encapsulate this logic in the charge object itself?

mrlumbu: sorry, bit of a drive by review there

SandorArpa: We're trying not to add business logic into data wrappers. Imho business logic should not reside in a data object. We had a discussion about it when we started working on corporate surcharge.

mrlumbu (changes requested):
Unless there is good reason not to, I think this logic belongs in the ChargeEntity

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
mrlumbu: Ok. I think this needs to be broadcast more widely, because there are several competing philosophies at play in connector. Let's have a huddle about this. I think if you don't want to encapsulate in the Entity (which already has a lot of logic in), then I think you should use the `CorporateSurchargeCalculator`

SandorArpa: 👍

mrlumbu (approved):

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
alexbishop1: Soooo… I have some views on this.

In general, I’m of the opinion that data objects shouldn’t have much in the way of business logic in them. That generally belongs in service objects. A domain object shouldn’t know how to perform business operations on itself.

However, I’m not opposed to having some logic in a domain object, if that logic relates to something _intrinsic_ to the object itself (rather than the business concerns). For example, I’m in favour of domain objects maintaining their invariants: if a domain object class `Person` has an `int` field called `age`, it’s desirable that it should be impossible for `age` to be set to a negative value, which would involve some logic in the class. That’s OK because the age not being negative is intrinsic to the concept of a person: it’s not something that can change depending on whatever business logic we want to apply to it this week.

I’m inclined to think that a total in a payment is in a similar vein. A total is always the amount plus any extras, almost by definition. Hence it’s intrinsic to the concept of a payment. The business logic may change but the meaning of the total won’t.

That said, there’s a complication here introduced by how we model domain objects in connector. The object in question, `ChargeEntity`, is used for three different purposes: a database entity object, an actual domain object, and a JSON response object. Both database entity objects and JSON response objects don’t want to have business logic in them: they should just be Java representations of what’s in the database and what’s getting sent in the response JSON respectively.

So in an ideal world, I’d like `ChargeEntity` to be three different objects: a database entity object, an actual domain object, and a JSON response object. In the normal flow of an operation, we might read what’s in the database to into a database entity object, convert it into an actual domain object for processing, then convert it into a JSON object for the response. The database entity object and the JSON response object would be dumb POJOs with just setters and getters. They database entity object would know nothing about the total value of the payment (because the database doesn’t store that), while the JSON response object would just be told what the total is when it’s constructed (because JSON is just a data representation and shouldn’t have anything fancy in it). The actual domain object, which is what most of the business logic services would interact with, could know what the total is because it’s intrinsic to the concept of a payment. This is the object that would be used when we’re figuring out what to send to a payment gateway. (I have plenty more thoughts on this if anyone’s interested.)

Unfortunately, we don’t live in an ideal world. So for this particular case, I think it’s debatable what the right answer is and I can see the argument from both sides. Looking at `ChargeEntity` as it is now, I see it does already have some logic in it, so I’d lean towards saying it’s OK for it to be able to report its own total, particularly if this is something that we need to use in several places (we don’t want the total = amount + surcharge logic to be repeated all over the shop).

As an aside, if the code is going to stay as it is, it could be expressed as:

```java
return String.valueOf(charge.getCorporateSurcharge()
    .map(surcharge -> charge.getAmount() + surcharge)
    .orElseGet(() -> charge.getAmount()));
```

PR #748, opened 2018-10-16
PP-4280 New validation pattern and validation testing

mrlumbu:
This commit illustrates how to use Dropwizard's built in validation
functionality to validate incoming requests. It is quite powerful, and
can do most of (probably all) the things we require of it.

In this commit I have taken one example endpoint
`/v1/frontend/charges/{chargeId}/status`, and refactored
validation for the PUT method.

I do not necessarily think that the validation logic is right
(in particular the constraint on the value of `new_status`),
but I have not changed it at this stage, just the implementation.

In order to change as little as possible, I have overridden the
dropwizard default exception mapper (see https://www.dropwizard.io/0.9.2/docs/manual/validation.html)
, so as  to be able to add my own `ConstraintViolationExceptionMapper`,
which allows me to map ConstraintViolationExceptions to 400 status
code. The returned message is slightly different from before, but
is not any worse IMHO.

In tandem with this change to the validation, I have changed how it
is tested. Instead of spinning up a whole application and db, I have
used the ResourceTestRule that provides a testable resource which can be
interacted with as if it is an http server, whilst allowing mocking
of it's dependencies. This allows tests to complete in ~400 ms as opposed to
~16 seconds before.

To do this I again had to do some fiddling with the exception mapper config,
(see https://www.dropwizard.io/1.3.5/docs/manual/testing.html)
which is the reason for the brillaintly named
`ResourceTestRuleWithCustomExceptionMappersBuilder`, which encapsulates the needed
changes.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java
@@ -1,0 +1,82 @@
+package uk.gov.pay.connector.charge.resource;
+
+import io.dropwizard.testing.junit.ResourceTestRule;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+import uk.gov.pay.connector.charge.dao.ChargeDao;
+import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.dao.CardTypeDao;
+import uk.gov.pay.connector.rules.ResourceTestRuleWithCustomExceptionMappersBuilder;
+
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.Response;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.hasItem;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ChargesFrontendResourceTest {
+    @Mock
+    private static ChargeService chargeService;
+    @Mock
+    private static  ChargeDao chargeDao;
+    @Mock
+    private static  CardTypeDao cardTypeDao;
+    
+    @ClassRule
+    public static ResourceTestRule resources = ResourceTestRuleWithCustomExceptionMappersBuilder.getBuilder()
+            .addResource(new ChargesFrontendResource(chargeDao, chargeService, cardTypeDao))
+            .build();
+
+    @Test
+    public void shouldReturn400_whenPutToChargeStatus_emptyPayload() {
+        Response response = resources.client()
+                .target("/v1/frontend/charges/irrelevant_charge_id/status")
+                .request()
+                .put(Entity.json(""));
+
+        assertThat(response.getStatus(), is(400));
+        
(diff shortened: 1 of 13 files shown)


oswaldquek:
👍 for using ResourceTestRule!!
Generally LGTM; no issues from me.

src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.List;
+import java.util.Map;
+
+import static org.hamcrest.CoreMatchers.hasItem;
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ChargesFrontendResourceTest {
```
georgeracu, line 26: 👍 This way of testing resources is faster than hitting the database each time. We need to make sure that we cover via unit tests the DAO objects used by this.

src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java
```diff
@@ -0,0 +1,82 @@
...
+                .request()
+                .put(Entity.json(Collections.singletonMap("new_status", "")));
+
+        assertEquals( 400, response.getStatus());
+        
+        List<String> listOfErrors = (List) response.readEntity(Map.class).get("message");
+        assertThat(listOfErrors.size(), is(2));
+        assertThat(listOfErrors, hasItem("invalid new status"));
+        assertThat(listOfErrors, hasItem("may not be empty"));
+
```
georgeracu, line 66: Maybe a few assertions that the DAO methods don't get called. Just like a reassurance that the validator works and stops execution.

src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java
```diff
@@ -0,0 +1,82 @@
...
+        Response response = resources.client()
+                .target("/v1/frontend/charges/irrelevant_charge_id/status")
+                .request()
+                .put(Entity.json(Collections.singletonMap("new_status", "invalid_status")));
+
+        assertEquals(400, response.getStatus());
+
+        List<String> listOfErrors = (List) response.readEntity(Map.class).get("message");
+        assertThat(listOfErrors.size(), is(1));
+        assertThat(listOfErrors, hasItem("invalid new status"));
```
georgeracu, line 80: Same as above on asserting the DAO methods not getting called

src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java
mrlumbu: at this point I was thinking a quick win would be to move all the validation tests to this style (and leave the ones that hit the db as they are). We could definitely go further, but that might be something to do more gradually

georgeracu (dismissed):
This looks like a good example to follow in testing our resources without having to hit the db each time. 👍

src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java
mrlumbu: given there are no methods defined on the mock, the whole thing would fail if it got anywhere near the dao I think, no?

georgeracu: That also works :)

georgeracu (approved):
LGTM :)

PR #750, opened 2018-10-16
PP-4280 Reorganise Notifications & Webhooks

kbottla:
## WHAT

- Organised user notification related classes into `usernotification` package and sub packages
- Organised payment gateway notifications to `webhooks` package and sub packages
- Introduced `common` package with few classes
- Mostly includes changes to package names & corresponding imports

Doesn't include


mrlumbu:
think we need to have a think about naming here. Historically people have referred to 'webhooks' as notifications in connector, and that's kind of leaked around the platform. So we might be better off reserving notification for that, and coming up with something else for email/sms stuff (communication?)

kbottla:
@mrlumbu @georgeracu Updated name to 'usernotification' which will avoid conflict with the notification terminology used for webhooks

mrlumbu:
Ok, let's go with this for now. If we are going for webhooks as package name we should do some further refactoring to make connector consistent, but that can be in a separate PR

mrlumbu (approved):

PR #751, opened 2018-10-17
PP-4128 Add contract test for corporate surcharge

SandorArpa:
## WHAT
- add pact test for a charge with corporate surcharge

with @danworth


georgeracu (approved):
LGTM

PR #752, opened 2018-10-17
PP-4306 Refactor response, rename corporate_surcharge

SandorArpa:
## WHAT
Rename corporate_surcharge to avoid discrepancy between connector and publicapi 
## HOW 
- rename corporate_surcharge -> corporate_card_surcharge
- rename relevant methods and fields


DanailMinchev (approved):
LGTM 👍

PR #754, opened 2018-10-18
PP-4130 Add a new Pact state for transactions list

DanailMinchev:
## WHAT

- We need to add one transaction to the list, because `pay-selfservice` Pact for Cypress has changed.

with @stephencdaly


SandorArpa (approved):
LOKTM

PR #755, opened 2018-10-18
PP-4130 Remove deprecated Pact state for transactions list

DanailMinchev:
## WHAT

- Remove deprecated Pact state for transactions list, because selfservice does not use it anymore.

This PR to be merged after https://github.com/alphagov/pay-selfservice/pull/928 is merged and deployed


SandorArpa (approved):
LOKTM

PR #756, opened 2018-10-19
PP-4280 Reorganise Cardtypes, Tokens & Reports

kbottla:
## WHAT
- Organised card type related classes into `cardtype` package and sub packages
- Organised token related classes into `token` package and sub packages
- Organised performance report & transaction summary classes into `report` package and sub packages

Doesn't include


mrlumbu:
Nice. The only change I would make is that I think the dashboard package imports UI concepts into connector. I would prefer it if the dashboard stuff went under `report`

kbottla:
Thanks @mrlumbu, agree and updated.

mrlumbu (approved):

src/main/java/uk/gov/pay/connector/cardtype/model/domain/UuidAbstractEntity.java
```diff
@@ -1,8 +1,13 @@
-package uk.gov.pay.connector.model.domain;
+package uk.gov.pay.connector.cardtype.model.domain;
```
alexbishop1, line 1: Not sure this class wanted to move

kbottla: Although the name suggests abstract, only place this is used is in CardTypeEntity and thought these can be close. But could be in `package uk.gov.pay.connector.model.domain` so something else can use it in future. Will move it in the next change

PR #762, opened 2018-10-22
PP-4293 Refactor validation of search params 

simad:
## WHAT
- Refactored existing search params, adding type objects `FromDate`, `ToDate`, `PageNumber` and `DisplaySize`. These are "implicitly" validated, so that it's been possible to remove the (quite ugly) validation we had before.
- The semantics of validation have changed a bit: we were collecting all errors and returning them in a bad request response, now we "fail fast" and let the dropwizard exception mapper return a bad request as soon as any validation fails

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/resource/SearchRefundsResource.java
+++ b/src/main/java/uk/gov/pay/connector/refund/resource/SearchRefundsResource.java
@@ -4,6 +4,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.charge.dao.SearchParams;
+import uk.gov.pay.connector.charge.model.DisplaySize;
+import uk.gov.pay.connector.charge.model.FromDate;
+import uk.gov.pay.connector.charge.model.PageNumber;
+import uk.gov.pay.connector.charge.model.ToDate;
 import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
 import uk.gov.pay.connector.refund.service.SearchRefundsService;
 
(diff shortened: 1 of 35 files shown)


mrlumbu:
A general comment: I am not convinced this is the best way to do this kind of validation. On the other hand, it is undoubtedly a vast improvement on the current implementation. So I am a bit torn as to what to do. I will make a suggestion here of a different approach that would also work, and maybe we can discuss it in person.
On the whole I am quite keen for us to use more Dropwizard powers, as I think one of the reasons we've ended up with so many divergent patterns is that we haven't leant enough on the many out of the box things that Dropwizard brings. In this case, there are annotations provided for validating incoming query parameters. So, for example for display size and page number you could do:
```
@QueryParam(PAGE) @Positive Optional<Long> pageNumber,
@QueryParam(DISPLAY_SIZE) @Positive Optional<Long> displaySize
```
to ensure that `page` and `displaySize` are positive, if they exist. 
This way we have to write barely any code at all to validate the inputs. 
It would also automatically mean we had an optional for a value that may of may not exist, which seems semantically nice. Finally, it would mean the domain objects would not have to throw ValidationExceptions, but could throw IllegalArgumentExceptions or something - i.e your resource does the up front validation and throws validation exceptions (which could be more extensive - e.g. checking that from date < to date etc.), whilst the domain objects are just making sure they are being created correctly (this distinction may not seem very clear in this case, admittedly). 
I am not saying this is the right way of doing it, but thought I'd point out that this is possible, and maybe preferable.

On a different note, I would encourage you (in a separate PR probably) to move any tests of validation at the resource level (e.g. `shouldShowValidationErrorForInvalidQueryParam`) to use a `ResourceTestRule` a la https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceTest.java.

simad:
Closing this PR for now to agree on a shared approach for validation with the team :)

src/main/java/uk/gov/pay/connector/refund/resource/SearchRefundsResource.java
```diff
@@ -53,10 +55,10 @@ public Response getRefundsByAccountId(@PathParam(ACCOUNT_ID) Long accountId,
                                           @Context UriInfo uriInfo) {
         logger.info("Getting all refunds for account id {}", accountId);
         SearchParams refundSearchParams = new SearchParams()
-                .withFromDate(parseDate(fromDate))
-                .withToDate(parseDate(toDate))
-                .withDisplaySize(displaySize != null ? displaySize : configuration.getTransactionsPaginationConfig().getDisplayPageSize())
-                .withPage(pageNumber != null ? pageNumber : 1);
+                .withFromDate(FromDate.ofNullable(fromDate))
```
cobainc0, line 58: `SearchParams` should be constructed with the GatewayAccountId. 
`withGatewayAccountId(gatewayAccountId)` to prevent a previous bug from being reintroduced.

PR #764, opened 2018-10-23
PP-4319 - Fix package of performance reporter dao

simad:
## WHAT
 - We have moved some classes around and the performance reporter dao
   could not find the class to unmarshall the db data to anymore
 - Added tests so that if this happens again, something will break

### DIFF
--- a/src/test/java/uk/gov/pay/connector/report/dao/PerformanceReportDaoITest.java
+++ b/src/test/java/uk/gov/pay/connector/report/dao/PerformanceReportDaoITest.java
@@ -21,10 +21,11 @@ import static org.junit.Assert.assertThat;
 public class PerformanceReportDaoITest extends DaoITestBase {
 
     private PerformanceReportDao performanceReportDao;
-    DatabaseFixtures.TestAccount testAccountFixture;
+    private DatabaseFixtures.TestAccount testAccountFixture;
 
     @Before
-    public void setUp()  {
+    public void setup() throws Exception {
+        super.setup();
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/report/dao/PerformanceReportDaoITest.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.hamcrest.number.BigDecimalCloseTo.closeTo;
+import static org.junit.Assert.assertThat;
+
+public class PerformanceReportDaoITest extends DaoITestBase {
+
+    private PerformanceReportDao performanceReportDao;
+    DatabaseFixtures.TestAccount testAccountFixture;
+
+    @Before
+    public void setUp()  {
```
mrlumbu: slight pedantry alert, but this method should probably override the `setup` method in the base class, then call that as super. This ensures the correct ordering of before operations. I think there is a general pattern of not having done this, but it's something George and Oz discovered can cause problems

simad: ah right. I have copied that from somewhere, so I'll fix both

mrlumbu: 👍 💟

mrlumbu (dismissed):

mrlumbu (approved):

PR #765, opened 2018-10-23
PP-4131 Available refund amount with surcharge

DanailMinchev:
## WHAT

Implemented available refund amount with surcharge

## HOW

- Move charge calculator classes to charge.util package
- Include corporate surcharge in total amount available for refund
- Added tests for refunds with corporate card surcharge

with @stephencdaly

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/util/RefundCalculator.java
+++ b/src/main/java/uk/gov/pay/connector/charge/util/RefundCalculator.java
@@ -1,6 +1,7 @@
-package uk.gov.pay.connector.util.charge;
+package uk.gov.pay.connector.charge.util;
 
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.common.service.charge.CorporateCardSurchargeCalculator;
 import uk.gov.pay.connector.refund.model.domain.RefundEntity;
 import uk.gov.pay.connector.refund.model.domain.RefundStatus;
 
--- a/src/main/java/uk/gov/pay/connector/charge/util/RefundCalculator.java
+++ b/src/main/java/uk/gov/pay/connector/charge/util/RefundCalculator.java
@@ -1,7 +1,6 @@
 package uk.gov.pay.connector.charge.util;
 
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
-import uk.gov.pay.connector.common.service.charge.CorporateCardSurchargeCalculator;
 import uk.gov.pay.connector.refund.model.domain.RefundEntity;
 import uk.gov.pay.connector.refund.model.domain.RefundStatus;
 
--- a/src/main/java/uk/gov/pay/connector/charge/util/RefundCalculator.java
+++ b/src/main/java/uk/gov/pay/connector/charge/util/RefundCalculator.java
@@ -13,23 +13,10 @@ public class RefundCalculator {
         // prevent Java for adding a public constructor
     }
 
-    /**
-     * Get the total amount available to be refunded. Will be the total amount, including any additional charges
-     * minus the amount that has already been refunded.
-     *
-     * @param chargeEntity The {@link ChargeEntity} to get the amount available to be refunded for
-     * @return long - amount available to be refunded
-     */
-    public static long getTotalAmountToBeRefunded(ChargeEntity chargeEntity) {
+    public static long getTotalAmountAvailableToBeRefunded(ChargeEntity chargeEntity) {
         return CorporateCardSurchargeCalculator.getTotalAmountFor(chargeEntity) - getRefundedAmount(chargeEntity);
     }
 
-    /**
-     * Get the total amount that has been refunded for the given {@link ChargeEntity}
-     *
-     * @param chargeEntity The {@link ChargeEntity} to get the amount refunded for
-     * @return long - refunded amount
-     */
     public static long getRefundedAmount(ChargeEntity chargeEntity) {
         return chargeEntity.getRefunds().stream()
                 .filter(p -> p.hasStatus(RefundStatus.CREATED, RefundStatus.REFUND_SUBMITTED, RefundStatus.REFUNDED))
--- a/src/test/java/uk/gov/pay/connector/charge/util/RefundCalculatorTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/util/RefundCalculatorTest.java
@@ -1,5 +1,83 @@
-import static org.junit.Assert.*;
+package uk.gov.pay.connector.charge.util;
+
+import org.junit.Test;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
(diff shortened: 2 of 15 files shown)


mrlumbu:
I think this PR raises the question of what pattern we want to use to add business logic to the connector domain. This is not something we have ever really discussed as a team, and certainly not come to a consensus about. I would prefer us to move towards a pattern of domain objects that encapsulate their logic and only expose the minimum information outwards. This is at odds with the pattern you are following here, that is using static utility functions to manipulate 'dumb entities'. This does not seem like a very idiomatic way of doing things in Java world (<- Dom's opinion alert)
I would like us as a team to make a decision about the way we want to do this stuff, so I propose that we have a huddle about this to try and get a team consensus. I appreciate that may block this story which you need to get out, so if need be you can merge this as is (review permitting), but we should definitely have a team conversation about this in the near future. I'm happy to organise that, or happy for you to organise it, whichever suits you.

DanailMinchev:
@mrlumbu 
Yes, as discussed we started with the logic encapsulated on the entity level (as you described in your comment), but then we saw that we use different pattern [`CorporateCardSurchargeCalculator`]( https://github.com/alphagov/pay-connector/blob/13a35e892dd8a9ae575259365591031bef5644e1/src/main/java/uk/gov/pay/connector/common/service/charge/CorporateCardSurchargeCalculator.java) and we were wondering what should we do, we decided to stick with the consistency as there were already discussion about this which we were not participating.
Lately, we agreed that we need to revisit this again.

mrlumbu:
cool, sounds good. Let's discuss soon

heathd:
if this works and delivers the functionality I think it would be fine to merge it as-is rather than waiting to have a wider discussion about design patterns (which we probably couldn't do until next Tuesday). 

We should definitely have that discussion and come back to refactor the implementation if we decide that this is not the appropriate pattern.

DanailMinchev:
We also should clarify which logic belongs to the `ChargeEntity`, because currently on `master` we have https://github.com/alphagov/pay-connector/blob/13a35e892dd8a9ae575259365591031bef5644e1/src/main/java/uk/gov/pay/connector/common/service/charge/CorporateCardSurchargeCalculator.java#L32
which probably should be part of `ChargeEntity` because the logic that method is linked to only one class which is `ChargeEntity` and we were wondering with @stephencdaly why it should be separate and not implemented in `ChargeEntity`.

On the other hand we have more complex logic https://github.com/alphagov/pay-connector/blob/13a35e892dd8a9ae575259365591031bef5644e1/src/main/java/uk/gov/pay/connector/common/service/charge/CorporateCardSurchargeCalculator.java#L18
which is more like service level logic (or separate service helper class like `CorporateCardSurchargeCalculator`).

So, I think more complex logic should live in `CorporateCardSurchargeCalculator` or in the service as it is not "pure entity logic".

The same discussion was raised in https://github.com/alphagov/pay-connector/pull/745 and the decision was to use calculators - https://github.com/alphagov/pay-connector/pull/745#discussion_r225218219 as commented by @mrlumbu if we are not going to use entities.

I think if we are going to refactor we could do more decoupled scenario proposed by @alexbishop1 in https://github.com/alphagov/pay-connector/pull/745#discussion_r225235016

Another similar PR is https://github.com/alphagov/pay-connector/pull/726 (cc @SandorArpa @danworth @georgeracu )

We should discuss the previous two PR decisions as well when finalising the approach.

danworth (dismissed):
Approved with the caveat that whether this is a pattern we wish to pursue is to be discussed and these changes could be refactored depending on the outcome of that decision.

src/test/java/uk/gov/pay/connector/charge/util/RefundCalculatorTest.java
```diff
@@ -0,0 +1,83 @@
...
+                .build();
+
+        long actualAmount = RefundCalculator.getTotalAmountToBeRefunded(chargeEntity);
+        long expectedAmount = chargeEntity.getAmount() + corporateSurcharge - refund1.getAmount() - refund2.getAmount();
+
+        assertThat(actualAmount, is(expectedAmount));
+    }
+
+    @Test
+    public void getRefundedAmount_shouldIncludeCreatedSubmittedAndRefundedRefunds() {
```
mrlumbu, line 66: readability of this test would be increased if you named them `createdRefund`,  `submittedRefund`, `refundedRefund`, `errorRefund` or similar

src/test/java/uk/gov/pay/connector/service/ChargeRefundServiceTest.java
```diff
@@ -380,21 +405,70 @@ public void shouldFailWhenChargeRefundIsNotAvailable() {
...
+                .withAmount(amount)
+                .withGatewayAccountEntity(account)
+                .withTransactionId("transactionId")
+                .withStatus(AUTHORISATION_SUCCESS)
+                .build();
+
+        when(mockChargeDao.findByExternalIdAndGatewayAccount(externalChargeId, accountId))
+                .thenReturn(Optional.of(charge));
+        expectedException.expect(RefundException.class);
+        expectedException.expectMessage("HTTP 412 Precondition Failed");
```
mrlumbu, line 434: an observation - this is probably the wrong http code to be sending back - HTTP 412 is a very specific use case https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412

src/test/java/uk/gov/pay/connector/charge/util/RefundCalculatorTest.java
stephencdaly: Good point, we'll change those names 👍

src/main/java/uk/gov/pay/connector/charge/util/RefundCalculator.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Holder for utility methods used to calculate refund amounts
+ */
+public class RefundCalculator {
+
+    private RefundCalculator() {
+        // prevent Java for adding a public constructor
+    }
+
+    /**
```
mrlumbu: Whilst I appreciate the effort of writing this javadoc commen, I'm kind of with uncle Bob on the dubious utility of javadocing unpublished classes. You are essentially duplicating information in the method signature, and not really adding any value - the method is perfectly understandable without it. Since we generally have not done this across Pay, I'd rather we didn't start now

src/test/java/uk/gov/pay/connector/service/ChargeRefundServiceTest.java
DanailMinchev: Currently it is returned by public api and it is described in the docs https://gds-payments.gelato.io/docs/versions/1.0.2/resources/payment-id/endpoints/submit-a-refund-for-a-payment as @stephencdaly noted, it may also be used by someone as it is in public api, not sure it is safe to change it directly to the correct `400 Bad Request`

SandorArpa (approved):
LOKTM

PR #767, opened 2018-10-24
PP-4301 Refactor gateway account resource 

IreneLau-GDS:
## WHAT
Currently all business logics live inside of the individual methods in the resources in quite a few places. Here are the things that we are doing which we want to improve:
* Deserialising the input into `JsonNode` with explicit inline validations
* All the business logic lives inside of the method

This story (PP-4031) is to start refactoring the code in the `createGatewayAccount` endpoint in `GatewayAccountResource` such that:
* We use DTO objects for input with validations instead of a JsonNode
* We use a service class with all the business logic
  * Transform the input DTO into the Entity object
  * Calls the persistence method(s)
  * Transform the Entity object to the output DTO object
* The resource method will only need to call the service to do all the work and it uses 

## HOW 
Since we are restructuring the call stack, we'll be implementing new tests to cover:
* The new service layer
* The JSON deserialisation & the validation

We'll also ensure that we regression tests what we've done have not broken anything by using the existing integration tests.


IreneLau-GDS:
@kbottla , re:
> Just a thought !! Feels like this method is responsible for both creating gateway account and also converting result to response which are independent ?

That's the idea. We want to avoid exposing the entity object to the resource class. Essentially the service layer should be doing all the business logic including the transformation between things. 

I personally would not want to expose the entity object outside of the service because:
* The service's method is transactional so if anything fails the database rollback is automatically taken care of. (Imagine the case we fail to do the last transform and it throws an exception, the REST call will return a 500 or whatever we have mapped it to. If we do the transform outside of the service layer it'd mean that things have already been persisted. This leaves the call in a half processed state.) - I understand that the transform is very lightweight here but if we are going to "start a trend to do it right" then we might as well going the whole nine yards.
* If we are using JPA (which I understand we aren't in connector but some other places yes), exposing the entity object outside of the transaction boundary could introduce the risk of someone doing some extra modification outside. I fear the possibility of the entity being modified that way... Same theory as above, I worry when we decide to do similar kind of refactoring somewhere and someone says "do what we have done in connector" then we'll be shooting ourselves in our foot.

I might have totally misunderstood you. Is that what you are concern about?

simad:
Just wanted to say this PR makes me feel warm and fuzzy inside 💟

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
```diff
@@ -299,7 +301,7 @@ public void createGatewayAccountWithMissingProviderUrlTypeCreatesTestType() {
         assertGettingAccountReturnsProviderName(response, "worldpay", TEST);
     }
 
-    @Test
+    @Ignore
```
kbottla: I think we no longer need this test as it is covered in GatewayAccountResourceValidationTest ?

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
```diff
@@ -310,7 +312,7 @@ public void createGatewayAccountWithIncorrectProviderUrlType() {
                 .body("message", is("Unsupported payment provider account type 'incorrect-type', should be one of (test, live)"));
     }
 
-    @Test
+    @Ignore
```
kbottla: I think we no longer need this test as it is covered in GatewayAccountResourceValidationTest ?

kbottla (commented):
Looking very nice now 👍. I am glad abstract class has been removed.
 
May be we can add few tests for new objects/methods introduced probably in a separate PR

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -211,67 +210,31 @@ public Response getGatewayAccountAcceptedCardTypes(@PathParam("accountId") Long
...
-        account.put("gateway_account_id", String.valueOf(entity.getId()));
-        account.put(PROVIDER_ACCOUNT_TYPE, entity.getType());
-        account.put(DESCRIPTION_FIELD_NAME, entity.getDescription());
-        account.put(SERVICE_NAME_FIELD_NAME, entity.getServiceName());
-        account.put(ANALYTICS_ID_FIELD_NAME, entity.getAnalyticsId());
-
-        addSelfLink(newLocation, account);
-
-        return Response.created(newLocation).entity(account).build();
+        return Response.created(gatewayAccountResponse.getLocation()).entity(gatewayAccountResponse).build();
```
mrlumbu, line 223: as an aside, I'm pretty sure nothing uses the header created in 
```
.created(gatewayAccountResponse.getLocation())
```
and if we remove that, we could just return gatewayAccountResponse without any of the wrapping. But this could be done separately, and also I might be wrong

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountObjectConvertor.java
```diff
@@ -0,0 +1,50 @@
+package uk.gov.pay.connector.gatewayaccount.service;
```
mrlumbu, line 1: I think I prefer the idea of having `from` methods on the relevant Classes to enable these kind of conversions, but this is fine too.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceValidationTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        assertThat(response.getStatus(), is(400));
+
+        String errorMessage = response.readEntity(JsonNode.class).get("message").get(0).textValue();
+        assertThat(errorMessage, is("Unsupported payment provider account type, should be one of (test, live)"));
+    }
+    
+    @Test
+    public void shouldReturn400WhenPaymentProviderIsInvalid() {
+        GatewayAccountRequest input = new GatewayAccountRequest(null, "blockchain", null, null, null);
```
mrlumbu, line 73: :trollface:

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
```diff
@@ -299,31 +299,6 @@ public void createGatewayAccountWithMissingProviderUrlTypeCreatesTestType() {
...
-
-        givenSetup()
-                .body(payload)
-                .post(ACCOUNTS_API_URL)
-                .then()
-                .statusCode(400)
-                .contentType(JSON)
-                .body("message", is(format("Unsupported payment provider %s.", testProvider)));
-    }
-
```
mrlumbu, line 326: yay

mrlumbu (approved):
Some comments on things to maybe consider in future iterations, but I think this looks 🥇

oswaldquek:
DO NOT MERGE!!! I think there is a problem with GatewayAccountRequest not having @JsonProperty on the fields. Closing.

PR #769, opened 2018-10-24
PP-4113 Add account id in search params

simad:
## WHAT
 - we have to make sure to filter by account id, not only when finding the gateway account
   but in the actual search params


cobainc0 (approved):
LVGTM

PR #772, opened 2018-10-25
PP 4301 - Implement logic to create a Stripe Account

oswaldquek:
## WHAT
This PR adds the ability to create a Stripe account, with optional credentials.

## HOW
The approach that we took was to use Jackson's polymorphism to deserialise the request into two different types:
- A default type of `GatewayAccountRequest` with no credentials, just like the existing behaviour
- A new type for stripe with stripe credentials.

By doing that we are delegating to Dropwizard using the Jackson library to deserialise thing into the correct object. 

### HOW DID WE TEST THIS?
- New resource tests

@oswaldquek @IreneLau-GDS


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountRequest.java
```diff
@@ -10,22 +10,27 @@
...
+    @JsonProperty("payment_provider")
     private String paymentProvider;
     
     public GatewayAccountRequest(@JsonProperty("type") String providerAccountType,
                                  @JsonProperty("payment_provider") String paymentProvider,
                                  @JsonProperty("service_name") String serviceName,
                                  @JsonProperty("description") String description,
-                                 @JsonProperty("analytics_id") String analyticsId) {
+                                 @JsonProperty("analytics_id") String analyticsId,
+                                 @JsonProperty("credentials") StripeCredentials credentials) {
```
mrlumbu: hmmm this could be a bit confusing - I feel like we should probably accept any credentials here, and have separate gateway specific validation, and maybe only write the Stripe validation + deserialisation for now? Maybe? 🤔

mrlumbu:
This looks like it will work for creating new `stripe` accounts with or without credentials. If I tried to create a new `worldpay` account and included credentials in the payload, what would happen? I think it would fail because the structure of `worldpay` credentials doesn't fit the structure you are expecting in your deserialisation. The fact that it would fail is fine (this story was not about adding ability to pass credentials to `worldpay` account on creation), but  the way in which it would fail seems a bit icky to me. 
A suggestion (which you may think is overkill) is to just deserialise the credentials to a `JsonNode` as far as the resource is concerned, and then allow the service to try and create `StripeCredentials` from that generic `JsonNode`, and fail appropriately if it can't . This would make it easy to extend this functionality to the other gateways - you'd just have to write a thing that tried to convert the credentials JSON into the appropriate provider credentials structure.

IreneLau-GDS:
@mrlumbu 
How about adding a validation rule to check something like `if the account type is stripe then we check for existence of credential` in the `GatewayAccountRequest`?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountRequest.java
```diff
@@ -6,29 +6,35 @@
...
+
     private String serviceName;
 
     private String description;
 
     private String analyticsId;
 
     private String paymentProvider;
+
+    private Optional<Credentials> credentials;
```
IreneLau-GDS: I know I have participated in making the changes but...

I feel a bit uncomfortable that the constructor takes in a `Credentials` object when the member variable and its getter are returning an optional. I feels... inconsistent. 

How does everyone feel about this?

kbottla: We have an existing functionality to update credentials in generic way. May be the same can be applied with some optimisation  ?
https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java#L246

IreneLau-GDS: Sorry @kbottla I think I confused you. I mean to not deserialising it into `Optional` in constructor.

That aside, Jackson allows inheritance: https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization

I have pushed a commit with that change in it.

IreneLau-GDS:
@mrlumbu 

So... as per discussed earlier, https://github.com/FasterXML/jackson-docs/wiki/JacksonPolymorphicDeserialization supports polymorphism. 

Commit https://github.com/alphagov/pay-connector/pull/772/commits/3736b79bbd9ac4f3cce9801e0ea87b7d3bd817da contains the change.

In order to use inheritance we have to declare a new field so that Jackson knows what to deserialise it into. The downside of it is that we must introduce a new attribute to the API call (I named it `credentials_type`.)

Given that this end point is only used between SelfService and Connector, I feel that it is acceptable? What do you guys think?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
```diff
@@ -0,0 +1,19 @@
...
+public class StripeCredentials extends Credentials {
+
+    private String accountId;
+
+    public StripeCredentials(@JsonProperty("account_id") String accountId) {
+        this.accountId = accountId;
+    }
+
+    public Map<String, String> toMap() {
+        return ImmutableMap.of("account_id", accountId);
```
kbottla: I think this should be `stripe_account_id` ? as per PP-4302

src/test/java/uk/gov/pay/connector/it/resources/CreateGatewayAccountResourceTest.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+    
+    @Test
+    public void createStripeGatewayAccountWithCredentials() throws Exception {
+        Map<String, Object> payload = ImmutableMap.of(
+                "type", "test",
+                "payment_provider", "stripe",
+                "service_name", "My shiny new stripe service",
+                "credentials", ImmutableMap.of(
+                        "credentials_type", "stripe",
```
kbottla: Does it mean we need this property (credentials_type) from selfservice ?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountRequest.java
kbottla: Sorry, I misunderstood the context. I think I am on same page with you now 👍

mrlumbu:
That's pretty neat, but it does change the contract as you say, and it duplicates information that's already in the request in the `type` attribute. But overall I think it's a nice solution. We will need to change what's been done on toolbox to reflect this. What happens now if I try to create a worldpay account with worldpay credentials? Do I get a validation error?

src/test/java/uk/gov/pay/connector/it/resources/CreateGatewayAccountResourceTest.java
IreneLau-GDS: Yup... Unfortunately. If we want to fully "use Jackson and Dropwizard to do the full deserialisation so we go all the way with using DTO no more JsonNode manipulation" thing.

IreneLau-GDS:
@mrlumbu 

>  it duplicates information that's already in the request in the type attribute 

Unfortunately not because the type of the object is declared inside of the object itself. Alternatively we introduce the inheritance at the Request object. (i.e. `GatewayAccountRequest` will be the base class that "everyone else" uses and `StripeGatewayAccountRequest` will be the subclass with the extra `StripeCredentails` attribute. This way we use the `type` attribute at the request level to drive how we want to deserialise things. I don't think this is a bad idea though...

> We will need to change what's been done on toolbox to reflect this.

Yes. Unless we do what is mentioned above.

> What happens now if I try to create a worldpay account with worldpay credentials? 

I thought we don't support credentials with other payment provider? (Credential is a completely new attribute at the request object, is it not??? Looking at the previous version of the resource (https://github.com/alphagov/pay-connector/pull/767/files#diff-5734a44a9d23e39fa4c38289fcf64026L230), it was always an empty `HashMap` so we weren't doing it.

If in the future we need to pass in Worldpay credentials we'll have to create a new subclass and repeat the same exercise.

> Do I get a validation error?

No because there is no such thing just yet to trigger a validation error. 😂

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
IreneLau-GDS: I'll make the change.

oswaldquek:
@mrlumbu @IreneLau-GDS how about using a Map instead of JsonNode for the credentials? See https://github.com/alphagov/pay-connector/pull/779/commits/098309a6e2277bf0d7c3eb5f1447c4de919675c5 for an example. That way we 1) get rid of duplicate information in the request 2) no need to update toolbox 3) no need to change the contract 4) if you try to create a worldpay account with credentials nothing happens, which is ok in my opinion.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
IreneLau-GDS: Done!

mrlumbu (dismissed):
🚢 it

kbottla (approved):
LOKTM!!!

PR #773, opened 2018-10-26
PP-4205 Detect docker host dynamically

mrlumbu:
We were requiring `DOCKER_HOST` and `DOCKER_CERT_PATH`
to be set to run unit tests. Turns out we don't need them
at all. They were used to figure out a host string, which
was then not used. Deleted all the offending stuff.🤓

### DIFF
--- a/src/test/java/uk/gov/pay/connector/rules/PostgresDockerRule.java
+++ b/src/test/java/uk/gov/pay/connector/rules/PostgresDockerRule.java
@@ -10,34 +10,10 @@ import org.junit.runners.model.Statement;
 import uk.gov.pay.connector.util.PostgresContainer;
 
 import java.io.IOException;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.util.Optional;
-
-import static org.junit.Assert.assertNotNull;
 
 public class PostgresDockerRule implements TestRule {
 
-    private static String host;
     private static PostgresContainer container;
-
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/rules/PostgresDockerRule.java
```diff
@@ -24,21 +24,6 @@
     private static final String DOCKER_HOST = "DOCKER_HOST";
     private static final String DOCKER_CERT_PATH = "DOCKER_CERT_PATH";
```
stephencdaly: remove the unused constants?

mrlumbu: good shout

stephencdaly: `private static String host;`

and 

`private static final String DOCKER_HOST = "DOCKER_HOST";`
`private static final String DOCKER_CERT_PATH = "DOCKER_CERT_PATH";`

are no longer used so can be removed is what I meant

mrlumbu: ah

mrlumbu: yes

stephencdaly (approved):

PR #774, opened 2018-10-26
PP-4261 Change charges sweep to handle AWAITING_CAPTURE_REQUEST differently

SandorArpa:
## WHAT
DVLA need longer than 90 minutes before they can do a capture request:
"EVL runs a clean up batch job to check the outcome of payments which have been initiated. If the payment has been authorised but EVL hasn’t been updated accordingly (e.g. the redirect back to us failed), it completes the transaction and settles the payment. I believe this batch job runs twice per day and picks up transactions that are pending payment and are more than 2 hours old.

Also our tachos service will have to go out to Tachonet to check if a card can be issued and may not get a response for up to 24 hours – so the attempt to settle the payment will be a lot more than 90 mins after the payment was created."

We agreed to change the sweep for charges in `AWAITING_CAPTURE_REQUEST` status.

See:
[JIRA story PP-4261](https://payments-platform.atlassian.net/browse/PP-4261)

## HOW
- change sweep process to expiry AWAITING_CAPTURE_REQUEST after 48 hours
- add relevant tests

with @simad

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
@@ -48,7 +48,8 @@ import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.Response.created;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static uk.gov.pay.connector.charge.model.TransactionType.inferTransactionTypeFrom;
-import static uk.gov.pay.connector.charge.service.ChargeExpiryService.EXPIRABLE_STATUSES;
+import static uk.gov.pay.connector.charge.service.ChargeExpiryService.EXPIRABLE_REGULAR_STATUSES;
+import static uk.gov.pay.connector.charge.service.ChargeExpiryService.EXPIRABLE_AWAITING_CAPTURE_REQUEST_STATUS;
 import static uk.gov.pay.connector.charge.service.SearchService.TYPE.CHARGE;
 import static uk.gov.pay.connector.charge.service.SearchService.TYPE.TRANSACTION;
 import static uk.gov.pay.connector.util.ResponseUtil.fieldsInvalidResponse;
@@ -87,7 +88,9 @@ public class ChargesApiResource {
     private static final String DISPLAY_SIZE = "display_size";
     private static final Set<String> CHARGE_REQUEST_KEYS_THAT_MAY_HAVE_PII = Collections.singleton("description");
     private static final int ONE_HOUR_AND_A_HALF = 5400;
+    private static final int FORTY_EIGHT_HOURS = 172800;
     private static final String CHARGE_EXPIRY_WINDOW = "CHARGE_EXPIRY_WINDOW_SECONDS";
+    private static final String AWAITING_DELAY_CAPTURE_EXPIRY_WINDOW = "AWAITING_DELAY_CAPTURE_EXPIRY_WINDOW";
     private static final Logger logger = LoggerFactory.getLogger(ChargesApiResource.class);
     public static int MIN_AMOUNT = 1;
     public static int MAX_AMOUNT = 10_000_000;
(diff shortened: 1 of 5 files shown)


mrlumbu:
Could you explain the reason for making this change, and could you squash the commits to remove the WIP commit

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -289,6 +299,15 @@ private ZonedDateTime getExpiryDate() {
         return ZonedDateTime.now().minusSeconds(chargeExpiryWindowSeconds);
     }
 
+    private ZonedDateTime getExpiryDateForAwaitingCaptureRequest() {
+        int chargeExpiryWindowSeconds = FORTY_EIGHT_HOURS;
+        if (StringUtils.isNotBlank(System.getenv(AWAITING_DELAY_CAPTURE_EXPIRY_WINDOW))) {
+            chargeExpiryWindowSeconds = Integer.parseInt(System.getenv(AWAITING_DELAY_CAPTURE_EXPIRY_WINDOW));
+        }
```
mrlumbu, line 306: I think this can be put in config, then we can have a default value

src/test/java/uk/gov/pay/connector/it/resources/ChargeExpiryResourceITest.java
```diff
@@ -217,4 +219,98 @@ public void shouldPreserveCardDetailsIfChargeExpires() {
...
+        assertTrue(isEqualCollection(events2,
+                asList(AWAITING_CAPTURE_REQUEST.getValue())));
+    }
+
+    @Test
+    public void shouldExpireChargesWhenAwaitingCaptureDelayIsMoreThan48Hours() {
+        String extChargeId1 = addCharge(CREATED, "ref", ZonedDateTime.now().minusMinutes(89), RandomIdGenerator.newId());
+        String extChargeId2 = addCharge(CREATED, "ref", ZonedDateTime.now().minusMinutes(90), RandomIdGenerator.newId());
+        String extChargeId3 = addCharge(AWAITING_CAPTURE_REQUEST, "ref", ZonedDateTime.now().minusHours(48L).minusMinutes(1L), RandomIdGenerator.newId());
+
```
mrlumbu, line 267: you seem to be testing 'normal' expiry and the new expiry in same test here. Not sure why

src/test/java/uk/gov/pay/connector/it/resources/ChargeExpiryResourceITest.java
```diff
@@ -217,4 +219,98 @@ public void shouldPreserveCardDetailsIfChargeExpires() {
...
+        assertTrue(isEqualCollection(events1,
+                asList(CREATED.getValue(), EXPIRED.getValue())));
+        assertTrue(isEqualCollection(events2,
+                asList(AWAITING_CAPTURE_REQUEST.getValue())));
+    }
+
+    @Test
+    public void shouldExpireChargesWhenAwaitingCaptureDelayIsMoreThan48Hours() {
+        String extChargeId1 = addCharge(CREATED, "ref", ZonedDateTime.now().minusMinutes(89), RandomIdGenerator.newId());
```
mrlumbu: you could name these charges better to amke it clearer what they are - `chargeNotToBeExpired`, `chargeToBeExpiredFromCreatedState`, `chargeToBeExpiredFromAwaitingCaptureRequestState`

src/test/java/uk/gov/pay/connector/it/resources/ChargeExpiryResourceITest.java
```diff
@@ -217,4 +219,98 @@ public void shouldPreserveCardDetailsIfChargeExpires() {
         assertThat(chargeCardDetails.get("address_country"), is(notNullValue()));
     }
 
+    @Test
+    public void shouldNotExpireChargesWhenAwaitingCaptureDelayIsLessThan48Hours() {
+        String extChargeId1 = addCharge(CREATED, "ref", ZonedDateTime.now().minusMinutes(90), RandomIdGenerator.newId());
```
mrlumbu: not sure what role this charge is playing in the test

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -274,13 +277,20 @@ public Response createNewCharge(@PathParam(ACCOUNT_ID) Long accountId, Map<Strin
     @Path("/v1/tasks/expired-charges-sweep")
     @Produces(APPLICATION_JSON)
     public Response expireCharges(@Context UriInfo uriInfo) {
```
mrlumbu, line 279: feels like this resource is now doing a lot of work. I would prefer this logic to be pushed down to a service, and then you can unit test more at that level, and not need too many integration tests. Thoughts?

mrlumbu (changes requested):

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
SandorArpa: The api document states: `The default value can be overridden by setting an environment variable CHARGE_EXPIRY_WINDOW_SECONDS in seconds.` I have searched for the variable `CHARGE_EXPIRY_WINDOW_SECONDS` but it is only present in this project. It seems that the original intention was to use config to override the default value.
IMHO if we want to use a config for the default value then we would need to open a separate ticket.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
SandorArpa: I agree, we can refactor and make a service call by a method without any parameters (eg `findAllChargesToBeExpired()`) and let the service handle the logic and time frames. However, if we decide to push the logic down then I'd like to do that in a separate PR, if possible.

src/test/java/uk/gov/pay/connector/it/resources/ChargeExpiryResourceITest.java
SandorArpa: 👍

src/test/java/uk/gov/pay/connector/it/resources/ChargeExpiryResourceITest.java
SandorArpa: To make sure that the combined search for charges to be expired works in both cases.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
mrlumbu: sounds sensible

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
mrlumbu: Not sure what api document you're referring to. I see that this code copies the original code that just got the expiry window from the environment directly, rather than from config, but it is pretty much the only bit of config that is read in directly from the environment, so would be nice to change it to conform to the way we have done it in all other cases. But if you don't want to do it here, that's fine

SandorArpa: I'm referring to the projects `api_specification.md`.

mrlumbu (approved):

PR #775, opened 2018-10-26
BAU: Java 1.8.0_191 test fix

DanailMinchev:
## WHAT

- When upgraded from Java `1.8.0_141` to `1.8.0_191` the messaging has changed. This PR makes sure we have the new one and we should remove the old one once we migrate to `Java >= 1.8.0_191`


DanailMinchev:
Looks like the exception was different because my ISP was resolving http://invalidone.invalid.
Going to close this PR.

PR #776, opened 2018-10-26
PP-4304 Added POJO for create charge request

georgeracu:
## WHAT
- Added a POJO object to represent the request object for creating a charge
- Moved validation from resource to POJO
- Updated service to get as input the new object
- Updated tests
- Removed unnecessary methods
- Added custom JSON deserializer for SupportedLanguage. Later on this will
be moved into java commons package
- Added unit test for the `toString()` representation without possible
personally identifiable information


src/main/java/uk/gov/pay/connector/gatewayaccount/model/ChargeCreateRequest.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @NotNull(message = "Field [return_url] cannot be null")
+    @JsonProperty("return_url")
+    private String returnUrl;
+
+    @Length(max = 254, message = "Field [email] can have a size between 0 and 254")
+    @JsonProperty("email")
+    private String email;
+
+    @JsonProperty("delayed_capture")
```
kbottla: Although the default is false, can we explicitly set it to 'false'  ? I just feel it can rule out assumptions

src/main/java/uk/gov/pay/connector/gatewayaccount/model/ChargeCreateRequestBuilder.java
```diff
@@ -0,0 +1,59 @@
+package uk.gov.pay.connector.gatewayaccount.model;
+
+import uk.gov.pay.commons.model.SupportedLanguage;
+
+public final class ChargeCreateRequestBuilder {
```
kbottla: Do we need this ? It is only used in test. Could probably be test helper if we want nice way to build test objects

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -76,12 +77,7 @@
     private static final long CHARGE_ENTITY_ID = 12345L;
     private static final String[] EXTERNAL_CHARGE_ID = new String[1];
 
-    private static final Map<String, String> CHARGE_REQUEST = new HashMap<String, String>() {{
-        put("amount", "100");
-        put("return_url", "http://return-service.com");
-        put("description", "This is a description");
-        put("reference", "Pay reference");
-    }};
+    private ChargeCreateRequestBuilder REQUEST_BUILDER;
```
kbottla: ALL CAPS ? is it intentional ? Since it is not a constant I feel we can use usual Java convention

src/main/java/uk/gov/pay/connector/gatewayaccount/model/ChargeCreateRequest.java
```diff
@@ -0,0 +1,106 @@
+package uk.gov.pay.connector.gatewayaccount.model;
```
kbottla: Should this be in package `uk.gov.pay.connector.charge.model`

kbottla: It should probably be in package `uk.gov.pay.connector.charge.model`

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
```diff
@@ -311,7 +316,12 @@ public void cannotMakeChargeForMissingFields() {
                 .contentType(JSON)
                 .header("Location", is(nullValue()))
                 .body(JSON_CHARGE_KEY, is(nullValue()))
-                .body(JSON_MESSAGE_KEY, is("Field(s) missing: [amount, description, reference, return_url]"));
+                .body(JSON_MESSAGE_KEY, containsInAnyOrder(
```
kbottla, line 319: Format of error messages is different to what we had before. But due to fact the publicapi doesn't process the response as it is from connector don't see any problem here

**Previous**
{    "message": "Field(s) missing: [reference]"}

**Current**
{    "message": [ 
        "Field [reference] cannot be null"
    ]}

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
georgeracu: Well spotted

src/main/java/uk/gov/pay/connector/gatewayaccount/model/ChargeCreateRequestBuilder.java
georgeracu: Will move it to tests, I didn't notice where it was created. We need a way to create an object like this, so I believe that a builder is a good fit.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/ChargeCreateRequest.java
georgeracu: Moved

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
georgeracu: 👍

src/main/java/uk/gov/pay/connector/gatewayaccount/model/ChargeCreateRequest.java
georgeracu: 👍

src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public String toStringWithoutPersonalIdentifiableInformation() {
+        return "ChargeCreateRequest{" +
+                "amount=" + amount +
+                ", reference='" + reference + '\'' +
+                ", returnUrl='" + returnUrl + '\'' +
+                delayedCapture != null ? ", delayed_capture=" + delayedCapture : "" +
+                language != null ? ", language=" + language.toString() : "" +
+                email != null ? ", email='" + email + '\'' : "" +
```
kbottla: I think we shouldn't log PIDs (email) ? At present we are logging reference, amount, return_url. New fields (delayed_capture, language) seems fine.

src/test/java/uk/gov/pay/connector/charge/model/ChargeCreateRequestTest.java
```diff
@@ -0,0 +1,19 @@
+package uk.gov.pay.connector.charge.model;
+
+import org.junit.Test;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class ChargeCreateRequestTest {
```
kbottla, line 8: Probably new test if we decided to remove email from toString

src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
mrlumbu: yes we should not log the email

georgeracu: Removed email

src/test/java/uk/gov/pay/connector/charge/model/ChargeCreateRequestTest.java
georgeracu: Added test for email

mrlumbu (approved):
This looks like a great tidying up of some fairly messy validation logic. Nice one

PR #777, opened 2018-10-29
PP-4276: Migrate test to use DropwizardJUnitRunner

oswaldquek:
This is the last test we can migrate to DropwizardJUnitRunner, might as well do
it (even if the story is done).

@oswaldquek


georgeracu (dismissed):
LGTM

georgeracu (approved):
LGTM

PR #778, opened 2018-10-29
PP-4261 Refactor ChargesApiResource, move 'sweep and expire' logic to ChargeExpiryService

SandorArpa:
## WHAT
Refactor `ChargesApiResource`. Move the 'sweep and expire' process to `ChargeExpiryService`. This is needed as the business logic has been extended [(previous PR for PP-4261)](https://github.com/alphagov/pay-connector/pull/774) and it got too complicated and too big to have it on the resource level.

## HOW 
- push logic of sweep and expiry of charges to `ChargeExpiryService`
- add relevant tests
- rearrange `ChargeExpiryServiceTest` tests, move common setups under `@Before`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
@@ -3,6 +3,7 @@ package uk.gov.pay.connector.charge.service;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Provider;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -174,4 +192,22 @@ private void logUnsuccessfulResponseReasons(ChargeEntity chargeEntity, GatewayRe
                 logger.error("Gateway error while cancelling the Charge - charge_external_id={}, gateway_error={}",
                         chargeEntity.getExternalId(), error));
     }
+
+    private ZonedDateTime getExpiryDateForRegularCharges() {
+        int chargeExpiryWindowSeconds = ONE_HOUR_AND_A_HALF;
+        if (StringUtils.isNotBlank(System.getenv(CHARGE_EXPIRY_WINDOW))) {
```
mrlumbu, line 198: Thinking a bit more about these calls to `getenv`: if the specified env var isn't present, this throws a `NullPointerException`, so I really think we should refactor these two methods to pull the expiry windows from config. That way we can set sensible defaults, and not run the risk of deploying to an env where it isn't set, and having it all blow up.

mrlumbu: Sorry, I'm totally wrong

mrlumbu (approved):

PR #780, opened 2018-10-30
PP-4261 Refactor charges sweep to use config file

SandorArpa:
## WHAT
Use config file for charge expiry windows instead of hard coded values.

## HOW 
- refactor ChargeExpiryService to use config file for expiry windows

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
@@ -9,7 +9,7 @@ import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 
-public class    ConnectorConfiguration extends Configuration {
+public class ConnectorConfiguration extends Configuration {
 
     @Valid
     @NotNull
@@ -62,6 +62,11 @@ public class    ConnectorConfiguration extends Configuration {
     @NotNull
     @JsonProperty("customJerseyClient")
     private CustomJerseyClientConfiguration customJerseyClient;
+    
+    @Valid
+    @NotNull
+    @JsonProperty("chargesSweepConfig")
+    private ChargeSweepConfig chargeSweepConfig;
 
     @NotNull
     private String graphiteHost;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/app/ChargesSweepTimeConfig.java
```diff
@@ -0,0 +1,17 @@
+package uk.gov.pay.connector.app;
+
+import io.dropwizard.Configuration;
+
+public class ChargesSweepTimeConfig extends Configuration {
```
mrlumbu: Can we call it `ChargeSweepConfig` (this applies for this class, but also to config etc)

src/main/java/uk/gov/pay/connector/app/ChargesSweepTimeConfig.java
```diff
@@ -0,0 +1,17 @@
...
+
+    private int chargeExpirySeconds;
```
mrlumbu: can we call this `defaultChargeExpiryThreshold` or similar, and make same changes across getters, config etc. This might seem a bit pernickety, but I think it better explains what it is

src/main/java/uk/gov/pay/connector/app/ChargesSweepTimeConfig.java
```diff
@@ -0,0 +1,17 @@
...
+    private int awaitingCaptureExpirySeconds;
```
mrlumbu: likewise can we call this `awaitingCaptureExpiryThreshold`

mrlumbu (changes requested):

src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
```diff
@@ -62,6 +62,11 @@
     @NotNull
     @JsonProperty("customJerseyClient")
     private CustomJerseyClientConfiguration customJerseyClient;
+    
+    @Valid
+    @NotNull
+    @JsonProperty("chargesSweepTimeConfig")
```
danworth: Should this JsonProperty be updated to `chargesSweepConfig` to match the name change of the class?

SandorArpa: 👍

danworth (approved):
Looks good, basing my review on-top of Dom's rather than starting again. Changes requested have been made and looks good.

mrlumbu (approved):

PR #782, opened 2018-10-30
PP-4297 return capture url when a charge is eligible

SandorArpa:
## WHAT
```
As a service that uses GOV.UK Pay for delayed capture payments
I want to be able to see the capture URL in a payment from public API if the payment can be captured
So that I can determine when I can capture a payment and what URL I need to use to do this
```
Return `capture` url when a charge state is in `AWAITING_CAPTURE_REQUEST`.

## HOW 
- add URL when eligible to ChargeResponse
- add relevant test

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
@@ -73,7 +73,7 @@ import static uk.gov.pay.connector.model.domain.ChargeEntityFixture.aValidCharge
 public class ChargeServiceTest {
 
     private static final String SERVICE_HOST = "http://my-service";
-    private static final long GATEWAY_ACCOUNT_ID = 1L;
+    private static final long GATEWAY_ACCOUNT_ID = 10L;
     private static final long CHARGE_ENTITY_ID = 12345L;
     private static final String[] EXTERNAL_CHARGE_ID = new String[1];
 
@@ -101,6 +101,8 @@ public class ChargeServiceTest {
     private PaymentProvider mockedPaymentProvider;
 
     private ChargeService service;
+    
+    private GatewayAccountEntity gatewayAccount;
 
     @Before
     public void setUp() {
@@ -111,7 +113,7 @@ public class ChargeServiceTest {
                 .withDescription("This is a description")
                 .withReference("Pay reference");
 
-        GatewayAccountEntity gatewayAccount = new GatewayAccountEntity("sandbox", new HashMap<>(), TEST);
+        gatewayAccount = new GatewayAccountEntity("sandbox", new HashMap<>(), TEST);
         gatewayAccount.setId(GATEWAY_ACCOUNT_ID);
 
         // Populate ChargeEntity with ID when persisting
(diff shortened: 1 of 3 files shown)


mrlumbu:
Possibly a silly question, but why is this needed?

SandorArpa:
@mrlumbu I added the reason from the story to the description.
The way we present a charge's state through publicapi is not clear enough to determine whether a charge is awaiting capture request.
```
"state": {
        "status": "submitted",
        "finished": false
    }
```
We do have a field `"delayed_capture": true` but even the combination of the two can refer to either`AWAITING_CAPTURE_REQUEST` or `AUTHORISATION_SUCCESS`

mrlumbu:
Ok cool. Can we talk about this? The thing I find a bit strange is that we are generating a URL in connector, but obviously ultimately the URL that will be required is a publicapi one. So really publicapi will have to look at whether this url is present in the response from the connector, then do something with it (i.e create a new url, which will probably be entirely different from this one), at which point this url is really just acting as a flag, so may as well be a flag. Quite possibly I'm missing something here

alexbishop1:
I don’t think you’re missing anything, @mrlumbu. This could be a flag, I suppose (I didn’t think of that when writing the story). A URL is nice from a RESTful perspective in that connector is telling you about operations you can perform on the URL, similar to how it includes other links like `self`, `refunds` and `next_url` (though `next_url` is a bit different because it’s passed straight through by public API without being converted from a connector URL to a public API URL).

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -414,6 +409,38 @@ public void shouldUpdateTransactionStatus_whenUpdatingChargeStatusFromInitialSta
         service.updateFromInitialStatus(createdChargeEntity.getExternalId(), ENTERING_CARD_DETAILS);
 
     }
+    
+    @Test
+    public void shouldFindChargeWithCaptureUrlAndNoNextUrl_whenChargeInAwaitingCaptureRequest() throws Exception{
+        Long chargeId = 101L;
+        Long accountId = 10L;
```
alexbishop1: Why do we have an `accountId` of 10 and various hard-coded account IDs of 1?

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -414,6 +409,38 @@ public void shouldUpdateTransactionStatus_whenUpdatingChargeStatusFromInitialSta
...
+        when(mockedChargeDao.findByExternalIdAndGatewayAccount(externalId, accountId)).thenReturn(chargeEntity);
+
+        Optional<ChargeResponse> chargeResponseForAccount = service.findChargeForAccount(externalId, accountId, mockedUriInfo);
+
+        verify(mockedTokenDao, never()).persist(any());
+
+        ChargeResponseBuilder expectedChargeResponse = chargeResponseBuilderOf(chargeEntity.get());
+        expectedChargeResponse.withLink("self", GET, new URI(SERVICE_HOST + "/v1/api/accounts/1/charges/" + externalId));
+        expectedChargeResponse.withLink("refunds", GET, new URI(SERVICE_HOST + "/v1/api/accounts/1/charges/" + externalId + "/refunds"));
+        expectedChargeResponse.withLink("capture", POST, new URI(SERVICE_HOST + "/v1/api/accounts/1/charges/" + externalId + "/capture"));
```
alexbishop1: I wonder if it’s worth having a test (or adding to another test) to check the capture URL isn’t included if the payment is in another state

alexbishop1 (changes requested):
Looks good — just a query about the test.

What’s going on with `ChargesApiResourceGetChargesJsonITest` and `ChargesApiResourceJsonPaginationITest`?

And let’s chat with Dom about this being a flag.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
SandorArpa: I did consider adding another test but given how the other tests are written (comparing the returned and an expected object), they would break if capture url surfaced in any other charge state.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
SandorArpa: copy-paste, will fix it

SandorArpa:
Both `ChargesApiResourceGetChargesJsonITest` and `ChargesApiResourceJsonPaginationITest` were empty classes. Looking back at the repository history their content was removed when we got rid of the unused endpoint they correspond to.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -111,7 +113,7 @@ public void setUp() {
                 .withDescription("This is a description")
                 .withReference("Pay reference");
 
-        GatewayAccountEntity gatewayAccount = new GatewayAccountEntity("sandbox", new HashMap<>(), TEST);
+        gatewayAccount = new GatewayAccountEntity("sandbox", new HashMap<>(), TEST);
```
alexbishop1, line 116: Nice change to enable to removal of duplicate code

alexbishop1 (approved):
Unfortunately, Dom isn’t working today so we can’t discuss the URL/flag issue with him but I’m happy with this, so approving

PR #783, opened 2018-10-30
PP-4113 Refund status should be external rather than internal

simad:
## WHAT
- We should always display external statuses in publicapi rather than internal ones.


alexbishop1 (approved):
Change looks good.

I think I would have preferred the code tidy-up to be separate from the actual functional changes but the whole thing is so small that it’s no big deal.

PR #784, opened 2018-10-30
PP-4341: Refactor Worldpay payment provider

oswaldquek:
## WHAT
Makes WorldpayPaymentProvider nicer:
- It can be creating by Guice via the @Inject, which means
- It is responsible for creating its own clients
- We can start getting rid of those `sendReceive` method in there
- Some lines of tests can be removed

@oswaldquek


mrlumbu (approved):
Looks like a great first step to me

PR #785, opened 2018-10-31
PP-4341: Refactor EpdqPaymentProvider

oswaldquek:
Similar to
https://github.com/alphagov/pay-connector/commit/dfef04523e22d78980cefa55d0a7015c4766e5ff
this PR makes EpdqPaymentProvider nicer:
- It can be creating by Guice via the @Inject, which means
- It is responsible for creating its own clients
- We can start getting rid of those `sendReceive` method in there
- Some lines of tests can be removed

We are a step closer to getting rid of BasePaymentProvider and its ugly
`EnumMap<GatewayOperation, GatewayClient> clients` field.

@oswaldquek

## WHAT

## HOW 
_Steps to test or reproduce:_

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProviderTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProviderTest.java
@@ -206,7 +206,7 @@ public class EpdqPaymentProviderTest extends BaseEpdqPaymentProviderTest {
 
         assertThat(provider.verifyNotification(mockNotification, mockGatewayAccountEntity), is(false));
     }
-
+    
     @Test
     public void shouldNotVerifyNotificationIfEmptyPayload() {
         when(mockNotification.getPayload()).thenReturn(Optional.empty());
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProviderTest.java
```diff
@@ -162,51 +162,6 @@ public void shouldNotRefundIfPaymentProviderReturnsNon200HttpStatusCode() {
                 UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY));
     }
 
-    @Test
```
mrlumbu: might be missing something here, but why are you deleting these three tests?

oswaldquek: it's because it's not testable. we used to pass a SignatureGenerator into the EpdqPaymentProvider constructor https://github.com/alphagov/pay-connector/blob/8452d22a58548caca05a237221deb4e99e5f3701/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java#L82 but I saw no point in passing it in as a constructor as it's instantiated with `new EpdqSha512SignatureGenerator()` https://github.com/alphagov/pay-connector/blob/4dcddc052832dab803846947db217dd3fc33ef31/src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java#L88 since it's no longer passed in the constructor, it's not mockable, and therefore not testable.

mrlumbu: I see. Is there a reason why we are not injecting the `EpdqSha512SignatureGenerator` - then we could keep these tests, that seem like worth keeping?

oswaldquek: I guess I'm not injecting `EpdqSha512SignatureGenerator` because it's such an easy class to instantiate with `new...`. I'll inject it into EpdqPaymentProvider so we can keep these tests.

mrlumbu (approved):

PR #787, opened 2018-10-31
[DO NOT MERGE] Experimental flattening refactor

heathd:
This is an experimental refactoring I did some time ago. It was focused around flattening the inheritance heirarchy used in notification handling.

I didn't write the commits carefully so I wanted to revisit it before merging it. But basically it worked. It's probably now hopelessly out of date, but some of the ideas might still be useful.

key points:

1. introduced [service objects](https://github.com/alphagov/pay-connector/compare/experimental-flattening-refactor?expand=1#diff-5cd35ce7e6595c627804ef13be020270R41) for handling notifications from each gateway
2. notification services handle the parsing of the notification; this removes the need for ‘status mappers’ because the logic is now localised to each payment gateway. The worldpay status mapper becomes [this](https://github.com/alphagov/pay-connector/commit/ed874f8b49f71fe00f2ac3dffe6523a34db6b978#diff-cad860456f2aaabe09ed02e1252caa24R98) for example
3. introduced [notification processors](https://github.com/alphagov/pay-connector/compare/experimental-flattening-refactor?expand=1#diff-90396682d87b92225d6083c88ca9ba76R17) which are the slice of behaviour common to all gateways. They perform the action once a notification has been parsed and interpreted.
4. introduced [modules to contain gateway-specific injected classes](https://github.com/alphagov/pay-connector/compare/experimental-flattening-refactor?expand=1#diff-cf4242aa182c59b1a026017116b41761R12)
5. started rewriting [PaymentProviders to receive individual providers as injected parameters rather than building them](https://github.com/alphagov/pay-connector/compare/experimental-flattening-refactor?expand=1#diff-01be8416a32c91605d811c7daea2dd3bR45)


kbottla:
Notifications design has been refactored
- There is a new notification service now for each payment gateway

Also, generics have been removed from PaymentProvider interface

PR #788, opened 2018-10-31
PP-4341: Refactor SmartpayPaymentProvider

oswaldquek:
## WHAT

Makes SmartpayPaymentProvider nicer:
- It can be created by Guice via the @Inject, which means
- It is responsible for creating its own clients, of which there is only ONE,
  unlike Worldpay and Epdq providers, which have FOUR -> simplicity!
- We can start getting rid of those sendReceive method in there

This commit reduces the number of GatewayClients from 4 previously to 1, as it
makes no sense to have 4 GatewayClients. Worldpay/Epdq payment providers have
4 because as shown in
https://github.com/alphagov/pay-connector/blob/master/src/main/resources/config/config.yaml#L29
they have different configured read timeouts which require a different client
for each operation (authorise, cancel, refund, capture). Smartpay doesn't have
these read timeouts so it makes sense to just use one GatewayClient. A side
effect of this is that GatewayClientFactory needs to be able to create a
GatewayClient without an operation parameter. This means the metric registry
names
(https://github.com/alphagov/pay-connector/blob/cde8bb205be1e3e33d8059055e994f5c290b4b2f/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java#L102)
for Smartpay are no longer "smartpay.authorise", "smartpay.cancel", etc but
simply "smartpay.all".

@oswaldquek

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
@@ -26,9 +26,12 @@ import uk.gov.pay.connector.util.TrustStoreLoader;
 import javax.inject.Inject;
 import javax.net.ssl.HostnameVerifier;
 import javax.ws.rs.client.Client;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
+import static org.glassfish.jersey.apache.connector.ApacheClientProperties.CONNECTION_MANAGER;
+import static org.glassfish.jersey.client.ClientProperties.READ_TIMEOUT;
 
 public class ClientFactory {
     private final Environment environment;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
```diff
@@ -40,15 +43,22 @@ public ClientFactory(Environment environment, ConnectorConfiguration conf) {
         this.conf = conf;
     }
 
-    public Client createWithDropwizardClient(PaymentGatewayName gateway, GatewayOperation operation, MetricRegistry metricRegistry) {
+    public Client createWithDropwizardClient(PaymentGatewayName gateway, Optional<GatewayOperation> operation, MetricRegistry metricRegistry) {
```
mrlumbu: Using Optionals as an argument is a bit smelly, and I think in this case is not necessary - why not have an overloaded `createWithDropwizardClient` (as you already have with `createGatewayClient`)?

oswaldquek: @mrlumbu fixed

mrlumbu: Ok. I would prefer if we removed the optional entirely, but probably getting into the weeds there. This looks fine for now

mrlumbu (approved):

PR #792, opened 2018-11-02
PP-4313 Add decrypter for apple pay payloads

simad:
## WHAT
A first stab at integrating Apple's suggested implementation of a decrypter. There are a few things which will be improved, like `private Map<String, String> paymentData;` (still unsure as to what values contained in the map we are actually going to use, so will be more strongly typed as other stories progress), and exception throwing/handling.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
@@ -1,0 +1,15 @@
+package uk.gov.pay.connector.app;
+
+
+public class ApplePayConfig extends GatewayConfig {
+    private String publicCertificate;
+    private String privateKey;
+
+    public byte[] getPublicCertificate() {
+        return publicCertificate.getBytes();
+    }
+
+    public byte[] getPrivateKey() {
+        return privateKey.getBytes();
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
@@ -1,15 +1,17 @@
 package uk.gov.pay.connector.app;
 
 
-public class ApplePayConfig extends GatewayConfig {
+import java.nio.charset.StandardCharsets;
+
+public class ApplePayConfig {
     private String publicCertificate;
     private String privateKey;
 
     public byte[] getPublicCertificate() {
-        return publicCertificate.getBytes();
+        return publicCertificate.getBytes(StandardCharsets.UTF_8);
     }
 
     public byte[] getPrivateKey() {
-        return privateKey.getBytes();
+        return privateKey.getBytes(StandardCharsets.UTF_8);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
@@ -1,10 +1,12 @@
 package uk.gov.pay.connector.app;
 
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 public class WorldpayConfig extends GatewayConfig {
     private boolean secureNotificationEnabled;
     private String notificationDomain;
-    private ApplePayConfig applePay;
+    private ApplePayConfig applePayConfig;
 
     public String getNotificationDomain() {
         return notificationDomain;
@@ -14,7 +16,8 @@ public class WorldpayConfig extends GatewayConfig {
         return secureNotificationEnabled;
     }
 
+    @JsonProperty("applePay")
     public ApplePayConfig getApplePayConfig() {
-        return applePay;
+        return applePayConfig;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
@@ -14,7 +14,7 @@ public class WorldpayConfig extends GatewayConfig {
         return secureNotificationEnabled;
     }
 
-    public ApplePayConfig getApplePay() {
+    public ApplePayConfig getApplePayConfig() {
         return applePay;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
@@ -1,8 +1,10 @@
 package uk.gov.pay.connector.app;
 
+
 public class WorldpayConfig extends GatewayConfig {
     private boolean secureNotificationEnabled;
     private String notificationDomain;
+    private ApplePayConfig applePay;
 
     public String getNotificationDomain() {
         return notificationDomain;
@@ -10,5 +12,9 @@ public class WorldpayConfig extends GatewayConfig {
 
     public boolean isSecureNotificationEnabled() {
         return secureNotificationEnabled;
+    }
+
+    public ApplePayConfig getApplePay() {
+        return applePay;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
@@ -15,15 +15,19 @@ import javax.crypto.spec.IvParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.security.KeyFactory;
 import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.Security;
 import java.security.cert.Certificate;
+import java.security.cert.CertificateException;
 import java.security.cert.CertificateFactory;
 import java.security.cert.X509Certificate;
 import java.security.interfaces.ECPublicKey;
+import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 
@@ -107,15 +111,14 @@ public class ApplePayDecrypter {
                 .digest(byteArrayOutputStream.toByteArray());
     }
 
-    private Certificate generateCertificate() throws Exception {
-        InputStream stream = new ByteArrayInputStream(publicCertificate);
-        CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
-        Certificate certificate = certificateFactory.generateCertificate(stream);
-        stream.close();
-        return certificate;
+    private Certificate generateCertificate() throws IOException, CertificateException {
+        try (InputStream stream = new ByteArrayInputStream(publicCertificate)) {
+            CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+            return certificateFactory.generateCertificate(stream);
+        }
     }
 
-    private PrivateKey generatePrivateKey() throws Exception {
+    private PrivateKey generatePrivateKey() throws NoSuchAlgorithmException, InvalidKeySpecException {
         return KeyFactory.getInstance("EC")
                 .generatePrivate(new PKCS8EncodedKeySpec(privateKeyBytes));
     }
--- a/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
@@ -2,13 +2,12 @@ package uk.gov.pay.connector.applepay;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
-import com.sun.org.apache.xerces.internal.impl.dv.util.HexBin;
 import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.bouncycastle.util.encoders.Hex;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ApplePayConfig;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
-import uk.gov.pay.connector.util.ResponseUtil;
 
 import javax.crypto.Cipher;
 import javax.crypto.KeyAgreement;
@@ -17,7 +16,6 @@ import javax.crypto.spec.SecretKeySpec;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
-import java.nio.charset.Charset;
 import java.security.KeyFactory;
 import java.security.MessageDigest;
 import java.security.PrivateKey;
@@ -29,10 +27,11 @@ import java.security.interfaces.ECPublicKey;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 public class ApplePayDecrypter {
     private static final Logger LOGGER = LoggerFactory.getLogger(ApplePayDecrypter.class);
 
-    private static final Charset UTF_8 = Charset.forName("UTF-8");
     private static final byte[] COUNTER = {0x00, 0x00, 0x00, 0x01};
     private static final byte[] APPLE_OEM = "Apple".getBytes(UTF_8);
     private static final byte[] ALG_IDENTIFIER_BYTES = "id-aes256-GCM".getBytes(UTF_8);
@@ -48,7 +47,7 @@ public class ApplePayDecrypter {
 
     @Inject
     public ApplePayDecrypter(ConnectorConfiguration configuration, ObjectMapper objectMapper) {
-        ApplePayConfig applePayConfig = configuration.getWorldpayConfig().getApplePay();
+        ApplePayConfig applePayConfig = configuration.getWorldpayConfig().getApplePayConfig();
         this.privateKeyBytes = applePayConfig.getPrivateKey();
         this.publicCertificate = applePayConfig.getPublicCertificate();
         this.objectMapper = objectMapper;
@@ -103,7 +102,7 @@ public class ApplePayDecrypter {
         byteArrayOutputStream.write(ALG_IDENTIFIER_BYTES);
         byteArrayOutputStream.write(APPLE_OEM);
         // Add Merchant Id
-        byteArrayOutputStream.write(HexBin.decode(new String((certificate.getExtensionValue("1.2.840.113635.100.6.32")), UTF_8).substring(4)));
+        byteArrayOutputStream.write(Hex.decode(new String((certificate.getExtensionValue("1.2.840.113635.100.6.32")), UTF_8).substring(4)));
         return MessageDigest.getInstance("SHA256", "BC")
                 .digest(byteArrayOutputStream.toByteArray());
     }
--- a/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
@@ -35,15 +35,15 @@ public class ApplePayDecrypter {
     private static final byte[] COUNTER = {0x00, 0x00, 0x00, 0x01};
     private static final byte[] APPLE_OEM = "Apple".getBytes(UTF_8);
     private static final byte[] ALG_IDENTIFIER_BYTES = "id-aes256-GCM".getBytes(UTF_8);
-
+    private static final String MERCHANT_ID_CERTIFICATE_OID = "1.2.840.113635.100.6.32";
     static {
         Security.addProvider(new BouncyCastleProvider());
     }
 
-    private byte[] privateKeyBytes;
-    private byte[] publicCertificate;
+    private final byte[] privateKeyBytes;
+    private final byte[] publicCertificate;
 
-    private ObjectMapper objectMapper;
+    private final ObjectMapper objectMapper;
 
     @Inject
     public ApplePayDecrypter(ConnectorConfiguration configuration, ObjectMapper objectMapper) {
@@ -66,8 +66,8 @@ public class ApplePayDecrypter {
             byte[] rawData = decrypt(certificate, privateKey, ephemeralKey, data);
             return objectMapper.readValue(new String(rawData, UTF_8), ApplePaymentData.class);
         } catch (Exception e) {
-            LOGGER.error("Error while trying to decrypt apple payload");
-            throw new InvalidKeyException("Error while trying to decrypt apple payload");
+            LOGGER.error("Error while trying to decrypt apple pay payload");
+            throw new InvalidKeyException("Error while trying to decrypt apple pay payload");
         }
     }
 
@@ -102,7 +102,7 @@ public class ApplePayDecrypter {
         byteArrayOutputStream.write(ALG_IDENTIFIER_BYTES);
         byteArrayOutputStream.write(APPLE_OEM);
         // Add Merchant Id
-        byteArrayOutputStream.write(Hex.decode(new String((certificate.getExtensionValue("1.2.840.113635.100.6.32")), UTF_8).substring(4)));
+        byteArrayOutputStream.write(Hex.decode(new String((certificate.getExtensionValue(MERCHANT_ID_CERTIFICATE_OID)), UTF_8).substring(4)));
         return MessageDigest.getInstance("SHA256", "BC")
                 .digest(byteArrayOutputStream.toByteArray());
     }
--- a/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
@@ -1,0 +1,124 @@
+package uk.gov.pay.connector.applepay;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Inject;
+import com.sun.org.apache.xerces.internal.impl.dv.util.HexBin;
+import org.bouncycastle.jce.provider.BouncyCastleProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.ApplePayConfig;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.util.ResponseUtil;
+
+import javax.crypto.Cipher;
+import javax.crypto.KeyAgreement;
+import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.InputStream;
+import java.nio.charset.Charset;
+import java.security.KeyFactory;
+import java.security.MessageDigest;
+import java.security.PrivateKey;
+import java.security.Security;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.security.interfaces.ECPublicKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+
+public class ApplePayDecrypter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ApplePayDecrypter.class);
+
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
+    private static final byte[] COUNTER = {0x00, 0x00, 0x00, 0x01};
+    private static final byte[] APPLE_OEM = "Apple".getBytes(UTF_8);
+    private static final byte[] ALG_IDENTIFIER_BYTES = "id-aes256-GCM".getBytes(UTF_8);
+
+    static {
+        Security.addProvider(new BouncyCastleProvider());
+    }
+
+    private byte[] privateKeyBytes;
+    private byte[] publicCertificate;
+
+    private ObjectMapper objectMapper;
+
+    @Inject
+    public ApplePayDecrypter(ConnectorConfiguration configuration, ObjectMapper objectMapper) {
+        ApplePayConfig applePayConfig = configuration.getWorldpayConfig().getApplePay();
+        this.privateKeyBytes = applePayConfig.getPrivateKey();
+        this.publicCertificate = applePayConfig.getPublicCertificate();
+        this.objectMapper = objectMapper;
+    }
+
+    public ApplePaymentData performDecryptOperation(byte[] data, byte[] ephemeralKey)  {
+        try {
+
+            PrivateKey privateKey = generatePrivateKey();
+            Certificate certificate = generateCertificate();
+
+            AsymmetricKeyVerifier verifier = new AsymmetricKeyVerifier(privateKey, certificate.getPublicKey());
+            if (!verifier.verify()) {
+                throw new InvalidKeyException("Asymmetric keys do not match!");
+            }
+            byte[] rawData = decrypt(certificate, privateKey, ephemeralKey, data);
+            return objectMapper.readValue(new String(rawData, UTF_8), ApplePaymentData.class);
+        } catch (Exception e) {
+            LOGGER.error("Error while trying to decrypt apple payload");
+            throw new InvalidKeyException("Error while trying to decrypt apple payload");
+        }
+    }
+
+    private byte[] decrypt(Certificate certificate, PrivateKey merchantPrivateKey, byte[] ephemeralPublicKeyBytes, byte[] data) throws Exception {
+        // Reconstitute Ephemeral Public Key
+        KeyFactory keyFactory = KeyFactory.getInstance("ECDH", "BC");
+        X509EncodedKeySpec encodedKeySpec = new X509EncodedKeySpec(ephemeralPublicKeyBytes);
+        ECPublicKey ephemeralPublicKey = (ECPublicKey) keyFactory.generatePublic(encodedKeySpec);
+        // Perform KeyAgreement
+        KeyAgreement agreement = KeyAgreement.getInstance("ECDH", "BC");
+        agreement.init(merchantPrivateKey);
+        agreement.doPhase(ephemeralPublicKey, true);
+        byte[] sharedSecret = agreement.generateSecret();
+
+        // Perform KDF
+        byte[] derivedSecret = performKeyDerivationFunction(((X509Certificate) certificate), sharedSecret);
+
+        // Use the derived secret to decrypt the data
+        SecretKeySpec key = new SecretKeySpec(derivedSecret, "AES");
+        byte[] ivBytes = new byte[16];
+        IvParameterSpec ivSpec = new IvParameterSpec(ivBytes);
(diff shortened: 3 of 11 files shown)


src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.pay.connector.app;
+
+
+public class ApplePayConfig extends GatewayConfig {
```
alexbishop1: Does this need to extend `GatewayConfig`?

src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
```diff
@@ -0,0 +1,15 @@
...
+    private String publicCertificate;
+    private String privateKey;
+
+    public byte[] getPublicCertificate() {
+        return publicCertificate.getBytes();
```
alexbishop1: This uses the platform default character encoding. It’s good to be explicit, so please use `publicCertificate.getBytes(StandardCharsets.UTF_8)` instead.

src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
```diff
@@ -0,0 +1,15 @@
...
+    }
+
+    public byte[] getPrivateKey() {
+        return privateKey.getBytes();
```
alexbishop1: As above: `privateKey.getBytes(StandardCharsets.UTF_8)`

src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
```diff
@@ -11,4 +13,8 @@ public String getNotificationDomain() {
     public boolean isSecureNotificationEnabled() {
         return secureNotificationEnabled;
     }
+
+    public ApplePayConfig getApplePay() {
```
alexbishop1: I imagine `getApplePayConfig()` would be better unless there’s an existing convention I‘m missing

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
+package uk.gov.pay.connector.applepay;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Inject;
+import com.sun.org.apache.xerces.internal.impl.dv.util.HexBin;
```
alexbishop1: We shouldn’t rely on `com.sun` packages because they’re JDK internals. I haven’t checked but I can imagine that this won’t continue to work on Java 11 (I know some other `com.sun.org.apache.xerces.internal` stuff doesn’t). Can you find an alternative? I imagine Apache Commons Codec or Guava has something.

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+import java.security.interfaces.ECPublicKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+
+public class ApplePayDecrypter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(ApplePayDecrypter.class);
+
+    private static final Charset UTF_8 = Charset.forName("UTF-8");
```
alexbishop1: This could be replaced by static-importing `java.nio.charset.StandardCharsets.UTF_8` and just using that everywhere

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+    private static final byte[] ALG_IDENTIFIER_BYTES = "id-aes256-GCM".getBytes(UTF_8);
+
+    static {
+        Security.addProvider(new BouncyCastleProvider());
+    }
+
+    private byte[] privateKeyBytes;
+    private byte[] publicCertificate;
+
+    private ObjectMapper objectMapper;
```
alexbishop1: Looks like these three fields can be `final`

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+            Certificate certificate = generateCertificate();
+
+            AsymmetricKeyVerifier verifier = new AsymmetricKeyVerifier(privateKey, certificate.getPublicKey());
+            if (!verifier.verify()) {
+                throw new InvalidKeyException("Asymmetric keys do not match!");
+            }
+            byte[] rawData = decrypt(certificate, privateKey, ephemeralKey, data);
+            return objectMapper.readValue(new String(rawData, UTF_8), ApplePaymentData.class);
+        } catch (Exception e) {
+            LOGGER.error("Error while trying to decrypt apple payload");
```
alexbishop1: It’s the “Apple Pay payload” 😀

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+            throw new InvalidKeyException("Error while trying to decrypt apple payload");
```
alexbishop1: We’re losing the original exception here. For example, if the `"Asymmetric keys do not match!`" `InvalidKeyException` is thrown, we won’t record that cause anywhere.

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+        byte[] derivedSecret = performKeyDerivationFunction(((X509Certificate) certificate), sharedSecret);
+
+        // Use the derived secret to decrypt the data
+        SecretKeySpec key = new SecretKeySpec(derivedSecret, "AES");
+        byte[] ivBytes = new byte[16];
+        IvParameterSpec ivSpec = new IvParameterSpec(ivBytes);
+        Cipher aesCipher = Cipher.getInstance("AES/GCM/NoPadding", "BC");
+        aesCipher.init(Cipher.DECRYPT_MODE, key, ivSpec);
+        return aesCipher.doFinal(data);
+    }
```
alexbishop1, line 99: I’m going to trust this is doing the right thing (I understand this code is basically just copied and pasted from Apple’s examples)

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private byte[] performKeyDerivationFunction(X509Certificate certificate, byte[] sharedSecret) throws Exception {
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+        byteArrayOutputStream.write(COUNTER);
+        byteArrayOutputStream.write(sharedSecret);
+        byteArrayOutputStream.write(ALG_IDENTIFIER_BYTES.length);
+        byteArrayOutputStream.write(ALG_IDENTIFIER_BYTES);
+        byteArrayOutputStream.write(APPLE_OEM);
+        // Add Merchant Id
+        byteArrayOutputStream.write(HexBin.decode(new String((certificate.getExtensionValue("1.2.840.113635.100.6.32")), UTF_8).substring(4)));
```
alexbishop1: Can we extract `"1.2.840.113635.100.6.32"` to a constant and give it a meaningful name?

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+        return MessageDigest.getInstance("SHA256", "BC")
+                .digest(byteArrayOutputStream.toByteArray());
+    }
+
+    private Certificate generateCertificate() throws Exception {
```
alexbishop1: Can we be more specific than `Exception`?

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -0,0 +1,124 @@
...
+        InputStream stream = new ByteArrayInputStream(publicCertificate);
+        CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+        Certificate certificate = certificateFactory.generateCertificate(stream);
+        stream.close();
+        return certificate;
```
alexbishop1: We can `try`-with-resources this:

```java
try (InputStream stream = new ByteArrayInputStream(publicCertificate)) {
    CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
    return certificateFactory.generateCertificate(stream);
}
```

src/main/java/uk/gov/pay/connector/applepay/ApplePaymentData.java
```diff
@@ -0,0 +1,44 @@
+package uk.gov.pay.connector.applepay;
+
+import uk.gov.pay.connector.gateway.model.AuthorisationDetails;
+
+import java.util.Map;
+
+public class ApplePaymentData implements AuthorisationDetails {
```
alexbishop1: Unless this absolutely has to extend `AuthorisationDetails`, I wouldn’t. It’s a basically useless interface (it defines no methods!) that was introduced to allow us to shoehorn authorisation-based-on-card-details and authorisation-based-on-a-3D-Secure-authentication-result into the same code path. It was a terrible idea. And I say that as one of the people who introduced it!

alexbishop1: I’m not sure how I feel that this has no constructor and is only ever created by an `ObjectMapper`. Though I can see the advantages from a code conciseness point of view, it very much ties it to the idea that it comes from a JSON representation.

Also, it’s not immediately obvious that this class is mapped from JSON, so someone could, say, decide to rename `applicationPrimaryAccountNumber` and break it because it wouldn’t match `application_primary_account_number` in the JSON any more (I’m assuming standard Jackson renaming is in play here).

So in terms of a concrete suggestion, I have two options:

1. Ponder how horrific it would be if this object were just a regular POJO with a constructor, with `ApplePayDecrypter` responsible for extracting the data from the JSON and instantiating this object via the constructor
2. Add `@JsonProperty` annotations to the fields with the JSON key names in them

src/main/java/uk/gov/pay/connector/applepay/AsymmetricKeyVerifier.java
```diff
@@ -0,0 +1,38 @@
...
+    private final PrivateKey privateKey;
+    private final PublicKey publicKey;
+
+    AsymmetricKeyVerifier(PrivateKey privateKey, PublicKey publicKey) {
+        this.privateKey = privateKey;
+        this.publicKey = publicKey;
+    }
+
+    boolean verify() throws Exception {
+        byte[] data = UUID.randomUUID().toString().getBytes();
```
alexbishop1: Another `getBytes(StandardCharsets.UTF_8)` opportunity

src/main/java/uk/gov/pay/connector/applepay/AsymmetricKeyVerifier.java
```diff
@@ -0,0 +1,38 @@
...
+        byte[] digitalSignature = signData(data);
+        return verifySig(data, digitalSignature);
+    }
+
+    private byte[] signData(byte[] data) throws Exception {
```
alexbishop1: It makes me a bit sad that we have `throws Exception` here. Can we list them out: `throws NoSuchAlgorithmException, InvalidKeyException, SignatureException`? Or are we expecting unchecked exceptions?

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
```diff
@@ -0,0 +1,96 @@
...
+    private WorldpayConfig mockWorldpayConfig;
+    @Mock
+    private ApplePayConfig mockApplePayConfig;
+    private ObjectMapper objectMapper = new ObjectMapper();
+    
+    private ApplePayDecrypter applePayDecrypter;
+    
+    private final byte[] data = Base64.decode( "0cWHZj3Py20Nxh8VrANfcXSFFaaNCJM9HpCPiUhb63GWi7+Aya0BmsKfELoK/Pp+1hvmJzO0DtkdYrtLQRnKRKUMX+KfEsQO3eKLEOaRm12qf1jgXG7HUE1r+BlrK9BC24QzyZkqYVbdTSbE8CTmDuiDSVjQi0fBwxo+MdjsWg+ap6RDlmSXVKXuGS1to5Ae/VDnwBBMuDNYJiJYSR9LWU7eO6HL6ke6+xjXcRhfxexeZ1y9XToTcDrC0M7xM3kAHkTyDV30m63MKdb7cpSV/7DVgj99AX9XrLlVJndAnBLI7jMsOFCTho86U0fJJ40XDklR8X5x43NKL+c2SimUNBMZkiZLygQSUrFD41cKb/7UIyB9c7Sk9UJmTM24FOeVt/RH2cIX+okRB6UzewVGZEFvV/PWbJqaOCWxISMjJc8HAkWa0Q1ARVKTzCS6ZgsPFZcao0Z3/j46kCxN/RYeYG7hfgrtaH8hqnvicac3khhFAU9RbjMZCmGdVzyuaxz/4SKGtDgN22y8sPsiWORi6NM+1As5nHWMgP7dO2ouI3wcuaxHADHfGm5aNQ==" );
+
+    private final byte[] ephemeralKey = Base64.decode( "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEl/XAbOgrSCupps/QbIxJ3u4QZ1PlbO5uGDD1zj/JGMoephYSEgw+63gHQHekx3T8duXN3CoYafUpuQlwOeK6/w==");
```
alexbishop1: These two can be constants, right?

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
```diff
@@ -0,0 +1,96 @@
...
...GA1UEAwwoTWVyY2hhbnQgSUQ6IG1lcmNoYW50LnJlZHRlYW0ud2Fyc2F3LmxibzETMBEGA1UECwwKTVlUVThZNURRTTETMBEGA1UECgwKQXBwbGUgSW5jLjELMAkGA1UEBhMCVVMwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAARdoBFEtnuapXFKw4DYWsW0yV4bavpdWKszkefi19AhlIRE3WSNWSn25W5tZNFjMWtLISBmqANyufx2xP19oRvyo4ICaDCCAmQwTwYIKwYBBQUHAQEEQzBBMD8GCCsGAQUFBzABhjNodHRwOi8vb2NzcC11YXQuY29ycC5hcHBsZS5jb20vb2NzcDA0LXRlc3R3d2RyY2FlY2MwHQYDVR0OBBYEFAV7nS4mNDLy1gxvOAcCS1hOgY4lMAwGA1UdEwEB/wQCMAAwHwYDVR0jBBgwFoAU1tbVWuX//cJ8NMND3r1odlw2qb4wggEdBgNVHSAEggEUMIIBEDCCAQwGCSqGSIb3Y2QFATCB/jCBwwYIKwYBBQUHAgIwgbYMgbNSZWxpYW5jZSBvbiB0aGlzIGNlcnRpZmljYXRlIGJ5IGFueSBwYXJ0eSBhc3N1bWVzIGFjY2VwdGFuY2Ugb2YgdGhlIHRoZW4gYXBwbGljYWJsZSBzdGFuZGFyZCB0ZXJtcyBhbmQgY29uZGl0aW9ucyBvZiB1c2UsIGNlcnRpZmljYXRlIHBvbGljeSBhbmQgY2VydGlmaWNhdGlvbiBwcmFjdGljZSBzdGF0ZW1lbnRzLjA2BggrBgEFBQcCARYqaHR0cDovL3d3dy5hcHBsZS5jb20vY2VydGlmaWNhdGVhdXRob3JpdHkvMEEGA1UdHwQ6MDgwNqA0oDKGMGh0dHA6Ly9jcmwtdWF0LmNvcnAuYXBwbGUuY29tL2FwcGxld3dkcmNhZWNjLmNybDAOBgNVHQ8BAf8EBAMCAygwTwYJKoZIhvdjZAYgBEIMQDU4MDJEMUM3NzRGMDk2MkY4MTEyNDhFNTM4REUzQkVGNjgwQzc5ODZCQjVCNERDRTBCNTYyNDlGMzdDQkI5NDMwCgYIKoZIzj0EAwIDRwAwRAIgTjtiYjk/BKp3V8Dg6mIlcm5FCOF06zub7Jsr6wCsvKACIH8U114DI5HnmfcNvwM4RXFDPToop4+jjMAPvidipKkg");
```
alexbishop1: Can we put the Base64 strings into constants please? It’s a bit hard to follow the code with occasional multi-thousand-character interruptions!

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
```diff
@@ -0,0 +1,96 @@
...
+        assertThat(applePaymentData.getCurrencyCode(), is("840"));
+        assertThat(applePaymentData.getDeviceManufacturerIdentifier(), is("050110030273"));
+        assertThat(applePaymentData.getPaymentDataType(), is("EMV"));
+        assertThat(applePaymentData.getTransactionAmount(), is("500"));
+        assertThat(applePaymentData.getPaymentData().get("emvData"), is("nyYIM2LtbOmXgn2fEBIBFKUAAAAAAAAAAAAAAAAAAACfNgIAA5UFAAAAAACfJwGAnzQDAQACnzcE0qE2H58CBgAAAAAFAJ8DBgAAAAAAAF8qAghAmgMWCAicAQCEB6AAAAAEEBBaCFGGAJMAgAgIXzQBAF8kAxkQMZ8aAghAggICgA=="));
+    }
+
+    @Test(expected = InvalidKeyException.class)
+    public void shouldThrowException_whenPublicCertificateIsInvalid() {
+        when(mockApplePayConfig.getPublicCertificate()).thenReturn("nope".getBytes());
```
alexbishop1: `"nope".getBytes(StandardCharsets.UTF_8)` please

alexbishop1 (changes requested):
Looks good overall but I won’t pretend to understand the cryptography stuff!

A few comments — I’ll be interested to see how this evolves as we start integrating it into our application

src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
simad: 🤦‍♀️

src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
simad: Nice catch. You're right!

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: Sure, I'll check.

simad: Bouncycastle (which is the dependency we introduced - has a hex encoder/decoder. In case, ApacheCommons does, too)

src/main/java/uk/gov/pay/connector/applepay/AsymmetricKeyVerifier.java
simad: yep

src/main/java/uk/gov/pay/connector/applepay/AsymmetricKeyVerifier.java
simad: We can. I was waiting to use this class in depth (and test out edge cases etc.) to check which exceptions are thrown during a "standard" journey. Will list them out for now and maybe revisit once I wire this class in 👍

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
simad: I also failed to spot those spaces next to those parenthesis. Oh dear

simad: Having them as a `byte[]` constant means having something like: ```[-47, -59, -121, 102, 61, -49, -53, 109, 13, -58, 31, 21, -84, 3, 95, 113, 116, -123, 21, -90, -115, 8, -109, 61, 30, -112, -113, -119, 72, 91, -21, 113, -106, -117, -65, -128, -55, -83, 1, -102, -62, -97, 16, -70, 10, -4, -6, 126, -42, 27, -26, 39, 51, -76, 14, -39, 29, 98, -69, 75, 65, 25, -54, 68, -91, 12, 95, -30, -97, 18, -60, 14, -35, -30, -117, 16, -26, -111, -101, 93, -86, 127, 88, -32, 92, 110, -57, 80, 77, 107, -8, 25, 107, 43, -48, 66, -37, -124, 51, -55, -103, 42, 97, 86, -35, 77, 38, -60, -16, 36, -26, 14, -24, -125, 73, 88, -48, -117, 71, -63, -61, 26, 62, 49, -40, -20, 90, 15, -102, -89, -92, 67, -106, 100, -105, 84, -91, -18, 25, 45, 109, -93, -112, 30, -3, 80, -25, -64, 16, 76, -72, 51, 88, 38, 34, 88, 73, 31, 75, 89, 78, -34, 59, -95, -53, -22, 71, -70, -5, 24, -41, 113, 24, 95, -59, -20, 94, 103, 92, -67, 93, 58, 19, 112, 58, -62, -48, -50, -15, 51, 121, 0, 30, 68, -14, 13, 93, -12, -101, -83, -52, 41, -42, -5, 114, -108, -107, -1, -80, -43, -126, 63, 125, 1, 127, 87, -84, -71, 85, 38, 119, 64, -100, 18, -56, -18, 51, 44, 56, 80, -109, -122, -113, 58, 83, 71, -55, 39, -115, 23, 14, 73, 81, -15, 126, 113, -29, 115, 74, 47, -25, 54, 74, 41, -108, 52, 19, 25, -110, 38, 75, -54, 4, 18, 82, -79, 67, -29, 87, 10, 111, -2, -44, 35, 32, 125, 115, -76, -92, -11, 66, 102, 76, -51, -72, 20, -25, -107, -73, -12, 71, -39, -62, 23, -6, -119, 17, 7, -91, 51, 123, 5, 70, 100, 65, 111, 87, -13, -42, 108, -102, -102, 56, 37, -79, 33, 35, 35, 37, -49, 7, 2, 69, -102, -47, 13, 64, 69, 82, -109, -52, 36, -70, 102, 11, 15, 21, -105, 26, -93, 70, 119, -2, 62, 58, -112, 44, 77, -3, 22, 30, 96, 110, -31, 126, 10, -19, 104, 127, 33, -86, 123, -30, 113, -89, 55, -110, 24, 69, 1, 79, 81, 110, 51, 25, 10, 97, -99, 87, 60, -82, 107, 28, -1, -31, 34, -122, -76, 56, 13, -37, 108, -68, -80, -5, 34, 88, -28, 98, -24, -45, 62, -44, 11, 57, -100, 117, -116, -128, -2, -35, 59, 106, 46, 35, 124, 28, -71, -84, 71, 0, 49, -33, 26, 110, 90, 53]``` which is a bit too verbose for my taste :D better ideas?

simad: I could generate the data myself and then encode it and make sure it is what we expect, which is also better. I'll try and see

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
simad: roger

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
simad: 👍

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: 👍

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: yo dawg

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: Yes, this is what I meant in the PR description with (will fix exception mapping) etc. I will make sure we bubble up what we want to bubble up properly when I actually wire the code in. If it's a burning issue right now I'll modify it

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: yes, I also quickly went through it with Ed.

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: sure

src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
```diff
@@ -11,4 +13,8 @@ public String getNotificationDomain() {
     public boolean isSecureNotificationEnabled() {
         return secureNotificationEnabled;
     }
+
+    public ApplePayConfig getApplePayConfig() {
+        return applePay;
```
alexbishop1: Can you rename the variable to `applePayConfig` to match please?

src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
simad: Renamed!

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: Done 👍

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: Indeed

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: 👍

src/main/java/uk/gov/pay/connector/applepay/ApplePaymentData.java
simad: I see. I was looking for a way to shoehorn apple pay related stuff in the same code path, too. Which proved to be quite a bad idea indeed. Will remove for now, let's see when I integrate the code how it looks like

simad: I understand completely what you mean but I am a bit ambivalent towards this. I could add JsonProperty annotations, indeed, but that would make this look like something we might expect to serialise. It is a POJO for what we are concerned, just a POJO which is currently deserialised by jackson rather than explicitly constructed.
The properties match what comes from apple (ie. `applicationPrimaryAccountNumber` is in the decrypted apple payload) - if somebody renames any property a test will break, which is ok.

I normally am quite against magic, but in this case instantiating the object explicitly feels a bit _wasteful_ , considering this is something which jackson does out of the box.

src/main/java/uk/gov/pay/connector/applepay/AsymmetricKeyVerifier.java
```diff
@@ -0,0 +1,39 @@
...
+    private final PrivateKey privateKey;
+    private final PublicKey publicKey;
+
+    AsymmetricKeyVerifier(PrivateKey privateKey, PublicKey publicKey) {
+        this.privateKey = privateKey;
+        this.publicKey = publicKey;
+    }
+
+    boolean verify() throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
+        byte[] data = UUID.randomUUID().toString().getBytes(UTF_8);
```
alexbishop1: I was going to comment on this in the first review round but then I forgot.

So here we’re generating a UUID, getting the string representation of it and then getting the UTF-8 bytes of that string and putting them in an array. That seems a little tortuous and isn’t going to give us a result where every byte is random. Is the intention here to just get an array of random bytes? We should probably just do that directly.

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
```diff
@@ -0,0 +1,104 @@
...
...8GA1UEAwwoTWVyY2hhbnQgSUQ6IG1lcmNoYW50LnJlZHRlYW0ud2Fyc2F3LmxibzETMBEGA1UECwwKTVlUVThZNURRTTETMBEGA1UECgwKQXBwbGUgSW5jLjELMAkGA1UEBhMCVVMwWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAARdoBFEtnuapXFKw4DYWsW0yV4bavpdWKszkefi19AhlIRE3WSNWSn25W5tZNFjMWtLISBmqANyufx2xP19oRvyo4ICaDCCAmQwTwYIKwYBBQUHAQEEQzBBMD8GCCsGAQUFBzABhjNodHRwOi8vb2NzcC11YXQuY29ycC5hcHBsZS5jb20vb2NzcDA0LXRlc3R3d2RyY2FlY2MwHQYDVR0OBBYEFAV7nS4mNDLy1gxvOAcCS1hOgY4lMAwGA1UdEwEB/wQCMAAwHwYDVR0jBBgwFoAU1tbVWuX//cJ8NMND3r1odlw2qb4wggEdBgNVHSAEggEUMIIBEDCCAQwGCSqGSIb3Y2QFATCB/jCBwwYIKwYBBQUHAgIwgbYMgbNSZWxpYW5jZSBvbiB0aGlzIGNlcnRpZmljYXRlIGJ5IGFueSBwYXJ0eSBhc3N1bWVzIGFjY2VwdGFuY2Ugb2YgdGhlIHRoZW4gYXBwbGljYWJsZSBzdGFuZGFyZCB0ZXJtcyBhbmQgY29uZGl0aW9ucyBvZiB1c2UsIGNlcnRpZmljYXRlIHBvbGljeSBhbmQgY2VydGlmaWNhdGlvbiBwcmFjdGljZSBzdGF0ZW1lbnRzLjA2BggrBgEFBQcCARYqaHR0cDovL3d3dy5hcHBsZS5jb20vY2VydGlmaWNhdGVhdXRob3JpdHkvMEEGA1UdHwQ6MDgwNqA0oDKGMGh0dHA6Ly9jcmwtdWF0LmNvcnAuYXBwbGUuY29tL2FwcGxld3dkcmNhZWNjLmNybDAOBgNVHQ8BAf8EBAMCAygwTwYJKoZIhvdjZAYgBEIMQDU4MDJEMUM3NzRGMDk2MkY4MTEyNDhFNTM4REUzQkVGNjgwQzc5ODZCQjVCNERDRTBCNTYyNDlGMzdDQkI5NDMwCgYIKoZIzj0EAwIDRwAwRAIgTjtiYjk/BKp3V8Dg6mIlcm5FCOF06zub7Jsr6wCsvKACIH8U114DI5HnmfcNvwM4RXFDPToop4+jjMAPvidipKkg";
```
alexbishop1: These and friends can all become `static` constants

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
alexbishop1: Looks like you forgot to do it!

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
```diff
@@ -0,0 +1,104 @@
...
+        applePayDecrypter = new ApplePayDecrypter(mockConfig, objectMapper);
+        applePayDecrypter.performDecryptOperation(
+                data,
+                ephemeralKey
+        );
+    }
+    
+    @Test(expected = InvalidKeyException.class)
+    public void shouldThrowException_whenPrivateKeyIsInvalid() {
+        when(mockApplePayConfig.getPrivateKey()).thenReturn("nope".getBytes());
```
alexbishop1: One more place for `getBytes(StandardCharsets.UTF_8)`

alexbishop1 (changes requested):
Just a couple of small things

src/main/java/uk/gov/pay/connector/app/WorldpayConfig.java
simad: :+1:

src/main/java/uk/gov/pay/connector/applepay/AsymmetricKeyVerifier.java
simad: I _assume_ the intent was that one, yes. I've replaced it with a direct `byte[]` generation, 16

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
simad: done

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
simad: they were so many :(

alexbishop1 (dismissed):

src/main/java/uk/gov/pay/connector/applepay/AsymmetricKeyVerifier.java
```diff
@@ -0,0 +1,38 @@
...
+    private final PrivateKey privateKey;
+    private final PublicKey publicKey;
+
+    AsymmetricKeyVerifier(PrivateKey privateKey, PublicKey publicKey) {
+        this.privateKey = privateKey;
+        this.publicKey = publicKey;
+    }
+
+    boolean verify() throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {
+        byte[] data = RandomUtils.nextBytes(16);
```
alexbishop1: Actually, come to think of it, should this use a cryptographically-secure random source?

simad: Done

src/main/java/uk/gov/pay/connector/gateway/util/XMLUnmarshaller.java
```diff
@@ -43,7 +44,7 @@
     }
 
     private static <T> T unmarshall(String payload, Class<T> clazz, XMLReader xmlReader) throws JAXBException {
-        InputSource inputSource = new InputSource(new ByteArrayInputStream(payload.getBytes()));
+        InputSource inputSource = new InputSource(new ByteArrayInputStream(payload.getBytes(UTF_8)));
```
alexbishop1, line 47: Completely tangential to what you’re currently doing but it would probably be more straightforward to just use a `StringReader` here:

```java
InputSource inputSource = new InputSource(new StringReader(payload));
```

alexbishop1 (approved):

PR #793, opened 2018-11-02
PP-4304 Replace `unstable` Google class

georgeracu:
## WHAT
- Replaced `Resources` from Google's library that is marked as `unstable` with Dropwizard's built in `fixture`.
Under the hoods, `fixture` is implemented using `Resources`, but it makes easier to be consistent and use the
recommended way. Also we don't have to handle exceptions thrown by `Resource` every time we use it.


mrlumbu (dismissed):
Nice

mrlumbu (approved):

PR #794, opened 2018-11-02
PP-4340 Remove CardService

kbottla:
## WHAT

Part of refactoring Card[3ds]AuthoriseService:

CardService is a superclass which has couple of methods (lockChargeForProcessing, getPaymentProviderFor) that can be moved to other relevant places. CardService itself can be removed as it will have no functionality once methods are moved out.

Changes include: 

- Moved `CardService.lockChargeForProcessing` to ChargeService
- Moved `CardService.getPaymentProviderFor` to Sub classes
- Removed `CardService` as it is no longer in use


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -239,6 +246,43 @@ private void setTransactionId(ChargeEntity chargeEntity, Optional<String> transa
...
+                    chargeEntity.getExternalId(), chargeEntity.getStatus(), getLegalStatusNames(legalStatuses));
+            throw new IllegalStateRuntimeException(chargeEntity.getExternalId());
+        }
+
+        chargeEntity.setStatus(lockingStatus);
+
+        return chargeEntity;
+    }
+
+    private String getLegalStatusNames(List<ChargeStatus> legalStatuses) {
```
IreneLau-GDS: This only gets used by line 272. Why don't we just do it as a local variable assignment before 272 gets called?

I understand that it was a method on its own before but maybe we can use this as an opportunity to get rid of a one line logic method?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -56,12 +60,23 @@
...
-        this.userNotificationService = userNotificationService;
+    public CardCaptureService(ChargeService chargeService, ChargeDao chargeDao, ChargeEventDao chargeEventDao, 
+                              PaymentProviders providers, UserNotificationService userNotificationService,
+                              Environment environment) {
+        this.chargeService = chargeService;
+        this.chargeDao = chargeDao;
+        this.chargeEventDao = chargeEventDao;
+        this.providers = providers;
+        this.metricRegistry = environment.metrics();
+         this.userNotificationService = userNotificationService;
```
IreneLau-GDS: Sorry to be pedantic... there is one extra whitespace here... 😢

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: 👍 Moved code to calling method

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
kbottla: 👍  Good spot. Formatted entire class.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
IreneLau-GDS: 👍 🙇‍♀️

IreneLau-GDS (approved):
Looks good to me! Our baby refactoring steps are getting us somewhere! Yay!

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -239,6 +246,41 @@ private void setTransactionId(ChargeEntity chargeEntity, Optional<String> transa
         });
     }
 
+    public ChargeEntity lockChargeForProcessing(ChargeEntity chargeEntity, OperationType operationType, List<ChargeStatus> legalStatuses, ChargeStatus lockingStatus) {
```
mrlumbu, line 249: this looks good. It feels like at the moment we've moved the method to `ChargeService`, but it's still being driven from the `CardAuthorisationService`, which is passing in all the relevant parameters (locking state, legal statuses etc). I would like to see if, in another PR, we can hide more of that from the authorise service. If there was a method that was just `ChargeEntity lockChargeFor(ChargeEntity charge, Operation operation)` or something, that would be cool, and the charge service can work out whether or not that is a legal move

PR #795, opened 2018-11-02
PP-4341 Remove Either

georgeracu:
## WHAT
- Remove usage of `Either` from `GatewayResponse`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayResponse.java
@@ -1,6 +1,5 @@
 package uk.gov.pay.connector.gateway.model.response;
 
-import fj.data.Either;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayResponse.java
```diff
@@ -1,63 +1,59 @@
...
     }
 
     private GatewayResponse(GatewayError error) {
-        this.response = left(error);
+        this.gatewayError = error;
     }
 
     public boolean isSuccessful() {
-        return response.isRight();
+        return baseResponse != null;
```
IreneLau-GDS, line 31: I know this class is here before... your changes are definitely making it a lot more readable for someone new like me!

I just want to say - I find this class is extremely baffling...

The assumptions: 
```
    public boolean isSuccessful() {
        return baseResponse != null;
    }

    public boolean isFailed() {
        return gatewayError != null;
    }
```

...means that we if we ever have both `baseResponse` and `gatewayError` not null... (I know what you are going to say - there is no way it could happen with what is in the class **right now** but all it takes is someone to add a `with` to set the response on top of it) then the outcome can be convoluted?

georgeracu: At the moment I don't have a better solution for this problem. The long term view is to get rid of this dual purpose class and return clear error/success response. To change it completely requires many changes in many places.

oswaldquek:
I'm not really against the concept of Either. I just hate the name and the left and right things. I wouldn't mind writing our own Monad that handles a valid and invalid response. Now before we all get scared about Monads and functional programming, we already use Monads: Optionals. for e.g. we can have an `Optional<String>.map(some transform).map(some transform) etc` and it handles the case if there was an empty in the first instance. Except for us we want to handle an error.

src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayResponse.java
georgeracu: Both constructors for this class are private and the factory method will allow only creation of error or success, you cannot have both in the same time. This limits the scope for having accidentally both cases.

IreneLau-GDS (approved):
It's definitely clearer than before.

PR #798, opened 2018-11-05
PP-4340 Make timeout test faster

mrlumbu:
Previously we checked the capture timeout behaviour using
the jersey client configured to timeout after 90000ms, then
configured the stub to timeout after 100000ms.
This commit overrides the timeout to be 5ms, and the stub
timeout to be 10ms, saving us 99990ms \o/


kbottla (approved):
🥇

IreneLau-GDS (approved):
This makes me so happy!

PR #800, opened 2018-11-05
PP-4313 Remove dependency on bouncy castle and tweak the decryption algorithm

simad:
## WHAT
 - Apparently bouncy castle is not safe to use. This means I had to tweak the encryption decryption process to use plain Java8 crypto libraries

### DIFF
--- a/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
@@ -2,8 +2,7 @@ package uk.gov.pay.connector.applepay;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
-import org.bouncycastle.jce.provider.BouncyCastleProvider;
-import org.bouncycastle.util.encoders.Hex;
+import org.apache.commons.codec.binary.Hex;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ApplePayConfig;
@@ -11,7 +10,7 @@ import uk.gov.pay.connector.app.ConnectorConfiguration;
 
 import javax.crypto.Cipher;
 import javax.crypto.KeyAgreement;
-import javax.crypto.spec.IvParameterSpec;
+import javax.crypto.spec.GCMParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -77,11 +72,11 @@ public ApplePaymentData performDecryptOperation(byte[] data, byte[] ephemeralKey
 
     private byte[] decrypt(Certificate certificate, PrivateKey merchantPrivateKey, byte[] ephemeralPublicKeyBytes, byte[] data) throws Exception {
         // Reconstitute Ephemeral Public Key
-        KeyFactory keyFactory = KeyFactory.getInstance("ECDH", "BC");
+        KeyFactory keyFactory = KeyFactory.getInstance("EC");
         X509EncodedKeySpec encodedKeySpec = new X509EncodedKeySpec(ephemeralPublicKeyBytes);
         ECPublicKey ephemeralPublicKey = (ECPublicKey) keyFactory.generatePublic(encodedKeySpec);
         // Perform KeyAgreement
-        KeyAgreement agreement = KeyAgreement.getInstance("ECDH", "BC");
+        KeyAgreement agreement = KeyAgreement.getInstance("ECDH");
```
IreneLau-GDS, line 79: KeyFactory is using just EC algo but KeyAgreement is specified to use ECDH, does it matter?

IreneLau-GDS (commented):
Do you have tests that decrypt sample messages (for the happy case and the unhappy case)?

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: That was one of my doubts - the java standard keyfactory does not support ECDH (bouncy castle does, tho) - I tried using plain EC and it decrypts stuff using the test data. Need to check if this will still hold true when using real stuff :D 

Tests: https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java

IreneLau-GDS: 🙆

IreneLau-GDS (approved):
I'm still new to the team so you might want one of the tech leads to review the changes as well...

IreneLau-GDS:
To add - According to one of the tech archs from Verify, BouncyCastle might now be okay to use. 

If you end up going down that path you should check with SecOps / IA who I heard has gotten the clearance from NCSC to use BouncyCastle.

PR #801, opened 2018-11-05
PP-4340 Move more logic into ChargeService

mrlumbu:
Changes the interface between the *AuthoriseServices
and ChargeService, so that ChargeEntity is fetched
by ChargeService, rather than having it passed in.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
@@ -80,15 +80,16 @@ public class CardAuthoriseService extends CardAuthoriseBaseService<AuthCardDetai
     @Override
     @Transactional
     public void processGatewayAuthorisationResponse(
-            ChargeEntity oldCharge,
+            String chargeExternalId,
+            ChargeStatus oldChargeStatus,
             AuthCardDetails authCardDetails,
             GatewayResponse<BaseAuthoriseResponse> operationResponse) {
 
-        Optional<String> transactionId = extractTransactionId(oldCharge.getExternalId(), operationResponse);
+        Optional<String> transactionId = extractTransactionId(chargeExternalId, operationResponse);
         ChargeStatus status = extractChargeStatus(operationResponse.getBaseResponse(), operationResponse.getGatewayError());
 
         ChargeEntity updatedCharge = chargeService.updateChargePostAuthorisation(
-                oldCharge.getExternalId(), 
+                chargeExternalId, 
                 status, 
                 transactionId,
                 extractAuth3dsDetails(operationResponse), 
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -82,48 +82,46 @@ private boolean cardBrandRequires3ds(String cardBrand) {
...
-
-            Optional<String> transactionId = operationResponse.getBaseResponse().map(BaseAuthoriseResponse::getTransactionId);
-            if (!transactionId.isPresent() || StringUtils.isBlank(transactionId.get())) {
-                logger.warn("AuthCardDetails authorisation response received with no transaction id. -  charge_external_id={}",
-                        chargeEntity.getExternalId());
-            }
-
-            ChargeStatus status = determineChargeStatus(operationResponse.getBaseResponse(), operationResponse.getGatewayError());
-
-            logger.info("Authorisation for {} ({} {}) for {} ({}) - {} .'. {} -> {}",
```
kbottla, line 97: We thought this is useful logging to find charge history, which I believe is used by support a lot and kept as is. Not sure if we log this type info anywhere else ?

mrlumbu: I put it back in

kbottla: Sorry, it was probably too early to comment. Logging should be before `updateChargePostAuthorisation` as we were logging the current status of charge and new status it is going to be in.

mrlumbu: agh

mrlumbu: ok, I have updated with a solution of sorts

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -81,49 +77,53 @@ private boolean cardBrandRequires3ds(String cardBrand) {
...
-                    auth3dsDetailsEntity, operationResponse.getSessionIdentifier(),
-                    authCardDetails);
-
-            metricRegistry.counter(String.format("gateway-operations.%s.%s.%s.authorise.result.%s",
-                    chargeEntity.getGatewayAccount().getGatewayName(),
-                    chargeEntity.getGatewayAccount().getType(),
-                    chargeEntity.getGatewayAccount().getId(),
-                    status.toString())).inc();
+    public void processGatewayAuthorisationResponse(
+            ChargeEntity oldCharge,
```
IreneLau-GDS: Instead of passing in the whole ChargeEntity updated in the original `preOp` and we are only using two fields, should we just pass in the status (a.k.a. prevStatus) and the external Id  into the original `postOp` method? (I get cringey when we pass the whole entity around...)

mrlumbu: yeah, good plan

mrlumbu: I didn't like that either

kbottla (approved):
LGTM!!

PR #802, opened 2018-11-06
[WIP] PP-4341 Separate authorise interface

mrlumbu:
In order to make payment providers more flexible, this
commit extracts an `AuthorisationProvider` interface from
the rather sprawling `PaymentProvider` interface. This allows
easier evolution of the authorisation path independently of
other paths.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/AuthorisationProviders.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/AuthorisationProviders.java
@@ -1,0 +1,31 @@
+package uk.gov.pay.connector.gateway;
+
+import uk.gov.pay.connector.gateway.epdq.EpdqPaymentProvider;
+import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/connector/gateway/AuthorisationProviders.java
```diff
@@ -0,0 +1,31 @@
...
+import uk.gov.pay.connector.gateway.sandbox.SandboxPaymentProvider;
+import uk.gov.pay.connector.gateway.smartpay.SmartpayPaymentProvider;
+import uk.gov.pay.connector.gateway.worldpay.WorldpayPaymentProvider;
+
+import javax.inject.Inject;
+import java.util.Map;
+
+import static jersey.repackaged.com.google.common.collect.Maps.newHashMap;
+
+public class AuthorisationProviders {
```
IreneLau-GDS, line 14: Oh nooooooooooooooo! Not another factory? Please tell me this is to replace the `PaymentProviders` at some point in the future. 

@kbottla and I just had a long chat about this. Maybe we should get on a call to discuss?

PR #803, opened 2018-11-06
[WIP] Pp 4304 create stripe payment gateway

oswaldquek:
## WHAT
Perform authorise operation on a stripe charge. We are making this PR based on the StripeResourceITest passing but we haven't tested hitting Stripe for real yet.

@georgeracu @oswaldquek


src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -114,9 +114,58 @@ public GatewayClient(Client client, Map<String, String> gatewayUrlMap,
         }
     }
 
+    public javax.ws.rs.core.Response postStripeRequestFor(GatewayAccountEntity account, GatewayOrder request, String path) throws GatewayError {
```
oswaldquek: This will likely be refactored in the near future to make it payment provider agnostic.

src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -114,9 +114,58 @@ public GatewayClient(Client client, Map<String, String> gatewayUrlMap,
...
+        javax.ws.rs.core.Response response = null;
+
+        String gatewayUrl = gatewayUrlMap.get(account.getType());
+
+        Stopwatch responseTimeStopwatch = Stopwatch.createStarted();
+        try {
+            logger.info("POSTing request for account '{}' with type '{}'", account.getGatewayName(), account.getType());
+            Builder requestBuilder = client.target(gatewayUrl + path)
+                    .request()
+                    .header(AUTHORIZATION, encode(
```
kbottla: we probably need different method of authorisation for stripe : `stripe api key + stripe connect account Id (stripe_account_id stored in database)` to link transaction to particular service.

mrlumbu: yep, this is a payment gateway specific thing

oswaldquek: according to the docs auth is performed by setting the `Authorization` header, i.e. `Authorization: Bearer sk_test_...etc`. i'm not sure where `stripe api key + stripe connect account Id (stripe_account_id stored in database)` is coming from?

mrlumbu: - the api key will be supplied as an environment variable (you should expose it to the app via the config)
- the account id is in the credentials column of the gatewayaccount. It is stored as a JSON blob, so you will need to figure out how this will fit with the current set up. I'm starting to think that using the Stripe client might be the way to go - all this prebuilt stuff isn't really helping

kbottla: This is example creating service level charges in stripe https://stripe.com/docs/connect/destination-charges

PR #804, opened 2018-11-06
PP-4341 Separate authorise interface

kbottla:
## WHAT
In order to make payment providers more flexible, this commit extracts an `AuthorisationProvider` interface from the rather sprawling `PaymentProvider` interface. This allows easier evolution of the authorisation path independently of other paths.

@mrlumbu @IreneLau-GDS @kbottla


kbottla:
Started https://github.com/alphagov/pay-connector/pull/806 and will revisit AuthHandler later

PR #805, opened 2018-11-06
Pp 4304 extract auth gateway req interface

georgeracu:
## WHAT

Extract AuthorisationGatewayRequest into an interface so the actual `common` implementation can be replaced with Stripe specific implementation in a different PR.


georgeracu:
Not required in the current context, therefore closing.

PR #809, opened 2018-11-08
PP-4352 Refactor CardCaptureService - doCapture

kbottla:
## WHAT

- Refactored `postOperation` method - moved ChargeDao and ChangeEntity related operations to ChargeService
- Renamed `preOperation, operation, postOperation` to `prepareChargeForCapture, capture, processGatewayCaptureResponse` respectively - inline with CardAuthoriseService
- Removed `TransactionalGatewayOperation` as it is not required


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -249,8 +252,24 @@ public ChargeEntity updateChargePost3dsAuthorisation(String chargeExternalId, Ch
...
-        
+                    chargeEntity.setStatus(nextStatus);
+                    chargeEventDao.persistChargeEventOf(chargeEntity, Optional.empty());
+
+                    //for sandbox, immediately move from CAPTURE_SUBMITTED to CAPTURED, as there will be no external notification
+                    if (chargeEntity.getPaymentGatewayName() == PaymentGatewayName.SANDBOX) {
+                        chargeEntity.setStatus(CAPTURED);
+                        ZonedDateTime gatewayEventTime = ZonedDateTime.now();
+                        chargeEventDao.persistChargeEventOf(chargeEntity, Optional.of(gatewayEventTime));
+                    }
```
mrlumbu, line 269: This is not ideal, but I think we can tolerate for now, but think about how to improve this in the future. I have some ideas, but don't want to overcomplicate this PR, which overall looks 👍

mrlumbu (approved):

PR #811, opened 2018-11-08
PP-4164 Handle missing address when creating Worldpay authorise payload

danworth:
- Modify payload xml template to skip adding 'cardAddress' element when
  address in CardEntity is null
- Add test for building payload when address is null
- Add integration test using Worldpay test account

with @danworth @stephencdaly


SandorArpa (approved):
LOKTM

PR #812, opened 2018-11-08
 PP-4164 Handle missing address when creating Smartpay authorise payload

stephencdaly:
## WHAT
Modify creation of the payload XML for a Smartpay authorisation order request to omit the address element when the address for the charge is empty in the database.

## HOW 
- Modify payload xml template to skip adding 'billingAddress' element when address in CardEntity is null
- Add test for building payload when input address is null
- Add no address scenario to the integration tests

with @danworth


SandorArpa (dismissed):
LOKTM

SandorArpa (approved):

PR #813, opened 2018-11-08
PP-4164 Handle missing address when creating ePDQ authorise payload

danworth:
- Modify payload creation to skip adding address properties to the
      payload parameter builder when the input address is null
- Add test for the parameter builder when the input address is null
- Add no address scenario to epdq integration test
- Added missing 3ds tests which mirror the functionality of the tests for the
   authorisation request without 3DS

@stephencdaly


src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3dsOrder.java
```diff
@@ -42,18 +43,25 @@
...
-                .add(ORDER_ID_KEY, templateData.getOrderId())
-                .add(OWNER_ADDRESS_KEY, concatAddressLines(templateData.getAuthCardDetails().getAddress().getLine1(),
-                        templateData.getAuthCardDetails().getAddress().getLine2()))
-                .add(OWNER_COUNTRY_CODE_KEY, templateData.getAuthCardDetails().getAddress().getCountry())
-                .add(OWNER_TOWN_KEY, templateData.getAuthCardDetails().getAddress().getCity())
-                .add(OWNER_ZIP_KEY, templateData.getAuthCardDetails().getAddress().getPostcode())
-                .add(PARAMPLUS_URL, "")
+                .add(ORDER_ID_KEY, templateData.getOrderId());
+
+        if (templateData.getAuthCardDetails().getAddress() != null) {
```
alexbishop1, line 48: I’m tempted to suggest making `AuthCardDetails.getAddress()` return an `Optional<Address>` but it might have significant ripple effects

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -86,6 +87,17 @@ public void shouldAuthoriseSuccessfully() {
         assertThat(response.isSuccessful(), is(true));
     }
 
+    @Test
+    public void shouldAuthoriseSuccessfullyWithNoAddressInRequest() {
+        setUpAndCheckThatEpdqIsUp();
+        PaymentProvider paymentProvider = getEpdqPaymentProvider();
+        AuthorisationGatewayRequest request = buildAuthorisationRequest(chargeEntity);
+        request.getAuthCardDetails().setAddress(null);
```
alexbishop1, line 136: Might be a bit nicer to have a `buildAuthorisationRequestWithoutAddress(ChargeEntity)` method instead

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -189,13 +195,17 @@ private PaymentProvider getEpdqPaymentProvider() {
         LinksConfig linksConfig = mock(LinksConfig.class);
         when(configuration.getLinks()).thenReturn(linksConfig);
         when(linksConfig.getFrontendUrl()).thenReturn("http://frontendUrl");
-        
+
+        GatewayConfig mockGatewayConfig = mock(GatewayConfig.class);
+        when(mockGatewayConfig.getUrls()).thenReturn(Collections.EMPTY_MAP);
```
alexbishop1: Please use `Collections.emptyMap()` because generics

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -189,13 +195,17 @@ private PaymentProvider getEpdqPaymentProvider() {
...
+        when(configuration.getGatewayConfigFor(EPDQ)).thenReturn(mockGatewayConfig);
```
alexbishop1: I’m confused as to why these three lines are suddenly needed? Is it because these tests are infrequently run and something has changed over the past few months that meant they no longer worked?

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -250,7 +260,9 @@ private void epdqSetupWithStatusCheck(boolean require3ds) {
             mockHistogram = mock(Histogram.class);
             mockCounter = mock(Counter.class);
             when(mockMetricRegistry.histogram(anyString())).thenReturn(mockHistogram);
-            when(mockMetricRegistry.counter(anyString())).thenReturn(mockCounter);
+            mockEnvironment = mock(Environment.class);
```
alexbishop1: Why is `mockEnvironment` now instantiated here rather than with `@Mock` annotation on the field?

alexbishop1 (changes requested):
This looks good — I’ve requested a couple of small changes and asked some questions to understand some changes in the tests.

I note that this will send attempt to send a billing address if only some required parts are there (e.g. just a postcode but nothing else) but this is no different to what happens currently: we rely on frontend to send us a complete address or none at all, so there’s no regression there.

It’s awesome that you added the missing ePDQ 3D Secure test but I think I would have preferred that to be in a separate PR.

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
stephencdaly: We changed this to keep consistency in the file. If using the @Mock annotation is the preferred way, we can change the other mocks to match

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
stephencdaly: Yes, they were broken so we had to do some extra stubbing to get them working again. These lines were present in the other payment provider tests but missing for ePDQ

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
stephencdaly: Would it be better to go the whole way and use the recently created AuthCardDetailsBuilder as opposed to all the various override methods in this class and in AuthUtils?

If so should this be another PR, as there are probably a few related places where it would be nice to use it?

alexbishop1: As discussed, look at migrating to `AuthCardDetailsBuilder` in a follow-up PR

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
alexbishop1: Cool. Great that you fixed it but I would have preferred it in a separate PR (easier to review and means we’re fixing things and adding new functionality at the same time, which reduces risk)

alexbishop1 (approved):

PR #815, opened 2018-11-08
PP-4352 Refactor CardCaptureService

kbottla:
## WHAT
Refactored CardCaptureService

- Moved ChargeEntity specific functionality to ChargeService
- Removed test `markChargeAsCaptureError_shouldIgnore_whenChargeDoesNotExist` as exception is raised if Charge is not found

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.connector.charge.service;
 
+import com.google.common.collect.ImmutableList;
 import com.google.inject.persist.Transactional;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
@@ -27,6 +28,7 @@ import uk.gov.pay.connector.charge.resource.ChargesApiResource;
 import uk.gov.pay.connector.charge.util.CorporateCardSurchargeCalculator;
 import uk.gov.pay.connector.charge.util.RefundCalculator;
 import uk.gov.pay.connector.chargeevent.dao.ChargeEventDao;
+import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
 import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
 import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
 import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
@@ -53,21 +55,36 @@ import java.util.Map;
 import java.util.Optional;
 
 import static com.google.common.collect.Lists.newArrayList;
+import static java.lang.String.format;
 import static javax.ws.rs.HttpMethod.GET;
 import static javax.ws.rs.HttpMethod.POST;
 import static javax.ws.rs.core.MediaType.APPLICATION_FORM_URLENCODED;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.connector.charge.model.ChargeResponse.aChargeResponseBuilder;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ABORTED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AWAITING_CAPTURE_REQUEST;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_APPROVED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_APPROVED_RETRY;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_READY;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_SUBMITTED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CREATED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.fromString;
 import static uk.gov.pay.connector.common.model.domain.NumbersInStringsSanitizer.sanitize;
 
 public class ChargeService {
     private static final Logger logger = LoggerFactory.getLogger(ChargeService.class);
 
     private static final List<ChargeStatus> CURRENT_STATUSES_ALLOWING_UPDATE_TO_NEW_STATUS = newArrayList(CREATED, ENTERING_CARD_DETAILS);
+
+    private static final List<ChargeStatus> IGNORABLE_CAPTURE_STATES = ImmutableList.of(
+            CAPTURE_APPROVED,
+            CAPTURE_APPROVED_RETRY,
+            CAPTURE_READY,
+            CAPTURE_SUBMITTED,
+            CAPTURED
+    );
 
     private final ChargeDao chargeDao;
     private final ChargeEventDao chargeEventDao;
@@ -118,7 +135,7 @@ public class ChargeService {
             return Optional.of(populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, chargeEntity).build());
         }).orElseGet(Optional::empty);
     }
-    
+
     @Transactional
     public void abortCharge(ChargeEntity charge) {
         charge.setStatus(AUTHORISATION_ABORTED);
@@ -244,12 +261,12 @@ public class ChargeService {
     }
 
     public ChargeEntity updateChargePost3dsAuthorisation(String chargeExternalId, ChargeStatus status,
-                                                 Optional<String> transactionId) {
+                                                         Optional<String> transactionId) {
         return chargeDao.findByExternalId(chargeExternalId).map(charge -> {
             charge.setStatus(status);
             setTransactionId(charge, transactionId);
             chargeEventDao.persistChargeEventOf(charge, Optional.empty());
-            
+
             return charge;
         }).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeExternalId));
     }
@@ -258,8 +275,7 @@ public class ChargeService {
         return chargeDao.findByExternalId(chargeId)
                 .map(chargeEntity -> {
 
-                    chargeEntity.setStatus(nextStatus);
-                    chargeEventDao.persistChargeEventOf(chargeEntity, Optional.empty());
+                    updateChargeStatus(chargeEntity, nextStatus);
 
                     //for sandbox, immediately move from CAPTURE_SUBMITTED to CAPTURED, as there will be no external notification
                     if (chargeEntity.getPaymentGatewayName() == PaymentGatewayName.SANDBOX) {
@@ -285,11 +301,6 @@ public class ChargeService {
             try {
                 chargeEntity.setStatus(operationType.getLockingStatus());
             } catch (InvalidStateTransitionException e) {
-                logger.warn("Could not transition charge {} from status {} to status {} - exception is {}",
-                        chargeEntity.getId(),
-                        chargeEntity.getStatus(),
-                        operationType.getLockingStatus(),
-                        e.getMessage());
                 if (chargeEntity.hasStatus(operationType.getLockingStatus())) {
                     throw new OperationAlreadyInProgressRuntimeException(operationType.getValue(), chargeEntity.getExternalId());
                 }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -299,6 +315,62 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
         }).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
     }
 
+    public ChargeEntity updateChargeStatus(ChargeEntity chargeEntity, ChargeStatus chargeStatus) {
+        chargeEntity.setStatus(chargeStatus);
```
oswaldquek, line 314: I'm not a fan of passing an object (chargeEntity) and a desired status into a method that mutates the object itself as it hides an important event from the calling class. @mrlumbu what are your thoughts? Also `updateChargeStatus` should be `private`.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -299,6 +315,62 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
...
+        chargeEventDao.persistChargeEventOf(chargeEntity, Optional.empty());
```
oswaldquek, line 315: i'm happy with this line being in this method as it has no side effect on the chargeEntity itself.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -299,6 +315,62 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
...
+        logger.info("{} for charge [charge_external_id={}]", chargeStatus, chargeEntity.getExternalId());
```
oswaldquek: this line was probably copied from somewhere else but I think the message could be improved, for eg `Charge with charge_external_id={} was updated from <old status> to <new status>`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -299,6 +315,62 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
...
+    }
+
+    public ChargeEntity markChargeAsEligibleForCapture(String externalId) {
+        return chargeDao.findByExternalId(externalId).map(charge -> {
+            ChargeStatus targetStatus = charge.isDelayedCapture() ? AWAITING_CAPTURE_REQUEST : CAPTURE_APPROVED;
+            ChargeStatus currentChargeStatus = fromString(charge.getStatus());
+
+            try {
+                updateChargeStatus(charge, targetStatus);
+            } catch (InvalidStateTransitionException e) {
```
oswaldquek, line 331: again i realise this code was copied from somewhere else but it's a red flag that calling `updateChargeStatus` here catches an `InvalidStateTransitionException`, whereas other places that call it don't. i don't think we need to catch it here as ChargeEntitiy.setStatus already throws it, so there's no point re-catching it. Also, ChargeEntity.setStatus's signature should be `public void setStatus(ChargeStatus targetStatus)` and not `public void setStatus(ChargeStatus targetStatus) throws InvalidStateTransitionException` as `InvalidStateTransitionException` is a RuntimeException.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -299,6 +315,62 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
...
+                logger.error("Charge with state " + currentChargeStatus + " cannot proceed to " + targetStatus +
```
oswaldquek: this logging can go into the ChargeEntity.setStatus method

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -299,6 +315,62 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
...
+            if (charge.hasStatus(IGNORABLE_CAPTURE_STATES)) {
+                logger.info("Skipping charge [charge_external_id={}] with status [{}] from marking as CAPTURE APPROVED", currentStatus, externalId);
+                return charge;
+            }
+
+            ChargeStatus targetStatus = CAPTURE_APPROVED;
+
+            try {
+                updateChargeStatus(charge, CAPTURE_APPROVED);
+            } catch (InvalidStateTransitionException e) {
```
oswaldquek, line 351: no need to catch this here

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -90,52 +73,19 @@ public ChargeEntity prepareChargeForCapture(String chargeId) {
...
-                throw new ConflictRuntimeException(charge.getExternalId(),
-                        format("attempt to perform delayed capture on charge not in %s state.", AWAITING_CAPTURE_REQUEST));
-            }
-
-            return changeChargeStatus(charge, targetStatus);
-        }).orElseThrow(() -> new ChargeNotFoundRuntimeException(externalId));
+        return chargeService.markChargeAsCaptureApproved(externalId);
     }
 
     public GatewayResponse<BaseCaptureResponse> capture(ChargeEntity chargeEntity) {
```
oswaldquek: this method can be made `private`

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
```diff
@@ -319,24 +322,30 @@ public void markChargeAsCaptureApproved_shouldThrowAnException_whenChargeIsNotFo
...
     public void markChargeAsCaptureApproved_shouldThrowAnIllegalStateRuntimeException_whenChargeIsNotInAuthorisationSuccess() {
         ChargeEntity chargeEntity = spy(createNewChargeWith("worldpay", 1L, CAPTURE_READY, "gatewayTxId"));
 
         when(mockedChargeDao.findByExternalId(chargeEntity.getExternalId())).thenReturn(Optional.of(chargeEntity));
 
+        cardCaptureService.markChargeAsEligibleForCapture(chargeEntity.getExternalId());
+    }
+
+    @Test
+    public void markChargeAsCaptureApproved_shouldLog_whenChargeIsNotInAuthorisationSuccess() {
```
oswaldquek: we can combine `markChargeAsCaptureApproved_shouldThrowAnIllegalStateRuntimeException_whenChargeIsNotInAuthorisationSuccess` and `markChargeAsCaptureApproved_shouldLog_whenChargeIsNotInAuthorisationSuccess` like so:

```
    @Test(expected = IllegalStateRuntimeException.class)
    public void markChargeAsCaptureApproved_shouldLog_whenChargeIsNotInAuthorisationSuccess() {
        ChargeEntity chargeEntity = spy(createNewChargeWith("worldpay", 1L, CAPTURE_READY, "gatewayTxId"));

        when(mockedChargeDao.findByExternalId(chargeEntity.getExternalId())).thenReturn(Optional.of(chargeEntity));

        try {
            cardCaptureService.markChargeAsEligibleForCapture(chargeEntity.getExternalId());
        } catch (IllegalStateRuntimeException e) {
            verify(mockedChargeDao).findByExternalId(chargeEntity.getExternalId());
            verifyNoMoreInteractions(mockedChargeDao);
            verify(mockAppender).doAppend(loggingEventArgumentCaptor.capture());
            throw e;
        }

    }
```
Less tests, the better :)

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
```diff
@@ -380,25 +389,13 @@ public void markChargeAsCaptureError_shouldSetChargeStatusToCaptureErrorAndWrite
...
-        verify(mockAppender).doAppend(loggingEventArgumentCaptor.capture());
+        verify(mockAppender, times(2)).doAppend(loggingEventArgumentCaptor.capture());
 
         List<LoggingEvent> logStatement = loggingEventArgumentCaptor.getAllValues();
         String expectedLogMessage = String.format("CAPTURE_ERROR for charge [charge_external_id=%s] - reached maximum number of capture attempts", charge.getExternalId());
         Assert.assertThat(logStatement.get(0).getFormattedMessage(), is(expectedLogMessage));
     }
 
-    @Test
-    public void markChargeAsCaptureError_shouldIgnore_whenChargeDoesNotExist() {
```
oswaldquek, line 391: should we not modify the test to verify an exception is thrown?

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: This is to support methods which already has `ChargeEntity` for processing. Updating charge status includes changing status and also adding event to ChargeEvent. This method extracts repeated code throughout

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: 👍

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: 👍

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: This is intentional as InvalidStateTransitionException is a run time exception and we will need to return `status 4xx` from Resource to frontend instead of `5xx`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: Agree, but it is not same message everywhere and could potentially generalise all messages https://github.com/alphagov/pay-connector/blob/4fffa1a593a4f7a379c4c0b1a2d650da428bacf4/src/main/java/uk/gov/pay/connector/webhook/service/NotificationService.java#L208

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: This is intentional as InvalidStateTransitionException is a run time exception and we will need to return `status 4xx` from Resource to frontend instead of `5xx`

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
kbottla: We already have a test to check exceptions in general

https://github.com/alphagov/pay-connector/blob/db28c7923843d3dbb7665c1bc0539b715c2b5dc2/src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java#L234

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: On closer look, don't see an impact of moving logging. Moved logging to ChargeEntity

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
kbottla: 👍

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
kbottla: Not a great fan of testing multiple aspects in single units test. But in this case it makes sense as it is testing exception and further interactions.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: We are already logging status change in ChargeEntity.setStatus. So removing logging from here.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: in my opinion we shouldn't extract common code just because we can. passing an object to a method to mutate that object's state is somewhat dangerous in my opinion. admittedly, this is java and not functional programming a la erlang/haskell but part of the reason they can handle concurrency so well (as opposed to java) is because objects are immutable and functions don't mutate state. we should adopt best practices where we can.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: 👍

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: how does `InvalidStateTransitionException` return a runtime exception? There is no exception mapper for it. i've traced the call all the way up to CardResource.authoriseCharge and can't see where it throws a 4xx as a result of `InvalidStateTransitionException` being thrown.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: see previous response about not throwing 4xx

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
oswaldquek: can you address this?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: `InvalidStateTransitionException extends IllegalStateException extends RuntimeException` which returns `5xx`. We are throwing `IllegalStateRuntimeException extends WebApplicationException` and returns `400` which is desired

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: `InvalidStateTransitionException extends IllegalStateException extends RuntimeException` which returns `5xx`. We are throwing `IllegalStateRuntimeException extends WebApplicationException` and returns `400` which is desired

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
mrlumbu: I agree in principle, but with Java you are fighting a losing battle if you try and keep things immutable, especially with database entities. At some point this entity needs to be mutated and saved - you can't make a copy of the entity with the new status and save it back in to the db. So in this case I don't have a problem with it, unless you can suggest a nicer way @oswaldquek? I hope Rich Hickey will forgive me

kbottla: In my opinion, it depends on where object is being passed around. Passing to a world outside or untrusted code will make it dangerous but don't think internal one will have an impact.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
kbottla: Sorry, I just realised I have un-staged changes locally

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: given this is a private method i guess i'm ok with this. keeping things immutable aside, it's still nicer to mutate a state from the calling code rather than passing the object to another method to do it.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: i think WebApplicationException returns a 500? https://docs.oracle.com/javaee/7/api/javax/ws/rs/WebApplicationException.html#WebApplicationException--

kbottla: WebApplicationException can return any of `4xx, 5xx` . But in our case we send `4xx` (see below for one example)  in specific cases rather than sending `500`
https://github.com/alphagov/pay-connector/blob/4fffa1a593a4f7a379c4c0b1a2d650da428bacf4/src/main/java/uk/gov/pay/connector/common/exception/IllegalStateRuntimeException.java#L10

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: ah ok i didn't see that.

oswaldquek (approved):

PR #816, opened 2018-11-08
PP-4352 - Remove generics from PaymentProvider interface

kbottla:
## WHAT
Removed generic type from PaymentProvider gateway operation signatures and replaced with relevant BaseXxx classes for the operation (authorise, capture, cancel, refund)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/PaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/PaymentProvider.java
@@ -4,10 +4,13 @@ import fj.data.Either;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability;
 import uk.gov.pay.connector.gateway.model.request.Auth3dsResponseGatewayRequest;
+import uk.gov.pay.connector.gateway.model.request.AuthorisationGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
-import uk.gov.pay.connector.gateway.model.response.BaseResponse;
+import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
+import uk.gov.pay.connector.gateway.model.response.BaseCancelResponse;
+import uk.gov.pay.connector.gateway.model.response.BaseRefundResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.usernotification.model.Notification;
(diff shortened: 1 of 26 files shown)


src/main/java/uk/gov/pay/connector/gateway/PaymentProvider.java
```diff
@@ -15,31 +15,31 @@
 
 import java.util.Optional;
 
-public interface PaymentProvider<T extends BaseResponse, R> {
+public interface PaymentProvider {
```
heathd: I think the type parameter `R` which propagates as `Notifications<R>` does actually have some significance. 

I think it would be worth keeping that for now.

mrlumbu: It's only ever `String` though?

mrlumbu: or am I missing something?

heathd: for Smartpay it's `Pair<String, Boolean>`

heathd: we could definitely replace `String` and `Pair<String, Boolean>` with concrete value-types representing the notifications from each payment gateway. That would make it clearer.

mrlumbu: ah ok, then yeah, maybe leave that bit as is, and just remove the `T`

mrlumbu: we can look at the other bit when we get to notifications

kbottla: ok, I have excluded notifications for now.

src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayResponse.java
```diff
@@ -56,47 +56,47 @@ public String toString() {
...
     }
 
-    public static <T extends BaseResponse> GatewayResponse<T> with(GatewayError gatewayError) {
+    public static GatewayResponse with(GatewayError gatewayError) {
         logger.error("Error received from gateway: {}", gatewayError);
-        return new GatewayResponse<>(gatewayError);
+        return new GatewayResponse(gatewayError);
     }
 
     public static class GatewayResponseBuilder<T extends BaseResponse> {
```
IreneLau-GDS, line 64: The type is no longer needed at the class declaration because the type T is not used anywhere. However, removing it means you'd have to touch 15 classes... (I tried, locally.)

mrlumbu (dismissed):

src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayResponse.java
kbottla: It is used now so we know `baseResponse` (of GatewayResponse) is of type that extends `BaseResponse`

IreneLau-GDS (dismissed):

IreneLau-GDS (approved):
No op approval after the rebase commit!

PR #818, opened 2018-11-09
PP-4164 Remove duplicate test fixture class for building AuthCardDetails

stephencdaly:
## WHAT
Rename AuthCardDetailsBuilder to AuthCardDetailsFixture
Remove CardFixture class and replace with usages of AuthCardDetailsFixture
Remove some test setup methods from Address business class


alexbishop1 (approved):
Yay! You removed more lines than you added!

Usually I’m a fan of doing renames of classes that are used in many places (such as `AuthCardDetailsBuilder` → `AuthCardDetailsFixture`) in their own PR but in this case the other changes are so small that it’s not hard to follow.

PR #819, opened 2018-11-12
PP-4162 - Allow no billing address in frontend card authorisation request

stephencdaly:
## WHAT
- Modify validation of request body to pass if no billing address is provided
- Modify charge service to allow persistence of card details with no billing address

with @danworth and @SandorArpa

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -361,11 +361,14 @@ public class ChargeService {
     private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails) {
         CardDetailsEntity detailsEntity = new CardDetailsEntity();
         detailsEntity.setCardBrand(sanitize(authCardDetails.getCardBrand()));
-        detailsEntity.setBillingAddress(new AddressEntity(authCardDetails.getAddress()));
         detailsEntity.setCardHolderName(sanitize(authCardDetails.getCardHolder()));
         detailsEntity.setExpiryDate(authCardDetails.getEndDate());
         detailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6)));
         detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
+
+        if (authCardDetails.getAddress() != null)
+            detailsEntity.setBillingAddress(new AddressEntity(authCardDetails.getAddress()));
+
         return detailsEntity;
     }
 
--- a/src/main/java/uk/gov/pay/connector/common/validator/AuthCardDetailsValidator.java
+++ b/src/main/java/uk/gov/pay/connector/common/validator/AuthCardDetailsValidator.java
@@ -21,22 +21,20 @@ public class AuthCardDetailsValidator {
         return isValidCardNumberLength(authCardDetails.getCardNo()) &&
                 isBetween3To4Digits(authCardDetails.getCvc()) &&
                 hasExpiryDateFormat(authCardDetails.getEndDate()) &&
-                hasAddress(authCardDetails.getAddress()) &&
+                isValidAddress(authCardDetails.getAddress()) &&
                 hasCardBrand(authCardDetails.getCardBrand()) &&
                 unlikelyToBeCvc(authCardDetails.getCardHolder()) &&
                 unlikelyToContainACardNumber(authCardDetails.getCardHolder()) &&
                 unlikelyToContainACardNumber(authCardDetails.getCardBrand()) &&
-                unlikelyToContainACardNumber(authCardDetails.getAddress().getLine1()) &&
-                unlikelyToContainACardNumber(authCardDetails.getAddress().getLine2()) &&
-                unlikelyToContainACardNumber(authCardDetails.getAddress().getCity()) &&
-                unlikelyToContainACardNumber(authCardDetails.getAddress().getCounty()) &&
-                unlikelyToContainACardNumber(authCardDetails.getAddress().getPostcode()) &&
-                unlikelyToContainACardNumber(authCardDetails.getAddress().getCountry());
+                addressUnlikelyToContainACardNumber(authCardDetails.getAddress());
     }
 
-    private static boolean hasAddress(Address address) {
-        return address != null &&
-                isNotBlank(address.getCity()) &&
+    private static boolean isValidAddress(Address address) {
+        return address == null || isAddressComplete(address);
+    }
+
+    private static boolean isAddressComplete(Address address) {
+        return isNotBlank(address.getCity()) &&
                 isNotBlank(address.getLine1()) &&
                 isNotBlank(address.getPostcode()) &&
                 isNotBlank(address.getCountry());
@@ -57,7 +55,17 @@ public class AuthCardDetailsValidator {
     private static boolean hasExpiryDateFormat(String date) {
         return notNullAndMatches(EXPIRY_DATE, date);
     }
-
+    
+    private static boolean addressUnlikelyToContainACardNumber(Address address) {
+        return address == null ||
+                (unlikelyToContainACardNumber(address.getLine1()) &&
+                        unlikelyToContainACardNumber(address.getLine2()) &&
+                        unlikelyToContainACardNumber(address.getCity()) &&
+                        unlikelyToContainACardNumber(address.getCounty()) &&
+                        unlikelyToContainACardNumber(address.getPostcode()) &&
+                        unlikelyToContainACardNumber(address.getCountry()));
+    }
+    
     private static boolean unlikelyToContainACardNumber(String field) {
         return nullOrDoesNotMatch(CONTAINS_MORE_THAN_11_NOT_NECESSARILY_CONTIGUOUS_DIGITS, field);
     }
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -361,11 +361,14 @@ public ChargeEntity markChargeAsCaptureApproved(String externalId) {
     private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails) {
         CardDetailsEntity detailsEntity = new CardDetailsEntity();
         detailsEntity.setCardBrand(sanitize(authCardDetails.getCardBrand()));
-        detailsEntity.setBillingAddress(new AddressEntity(authCardDetails.getAddress()));
         detailsEntity.setCardHolderName(sanitize(authCardDetails.getCardHolder()));
         detailsEntity.setExpiryDate(authCardDetails.getEndDate());
         detailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6)));
         detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
+
+        if (authCardDetails.getAddress() != null)
```
alexbishop1, line 369: I’m getting more convinced that making `AuthCardDetails.getAddress()` return an `Optional<Address>` is a good idea. It would make the code much nicer. We shouldn’t make that change in this PR though.

src/main/java/uk/gov/pay/connector/common/validator/AuthCardDetailsValidator.java
```diff
@@ -21,22 +21,20 @@ public static boolean isWellFormatted(AuthCardDetails authCardDetails) {
...
     }
 
-    private static boolean hasAddress(Address address) {
-        return address != null &&
-                isNotBlank(address.getCity()) &&
+    private static boolean isValidAddress(Address address) {
+        return address == null || isAddressComplete(address);
+    }
+
+    private static boolean isAddressComplete(Address address) {
```
alexbishop1: I think `isAddressComplete(Address)` can just be merged into `isValidAddress(Address)`. This would eliminate the inconsistency that some of the `Address` validation methods handle `null` while others don’t

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceITest.java
```diff
@@ -65,6 +70,26 @@ public void shouldAuthoriseChargeWithCorporateCard_ForValidAuthorisationDetails(
...
+
+        String chargeId = createNewChargeWithNoTransactionId(ENTERING_CARD_DETAILS);
+        worldpayMockClient.mockAuthorisationSuccess();
+
+        AuthCardDetails authCardDetails = AuthCardDetailsBuilder.anAuthCardDetails()
+                .withAddress(null)
+                .build();
+
+        givenSetup()
+                .body(objectMapper.writeValueAsString(authCardDetails))
```
alexbishop1: It would be more consistent with the other tests to add a new static method to `JsonRequestHelper`

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -152,6 +153,30 @@ public void doAuthoriseWithNonCorporateCard_shouldRespondAuthorisationSuccess()
         assertThat(charge.getCorporateSurcharge().isPresent(), is(false));
     }
 
+    @Test
+    public void doAuthoriseWithoutBillingAddress_shouldRespondAuthorisationSuccess() {
+
+        providerWillAuthorise();
+        AuthCardDetails authCardDetails = AuthCardDetailsBuilder.anAuthCardDetails()
+                .withAddress(null)
+                .build();
```
alexbishop1, line 162: This is inconsistent with the other tests but I’m assuming we want to gradually move towards using `AuthCardDetailsBuilder` rather than `AuthUtils` (how did we end up with so many different ways to build these objects?!)

src/test/java/uk/gov/pay/connector/resources/AuthCardDetailsValidatorTest.java
```diff
@@ -223,6 +223,13 @@ public void validationSucceedIfAddressLine2IsNull() {
         assertTrue(AuthCardDetailsValidator.isWellFormatted(authCardDetails));
     }
 
+    @Test
+    public void validationSucceedIfAddressIsNull() {
+        AuthCardDetails authCardDetails = aValidAuthorisationDetails();
+        authCardDetails.setAddress(null);
```
alexbishop1, line 221: I think it might be worth adding a `buildAuthCardDetailsWithoutAddressFor` static method that never has an address in the first place

src/test/java/uk/gov/pay/connector/resources/AuthCardDetailsValidatorTest.java
```diff
@@ -223,6 +223,13 @@ public void validationSucceedIfAddressLine2IsNull() {
...
+        assertTrue(AuthCardDetailsValidator.isWellFormatted(authCardDetails));
+    }
+
```
alexbishop1, line 224: There is a test in this class called `validationFailsForMissingAddress`, which still passes. Please look into what’s going on here

alexbishop1 (changes requested):
This looks straightforward. I’ve left a few comments but most of them are just things to consider rather than hard requests. The only showstopper relates to `validationFailsForMissingAddress` in `AuthCardDetailsValidatorTest`.

src/test/java/uk/gov/pay/connector/resources/AuthCardDetailsValidatorTest.java
stephencdaly: Ah yes, I noticed this when I was doing my other refactoring but forgot to come back to it here.
It's because there is an invalid expiry date in the input card details, so it was never getting to the address checks. We'll double check that's the case then remove it.

src/test/java/uk/gov/pay/connector/resources/AuthCardDetailsValidatorTest.java
stephencdaly: This will be replaced by using the fixture in my other refactoring, is it okay to leave it and do it there as the refactoring changes all the other tests in this class?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
stephencdaly: Yes the refactoring will get rid of the inconsistencies

src/test/java/uk/gov/pay/connector/resources/AuthCardDetailsValidatorTest.java
alexbishop1: Yeah, sure

src/test/java/uk/gov/pay/connector/resources/AuthCardDetailsValidatorTest.java
alexbishop1: Looks like `validationFailsForMissingCityAddress`, `validationFailsForMissingLine1Address`, `validationFailsForMissingCountryAddress` and `validationFailsForMissingPostcodeAddress` also have invalid expiry dates so don’t check what they think they are 🤦‍♂️

stephencdaly: We have fixed those too

src/test/java/uk/gov/pay/connector/it/JsonRequestHelper.java
```diff
@@ -95,35 +95,57 @@ public static String buildJsonAuthorisationDetailsWithFullAddress() {
...
-                                                                    String expiryDate, 
+    public static String buildJsonAuthorisationDetailsWithoutAddress() {
+        JsonObject authorisationDetails = buildJsonAuthorisationDetailsWithoutAddress(
+                CARD_HOLDER_NAME,
+                CARD_NUMBER,
+                CVC,
+                EXPIRY_DATE,
+                CARD_BRAND,
+                Boolean.TRUE,
+                null);
```
alexbishop1: I don’t think `PayersCardType` is going to be `null` realistically any more so we may as well set it to something

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceITest.java
```diff
@@ -20,12 +22,14 @@
...
 import static uk.gov.pay.connector.it.JsonRequestHelper.buildCorporateJsonAuthorisationDetailsFor;
 import static uk.gov.pay.connector.it.JsonRequestHelper.buildJsonAuthorisationDetailsFor;
+import static uk.gov.pay.connector.it.JsonRequestHelper.buildJsonAuthorisationDetailsWithoutAddress;
 
 @RunWith(DropwizardJUnitRunner.class)
 @DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml")
 public class WorldpayCardResourceITest extends ChargingITestBase {
 
     private String validAuthorisationDetails = buildJsonAuthorisationDetailsFor("4444333322221111", "visa");
+    private ObjectMapper objectMapper = new ObjectMapper();
```
alexbishop1: I don’t think you need this any more

alexbishop1 (changes requested):
Awesome, thanks for making those changes. Just two trivial comments.

alexbishop1 (approved):
Awesomesauce! Looking forward to the refactoring that will tidy this up a bit more

PR #820, opened 2018-11-12
PP-4162 Add tests for get charge requests when there is no address

stephencdaly:
## WHAT
- The response already omits the 'billing_address' field when no billing address is present so no changes were required for this
- Add tests for the get charges endpoints to ensure that when there is no address for the charge in the database, the 'billing_address' property is not present in the response

with @SandorArpa


src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -169,6 +171,27 @@ public void shouldReturnEmptyCardBrandLabel_whenChargeIsAuthorisedAndBrandUnknow
                 .body("card_details.card_brand", is(""));
     }
 
+    @Test
+    public void shouldNotReturnBillingAddress_whenNoAddressDetailsPresentInDB() {
```
alexbishop1, line 175: I’d lose `InDB`: that’s an implementation detail at this level

alexbishop1 (approved):
No new production code because it works already! 🎉 

One suggestion about test names (applies to both of them) but I’m not going to hold off approval on it

PR #822, opened 2018-11-12
PP-4312 Add endpoint for authorising e-wallet payments

simad:
## WHAT
- Adding a new endpoint in the card resource to receive e-wallets (apple pay, for now) payment data from frontend
- Using Dropwizard's built it jersey support do deserialise the payload sent by frontend containing both payment info data and encrypted data coming from the e-wallet payment provider itself
- Cardholder name and email seem to be optional coming from apple pay, so either or both of them might be null. This is ok as both are nullable in connector's db
- We are currently duplicating a bit of code to move the handling of the auth response out from the resource - this will be refactored to be in a better shape

with @cobainc0

### DIFF
--- a/src/main/java/uk/gov/pay/connector/applepay/ApplePayService.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/ApplePayService.java
@@ -36,11 +36,15 @@ public class ApplePayService {
 
         LOGGER.info("Authorising apple pay charge with id {} ", chargeId);
         GatewayResponse<BaseAuthoriseResponse> response = authoriseService.doAuthorise(chargeId, data);
-        return handleGatewayAuthoriseResponse(response);
+        if (isAuthorisationSubmitted(response)) {
+            return badRequestResponse("This transaction was deferred.");
+        }
+
+        return isAuthorisationDeclined(response) ? badRequestResponse("This transaction was declined.") : handleGatewayAuthoriseResponse(response);
     }
 
 
-    //PP-4314 This is duplicated from the CardResource. This kind of handling shouldn't be there in the first place, so will refactor to be embedded in services rather than at resource level
+    //PP-4314 These methods duplicated from the CardResource. This kind of handling shouldn't be there in the first place, so will refactor to be embedded in services rather than at resource level
     private Response handleGatewayAuthoriseResponse(GatewayResponse<? extends BaseAuthoriseResponse> response) {
         return response.getGatewayError()
                 .map(this::handleError)
@@ -48,6 +52,7 @@ public class ApplePayService {
                         .map(r -> ResponseUtil.successResponseWithEntity(ImmutableMap.of("status", r.authoriseStatus().getMappedChargeStatus().toString())))
                         .orElseGet(() -> ResponseUtil.serviceErrorResponse("InterpretedStatus not found for Gateway response")));
     }
+    
     private Response handleError(GatewayError error) {
         switch (error.getErrorType()) {
             case UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY:
@@ -59,5 +64,18 @@ public class ApplePayService {
             default:
                 return badRequestResponse(error.getMessage());
         }
+    }
+    
+    private static boolean isAuthorisationSubmitted(GatewayResponse<BaseAuthoriseResponse> response) {
+        return response.getBaseResponse()
+                .filter(baseResponse -> baseResponse.authoriseStatus() == BaseAuthoriseResponse.AuthoriseStatus.SUBMITTED)
+                .isPresent();
+    }
+
+    private static boolean isAuthorisationDeclined(GatewayResponse<BaseAuthoriseResponse> response) {
+        return response.getBaseResponse()
+                .filter(baseResponse -> baseResponse.authoriseStatus() == BaseAuthoriseResponse.AuthoriseStatus.REJECTED ||
+                        baseResponse.authoriseStatus() == BaseAuthoriseResponse.AuthoriseStatus.ERROR)
+                .isPresent();
     }
 }
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
@@ -3,6 +3,8 @@ package uk.gov.pay.connector.paymentprocessor.resource;
 import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.applepay.ApplePayService;
+import uk.gov.pay.connector.applepay.api.ApplePayToken;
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -49,12 +50,22 @@ public CardResource(CardAuthoriseService cardAuthoriseService, Card3dsResponseAu
...
     }
 
+    @POST
+    @Path("/v1/frontend/charges/{chargeId}/wallets")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    public Response authoriseCharge(@PathParam("chargeId") String chargeId, ApplePayToken jsonNode) {
+        logger.info("Received encrypted payload for charge with id {} ", chargeId);
+        return Response.ok()
+                .entity(ImmutableMap.of("status", "AUTHORISATION SUCCESS"))
```
alexbishop1: I think I would feel more comfortable if this is `AUTHORISATION ERROR` because it’s not really been successful.

However, I recognise that if it is `AUTHORISATION ERROR` then the tests will have to check for this and then be modified later.

src/test/java/uk/gov/pay/connector/it/resources/CardAuthoriseResourceITest.java
```diff
@@ -235,7 +246,18 @@ private String shouldAuthoriseChargeFor(String cardDetails) {
         assertFrontendChargeStatusIs(chargeId, AUTHORISATION_SUCCESS.getValue());
         return chargeId;
     }
+    private String shouldAuthoriseChargeForApplePay(String cardDetails) {
```
alexbishop1: The test name says `Authorise` but it checks the status is `ENTERING CARD DETAILS`

alexbishop1 (commented):
I’m uncomfortable with returning `AUTHORISATION SUCCESS` when there it wasn’t actually successful and the real status of the payment did not change. I probably would have gone down the route of getting the back-end parts working and then adding the endpoint. Has it been done in this order so that the frontend work can proceed?

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
simad: Can do. The frontend is not wired anyway, so nothing is actually sent to connector

src/test/java/uk/gov/pay/connector/it/resources/CardAuthoriseResourceITest.java
simad: Changed it, now it actually does work :D

src/main/java/uk/gov/pay/connector/applepay/ApplePayService.java
```diff
@@ -36,18 +36,23 @@ public Response authorise(String chargeId, ApplePayToken applePayToken) {
 
         LOGGER.info("Authorising apple pay charge with id {} ", chargeId);
         GatewayResponse<BaseAuthoriseResponse> response = authoriseService.doAuthorise(chargeId, data);
-        return handleGatewayAuthoriseResponse(response);
+        if (isAuthorisationSubmitted(response)) {
+            return badRequestResponse("This transaction was deferred.");
+        }
+
+        return isAuthorisationDeclined(response) ? badRequestResponse("This transaction was declined.") : handleGatewayAuthoriseResponse(response);
```
alexbishop1, line 43: Shame that this duplicates `CardResource` — one for later tidy-up, maybe

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java
```diff
@@ -112,6 +112,8 @@ public ExecutorService getExecutor() {
         } catch (ExecutionException | InterruptedException exception) {
             if (exception.getCause() instanceof WebApplicationException) {
                 throw (WebApplicationException) exception.getCause();
+            } else if (exception.getCause() instanceof UnsupportedOperationException) { //ooof
```
alexbishop1, line 115: Why do we have to do this?

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -39,22 +41,32 @@
...
         this.chargeCancelService = chargeCancelService;
+        this.applePayService = applePayService;
+    }
+
+    @POST
+    @Path("/v1/frontend/charges/{chargeId}/wallets")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    public Response authoriseCharge(@PathParam("chargeId") String chargeId, ApplePayToken applePayToken) {
+        logger.info("Received encrypted payload for charge with id {} ", chargeId);
```
alexbishop1, line 61: For searchability, having “Apple Pay” in here might be good

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceAuthorizeITest.java
```diff
@@ -233,10 +252,6 @@ private void addGatewayAccount(Map credentials) {
         databaseTestHelper.addGatewayAccount(accountId, paymentProvider, credentials);
     }
 
-    private String authoriseChargeUrlFor(String chargeId) {
-        return "/v1/frontend/charges/{chargeId}/cards".replace("{chargeId}", chargeId);
-    }
-
```
alexbishop1, line 239: I assume this was unused

alexbishop1 (approved):
This all looks straightforward enough — I have a couple of comments but nothing to withhold approval

src/main/java/uk/gov/pay/connector/applepay/ApplePayService.java
simad: yes :( will definitely refactor that out

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java
simad: I want to use the exception mapper to throw a bad request response, but because we capture the exceptions there, we return a failed promise from the executor service, which is very sad. So I need to add that line to let the proper exception bubble up so that dropwizard can do its thing. This again should be improved, i never realised we were doing this

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
simad: will change that in another pr 👍

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceAuthorizeITest.java
simad: yep

PR #824, opened 2018-11-12
Pp 4341 auth error test

oswaldquek:
## WHAT
This PR adds a test to handle authorization error from Stripe.

@oswaldquek

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
@@ -22,6 +22,7 @@ import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.gateway.stripe.json.StripeError;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.usernotification.model.Notification;
 import uk.gov.pay.connector.usernotification.model.Notifications;
@@ -37,6 +38,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.UUID;
 
 import static java.lang.String.format;
 import static java.util.stream.Collectors.joining;
@@ -57,11 +59,7 @@ public class StripePaymentProvider implements PaymentProvider<BaseResponse, Stri
                                  Environment environment,
                                  ConnectorConfiguration configuration) {
         this.stripeGatewayConfig = configuration.getStripeConfig();
-        this.client = gatewayClientFactory.createStripeGatewayClient(
-                PaymentGatewayName.STRIPE,
-                GatewayOperation.AUTHORISE,
-                environment.metrics()
-        );
+        this.client = gatewayClientFactory.createStripeGatewayClient(STRIPE, GatewayOperation.AUTHORISE, environment.metrics());
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -90,6 +84,14 @@ public GatewayResponse authorise(AuthorisationGatewayRequest request) {
                 sourcePayload(request),
                 getAuthHeaderValue(),
                 APPLICATION_FORM_URLENCODED_TYPE);
+
+        if (sourceResponse.getStatusInfo().getFamily() == Response.Status.Family.CLIENT_ERROR) {
```
kbottla, line 88: Although I like parsing error response from payment provider (which I was trying to in StripeGatewayClient) I think we should catch this exception in StripeGatewayClient as CLIENT_ERROR is possible in multiple circumstances

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceITest.java
```diff
@@ -89,26 +90,36 @@ public void authoriseCharge() {
...
+        addGatewayAccount(ImmutableMap.of("stripe_account_id", stripeAccountId));
+
+        String externalChargeId = addCharge();
+
+        given().port(testContext.getPort())
+                .contentType(JSON)
+                .body(validAuthorisationDetails)
+                .post(authoriseChargeUrlFor(externalChargeId))
+                .then()
+                .statusCode(500)
```
kbottla, line 121: Not a big deal, but I think we get 4xx (CLIENT_ERROR) for unauthorised response.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeError.java
```diff
@@ -0,0 +1,34 @@
...
+/** Example response from Stripe:
+{
+  "error": {
+    "message": "Invalid API Key provided: sk_test_************S8wL",
+    "type": "invalid_request_error"
+  }
+}
+ **/
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class StripeError {
```
kbottla, line 15: If CLIENT_ERROR handling moves to StripeGatewayClient, this might not be necessary at this time.

kbottla (changes requested):

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
oswaldquek: sorry I don't follow... haven't we agreed with @mrlumbu that Stripe specific stuff should be done in the payment provider? we need to catch the client error here and not in the GatewayClient so we can deserialize the response into a StripeError.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeError.java
oswaldquek: see previous comment

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceITest.java
oswaldquek: The status code returned to the caller should be 500, as from their point of view it's a server (i.e. pay-connector) error. As the caller doesn't provide credentials, it's pay-connector's fault, not the caller's.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeError.java
oswaldquek: why is this class unnecessary? we need this StripeError here so we can retrieve the error message and log it.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: Agree that stripe specific stuff shouldn't be there. But CLIENT_ERROR is also possible when creating charges or other operations. For example for `card_declined`, stripe returns CLIENT_ERROR. We probably need to have wrapper to handle client errors  or have some thing like `response.getStatusInfo().getFamily() == Response.Status.Family.SERVER_ERROR || response.getStatusInfo().getFamily() == Response.Status.Family.CLIENT_ERROR` to throw exception. Hope I am clear.

oswaldquek: yes i would definitely have to look at other errors that could happen, thanks for the shout. however those would go into another PR which would then make the case for a more general mechanism for catching client errors clearer.

mrlumbu: Ok, so let's leave like this for time being, and we can extract out common stripe response error handling as needed

kbottla: 👍 Sounds good to me

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeError.java
kbottla: Not any more, we need it now

kbottla (approved):
LGTM!!

oswaldquek:
thanks @kbottla

PR #827, opened 2018-11-13
PP-4313 Change decrypter test data to be 3DSecure 

simad:
## WHAT
 - We have now valid test private key / certs / ephemeral key and a payload which deserialises to a payment of type 3DSecure, which is what we are going to use in anger
 - Moreover, having those test keys will allow us to complete an end to end flow for local dev, which is 👌


src/main/java/uk/gov/pay/connector/applepay/AppleDecryptedPaymentData.java
```diff
@@ -0,0 +1,90 @@
+package uk.gov.pay.connector.applepay;
+
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import uk.gov.pay.connector.applepay.api.AppleCardExpiryDate;
+import uk.gov.pay.connector.applepay.api.AppleCardExpiryDateDeserialiser;
+import uk.gov.pay.connector.applepay.api.PaymentInfo;
+import uk.gov.pay.connector.gateway.model.AuthorisationDetails;
+
+public class AppleDecryptedPaymentData implements AuthorisationDetails {
```
alexbishop1, line 11: I assume we’ve sadly come to the conclusion that `AppleDecryptedPaymentData` has to implement `AuthorisationDetails`?

src/main/java/uk/gov/pay/connector/applepay/AppleDecryptedPaymentData.java
```diff
@@ -0,0 +1,90 @@
...
+    private PaymentInfo paymentInfo;
+    private String applicationPrimaryAccountNumber;
+    private String currencyCode;
+    private String transactionAmount;
```
alexbishop1: The amount is a string?

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -3,10 +3,12 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
 import org.apache.commons.codec.binary.Hex;
+import org.glassfish.jersey.internal.util.Base64;
```
alexbishop1: `internal` 🙀
Looks like `java.util.Base64` can be used here

src/main/java/uk/gov/pay/connector/applepay/ApplePaymentData.java
```diff
@@ -1,25 +1,41 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import uk.gov.pay.connector.applepay.api.AppleCardExpiryDate;
+import uk.gov.pay.connector.applepay.api.AppleCardExpiryDateDeserialiser;
+import uk.gov.pay.connector.applepay.api.PaymentInfo;
+import uk.gov.pay.connector.gateway.model.AuthorisationDetails;
 
-import java.util.Map;
-
-public class ApplePaymentData {
+public class ApplePaymentData implements AuthorisationDetails {
```
alexbishop1: I’m having trouble understanding what the difference is between `ApplePaymentData` and `AppleDecryptedPaymentData` is

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDate.java
```diff
@@ -0,0 +1,41 @@
+package uk.gov.pay.connector.applepay.api;
+
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+
+public class AppleCardExpiryDate {
+    private LocalDate date;
```
alexbishop1: This can probably be `final`

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDate.java
```diff
@@ -0,0 +1,41 @@
...
+    private final static DateTimeFormatter FORMAT = DateTimeFormatter.ofPattern("yyMMdd");
+
+    private static LocalDate from(String applicationExpirationDate){
+        return LocalDate.parse(applicationExpirationDate, FORMAT);
+    }
+
+    public AppleCardExpiryDate(String date) {
+        this.date = from(date);
```
alexbishop1: If this is never going to be `null`, we can use `Objects.requireNonNull` to ensure this, which would also eliminate the need for the `null` checks in the `equals` and `hashCode` methods

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDate.java
```diff
@@ -0,0 +1,41 @@
...
+        AppleCardExpiryDate that = (AppleCardExpiryDate) o;
+
+        return date != null ? date.equals(that.date) : that.date == null;
+    }
+
+    @Override
+    public int hashCode() {
+        return date != null ? date.hashCode() : 0;
+    }
+}
```
alexbishop1: What is this class giving us that we can’t just get using a `LocalDate`? Is it just the string (de)serialisation? Is it feasible to use `LocalDate` directly and just de-serialise from and serialise to the appropriate formats (which hardly look exotic) where necessary?

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDateDeserialiser.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public AppleCardExpiryDateDeserialiser(Class<?> vc) {
+        super(vc);
+    }
+
+    @Override
+    public AppleCardExpiryDate deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
+        JsonNode node = jsonParser.getCodec().readTree(jsonParser);
+        return new AppleCardExpiryDate(node.asText());
```
alexbishop1: There doesn’t seem to be much logic here: it just takes the text and relies on `AppleCardExpiryDate` to parse it. I’m surprised we can’t use something generic built in to Jackson to do that.

src/main/java/uk/gov/pay/connector/applepay/api/PaymentInfo.java
```diff
@@ -0,0 +1,54 @@
+package uk.gov.pay.connector.applepay.api;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import uk.gov.pay.connector.gateway.model.PayersCardType;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class PaymentInfo {
```
alexbishop1: This is the information we get from Apple Pay that isn’t encrypted, right? I’d prefer a less generic name than `PaymentInfo` but I can’t think of one right now. Might want to at least prefix it with `Apple` though.

src/main/java/uk/gov/pay/connector/applepay/api/PaymentInfo.java
```diff
@@ -0,0 +1,54 @@
...
+        return cardType;
+    }
+
+    @Override
+    public String toString() {
+        return "PaymentInfo{" +
+                "lastDigitsCardNumber='" + lastDigitsCardNumber + '\'' +
+                ", brand='" + brand + '\'' +
+                ", cardType=" + cardType +
+                '}';
```
alexbishop1: I assume the cardholder name and email are excluded because they’re personally-identifiable information and we don’t want that ending up in the logs if possible? It might be worth adding a comment to state that we’re doing that so someone doesn’t “helpfully” add them.

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
```diff
@@ -48,52 +56,28 @@ public void setUp() {
...
-                "nope".getBytes(UTF_8)
-        );
-    }
-    
-    @Test(expected = InvalidKeyException.class)
-    public void shouldThrowException_whenDataIsInvalid() {
-        applePayDecrypter.performDecryptOperation(
-                "nope".getBytes(UTF_8),
-                EPHEMERAL_KEY
-        );
```
alexbishop1, line 97: Do we have tests elsewhere that cover this stuff?

alexbishop1 (changes requested):
Good to see this functionality evolving towards where we want it to be

src/main/java/uk/gov/pay/connector/applepay/AppleDecryptedPaymentData.java
simad: Yes, I was hoping to avoid this but in another branch I am carrying out the work to wire it up and it seems we have to. I'll still try to avoid this as much as possible - also because this are in connector is undergoing a big refactoring, so 🤞

src/main/java/uk/gov/pay/connector/applepay/AppleDecryptedPaymentData.java
simad: `transactionAmount | number | Transaction amount.`

nope :D fixed, it is a `long`. We are not using this value, tho, for now, as we get the value from the charge entity

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: oopsy

src/main/java/uk/gov/pay/connector/applepay/ApplePaymentData.java
simad: Let me explain. Please sit down. It's a very subtle difference, I thought a lot about this - so, when the Universe was created ..
.
.
.
I just renamed the class and the old one has been resurrected by a rebase. Damn zombies.

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDate.java
simad: yep

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDate.java
simad: 👍

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDateDeserialiser.java
simad: yes I was thinking exactly this when going through the changeset, I'll try to figure this out *opens up jackson's docs*

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDate.java
simad: Nice catch - this class will be enhanced a bit when wiring things up **spoiler alert** I will introduce a getFormattedMonth() (0 padded month) and getFormattedYear() which will be used in the worldpay xml templates (instead of having this logic directly in the xml, like we do for other templates, and it's a bit 😱 )

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDateDeserialiser.java
simad: Turns up the deserialised was not even needed! Hurray. Also, shame on me

src/main/java/uk/gov/pay/connector/applepay/api/PaymentInfo.java
simad: I thought about keeping it generic because this is created by frontend and _I imagine_ we will be able to share the same info between e-wallets. I'll rename it to `ApplePaymentInfo` and then maybe rename it back when and if we introduce Google Pay and it makes sense to do so

src/main/java/uk/gov/pay/connector/applepay/api/PaymentInfo.java
simad: right 👍

src/test/java/uk/gov/pay/connector/applepay/ApplePayDecrypterTest.java
simad: I have added the tests back - I kind of assumed those cases would not happen because that payload comes from apple, and I thought it was all encrypted so hardly temper-able. It still comes from client side, so we should check for those cases.

src/main/java/uk/gov/pay/connector/applepay/api/AppleCardExpiryDate.java
alexbishop1: I’m not convinced we should be creating our own wrapper domain object just so we can add formatting convenience methods. For example, `getFormattedMonth()` could be achieved with just `localDate.format(DateTimeFormatter.ofPattern("MM"))`, which could be done immediately before passing the data to the XML template (doing it de-serialisation time as we are here feels a bit like mixing concerns).

I’m happy to trust your judgement on this but I think the bar should be pretty high for implementing our own domain model type that is — in terms of the data it can hold — identical to one of the Java built-in types.

Personally, I’d see how far I could get just using `LocalDate` but you’ve obviously got a lot more context on this and have a better idea of where this is going so I’m happy to defer to your decision here.

simad: I see and agree. I'll remove the class, let's see if things still smooth later on, I can always reintroduce it. Cheers

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
```diff
@@ -43,7 +45,7 @@
     private final byte[] publicCertificate;
 
     private final ObjectMapper objectMapper;
-
+    private final static Base64.Decoder BASE64_DECODER = Base64.getDecoder();
```
alexbishop1, line 48: I just checked if `Base64.Decoder` is threadsafe and you’re good, it is!

Also, I think we usually follow the Java Language Specification, which is `private static final` rather than `private final static`.

src/main/java/uk/gov/pay/connector/applepay/AppleDecryptedPaymentData.java
```diff
@@ -0,0 +1,93 @@
...
+    private ApplePaymentInfo paymentInfo;
+    private String applicationPrimaryAccountNumber;
+    private String currencyCode;
+    private long transactionAmount;
+    private String deviceManufacturerIdentifier;
+    private String paymentDataType;
+    private PaymentData paymentData;
+    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyMMdd")
+    @JsonDeserialize(using = LocalDateDeserializer.class)
+    private LocalDate applicationExpirationDate;
```
alexbishop1, line 21: Neat, I didn’t know it was possible to de-serialise to a `LocalDate` with a custom pattern like this

alexbishop1 (approved):
It’s awesome how the second version of this PR actually had fewer lines than the first!

src/main/java/uk/gov/pay/connector/applepay/ApplePayDecrypter.java
simad: 👍 will fix in the next PR

src/main/java/uk/gov/pay/connector/applepay/AppleDecryptedPaymentData.java
simad: yes, pretty cool. Learned a new thing myself :D

PR #830, opened 2018-11-13
PP-4164 Use AuthCardDetailsFixture instead of AuthUtils in tests

stephencdaly:
## WHAT
- Replace all usages of methods in AuthUtils to create AuthCardDetails with use of AuthCardDetailsFixture
- Remove now unused methods in AuthUtils

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
+++ b/src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
@@ -15,7 +15,6 @@ import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.app.GatewayConfig;
 import uk.gov.pay.connector.app.LinksConfig;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
-import uk.gov.pay.connector.common.model.domain.Address;
 import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayClientFactory;
 import uk.gov.pay.connector.gateway.GatewayOperation;
@@ -37,6 +36,7 @@ import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.model.domain.AuthCardDetailsFixture;
 import uk.gov.pay.connector.refund.model.domain.RefundEntity;
 import uk.gov.pay.connector.util.TestClientFactory;
 
@@ -60,7 +60,6 @@ import static uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse.
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity.Type.TEST;
 import static uk.gov.pay.connector.model.domain.ChargeEntityFixture.aValidChargeEntity;
 import static uk.gov.pay.connector.model.domain.RefundEntityFixture.userExternalId;
-import static uk.gov.pay.connector.util.AuthUtils.aValidAuthorisationDetails;
 import static uk.gov.pay.connector.util.SystemUtils.envOrThrow;
 
 @Ignore("Ignoring as this test is failing in Jenkins because it's failing to locate the certificates - PP-1707")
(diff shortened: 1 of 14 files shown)


src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -132,8 +131,11 @@ public void shouldAuthoriseSuccessfully() {
     @Test
     public void shouldAuthoriseSuccessfullyWithNoAddressInRequest() {
         setUpAndCheckThatEpdqIsUp();
-        AuthorisationGatewayRequest request = buildAuthorisationRequest(chargeEntity);
-        request.getAuthCardDetails().setAddress(null);
+        AuthCardDetails authCardDetails = AuthCardDetailsFixture.anAuthCardDetails()
+                .withAddress(null)
```
alexbishop1, line 135: It would be nice if `AuthCardDetailsFixture` had a method like `anAuthCardDetailsWithoutAddress()` so you could create an addressless `AuthCardDetails` from the get-go, since it’s a perfectly valid state for `AuthCardDetails` to be in

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -223,7 +230,8 @@ public void shouldRefundSuccessfully() {
     }
 
     private static AuthorisationGatewayRequest buildAuthorisationRequest(ChargeEntity chargeEntity) {
-        return buildAuthorisationRequest(chargeEntity, "Mr. Payment");
+        AuthCardDetails authCardDetails = AuthCardDetailsFixture.anAuthCardDetails().build();
```
alexbishop1, line 233: Have you checked this still works? We’re changing the card data here and this test actually sends stuff to ePDQ: payment providers can be quite sensitive about things like card numbers and expiry dates in their test gateways

src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
```diff
@@ -138,7 +142,11 @@ public void shouldBeAbleToSendAuthorisationRequestForMerchantWithoutAddress() th
     @Test
     public void shouldBeAbleToSendAuthorisationRequestForMerchantUsing3ds() throws Exception {
         WorldpayPaymentProvider paymentProvider = getValidWorldpayPaymentProvider();
-        AuthorisationGatewayRequest request = getCardAuthorisationRequestWithRequired3ds();
+        AuthCardDetails authCardDetails = AuthCardDetailsFixture.anAuthCardDetails()
+                .withCardHolder(MAGIC_CARDHOLDER_NAME_THAT_MAKES_WORLDPAY_TEST_REQUIRE_3DS)
```
alexbishop1, line 146: This remains my favourite cardholder name

src/test/java/uk/gov/pay/connector/model/domain/AddressFixture.java
```diff
@@ -8,7 +8,7 @@
     private String postcode = "WC2B 6NH";
     private String city = "London";
     private String county = "London";
-    private String country = "United Kingdom";
+    private String country = "GB";
```
alexbishop1, line 11: Nice catch

alexbishop1 (commented):
This looks like a nice reduction in duplication. I’m slightly disappointed that it didn’t result in a net reduction of lines.

I’m a little confused by when you chose to use `anAuthCardDetails()` as is versus when you overrode it with what was already in the test. In most cases, you’ve just used what `anAuthCardDetails()` returns, which changes the tests but I think is fine unless the test actually depends something particular about the authorisation data. However, there are a few places where you’ve ensured what the test previously used is retained (such as `EpdqOrderRequestBuilderTest`) and I’m not quite sure why. The only place I think we need to be really cautious is the tests that actually hit test payment gateways because they can be finicky about what they accept: we should definitely make sure those tests are passing if the data changes.

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
stephencdaly: Yes I've checked this still passes.

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
stephencdaly: I guess so, but then where do you stop with adding these extra methods?

stephencdaly:
For the ...OrderRequestBuilderTests, I have built with the existing specific cases because we're validating against some text/xml files, so we need the card details to exactly match the details in these files.
This is also the case for BaseEpdqPaymentProviderTest.java and WorldpayPaymentProviderTest.java.

For SmartpayPaymentProviderTest.java we aren't validating against any particular card details, so I can remove these here.

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
alexbishop1: Fundamentally, a valid `AuthCardDetails` either has an address or doesn’t, so there’s an argument that it should be easy to construct the non-address variant. That said, adding `.withAddress(null)` isn’t exactly onerous, so I’m happy to leave this as it is.

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
alexbishop1: Did you check all three providers?

stephencdaly: Yep

stephencdaly:
I've modified SmartpayPatmentProviderTest to use the builder defaults

alexbishop1:
We can, of course, modify the test XML files but I recognise this can be a pain because they can be shared between multiple tests, so I’m not too fussed for now

alexbishop1 (approved):

PR #832, opened 2018-11-13
PP-4397 Pull encrypted secrets from S3

ejrowley:
## WHAT
The secrets are encrypted in S3 with a symmetric key, the access to the
S3 is restricted but it is not considered the control point for access
to the secret.

The control point for access of the secret is access to the decryption
key.

The decryption key is stored in parameter store. So the secret has the
same control point as existing secrets stored in parameter store.

## HOW
Currently this will not change how the container is ran, the entry point is still run-with-chamber.sh

run-with-chamber.sh will by default run docker-startup.sh this bypasses the S3 workflow.

to test this code the entrypoint needs to be changed to `run-with-chamber.sh run-with-aws-credentials.sh fetch-from-pay-secrets.sh`

### DIFF
--- a/fetch-from-pay-secrets.sh
+++ b/fetch-from-pay-secrets.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -eu
+
+PREFIX=PAY_SECRETS_DECRYPT__
+
+for i in $(env | grep "^${PREFIX}")
+do
+    key=${i%=*}
+    target_key=${key/$PREFIX/}
+
+    secret=$(
+        BUCKET="${PAY_SECRETS_BUCKET}" \
+        OBJECT="${ECS_SERVICE}"/"${target_key}" \
+        ./get_bucket_contents.sh | gpg --quiet -d --batch --yes --passphrase "${!key}")
+
+    unset key
+    export ${target_key}="${secret}"
+done
+
+if [[ $# -eq 0 ]]
+then
+    exec ./docker-startup.sh
+else
+    exec "$@"
+fi
--- a/get_bucket_contents.sh
+++ b/get_bucket_contents.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -eu
+
+function hmac_sha256 () {
+  key="$1"
+  data="$2"
+  printf '%b' "$data" | openssl dgst -sha256 -mac HMAC -macopt "$key" | sed 's/^.* //'
+}
+
+service=s3
+region=eu-west-1
+
+http_method=GET
+canonical_url="/${OBJECT}"
+canonical_query_string=''
+hashed_payload=e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+
+timestamp=$(date -u +"%Y%m%dT%H%M00Z")
+date=$(date -u +"%Y%m%d")
+
+host="${BUCKET}.s3.${region}.amazonaws.com"
+
+date_key=$(hmac_sha256 "key:AWS4${AWS_SECRET_ACCESS_KEY}" "${date}")
+date_region_key=$(hmac_sha256 "hexkey:${date_key}" "${region}")
+date_region_service_key=$(hmac_sha256 "hexkey:${date_region_key}" "${service}")
+signing_key=$(hmac_sha256 "hexkey:${date_region_service_key}" "aws4_request")
+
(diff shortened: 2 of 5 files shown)


get_bucket_contents.sh
```diff
@@ -0,0 +1,50 @@
...
+date_region_service_key=$(hmac_sha256 "hexkey:${date_region_key}" "${service}")
+signing_key=$(hmac_sha256 "hexkey:${date_region_service_key}" "aws4_request")
+
+canonical_headers="host:$host\nx-amz-content-sha256:$hashed_payload\nx-amz-date:$timestamp\nx-amz-security-token:$AWS_SESSION_TOKEN\n"
+
+signed_headers='host;x-amz-content-sha256;x-amz-date;x-amz-security-token'
+
+canonical_request="${http_method}\n${canonical_url}\n${canonical_query_string}\n${canonical_headers}\n${signed_headers}\n${hashed_payload}"
+
+canonical_request_hash=$(printf '%b' "${canonical_request}" | openssl sha256 | sed 's/^.* //')
```
rhowe-gds, line 35: I think you don't need the sed here.. at least with the output of my openssl - if you were using `sha256` then yes, but openssl doesn't print a filename before the hash

run-with-aws-credentials.sh
```diff
@@ -0,0 +1,27 @@
...
+
+    regex='"'"${key}"'"\s*:\s*"([^"]*)'
+
+    if [[ ${json} =~ ${regex} ]]
+    then
+	echo ${BASH_REMATCH[1]}
+    fi
+}
+
+response=$(curl 169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI)
```
rhowe-gds, line 21: `curl --fail` and quote the URL

run-with-aws-credentials.sh
```diff
@@ -0,0 +1,27 @@
...
+
+export AWS_ACCESS_KEY_ID=$(get_value_for_key 'AccessKeyId' $response)
```
rhowe-gds, line 23: Quote `$response`

fetch-from-pay-secrets.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -eu
+
+PREFIX=PAY_SECRETS_DECRYPT__
+
+for i in $(env | grep "^${PREFIX}")
+do
+    key=${i%=*}
+    target_key=${key/$PREFIX/}
```
rhowe-gds, line 10: `${key#$PREFIX}`

fetch-from-pay-secrets.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -eu
+
+PREFIX=PAY_SECRETS_DECRYPT__
+
+for i in $(env | grep "^${PREFIX}")
+do
+    key=${i%=*}
```
rhowe-gds, line 9: I think `%%` instead of `%` - allow `=` in the value, but not the name

fetch-from-pay-secrets.sh
```diff
@@ -0,0 +1,26 @@
...
+    target_key=${key/$PREFIX/}
+
+    secret=$(
+        BUCKET="${PAY_SECRETS_BUCKET}" \
+        OBJECT="${ECS_SERVICE}"/"${target_key}" \
+        ./get_bucket_contents.sh | gpg --quiet -d --batch --yes --passphrase "${!key}")
+
+    unset key
```
rhowe-gds, line 17: meh, it's going to get overwritten at the start of the loop anyway and nothing else in the script uses this var

fetch-from-pay-secrets.sh
```diff
@@ -0,0 +1,26 @@
...
+    export ${target_key}="${secret}"
```
rhowe-gds, line 18: sh is insane

fetch-from-pay-secrets.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -eu
```
rhowe-gds, line 3: `set -euo pipefail`

get_bucket_contents.sh
ejrowley: The version of openssl within the containers did print the filename, at least at the time this was first written

whpearson:
Closing as we no longer need large secrets

PR #834, opened 2018-11-14
PP-4341 Make StripeGatewayClient more generic

mrlumbu:
The StripeGatewayClient is intended to eventually become the GatewayClient
for all gateways. For this to work, it needs to only now about 'clienty' things: urls, payloads, headers... So this commit refactors the StripeGatewayClient
so that it only takes as arguments things that make sense for a generic client
to know about. The only mildly controversial one is the metrics prefix, which
we use within the client to emit correctly namespaced metrics. I think this is
fine.

This change in the client has prompted a small refactoring of StripePaymentProvider; it now catches a checked exception from the client before deserialising
StripeErrorResponse - I think this is the right place for this to happen.

I have also taken the opportunity to extract out smaller methods to make the main
authorise method easier to read.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/GatewayClientException.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/GatewayClientException.java
@@ -1,0 +1,16 @@
+package uk.gov.pay.connector.gateway.stripe;
+
+import javax.ws.rs.core.Response;
+
+public class GatewayClientException extends Exception{
+    private final transient Response response;
+    
+    GatewayClientException(String message, Response response) {
+        super(message);
+        this.response = response;
+    }
+
+    public Response getResponse() {
+        return response;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
@@ -4,23 +4,25 @@ import com.codahale.metrics.MetricRegistry;
 import com.google.common.base.Stopwatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import uk.gov.pay.connector.gateway.model.OrderRequestType;
-import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 
 import javax.ws.rs.ProcessingException;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.MultivaluedHashMap;
 import javax.ws.rs.core.Response;
 import java.net.SocketException;
 import java.net.SocketTimeoutException;
 import java.net.URI;
 import java.net.UnknownHostException;
+import java.util.Map;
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/GatewayClientException.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.pay.connector.gateway.stripe;
+
+import javax.ws.rs.core.Response;
+
+public class GatewayClientException extends Exception{
```
oswaldquek, line 5: very pedantic but there should be a space between `Exception` and `{`. sorry!

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -38,30 +40,23 @@ public StripeGatewayClient(Client client, MetricRegistry metricRegistry) {
...
+                    .headers(new MultivaluedHashMap<>(headers))
                     .post(Entity.entity(payload, mediaType));
-
-            if (response.getStatusInfo().getFamily() == Response.Status.Family.SERVER_ERROR) {
-                logger.error("Stripe gateway returned server error: {}, for gateway url={} with type {}", response.getStatus(), url.toString(), account.getType());
-                incrementFailureCounter(metricRegistry, metricsPrefix);
-                throw new WebApplicationException("Unexpected HTTP status code " + response.getStatus() + " from gateway");
-            }
+            
+            throwIfErrorResponse(response, metricsPrefix);
```
oswaldquek, line 59: 👍 for throwing a GatewayClientException here but be careful as its currently swallowed by the `catch (Exception e)` below

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -86,43 +87,77 @@ public StatusMapper getStatusMapper() {
...
+            Response authorisationResponse = createCharge(
+                    request,
+                    sourceResponse.readEntity(StripeSourcesResponse.class).getId()
+            );
+            
+            return responseBuilder.withResponse(StripeAuthorisationResponse.of(authorisationResponse)).build();
+        } catch(GatewayClientException e) {
+            logger.error(
+                    "There was error calling Stripe. Reason: {}",
+                    e.getResponse().readEntity(StripeErrorResponse.class).getError().getMessage()
```
oswaldquek, line 112: Given we don't really know whether the stripe error response will be of the form 
```
{
  "error": {
    "message": "Invalid API Key provided: sk_test_************S8wL",
    "type": "invalid_request_error"
  }
}
``` 
or 
```
com.stripe.model.Charge JSON: {
  "failure_code": ...,
  "failure_message": ...,
  "status": "failure"
```
i wonder if it's best to just read the entity as String.class...
I guess this might become clearer following the meet with Stripe next Monday.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
mrlumbu: ah, good spot

oswaldquek (approved):
Not sure if this PR is open for more discussion but LGTM.

PR #838, opened 2018-11-15
PP-4352 StripeGatewayClient exception handling

kbottla:
## WHAT
`StripeGatewayClient` was throwing WebApplicationException for exception from payment providers which propagates to resources and result in 500 status. But we need to pass relevant exceptions as `GatewayError` so that CardAuthoriseService, CardCaptureService and so on can set the charge status appropriately.

Exception handling has now been passed to PaymentProvider which can handle and return `GatewayError` appropriate to the operation.


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -45,42 +39,24 @@ public Response postRequest(URI url,
...
+
             Response response = clientBuilder
                     .post(Entity.entity(payload, mediaType));
-            
+
             throwIfErrorResponse(response, metricsPrefix);
-            
+
             return response;
-        } catch (ProcessingException pe) {
```
oswaldquek, line 62: I think we should be explicitly throwing ProcessingException, so for e.g.
```
public Response postRequest(URI url,
                                String payload,
                                Map<String, String> headers,
                                MediaType mediaType,
                                String metricsPrefix) throws GatewayClientException, ProcessingException
```
The reason is because according to the javadoc, `clientBuilder.post(Entity.entity(payload, mediaType))` throws a ProcessingException so we should explicitly handle that.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -45,42 +39,24 @@ public Response postRequest(URI url,
...
-                    throw new WebApplicationException("Gateway connection socket error");
-                }
-            }
-            logger.error(format("Exception for gateway url=%s", url.toString()), pe);
-            throw new WebApplicationException(pe.getMessage());
         } catch (RuntimeException e) {
             metricRegistry.counter(metricsPrefix + ".failures").inc();
             logger.error(format("Exception for gateway url=%s", url.toString()), e);
-            throw new WebApplicationException(e.getMessage());
+            throw new GatewayClientRuntimeException(url.toString(), e);
```
oswaldquek, line 58: As mentioned before, I believe we should explicitly throw ProcessingException as that is what the javadoc for `clientBuilder.post(Entity.entity(payload, mediaType))` says. This catching of any other RuntimeExceptions are for anything else that we do not know about, so I think it should revert to throwing `WebApplicationException` here.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -58,8 +64,11 @@ public GatewayResponse capture(CaptureGatewayRequest request) {
                     .readEntity(StripeErrorResponse.class)
                     .getError()
                     .getMessage());
-            
+
             return responseBuilder.withGatewayError(gatewayError).build();
+        } catch (GatewayClientRuntimeException e) {
+            return responseBuilder.withGatewayError(processGatewayException(e)).build();
```
oswaldquek: What are your thoughts about overloading the responseBuilder's `withGatewayError` method with an exception? so for example `responseBuilder.withGatewayError(e)`.

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
oswaldquek: It would make the ResponseBuilder more "clever".

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
kbottla: It can throw ProcessingException and ResponseProcessingException which are RuntimeExceptions. So catching RuntimeException and throw custom one (GatewayClientRuntimeException) as required.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
kbottla: It can throw ProcessingException and ResponseProcessingException which are RuntimeExceptions. So catching RuntimeException and throw custom one (GatewayClientRuntimeException) as required.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
kbottla: Sounds good but I think additional processing in ResponseBuilder will add multiple responsibilities to it. May be in `GatewayError` as it more related to returning this instance ?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
oswaldquek: That makes me feel uncomfortable because any of the following calls can throw RuntimeException: `client.target`, `client.target(url.toString()).request()`, `Entity.entity`, `url.toString()` etc. But obviously from the code we are interested in catching ProcessingException explicitly because they map to various GatewayErrors (https://github.com/alphagov/pay-connector/blob/490ec960b43a856d893565cdaffb5c0de357de74/src/main/java/uk/gov/pay/connector/gateway/model/GatewayError.java).
We shouldn't even have to catch other RuntimeExceptions really, as these will automatically throw a 500 response back and the stack trace should be logged.

kbottla: We need to catch all possible exceptions so we can update charge status accordingly. We map error to status even when it is not ProcessingException.
ProcessingException - helps when we can't reach payment providers and if we want to retry the operation ( like Capture retry). In other cases status is mapped to something like CAPTURE_ERROR

oswaldquek (approved):

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
oswaldquek: ok as I understand now GatewayClient is catching all exceptions so am ok with this.

PR #839, opened 2018-11-15
PP-4341: Handle a card authorisation failure

oswaldquek:
Technically a 400 is a response to a malformed syntax but for consistentcy with
    the current code in CardResource.authoriseCharge:

```
return isAuthorisationDeclined(response) ? badRequestResponse("This transaction was declined.")
```

a relevant test
    (StripeResourceITest.cardAuthorisationFailureShouldReturnBadRequest) has been
    added to handle a card auth failure (this covers the scenarios in
    https://stripe.com/docs/testing#cards-responses).

From manual testing, auth failure responses from Stripe have a 402 status code and a json body that
looks like
```
    {
      "error": {
        "charge": "ch_1DYXEIHj08j2jFuBsOaFNQ7W",
        "code": "expired_card",
        "doc_url": "https://stripe.com/docs/error-codes/expired-card",
        "message": "Your card has expired.",
        "param": "exp_month",
        "type": "card_error"
      }
    }
```

I've also taken the opportunity to improve the logging by logging the relevant external charge id as well.

@oswaldquek

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/DownstreamException.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/DownstreamException.java
@@ -1,4 +1,23 @@
 package uk.gov.pay.connector.gateway.stripe;
 
-public class DownstreamException {
+/**
+ * This class represents 5xx exceptions
+ */
+public class DownstreamException extends Exception {
+    private final int statusCode;
+    private final String jsonErrorEntity;
+
+    public DownstreamException(int statusCode, String jsonErrorEntity) {
+        this.statusCode = statusCode;
+        this.jsonErrorEntity = jsonErrorEntity;
+    }
+
+    public int getStatusCode() {
+        return statusCode;
+    }
+
+    @Override
+    public String getMessage() {
+        return jsonErrorEntity;
+    }
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/DownstreamException.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/DownstreamException.java
@@ -1,0 +1,4 @@
+package uk.gov.pay.connector.gateway.stripe;
+
+public class DownstreamException {
+}
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
@@ -13,7 +13,6 @@ import uk.gov.pay.connector.gateway.CaptureHandler;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.PaymentProvider;
 import uk.gov.pay.connector.gateway.StatusMapper;
-import uk.gov.pay.connector.gateway.model.ErrorType;
 import uk.gov.pay.connector.gateway.model.GatewayError;
 import uk.gov.pay.connector.gateway.model.request.Auth3dsResponseGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
@@ -22,9 +21,10 @@ import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
 import uk.gov.pay.connector.gateway.stripe.handler.StripeCaptureHandler;
+import uk.gov.pay.connector.gateway.stripe.json.StripeCreateChargeResponse;
+import uk.gov.pay.connector.gateway.stripe.json.StripeErrorResponse;
 import uk.gov.pay.connector.gateway.stripe.json.StripeSourcesResponse;
 import uk.gov.pay.connector.gateway.stripe.json.StripeTokenResponse;
-import uk.gov.pay.connector.gateway.stripe.response.StripeErrorResponse;
 import uk.gov.pay.connector.gateway.stripe.util.StripeAuthUtil;
 import uk.gov.pay.connector.gateway.util.DefaultExternalRefundAvailabilityCalculator;
 import uk.gov.pay.connector.gateway.util.ExternalRefundAvailabilityCalculator;
@@ -50,6 +50,8 @@ import static java.util.stream.Collectors.joining;
 import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
 import static javax.ws.rs.core.MediaType.APPLICATION_FORM_URLENCODED_TYPE;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.STRIPE;
+import static uk.gov.pay.connector.gateway.model.ErrorType.GENERIC_GATEWAY_ERROR;
+import static uk.gov.pay.connector.gateway.model.ErrorType.UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY;
 
 
 @Singleton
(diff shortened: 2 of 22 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -105,18 +106,29 @@ public GatewayResponse authorise(AuthorisationGatewayRequest request) {
...
-            logger.error(
-                    "There was error calling Stripe. Reason: {}",
-                    e.getResponse().readEntity(StripeErrorResponse.class).getError().getMessage()
-            );
+            
+            Response response = e.getResponse();
+            int status = response.getStatus();
+            
+            switch (status) {
+                case 400:
```
kbottla: We may not need this here as 400 can be due to incorrect payload and other error are given below:
https://github.com/stripe/stripe-java/blob/6cd60ea620c02044e06de85be700a165f923a320/src/main/java/com/stripe/net/LiveStripeResponseGetter.java#L699

In case if authorisation declined / payment failed, I think stripe returns charge with `outcome` populated and status 2xx (a question to confirm with stripe team)
https://stripe.com/docs/declines

oswaldquek: it may very well be a 402, as that throws a CardException https://github.com/stripe/stripe-java/blob/6cd60ea620c02044e06de85be700a165f923a320/src/main/java/com/stripe/exception/CardException.java
but we should confirm with the stripe team on these things.

oswaldquek: maybe this is one of the reasons to use the Stripe client, as we don't have to worry about what json is returned...

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -105,18 +106,29 @@ public GatewayResponse authorise(AuthorisationGatewayRequest request) {
...
+                    StripeAuthorisationResponse stripeAuthResponse = new StripeAuthorisationResponse(response.readEntity(StripeCreateChargeResponse.class));
+                    logger.error("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}.",
+                            request.getChargeExternalId(), stripeAuthResponse.getErrorCode(), stripeAuthResponse.getErrorMessage());
```
georgeracu: Given that there are actually 3 calls that potentially throw the same Exception type, when calling `stripeAuthResponse.` you might get a `NullPointerException` as this `GatewayClientException` could have bee thrown creating the token or the source. It feels like we need to handle each of this calls into their own `try catch` blocks so we can handle exceptions for each one. Probably a good point to ask ourselves if using Stripe's client is more useful that us doing all this work 🤔

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
oswaldquek: After doing manual testing with error scenarios as documented in https://stripe.com/docs/testing#cards-responses, I can confirm the error code returned is 402 and the response looks something like
```
{
  "error": {
    "charge": "ch_1DYXGiHj08j2jFuB0QNS2Qm9",
    "code": "processing_error",
    "doc_url": "https://stripe.com/docs/error-codes/processing-error",
    "message": "An error occurred while processing your card. Try again in a little bit.",
    "type": "card_error"
  }
}
```

src/main/java/uk/gov/pay/connector/gateway/stripe/DownstreamException.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * This class represents 5xx exceptions
+ */
+public class DownstreamException extends Exception {
+    private final String message;
+    private final String errorEntity;
+
+    public DownstreamException(String message, String errorEntity) {
+        this.message = message;
```
kbottla: Can it be `super(message)`?  A call on exception e.getMessage() not likely to return anything

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeCreateChargeResponse.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.connector.gateway.stripe.json;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class StripeCreateChargeResponse {
```
kbottla, line 7: Can we please name it to StripeChargeResponse to be consistent with all other response object naming

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -111,26 +114,44 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+
+            if (status == HttpStatus.SC_UNAUTHORIZED) {
+                return unexpectedGatewayResponse(request.getChargeExternalId(), responseBuilder, e.getMessage());
+            }
+
+            StripeErrorResponse stripeErrorResponse = response.readEntity(StripeErrorResponse.class);
+            String errorId = UUID.randomUUID().toString();
+            logger.error("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
+                    request.getChargeExternalId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
+            GatewayError gatewayError = new GatewayError(stripeErrorResponse.getError().getMessage(), GENERIC_GATEWAY_ERROR);
```
kbottla: This works, but we have static factory methods for specific use cases. `GatewayError.baseError(msg)` ?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -111,26 +114,44 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
             return responseBuilder.withGatewayError(GatewayError.of(e)).build();
         }
     }
 
-    private Response create3dsSource(CardAuthorisationGatewayRequest request, String sourceId) throws GatewayClientException, GatewayException {
+    private GatewayResponse unexpectedGatewayResponse(String chargeExternalId, GatewayResponse.GatewayResponseBuilder<BaseResponse> responseBuilder, String errorMessage) {
+        String errorId = UUID.randomUUID().toString();
+        logger.error("Authorisation failed for charge {}. Reason: {}. ErrorId: {}",
+                chargeExternalId, errorMessage, errorId);
+        GatewayError gatewayError = new GatewayError("There was an internal server error. ErrorId: " + errorId, UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY);
```
kbottla: This works, but we have static factory methods for specific use cases. `GatewayError.unexpectedStatusCodeFromGateway(msg)` ?

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -54,15 +64,25 @@ public GatewayResponse capture(CaptureGatewayRequest request) {
...
-                    .getError()
-                    .getMessage());
-
+            
+            Response response = e.getResponse();
+            StripeErrorResponse stripeErrorResponse = response.readEntity(StripeErrorResponse.class);
+            String errorId = UUID.randomUUID().toString();
+            logger.error("Capture failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
+                    request.getTransactionId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
+            GatewayError gatewayError = new GatewayError(stripeErrorResponse.getError().getMessage(), GENERIC_GATEWAY_ERROR);
```
kbottla: Same as above with using static factory methods.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -54,15 +64,25 @@ public GatewayResponse capture(CaptureGatewayRequest request) {
...
+        
         } catch (GatewayException e) {
+        
             return responseBuilder.withGatewayError(GatewayError.of(e)).build();
+        
+        } catch (DownstreamException e) {
+            String errorId = UUID.randomUUID().toString();
+            logger.error("Capture failed for transaction id {}. Reason: {}. ErrorId: {}",
+                    request.getTransactionId(), e.getMessage(), errorId);
+            GatewayError gatewayError = new GatewayError("An internal server error occurred. ErrorId: " + errorId, UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY);
```
kbottla: Same as above with using static factory methods.

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceITest.java
```diff
@@ -89,6 +89,23 @@ public void setup() {
...
+
+        String externalChargeId = addCharge();
+
+        given().port(testContext.getPort())
+                .contentType(JSON)
+                .body(validAuthorisationDetails)
+                .post(authoriseChargeUrlFor(externalChargeId))
+                .then()
+                .statusCode(400)
+                .body("message", is("Your card has expired."));
```
kbottla, line 106: Not sure how this message gets propagated to frontend. Do we present as is to the user in frontend?

kbottla (changes requested):
It looks good overall 👍. Very minor comments/suggestions.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeCreateChargeResponse.java
oswaldquek: I'd prefer it to be StripeChargeResponse as it makes it clear the object is meant to be deserialized from the action of creating a charge. It would make it confusing in the future when we implement refund and cancel if this object were named StripeChargeResponse.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
oswaldquek: Personally I'd like to keep it `new GatewayError(stripeErrorResponse.getError().getMessage(), GENERIC_GATEWAY_ERROR);` as it is very readable in the sense that I know we're returning a `GENERIC_GATEWAY_ERROR` type and I don't have to go into the `GatewayError.baseError` method to ascertain the error type. Also if I'm not meant to use the `new GatewayError` constructor it shouldn't be there for me to use...

src/main/java/uk/gov/pay/connector/gateway/stripe/DownstreamException.java
oswaldquek: good spot, will change

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeCreateChargeResponse.java
kbottla: 👍

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: In my opinion, this is probably leading to multiple approaches to the way we construct GatewayError in codebase and also have to update code throughout if we decide to change `GENERIC_GATEWAY_ERROR` to something else.

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceITest.java
oswaldquek: It's up to the self-service and publicapi to decide how they would want to deal with errors and it's not really our concern here.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
oswaldquek: If we want to lock down how we construct GatewayError that's probably a discussion for the wider team and would be out of scope for this PR. Refactoring `GENERIC_GATEWAY_ERROR` wouldn't be hard either as IntelliJ will do it nicely for us.

oswaldquek: but i agree it would be worth bringing this up at a code quality meeting or something!

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceITest.java
kbottla: The reason for asking is - 
If this is used as is, we may end up showing error stripe specific error messages received to the users. It is fine, if not used.
Don't think it is used as is by frontend which is fine

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
oswaldquek: i have changed it to use the static method :) i didn't make the GatewayError constructor private though as it would introduce a lot of code changes to this PR. that can be done in a separate PR.

kbottla: 👍

kbottla (approved):

PR #842, opened 2018-11-16
PP-4314 Add worldpay templates and test secrets to config

simad:
## WHAT
- Splitting https://github.com/alphagov/pay-connector/pull/841 up, because it's a bit of a monster (😱 )
- Adding fake secrets for our tests. The ones in test-it-config are valid will allow us to run proper decryption tests / run local payment journeys
- Adding xml templates for worldpay
- Decoding private key and certificate in the decrypter, so retrieving them as plain strings from the config

with @cobainc0


brid (approved):
Looks fine to me. Approved!

PR #843, opened 2018-11-16
PP-4352 StripeGatewayClient - Checked Exception

kbottla:
## WHAT

`StripeGatewayClient` is throwing RuntimeException, for which gateway operations can get away without handing the exception and may result in unexpected behaviour. Throwing checked exception will enforce gateway operations to catch and handle exception appropriately.


georgeracu (approved):
LGTM :)

PR #849, opened 2018-11-19
PP-4314 Split the AuthorisationGatewayRequest between Card and Apple Pay

simad:
## WHAT
- We are about to introduce an authorisation service which deals with apple pay (and probably other wallets, in the future) payments. We need to authorise a request with different data (ie. there's no card number but a token number) so we are introducing a separate gateway request which will be use to build the correct xml template to be sent to Worldpay.


src/main/java/uk/gov/pay/connector/applepay/AuthorisationApplePayGatewayRequest.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.connector.applepay;
+
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.gateway.model.request.BaseAuthorisationGatewayRequest;
+
+public class AuthorisationApplePayGatewayRequest extends BaseAuthorisationGatewayRequest {
```
alexbishop1: It’s really an `ApplePayAuthorisationGatewayRequest`

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
```diff
@@ -1,53 +1,22 @@
...
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
-import uk.gov.pay.connector.charge.util.CorporateCardSurchargeCalculator;
-import uk.gov.pay.connector.gateway.GatewayOperation;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
-import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 
-import java.util.Optional;
-
-public class AuthorisationGatewayRequest implements GatewayRequest {
+public class AuthorisationGatewayRequest extends BaseAuthorisationGatewayRequest {
```
alexbishop1: `CardAuthorisationGatewayRequest`?

src/main/java/uk/gov/pay/connector/gateway/model/request/BaseAuthorisationGatewayRequest.java
```diff
@@ -0,0 +1,46 @@
...
+    public String getAmount() {
+        return String.valueOf(CorporateCardSurchargeCalculator.getTotalAmountFor(charge));
+    }
+
+    public String getDescription() {
+        return charge.getDescription();
+    }
+
+    public String getChargeExternalId() {
+        return String.valueOf(charge.getExternalId());
```
alexbishop1: Isn’t this already a `String`?

src/test/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequestTest.java
```diff
@@ -16,7 +16,7 @@ public void shouldReturnTotalAmount_whenThereIsACorporateSurcharge() {
                 .withAmount(2000L)
                 .withCorporateSurcharge(250L)
                 .build();
-        AuthorisationGatewayRequest gatewayRequest = new AuthorisationGatewayRequest(chargeEntity, null);
+        BaseAuthorisationGatewayRequest gatewayRequest = new AuthorisationGatewayRequest(chargeEntity, null);
```
alexbishop1: Not convinced declaring the abstract supertype is aiding clarity here

alexbishop1 (changes requested):
This looks like a nice improvement.

The only showstopper comment is a bit where we’re doing `String.valueOf(…)` on a `String`.

I’ve left a few comments about the naming. We might want to consider renaming `AuthorisationGatewayRequest` to `CardAuthorisationGatewayRequest` (though I recognise this will have ripple effects). This would also offer the possibility of using the name `AuthorisationGatewayRequest` for what’s now `BaseAuthorisationGatewayRequest`. Something to think about but I’m not going to insist on any name changes (they’re easy enough to do later).

src/main/java/uk/gov/pay/connector/applepay/AuthorisationApplePayGatewayRequest.java
simad: 👍

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
simad: I originally decide not to do this right now to avoid adding 100 lines of changes - since I've been splitting the thing anyway, I'll do it :D it's a mechanical change anyway

src/main/java/uk/gov/pay/connector/gateway/model/request/BaseAuthorisationGatewayRequest.java
simad: I just copied the methods which were already there, didn't actually read what they are doing. Soz. Encapsulation doing too well.

src/test/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequestTest.java
simad: It's a test for the superclass apparently (again, just moved stuff around)

simad: renamed to make sense

src/main/java/uk/gov/pay/connector/applepay/AuthorisationApplePayGatewayRequest.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.connector.applepay;
+
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.gateway.model.request.AuthorisationGatewayRequest;
+
+public class AuthorisationApplePayGatewayRequest extends AuthorisationGatewayRequest {
```
alexbishop1: I think this wants to be `ApplePayAuthorisationGatewayRequest` (to match `CardAuthorisationGatewayRequest`)

src/test/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequestTest.java
```diff
@@ -16,7 +16,7 @@ public void shouldReturnTotalAmount_whenThereIsACorporateSurcharge() {
                 .withAmount(2000L)
                 .withCorporateSurcharge(250L)
                 .build();
-        AuthorisationGatewayRequest gatewayRequest = new AuthorisationGatewayRequest(chargeEntity, null);
+        AuthorisationGatewayRequest gatewayRequest = new CardAuthorisationGatewayRequest(chargeEntity, null);
```
alexbishop1: Is this test now `CardAuthorisationGatewayRequestTest`?

alexbishop1 (changes requested):
Couple of naming things

src/test/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequestTest.java
simad: No, it is `AuthorisationGatewayRequestTest` - I'll rename it to that.

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
alexbishop1: Yeah, it’s good to avoid huge ripple-effect renaming mixed in with other changes — perhaps I should have suggested you could do the renaming in a follow-up PR.

In this case, the other changes are sufficiently small that it’s not too hard to review both them and the renaming together.

alexbishop1 (approved):
Good step forward

PR #850, opened 2018-11-19
PP-4389: Switch to using our openjdk-jre-base image

rhowe-gds:
Switch to using Alpine's openjdk base JRE, which doesn't include image and
audio decoders.

This shrinks our containers slightly and reduces the potential for false
security alerts.


alexbishop1 (approved):
How will I use connector to listen to my tunes now? 🔈🎶

PR #851, opened 2018-11-19
PP-4314 Add ApplePayAuthorisationHandler 

simad:
- Apple pay orders are authorised similarly to how "normal" card orders are authorised, but the payload we send to worldpay is different. With the introduction of apple pay there are basically two ways to move forward: either by shoeorning an apple pay auth order in the normal "card" authorise path, or by creating a separate interface.
 - As we are currently refactoring connector and will probably get rid of the `PaymentProviders` interface, I have opted for separating the two paths. We are currently splitting the capabilities of a payment provider using multiple handlers (ie. CaptureHandler). This follows the same pattern, by introducing an `ApplePayAuthorisationHandler`. 
Unfortunately we are bringing in a bit of duplication which will hopefully be extracted away as we move forward with the refactoring.
- Made so that the Sandbox provider switches on the last 4 digits of card rather than the token number, as the latter is encrypted so it would be a bit clunky to generate proper payloads with

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -355,13 +355,12 @@ public class ChargeService {
         }).orElseThrow(() -> new ChargeNotFoundRuntimeException(externalId));
     }
 
-
     private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails) {
         CardDetailsEntity detailsEntity = new CardDetailsEntity();
         detailsEntity.setCardBrand(sanitize(authCardDetails.getCardBrand()));
         detailsEntity.setCardHolderName(sanitize(authCardDetails.getCardHolder()));
         detailsEntity.setExpiryDate(authCardDetails.getEndDate());
-        if (authCardDetails.getCardNo().length() > 6) {
+        if (hasFullCardNumber(authCardDetails)) {
             detailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6)));
         }
         detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
@@ -372,6 +371,10 @@ public class ChargeService {
         return detailsEntity;
     }
 
+    private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
+        return authCardDetails.getCardNo().length() > 6;
+    }
+    
     private TokenEntity createNewChargeEntityToken(ChargeEntity chargeEntity) {
         TokenEntity token = TokenEntity.generateNewTokenFor(chargeEntity);
         tokenDao.persist(token);
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -361,7 +361,9 @@ public class ChargeService {
         detailsEntity.setCardBrand(sanitize(authCardDetails.getCardBrand()));
         detailsEntity.setCardHolderName(sanitize(authCardDetails.getCardHolder()));
         detailsEntity.setExpiryDate(authCardDetails.getEndDate());
-        detailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6)));
+        if (authCardDetails.getCardNo().length() > 6) {
+            detailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6)));
+        }
         detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
 
         if (authCardDetails.getAddress().isPresent())
--- a/src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
@@ -38,7 +38,7 @@ public class PaymentProviders {
         return cardPaymentProviders.get(gateway);
     }
     
-    public ApplePayAuthoriser getAppleAuthoriserFor(ChargeEntity chargeEntity) {
+    public ApplePayAuthoriser getApplePayAuthoriserFor(ChargeEntity chargeEntity) {
         return applePayPaymentProviders.get(chargeEntity.getPaymentGatewayName());
     }
 
--- a/src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
@@ -1,8 +1,7 @@
 package uk.gov.pay.connector.gateway;
 
-import uk.gov.pay.connector.applepay.ApplePayAuthoriser;
-import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.gateway.epdq.EpdqPaymentProvider;
+import uk.gov.pay.connector.gateway.model.response.BaseResponse;
 import uk.gov.pay.connector.gateway.sandbox.SandboxPaymentProvider;
 import uk.gov.pay.connector.gateway.smartpay.SmartpayPaymentProvider;
 import uk.gov.pay.connector.gateway.stripe.StripePaymentProvider;
@@ -15,8 +14,7 @@ import static jersey.repackaged.com.google.common.collect.Maps.newHashMap;
 
 public class PaymentProviders {
 
-    private final Map<PaymentGatewayName, PaymentProvider> cardPaymentProviders = newHashMap();
-    private final Map<PaymentGatewayName, ApplePayAuthoriser> applePayPaymentProviders = newHashMap();
+    private final Map<PaymentGatewayName, PaymentProvider> paymentProviders = newHashMap();
     
     @Inject
     public PaymentProviders(WorldpayPaymentProvider worldpayPaymentProvider,
@@ -24,22 +22,15 @@ public class PaymentProviders {
                             SmartpayPaymentProvider smartpayPaymentProvider,
                             SandboxPaymentProvider sandboxPaymentProvider,
                             StripePaymentProvider stripePaymentProvider) {
-        cardPaymentProviders.put(PaymentGatewayName.WORLDPAY, worldpayPaymentProvider);
-        cardPaymentProviders.put(PaymentGatewayName.SANDBOX, sandboxPaymentProvider);
-        cardPaymentProviders.put(PaymentGatewayName.SMARTPAY, smartpayPaymentProvider);
-        cardPaymentProviders.put(PaymentGatewayName.EPDQ, epdqPaymentProvider);
-        cardPaymentProviders.put(PaymentGatewayName.STRIPE, stripePaymentProvider);
+        paymentProviders.put(PaymentGatewayName.WORLDPAY, worldpayPaymentProvider);
+        paymentProviders.put(PaymentGatewayName.SANDBOX, sandboxPaymentProvider);
+        paymentProviders.put(PaymentGatewayName.SMARTPAY, smartpayPaymentProvider);
+        paymentProviders.put(PaymentGatewayName.EPDQ, epdqPaymentProvider);
+        paymentProviders.put(PaymentGatewayName.STRIPE, stripePaymentProvider);
 
-        applePayPaymentProviders.put(PaymentGatewayName.WORLDPAY, worldpayPaymentProvider);
-        applePayPaymentProviders.put(PaymentGatewayName.SANDBOX, sandboxPaymentProvider);
     }
 
-    public PaymentProvider byName(PaymentGatewayName gateway) {
-        return cardPaymentProviders.get(gateway);
+    public PaymentProvider<BaseResponse> byName(PaymentGatewayName gateway) {
+        return paymentProviders.get(gateway);
     }
-    
-    public ApplePayAuthoriser getApplePayAuthoriserFor(ChargeEntity chargeEntity) {
-        return applePayPaymentProviders.get(chargeEntity.getPaymentGatewayName());
-    }
-
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.gateway;
 
+import uk.gov.pay.connector.applepay.ApplePayAuthoriser;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.gateway.epdq.EpdqPaymentProvider;
 import uk.gov.pay.connector.gateway.sandbox.SandboxPaymentProvider;
 import uk.gov.pay.connector.gateway.smartpay.SmartpayPaymentProvider;
@@ -13,22 +15,31 @@ import static jersey.repackaged.com.google.common.collect.Maps.newHashMap;
 
 public class PaymentProviders {
 
-    private final Map<PaymentGatewayName, PaymentProvider> paymentProviders = newHashMap();
-
+    private final Map<PaymentGatewayName, PaymentProvider> cardPaymentProviders = newHashMap();
+    private final Map<PaymentGatewayName, ApplePayAuthoriser> applePayPaymentProviders = newHashMap();
+    
     @Inject
     public PaymentProviders(WorldpayPaymentProvider worldpayPaymentProvider,
                             EpdqPaymentProvider epdqPaymentProvider,
                             SmartpayPaymentProvider smartpayPaymentProvider,
                             SandboxPaymentProvider sandboxPaymentProvider,
                             StripePaymentProvider stripePaymentProvider) {
-        paymentProviders.put(PaymentGatewayName.WORLDPAY, worldpayPaymentProvider);
-        paymentProviders.put(PaymentGatewayName.SMARTPAY, smartpayPaymentProvider);
-        paymentProviders.put(PaymentGatewayName.SANDBOX, sandboxPaymentProvider);
-        paymentProviders.put(PaymentGatewayName.EPDQ, epdqPaymentProvider);
-        paymentProviders.put(PaymentGatewayName.STRIPE, stripePaymentProvider);
+        cardPaymentProviders.put(PaymentGatewayName.WORLDPAY, worldpayPaymentProvider);
+        cardPaymentProviders.put(PaymentGatewayName.SANDBOX, sandboxPaymentProvider);
+        cardPaymentProviders.put(PaymentGatewayName.SMARTPAY, smartpayPaymentProvider);
+        cardPaymentProviders.put(PaymentGatewayName.EPDQ, epdqPaymentProvider);
+        cardPaymentProviders.put(PaymentGatewayName.STRIPE, stripePaymentProvider);
+
+        applePayPaymentProviders.put(PaymentGatewayName.WORLDPAY, worldpayPaymentProvider);
+        applePayPaymentProviders.put(PaymentGatewayName.SANDBOX, sandboxPaymentProvider);
     }
 
     public PaymentProvider byName(PaymentGatewayName gateway) {
-        return paymentProviders.get(gateway);
+        return cardPaymentProviders.get(gateway);
     }
+    
+    public ApplePayAuthoriser getAppleAuthoriserFor(ChargeEntity chargeEntity) {
+        return applePayPaymentProviders.get(chargeEntity.getPaymentGatewayName());
+    }
+
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
@@ -7,6 +7,8 @@ import org.apache.http.NameValuePair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.applepay.ApplePayAuthorisationGatewayRequest;
+import uk.gov.pay.connector.applepay.ApplePayAuthorisationHandler;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability;
 import uk.gov.pay.connector.gateway.GatewayClient;
@@ -139,6 +141,11 @@ public class EpdqPaymentProvider implements PaymentProvider<String> {
         return GatewayResponseGenerator.getEpdqGatewayResponse(cancelClient, response, EpdqCancelResponse.class);
     }
 
+    @Override
+    public GatewayResponse<BaseAuthoriseResponse> authoriseApplePay(ApplePayAuthorisationGatewayRequest request) {
+        throw new UnsupportedOperationException("Apple Pay is not supported for EPDQ");
+    }
+    
     @Override
     public GatewayResponse<BaseAuthoriseResponse> authorise3dsResponse(Auth3dsResponseGatewayRequest request) {
         Either<GatewayError, GatewayClient.Response> response = authoriseClient.postRequestFor(ROUTE_FOR_QUERY_ORDER, request.getGatewayAccount(), buildQueryOrderRequestFor(request));
--- a/src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
@@ -13,35 +13,34 @@ class SandboxCardNumbers {
 
     static boolean isValidCard(String cardNumber) {
         return GOOD_CARDS.contains(cardNumber) ||
-                GOOD_CORPORATE_CARDS.contains(cardNumber);
+                GOOD_CORPORATE_CARDS.contains(cardNumber) || GOOD_APPLE_PAY_LAST_DIGITS_CARD_NUMBER.equals(cardNumber);
     }
 
     static boolean isRejectedCard(String cardNumber) {
         return REJECTED_CARDS
                 .keySet()
                 .stream()
-                .anyMatch(a -> a.contains(cardNumber));
+                .anyMatch(rejectedCards -> rejectedCards.contains(cardNumber));
     }
 
     static boolean isErrorCard(String cardNumber) {
         return ERROR_CARDS
                 .keySet()
                 .stream()
-                .anyMatch(a -> a.contains(cardNumber));
+                .anyMatch(errorCards -> errorCards.contains(cardNumber));
     }
 
     static CardError cardErrorFor(String cardNumber) {
         return ERROR_CARDS
                 .entrySet()
                 .stream()
-                .filter(a -> a.getKey().contains(cardNumber))
+                .filter(errorCards -> errorCards.getKey().contains(cardNumber))
                 .findFirst()
                 .map(Map.Entry::getValue)
                 .orElse(null);
     }
 
     private static final List<String> GOOD_CARDS = ImmutableList.of(
-            "4242",
             "4444333322221111",
             "4242424242424242",
             "4917610000000000003",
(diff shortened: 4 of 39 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -362,7 +362,9 @@ private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails
         detailsEntity.setCardBrand(sanitize(authCardDetails.getCardBrand()));
         detailsEntity.setCardHolderName(sanitize(authCardDetails.getCardHolder()));
         detailsEntity.setExpiryDate(authCardDetails.getEndDate());
-        detailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6)));
+        if (authCardDetails.getCardNo().length() > 6) {
```
alexbishop1: I assume for Apple Pay this is going to be under six digits (because we don’t have a full card number, just the last four digits) but I don’t fully understand what’s going on. Could you explain it to me please?

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
```diff
@@ -9,26 +9,39 @@
...
                 GOOD_CORPORATE_CARDS.contains(cardNumber);
     }
 
-    public static boolean isRejectedCard(String cardNumber) {
-        return REJECTED_CARDS.containsKey(cardNumber);
+    static boolean isRejectedCard(String cardNumber) {
+        return REJECTED_CARDS
+                .keySet()
+                .stream()
+                .anyMatch(a -> a.contains(cardNumber));
```
alexbishop1: Not sure why we’re doing substring matches now

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
```diff
@@ -42,19 +55,26 @@ public static CardError cardErrorFor(String cardNumber) {
...
-            PROCESSING_ERROR_CARD_NUMBER, new CardError(AUTHORISATION_ERROR, "This transaction could be not be processed."));
-
-    private static final Map<String, CardError> REJECTED_CARDS = ImmutableMap.of(
-            DECLINED_CARD_NUMBER, new CardError(AUTHORISATION_REJECTED, "This transaction was declined."),
-            EXPIRED_CARD_NUMBER, new CardError(AUTHORISATION_REJECTED, "The card is expired."),
-            CVC_ERROR_CARD_NUMBER, new CardError(AUTHORISATION_REJECTED, "The CVC code is incorrect."));
+            "5101180000000007");
+    private static final List<String> DECLINED_CARD_NUMBERS = ImmutableList.of(
+            "4000000000000002",
+            "0002");
```
alexbishop1: I think we’ve reached the stage where we need some explanation of what these numbers are — maybe we could extract each of them to appropriately-named constants

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProvider.java
```diff
@@ -31,41 +35,57 @@
...
 
-    @Override
-    public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
-        String cardNumber = request.getAuthCardDetails().getCardNo();
-        GatewayResponseBuilder<BaseResponse> gatewayResponseBuilder = responseBuilder();
+    private GatewayResponse<BaseAuthoriseResponse> authorise(String cardNumber) {
+        GatewayResponseBuilder<BaseAuthoriseResponse> gatewayResponseBuilder = responseBuilder();
 
         if (SandboxCardNumbers.isErrorCard(cardNumber)) {
+            LOGGER.info("sandbox auth has been called with an error card number");
```
alexbishop1: Not sure logging like this is going to be very useful in non-local environments with an identifier for the payment

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProvider.java
```diff
@@ -31,41 +35,57 @@
...
+    public GatewayResponse<BaseAuthoriseResponse> authorise(CardAuthorisationGatewayRequest request) {
+        LOGGER.info("sandbox card auth");
+        String cardNumber = request.getAuthCardDetails().getCardNo();
+        return authorise(cardNumber);
+    }
+    
     @Override
-    public GatewayResponse<BaseResponse> authorise3dsResponse(Auth3dsResponseGatewayRequest request) {
-        GatewayResponseBuilder<BaseResponse> gatewayResponseBuilder = responseBuilder();
+    public GatewayResponse<BaseAuthoriseResponse> authorise(ApplePayAuthorisationGatewayRequest request) {
```
alexbishop1: We now have three methods named `authorise`, which I found quite hard to follow. I understand the public ones may need to share the name because polymorphism but I think at least the `private` one could be renamed. Also, can we have the public methods at the top with the private methods at the bottom?

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProvider.java
```diff
@@ -31,41 +35,57 @@
...
+        LOGGER.info("sandbox apple pay auth");
+        String lastDigitsCardNumber = request.getAppleDecryptedPaymentData().getPaymentInfo().getLastDigitsCardNumber();
+        return authorise(lastDigitsCardNumber);
```
alexbishop1: Not convinced about sometimes passing a full card number to the private `authorise(String)` method and sometimes passing the last four digits. Could these be separate methods? I think this would also mean we don’t need to do the substring stuff in `SandboxCardNumbers` (at the expense of a little code duplication).

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -185,7 +186,7 @@ private Response postToStripe(String path, String payload, String metricsPrefix)
     }
 
     @Override
-    public GatewayResponse<BaseResponse> authorise3dsResponse(Auth3dsResponseGatewayRequest request) {
+    public GatewayResponse<BaseAuthoriseResponse> authorise3dsResponse(Auth3dsResponseGatewayRequest request) {
```
alexbishop1: Hooray for more precise types!

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -178,6 +189,18 @@ private GatewayOrder buildAuthoriseOrder(CardAuthorisationGatewayRequest request
                 .build();
     }
 
+    private GatewayOrder buildApplePayAuthoriseOrder(ApplePayAuthorisationGatewayRequest request) {
+        return aWorldpayAuthoriseApplePayOrderRequestBuilder()
+                .withApplePayTemplateData(ApplePayTemplateData.from(request.getAppleDecryptedPaymentData()))
+                .withSessionId(request.getChargeExternalId())
+                .with3dsRequired(request.getGatewayAccount().isRequires3ds())
```
alexbishop1: A bit tangential, but why does this matter? I thought with Apple Pay we send through 3D Secure authentication stuff (or Apple Pay authentication stuff disguised as 3D Secure authentication stuff) with every authorisation request.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -136,6 +136,7 @@ private ChargeStatus determineNextStatus(GatewayResponse<BaseCaptureResponse> op
     }
 
     PaymentProvider<BaseCaptureResponse, ?> getPaymentProviderFor(ChargeEntity chargeEntity) {
-        return providers.byName(chargeEntity.getPaymentGatewayName());
+        PaymentProvider<BaseCaptureResponse, ?> paymentProvider = providers.byName(chargeEntity.getPaymentGatewayName());
+        return paymentProvider;
```
alexbishop1: This change isn’t really doing anything

src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
```diff
@@ -13,22 +15,31 @@
...
+        applePayPaymentProviders.put(PaymentGatewayName.WORLDPAY, worldpayPaymentProvider);
+        applePayPaymentProviders.put(PaymentGatewayName.SANDBOX, sandboxPaymentProvider);
     }
 
     public PaymentProvider byName(PaymentGatewayName gateway) {
-        return paymentProviders.get(gateway);
+        return cardPaymentProviders.get(gateway);
     }
+    
+    public ApplePayAuthoriser getAppleAuthoriserFor(ChargeEntity chargeEntity) {
```
alexbishop1: `getApplePayAuthoriserFor`

src/test/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProviderTest.java
```diff
@@ -79,10 +77,46 @@ public void parseNotification_shouldFailParsingNotification() throws Exception {
 
         provider.parseNotification(notification);
     }
+    
+    @Test
+    public void authorise_shouldBeAuthorisedWhenApplePayTokenNumberIsExpectedToSucceedForAuthorisation() {
+        AppleDecryptedPaymentData applePaymentData = AuthUtils.ApplePay.buildDecryptedPaymentData("Mr. Payment", "mr@payment.test", "4242");
```
alexbishop1: It would be nice if `"4242"` could be an appropriately-named constant

src/test/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProviderTest.java
```diff
@@ -79,10 +77,46 @@ public void parseNotification_shouldFailParsingNotification() throws Exception {
...
+        assertThat(gatewayResponse.getGatewayError().isPresent(), is(false));
+        assertThat(gatewayResponse.getBaseResponse().isPresent(), is(true));
+        assertThat(gatewayResponse.getBaseResponse().get() instanceof BaseAuthoriseResponse, is(true));
+
+        BaseAuthoriseResponse authoriseResponse = (BaseAuthoriseResponse) gatewayResponse.getBaseResponse().get();
+        assertThat(authoriseResponse.authoriseStatus(), is(AuthoriseStatus.REJECTED));
+        assertThat(authoriseResponse.getTransactionId(), is(notNullValue()));
+        assertThat(authoriseResponse.getErrorCode(), is(nullValue()));
+        assertThat(authoriseResponse.getErrorMessage(), is(nullValue()));
+    }
```
alexbishop1: Can we do a test for an authorisation error?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -132,7 +140,34 @@ public void testRefundRequestContainsReference() {
...
+        when(mockGatewayAccountEntity.isRequires3ds()).thenReturn(false);
+        when(mockGatewayClient.postRequestFor(isNull(), any(GatewayAccountEntity.class), any(GatewayOrder.class)))
+                .thenReturn(left(unexpectedStatusCodeFromGateway("Unexpected HTTP status code 400 from gateway")));
+
+        GatewayClientFactory gatewayClientFactory = mock(GatewayClientFactory.class);
+        when(gatewayClientFactory.createGatewayClient(any(PaymentGatewayName.class), any(GatewayOperation.class), any(Map.class), any(BiFunction.class), any())).thenReturn(mockGatewayClient);
+
+        WorldpayPaymentProvider worldpayPaymentProvider = new WorldpayPaymentProvider(configuration, gatewayClientFactory, environment);
+
+        worldpayPaymentProvider.authorise(getApplePayAuthorisationRequest(chargeEntity));
```
alexbishop1: I have to confess that I was staring at this test for ages before I realised, “This is the Apple Pay bit! This is what makes this test different from the others!”

I think as we’ve created more variants of the authorisation tests, a lot of boilerplate has been copied and pasted between them. We could perhaps try to consolidate some of it into a `@Before` method in a follow-up PR.

src/test/java/uk/gov/pay/connector/util/AuthUtils.java
```diff
@@ -35,14 +35,14 @@ private static AppleDecryptedPaymentData buildDecryptedPaymentData(String cardHo
             );
         }
 
-        public static AppleDecryptedPaymentData buildDecryptedPaymentData(String cardHolderName, String email, String tokenNumber) {
-            return buildDecryptedPaymentData(cardHolderName, email, tokenNumber, new AppleDecryptedPaymentData.PaymentData(
+        public static AppleDecryptedPaymentData buildDecryptedPaymentData(String cardHolderName, String email, String lastFourDigitsCardNumber) {
```
alexbishop1, line 38: Recently (very recently: perhaps after you started doing this work), we’ve done some work to remove a lot of test fixture stuff from `AuthUtils` in favour of `AuthCardDetailsFixture`. Could we do something the Apple Pay authorisation test fixtures? Perhaps in a follow-up PR?

alexbishop1 (changes requested):
This looks good.

I’m still not sure about the `ApplePayAuthoriser` interface itself. I can see that it’s nice that we indicate which payment gateways support Apple Pay but it doesn’t really have a plain card payment parallel and it’s a shame that the actual implementation ends up in the already-doing-more-than-it-probably-should-be `SandboxPaymentProvider` and `WorldpayPaymentProvider` I may ponder on this a bit more but I have no qualms with the functionality itself.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
simad: This is where we are setting the details for last 4 digits and first 6 digits - you're right, in case of apple pay cardNo will have 4 digits, so that `FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6))` would not pass the validation because it assumes we have 6 digits. So, I am not setting the first 6 digits in case we don't have the full card number.
This means first six digits will be null for apple pay payments

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
simad: it's not a substring match, I forgot to rename `a` to a properly named rejectedCardsList or something - that function is just checking that the list of rejected cards contains cardNumber. Will rename, sorry!

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
simad: 👍 sounds good

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
alexbishop1: Of course, it’s doing `contains` on a `List`, not a `String` (`String` doesn’t even have `contains`!)

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: It’s a shame that it’s not obvious from this specific code under which circumstances the card number won’t have the first six digits and it’s unfortunate that the card number is sometimes a card number and other times only has the last four digits. More broadly, I don’t really understand why we have two `AuthorisationDetails` implementations used for Apple Pay: `AuthCardDetails` and `AppleDecryptedPaymentData`. I would have hoped we could have one class that contains just what’s available.

simad: AuthCardDetails is what is used by jpa to put stuff in the database, that's the problem with the mashup of json + jpa + domain object we have. So I have to get our model AppleDecryptedPaymentData and kind of match it to what we store in the db (AuthCardDetails).

simad: If you have a better idea / better way to do this, it would be awesome!

alexbishop1: Ah, our wonderful domain objects that are JSON, JPA, floor wax and dessert topping! At least we have a rough plan for gradually unpicking this now.

So is it that `AppleDecryptedPaymentData` is the incoming JSON, which then gets converted to `AuthCardDetails` (which can also represent incoming JSON but isn’t in this case) so it can be persisted to the database?

I guess leave it as it is for now and hopefully we’ll make it better later. It might be an idea to put a comment next to `if (authCardDetails.getCardNo().length() > 6)` explaining that for Apple Pay authorisations the card number is only the four digits (a comment that will hopefully become unnecessary when future refactoring makes our code wonderfully self-describing and comprehensible).

simad: precisely :( `AppleDecryptedPaymentData` is the decrypted version of the incoming encrypted json payload, and it is a fairly pure pojo, but AuthCardDetails is both what comes from frontend AND what's stored in the db :(
I'll try to think of a way to make `authCardDetails.getCardNo().length() > 6` not needing a comment, but if it takes too long I might just leave it and improve things later on. Thank you :)

alexbishop1:
As promised, I pondered a bit more about the `ApplePayAuthoriser` interface. I like the direction it’s going in, which is similar to the new `CaptureHandler` interface in that it encapsulates the methods for a single payment gateway operation.

The main difference is in the implementations: while the concrete implementation of`ApplePayAuthoriser` is`WorldpayPaymentProvider` (or equivalent for other gateways), the concrete implementation of `CaptureHandler` is `WorldpayCaptureHandler` (or equivalent for other gateways). Furthermore, while the appropriate `ApplePayAuthoriser` implementation is obtained from `PaymentProviders`, the appropriate `CaptureHandler` is obtained from the `WorldpayPaymentProvider` or whatever.

Could you make `ApplePayAuthoriser` behave a bit more like `CaptureHandler` please? That is, rename it to `ApplePayAuthorisationHandler`, put the concrete implementations in new `WorldpayApplePayAuthorisationHandler` and `SandboxApplePayAuthorisationHandler` classes (rather than `WorldpayPaymentProvider` and `SandboxPaymentProvider` respectively) and make it so that the `ApplePayAuthorisationHandler` instance is obtained from the `WorldpayPaymentProvider` and `SandboxPaymentProvider` (rather than the `PaymentProviders` class).

I hope that we’re intending to continue down this path and will eventually have interfaces like `CardAuthorisationHandler` (for plain card authorisations), `Card3dsResponseAuthorisationHandler` (for authorisations using the response of 3D Secure authentication) and `RefundHandler` etc., each with their own gateway-specific implementations. This would stop `WorldpayPaymentProvider` and friends from being such huge beasts (and I’m quite looking forward to breaking up the monstrous test classes: not only would they be smaller, but it would be easier to share common code in a `@Before` method because each test class would only be dealing with a single operation).

src/test/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProviderTest.java
simad: 👍

src/main/java/uk/gov/pay/connector/gateway/PaymentProviders.java
simad: 👍

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProvider.java
simad: removed those lines

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
simad: It is still a super clunky interface but it didn't make any sense to have a generic response when we know that authorise will have an authorise response :(

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxPaymentProvider.java
simad: they could be separate of course, but it is literally the same code

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
simad: I had originally removed the getPaymentProviderFor and then put it back to reduce the change, copy paste fail

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
simad: true, it does not matter at all. copy paste again :)

src/test/java/uk/gov/pay/connector/util/AuthUtils.java
simad: Sure, I'll do it in the next PR. Cheers :)

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -371,6 +372,10 @@ private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails
         return detailsEntity;
     }
 
+    private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
+        return authCardDetails.getCardNo().length() > 6;
```
alexbishop1, line 375: Maybe a comment like `// Apple Pay etc. don’t give us a full card number, just the last four digits here` might make the _why_ clearer?

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -139,6 +141,11 @@ public PaymentGatewayName getPaymentGatewayName() {
         return GatewayResponseGenerator.getEpdqGatewayResponse(cancelClient, response, EpdqCancelResponse.class);
     }
 
+    @Override
+    public GatewayResponse<BaseAuthoriseResponse> authoriseApplePay(ApplePayAuthorisationGatewayRequest request) {
+        throw new UnsupportedOperationException("Apple Pay is not supported for EPDQ");
```
alexbishop1, line 146: Minor, but it’s “ePDQ”

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
```diff
@@ -39,22 +51,28 @@ public static CardError cardErrorFor(String cardNumber) {
...
-            EXPIRED_CARD_NUMBER, new CardError(AUTHORISATION_REJECTED, "The card is expired."),
-            CVC_ERROR_CARD_NUMBER, new CardError(AUTHORISATION_REJECTED, "The CVC code is incorrect."));
-    
+    private static final Map<List<String>, CardError> REJECTED_CARDS = ImmutableMap.of(
+            ImmutableList.of(DECLINED_CARD_NUMBER, DECLINED_APPLE_PAY_LAST_DIGITS_CARD_NUMBER),
+            new CardError(AUTHORISATION_REJECTED, "This transaction was declined."),
+            ImmutableList.of(EXPIRED_CARD_NUMBER, EXPIRED_APPLE_PAY_LAST_DIGITS_CARD_NUMBER),
+            new CardError(AUTHORISATION_REJECTED, "The card is expired."),
+            ImmutableList.of(CVC_ERROR_CARD_NUMBER, CVC_ERROR_APPLE_PAY_LAST_DIGITS_CARD_NUMBER),
+            new CardError(AUTHORISATION_REJECTED, "The CVC code is incorrect."));
```
alexbishop1, line 77: I find this change from the “bad” card maps being `Map<String, CardError>` to `Map<List<String>, CardError>` makes the look-up logic hard to follow and we kind of lose the benefit of it being a map because we have to sequentially look at each key until we find the one we want.

I think it was clearer when it was `Map<String, CardError>`. I assume the change to `Map<List<String>, CardError>` was to avoid lots of repitition from many card numbers being mapped to the same `CardError`. We could do something like this to handle that:

```java

private static Map<String, CardError> ERROR_CARDS = Stream.of(
        PROCESSING_ERROR_CARD_NUMBER,
        PROCESSING_ERROR_APPLE_PAY_LAST_DIGITS_CARD_NUMBER)
        .collect(toMap(
                Function.identity(),
                cardNumber -> new CardError(AUTHORISATION_ERROR, "This transaction could be not be processed.")));
```

Then `isErrorCard(String)` and `cardErrorFor(String)` can go back to how they used to be.

src/main/java/uk/gov/pay/connector/gateway/sandbox/applepay/SandboxApplePayAuthorisationHandler.java
```diff
@@ -0,0 +1,43 @@
...
+
+import static uk.gov.pay.connector.gateway.model.ErrorType.GENERIC_GATEWAY_ERROR;
+import static uk.gov.pay.connector.gateway.model.response.GatewayResponse.GatewayResponseBuilder.responseBuilder;
+
+public class SandboxApplePayAuthorisationHandler implements ApplePayAuthorisationHandler {
+    private final static Logger LOGGER = LoggerFactory.getLogger(SandboxApplePayAuthorisationHandler.class);
+
+    @Override
+    public GatewayResponse<BaseAuthoriseResponse> authorise(ApplePayAuthorisationGatewayRequest request) {
+        LOGGER.info("sandbox apple pay auth");
```
alexbishop1, line 22: I thought this line was going away

src/test/java/uk/gov/pay/connector/gateway/sandbox/applepay/SandboxApplePayAuthorisationHandlerTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+        BaseAuthoriseResponse authoriseResponse = (BaseAuthoriseResponse) gatewayResponse.getBaseResponse().get();
+        assertThat(authoriseResponse.authoriseStatus(), is(BaseAuthoriseResponse.AuthoriseStatus.REJECTED));
+        assertThat(authoriseResponse.getTransactionId(), is(notNullValue()));
+        assertThat(authoriseResponse.getErrorCode(), is(nullValue()));
+        assertThat(authoriseResponse.getErrorMessage(), is(nullValue()));
+    }
+
+    @Test
+    public void authorise_shouldGetGatewayErrorWhenLastDigitsCardNumbersAreExpectedToFailForAuthorisation_forApplePay() {
```
alexbishop1, line 70: Thanks for adding an authorisation error test

src/test/java/uk/gov/pay/connector/gateway/sandbox/applepay/SandboxApplePayAuthorisationHandlerTest.java
```diff
@@ -0,0 +1,83 @@
...
+        assertThat(gatewayResponse.isSuccessful(), is(false));
+        assertThat(gatewayResponse.isFailed(), is(true));
+        assertThat(gatewayResponse.getGatewayError().isPresent(), is(true));
+        assertThat(gatewayResponse.getBaseResponse().isPresent(), is(false));
+
+        GatewayError gatewayError = (GatewayError) gatewayResponse.getGatewayError().get();
+        assertThat(gatewayError.getErrorType(), is(GENERIC_GATEWAY_ERROR));
+        assertThat(gatewayError.getMessage(), is("This transaction could be not be processed."));
+    }
+}
```
alexbishop1, line 83: Hooray! This stuff this is testing is small enough that the test class only needs to be 83 lines!

alexbishop1 (approved):
Thanks for making the changes!

I think moving the Apple Pay authorisation logic into separate gateway-specific classes worked out well and it’s great that this meant that the changes to `WorldpayPaymentProvider` and `SandboxPaymentProvider` basically thinned out to nothing and that some of my review comments ceased to be relevant.

I’ve left a few comments but there’s nothing serious enough to withhold approval. Could you look at addressing them in a follow-up PR (perhaps along with `AuthUtils` to `AuthCardDetailsFixture` migration)?

src/main/java/uk/gov/pay/connector/gateway/sandbox/applepay/SandboxApplePayAuthorisationHandler.java
simad: It will in the next pr 🗡

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
simad: oki doki

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
simad: rOGER

PR #852, opened 2018-11-20
PP-4315 Add wallet column to charges table

simad:
## WHAT
 - This column will be used to flag if the payment has been done using an electronic wallet like `APPLE_PAY`. It will be NULL by default for all of the other payments

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -949,4 +949,12 @@
         </update>
     </changeSet>
 
+    <changeSet id="add wallet column to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="wallet" type="varchar(50)">
+                <constraints nullable="true" unique="false"/>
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -949,4 +949,12 @@
         </update>
     </changeSet>
 
+    <changeSet id="add wallet column to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="wallet" type="varchar(100)">
```
alexbishop1: I’m still a bit annoyed there isn’t a better term than “wallet” for this, what with Apple and their ilk all maintaining that each of their solutions is _sui generis_ and couldn’t possibly be lumped together with similar solutions!

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #856, opened 2018-11-21
BAU remove logic from charge entity

georgeracu:
## WHAT
* Removed logic from `ChargeEntity`
* Refactored long conditionals into smaller and meaningful methods

## HOW 
BAU - Extract conditionals into smaller methods

    - Long if statements can be expressed better by encapsulating them into
    one line methods. Beside making the line shorter, the name of the
    method is meaningful in the context of the if statement, instead of trying
    to figure out what's going on with status mappers, equality etc.

BAU - Remove logic from `ChargeEntity`

    - I removed the two methods `hasStatus` from the `ChargeEntity` as this feels wrong.
    The entity already has a `getStatus()` method which returns the current status, while
    the two methods were used to actually check that the current status is in a list of
    already known states. This can be easily accomplished by just using `contains` of
    Collections framework
    - Added a private constructor on `CancelServiceFunctions` to prevent people from
    instantiating it as this class has only static methods
    - Updated relevant tests to test on the actual String value that gets returned by the
    `getStatus()` method, instead trying to compare a String (returned by the entity) with
    a ChargeStatus value. Assert on the value you get and not on some other object.
    - Reordered modifiers to comply with JLS. Will make it easier for future reading and
    understanding of the code.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/CancelServiceFunctions.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/CancelServiceFunctions.java
@@ -21,6 +21,7 @@ import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 
 import java.time.ZonedDateTime;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
@@ -35,6 +36,10 @@ import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.fromString;
 class CancelServiceFunctions {
 
     private static final Logger logger = LoggerFactory.getLogger(CancelServiceFunctions.class);
+
+    private CancelServiceFunctions() {
+        // prevent people to instantiate this class, as it has only static methods
+    }
 
     static TransactionalOperation<TransactionContext, ChargeEntity> changeStatusTo(ChargeDao chargeDao, ChargeEventDao chargeEventDao, String chargeId, ChargeStatus targetStatus, Optional<ZonedDateTime> generationTimeOptional) {
         return context -> chargeDao.findByExternalId(chargeId)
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/charge/service/CancelServiceFunctions.java
```diff
@@ -36,6 +37,10 @@
 
     private static final Logger logger = LoggerFactory.getLogger(CancelServiceFunctions.class);
 
+    private CancelServiceFunctions() {
+        // prevent people to instantiate this class, as it has only static methods
```
oswaldquek: do we need this comment here? i think the comment is implied by virtue of having the private constructor

src/main/java/uk/gov/pay/connector/charge/service/CancelServiceFunctions.java
```diff
@@ -52,13 +57,13 @@
...
                                                                                            String chargeId,
                                                                                            StatusFlow statusFlow
-                                                                                           ) {
+    ) {
         return context -> chargeDao.findByExternalId(chargeId).map(chargeEntity -> {
             ChargeStatus newStatus = statusFlow.getLockState();
-            if (!chargeEntity.hasStatus(statusFlow.getTerminatableStatuses())) {
-                if (chargeEntity.hasStatus(newStatus)) {
+            if (!chargeIsInTerminatableStatus(statusFlow, chargeEntity)) {
+                if (newStatus.equals(ChargeStatus.fromString(chargeEntity.getStatus()))) {
```
oswaldquek: i think the code could be cleaner if `ChargeStatus.fromString(chargeEntity.getStatus())` were extracted. then `chargeIsInTerminatableStatus` could take in that ChargeStatus as well instead of having to pass in the whole chargeEntity.

oswaldquek (changes requested):
Just two minor comments!

oswaldquek (dismissed):

oswaldquek (approved):

PR #859, opened 2018-11-22
PP-4374 Introduced EpdqNotificationService

kbottla:
## WHAT

Part of refactoring notification handling (see https://github.com/alphagov/pay-connector/pull/787 for the original full attempt...). Introducing gateway specific service objects to handle notifications instead of generalised notification handler (`NotificationService`). We don't need a generalised `NotificationService` as each gateway notification has dedicated notification resource endpoint. Having separate service object will make it easier to understand and follow through the path.  

- Added `EpdqNotificationService` to delegate EPDQ notification processing to service instead of a generic handler.
- NotificationResource for EPDQ notifications now uses `EpdqNotificationService`

Note that any existing code that handles EPDQ notifications still stays and will be removed in subsequent PRs

## HOW 
- Existing tests in `EpdqNotificationResourceITest` should continue to pass
- All tests in `EpdqNotificationServiceTest`, `EpdqNotificationServiceStatusMapperTest` should pass

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotification.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotification.java
@@ -3,9 +3,10 @@ package uk.gov.pay.connector.gateway.epdq;
 import com.google.common.base.Strings;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.utils.URLEncodedUtils;
-import uk.gov.pay.connector.usernotification.model.ChargeStatusRequest;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.usernotification.model.ChargeStatusRequest;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -26,9 +27,41 @@ public class EpdqNotification implements ChargeStatusRequest {
     private final String payIdSub;
     private final String shaSign;
 
+    public enum StatusCode {
+        EPDQ_AUTHORISATION_REFUSED("2"),
+        EPDQ_AUTHORISED("5"),
+        EPDQ_AUTHORISED_CANCELLED("6"),
+        EPDQ_PAYMENT_DELETED("7"),
+        EPDQ_DELETION_REFUSED("73"),
+        EPDQ_REFUND("8"),
+        EPDQ_REFUND_REFUSED("83"),
+        EPDQ_PAYMENT_REQUESTED("9"),
+        EPDQ_REFUND_DECLINED_BY_ACQUIRER("94"),
+        UNKNOWN("");
+
+        public String getCode() {
+            return code;
+        }
+
+        public final String code;
+
+        StatusCode(final String code) {
+            this.code = code;
+        }
+
+        public static StatusCode byCode(String code) {
+            return Arrays.stream(StatusCode.values()).filter(c -> c.getCode().equals(code)).findFirst().orElse(UNKNOWN);
+        }
+
+        @Override
+        public String toString() {
+            return this.name();
+        }
+    }
+
     private Optional<ChargeStatus> chargeStatus = Optional.empty();
 
-    public EpdqNotification(String payload) {
+    public EpdqNotification(String payload) throws EpdqParseException {
         try {
             paramsList = URLEncodedUtils.parse(payload, EpdqPaymentProvider.EPDQ_APPLICATION_X_WWW_FORM_URLENCODED_CHARSET);
 
@@ -40,8 +73,7 @@ public class EpdqNotification implements ChargeStatusRequest {
             payIdSub = params.get(PAYIDSUB_KEY);
             shaSign = params.get(SHASIGN_KEY);
         } catch (Exception e) {
-            throw new IllegalArgumentException("Could not decode ePDQ notification payload as "
-                    + EpdqPaymentProvider.EPDQ_APPLICATION_X_WWW_FORM_URLENCODED_CHARSET.name() + " application/x-www-form-urlencoded");
+            throw new EpdqParseException("Could not decode ePDQ notification payload as UTF-8 application/x-www-form-urlencoded");
         }
     }
 
@@ -72,6 +104,18 @@ public class EpdqNotification implements ChargeStatusRequest {
         return status;
     }
 
+    public StatusCode getStatusCode() {
+        return StatusCode.byCode(status);
+    }
+
+    public String describeStatusCode() {
+        final String description = getStatusCode().name();
+        if (getStatusCode() == StatusCode.UNKNOWN) {
+            return description + " (" + status + ")";
+        }
+        return description;
+    }
+
     @Override
     public String getTransactionId() {
         return getPayId();
@@ -84,6 +128,17 @@ public class EpdqNotification implements ChargeStatusRequest {
 
     public void setChargeStatus(Optional<ChargeStatus> chargeStatus) {
         this.chargeStatus = chargeStatus;
+    }
+
+    @Override
+    public String toString() {
+        return "EpdqNotification{" +
+                "status='" + status + "' " +
+                "(" + describeStatusCode() + ")" +
+                ", payId='" + payId + '\'' +
+                ", payIdSub='" + payIdSub + '\'' +
+                ", chargeStatus=" + chargeStatus +
+                '}';
     }
 
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
@@ -1,0 +1,182 @@
+package uk.gov.pay.connector.gateway.epdq;
+
+import com.google.inject.Inject;
+import com.google.inject.persist.Transactional;
+import org.apache.http.NameValuePair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.dao.ChargeDao;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.gateway.PaymentGatewayName;
+import uk.gov.pay.connector.gateway.processor.ChargeNotificationProcessor;
+import uk.gov.pay.connector.gateway.processor.RefundNotificationProcessor;
+import uk.gov.pay.connector.refund.model.domain.RefundStatus;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
+import static uk.gov.pay.connector.charge.service.StatusFlow.EXPIRE_FLOW;
+import static uk.gov.pay.connector.charge.service.StatusFlow.SYSTEM_CANCELLATION_FLOW;
+import static uk.gov.pay.connector.charge.service.StatusFlow.USER_CANCELLATION_FLOW;
+import static uk.gov.pay.connector.gateway.epdq.EpdqNotification.SHASIGN_KEY;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_SHA_OUT_PASSPHRASE;
+import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUNDED;
+import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUND_ERROR;
+
+public class EpdqNotificationService {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private final ChargeDao chargeDao;
+    private final SignatureGenerator signatureGenerator;
+    private final ChargeNotificationProcessor chargeNotificationProcessor;
+    private final RefundNotificationProcessor refundNotificationProcessor;
+
+    private static final String PAYMENT_GATEWAY_NAME = PaymentGatewayName.EPDQ.getName();
+
+    @Inject
+    public EpdqNotificationService(ChargeDao chargeDao,
+                                   SignatureGenerator signatureGenerator,
+                                   ChargeNotificationProcessor chargeNotificationProcessor,
+                                   RefundNotificationProcessor refundNotificationProcessor) {
+        this.chargeDao = chargeDao;
+        this.signatureGenerator = signatureGenerator;
+        this.chargeNotificationProcessor = chargeNotificationProcessor;
+        this.refundNotificationProcessor = refundNotificationProcessor;
+    }
+
+    @Transactional
+    public void handleNotificationFor(String payload) {
+        logger.info("Parsing {} notification", PAYMENT_GATEWAY_NAME);
+
+        EpdqNotification notification;
+        try {
+            notification = new EpdqNotification(payload);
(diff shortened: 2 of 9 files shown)


oswaldquek:
Can there be an explaining of how this is a refactor? It seems to be adding new functionality (the EpdqNotificationService) rather than refactoring existing functionality. At least it's not clear to me...

kbottla:
@oswaldquek Added more info to description. Hope this helps.

oswaldquek:
@kbottla thanks, that helped

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -0,0 +1,182 @@
...
+        Optional<ChargeEntity> maybeCharge = chargeDao.findByProviderAndTransactionId(PAYMENT_GATEWAY_NAME, notification.getTransactionId());
+
+        if (!maybeCharge.isPresent()) {
+            logger.error("{} notification {} could not be verified (associated charge entity not found)",
+                    PAYMENT_GATEWAY_NAME, notification);
+            return;
+        }
+
+        ChargeEntity charge = maybeCharge.get();
+        if (!verifyNotificationSignature(notification, charge)) {
```
oswaldquek: i think this line would read better as `if (!notificationSignatureVerififed(notification, charge)) {`

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -0,0 +1,182 @@
...
+                .stream()
+                .collect(Collectors.partitioningBy(p -> p.getName().equalsIgnoreCase(SHASIGN_KEY)))
+                .get(withShaSignature);
+    }
+
+    private String getShaOutPassphrase(ChargeEntity charge) {
+        return charge.getGatewayAccount().getCredentials().get(CREDENTIALS_SHA_OUT_PASSPHRASE);
+    }
+
+    static Optional<ChargeStatus> newChargeStateForChargeNotification(String notificationStatus, ChargeStatus chargeStatus) {
```
oswaldquek: why is this method not private? it's not used outside of this class

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -0,0 +1,182 @@
...
+                        return Optional.empty();
+                }
+            case EPDQ_PAYMENT_REQUESTED:
+                return Optional.of(CAPTURED);
+            default:
+                return Optional.empty();
+        }
+    }
+
+    static Optional<RefundStatus> newRefundStateForRefundNotification(String notificationStatus) {
```
oswaldquek: why is this method not private? it's not used outside of this class

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotification.java
```diff
@@ -26,9 +27,47 @@
...
+        @Override
+        public String toString() {
+            return this.name();
+        }
+    }
+
     private Optional<ChargeStatus> chargeStatus = Optional.empty();
 
-    public EpdqNotification(String payload) {
+    public class EpdqParseException extends Exception {
```
oswaldquek: Can this be moved to its own class? It feels like the responsibility of `EpdqNotification` is to hold information about the notification from epdq, not to hold exception classes.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
```diff
@@ -0,0 +1,129 @@
...
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.EXPIRE_CANCEL_SUBMITTED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.SYSTEM_CANCEL_SUBMITTED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.USER_CANCEL_SUBMITTED;
+import static uk.gov.pay.connector.charge.service.StatusFlow.EXPIRE_FLOW;
+import static uk.gov.pay.connector.charge.service.StatusFlow.SYSTEM_CANCELLATION_FLOW;
+import static uk.gov.pay.connector.charge.service.StatusFlow.USER_CANCELLATION_FLOW;
+import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUNDED;
+import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUND_ERROR;
+
+public class EpdqNotificationServiceStatusMapperTest {
```
oswaldquek: Are you aware of this test? https://github.com/alphagov/pay-connector/blob/32a0acca2b169c9c4ca34d2e8d54bcf45f293ba4/src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapperTest.java
It looks like there's an already existing `EpdqStatusMapper`: https://github.com/alphagov/pay-connector/blob/32a0acca2b169c9c4ca34d2e8d54bcf45f293ba4/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapper.java

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceTest.java
```diff
@@ -0,0 +1,163 @@
...
+        );
+        when(mockCharge.getStatus()).thenReturn(CAPTURE_APPROVED.getValue());
+        when(mockCharge.getGatewayAccount()).thenReturn(mockGatewayAccountEntity);
+        when(mockGatewayAccountEntity.getCredentials()).thenReturn(ImmutableMap.of(CREDENTIALS_SHA_OUT_PASSPHRASE, shaPhraseOut));
+
+        when(mockChargeDao.findByProviderAndTransactionId(EPDQ.getName(), payId)).thenReturn(Optional.of(mockCharge));
+    }
+
+    @Test
+    public void givenAChargeCapturedNotification_chargeNotificationProcessorInvokedWithNotificationAndCharge() throws EpdqNotification.EpdqParseException {
```
oswaldquek: `EpdqNotification.EpdqParseException` does not need to be thrown

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceTest.java
```diff
@@ -0,0 +1,163 @@
...
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_APPROVED;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.EPDQ;
+import static uk.gov.pay.connector.gateway.epdq.EpdqNotification.StatusCode.EPDQ_PAYMENT_REQUESTED;
+import static uk.gov.pay.connector.gateway.epdq.EpdqNotification.StatusCode.EPDQ_REFUND;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_SHA_OUT_PASSPHRASE;
+
+@RunWith(MockitoJUnitRunner.class)
+public class EpdqNotificationServiceTest {
+    private EpdqNotificationService notificationService;
```
oswaldquek: There doesn't seem to be any test where the notification signature is not verified.

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceITest.java
```diff
@@ -61,6 +64,42 @@ public void shouldHandleACaptureNotification(ChargeStatus chargeStatus) throws E
         assertFrontendChargeStatusIs(chargeId, CAPTURED.getValue());
     }
 
+    @Test
+    public void shouldHandleAnAuthorisedNotification_whenChargeIsInAuthorisationSubmittedState() throws Exception {
+        String transactionId = "transaction-id";
+        String chargeId = createNewChargeWith(AUTHORISATION_SUBMITTED, transactionId);
+        String epdqAuthorisedNotificationCode = "5";
```
oswaldquek: It would be clearer if `"5"` was replaced with `EpdqStatusMapper.EPDQ_AUTHORISED`

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceITest.java
```diff
@@ -61,6 +64,42 @@ public void shouldHandleACaptureNotification(ChargeStatus chargeStatus) throws E
...
+                .asString();
+        assertThat(response, is(RESPONSE_EXPECTED_BY_EPDQ));
+        assertFrontendChargeStatusIs(chargeId, AUTHORISATION_SUCCESS.getValue());
+    }
+
+    @Test
+    public void shouldHandleAnAuthorisedCancelledNotification_whenChargeIsInUserCancelSubmittedState() throws Exception {
+        String transactionId = "transaction-id";
+        String chargeId = createNewChargeWith(USER_CANCEL_SUBMITTED, transactionId);
+        String epdqAuthorisedCancelledNotificationCode = "6";
```
oswaldquek: It would be clearer if "6" was replaced with EpdqStatusMapper.EPDQ_AUTHORISED_CANCELLED

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
oswaldquek: this method might not be needed as there is https://github.com/alphagov/pay-connector/blob/32a0acca2b169c9c4ca34d2e8d54bcf45f293ba4/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapper.java

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
oswaldquek: this method might not be needed as there is https://github.com/alphagov/pay-connector/blob/32a0acca2b169c9c4ca34d2e8d54bcf45f293ba4/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapper.java

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
kbottla: Yes, I am aware of it. It was used in the previous handler model. We now have status mapping with in EpdqNotificationService which handles charge and refund statuses separately (`newChargeStateForChargeNotification(), newRefundStateForRefundNotification()`).

We will eventually be getting rid of existing unused code but in separate PR. But including it current PR will make it too big.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceTest.java
kbottla: 👍

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceITest.java
kbottla: 👍 Good spot. Updated, but to `EpdqNotification.StatusCode.EPDQ_AUTHORISED;`

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceITest.java
kbottla: Updated, but to `EpdqNotification.StatusCode.EPDQ_AUTHORISED_CANCELLED;`

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
kbottla: This is slightly inconvenient for me too but need to keep it package private to be able to add tests specific for Status mapping `EpdqNotificationServiceStatusMapperTest`. I am happy to change it if there is any better way to test status mapping only.
 
`EpdqStatusMapper` will be removed as it is no longer required by new EpdqNotificationService but in a separate PR.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
kbottla: This is slightly inconvenient for me too but need to keep it package private to be able to add tests specific for Status mapping `EpdqNotificationServiceStatusMapperTest`. I am happy to change it if there is any better way to test status mapping only.
 
`EpdqStatusMapper` will be removed as it is no longer required by new EpdqNotificationService but in a separate PR.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotification.java
kbottla: 👍

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
kbottla: I think `isValidNotificationSignature` defines it well.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceTest.java
kbottla: New test available `shouldNotUpdateIfShaPhraseExpectedIsIncorrect`

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
oswaldquek: So one way to do it is to extract the `newChargeStateForChargeNotification` and `newRefundStateForRefundNotification` into another class, just like the existing `EpdqStatusMapper`. Another way would be to test it by calling `EpdqNotificationService.handleNotificationFor` and verifying `chargeNotificationProcessor` was invoked with the correct new charge status.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
oswaldquek: ok, understood 👍

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
oswaldquek: i don't think it's good practice to make method non-private for the sake of testing

kbottla: Extracting into new classes will give same visibility as it is now. Doesn't make much difference.

testing `handleNotificationFor` seems to be an overkill to test different statuses exhaustively compared to current testing.

oswaldquek: it doesn't set a good precedent if we make methods non-private just because it will make testing easier... we should test things the right way even if it's more troublesome. we shouldn't modify production code just to make testing code easier.

kbottla: Package-private still encapsulates a level of detail. So personally won't be an issue to have  `static` methods as package-private given they don't have any state and also doesn't alter state. It all comes down to context. But to get out of the deadlock I have changed them to private 😄

mrlumbu: yeah, I don't think we need to be precious about private versus package private. Ultimately we're not writing library code here - if something is a tad more visible than it strictly needs to be, we're not going to be paying for it in 10 years time

oswaldquek (approved):

kbottla:
Thanks @oswaldquek

PR #861, opened 2018-11-22
PP-4399 remove cancel service functions

georgeracu:
## WHAT
Another step in refactoring `ChargeCancelService` to not use `CancelServiceFunctions`.

with @oswaldquek

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
@@ -1,31 +1,33 @@
 package uk.gov.pay.connector.charge.service;
 
 import com.google.common.collect.ImmutableList;
-import com.google.inject.Provider;
 import com.google.inject.persist.Transactional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
-import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
-import uk.gov.pay.connector.charge.service.transaction.TransactionContext;
-import uk.gov.pay.connector.charge.service.transaction.TransactionFlow;
-import uk.gov.pay.connector.charge.service.transaction.TransactionalOperation;
 import uk.gov.pay.connector.chargeevent.dao.ChargeEventDao;
+import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
+import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
+import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
 import uk.gov.pay.connector.gateway.PaymentProviders;
+import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseCancelResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 
 import javax.inject.Inject;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CREATED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
-import static uk.gov.pay.connector.charge.service.CancelServiceFunctions.doGatewayCancel;
-import static uk.gov.pay.connector.charge.service.CancelServiceFunctions.prepareForTerminate;
 import static uk.gov.pay.connector.charge.service.StatusFlow.SYSTEM_CANCELLATION_FLOW;
 import static uk.gov.pay.connector.charge.service.StatusFlow.USER_CANCELLATION_FLOW;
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -46,53 +43,63 @@
...
-                .map(doCancel(chargeId, SYSTEM_CANCELLATION_FLOW))
-                .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
+    @Transactional
+    public void doSystemCancel(String chargeId, Long accountId) {
+        final Optional<ChargeEntity> maybeCharge = chargeDao.findByExternalIdAndGatewayAccount(chargeId, accountId);
+        if (maybeCharge.isPresent()) {
+            doCancel(maybeCharge.get(), SYSTEM_CANCELLATION_FLOW);
+        } else {
+            throw new ChargeNotFoundRuntimeException(chargeId);
+        }
```
mrlumbu: I don't think we want to get rid of the mapping over optional pattern - to me it seems significantly nicer than your replacement, but even if you disagree, it will be a LOT of work to replace every instance, and we'll be stuck between two patterns.

georgeracu: The reasoning came from the fact that we return an `Optional<GatewayResponse<BaseCancelResponse>>` even if we don't call the gateway. This means that we need to have another object that will wrap a success or an error based on the answer we get from the gateway or from the local cancellation. On the other hand, we were breaking early by throwing an exception if the charge is not found at the service level.
The resource was returning the same `NoContent` no matter what the answer was from the gateway, so I didn't see any reason in actually returning an object that will be ignored. The functionality was kept, the error from the gateway is logged at the service level now.

mrlumbu: My point was just that the code you have written could be written like this:
```
chargeDao.findByExternalIdAndGatewayAccount(chargeId, accountId)
    .map(charge -> doCancel(charge, SYSTEM_CANCELLATION_FLOW))
    .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
```
which is how we have written similar code everywhere else, and I think is quite nice

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -40,74 +42,67 @@
...
 
     @Transactional
-    public void doSystemCancel(String chargeId, Long accountId) {
-        final Optional<ChargeEntity> maybeCharge = chargeDao.findByExternalIdAndGatewayAccount(chargeId, accountId);
-        if (maybeCharge.isPresent()) {
-            doCancel(maybeCharge.get(), SYSTEM_CANCELLATION_FLOW);
-        } else {
-            throw new ChargeNotFoundRuntimeException(chargeId);
-        }
+    public Optional<ChargeEntity> doSystemCancel(String chargeId, Long accountId) {
```
kbottla, line 56: I am under the impression we are trying to keep the scope of `<ChargeEntity>` within Service objects and not propagate to Resource. We are sending lot of information than needed to the Resource and none is used with in Resource method.

We could throw `ChargeNotFoundRuntimeException` here which won't break existing functionality and will be inline with other exceptions thrown (`IllegalStateRuntimeException`, ...) from the service.

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -40,74 +42,67 @@
...
 
     @Transactional
-    public void doUserCancel(String chargeId) {
-        final Optional<ChargeEntity> maybeCharge = chargeDao.findByExternalId(chargeId);
-        if (maybeCharge.isPresent()) {
-            doCancel(maybeCharge.get(), USER_CANCELLATION_FLOW);
-        } else {
-            throw new ChargeNotFoundRuntimeException(chargeId);
-        }
+    public Optional<ChargeEntity> doUserCancel(String chargeId) {
```
kbottla, line 65: Same as above for the object scope

kbottla (commented):
Looks good overall. Just a few comments on the way ChargeEntity is passed to Resource method.

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
georgeracu: We had the model you are suggesting before. After chatting with @mrlumbu we decided to let the resource decide how to deal with exceptions. In the end we were returning a 404 from the service level (by throwing `ChargeNotFoundRuntimeException` and then getting caught by the exception handler) while now we are making explicit that we return a 404 from the resource.

georgeracu: Dropwizard suggests both approaches to signal that something went wrong at the resource level: throw exception or return response. They are in favour of throwing exceptions over returning explicit response
`If at all possible, prefer throwing Exception instances to returning Response objects, as that will make resource endpoints more self describing and easier to test.`

kbottla: We will end up handling exceptions both in Service and Resource. Though I am more concerned about the ChargeEntity being available in Resource which is not being used at all except if it is available or not.

georgeracu: Just to make myself clear, I prefer being explicit about what you return. I don't like using exceptions for flow control and throwing them levels down in the application just to be caught by the exception mapper and translated into the same response that we explicitly write at the resource level. This to me feels like magic behind curtains. 
What would solve the problem of passing around an entity is a construct like `Either<Error, Success>` where we can put all the exceptions we get from the service and let the resource decide how to deal with them. Atm we don't use such a construct so we simulate it with the presence of `Optional`. 

I agree that we should not send the charge back to the resource, just to be ignored.

oswaldquek: if a charge doesn't exist, it's highly arguable that it constitutes as an exception at all. i don't think it's an application or system exception at all because we have an endpoint `/v1/api/accounts/{accountId}/charges/{chargeId}/cancel` that allows the caller to pass in any chargeId.

if we look at CancelChargeService's contract `public Optional<ChargeEntity> doSystemCancel(String chargeId, Long accountId)` it is perfectly reasonable to return an `Optional<ChargeEntity>` as the ChargeEntity might have changed as a result of it being canceled. so in my view that is ok. it just happens that the caller, CardResource doesn't want to do anything with it.

given that we've discussed this with @mrlumbu there has already been consensus about this approach so i think we should accept the solution here.

mrlumbu: Returning entities to the resource is fine until we have some better pattern for passing things around. It makes more sense to pass charge entity up than to say pass the gateway response up which is what we were doing before. The fact that the resource ignores the entity is a quirk of our api - it would actually make more sense if it returned it, as in all the dropwizard example code. So I think this is fine for now

kbottla (approved):
Approved as per team consensus 😄. We will revisit exception handling across code base and have unique pattern

PR #864, opened 2018-11-23
PP-4314 Add ApplePayService to authorise an apple pay charge

simad:
## WHAT
- Adding `ApplePayService` which decrypts a payload and then delegates to `AppleAuthoriseService` to authorise an apple pay charge. There is some duplication with CardResource on handling the response from the authorise service - another refactor will follow in which we embed that logic in services and move it out of the resource (also, using better practices like exception mappers).
- The logic is extremely similar to CardAuthoriseService - we can definitely extract commonalities / use one service. As this area is going through a big refactoring, I am introducing a bit of duplication in the hope of not interfering (right now) with it
- A charge is marked with WalletType of `APPLE_PAY` when authorised through this new service
- The tests are very verbose (are similar to ones in the card authorise service). The pending refactoring should help shrinking those tests.
- Simplified logic for SandboxCardNumbers

### DIFF
--- a/src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
@@ -47,21 +47,14 @@ public class AppleAuthoriseService extends CardAuthoriseBaseService<AppleDecrypt
         Optional<String> transactionId = extractTransactionId(chargeExternalId, operationResponse);
         ChargeStatus status = extractChargeStatus(operationResponse.getBaseResponse(), operationResponse.getGatewayError());
 
-        AuthCardDetails authCardDetailsToBePersisted = new AuthCardDetails();
-        authCardDetailsToBePersisted.setCardHolder(applePaymentData.getPaymentInfo().getCardholderName());
-        authCardDetailsToBePersisted.setCardNo(applePaymentData.getPaymentInfo().getLastDigitsCardNumber());
-        authCardDetailsToBePersisted.setPayersCardType(applePaymentData.getPaymentInfo().getCardType());
-        authCardDetailsToBePersisted.setCardBrand(applePaymentData.getPaymentInfo().getBrand());
-        authCardDetailsToBePersisted.setEndDate(applePaymentData.getApplicationExpirationDate().format(EXPIRY_DATE_FORMAT));
-        authCardDetailsToBePersisted.setCorporateCard(false);
-        ChargeEntity updatedCharge = chargeService.updateChargePostAuthorisation(
+        AuthCardDetails authCardDetailsToBePersisted = authCardDetailsFor(applePaymentData);
+        ChargeEntity updatedCharge = chargeService.updateChargePostApplePayAuthorisation(
                 chargeExternalId,
                 status,
                 transactionId,
                 Optional.empty(),
                 operationResponse.getSessionIdentifier(),
-                authCardDetailsToBePersisted,
-                Optional.of(WalletType.APPLE_PAY));
+                authCardDetailsToBePersisted);
 
         logger.info("Authorisation for {} ({} {}) for {} ({}) - {} .'. {} -> {}",
                 updatedCharge.getExternalId(), updatedCharge.getPaymentGatewayName().getName(),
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
```diff
@@ -0,0 +1,88 @@
...
+            String chargeExternalId,
+            ChargeStatus oldChargeStatus,
+            AppleDecryptedPaymentData applePaymentData,
+            GatewayResponse<BaseAuthoriseResponse> operationResponse) {
+
+        logger.info("Processing gateway auth response for apple pay");
+        Optional<String> transactionId = extractTransactionId(chargeExternalId, operationResponse);
+        ChargeStatus status = extractChargeStatus(operationResponse.getBaseResponse(), operationResponse.getGatewayError());
+
+        AuthCardDetails authCardDetailsToBePersisted = new AuthCardDetails();
```
mrlumbu: I think a converter method would be nice here? `AuthCardDetails.from(blah` or something?

src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
```diff
@@ -0,0 +1,88 @@
...
+        authCardDetailsToBePersisted.setEndDate(applePaymentData.getApplicationExpirationDate().format(EXPIRY_DATE_FORMAT));
+        authCardDetailsToBePersisted.setCorporateCard(false);
+        ChargeEntity updatedCharge = chargeService.updateChargePostAuthorisation(
+                chargeExternalId,
+                status,
+                transactionId,
+                Optional.empty(),
+                operationResponse.getSessionIdentifier(),
+                authCardDetailsToBePersisted,
+                Optional.of(WalletType.APPLE_PAY));
```
mrlumbu: Could we just overload `updateChargePostAuthorisation` or have a `updateChargePostAppleAuthorisation` method? Having an optional of a constant as an argument feels wrong (having any optional as an argument feels wrong, really)

src/test/java/uk/gov/pay/connector/applepay/AppleAuthoriseServiceTest.java
```diff
@@ -0,0 +1,366 @@
...
+import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
+import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
+import uk.gov.pay.connector.gateway.model.GatewayError;
+import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse.AuthoriseStatus;
+import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.gateway.model.response.GatewayResponse.GatewayResponseBuilder;
+import uk.gov.pay.connector.gateway.worldpay.WorldpayOrderStatusResponse;
+import uk.gov.pay.connector.paymentprocessor.service.CardExecutorService;
+import uk.gov.pay.connector.paymentprocessor.service.CardServiceTest;
+import uk.gov.pay.connector.util.AuthUtils;
```
mrlumbu: this is unused, and there are a few other unused imports around

src/test/java/uk/gov/pay/connector/applepay/AppleAuthoriseServiceTest.java
```diff
@@ -0,0 +1,366 @@
...
+        assertThat(charge.getStatus(), is(AUTHORISATION_SUCCESS.getValue()));
+        assertThat(charge.getGatewayTransactionId(), is(TRANSACTION_ID));
+        verify(mockedChargeEventDao).persistChargeEventOf(charge);
+        assertThat(charge.get3dsDetails(), is(nullValue()));
+        assertThat(charge.getCardDetails(), is(notNullValue()));
+        assertThat(charge.getCorporateSurcharge().isPresent(), is(false));
+    }
+
+    @Test
+    public void doAuthorise_shouldRespondAuthorisationSuccess_overridingGeneratedTransactionId() {
```
mrlumbu: don't think this is correctly named

src/test/java/uk/gov/pay/connector/applepay/AppleAuthoriseServiceTest.java
```diff
@@ -0,0 +1,366 @@
...
+
+        providerWillAuthorise();
+
```
mrlumbu: pedantry alert! Could we group things in the same way in each test? Makes it easier to figure out the differences

mrlumbu (changes requested):
Looking good. Have suggested some small changes

src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
simad: 👍 yup, I think would save a bit of code in many places

simad: no i lied, it's used only there. But well :D

simad: Putting it in a private method in the class itself, I wouldn't want to "pollute" the AuthCardDetails class with apple pay specific stuff, but it's debatable and I will move it if you don't like it

src/test/java/uk/gov/pay/connector/applepay/AppleAuthoriseServiceTest.java
simad: yes, this was a copy from the tests we have for the card auth service. The test itself is so verbose with all the mocks setups. I'll take a look

src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
simad: yup, added `updateChargePostAppleAuthorisation` as we do for 3ds

src/test/java/uk/gov/pay/connector/applepay/AppleAuthoriseServiceTest.java
simad: yes, it was actually not doing anything different from the basic "should authorise" test

src/test/java/uk/gov/pay/connector/applepay/AppleAuthoriseServiceTest.java
simad: 👍

mrlumbu (approved):

PR #866, opened 2018-11-23
PP-4373 Refactor authorisation services to remove base authorise service

mrlumbu:
This commit reconfigures the services that do something authorisation related,
so that they no longer inherit from a base class, but instead use that base
class as an injected dependency. This makes the services more flexible - they
no longer have to aobey the same contract, so can do what they actually need
to do. There is still some duplication or near duplication which I will
try to address in future PRs.
One thing I have done which is worth being aware of is moving all `@Transactional`
annotations into charge service - this feels like where they really belong,
but does lead to transaction nesting in some cases. From my understanding, Guice
basically ignores all but the top level of the transaction, so I think this is
fine.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
@@ -1,11 +1,16 @@
 package uk.gov.pay.connector.applepay;
 
+import com.codahale.metrics.MetricRegistry;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
 import io.dropwizard.setup.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.cardtype.dao.CardTypeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.gateway.PaymentProvider;
 import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
```diff
@@ -89,4 +116,8 @@ private AuthCardDetails authCardDetailsFor(AppleDecryptedPaymentData applePaymen
         return authCardDetails;
     }
 
+    private PaymentProvider<BaseAuthoriseResponse> getPaymentProviderFor(ChargeEntity chargeEntity) {
```
simad, line 119: I see you've brought back all of these private `getPaymentProviderFor` methods. Why do you think they bring any advantage over calling `providers.byName(chargeEntity.getPaymentGatewayName())` directly inline? (Genuine question! :D )

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseBaseService.java
```diff
@@ -27,38 +22,18 @@
...
-    protected MetricRegistry metricRegistry;
+    protected final Logger logger = LoggerFactory.getLogger(getClass());
 
-    protected CardAuthoriseBaseService(PaymentProviders providers,
-                             CardExecutorService cardExecutorService,
-                             ChargeService chargeService,
-                             Environment environment) {
-        this.providers = providers;
+    @Inject
+    public CardAuthoriseBaseService(CardExecutorService cardExecutorService) {
```
simad, line 30: nice :D

simad (approved):
Good progress! A question but will approve anyway - it can be addressed in another PR

src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
mrlumbu: Ha, the reason is that this was the easiest way to tell the compiler we are dealing with `PaymentProvider<BaseAuthoriseResponse>` as opposed to the raw type `PaymentProvider`. I think the only other way is to cast it, and I preferred this way. Not the greatest reason in the world - if you have any suggestions, lemme know.

mrlumbu: This changed because we no longer pass the type parameter into the service

PR #869, opened 2018-11-26
PP-4374 Introduced SmartpayNotificationService

kbottla:
## WHAT

Part of refactoring notification handling (see https://github.com/alphagov/pay-connector/pull/787 for the original full attempt...). Introducing gateway specific service objects to handle notifications instead of generalised notification handler (`NotificationService`). We don't need a generalised `NotificationService` as each gateway notification has dedicated notification resource endpoint. Having separate service object will make it easier to understand and follow through the path.  

- Added `SmartpayNotificationService` to delegate Smartpay notification processing to service instead of a generic handler.
- NotificationResource for Smartpay notifications now uses `SmartNotificationService`
- Also removed handleNotification() from NotificationResource as it is no longer required

Note that any existing code that handles Smartpay notifications still stays and will be removed in subsequent PRs

## HOW 
- Existing tests in `SmartpayNotificationResourceWithAccountSpecificAuthITest` should continue to pass
- All tests in `SmartpayNotificationServiceTest` should pass

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
@@ -1,0 +1,144 @@
+package uk.gov.pay.connector.gateway.smartpay;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.persist.Transactional;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.dao.ChargeDao;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.gateway.model.status.InterpretedStatus;
+import uk.gov.pay.connector.gateway.model.status.MappedChargeStatus;
+import uk.gov.pay.connector.gateway.model.status.MappedRefundStatus;
+import uk.gov.pay.connector.gateway.processor.ChargeNotificationProcessor;
+import uk.gov.pay.connector.gateway.processor.RefundNotificationProcessor;
+
+import javax.inject.Inject;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.SMARTPAY;
+
+public class SmartpayNotificationService {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private final ChargeDao chargeDao;
+    private final ChargeNotificationProcessor chargeNotificationProcessor;
+    private final RefundNotificationProcessor refundNotificationProcessor;
+
+    private static final String PAYMENT_GATEWAY_NAME = SMARTPAY.getName();
+
+    @Inject
+    public SmartpayNotificationService(ChargeDao chargeDao,
+                                       ChargeNotificationProcessor chargeNotificationProcessor,
+                                       RefundNotificationProcessor refundNotificationProcessor) {
+        this.chargeDao = chargeDao;
+        this.chargeNotificationProcessor = chargeNotificationProcessor;
+        this.refundNotificationProcessor = refundNotificationProcessor;
+    }
+
+    @Transactional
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private final ObjectMapper objectMapper;
+    private final ChargeDao chargeDao;
+    private final ChargeNotificationProcessor chargeNotificationProcessor;
+    private final RefundNotificationProcessor refundNotificationProcessor;
+
+    private static final String PAYMENT_GATEWAY_NAME = SMARTPAY.getName();
+
+    @Inject
+    public SmartpayNotificationService(ObjectMapper objectMapper, ChargeDao chargeDao,
```
georgeracu: `ObjectMapper` can be removed from the constructor and can be created when needed. It looks like is being used only in 1 place in this class.

src/main/java/uk/gov/pay/connector/webhook/resource/NotificationResource.java
```diff
@@ -45,7 +43,10 @@ public NotificationResource(WorldpayNotificationService worldpayNotificationServ
     @PermitAll
     @Path("/v1/api/notifications/smartpay")
     public Response authoriseSmartpayNotifications(String notification) {
-        return handleNotification("not-required", "smartpay", notification);
+        smartpayNotificationService.handleNotificationFor(notification);
+        String response = "[accepted]";
+        logger.info("Responding to notification from provider={} with 200 {}", "smartpay", response);
```
georgeracu: Inlining `smartpay` would make it simpler

src/test/java/uk/gov/pay/connector/util/TestTemplateResourceLoader.java
```diff
@@ -69,11 +69,11 @@
     public static final String SMARTPAY_VALID_CAPTURE_SMARTPAY_REQUEST = SMARTPAY_BASE_NAME + "/valid-capture-smartpay-request.xml";
 
     public static final String SMARTPAY_MULTIPLE_NOTIFICATIONS = SMARTPAY_BASE_NAME + "/multiple-notifications.json";
-    public static final String SMARTPAY_MULTIPLE_NOTIFICATIONS_DIFFERENT_DATES = SMARTPAY_BASE_NAME + "/multiple-notifications-different-dates.xml";
-    public static final String SMARTPAY_NOTIFICATION_AUTHORISATION = SMARTPAY_BASE_NAME + "/notification-authorisation.xml";
-    public static final String SMARTPAY_NOTIFICATION_CAPTURE_WITH_UNKNOWN_STATUS = SMARTPAY_BASE_NAME + "/notification-capture-with-unknown-status.xml";
-    public static final String SMARTPAY_NOTIFICATION_CAPTURE = SMARTPAY_BASE_NAME + "/notification-capture.xml";
-    public static final String SMARTPAY_NOTIFICATION_REFUND = SMARTPAY_BASE_NAME + "/notification-refund.xml";
+    public static final String SMARTPAY_MULTIPLE_NOTIFICATIONS_DIFFERENT_DATES = SMARTPAY_BASE_NAME + "/multiple-notifications-different-dates.json";
```
georgeracu, line 72: I wonder how were this working with `.xml` as I see in master they are `.json` 🤔

georgeracu (changes requested):
Nothing major, just a comment on removing a constructor argument that is not required to be injected. Nice one 👍

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
kbottla: 👍

src/test/java/uk/gov/pay/connector/util/TestTemplateResourceLoader.java
kbottla: These constants were never used and were hardcoded 
https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/smartpay/SmartpayNotificationResourceWithAccountSpecificAuthITest.java#L227

src/main/java/uk/gov/pay/connector/webhook/resource/NotificationResource.java
kbottla: 👍

georgeracu (approved):
LGTM

PR #872, opened 2018-11-26
PP-4374 Remove Notifications from PaymentProvider

kbottla:
## WHAT

Part of refactoring notification handling (see https://github.com/alphagov/pay-connector/pull/787 for the original full attempt...). 

- Notifications related code removed from PaymentProvider interface and sub classes as it is no longer required ( Each payment provider now has separate Notification service to handle notifications)


georgeracu (approved):
LGTM 👍

PR #887, opened 2018-11-29
PP-4375 Validate Stripe notifications

kbottla:
## WHAT
Validate notification payload received from Stripe using stripe-java library. Signature is available in notification request header ('Stripe-Signature') from Stripe.

For info :

Format of the signature header is:
`t=1492774577, v1=5257a869e7ec,v0=aba816edb039989a39`

**where**
t - timestamp. Tolerance (suggested by Stripe) for timestamp is 5 minutes. Any payload with time difference above tolerance is invalid
v1 - scheme which has valid signature. Multiple values are possible when secret keys are rotated. So payload is valid if at least one signature is valid.
v0 - is ignored and sent only for test mode events

See Stripe docs for more details :
https://stripe.com/docs/webhooks/signatures

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
@@ -4,8 +4,11 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
 import com.google.inject.persist.Transactional;
+import com.stripe.exception.SignatureVerificationException;
+import com.stripe.net.Webhook;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.StripeGatewayConfig;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
@@ -14,6 +17,7 @@ import uk.gov.pay.connector.gateway.processor.ChargeNotificationProcessor;
 import uk.gov.pay.connector.gateway.stripe.json.StripeSourcesResponse;
 import uk.gov.pay.connector.gateway.stripe.response.StripeNotification;
 
+import javax.ws.rs.WebApplicationException;
 import java.util.List;
 import java.util.Optional;
 
@@ -28,6 +32,7 @@ import static uk.gov.pay.connector.gateway.stripe.StripeNotificationType.SOURCE_
 
 public class StripeNotificationService {
 
+
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
     private final List<StripeNotificationType> sourceTypes = ImmutableList.of(SOURCE_CANCELED, SOURCE_CHARGEABLE, SOURCE_FAILED);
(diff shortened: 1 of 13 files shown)


src/test/java/uk/gov/pay/connector/gateway/stripe/util/StripeNotificationUtilTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        if (timestamp != null && timestamp.length > 0) {
+            currentTimestamp = timestamp[0];
+        }
+
+        final String payloadToSign = String.format("%d.%s", currentTimestamp, message);
+        String signature = null;
+        try {
+            signature = StripeHmacSha256SignatureGenerator.computeHmacSha256(webhookSigningSecret, payloadToSign);
+        } catch (Exception e) {
```
oswaldquek: since StripeNotificationUtilTest is just a test, we don't need to catch exceptions here. It's ok to just the test throw an exception.

oswaldquek (changes requested):
Just one comment at the moment. Awaiting for the PR to be updated with usage of stripe-java library.

src/test/java/uk/gov/pay/connector/gateway/stripe/util/StripeNotificationUtilTest.java
kbottla: That is intentional and you may see the same in updated PR. compute function throws checked exceptions and we will end up declaring the exceptions in every method and above if not caught explicitly.

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeNotificationResourceITest.java
```diff
@@ -132,14 +133,36 @@ public void shouldFailAStripeNotificationWithAnUnexpectedContentType() {
...
         return given()
                 .port(testContext.getPort())
                 .body(payload)
+                .header("Stripe-Signature", header)
                 .contentType(APPLICATION_JSON)
                 .post(NOTIFICATION_PATH);
     }
 
+    private Response notifyConnector(String payload) {
+        return notifyConnectorWithHeader(payload, StripeNotificationUtilTest.generateSigHeader(
```
oswaldquek, line 162: we could use `com.stripe.net.Webhook.Util.computeHmacSha256` instead of `StripeNotificationUtilTest.generateSigHeader`? this would be better as we can be sure we're using what Stripe uses to generate the header.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -57,16 +63,21 @@ public void setup() {
...
                 sourceId, SOURCE_CHARGEABLE);
-        notificationService.handleNotificationFor(payload);
+
+        notificationService.handleNotificationFor(payload, signPayload(payload));
 
         verify(mockChargeNotificationProcessor).invoke(sourceId, mockCharge, AUTHORISATION_3DS_READY, null);
     }
 
+    private String signPayload(String payload) {
+        return StripeNotificationUtilTest.generateSigHeader(webhookSigningSecret, payload);
```
oswaldquek, line 73: we could use `com.stripe.net.Webhook.Util.computeHmacSha256` instead of `StripeNotificationUtilTest.generateSigHeader`? this would be better as we can be sure we're using what Stripe uses to generate the header.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -28,28 +32,37 @@
...
+        this.stripeGatewayConfig = stripeGatewayConfig;
     }
 
     @Transactional
-    public void handleNotificationFor(String payload) {
+    public void handleNotificationFor(String payload, String signatureHeader) {
         logger.info("Parsing {} notification", PAYMENT_GATEWAY_NAME);
 
+        if (!isValidNotificationSignature(payload, signatureHeader)) {
+            throw new WebApplicationException(String.format("Invalid notification signature from %s [%s]", PAYMENT_GATEWAY_NAME, signatureHeader));
```
oswaldquek, line 63: Think we need to add a 400 status when constructing this WebApplicationException. if you look at https://stripe.com/docs/webhooks/signatures#verify-official-libraries the example Java code returns a 400 back to Stripe in the event of an invalid signature.

oswaldquek (changes requested):

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeNotificationResourceITest.java
kbottla: They are different
- computeHmacSha256 : will only encodes payload
- generateSigHeader : encodes payload + puts it in the format required (t=123123,v1=dfsddfgdf)

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
kbottla: They are different
- computeHmacSha256 : will only encodes payload
- generateSigHeader : encodes payload + puts it in the format required (t=123123,v1=dfsddfgdf)

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
kbottla: Any status is fine. Stripe doesn't really care what we send as long as it is non 2xx

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeNotificationResourceITest.java
oswaldquek: ah ok 👍

oswaldquek (approved):

PR #891, opened 2018-12-03
PP-4129 Calculate total amount depending on corporate surcharge for email notification

SandorArpa:
## WHAT

- calculate total amount when corporate card surcharge is present
- update relevant test

### DIFF
--- a/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
@@ -8,6 +8,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.util.CorporateCardSurchargeCalculator;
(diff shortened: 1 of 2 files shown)


alexbishop1 (changes requested):
Logic looks fine but I imagine we want to reuse the `CorporateSurchargeCalculator` here if possible

src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
```diff
@@ -157,7 +158,10 @@ private void readEmailConfig(ConnectorConfiguration configuration) {
 
         map.put("serviceReference", charge.getReference().toString());
         map.put("date", DateTimeUtils.toUserFriendlyDate(charge.getCreatedDate()));
-        map.put("amount", formatToPounds(charge.getAmount()));
+        Long amount = charge.getCorporateSurcharge()
+                .map(corporateSurcharge -> CorporateCardSurchargeCalculator.getTotalAmountFor(charge))
+                .orElse(charge.getAmount());
```
alexbishop1: I think we can just do `CorporateCardSurchargeCalculator.getTotalAmountFor(charge)`: it will return the correct total if there isn’t a surcharge

alexbishop1 (approved):

alexbishop1 (changes requested):
I pressed _Approve_ by mistake!

alexbishop1 (approved):

PR #892, opened 2018-12-03
bau: upgrade pay-java-commons

oswaldquek:
pay-java-commons uses mockserver 5.5.0, so it had to be upgraded here as well.

@oswaldquek


georgeracu (approved):
LGTM

PR #893, opened 2018-12-03
BAU: Do not run dependency-check maven plugin in verify phase

alexbishop1:
By default, the dependency-check plugin is tied to the `verify` phase (when used as a build plugin). We don’t want this; we just want to run it manually when required. So we define a property called `dependency-check.skip`, set it to `true`, and use this property’s value to set the dependency-check plugin’s own `skip` property. To execute the plugin manually, override the `dependency-check.skip` property:

```
$ mvn dependency-check:check -Ddependency-check.skip=false
```

Also upgrade from version 3.3.4 to 4.0.0.


georgeracu (approved):
LGTM

PR #894, opened 2018-12-04
BAU: Replace postgres container code with pay-java-commons

rhowe-gds:
pay-java-commons has code for spinning up a postgres container in order to run
tests against it. Use that instead of the copies in here.


georgeracu (approved):
LGTM

PR #895, opened 2018-12-04
PP-4375 Stripe Notification - 3DS authorisation

kbottla:
## WHAT
When a user completes 3DS authorisation using Stripe, the status of the transaction is not available immediately. Stripe sends the status using webhooks(notifications) and charge will be in AUTHORISATION_3DS_REQUIRED / AUTHORISATION_3DS_READY state until such notification is available. Once a source notification is available, the status is used to update charge or authorise the same with Stripe.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -275,10 +275,16 @@ public class ChargeService {
                                                          OperationType operationType,
                                                          Optional<String> transactionId) {
         return chargeDao.findByExternalId(chargeExternalId).map(charge -> {
-            charge.setStatus(status);
-            setTransactionId(charge, transactionId);
-            chargeEventDao.persistChargeEventOf(charge);
-
+            try {
+                charge.setStatus(status);
+                setTransactionId(charge, transactionId);
+                chargeEventDao.persistChargeEventOf(charge);
+            } catch (InvalidStateTransitionException e) {
+                if (chargeIsInLockedStatus(operationType, charge)) {
+                    throw new OperationAlreadyInProgressRuntimeException(operationType.getValue(), charge.getExternalId());
+                }
+                throw new IllegalStateRuntimeException(charge.getExternalId());
+            }
             return charge;
         }).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeExternalId));
     }
@@ -322,6 +328,11 @@ public class ChargeService {
         }).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
     }
 
+    public ChargeEntity findChargeById(String chargeId) {
+        return chargeDao.findByExternalId(chargeId)
+                .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
+    }
+
     private ChargeEntity updateChargeStatus(ChargeEntity chargeEntity, ChargeStatus chargeStatus) {
         chargeEntity.setStatus(chargeStatus);
         chargeEventDao.persistChargeEventOf(chargeEntity);
@@ -332,6 +343,10 @@ public class ChargeService {
         return chargeDao.findByExternalId(chargeId).map(chargeEntity ->
                 updateChargeStatus(chargeEntity, chargeStatus)
         ).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
+    }
+
+    public Optional<ChargeEntity> findByProviderAndTransactionId(String paymentGatewayName, String transactionId) {
+        return chargeDao.findByProviderAndTransactionId(paymentGatewayName, transactionId);
     }
 
     public ChargeEntity markChargeAsEligibleForCapture(String externalId) {
@@ -388,7 +403,7 @@ public class ChargeService {
     private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
         return authCardDetails.getCardNo().length() > 6;
     }
-    
+
     private TokenEntity createNewChargeEntityToken(ChargeEntity chargeEntity) {
         TokenEntity token = TokenEntity.generateNewTokenFor(chargeEntity);
         tokenDao.persist(token);
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
@@ -129,6 +129,7 @@ public class StripeNotificationService {
         }
 
         return Optional.empty();
+        
     }
 
     private StripeNotification parseNotification(String payload) throws StripeParseException {
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
@@ -3,19 +3,20 @@ package uk.gov.pay.connector.gateway.stripe;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
-import com.google.inject.persist.Transactional;
 import com.stripe.exception.SignatureVerificationException;
 import com.stripe.net.Webhook;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
-import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
-import uk.gov.pay.connector.gateway.processor.ChargeNotificationProcessor;
+import uk.gov.pay.connector.gateway.model.Auth3dsDetails;
 import uk.gov.pay.connector.gateway.stripe.json.StripeSourcesResponse;
 import uk.gov.pay.connector.gateway.stripe.response.StripeNotification;
+import uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService;
 
 import javax.ws.rs.WebApplicationException;
 import java.util.List;
@@ -24,21 +25,20 @@ import java.util.Optional;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_CANCELLED;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
 import static uk.gov.pay.connector.gateway.stripe.StripeNotificationType.SOURCE_CANCELED;
 import static uk.gov.pay.connector.gateway.stripe.StripeNotificationType.SOURCE_CHARGEABLE;
 import static uk.gov.pay.connector.gateway.stripe.StripeNotificationType.SOURCE_FAILED;
+import static uk.gov.pay.connector.paymentprocessor.model.OperationType.AUTHORISATION_3DS;
 
 public class StripeNotificationService {
-
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
     private final List<StripeNotificationType> sourceTypes = ImmutableList.of(SOURCE_CANCELED, SOURCE_CHARGEABLE, SOURCE_FAILED);
+    private final List<ChargeStatus> threeDSAuthorisableStates = ImmutableList.of(AUTHORISATION_3DS_REQUIRED, AUTHORISATION_3DS_READY);
 
-    private final ChargeDao chargeDao;
-    private final ChargeNotificationProcessor chargeNotificationProcessor;
+    private final ChargeService chargeService;
+    private final Card3dsResponseAuthService card3dsResponseAuthService;
     private final ObjectMapper objectMapper;
     private final StripeGatewayConfig stripeGatewayConfig;
 
@@ -46,16 +46,15 @@ public class StripeNotificationService {
     private static final long DEFAULT_TOLERANCE = 300L;
 
     @Inject
-    public StripeNotificationService(ChargeDao chargeDao,
-                                     ChargeNotificationProcessor chargeNotificationProcessor,
+    public StripeNotificationService(Card3dsResponseAuthService card3dsResponseAuthService,
+                                     ChargeService chargeService,
                                      StripeGatewayConfig stripeGatewayConfig) {
-        this.chargeDao = chargeDao;
-        this.chargeNotificationProcessor = chargeNotificationProcessor;
+        this.card3dsResponseAuthService = card3dsResponseAuthService;
+        this.chargeService = chargeService;
         objectMapper = new ObjectMapper();
         this.stripeGatewayConfig = stripeGatewayConfig;
     }
 
-    @Transactional
     public void handleNotificationFor(String payload, String signatureHeader) {
         logger.info("Parsing {} notification", PAYMENT_GATEWAY_NAME);
 
@@ -88,7 +87,7 @@ public class StripeNotificationService {
 
             logger.info("Evaluating {} for source notification [{}]", PAYMENT_GATEWAY_NAME, stripeSourcesResponse.getTransactionId());
 
-            Optional<ChargeEntity> maybeCharge = chargeDao.findByProviderAndTransactionId(PAYMENT_GATEWAY_NAME, stripeSourcesResponse.getTransactionId());
+            Optional<ChargeEntity> maybeCharge = chargeService.findByProviderAndTransactionId(PAYMENT_GATEWAY_NAME, stripeSourcesResponse.getTransactionId());
 
             if (!maybeCharge.isPresent()) {
                 logger.error("{} notification for source [{}] could not be verified (associated charge entity not found)",
@@ -98,13 +97,35 @@ public class StripeNotificationService {
 
             ChargeEntity charge = maybeCharge.get();
 
-            final Optional<ChargeStatus> newChargeStatus = newChargeStateForSourceNotification(notification.getType(), ChargeStatus.fromString(charge.getStatus()));
-
-            if (newChargeStatus.isPresent()) {
-                chargeNotificationProcessor.invoke(stripeSourcesResponse.getTransactionId(), charge, newChargeStatus.get(), null);
+            if (isChargeIn3DSRequiredOrReadyState(ChargeStatus.fromString(charge.getStatus()))) {
+                authorise3DSSource(charge, notification.getType());
             }
+
         } catch (StripeParseException e) {
             logger.error("{} notification parsing for source object failed: {}", PAYMENT_GATEWAY_NAME, e);
+        }
+    }
+
+    private void authorise3DSSource(ChargeEntity charge, String notificationEventType) {
+        try {
+            final StripeNotificationType type = StripeNotificationType.byType(notificationEventType);
+
+            if (AUTHORISATION_3DS_REQUIRED.equals(ChargeStatus.fromString(charge.getStatus()))) {
+                logger.info("Updating charge for notification - " +
+                                "charge_external_id={}, status={}, status_to={}, transaction_id={}, " +
+                                "account_id={}, provider={}, provider_type={}",
+                        charge.getExternalId(), charge.getStatus(), charge.getStatus(), charge.getGatewayTransactionId(),
+                        charge.getGatewayAccount().getId(), charge.getGatewayAccount().getGatewayName(), charge.getGatewayAccount().getType());
+                chargeService.lockChargeForProcessing(charge.getExternalId(), AUTHORISATION_3DS);
+            }
+
+            Auth3dsDetails auth3dsDetails = new Auth3dsDetails();
+            auth3dsDetails.setAuth3dsResult(getMappedAuth3dsResult(type));
+            card3dsResponseAuthService.process3DSecureAuthorisationWithoutLocking(charge.getExternalId(), auth3dsDetails);
+        } catch (OperationAlreadyInProgressRuntimeException e) {
+            // CardExecutorService is asynchronous and sends back 'OperationAlreadyInProgressRuntimeException' 
+            // exception while the charge is being authorised. Catch this exception to send a response with 
+            // http status 200 instead of depending on the status returned by Exception 
         }
     }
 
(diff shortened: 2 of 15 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -98,37 +100,63 @@ private void processSourceNotification(StripeNotification notification) {
...
         } catch (StripeParseException e) {
             logger.error("{} notification parsing for source object failed: {}", PAYMENT_GATEWAY_NAME, e);
         }
     }
 
+    private void authorise3DSSource(ChargeEntity charge, String notificationEventType) {
+        try {
+            final StripeNotificationType type = StripeNotificationType.byType(notificationEventType);
+
+            if (AUTHORISATION_3DS_REQUIRED.equals(ChargeStatus.fromString(charge.getStatus()))) {
```
mrlumbu, line 113: is this not just locking the charge? Don't we already have a method for that in charge service

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -24,38 +26,38 @@
...
 
     private static final String PAYMENT_GATEWAY_NAME = PaymentGatewayName.STRIPE.getName();
     private static final long DEFAULT_TOLERANCE = 300L;
 
     @Inject
-    public StripeNotificationService(ChargeDao chargeDao,
-                                     ChargeNotificationProcessor chargeNotificationProcessor,
+    public StripeNotificationService(Card3dsResponseAuthService card3dsResponseAuthService,
+                                     ChargeService chargeService,
+                                     ChargeDao chargeDao,
```
mrlumbu: ideally shouldn't be accessing ChargeDao from here

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -144,6 +144,59 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+    private Gateway3DSAuthorisationResponse authorise3DSSource(Auth3dsResponseGatewayRequest request) {
+        try {
+            Response authorisationResponse = createChargeFor3DSSource(request, request.getTransactionId().get());
+            StripeAuthorisationResponse stripeAuthResponse = new StripeAuthorisationResponse(authorisationResponse.readEntity(StripeCreateChargeResponse.class));
+
+            return Gateway3DSAuthorisationResponse.of(stripeAuthResponse.authoriseStatus(), stripeAuthResponse.getTransactionId());
+        } catch (GatewayClientException e) {
+
+            Response response = e.getResponse();
+            int status = response.getStatus();
```
mrlumbu: not sure why we need to do this

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -144,6 +144,59 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+            }
+
+            StripeErrorResponse stripeErrorResponse = response.readEntity(StripeErrorResponse.class);
+            String errorId = UUID.randomUUID().toString();
+            logger.error("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
+                    request.getChargeExternalId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
+
+            return Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.REJECTED);
+
+        } catch (DownstreamException e) {
```
mrlumbu: think the logging here is inaccurate - should be 3D authorisation?
Also the logging could be done in StripeGatewayClient - then the handling of these two exceptions is identical

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
kbottla: 👍

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
kbottla: 👍 yeah could use locking function, but still need to be invoked conditionally. As locking function throws exception (`OperationAlreadyInProgressRuntimeException`) if charge is already is locking state

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: Good spot. Removed in 3DS authorise as well as authorise.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: 👍 Moved logging and combined both exceptions

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -322,18 +328,28 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
...
+                .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
+    }
+
     private ChargeEntity updateChargeStatus(ChargeEntity chargeEntity, ChargeStatus chargeStatus) {
         chargeEntity.setStatus(chargeStatus);
         chargeEventDao.persistChargeEventOf(chargeEntity);
         return chargeEntity;
     }
 
+    @Transactional
```
IreneLau-GDS: But this is already being called by a method that has the transactional annotation. Are we doing it just to be extra sure? It's sister method doesn't have that the transactional annotation.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -98,37 +97,58 @@ private void processSourceNotification(StripeNotification notification) {
...
+                                "charge_external_id={}, status={}, status_to={}, transaction_id={}, " +
+                                "account_id={}, provider={}, provider_type={}",
+                        charge.getExternalId(), charge.getStatus(), charge.getStatus(), charge.getGatewayTransactionId(),
+                        charge.getGatewayAccount().getId(), charge.getGatewayAccount().getGatewayName(), charge.getGatewayAccount().getType());
+                chargeService.lockChargeForProcessing(charge.getExternalId(), AUTHORISATION_3DS);
+            }
+
+            Auth3dsDetails auth3dsDetails = new Auth3dsDetails();
+            auth3dsDetails.setAuth3dsResult(getMappedAuth3dsResult(type));
+            card3dsResponseAuthService.process3DSecureAuthorisationWithoutLocking(charge.getExternalId(), auth3dsDetails);
```
IreneLau-GDS, line 124: I'm not fully understanding what is going on between here and line 119. 

This whole process (as in further up the stack) is transactional. We then do multiple update to the charge in the database - first lock it then look up the change again later and update it???

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -98,37 +97,58 @@ private void processSourceNotification(StripeNotification notification) {
...
+            case SOURCE_CHARGEABLE:
+                return Auth3dsDetails.Auth3dsResult.AUTHORISED.toString();
+            case SOURCE_CANCELED:
+                return Auth3dsDetails.Auth3dsResult.CANCELED.toString();
+            case SOURCE_FAILED:
+                return Auth3dsDetails.Auth3dsResult.DECLINED.toString();
+            default:
+                Auth3dsDetails.Auth3dsResult.ERROR.toString();
         }
+        return Auth3dsDetails.Auth3dsResult.ERROR.toString();
```
IreneLau-GDS: If you get line 145 to return then you won't need this line.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -136,14 +135,60 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+                case DECLINED:
+                    return Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.REJECTED);
+                case AUTHORISED:
+                    return authorise3DSSource(request);
+            }
+        }
+
+        // if Auth3DSResult is not available, return response as AUTH_3DS_READY
+        // (to keep the transaction in auth waiting until a notification is received from Stripe for 3DS authorisation)
+        return Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.AUTH_3DS_READY);
```
IreneLau-GDS, line 163: Questions...
What would the charge status be at this point? 
What if a notification is never received?

The reason why I ask is... would the charge be stuck in a status that never gets cleaned up if a notification never arrives?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
kbottla: We need this line (otherwise it will result in compile error) and may not need 145. But it is just a code smell without default clause

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: Charge state will be 'AUTHORISATION_3DS_READY'. 
If a notification is never received, it will be cleaned up by charge expiry service

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: Removed annotation.
Added it to be used by another non-transactional method which was removed later. So this is not adding any value now.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
kbottla: Yes it does multiple updates depending on charge status. 

If charge is not already locked (frontend can lock the charge before a notification is received) it is locked and then charge is authorised. Depending on the authorisation status, it is updated to new state.

kbottla: Also 
- locking step is synchronous
- looking up charge and authorise/cancel/so on and update charge is asynchronous

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
IreneLau-GDS: Not if you do:

```
            default:
                return Auth3dsDetails.Auth3dsResult.ERROR.toString();
```

IreneLau-GDS: For completeness' sake:
```
    private String getMappedAuth3dsResult(StripeNotificationType type) {
        switch (type) {
            case SOURCE_CHARGEABLE:
                return Auth3dsDetails.Auth3dsResult.AUTHORISED.toString();
            case SOURCE_CANCELED:
                return Auth3dsDetails.Auth3dsResult.CANCELED.toString();
            case SOURCE_FAILED:
                return Auth3dsDetails.Auth3dsResult.DECLINED.toString();
            default:
                return Auth3dsDetails.Auth3dsResult.ERROR.toString();
        }
    }
```

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: I don't think it does because in `ChargeExpiryService` we have:
```
    public static final List<ChargeStatus> EXPIRABLE_REGULAR_STATUSES = ImmutableList.of(
            CREATED,
            ENTERING_CARD_DETAILS,
            AUTHORISATION_3DS_REQUIRED,
            AUTHORISATION_SUCCESS);
```
for statuses that we'll mark as expired and
```
    static final List<ChargeStatus> GATEWAY_CANCELLABLE_STATUSES = Arrays.asList(
            AUTHORISATION_SUCCESS,
            AWAITING_CAPTURE_REQUEST
    );
```
for statuses that we will call out to the gateway for real cancellation.

You might want to add AUTHORISATION_3DS_READY to the `EXPIRABLE_REGULAR_STATUSES` list in `ChargeExpiryService` and add a test to cover that?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
kbottla: oh blimey❗️how did I miss return on default. Updated now 😄

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: Good catch. Added AUTHORISATION_3DS_READY to EXPIRABLE_REGULAR_STATUSES.

There is a test already ```ChargeExpiryServiceTest.shouldExpireChargesWithoutCallingProviderToCancel``` which failed for this change due to invalid transition. So allowed new transition now
```PaymentGatewayStateTransitions : graph.putEdgeValue(AUTHORISATION_3DS_READY, EXPIRED, "ChargeExpiryService");```

File: src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java [hunk 9]
```diff
@@ -136,14 +135,60 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+            Response authorisationResponse = createChargeFor3DSSource(request, request.getTransactionId().get());
+            StripeAuthorisationResponse stripeAuthResponse = new StripeAuthorisationResponse(authorisationResponse.readEntity(StripeCreateChargeResponse.class));
+
+            return Gateway3DSAuthorisationResponse.of(stripeAuthResponse.authoriseStatus(), stripeAuthResponse.getTransactionId());
+        } catch (GatewayClientException e) {
+
+            Response response = e.getResponse();
+
+            if (response.getStatus() == HttpStatus.SC_UNAUTHORIZED) {
+                Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.ERROR);
```
IreneLau-GDS: Shouldn't this guy be returned? Otherwise this whole `if` block doesn't do anything. Probably means we haven't had a test that cover this path.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -136,14 +135,60 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+        } catch (GatewayClientException e) {
+
+            Response response = e.getResponse();
+
+            if (response.getStatus() == HttpStatus.SC_UNAUTHORIZED) {
+                return Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.ERROR);
+            }
+
+            StripeErrorResponse stripeErrorResponse = response.readEntity(StripeErrorResponse.class);
+            String errorId = UUID.randomUUID().toString();
```
IreneLau-GDS: We generate an error ID here, what is the purpose of that? It doesn't seem to go anywhere but gets logged to the log file.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: oops! don't know how I missed it 😄 . Corrected and added test `StripePaymentProviderTest.shouldMark3DSChargeAsError_whenGatewayOperationResultedInUnauthorisedException` to cover it

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -136,14 +135,60 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+            logger.error("3DS Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
+                    request.getChargeExternalId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
+
+            return Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.REJECTED);
+
+        } catch (DownstreamException | GatewayException e) {
+            return Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.EXCEPTION);
```
IreneLau-GDS, line 187: I think it would be useful to log the exception here because otherwise all we know is at the consumer level that there is an exception but what kind and more importantly, where in the stack...

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -260,21 +311,34 @@ private String threeDSecurePayload(CardAuthorisationGatewayRequest request, Stri
...
 
         params.add(new BasicNameValuePair("destination[account]", stripeAccountId));
         return URLEncodedUtils.format(params, UTF_8);
     }
 
+    private String getStripeAccountId(GatewayAccountEntity gatewayAccount) {
+        String stripeAccountId = gatewayAccount.getCredentials().get("stripe_account_id");
+
+        if (StringUtils.isBlank(stripeAccountId))
+            throw new WebApplicationException(format("There is no stripe_account_id for gateway account with id %s", gatewayAccount.getId()));
```
IreneLau-GDS, line 339: Could this ever happen? If so, I feel like we'll need the information of which charge is also affected by this so that we can do what is needed to rectify the now stuck status of those changes.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -37,26 +38,33 @@
 public class StripeNotificationServiceTest {
     private StripeNotificationService notificationService;
 
+    @Mock
+    private Card3dsResponseAuthService mockCard3dsResponseAuthService;
     @Mock
     private ChargeDao mockChargeDao;
```
IreneLau-GDS: This is now unused.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: Slightest chance, if it is overwritten or not setup (as it is also used during authorisation phase). I think logs will show this already up the stack as it is `WebApplicationException`

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: We log both these exceptions here https://github.com/alphagov/pay-connector/blob/9934c2f7a0b583035e7cb93306ce6a82b2f95561/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java#L75 and https://github.com/alphagov/pay-connector/blob/9934c2f7a0b583035e7cb93306ce6a82b2f95561/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java#L54

It is common for all gateway operations

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: I mean as in which charge(s) because we are not processing the notifications of those charges so they'd be stuck in certain state and ultimately gets marked as expired. Since the log won't have the charge Id of the original thing, it'd be hard for us to piece it together by looking at the logs. 

Sure we'll eventually look at Stripe Connect and probably find out from there but in terms of support workflow people will look at the logs which are already very hard to trace because of the amount of in-between log messages. It'd be really useful if we have something like `Unable to process notification for chargeId blahblahblahblah because there is no stripe_account_id for gateway account with Id 12345`.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: 👍

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
oswaldquek: @IreneLau-GDS this was actually my idea. It would be useful in the case where we only provide a summary back to the client but log the error in greater detail, which a person on support might want to look into. This is only useful if the errorId is passed back in the response, but we don't.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
kbottla: 👍 Added log

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: When I was on support I find the most useful thing is the external payment Id. This is because all I have to search is that single string then I'll get all the events that happened to it.

I feel that even if we pass the error Id, it'd mean the support person needs to do two searches. First search to find the charge Id using the error Id then search again for all the events.

I am talking like a trader now, but sometimes an extra click really is painful. This ZD tix 3477859 is an example. I was given a list of references and I had to use them to get the external payment Id. That was... painful.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
kbottla: 🥇 Removed

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: Anyways - it doesn't go anywhere, so shall we just not do it? I feel that if we log another thing that doesn't come from or go anywhere will just confuse people...

kbottla: I think having an errorId is a widely used practice for good reasons. We need to explore on how we can propagate back to response. 
- having only chargeId means, search brings up everything happened through out its lifecycle.
- external payment Id is not always available. (ie., if charge is still non-gateway related states, or some gateway/(s) operations doesn't always return paymentId, and especially in stripe - payment id is different for preauthorise and authorise )
- errorId - we can track particular transaction in question based on Id. It can filter out all other charge processing states not relevant.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: In...? Sorry I can't find it.

kbottla: May need a refresh.
https://github.com/alphagov/pay-connector/pull/895/commits/172b11d0ac7cb3815d8f972d4f9668e41bba196e#diff-88deef0983d769bc800a12345ea17fceR338

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: I understand that it is for a good reason but only if we bubble it all the way up to a response that ultimately shows up in the user's browser so that the user can call up the support line / submit a help request and say - hey I have this error Id, can you see what's going on.

However it is not being propagated and it is a UUID which only appears in this log message and only once. 

If we are going to explore how we can propagate back to the response and land on the user's browser then can we have a new story and actually implement the logic there. 

Being on support and having to look at the amount of log that we have is extremely painful. This team knows from the code that this is a UUID we generates because we wrote / review it but if someone in the other teams on support and they see this value it would not be immediately apparent to them that it is a value that is currently not in use and that it doesn't come from Stripe (because I would assume that it is if I didn't know!)

oswaldquek: In this case we have the charge external id so that should suffice. we don't need the errorId. this is my bad.

kbottla: Agreed and removed.

IreneLau-GDS: It's not anyone's bad! It's all good intentions. We just need to follow it through. 

Looking at the logs in the past 2 weeks while I was on support was extremely excruciating. I just don't want to make it any worse for anybody...

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
IreneLau-GDS: 🍬 👍

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
IreneLau-GDS: Now you have the import unused! 😱

kbottla: I have had auto remove on imports 😩Looks like I can't trust IDE anymore

IreneLau-GDS (approved):
I'll create a PR to remove the UUID error Id in `StripeServiceProvider` L130-L131.

PR #896, opened 2018-12-04
PP-4439 remove CancelServiceFunctions

georgeracu:
## WHAT
- Remove last traces of CancelServiceFunctions and co.
- Make public methods annotated with `@Transactional` so they will be picked up by GuicePersist
- Remove custom `merge()` calls as they are now handled by JPA


src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -119,74 +115,56 @@ private int expireChargesWithCancellationNotRequired(List<ChargeEntity> nonAuthS
...
+            logger.info("Charge status to update - charge_external_id={}, status={}, to_status={}",
+                    chargeEntity.getExternalId(), chargeEntity.getStatus(), status);
+            chargeEntity.setStatus(status);
+            chargeEventDao.persistChargeEventOf(chargeEntity);
+            return chargeEntity;
+        }).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeExternalId));
     }
 
-    private void logUnsuccessfulResponseReasons(ChargeEntity chargeEntity, GatewayResponse gatewayResponse) {
+    private void logUnsuccessfulResponseReasons(ChargeEntity chargeEntity, GatewayResponse<BaseCancelResponse> gatewayResponse) {
```
oswaldquek, line 167: not a big deal but does this really need to be in its own method? it's pretty much a one-liner. i think i'm just trying to draw a line when things should or should not be refactored into its own method and i think code would be neater if there were less methods.

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
```diff
@@ -193,10 +195,11 @@ private void checkIfChargeIsRefundableOrTerminate(ChargeEntity reloadedCharge, E
...
     private String getRefundReference(RefundEntity refundEntity, GatewayResponse<BaseRefundResponse> gatewayResponse) {
         if (gatewayResponse.isSuccessful()) {
-
-            return gatewayResponse.getBaseResponse().get().getReference().orElse(refundEntity.getExternalId());
+            return gatewayResponse.getBaseResponse().flatMap(BaseRefundResponse::getReference)
+                    .orElseGet(refundEntity::getExternalId);
         }
-        /**
+        /*
+         * todo: refactor this code to not return empty String
```
oswaldquek: i don't think it's too difficult to do this refactor in this PR? might as well ;)

oswaldquek (changes requested):
Not so much a request of changes, but some things to consider. I'm happy to approve if you don't think my suggestions are necessary.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
georgeracu: 👍

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
georgeracu: 👍

oswaldquek (dismissed):

oswaldquek (approved):

PR #898, opened 2018-12-06
PP-4352: Refactor PaymentProvider's capture interface

oswaldquek:
This commit simplifies the response from the capture method bigly. From
GatewayResponse<BaseCaptureResponse> -> CaptureResponse.

ChargeService no longer has knowledge of any specific payment provider.

@oswaldquek


alexbishop1:
I’m concerned about the logging changes in here. We log the result of the capture request in the `CardCaptureService.processGatewayCaptureResponse(…)` method (we do similar logging for authorisations etc.). On master, the log line `toString()`s the `GatewayResponse<BaseCaptureResponse>`, which will in turn `toString()` either the `BaseCaptureResponse` if present or the `GatewayError` if not (and stringification of the `GatewayError` will contain the stringification of the error `BaseCaptureResponse` if present).

The `BaseCaptureResponse` implementation `toString()`s (`WorldpayCaptureResponse` etc.) contain a summarisation of the response that’s intentionally quite close to what the payment gateway returns (the exact data and how it’s labelled varies between gateways and we use their terminology etc. to make cross-referencing with their documentation easier). In the past, we’ve found having this information really, really useful (for authorisations, at least — captures tend to be far less problematic).

With this PR, the `CardCaptureService.processGatewayCaptureResponse(…)` logging will instead `toString()` `CaptureResponse`, which doesn’t override the`toString()` method. Even if it did, it doesn’t have access to the `BaseCaptureResponse` (if present) and it looks like changes to the way `GatewayError` is built (if present) means that it doesn’t have as much gateway-specific information as it used to.

oswaldquek:
@alexbishop1 thanks for pointing this out. I think I can still make work. GatewayError's `toString()` is simply

https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/model/GatewayError.java#L88

and the `CaptureResponse` already contains a `GatewayError`.

WorldpayCaptureResponse's `toString()` only requires the txId, errorCode and errorMessage:

https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCancelResponse.java#L31-L42

information which `CaptureResponse` can have by passing the `WorldpayCaptureResponse` into its constructor.

I hope that will be a good compromise? `CaptureResponse` simplifies a lot of code and also I don't think we should be passing a complicated `GatewayResponse<BaseCaptureResponse>` back to `CardCaptureService.processGatewayCaptureResponse` for the sake of logging.

mrlumbu:
@alexbishop1 Just to throw my two cents in - I agree that this kind of logging is useful, but I just think it is a case of logging the right thing in the right place. I feel like logging the response from the gateway should be something that the relevant gateway specific thing should do, rather than passing up an overcomplicated thing to be logged higher up in the stack. I can imagine cases where you want to accumulate bits of information through the stack before logging it all in one go, but I'm not sure this case is one of them?

alexbishop1:
@oswaldquek Yes, `GatewayError.toString()` is fine but the way the `message` is built has changed. However, I think this would be easy to fix (and actually reduce the amount of code we have): in https://github.com/alphagov/pay-connector/blob/c7b6dae7ecc1167d75449416f1a8cfc51e86a277/src/main/java/uk/gov/pay/connector/gateway/CaptureHandler.java#L26-L27, just change string passed in to be `captureResponse.toString()`.

For the non-error case, replacing `CaptureResponse.fromTransactionId(String, ChargeState)` with something that takes the `BaseCaptureResponse` (or its stringification) would work, which I think is what you were suggesting.

alexbishop1:
@mrlumbu I think logging a line with an accumulation of information is useful in this case. Back in the distant past, the logging that’s now in `CardCaptureService.processGatewayCaptureResponse(…)` basically just said what status the payment was changing to as a result of the interaction with the gateway but it didn’t say why. There was sometimes also some logging closer to the actual interaction (depending on what actually happened) but it was inconsistent and often didn’t include useful things like the GOV.UK Pay payment ID. So finding out why a capture request failed (for example) could be quite frustrating.

While information is being passed up through the stack, the higher levels basically only need to `toString()` an object when logging; they don’t need to understand the intricacies of what happened lower in the stack.

oswaldquek:
@alexbishop1 I've updated the `CaptureResponse` class to accept a `BaseCaptureResponse`. I've chosen to reconstruct the toString as it's a code smell to assume an implementation of `BaseCaptureResponse` has overridden the `toString` method. This is especially pertinent as Stripe doesn't really utilise the concept of `BaseCaptureResponse`.

alexbishop1:
@oswaldquek I’m not sure that I agree it’s a code smell to assume implementations of `BaseCaptureResponse` have overridden `toString()`. I think it’s reasonable for us to declare that the contract of `BaseCaptureResponse` requires that there’s a sensible `toString()`.

The nice thing about using `toString()` is that `CaptureResponse`doesn’t have to know anything about each implementation of `BaseCaptureResponse` in order to output something useful. If `CaptureResponse` does the stringification itself, then it either has to be aware of the differences between `BaseCaptureResponse` implementations or overgeneralise (which is what we’re trying to get away from).

I note from looking at the `CaptureResponse` stringification code that it always calls the provider’s ID an `orderCode`, which is a Worldpay-specific term (for Smartpay it’s a `pspReference` — but this is actually per operation rather than per payment — and for ePDQ it’s a `PAYID`). We’re also missing some of the extra information that some payment providers can provide (which can be really useful for debugging). And it seems that the `CaptureResponse.fromBaseCaptureResponse(…)` method _does_ assume it can `toString()` `BaseCaptureResponse` when there’s an error.

oswaldquek:
@alexbishop1 ok, I can see it's a nice thing that `CaptureResponse` doesn’t have to know anything about each implementation of `BaseCaptureResponse`. my only gripe is that implementations of `BaseCaptureResponse` aren't FORCED to override `toString` by virtue of it being a method that every Object implements by default. One could very well make an anonymous class instance of `BaseCaptureResponse` without realising that overriding the `toString` is important for logging purposes. It all sounds quite messy. I'd prefer if implementations were forced to override another method that is not called `toString`.

alexbishop1:
@oswaldquek I could live with `BaseCaptureResponse` (and probably the other `BaseResponse`s) defining an abstract `stringify()` method (or something) so that implementers would be forced to implement it. That way if someone forgot to write provide a `stringify()` method, it would fail to compile. The `toString()` in `BaseCaptureResponse` could invoke this `stringify()` method to get maximum ergonomics.

oswaldquek:
@alexbishop1 sounds like a plan

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -330,8 +315,15 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
     }
 
     private ChargeEntity updateChargeStatus(ChargeEntity chargeEntity, ChargeStatus chargeStatus) {
-        chargeEntity.setStatus(chargeStatus);
-        chargeEventDao.persistChargeEventOf(chargeEntity);
+        if (chargeStatus == CAPTURED) {
+            chargeEntity.setStatus(CAPTURE_SUBMITTED);
+            chargeEventDao.persistChargeEventOf(chargeEntity);
+            chargeEntity.setStatus(CAPTURED);
+            chargeEventDao.persistChargeEventOf(chargeEntity, ZonedDateTime.now());
```
IreneLau-GDS, line 322: Do we expect the version of the charge entity to be incremented by 2 by the end of it?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -330,8 +315,15 @@ public ChargeEntity lockChargeForProcessing(String chargeId, OperationType opera
     }
 
     private ChargeEntity updateChargeStatus(ChargeEntity chargeEntity, ChargeStatus chargeStatus) {
-        chargeEntity.setStatus(chargeStatus);
-        chargeEventDao.persistChargeEventOf(chargeEntity);
+        if (chargeStatus == CAPTURED) {
```
IreneLau-GDS, line 318: Is the behaviour applicable to all payment service providers?

oswaldquek:
@alexbishop1 i've made the changes

src/main/java/uk/gov/pay/connector/gateway/epdq/model/response/EpdqCaptureResponse.java
```diff
@@ -43,6 +43,11 @@ public String getErrorMessage() {
 
     @Override
     public String toString() {
```
IreneLau-GDS, line 45: On a unrelated note - `EpdqBaseResponse` doesn't need to override the `toString()` method with an abstract method. I understand that it was there for over a year so it's not part of this commit. I just saw it while I traced through the code.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqCaptureHandler.java
```diff
@@ -2,15 +2,14 @@
...
 import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayOrder;
 import uk.gov.pay.connector.gateway.epdq.model.response.EpdqCaptureResponse;
 import uk.gov.pay.connector.gateway.model.GatewayError;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
-import uk.gov.pay.connector.gateway.model.response.BaseCaptureResponse;
-import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
-import uk.gov.pay.connector.gateway.util.GatewayResponseGenerator;
 
+import static java.lang.String.format;
```
IreneLau-GDS: This is now unused.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -59,29 +61,30 @@ public StripeCaptureHandler(StripeGatewayClient client,
...
-            logger.error("Capture failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
-                    request.getTransactionId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
-            GatewayError gatewayError = genericGatewayError(stripeErrorResponse.getError().getMessage());
-            return responseBuilder.withGatewayError(gatewayError).build();
+            String errorCode = stripeErrorResponse.getError().getCode();
+            String errorMessage = stripeErrorResponse.getError().getMessage();
+            logger.error("Capture failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. External Charge id: {}. Response code from Stripe: {}",
+                    transactionId, errorCode, errorMessage, request.getExternalId(), response.getStatus());
+
+            return fromBaseCaptureResponse(new StripeCaptureResponse(transactionId, errorCode, errorMessage), COMPLETE);
```
IreneLau-GDS: So this is still of COMPLETE status?

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeCaptureResponse.java
```diff
@@ -1,16 +1,28 @@
 package uk.gov.pay.connector.gateway.stripe.response;
 
+import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.connector.gateway.model.response.BaseCaptureResponse;
 
 import javax.ws.rs.core.Response;
```
IreneLau-GDS: You have unused imports in L6-L7

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCaptureHandler.java
```diff
@@ -4,14 +4,13 @@
...
+import uk.gov.pay.connector.gateway.CaptureResponse;
 import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayOrder;
 import uk.gov.pay.connector.gateway.model.GatewayError;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
-import uk.gov.pay.connector.gateway.model.response.BaseCaptureResponse;
-import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
-import uk.gov.pay.connector.gateway.util.GatewayResponseGenerator;
 
+import static java.lang.String.format;
```
IreneLau-GDS: Unused

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureProcess.java
```diff
@@ -11,13 +11,17 @@
 import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
+import uk.gov.pay.connector.gateway.CaptureResponse;
+import uk.gov.pay.connector.gateway.model.GatewayError;
```
IreneLau-GDS: L15-L16 unused

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -109,21 +110,15 @@ public void processGatewayCaptureResponse(String chargeId, String oldStatus, Gat
...
-        if (operationResponse.isSuccessful()) {
-            return CAPTURE_SUBMITTED;
-        } else {
-            return operationResponse.getGatewayError()
-                    .map(timeoutError -> CAPTURE_APPROVED_RETRY)
-                    .orElse(CAPTURE_ERROR);
-        }
+    private ChargeStatus determineNextStatus(CaptureResponse operationResponse) {
+        if (operationResponse.getError().isPresent()) return CAPTURE_APPROVED_RETRY;
+        if (operationResponse.state().equals(PENDING)) return CAPTURE_SUBMITTED; else return CAPTURED;
```
IreneLau-GDS, line 122: So capture error is no longer a thing? (Sorry about the many questions, I don't know the flow well so would like to understand.)

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqCaptureHandlerTest.java
```diff
@@ -67,9 +66,10 @@ public void shouldCapture() throws Exception {
         Either<GatewayError, EpdqCaptureResponse> unmarshalledResponse = right(XMLUnmarshaller.unmarshall(load(EPDQ_CAPTURE_SUCCESS_RESPONSE), EpdqCaptureResponse.class));
         when(client.unmarshallResponse(any(TestResponse.class), any(Class.class))).thenReturn(unmarshalledResponse);
         
-        GatewayResponse<BaseCaptureResponse> gatewayResponse = epdqCaptureHandler.capture(buildTestCaptureRequest());
+        CaptureResponse gatewayResponse = epdqCaptureHandler.capture(buildTestCaptureRequest());
         assertTrue(gatewayResponse.isSuccessful());
-        assertThat(gatewayResponse.getBaseResponse().get().getTransactionId(), is("3014644340"));
+        assertThat(gatewayResponse.getTransactionId().isPresent(), is(true));
+        assertThat(gatewayResponse.getTransactionId().get(), is("3014644340"));
     }
```
IreneLau-GDS, line 74: Should we assert the state as well, now that the `CaptureResponse` also contains the state?

src/test/java/uk/gov/pay/connector/gateway/smartpay/SmartpayCaptureHandlerTest.java
```diff
@@ -61,7 +60,7 @@ public void shouldCaptureAPaymentSuccessfully() {
                 .build();
 
         CaptureGatewayRequest request = CaptureGatewayRequest.valueOf(chargeEntity);
-        GatewayResponse gatewayResponse = smartpayCaptureHandler.capture(request);
+        CaptureResponse gatewayResponse = smartpayCaptureHandler.capture(request);
         assertTrue(gatewayResponse.isSuccessful());
```
IreneLau-GDS, line 66: Assert some extra things?

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeCaptureHandlerTest.java
```diff
@@ -74,9 +73,10 @@ public void shouldCapture() throws Exception {
 
         when(stripeGatewayClient.postRequest(any(URI.class), anyString(), any(Map.class), any(MediaType.class), anyString())).thenReturn(response);
 
-        GatewayResponse<BaseCaptureResponse> response = stripeCaptureHandler.capture(captureGatewayRequest);
+        CaptureResponse response = stripeCaptureHandler.capture(captureGatewayRequest);
         assertTrue(response.isSuccessful());
-        assertThat(response.getBaseResponse().get().getTransactionId(), is("ch_123456"));
+        assertThat(response.getTransactionId().isPresent(), is(true));
+        assertThat(response.getTransactionId().get(), is("ch_123456"));
     }
```
IreneLau-GDS, line 81: Same goes, shall we assert the state too?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCaptureHandlerTest.java
```diff
@@ -63,7 +62,7 @@ public void shouldCaptureAPaymentSuccessfully() throws Exception {
         Either<GatewayError, WorldpayCaptureResponse> unmarshalledResponse = right(XMLUnmarshaller.unmarshall(successCaptureResponse(), WorldpayCaptureResponse.class));
         when(client.unmarshallResponse(any(TestResponse.class), any(Class.class))).thenReturn(unmarshalledResponse);
 
-        GatewayResponse gatewayResponse = worldpayCaptureHandler.capture(getCaptureRequest());
+        CaptureResponse gatewayResponse = worldpayCaptureHandler.capture(getCaptureRequest());
         assertTrue(gatewayResponse.isSuccessful());
```
IreneLau-GDS, line 66: Broken record... extra asserts?

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
```diff
@@ -197,7 +197,7 @@ public void shouldCaptureSuccessfully() {
         String transactionId = response.getBaseResponse().get().getTransactionId();
         assertThat(transactionId, is(not(nullValue())));
         CaptureGatewayRequest captureRequest = buildCaptureRequest(chargeEntity, transactionId);
-        GatewayResponse<BaseCaptureResponse> captureResponse = paymentProvider.capture(captureRequest);
+        CaptureResponse captureResponse = paymentProvider.capture(captureRequest);
         assertThat(captureResponse.isSuccessful(), is(true));
```
IreneLau-GDS, line 202: Extra asserts?

src/test/java/uk/gov/pay/connector/it/contract/SmartpayPaymentProviderTest.java
```diff
@@ -194,7 +194,7 @@ public void shouldSuccessfullySendACaptureRequest() {
 
         chargeEntity.setGatewayTransactionId(transactionId);
 
-        GatewayResponse<BaseCaptureResponse> captureGatewayResponse = paymentProvider.capture(CaptureGatewayRequest.valueOf(chargeEntity));
+        CaptureResponse captureGatewayResponse = paymentProvider.capture(CaptureGatewayRequest.valueOf(chargeEntity));
         assertTrue(captureGatewayResponse.isSuccessful());
```
IreneLau-GDS, line 200: Extra asserts?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
```diff
@@ -106,20 +110,13 @@ public void beforeTest() {
...
-    private void worldpayWillRespondWithSuccess(String transactionId, String worldpayErrorCode) {
-        WorldpayCaptureResponse worldpayResponse = mock(WorldpayCaptureResponse.class);
-        when(worldpayResponse.getTransactionId()).thenReturn(transactionId);
-        when(worldpayResponse.getErrorCode()).thenReturn(worldpayErrorCode);
-        GatewayResponseBuilder<WorldpayCaptureResponse> gatewayResponseBuilder = responseBuilder();
-        GatewayResponse captureResponse = gatewayResponseBuilder.withResponse(worldpayResponse).build();
-        when(mockedPaymentProvider.capture(any())).thenReturn(captureResponse);
+    private void worldpayWillRespondWithSuccess() {
+        when(mockedPaymentProvider.capture(any())).thenReturn(
+                fromBaseCaptureResponse(BaseCaptureResponse.fromTransactionId(randomUUID().toString(), WORLDPAY), PENDING));
```
IreneLau-GDS, line 115: Should state be PENDING? Or it doesn't matter?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: Yes

src/main/java/uk/gov/pay/connector/gateway/epdq/model/response/EpdqCaptureResponse.java
oswaldquek: Unfortunately it does :( It won't compile if it doesn't.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
oswaldquek: very good spot. will fix

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
oswaldquek: I think it's no longer a thing. If you look at the code on the left, if the operationResponse has a gatewayError it is mapped to a CAPTURE_APPROVED_RETRY:
```
return operationResponse.getGatewayError()	
                    .map(timeoutError -> CAPTURE_APPROVED_RETRY)
```
There is an `orElse` bit:
```
.orElse(CAPTURE_ERROR)
```
but if there isn't a capture error, it would be successful and that is covered by
```
if (operationResponse.isSuccessful()) {	        if (operationResponse.getError().isPresent()) return CAPTURE_APPROVED_RETRY;
            return CAPTURE_SUBMITTED;
```

src/main/java/uk/gov/pay/connector/gateway/epdq/model/response/EpdqCaptureResponse.java
IreneLau-GDS: It compiles on mine. I meant removing the one in the base class (`EpdqBaseReaponse`) where this class extends.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
oswaldquek: worldpay's response should be PENDING. it's only complete when we receive a notification from worldpay.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
IreneLau-GDS: The fact that nothing failed means we will need a test or two for this scenario...

oswaldquek: There is a test in StripeCaptureHandler.shouldCapture. `assertThat(response.state(), is(CaptureResponse.ChargeState.COMPLETE));`

src/main/java/uk/gov/pay/connector/gateway/epdq/model/response/EpdqCaptureResponse.java
oswaldquek: We shouldn't remove it from the `EpdqBaseReaponse` as there are other classes that extend it. We could remove it in a future PR.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: Where do we test for charge entity versions currently?

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqCaptureHandlerTest.java
oswaldquek: this is done now

src/test/java/uk/gov/pay/connector/gateway/smartpay/SmartpayCaptureHandlerTest.java
oswaldquek: the state is now asserted

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeCaptureHandlerTest.java
oswaldquek: done

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCaptureHandlerTest.java
oswaldquek: the state is now asserted

src/test/java/uk/gov/pay/connector/it/contract/EpdqPaymentProviderTest.java
oswaldquek: the state is now asserted

src/test/java/uk/gov/pay/connector/it/contract/SmartpayPaymentProviderTest.java
oswaldquek: the state is now asserted

src/main/java/uk/gov/pay/connector/gateway/epdq/model/response/EpdqCaptureResponse.java
IreneLau-GDS: Right... so basically we are doing it in the base class to enforce the subclasses to implement it. It seems a bit... cringey... (Sorry...)

Maybe in the future PR, we can do something like:

```
    public abstract class BaseClass {
        
        protected abstract String formatToStringOutput();
        
        public final String toString() {
            return formatToStringOutput();
        }
    }

    public class SubClass extends BaseClass {

        @Override
        public String formatToStringOutput() {
            return "I would like to format my toString() like this!";
        }
    }
```

oswaldquek: yes it is cringey. if you read the PR message history I said: 

```"my only gripe is that implementations of BaseCaptureResponse aren't FORCED to override toString by virtue of it being a method that every Object implements by default".```

To which Alex responded: ```"I could live with BaseCaptureResponse (and probably the other BaseResponses) defining an abstract stringify() method (or something) so that implementers would be forced to implement it. That way if someone forgot to write provide a stringify() method, it would fail to compile"```

File: src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java (hunk 1 above)
Comment (IreneLau-GDS) [reply src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java:322]: I wasn't asking about the test. I am trying to understand what behaviour we expect it to be. 

What looks to me is that we are trying to insert two events but the only way to do it is by setting the status of the `chargeEntity` object twice and pass it into the `persistChargeEventOf` method in the `ChargeEventDao` class. 

So is it supposed to be one event that does two things? Or is it really two actual events? The way we have done it seems a bit strange logically... Probably just me being simple minded.

Comment (oswaldquek) [reply src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java:322]: yes we want two events - a CAPTURE_SUBMITTED and CAPTURED event. the way to do this is by setting the status on the charge to CAPTURE_SUBMITTED first, then calling `chargeEventDao.persistChargeEventOf`. then set the status to CAPTURED and call `chargeEventDao.persistChargeEventOf` again. i did think about changing the `chargeEventDao.persistChargeEventOf` interface to accept a status but decided against it.

Comment (IreneLau-GDS) [reply src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java:322]: So is it actually two events or one event with two side effects? The reason why I ask is that an event generally indicates one thing that happens. However, it looks like capture submitted and captured happen all in one go. I'm trying to understand what's going on.

File: src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java (hunk 5 above)
Comment (IreneLau-GDS) [reply src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java]: Should it be in COMPLETE status or not? (Coz your reply seems to imply that it shouldn't be.)

File: src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java (hunk 1 above)
Comment (oswaldquek) [reply src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java:322]: I've tagged you in the relevant slack conversation. Short answer is yes, two events happen in one go.

File: src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java (hunk 5 above)
Comment (oswaldquek) [reply src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java]: oh sorry, i got confused. i thought it was a success scenario. it should just wrap a GatewayError inside the returned CaptureResponse.

File: src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java (hunk 1 above)
Comment (IreneLau-GDS) [reply src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java:322]: Have discussed with Oz. Thx for explaining. Just to make a note... the capture flow for stripe is synchronous for Stripe (and Sandbox), then that single "action" kind of implies two events because the request is the submit, the response is the capture..., although the mechanism is fundamentally different from the remaining payment gateways.

File: src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java [hunk 17]
```diff
@@ -59,29 +61,30 @@ public StripeCaptureHandler(StripeGatewayClient client,
...
-            logger.error("Capture failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
-                    request.getTransactionId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
-            GatewayError gatewayError = genericGatewayError(stripeErrorResponse.getError().getMessage());
-            return responseBuilder.withGatewayError(gatewayError).build();
+            String errorCode = stripeErrorResponse.getError().getCode();
+            String errorMessage = stripeErrorResponse.getError().getMessage();
+            logger.error("Capture failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. External Charge id: {}. Response code from Stripe: {}",
+                    transactionId, errorCode, errorMessage, request.getExternalId(), response.getStatus());
+
+            return fromBaseCaptureResponse(new StripeCaptureResponse(transactionId, errorCode, errorMessage), null);
```
IreneLau-GDS, line 76: Instead of passing in a `null` value for ChargeStatus, we can potentially do something clearer in `CaptureResponse` to default it to `null`. But that is a conversation another time.

IreneLau-GDS (approved):
I have comments around how things are currently done but that's not something should be addressed in this PR or it will never ends.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
oswaldquek: I did deliberate about this. I think it's better to force the developer to consider whether a status needs to be passed in. I'm against passing in an Optional.

PR #899, opened 2018-12-06
WIP PP 4439 refactor card authorise service response

georgeracu:
## WHAT
- At the moment we propagate `GatewayResponse<BaseAuthoriseResponse>` all
    the way up to the resource. This is not a desired behaviour as this object
    has a pretty convoluted structure, also issues compilation warnings and most
    importantly, a service or a resource don't need to know about the response
    from the payment provider (gateway). Also we were passing quite a few optional
    method arguments around, which brings a bit of code smell. Here I extracted
    into an object what a service needs to know from the payment gateway in order
    to update the charge. This object will be converted into an even lighter one
    that gets sent to the resource where the resource will deal with how to respond.
    - In this craziness I also moved some specific methods from CardAuthoriseBaseService
    into Card3dsResponseAuthService as it were specific to this class
    - I extracted logic onto how to generate a `PaymentProviderAuthorisationResponse` into
    this class as it will be into one place and anyone who needs one of this objects can
    easily pass a gateway response into the factory method and voila, you a new object
    - Simplified `AuthorisationResponse` to make use of the Java 8 Optional.ifPresent()
    - Changed Apple related services to use the new object
    - Inlined strings for logs and metrics
    - There is also more work to be done to get rid of the generic object that gets
    returned from the payment provider

### DIFF
--- a/src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/applepay/AppleAuthoriseService.java
@@ -16,9 +16,9 @@ import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 import uk.gov.pay.connector.paymentprocessor.service.CardAuthoriseBaseService;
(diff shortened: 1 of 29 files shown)


alexbishop1:
I think a lot of what I said about logging in https://github.com/alphagov/pay-connector/pull/898 also applies here

georgeracu:
I closed this PR as it can be submitted into smaller chunks.

PR #900, opened 2018-12-06
BAU Don't run e2e tests on PR by default

mrlumbu:
As a result of extensive polling in the Pay team, we decided to
try setting the default for running e2e tests on a PR to false.
We believe this will lead to increase in speed, possibly at cost of
increase in broken builds. If the experiment proves
successful we will roll out to other services.


georgeracu (approved):
LGTM

PR #871, opened 2018-11-26
PP-4373 Simplify auth 3ds interface

mrlumbu:
This PR does two things:
- Pushes some functionality in `Auth3DSResponseService` into `CarAuthoriseBaseService` - basically logging and metrics, which we can also use from any other authorise service
- Simplifies the interface of the payment provider method `authorise3dsResponse` to respond with a simpler object, that only contains the necessary information for doing the things that the service and resource need to do. This is still not perfect, but is massively simpler and cleaner, and can be refined in further work.

PR #902, opened 2018-12-07
PP-4439 Move logging and metrics methods inside Card3dsResponse...

georgeracu:
## WHAT
- Methods from the base service are being used only by the Card3dsResponseAuthService
and there was no reason to have them in the base service
- In-lined log and metrics markers
- This is part of a bigger change but is easier to submit it in smaller chunks
than a 600 lines change review

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
@@ -1,5 +1,9 @@
 package uk.gov.pay.connector.paymentprocessor.service;
 
+import com.codahale.metrics.MetricRegistry;
+import io.dropwizard.setup.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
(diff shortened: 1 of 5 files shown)


kbottla:
Logging was moved recently (https://github.com/alphagov/pay-connector/pull/871/files#diff-787957e0de43712f1eab198ac5fbb790R87) to BaseAuthoriseService as part of 3DS refactoring. I think the plan was to eventually use it in CardAuthoriseService too.

Also metric format has been corrected this morning (https://github.com/alphagov/pay-connector/commit/7b082e99a3013d2da3851b227b01b9da69002097)

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
```diff
@@ -58,7 +66,33 @@ private void processGateway3DSecureResponse(
...
+                updatedCharge.getGatewayAccount().getAnalyticsId(),
+                updatedCharge.getGatewayAccount().getId(),
+                oldChargeStatus,
+                updatedCharge.getStatus()
+        );
+    }
+
+    private void emitAuthorisationMetric(ChargeEntity charge) {
+        metricRegistry.counter(String.format("gateway-operations.%s.%s.%s.%s.result.%s",
+                "authorise-3ds",
```
alexbishop1, line 91: Please don’t break my fix! https://github.com/alphagov/pay-connector/pull/901

PR #904, opened 2018-12-07
PP-4439 Stop BaseAuthoriseResponse from bubbling up

georgeracu:
## WHAT
- Stop BaseAuthoriseResponse from bubbling all the way up to the surface of a
resource. Added a custom object that is returned by the service and the
response is built at the resource level, based on error or success case

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/api/AuthorisationResponse.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/api/AuthorisationResponse.java
@@ -1,0 +1,25 @@
+package uk.gov.pay.connector.paymentprocessor.api;
+
+import uk.gov.pay.connector.gateway.model.GatewayError;
+import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
+import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+
+import java.util.Optional;
+
+public class AuthorisationResponse {
+    private BaseAuthoriseResponse.AuthoriseStatus authoriseStatus;
+    private GatewayError gatewayError;
+
+    public AuthorisationResponse(GatewayResponse<BaseAuthoriseResponse> gatewayResponse) {
+        gatewayResponse.getBaseResponse().ifPresent(baseAuthoriseResponse -> authoriseStatus = baseAuthoriseResponse.authoriseStatus());
+        gatewayResponse.getGatewayError().ifPresent(error -> gatewayError = error);
+    }
+
+    public Optional<GatewayError> getGatewayError() {
+        return Optional.ofNullable(gatewayError);
+    }
+
+    public Optional<BaseAuthoriseResponse.AuthoriseStatus> getAuthoriseStatus() {
+        return Optional.ofNullable(authoriseStatus);
+    }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/api/AuthorisationResponse.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.pay.connector.paymentprocessor.api;
+
+import uk.gov.pay.connector.gateway.model.GatewayError;
+import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
+import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+
+import java.util.Optional;
+
+public class AuthorisationResponse {
+    private BaseAuthoriseResponse.AuthoriseStatus authoriseStatus;
```
IreneLau-GDS: If the getters are always going to return an Optional, you might as well have the member variables as Optional too?

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -58,13 +57,21 @@ public Response authoriseCharge(@PathParam("chargeId") String chargeId, AuthCard
         if (!isWellFormatted(authCardDetails)) {
             return badRequestResponse("Values do not match expected format/length.");
         }
-        GatewayResponse<BaseAuthoriseResponse> response = cardAuthoriseService.doAuthorise(chargeId, authCardDetails);
+        AuthorisationResponse response = cardAuthoriseService.doAuthorise(chargeId, authCardDetails);
 
-        if (isAuthorisationSubmitted(response)) {
-            return badRequestResponse("This transaction was deferred.");
-        }
+        return response.getGatewayError().map(this::handleError)
```
IreneLau-GDS, line 62: I actually find the old mini methods breakdown a bit easier to read...

IreneLau-GDS:
Okay so I look forward reviewing the PR to get rid of the other `Base*Reponse` classes!

src/main/java/uk/gov/pay/connector/paymentprocessor/api/AuthorisationResponse.java
georgeracu: I don't see any value in having the member variables optional. Because when you call get() you have to make sure that you instantiate it so you don't return null, instead of Optional<T>.

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
georgeracu: 👍

src/main/java/uk/gov/pay/connector/paymentprocessor/api/AuthorisationResponse.java
IreneLau-GDS: I do see some values, they are:
* Your member variable can be `final` because the constructor is the only place that sets the value
* Your getter just returns the member variable

i.e.
```
public class AuthorisationResponse {
    private final Optional<BaseAuthoriseResponse.AuthoriseStatus> authoriseStatus;
    private final Optional<GatewayError> gatewayError;

    public AuthorisationResponse(GatewayResponse<BaseAuthoriseResponse> gatewayResponse) {
        authoriseStatus = gatewayResponse.getBaseResponse().map(BaseAuthoriseResponse::authoriseStatus);
        gatewayError = gatewayResponse.getGatewayError();
    }

    public Optional<GatewayError> getGatewayError() {
        return gatewayError;
    }

    public Optional<BaseAuthoriseResponse.AuthoriseStatus> getAuthoriseStatus() {
        return authoriseStatus;
    }
}
```

IreneLau-GDS (approved):
Sorry for being picky...

PR #905, opened 2018-12-11
PP-4508 Do not show corporate card surcharge for refunds

DanailMinchev:
## WHAT

This PR fixes the issue when we display extra surcharge on `/v2/api/accounts/{accountId}/charges` endpoint.

## HOW 

 - modified `TransactionSearchStrategy` to exclude corporate surcharge properties
in refunds, when using: `/v2/api/accounts/{accountId}/charges` endpoint
- Refactored `ChargesApiResourceITest` to have common method for creating a charge
- Code Formatting

with @cobainc0

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/TransactionSearchStrategy.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/TransactionSearchStrategy.java
@@ -10,10 +10,10 @@ import uk.gov.pay.connector.charge.model.TransactionType;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.model.domain.PersistedCard;
 import uk.gov.pay.connector.charge.model.domain.Transaction;
+import uk.gov.pay.connector.charge.util.CorporateCardSurchargeCalculator;
 import uk.gov.pay.connector.common.model.api.ExternalChargeState;
 import uk.gov.pay.connector.common.model.api.ExternalRefundStatus;
 import uk.gov.pay.connector.common.model.api.ExternalTransactionState;
-import uk.gov.pay.connector.charge.util.CorporateCardSurchargeCalculator;
 import uk.gov.pay.connector.common.service.search.AbstractSearchStrategy;
 import uk.gov.pay.connector.common.service.search.SearchStrategy;
 import uk.gov.pay.connector.refund.model.domain.RefundStatus;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/charge/service/TransactionSearchStrategy.java
```diff
@@ -95,13 +95,15 @@ public ChargeResponse buildResponse(UriInfo uriInfo, Transaction transaction) {
...
         }
 
-        transaction.getCorporateCardSurcharge().ifPresent(surcharge -> {
-            if (surcharge > 0) {
-                transactionResponseBuilder
-                        .withCorporateCardSurcharge(surcharge)
-                        .withTotalAmount(CorporateCardSurchargeCalculator.getTotalAmountFor(transaction));
-            }
-        });
+        if (!TransactionType.REFUND.getValue().equals(transaction.getTransactionType())) {
```
alexbishop1: Maybe this should be a non-reversed check for `CHARGE` (i.e. say we only want to do this if it’s a charge rather than not do it if it’s a refund) so we don’t get any surprises with hypothetical future new transaction types

alexbishop1 (dismissed):
Nice work

src/main/java/uk/gov/pay/connector/charge/service/TransactionSearchStrategy.java
DanailMinchev: Good point, we changed it to compare with `charge` 👍

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/charge/service/TransactionSearchStrategy.java
```diff
@@ -84,24 +84,26 @@ public ChargeResponse buildResponse(UriInfo uriInfo, Transaction transaction) {
...
         }
 
-        transaction.getCorporateCardSurcharge().ifPresent(surcharge -> {
-            if (surcharge > 0) {
-                transactionResponseBuilder
-                        .withCorporateCardSurcharge(surcharge)
-                        .withTotalAmount(CorporateCardSurchargeCalculator.getTotalAmountFor(transaction));
-            }
-        });
+        if ("charge".equals(transaction.getTransactionType())) {
```
stephencdaly, line 98: Should the transaction types "charge" and "refund" exist as an enum?

DanailMinchev: yes, we can do that as a separate PR, we discussed that before with @alexbishop1 not to change that, but as discussed again we can try to see if it will break anything

PR #907, opened 2018-12-12
PP-4509: Remove errorId

oswaldquek:
errorId was originally put there to enable a support person to dig deeper into
a cause of failure without returning the cause back to the client. However
using the charge external id will do just fine.

@oswaldquek

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
@@ -9,13 +9,13 @@ public class RefundGatewayRequest implements GatewayRequest {
     private final GatewayAccountEntity gatewayAccountEntity;
     private final String amount;
     private final String transactionId;
-    private final String reference;
+    private final String refundExternalId;
 
-    private RefundGatewayRequest(String transactionId, GatewayAccountEntity gatewayAccount, String amount, String reference) {
+    private RefundGatewayRequest(String transactionId, GatewayAccountEntity gatewayAccount, String amount, String refundExternalId) {
         this.transactionId = transactionId;
         this.gatewayAccountEntity = gatewayAccount;
         this.amount = amount;
-        this.reference = reference;
+        this.refundExternalId = refundExternalId;
     }
 
     public String getAmount() {
@@ -36,6 +36,10 @@ public class RefundGatewayRequest implements GatewayRequest {
         return GatewayOperation.REFUND;
     }
 
+    public String getRefundExternalId() {
+        return refundExternalId;
+    }
+    
     /**
      * <p>
      * For Worldpay ->
@@ -61,8 +65,16 @@ public class RefundGatewayRequest implements GatewayRequest {
                 String.valueOf(refundEntity.getAmount()),
                 refundEntity.getExternalId());
     }
-
-    public String getReference() {
-        return reference;
+    
+    @Override
+    public String toString() {
+        return new StringBuilder()
+                .append("RefundGatewayRequest[\n")
+                .append("transactionId: ")
+                .append(transactionId)
+                .append("\nrefundExternalId: ")
+                .append(refundExternalId)
+                .append("]")
+                .toString();
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCancelHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCancelHandler.java
@@ -88,9 +88,8 @@ public class StripeCancelHandler {
             
             Response response = e.getResponse();
             StripeErrorResponse stripeErrorResponse = response.readEntity(StripeErrorResponse.class);
-            String errorId = UUID.randomUUID().toString();
-            logger.error("Cancel failed for gateway transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
-                    request.getTransactionId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
+            logger.error("Cancel failed for gateway transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. Charge External Id: {}. Response code from Stripe: {}",
+                    request.getTransactionId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), request.getExternalChargeId(), response.getStatus());
             GatewayError gatewayError = genericGatewayError(stripeErrorResponse.getError().getMessage());
             return responseBuilder.withGatewayError(gatewayError).build();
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
```diff
@@ -9,13 +9,13 @@
...
-    private final String reference;
+    private final String externalChargeId;
 
-    private RefundGatewayRequest(String transactionId, GatewayAccountEntity gatewayAccount, String amount, String reference) {
+    private RefundGatewayRequest(String transactionId, GatewayAccountEntity gatewayAccount, String amount, String externalChargeId) {
         this.transactionId = transactionId;
         this.gatewayAccountEntity = gatewayAccount;
         this.amount = amount;
-        this.reference = reference;
+        this.externalChargeId = externalChargeId;
```
alexbishop1: Really not sure about this. Refunds have references (identifiers also shared with the payment gateways) but they are not the same as the external ID of the associated charge.

oswaldquek: I'll revert this back to "reference". 

When a refund entity first gets, created, the reference is not populated: https://github.com/alphagov/pay-connector/blob/8452d22a58548caca05a237221deb4e99e5f3701/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundEntity.java#L87.

However when we make a RefundGatewayRequest, it's created with
```
    public static RefundGatewayRequest valueOf(RefundEntity refundEntity) {
        return new RefundGatewayRequest(
                refundEntity.getChargeEntity().getGatewayTransactionId(),
                refundEntity.getChargeEntity().getGatewayAccount(),
                String.valueOf(refundEntity.getAmount()),
                refundEntity.getExternalId());
    }
```
At this point the refundEntity's external id is null?? Something seems off here.

alexbishop1: I think it works like this:

1. When a `RefundEntity` is created, it gets assigned a random new `externalId` (this is not the same as the associated charge’s `externalId`)
2. When we send the refund to the gateway (using the `RefundGatewayRequest`), we sometimes (depending on the gateway: we definitely do this for Worldpay) include the `RefundEntity`’s `externalId`, which we refer to in some of our variables and comments as our `reference`
3. When we receive the response from the gateway in `ChargeRefundService`, we populate the `reference` on the `RefundEntity` with a reference provided by the gateway — if the gateway does not provide one (Worldpay doesn’t), we use the `RefundEntity`’s `externalId` 
4. We then rely on the gateway providing this latter `reference` in subsequent notifications so we can match them up to the refund

So I think the appropriate thing to call this variable is `refundExternalId` and we should update the comments in this class so they don’t talk about sending “our reference”.

oswaldquek: @alexbishop1 thanks for the clarification. i've updated RefundGatewayRequest's reference field to refundExternalId. also i realised i was not looking properly when i said the refundEntity's external id is null!

src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
```diff
@@ -61,8 +65,17 @@ public static RefundGatewayRequest valueOf(RefundEntity refundEntity) {
...
                 refundEntity.getExternalId());
     }
-
-    public String getReference() {
-        return reference;
+    
+    @Override
+    public String toString() {
+        return new StringBuilder()
+                .append("EvaluatedGatewayRefundRequest[\n")
```
alexbishop1: Name doesn’t look right

src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
```diff
@@ -61,8 +65,17 @@ public static RefundGatewayRequest valueOf(RefundEntity refundEntity) {
...
+                .append("transactionId: ")
+                .append(transactionId)
+                .append("\nrefundExternalId: ")
+                .append(refundExternalId)
+                .append("\nchargeExternalId:")
```
alexbishop1: This label seems lost

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCancelHandler.java
```diff
@@ -100,10 +99,9 @@ public CancelStatus cancelStatus() {
             
         } catch (DownstreamException e) {
             
-            String errorId = UUID.randomUUID().toString();
-            logger.error("Cancel failed for transaction id {}. Reason: {}. Status code from Stripe: {}. ErrorId: {}",
-                    request.getTransactionId(), e.getMessage(), e.getStatusCode(), errorId);
-            GatewayError gatewayError = unexpectedStatusCodeFromGateway("An internal server error occurred. ErrorId: " + errorId);
+            logger.error("Cancel failed for transaction id {}. Reason: {}. Status code from Stripe: {}. Charge External Id: {}",
+                    request.getTransactionId(), e.getMessage(), e.getStatusCode(), request.getExternalChargeId());
+            GatewayError gatewayError = unexpectedStatusCodeFromGateway("An internal server error occurred while canceling charge_external_id: " + request.getExternalChargeId());
```
alexbishop1: Don’t really need the underscores; it’s not a variable name

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeCancelHandlerTest.java
```diff
@@ -86,7 +86,7 @@ public void shouldHandle5xxFromStripeGateway() throws Exception {
         final GatewayResponse<BaseCancelResponse> gatewayResponse = stripeCancelHandler.cancel(request);
         assertThat(gatewayResponse.isFailed(), is(true));
         assertThat(gatewayResponse.getGatewayError().isPresent(), Is.is(true));
-        assertThat(gatewayResponse.getGatewayError().get().getMessage(), containsString("An internal server error occurred. ErrorId:"));
+        assertThat(gatewayResponse.getGatewayError().get().getMessage(), containsString("An internal server error occurred while canceling charge_external_id: " + request.getExternalChargeId()));
```
alexbishop1: Two “l”s in “cancelling” in British English

alexbishop1 (changes requested):
Nice change and some good logging improvements

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeCancelHandlerTest.java
oswaldquek: but the less code the better! 😛

alexbishop1 (approved):
A few `charge_external_id`s hanging around in logging and exception messages but I’m not going to be petty about that

PR #908, opened 2018-12-13
BAU: Rearrange Dockerfile to reduce the number of rebuilt layers

rhowe-gds:
The thing which changes most often is the .jar, so add that to the docker image
last. This means that in the common case, we only rebuild one layer rather than
several.


simad (approved):
💯

PR #911, opened 2018-12-13
BAU: Add extra logging to Stripe HTTP Client

IreneLau-GDS:
# WHAT
We were experiencing problems with test where HTTP calls out to Stripe don't seem to reach them. We need to up the logging to see what is happening.

I also squeezed two lots of changes:
* Removed the error Id generation and logging that goes nowhere because there is a separate story to implement it properly. 
* Moved where we call `throwIfErrorResponse` because of potential NPE.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
@@ -15,8 +15,6 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
-import static javax.ws.rs.core.Response.Status.Family.CLIENT_ERROR;
-import static javax.ws.rs.core.Response.Status.Family.SERVER_ERROR;
 
 /**
  * This class, while named StripeGatewayClient, is meant to be payment provider agnostic. It will be used by all
@@ -43,7 +41,7 @@ public class StripeGatewayClient {
 
         Response response;
         try {
-
+            logger.info("POSTing request for gateway url={}", url);
             Invocation.Builder clientBuilder = client.target(url.toString()).request();
             headers.keySet().forEach(headerKey -> clientBuilder.header(headerKey, headers.get(headerKey)));
 
@@ -63,18 +61,24 @@ public class StripeGatewayClient {
     }
 
     private void throwIfErrorResponse(Response response, String metricsPrefix) throws GatewayClientException, DownstreamException {
-        if (CLIENT_ERROR == response.getStatusInfo().getFamily()) {
-            metricRegistry.counter(metricsPrefix + ".failures").inc();
-            throw new GatewayClientException(
-                    "Unexpected HTTP status code " + response.getStatus() + " from gateway",
-                    response);
-        }
-        if (SERVER_ERROR == response.getStatusInfo().getFamily()) {
-            metricRegistry.counter(metricsPrefix + ".failures").inc();
-            String responseMessage = response.readEntity(String.class);
-            logger.error("Unexpected HTTP status code {} from gateway, error=[{}]",
-                    response.getStatus(), responseMessage);
-            throw new DownstreamException(response.getStatus(), responseMessage);
+
+        switch (response.getStatusInfo().getFamily()) {
+
+            case SUCCESSFUL:
+                break;
+
+            case CLIENT_ERROR:
+                metricRegistry.counter(metricsPrefix + ".failures").inc();
+                throw new GatewayClientException("Unexpected HTTP status code " + response.getStatus() + " from gateway", response);
+
+            case SERVER_ERROR:
+                metricRegistry.counter(metricsPrefix + ".failures").inc();
+                String responseMessage = response.readEntity(String.class);
+                logger.error("Unexpected HTTP status code {} from gateway, error=[{}]", response.getStatus(), responseMessage);
+                throw new DownstreamException(response.getStatus(), responseMessage);
+
+             default:
+                 logger.error("Other HTTP status from gateway=[stripe], response=[]", response);
         }
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
@@ -127,9 +127,8 @@ public class StripePaymentProvider implements PaymentProvider {
             }
 
             StripeErrorResponse stripeErrorResponse = response.readEntity(StripeErrorResponse.class);
-            String errorId = UUID.randomUUID().toString();
-            logger.error("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
-                    request.getChargeExternalId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
+            logger.error("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -76,5 +76,7 @@ private void throwIfErrorResponse(Response response, String metricsPrefix) throw
                     response.getStatus(), responseMessage);
             throw new DownstreamException(response.getStatus(), responseMessage);
         }
+
+        logger.error("Other unexpected HTTP status code {} from gateway=[stripe]", response.getStatus());
```
oswaldquek: might it be useful to log the response body as well?

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
IreneLau-GDS: Do they even reply with a body? I can only see in their doc that they send back specific error code. 

If not, there is no point to try to cast the `Response` object to `HttpResponse` just to get a body that has nothing.

oswaldquek: I'm not sure they reply with a body but it can't hurt to log `response.getEntity(String.class)` which will get the body. if there is a specific error code from them it'll be in the body too.

IreneLau-GDS: If you say so...

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -43,11 +43,12 @@ public Response postRequest(URI url,
 
         Response response;
         try {
-
+            logger.info("POSTing request for gateway url={}", url);
             Invocation.Builder clientBuilder = client.target(url.toString()).request();
             headers.keySet().forEach(headerKey -> clientBuilder.header(headerKey, headers.get(headerKey)));
 
             response = clientBuilder.post(Entity.entity(payload, mediaType));
+            throwIfErrorResponse(response, metricsPrefix);
```
kbottla: Don't think this move works. `throwIfErrorResponse` throws few checked exceptions based on response status (4xx,5xx) which will be caught here by `catch (Exception e)` which is not desired.
We want to throw exceptions so payment provider can deal with them.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -76,5 +76,8 @@ private void throwIfErrorResponse(Response response, String metricsPrefix) throw
                     response.getStatus(), responseMessage);
             throw new DownstreamException(response.getStatus(), responseMessage);
         }
+
+        logger.error("Other unexpected HTTP status code {} from gateway=[stripe], responseBody=[]",
+                response.getStatus(), response.readEntity(String.class));
```
kbottla: This will be logged even for 2xx & 3xx errors. Is this what it is intended for ?

kbottla (changes requested):

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
IreneLau-GDS: Okay then I'll add a null check because where it was before can throw an NPE.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
IreneLau-GDS: If you look at the existing `GatewayClient`, I'm afraid that's exactly what it is doing.

https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java#L81-L87 

That raise a question though, why are we throwing exceptions for certain status code for stripe but not for the other gateway?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
oswaldquek: you can always `catch (GatewayClientException |  DownstreamException e) { throw e }` if you want to avoid catching the NPE

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
kbottla: Yes, `GatewayClient` is logging non 200 responses.

Current log statement will log `everything` including 200 which is not unexpected

kbottla: > That raise a question though, why are we throwing exceptions for certain status code for stripe but not for the other gateway?

Reason is error messages can be in the format specific to Payment provider (json in case of Stripe). To keep StripeGateway free from handling gateway specific error codes, exceptions are propagated and relevant PaymentProvider can process error messages

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
IreneLau-GDS: Then rethrow? That would be nasty. I'd rather do a null check.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -58,7 +58,10 @@ public Response postRequest(URI url,
             metricRegistry.histogram(metricsPrefix + ".response_time").update(responseTimeStopwatch.elapsed(TimeUnit.MILLISECONDS));
         }
 
-        throwIfErrorResponse(response, metricsPrefix);
+        if (response != null) {
```
oswaldquek: i don't like this null check... why would it be null? it's successful at this point. this method MUST return a valid Response object if there was no error. if you're that concerned about the null then moving `throwIfErrorResponse` into the `try` block and doing `catch (GatewayClientException| DownstreamException e) { throw e }` is the lesser evil. in my opinion.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
IreneLau-GDS: @oswaldquek that didn't work BTW. So I've checked it to just print the whole response object. The default toString did print the output, not the reference to the object.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
IreneLau-GDS: I think you are right. it will not happen. Let me remove it.

oswaldquek (dismissed):

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
```diff
@@ -76,5 +76,7 @@ private void throwIfErrorResponse(Response response, String metricsPrefix) throw
                     response.getStatus(), responseMessage);
             throw new DownstreamException(response.getStatus(), responseMessage);
         }
+
+        logger.error("Other unexpected HTTP status gateway=[stripe], response=[]", response);
```
kbottla: We will be logging this for all statuses (including 2xx). All won't be `unexpected`.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClient.java
IreneLau-GDS: Removed.

oswaldquek (dismissed):

kbottla (approved):
LGTM!!!

PR #912, opened 2018-12-13
PP-4508 Renamed TransactionType to TransactionSearchStrategyTransactionType

DanailMinchev:
## WHAT

- The name `TransactionType` is too generic and we want to introduce another
  enum `TransactionType` which will have `CHARGE` instead of `PAYMENT`.

with @alexbishop1 and @stephencdaly


SandorArpa (approved):
LOKTM

PR #913, opened 2018-12-13
PP-4508 TransactionType enum refactoring

DanailMinchev:
## WHAT

- Added new `TransactionType` enum to be used in `Transaction` instead of `String`
- Modified `TransactionSearchStrategy` to use the new enum
- Modified tests

related PR: https://github.com/alphagov/pay-connector/pull/912

with @alexbishop1


stephencdaly:
Looks good, my only question is whether we should use the enum values when constructing the query in TransactionDao. Although I've just been looking through some of that code and I don't fully understand it...

DanailMinchev:
I changed the `TransactionDao` to use the `TransactionType` enum and as discussed with @stephencdaly and @alexbishop1 we also don't fully understand why we have following conditions:
https://github.com/alphagov/pay-connector/blob/ca0ff467dfcd4fd48de8253c748640125e179bf0/src/main/java/uk/gov/pay/connector/charge/dao/TransactionDao.java#L156
https://github.com/alphagov/pay-connector/blob/ca0ff467dfcd4fd48de8253c748640125e179bf0/src/main/java/uk/gov/pay/connector/charge/dao/TransactionDao.java#L160
instead of making the false explicitly and not to execute the query.
We decided not to change the logic as it needs more time for research and we have other priorities currently

stephencdaly (approved):

PR #915, opened 2018-12-17
PP 4439 - Refactor PaymentProvider.refund() response

kbottla:
## WHAT
- This change simplifies the response object (GatewayResponse<BaseRefundResponse> --> GatewayRefundResponse) returned from PaymentProvider refund operation (inline with capture refactoring PR #898).
- ChargeRefundService has also been refactored to update refund based on state returned by PaymentProvider and not based on specific payment provider

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayRefundResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayRefundResponse.java
@@ -1,0 +1,60 @@
+package uk.gov.pay.connector.gateway.model.response;
+
+import uk.gov.pay.connector.gateway.model.GatewayError;
+
+import java.util.Optional;
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+import static uk.gov.pay.connector.gateway.model.GatewayError.genericGatewayError;
+
+public class GatewayRefundResponse {
+
+    private final Optional<String> reference;
+    private final GatewayRefundResponse.RefundState refundState;
+    private final GatewayError gatewayError;
+    private final String stringified;
+
+    private GatewayRefundResponse(Optional<String> reference, RefundState refundState,
+                                  GatewayError gatewayError, String stringified) {
+        this.reference = reference;
+        this.refundState = refundState;
+        this.gatewayError = gatewayError;
+        this.stringified = stringified;
+    }
+
+    public static GatewayRefundResponse fromGatewayError(GatewayError gatewayError) {
+        return new GatewayRefundResponse(Optional.empty(), RefundState.ERROR, gatewayError, gatewayError.toString());
+    }
+
+    public static GatewayRefundResponse fromBaseRefundResponse(BaseRefundResponse refundResponse, RefundState refundState) {
+        if (isNotBlank(refundResponse.getErrorCode()) || isNotBlank(refundResponse.getErrorMessage())) {
+            return new GatewayRefundResponse(refundResponse.getReference(), refundState, genericGatewayError(refundResponse.stringify()), refundResponse.stringify());
+        } else
+            return new GatewayRefundResponse(refundResponse.getReference(), refundState, null, refundResponse.stringify());
+    }
+
+    public Optional<GatewayError> getError() {
+        return Optional.ofNullable(gatewayError);
+    }
+
+    public GatewayRefundResponse.RefundState state() {
+        return refundState;
+    }
+
+    public Optional<String> getReference() {
+        return reference;
+    }
+
+    public boolean isSuccessful() {
+        return gatewayError == null;
+    }
+
+    public enum RefundState {
+        COMPLETE, PENDING, ERROR
+    }
+
+    @Override
+    public String toString() {
+        return stringified;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
@@ -8,9 +8,7 @@ import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
 import uk.gov.pay.connector.gateway.model.GatewayError;
 import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
-import uk.gov.pay.connector.gateway.model.response.BaseRefundResponse;
-import uk.gov.pay.connector.gateway.model.response.BaseResponse;
-import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.gateway.model.response.GatewayRefundResponse;
 import uk.gov.pay.connector.gateway.stripe.DownstreamException;
 import uk.gov.pay.connector.gateway.stripe.GatewayClientException;
 import uk.gov.pay.connector.gateway.stripe.GatewayException;
@@ -23,14 +21,15 @@ import javax.ws.rs.core.Response;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 
 import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
 import static javax.ws.rs.core.MediaType.APPLICATION_FORM_URLENCODED_TYPE;
-import static uk.gov.pay.connector.gateway.model.GatewayError.genericGatewayError;
 import static uk.gov.pay.connector.gateway.model.GatewayError.unexpectedStatusCodeFromGateway;
+import static uk.gov.pay.connector.gateway.model.response.GatewayRefundResponse.fromBaseRefundResponse;
 import static uk.gov.pay.connector.gateway.stripe.util.StripeAuthUtil.getAuthHeaderValue;
 
 public class StripeRefundHandler {
@@ -43,11 +42,9 @@ public class StripeRefundHandler {
         this.stripeGatewayConfig = stripeGatewayConfig;
     }
 
-    public GatewayResponse<BaseRefundResponse> refund(RefundGatewayRequest request) {
+    public GatewayRefundResponse refund(RefundGatewayRequest request) {
         String url = stripeGatewayConfig.getUrl() + "/v1/refunds";
         GatewayAccountEntity gatewayAccount = request.getGatewayAccount();
-
-        GatewayResponse.GatewayResponseBuilder<BaseResponse> responseBuilder = GatewayResponse.GatewayResponseBuilder.responseBuilder();
 
         try {
             String payload = URLEncodedUtils.format(buildPayload(request.getTransactionId(), request.getAmount()), UTF_8);
@@ -58,28 +55,28 @@ public class StripeRefundHandler {
                     APPLICATION_FORM_URLENCODED_TYPE,
                     format("gateway-operations.%s.%s.refund", gatewayAccount.getGatewayName(), gatewayAccount.getType()));
 
-            return responseBuilder.withResponse(StripeRefundResponse.of(response)).build();
+            return fromBaseRefundResponse(StripeRefundResponse.of((String) response.readEntity(Map.class).get("id")),
+                    GatewayRefundResponse.RefundState.COMPLETE);
 
         } catch (GatewayClientException e) {
 
             Response response = e.getResponse();
             StripeErrorResponse stripeErrorResponse = response.readEntity(StripeErrorResponse.class);
-            String errorId = UUID.randomUUID().toString();
-            logger.error("Refund failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
-                    request.getTransactionId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
-            GatewayError gatewayError = genericGatewayError(stripeErrorResponse.getError().getMessage());
-            return responseBuilder.withGatewayError(gatewayError).build();
+            StripeErrorResponse.Error error = stripeErrorResponse.getError();
+            logger.error("Refund failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
+                    request.getTransactionId(), error.getCode(), error.getMessage(), response.getStatus());
 
+            return fromBaseRefundResponse(
+                    StripeRefundResponse.of(request.getTransactionId(), error.getCode(), error.getMessage()),
+                    GatewayRefundResponse.RefundState.ERROR);
         } catch (GatewayException e) {
-
-            return responseBuilder.withGatewayError(GatewayError.of(e)).build();
-
+            return GatewayRefundResponse.fromGatewayError(GatewayError.of(e));
         } catch (DownstreamException e) {
             String errorId = UUID.randomUUID().toString();
             logger.error("Refund failed for transaction id {}. Reason: {}. Status code from Stripe: {}. ErrorId: {}",
                     request.getTransactionId(), e.getMessage(), e.getStatusCode(), errorId);
             GatewayError gatewayError = unexpectedStatusCodeFromGateway("An internal server error occurred. ErrorId: " + errorId);
-            return responseBuilder.withGatewayError(gatewayError).build();
+            return GatewayRefundResponse.fromGatewayError(gatewayError);
         }
     }
 
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
@@ -67,7 +67,7 @@ public class StripeRefundHandler {
                     request.getTransactionId(), error.getCode(), error.getMessage(), response.getStatus());
 
             return fromBaseRefundResponse(
-                    StripeRefundResponse.of(request.getTransactionId(), error.getCode(), error.getMessage()),
+                    StripeRefundResponse.of(error.getCode(), error.getMessage()),
                     GatewayRefundResponse.RefundState.ERROR);
         } catch (GatewayException e) {
             return GatewayRefundResponse.fromGatewayError(GatewayError.of(e));
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
@@ -1,10 +1,10 @@
 package uk.gov.pay.connector.gateway.stripe.response;
 
+import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.connector.gateway.model.response.BaseRefundResponse;
 
-import javax.ws.rs.core.Response;
-import java.util.Map;
 import java.util.Optional;
+import java.util.StringJoiner;
 
 public class StripeRefundResponse implements BaseRefundResponse {
     private String reference;
@@ -15,14 +15,43 @@ public class StripeRefundResponse implements BaseRefundResponse {
         this.reference = reference;
     }
 
-    public static StripeRefundResponse of(Response response) {
-        final String id = response.readEntity(Map.class).get("id").toString();
-        return new StripeRefundResponse(id);
+    private StripeRefundResponse(String reference, String errorCode, String errorMessage) {
+        this.reference = reference;
+        this.errorCode = errorCode;
+        this.errorMessage = errorMessage;
+    }
+
+    public static StripeRefundResponse of(String reference, String errorCode, String errorMessage) {
+        return new StripeRefundResponse(reference, errorCode, errorMessage);
+    }
+
+    public static StripeRefundResponse of(String reference) {
+        return new StripeRefundResponse(reference);
     }
 
     @Override
     public Optional<String> getReference() {
         return Optional.ofNullable(reference);
+    }
+
+    @Override
+    public String stringify() {
+        return toString();
+    }
+
+    @Override
+    public String toString() {
+        StringJoiner joiner = new StringJoiner(", ", "Stripe refund response (", ")");
+        if (StringUtils.isNotBlank(reference)) {
+            joiner.add("Refund gateway reference id: " + reference);
+        }
+        if (StringUtils.isNotBlank(errorCode)) {
+            joiner.add("error code: " + errorCode);
+        }
+        if (StringUtils.isNotBlank(errorMessage)) {
+            joiner.add("error: " + errorMessage);
+        }
+        return joiner.toString();
     }
 
     @Override
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
@@ -21,8 +21,8 @@ public class StripeRefundResponse implements BaseRefundResponse {
         this.errorMessage = errorMessage;
     }
 
-    public static StripeRefundResponse of(String reference, String errorCode, String errorMessage) {
-        return new StripeRefundResponse(reference, errorCode, errorMessage);
+    public static StripeRefundResponse of(String errorCode, String errorMessage) {
+        return new StripeRefundResponse(null, errorCode, errorMessage);
     }
 
     public static StripeRefundResponse of(String reference) {
--- a/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
+++ b/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
@@ -6,6 +6,11 @@ import uk.gov.pay.connector.common.model.domain.UTCDateTimeConverter;
 import java.sql.Timestamp;
 import java.time.ZonedDateTime;
 
+/**
+ * Entity which represents Refunds history (a record for each event for a particular refund)
+ * Each time a new status is set for RefundEntity a new record is created in RefundsHistory (Refunds_History table in database)
+ * which happens magically by @Customizer(HistoryCustomizer.class) annotation on RefundEntity
+ */
 public class RefundHistory extends RefundEntity {
 
     private ZonedDateTime historyStartDate;
--- a/src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
+++ b/src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
@@ -11,8 +11,7 @@ import uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
-import uk.gov.pay.connector.gateway.model.response.BaseRefundResponse;
-import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.gateway.model.response.GatewayRefundResponse;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.refund.dao.RefundDao;
 import uk.gov.pay.connector.refund.exception.RefundException;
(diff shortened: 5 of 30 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayRefundResponse.java
```diff
@@ -0,0 +1,60 @@
...
+    public Optional<GatewayError> getError() {
+        return Optional.ofNullable(gatewayError);
+    }
+
+    public GatewayRefundResponse.RefundState state() {
+        return refundState;
+    }
+
+    public Optional<String> getReference() {
+        return reference;
```
oswaldquek, line 45: this should be `return Optional.ofNullable(reference)` as you're passing in a null reference in line 26. also if there's an error it's not guaranteed the `refundResponse.getReference()` is not null (line 31).

oswaldquek (changes requested):
haven't gotten round to reviewing everything yet but just one comment so far

src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayRefundResponse.java
kbottla: 👍 Good spot. I have changed factory method to pass `Optional.empty()` instead of null. As reference is already Optional, can't use `return Optional.ofNulllable` straight away

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
```diff
@@ -58,28 +55,28 @@ public StripeRefundHandler(StripeGatewayClient client, StripeGatewayConfig strip
...
-            logger.error("Refund failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. ErrorId: {}. Response code from Stripe: {}",
-                    request.getTransactionId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), errorId, response.getStatus());
-            GatewayError gatewayError = genericGatewayError(stripeErrorResponse.getError().getMessage());
-            return responseBuilder.withGatewayError(gatewayError).build();
+            StripeErrorResponse.Error error = stripeErrorResponse.getError();
+            logger.error("Refund failed for transaction id {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
+                    request.getTransactionId(), error.getCode(), error.getMessage(), response.getStatus());
 
+            return fromBaseRefundResponse(
+                    StripeRefundResponse.of(request.getTransactionId(), error.getCode(), error.getMessage()),
```
oswaldquek: I don't think passing in the `request.getTransactionId()` as the reference is correct here. The reference is an id returned by Stripe as a result of a successful refund request. The `request.getTransactionId()` is actually the charge entity's gateway transaction id: https://github.com/alphagov/pay-connector/blob/32a0acca2b169c9c4ca34d2e8d54bcf45f293ba4/src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java#L59 and is not the reference.

since `GatewayClientException` is a 4xx error and Stripe seems to return something like
```
{
  "error": {
    "message": "The transfer tr_blah_blah_blah is already fully reversed.",
    "type": "invalid_request_error"
  }
}
```
 i think the reference should be null in this case.

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
```diff
@@ -15,16 +15,45 @@ private StripeRefundResponse(String reference) {
...
-    public static StripeRefundResponse of(Response response) {
-        final String id = response.readEntity(Map.class).get("id").toString();
-        return new StripeRefundResponse(id);
+    private StripeRefundResponse(String reference, String errorCode, String errorMessage) {
+        this.reference = reference;
+        this.errorCode = errorCode;
+        this.errorMessage = errorMessage;
+    }
+
+    public static StripeRefundResponse of(String reference, String errorCode, String errorMessage) {
```
oswaldquek: since reference might be null in the event of an error it's likely you'll need a constructor that only takes an errorCode and errorMessage

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
```diff
@@ -104,34 +104,47 @@ public RefundEntity createRefund(Long accountId, String chargeId, RefundRequest
...
 
-            refundEntity.setStatus(status);
-            getRefundReference(refundEntity, gatewayResponse).ifPresent(refundEntity::setReference);
+            if (refundStatus == REFUNDED) {
+                refundEntity.setStatus(REFUND_SUBMITTED);
+            }
+            if (chargeEntity.getPaymentGatewayName() == PaymentGatewayName.SANDBOX
+                    && refundEntity.hasStatus(RefundStatus.REFUND_SUBMITTED)) {
+                userNotificationService.sendRefundIssuedEmail(refundEntity);
+            }
```
oswaldquek, line 135: lines 121-127. one of the purposes of https://github.com/alphagov/pay-connector/pull/898 really is to make the `CardCaptureService` payment provider agnostic. i think you can do the same here instead of relying on the payment provider name? we should do this because Stripe and Sandbox payment providers both return a COMPLETE status, and therefore a REFUNDED status here.

oswaldquek: see https://github.com/alphagov/pay-connector/blob/beea6fa714eaf76f1531a536cabf2469d6484a18/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java#L344-L355 as an example of making it payment provider agnostic.

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gateway/model/response/GatewayRefundResponse.java
oswaldquek: 👍

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
kbottla: 👍 Removed reference

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
kbottla: Added new factory method for only errorCode and errorMessage

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
kbottla: As far the refund status is concerned, we are not depending anymore on payment provider and relying on refund state returned.
Only reason we still see payment provider and can't extract entire update logic is because of email notifications sent only for sandbox.

oswaldquek: I don't think sending a notification is a sandbox specific thing. It seems logical to me that if a refund's status is REFUNDED, the user should expect an email notification, regardless of whether the PSP was Stripe or Sandbox. but check with @mrlumbu or @alexbishop1

oswaldquek: Also I think lines 121-123 
```
if (refundStatus == REFUNDED) {
             getRefundReference(refundEntity, gatewayResponse).ifPresent(refundEntity::setReference);	                 refundEntity.setStatus(REFUND_SUBMITTED);
             }
```
is very confusing. The question is why is this being done??

Generally this section of code needs a bit more clarity.

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
oswaldquek: 👍 
`public static StripeRefundResponse of(String reference, String errorCode, String errorMessage)` isn't being used anymore though so can be removed.

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
kbottla: This is not a new functionality introduced.

https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java#L128-L132

kbottla: Not sure how lines 121-123 became
`if (refundStatus == REFUNDED) {
             getRefundReference(refundEntity, gatewayResponse).ifPresent(refundEntity::setReference);	                 refundEntity.setStatus(REFUND_SUBMITTED);
             }`

This is the code that I can see
`if (refundStatus == REFUNDED) {
                refundEntity.setStatus(REFUND_SUBMITTED);
            }`

oswaldquek: Sorry, not sure how I copied that wrong. but `if (refundStatus == REFUNDED) { refundEntity.setStatus(REFUND_SUBMITTED); }` is very confusing. why is this being done? there doesn't seem to be any rationale behind this code.

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeRefundResponse.java
kbottla: Oops! Removed unused factory method

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
oswaldquek: "This is not a new functionality introduced."

Yes, of course. But we've introduced a new payment provider Stripe. It makes sense to send email notifications for any provider which returns a COMPLETE status.

kbottla: Depends on what our functionality is. 
- Do we send refund email notifications to users (at any point) ? cc @alexbishop1 @mrlumbu.

As far as I can see it is only for Sandbox that we send notifications.

kbottla: > if (refundStatus == REFUNDED) { refundEntity.setStatus(REFUND_SUBMITTED); } is very confusing. why is this being done? there doesn't seem to be any rationale behind this code.

This will enable to store history of events for refunds and also displayed (for services) in frontend as:
```
Refunded                           amount   date 
Refund Submitted by [user email]   amount   date 
```

mrlumbu: @kbottla no, we definitely send them for all gateways, it's just that I believe we send them once we have received a successful refund notification for ePDQ, Worldpay, Smartpay, which is why we have that hacky bit of code for sandbox. I can dig out the code in a bit, or I'm sure you can find it

kbottla: We don't at the moment. It is been missed out on refactoring notifications. Just raised another PR for the fix

oswaldquek: @kbottla the other PR is for worldpay. i believe we should send refund notification for Stripe which means we can get rid of references to the Sandbox payment provider here.

oswaldquek: basically, `if (refundStatus == REFUNDED)`, send refund notification.

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
```diff
@@ -104,34 +103,44 @@ public RefundEntity createRefund(Long accountId, String chargeId, RefundRequest
...
                     refundEntity.getExternalId(), chargeEntity.getPaymentGatewayName().getName(), refundEntity.getReference(),
                     chargeEntity.getExternalId(), chargeEntity.getPaymentGatewayName().getName(), chargeEntity.getGatewayTransactionId(),
                     chargeEntity.getGatewayAccount().getAnalyticsId(), chargeEntity.getGatewayAccount().getId(),
-                    gatewayResponse, refundEntity.getStatus(), status);
+                    gatewayRefundResponse, refundEntity.getStatus(), refundStatus);
 
-            refundEntity.setStatus(status);
-            getRefundReference(refundEntity, gatewayResponse).ifPresent(refundEntity::setReference);
+            if (refundStatus == REFUNDED) {
+                refundEntity.setStatus(REFUND_SUBMITTED);
```
oswaldquek: i understand this is for storing history of events for refunds, but where is this history stored? 

if you look at 

https://github.com/alphagov/pay-connector/blob/beea6fa714eaf76f1531a536cabf2469d6484a18/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java#L344-L355

it's very clear about this.

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
kbottla: The other PR is for RefundNotificationsProcessor (which is a fix). I have already made suggested changes (`if (refundStatus == REFUNDED)`) for Stripe.

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
kbottla: RefundHistory is being handled by Customiser and eclipselink here
https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundEntity.java#L52

We can look at refactoring history writing separately if any improvements are needed.

oswaldquek: cool, that HistoryCustomizer looks interesting. i don't think any refactoring is needed. i'm more concerned about readability and understanding code - reading it as if i were a new developer on the project. i think there should be some comment then, about why you're setting the RefundEntity's status to REFUND_SUBMITTED. 
i think it'll also be good to modify SandboxRefundITest to assert the RefundHistory of a refund processed by Sandbox payment provider has a REFUND_SUBMITTED and REFUNDED status.

basically i think everything could be collapsed to:
```
if (refundStatus == REFUNDED) {
                refundEntity.setStatus(REFUND_SUBMITTED); //some comment about why we're doing this
                userNotificationService.sendRefundIssuedEmail(refundEntity);
                refundEntity.setStatus(REFUNDED);
            } else refundEntity.setStatus(refundStatus);
```

oswaldquek: sorry to be a pain

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
```diff
@@ -104,34 +103,47 @@ public RefundEntity createRefund(Long accountId, String chargeId, RefundRequest
...
                     refundEntity.getExternalId(), chargeEntity.getPaymentGatewayName().getName(), refundEntity.getReference(),
                     chargeEntity.getExternalId(), chargeEntity.getPaymentGatewayName().getName(), chargeEntity.getGatewayTransactionId(),
                     chargeEntity.getGatewayAccount().getAnalyticsId(), chargeEntity.getGatewayAccount().getId(),
-                    gatewayResponse, refundEntity.getStatus(), status);
+                    gatewayRefundResponse, refundEntity.getStatus(), refundStatus);
+
+            if (refundStatus == REFUNDED) {
+                // If the gateway confirms refunds immediately, insert history (REFUND_SUBMITTED) to record
+                // additional event for refund submitted date which also includes the user submitting the
+                // refund. This will also help services to view refund history in detail in self service
```
oswaldquek: Can this be changed to ```If the gateway confirms refunds immediately, the refund status needs to be set to REFUND_SUBMITTED and then REFUNDED. This will help services to view refund history in detail in self service. By setting the refund status like `refundEntity.setStatus(REFUND_SUBMITTED)` for example, the refund status history is magically captured by the @Customizer(HistoryCustomizer.class) annotation on RefundEntity.```

Again I'm approaching this with the eyes of a new developer and I think it's important to tell HOW this history is persisted in the database, rather than saying "insert history to record".

oswaldquek (changes requested):
sorry just one change requested to make the comment clearer for developers.

src/main/java/uk/gov/pay/connector/refund/service/ChargeRefundService.java
kbottla: I see your point. But I am not comfortable mixing up why (business desc) and how (tech choices) and also anytime (which is in multiple places) we use `refund.setStatus` \@Customiser comes into action. I feel this may not be right place to describe tech implementation

src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
```diff
@@ -6,6 +6,11 @@
 import java.sql.Timestamp;
 import java.time.ZonedDateTime;
 
+/**
+ * Entity which represents Refunds history (a record for each event for a particular refund)
+ * Each time a new status is set for RefundEntity a new record is created in RefundsHistory (Refunds_History table in database)
+ * which happens magically by @Customizer(HistoryCustomizer.class) annotation on RefundEntity
+ */
```
oswaldquek, line 13: we don't need this comment anymore. i originally thought it was the `@Customizer(HistoryCustomizer.class)` that updates the refunds_history table but it's actually the `@Transactional` annotation on a method that causes the refund entity and the history table to be updated.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -101,6 +108,8 @@ public void shouldRespond_412_WhenSecondRefundRequestAmountAvailableMismatches()
         List<Map<String, Object>> refundsFoundByChargeId = databaseTestHelper.getRefundsByChargeId(defaultTestCharge.getChargeId());
         assertThat(refundsFoundByChargeId.size(), is(1));
         assertThat(refundsFoundByChargeId, hasItems(aRefundMatching(refundId, is(notNullValue()), defaultTestCharge.getChargeId(), refundAmount, "REFUNDED")));
+
+        assertRefundsHistoryInDBForSuccessfulOrPartialRefund(defaultTestCharge);
```
oswaldquek: this can be deleted as it is already asserted in the `shouldBeAbleToRequestARefund_partialAmount` test which is sufficient. no need to go overboard here.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -114,6 +123,8 @@ public void shouldBeAbleToRequestARefund_fullAmount() {
...
```
oswaldquek: this can be deleted as it is already asserted in the `shouldBeAbleToRequestARefund_partialAmount` test which is sufficient. no need to go overboard here.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -136,6 +147,8 @@ public void shouldBeAbleToRequestARefund_multiplePartialAmounts_andRefundShouldB
                 aRefundMatching(secondRefundId, is(notNullValue()), chargeId, secondRefundAmount, "REFUNDED"),
                 aRefundMatching(firstRefundId, is(notNullValue()), chargeId, firstRefundAmount, "REFUNDED")));
 
+        assertRefundsHistoryInDBForTwoRefunds(defaultTestCharge);
```
oswaldquek: can delete this as already asserted in `shouldBeAbleToRefundTwoRequestsWhereAmountAvailableMatch`.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -176,13 +189,15 @@ public void shouldFailRequestingARefund_whenChargeRefundIsFull() {
         postRefundFor(externalChargeId, refundAmount, defaultTestCharge.getAmount())
                 .statusCode(ACCEPTED.getStatusCode());
 
+        List<Map<String, Object>> refundsFoundByChargeId = databaseTestHelper.getRefundsByChargeId(chargeId);
+        assertThat(refundsFoundByChargeId.size(), is(1));
```
oswaldquek, line 193: don't think we need this assertion.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -176,13 +189,15 @@ public void shouldFailRequestingARefund_whenChargeRefundIsFull() {
...
+
+        assertRefundsHistoryInDBForSuccessfulOrPartialRefund(defaultTestCharge);
```
oswaldquek: don't think we need this assertion.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -196,6 +211,9 @@ public void shouldFailRequestingARefund_whenAmountIsBiggerThanChargeAmount() {
 
         List<Map<String, Object>> refundsFoundByChargeId = databaseTestHelper.getRefundsByChargeId(defaultTestCharge.getChargeId());
         assertThat(refundsFoundByChargeId.size(), is(0));
+
+        List<String> refundsHistory = databaseTestHelper.getRefundsHistoryByChargeId(defaultTestCharge.getChargeId()).stream().map(x -> x.get("status").toString()).collect(Collectors.toList());
+        assertThat(refundsHistory.size(), is(0));
```
oswaldquek, line 214: this can be simplified by using `assertThat(refundDao.searchHistoryByChargeId(defaultTestCharge.getChargeId()), hasSize(0))`

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -210,6 +228,9 @@ public void shouldFailRequestingARefund_whenAmountIsBiggerThanAllowedChargeAmoun
 
         List<Map<String, Object>> refundsFoundByChargeId = databaseTestHelper.getRefundsByChargeId(defaultTestCharge.getChargeId());
         assertThat(refundsFoundByChargeId.size(), is(0));
+        
+        List<String> refundsHistory = databaseTestHelper.getRefundsHistoryByChargeId(defaultTestCharge.getChargeId()).stream().map(x -> x.get("status").toString()).collect(Collectors.toList());
+        assertThat(refundsHistory.size(), is(0));
```
oswaldquek, line 231: don't need this assertion as already tested

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -224,6 +245,9 @@ public void shouldFailRequestingARefund_whenAmountIsLessThanOnePence() {
 
         List<Map<String, Object>> refundsFoundByChargeId = databaseTestHelper.getRefundsByChargeId(defaultTestCharge.getChargeId());
         assertThat(refundsFoundByChargeId.size(), is(0));
+
+        List<String> refundsHistory = databaseTestHelper.getRefundsHistoryByChargeId(defaultTestCharge.getChargeId()).stream().map(x -> x.get("status").toString()).collect(Collectors.toList());
+        assertThat(refundsHistory.size(), is(0));
```
oswaldquek, line 248: don't need this assertion as already tested

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -246,6 +270,8 @@ public void shouldFailRequestingARefund_whenAPartialRefundMakesTotalRefundedAmou
         List<Map<String, Object>> refundsFoundByChargeId1 = databaseTestHelper.getRefundsByChargeId(defaultTestCharge.getChargeId());
         assertThat(refundsFoundByChargeId1.size(), is(1));
         assertThat(refundsFoundByChargeId1, hasItems(aRefundMatching(firstRefundId, is(notNullValue()), defaultTestCharge.getChargeId(), firstRefundAmount, "REFUNDED")));
+
+        assertRefundsHistoryInDBForSuccessfulOrPartialRefund(defaultTestCharge);
```
oswaldquek: don't need this assertion as already tested

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -280,4 +306,16 @@ private String assertRefundResponseWith(Long refundAmount, ValidatableResponse v
 
         return refundId;
     }
+
+    private void assertRefundsHistoryInDBForSuccessfulOrPartialRefund(DatabaseFixtures.TestCharge defaultTestCharge) {
+        List<String> refundsHistory = databaseTestHelper.getRefundsHistoryByChargeId(defaultTestCharge.getChargeId()).stream().map(x -> x.get("status").toString()).collect(Collectors.toList());
```
oswaldquek, line 309: i'd rather `refundDao.searchHistoryByChargeId` was used as you get the `List<RefundHistory>` returned. it will be quite easy to order the list based on historyStartDate. because i'm not sure asserting events "containsInAnyOrder" is actually correct here.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -280,4 +306,16 @@ private String assertRefundResponseWith(Long refundAmount, ValidatableResponse v
...
+        assertThat(refundsHistory.size(), is(3));
+        assertThat(refundsHistory, containsInAnyOrder("REFUND SUBMITTED", "REFUNDED", "CREATED"));
+    }
+
+    private void assertRefundsHistoryInDBForTwoRefunds(DatabaseFixtures.TestCharge defaultTestCharge) {
+        List<String> refundsHistory = databaseTestHelper.getRefundsHistoryByChargeId(defaultTestCharge.getChargeId()).stream().map(x -> x.get("status").toString()).collect(Collectors.toList());
+        assertThat(refundsHistory.size(), is(6));
+        assertThat(refundsHistory, containsInAnyOrder("REFUND SUBMITTED", "REFUNDED", "CREATED", "REFUND SUBMITTED", "REFUNDED", "CREATED"));
```
oswaldquek: same comment as above

oswaldquek (changes requested):
i'm happy with the implementation. just two things: 1) testing the event history once is enough... we don't need to assert the event history for every test as it will further pollute existing tests and affect readability. in fact i'd prefer if you just created an assertEventHistoryForASuccessfulRefund test 2) it's better to use the refundDao to get event history as you can verify the order in which events happen as they have a timestamp which you can use to sort.

src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
kbottla: It is still @Customizer(HistoryCustomizer.class) that is responsible to insert into history.

Transactional is more about enabling boundaries for the operations and if we do multiple operations in a single transaction they won't be treated individually.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
kbottla: We do it refunds record. It is better to have refunds history together. 
Having only one test won't cover full/partial refunds and also if that tests gets removed we are back to square 1. I feel we should add them wherever we check for refunds.

src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
oswaldquek: ok, i understand 👍

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
oswaldquek: ?? it's already tested in `shouldBeAbleToRequestARefund_partialAmount`. all i'm saying is you don't have to go overboard and re-test it here (and for nearly every method in this test class).

kbottla: May be it is confusing to have common method ?
My intention is, if each test is considered on its own it should cover `issue refund, check response, check refunds & refunds history in DB for that particular transaction`

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
kbottla: Not against using history_start_date. Updated query to order by statuses and removed  containsInAnyOrder.

kbottla:
> 1) testing the event history once is enough... we don't need to assert the event history for every test as it will further pollute existing tests and affect readability. in fact i'd prefer if you just created an assertEventHistoryForASuccessfulRefund test 

I don't think it as a right place to test refundHistory alone as these are resource tests. As each test is for different scenario we verify `response, refund record from DB` in each test. I have extended tests to check for refunds history. Anyone looking at an individual test can understand different aspects to be looked out for these tests.

> 2) it's better to use the refundDao to get event history as you can verify the order in which events happen as they have a timestamp which you can use to sort.

Updated helper to sort by date and checking the statuses in order. Felt it a bit like polluting test class if mixing helper and dao for different reasons.

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundITest.java
```diff
@@ -176,13 +189,15 @@ public void shouldFailRequestingARefund_whenChargeRefundIsFull() {
         postRefundFor(externalChargeId, refundAmount, defaultTestCharge.getAmount())
                 .statusCode(ACCEPTED.getStatusCode());
 
+        List<Map<String, Object>> refundsFoundByChargeId = databaseTestHelper.getRefundsByChargeId(chargeId);
+        assertThat(refundsFoundByChargeId.size(), is(1));
+
+        assertRefundsHistoryInOrderInDBForSuccessfulOrPartialRefund(defaultTestCharge);
```
oswaldquek: we definitely don't need this assertion here as this is already done in the `shouldBeAbleToRequestARefund_fullAmount` test. this test (`shouldFailRequestingARefund_whenChargeRefundIsFull`) makes a full refund as well before making another refund. so the event history is already tested.

oswaldquek: we should keep code as clean as possible and not add unnecessary tests

kbottla: 👍 Removed.

oswaldquek (approved):

oswaldquek:
I actually think we should test refund history as it's own test as the end user can see this history. This is got by calling the `"/v1/api/accounts/{accountId}/charges/{chargeId}/events"` endpoint. So it makes perfect sense to have a resource level test where a refund is made and we assert the event history. This PR has gone on for long enough though so will approve. I will create the test separately.

kbottla:
There is probably a test already although I haven't looked into its details
https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceITest.java#L92

PR #916, opened 2018-12-17
bau: extra charge logging

oswaldquek:
We are encountering issues when making Stripe charges, where if we make enough
requests the payment hangs at the "auth waiting" window in frontend
indefinitely. Adding extra logging to ascertain where this is happening. We
suspect a timeout error but can't see this being caught anywhere.

@oswaldquek

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
@@ -101,7 +101,9 @@ public class StripePaymentProvider implements PaymentProvider {
                 .responseBuilder();
 
         try {
+            logger.info("Creating token for charge external id " + request.getChargeExternalId()); //FIXME remove when Stripe performance issues have been sorted out. Ref: https://github.com/alphagov/pay-connector/pull/916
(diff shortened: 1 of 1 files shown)


IreneLau-GDS (commented):
If these are just for debugging these issues, we should probably put a FIXME in the code or create a story to remove them. Otherwise we will forget and the log will be extremely noise after we go-live.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -136,6 +139,7 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
         } catch (DownstreamException e) {
             return unexpectedGatewayResponse(request.getChargeExternalId(), responseBuilder, e.getMessage(), e.getStatusCode());
```
IreneLau-GDS, line 140: Does this log as well?

oswaldquek: yes the `unexpectedGatewayResponse` logs already

IreneLau-GDS (approved):

PR #918, opened 2018-12-17
PP-4548 Restore 3D Secure authorisation response logging

alexbishop1:
When logging the result of an attempt to authorise the result of a 3D Secure authentication with Worldpay, Smartpay or ePDQ, log more information from the response itself or the error that meant we could not get a response.

This restores the logging to a similar level of detail to what we used to record (though for implementation reasons the logging is not identical to previously).

Stripe never had this level of logging and this commit does not attempt to change that (though we should examine this before we go live with Stripe).

The logging relies on new stringification code in `Gateway3DSAuthorisationResponse`. It would be nice if the new code were similar to the code in the new `CaptureResponse` class that helps with logging the results of attempts to capture, but the two classes seem to have different ideas about how much they want to know about the `BaseResponse`. We should attempt to bridge this divide in the future.

Ideally there would be some tests for the logging (to stop it being accidentally degraded again) but this commit does not add any in the interests of restoring our logging detail quickly.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
@@ -151,11 +151,12 @@ public class EpdqPaymentProvider implements PaymentProvider {
                     authoriseStatus.name()))
                     .inc();
             gatewayResponse = reconstructErrorBiasedGatewayResponse(gatewayResponse, authoriseStatus, auth3DResult);
-
         }
-
-        return gatewayResponse.getBaseResponse().map(baseResponse -> Gateway3DSAuthorisationResponse.of(baseResponse.authoriseStatus(), baseResponse.getTransactionId()))
-                .orElseGet(() -> Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.EXCEPTION));
+        
+        String gatewayResponseString = gatewayResponse.toString();
+        return gatewayResponse.getBaseResponse()
+                .map(baseResponse -> Gateway3DSAuthorisationResponse.of(gatewayResponseString, baseResponse.authoriseStatus(), baseResponse.getTransactionId()))
+                .orElseGet(() -> Gateway3DSAuthorisationResponse.of(gatewayResponseString, BaseAuthoriseResponse.AuthoriseStatus.EXCEPTION));
     }
 
     @Override
--- a/src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
@@ -9,22 +9,28 @@ import static uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse.
 public class Gateway3DSAuthorisationResponse {
     private final BaseAuthoriseResponse.AuthoriseStatus authorisationStatus;
     private final String transactionId;
+    private final String stringified;
 
-    private Gateway3DSAuthorisationResponse(BaseAuthoriseResponse.AuthoriseStatus authorisationStatus, String transactionId) {
+    private Gateway3DSAuthorisationResponse(BaseAuthoriseResponse.AuthoriseStatus authorisationStatus, String transactionId, String stringified) {
         this.transactionId = transactionId;
         this.authorisationStatus = authorisationStatus;
+        this.stringified = stringified;
+    }
+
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
```diff
@@ -48,4 +54,9 @@ public ChargeStatus getMappedChargeStatus() {
     public Optional<String> getTransactionId() {
         return Optional.ofNullable(transactionId);
     }
+
```
georgeracu, line 57: Missing an `@Override` to signal our intention to override the base method.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -151,11 +151,12 @@ public Gateway3DSAuthorisationResponse authorise3dsResponse(Auth3dsResponseGatew
                     authoriseStatus.name()))
                     .inc();
             gatewayResponse = reconstructErrorBiasedGatewayResponse(gatewayResponse, authoriseStatus, auth3DResult);
-
         }
-
-        return gatewayResponse.getBaseResponse().map(baseResponse -> Gateway3DSAuthorisationResponse.of(baseResponse.authoriseStatus(), baseResponse.getTransactionId()))
-                .orElseGet(() -> Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.EXCEPTION));
+        
+        String gatewayResponseString = gatewayResponse.toString();
```
georgeracu, line 156: I think that sending this string representation one level up to the service, just to be logged shows that we don't separate our layers very well. This string can be logged at this level and the service doesn't need to know about the gateway response at all. For the sake of re-enabling logging I will leave it as it is.

src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
```diff
@@ -9,22 +9,28 @@
 public class Gateway3DSAuthorisationResponse {
     private final BaseAuthoriseResponse.AuthoriseStatus authorisationStatus;
     private final String transactionId;
+    private final String stringified;
 
-    private Gateway3DSAuthorisationResponse(BaseAuthoriseResponse.AuthoriseStatus authorisationStatus, String transactionId) {
+    private Gateway3DSAuthorisationResponse(BaseAuthoriseResponse.AuthoriseStatus authorisationStatus, String transactionId, String stringified) {
         this.transactionId = transactionId;
         this.authorisationStatus = authorisationStatus;
+        this.stringified = stringified;
```
georgeracu, line 17: This string representation is not an actual representation of this object. Calling it `stringified` is misleading. If we override the `toString()` method doesn't return the same value as `stringified` holds. As I said above, this doesn't have to bubble all the way up to the service level. We might very well call it something like `gatewayRawResponse` to get a better description of what it is. For the sake of re-enabling logging I will leave it as it is.

georgeracu (approved):
This enabled a more verbose logging of the response (or of the raw response from the gateway). There are a few comments related to the way we propagate things outside the context boundaries. It would be useful to have a test for this fix. For the sake of re-enabling logging I will leave it as it is.

src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
alexbishop1: It’s not a great name; I picked it for consistency with `CaptureResponse`

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
alexbishop1: The advantage of logging at the service level is that we have enough information at that point to say (consistently across all gateways and for all possible outcomes), “We tried to do a 3D Secure authorisation for _this payment_ and we got _this result_, therefore we’re changing the payment from _this status_ to _this status_.”

But I’m certainly no fan of passing strings about like this (or objects representing the response, which is what we do elsewhere in the code).

src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
alexbishop1: Whoops, thought my IDE would have added that for me: time to tinker with the settings!

PR #922, opened 2018-12-19
PP-4522: Add test and live token configuration

oswaldquek:
This commit allows the configuration of a test and live token. We want to be able to support test and live accounts in production. The idea is that we can do smoke test payments in production using a test token. At the time of this commit, we aren't taking stripe payments in production yet. This commit only concerns itself with adding the live token configuration without breaking  existing functionality. As such, all tokens used will be the test token.

@oswaldquek

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/StripeAuthTokens.java
+++ b/src/main/java/uk/gov/pay/connector/app/StripeAuthTokens.java
@@ -1,0 +1,24 @@
+package uk.gov.pay.connector.app;
+
+import io.dropwizard.Configuration;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+public class StripeAuthTokens extends Configuration {
+
+    @Valid
+    @NotNull
+    private String test;
+
+    @Valid
+    private String live;
+
+    public String getTest() {
+        return test;
+    }
+
+    public String getLive() {
+        return live;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/util/StripeAuthUtil.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/util/StripeAuthUtil.java
@@ -8,7 +8,7 @@ public class StripeAuthUtil {
     }
 
     public static String getAuthHeaderValue(StripeGatewayConfig stripeGatewayConfig) {
-        return "Bearer " + stripeGatewayConfig.getAuthToken();
+        return "Bearer " + stripeGatewayConfig.getAuthTokens().getTest();
     }
 
 }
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/util/StripeAuthUtil.java
```diff
@@ -8,7 +8,7 @@ private StripeAuthUtil() {
     }
 
     public static String getAuthHeaderValue(StripeGatewayConfig stripeGatewayConfig) {
-        return "Bearer " + stripeGatewayConfig.getAuthToken();
+        return "Bearer " + stripeGatewayConfig.getAuthToken().getTest();
```
kbottla: Won't it always be test token ?

src/main/java/uk/gov/pay/connector/app/StripeAuthTokens.java
```diff
@@ -0,0 +1,24 @@
...
+import javax.validation.constraints.NotNull;
+
+public class StripeAuthTokens extends Configuration {
+
+    @Valid
+    @NotNull
+    private String test;
+
+    @Valid
+    private String live;
```
kbottla, line 15: What is expected when live is not available on production but test is ?

src/main/java/uk/gov/pay/connector/gateway/stripe/util/StripeAuthUtil.java
kbottla: We have something similar https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java#L62
so can return based on gateway account type

src/main/java/uk/gov/pay/connector/app/StripeAuthTokens.java
kbottla: We can probably make this NonNull too (with defaulted to some string). For live gateway accounts test token can be ignored and vice versa.

oswaldquek: if production token is not available, and a live stripe gateway account exists in connector and a payment is made with that account, a 500 will be returned. Please see the acceptance criteria here https://payments-platform.atlassian.net/browse/PP-4522. The purpose of this PR is really just to add the configuration without breaking existing functionality.

The production token can't be NonNull as we don't want to populate it for test, perf, and staging environments.

src/main/java/uk/gov/pay/connector/gateway/stripe/util/StripeAuthUtil.java
oswaldquek: it won't always be test token as we have to make way for the possibility for a live token now, depending on whether the gateway account is a test or live one

kbottla (dismissed):

kbottla (dismissed):

kbottla (approved):

PR #923, opened 2018-12-20
PP-4522: Use test or live token according to gateway account type

oswaldquek:
@oswaldquek

## HOW 
nothing much to say, the title is self explanatory.


georgeracu (approved):
Only a small comment but everything else is 👍

PR #924, opened 2018-12-21
PP-4553 Tests for User email notifications (Refunds)

kbottla:
## WHAT

Additional asserts to test user email notifications. Although integration (Resource) tests would be ideal this will need further investigation as per https://payments-platform.atlassian.net/browse/PP-4569

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
@@ -21,7 +21,7 @@ public class RefundNotificationProcessor {
     private UserNotificationService userNotificationService;
 
     @Inject
-    RefundNotificationProcessor(RefundDao refundDao,
+    public RefundNotificationProcessor(RefundDao refundDao,
                                 UserNotificationService userNotificationService) {
         this.refundDao = refundDao;
         this.userNotificationService = userNotificationService;
--- a/src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqNotificationServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqNotificationServiceTest.java
@@ -12,11 +12,15 @@ import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.gateway.processor.ChargeNotificationProcessor;
 import uk.gov.pay.connector.gateway.processor.RefundNotificationProcessor;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.refund.dao.RefundDao;
+import uk.gov.pay.connector.refund.model.domain.RefundEntity;
+import uk.gov.pay.connector.usernotification.service.UserNotificationService;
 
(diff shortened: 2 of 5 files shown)


oswaldquek:
Hmm...what is the issue that has caused this PR to be raised? 
It seems like we're already verifying the emails get sent indirectly. For example, in `EpdqNotificationServiceTest`, we already do this verification:
```
verify(mockRefundNotificationProcessor).invoke(EPDQ, RefundStatus.REFUNDED, payId + "/" + payIdSub, payId);
```
Then in `RefundNotificationProcessorTest` we already do this verification:
```
verify(userNotificationService).sendRefundIssuedEmail(refundEntity);
```

I also question the usage of the `Mockito.spy`. It's javadoc says `Real spies should be used carefully and occasionally, for example when dealing with legacy code.... However, I wouldn't use partial mocks for new, test-driven & well-designed code.`

kbottla:
Issue was `missing email notifications` for successful refunds. The below block was missing (before the fix) and couldn't be detected by existing tests

>  if (RefundStatus.REFUNDED.equals(newStatus)) {
>             userNotificationService.sendRefundIssuedEmail(refundEntity);
>         }

`verify(userNotificationService).sendRefundIssuedEmail(refundEntity);`
 has been added recently when fixing the bug. But rewriting the RefundNotificationProcessor (which happened during Notification refactoring) could potentially introduce the bug if not caught at high level. 

Integration would have been ideal but is currently blocked due to issues as per https://payments-platform.atlassian.net/browse/PP-4569.

I am aware it is crossing the boundary of unit tests but welcome any other suggestions to test this with integration.

oswaldquek:
Yes, this is crossing the boundary of unit tests. Unit testing of the `EpdqNotificationService` should only care that the `refundNotificationProcessor` was invoked. It is not the responsibility of the `EpdqNotificationServiceTest` to verify the refund email was sent. It over complicates the test and it will potentially cause confusion for developers reading the test. Personally I think the existing `verify(userNotificationService).sendRefundIssuedEmail(refundEntity)` will suffice.

If the wiremock https thing doesn't work, just make it http?

kbottla:
Having an integration test is ideal. As I mentioned before, if someone replaces or factor out RefundNotificationProcessor (which happened once) there is nothing else that tests if emails are being sent for notifications.

We can't use http. NotificationClient (notifications-java-client library) only accepts https url.

oswaldquek:
But you already have
```
    @Test
    public void shouldInvokeSendEmailNotificationsForSuccessfulRefunds() {
        refundNotificationProcessor.invoke(paymentGatewayName, RefundStatus.REFUNDED, reference, transactionId);
        verify(userNotificationService).sendRefundIssuedEmail(refundEntity);
    }
```
I don't see how this is insufficient to catch out any refactoring.

kbottla:
It could happen if `RefundNotificationProcessor` is removed and code is copied over to something like `Charge/RefundService` (without carrying over `sendRefundIssuedEmail`) which is what happened in refactoring that introduced this bug. 

Although, it is very unlikely we will do the above anytime soon/ever, but it is still a possibility.

I am happy to close this PR if the only way we want is to have a proper integration test (with HTTPS request mocked).

kbottla:
Closing as there is an integration test https://github.com/alphagov/pay-connector/pull/935 now

PR #926, opened 2018-12-27
PP-4550 Fix Stripe Client bug

georgeracu:
## WHAT

- Fixed the case when `StripeGatewayClient` was hanging on requests after a while.
This behaviour has been linked to the Jersey Client where we didn't close
the `Response`.
- Given that we were propagating the `Response` object from the client a few
levels up to be handled by the caller of the client, the lifespan of this
object was long and it was keeping busy HTTP connections. The client now
returns a custom Response object, not a `javax.ws.rs.core.Response` one.
This new object type holds the response body, the status and family of the
status from the initial Response. The client will send back this new custom
object and any classes interested in using it will use the String payload
to map into the object type they need (if the payload is returned).
- Updated the tests to match the new response type
- Replaced a `WireMockRule` with a class rule, to keep the server running
between test classes. This change fixed a random test failure, that was
failing because the WireMock server failed to respond to the HTTP request
- Updated tests such that mocks will match against known URIs and not `any()`
- Deleted base test class as it was used only in one child class
- Upgraded WireMock to 2.20.0


src/main/java/uk/gov/pay/connector/gateway/stripe/Stripe3dsSourceAuthorisationResponse.java
```diff
@@ -44,8 +51,13 @@ public String getErrorMessage() {
         return null;
     }
 
-    public static Stripe3dsSourceAuthorisationResponse of(Response response) {
-        return new Stripe3dsSourceAuthorisationResponse(response.readEntity(Stripe3dsSourceResponse.class));
+    public static Stripe3dsSourceAuthorisationResponse from(String jsonResponse) {
+        try {
+            Stripe3dsSourceResponse stripe3dsSourceResponse = new ObjectMapper().readValue(jsonResponse, Stripe3dsSourceResponse.class);
```
kbottla: It may be big change now. But `new ObjectMapper()` is expensive operation as per https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClientResponse.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.pay.connector.gateway.stripe;
+
+public class StripeGatewayClientResponse {
+    private javax.ws.rs.core.Response.Status.Family statusFamily;
```
kbottla: Is there any reason full package name is used ? I think it would read nice without full package name.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClientResponse.java
```diff
@@ -0,0 +1,25 @@
...
+    private String payload;
+    private int status;
+
+    StripeGatewayClientResponse(javax.ws.rs.core.Response response) {
```
kbottla: Same as above for full package name

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClientResponse.java
```diff
@@ -0,0 +1,25 @@
...
+        this.payload = response.readEntity(String.class);
+        this.statusFamily = response.getStatusInfo().getFamily();
+        this.status = response.getStatus();
+    }
+
+    javax.ws.rs.core.Response.Status.Family getFamily() {
```
kbottla: Same as above

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -364,4 +355,13 @@ private String tokenPayload(CardAuthorisationGatewayRequest request) {
 
         return URLEncodedUtils.format(params, UTF_8);
     }
+
+    public static <T> T getObjectFromJson(String jsonResponse, Class<T> targetType) {
```
kbottla: If we want to avoid `new ObjectMapper`, This could be in another object (injected into PaymentProvider and Handlers) with ObjectMapper injected.

Common catch block can also be in one place.

> catch (IOException e) {
             logger.info("There was an exception parsing the payload [{}] into an [{}]", jsonResponse, targetType);
             throw new WebApplicationException(format("There was an exception parsing the payload [%s] into an [%s]", jsonResponse, targetType));
         }

kbottla (changes requested):
Looks good overall. Some minor comments and a comment about ObjectMapper instantiation.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClientResponse.java
georgeracu: A leftover from when the `StripeGatewayClientResponse` was called `Response` and this line required to fully qualify the name. Will fix it :)

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClientResponse.java
georgeracu: 👍

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeGatewayClientResponse.java
georgeracu: 👍

kbottla (dismissed):
Thanks for the changes. Looks like `getObjectFromJson` method doesn't belong to StripeGatewayClient, but we can look at it separately.

My initial thought is to have a class like `JsonObjectMapper` with ObjectMapper injected
```
    public JsonObjectMapper(ObjectMapper objectMapper) {
    }

   ..... getObjectFromJson method ...
```

and inject the above into gateway handlers


```
    public StripePaymentProvider(StripeGatewayClient stripeGatewayClient,
                                 ConnectorConfiguration configuration,
                                 JsonObjectMapper jsonObjectMapper
    )
```

kbottla (approved):
Looks good 👍 🏅

PR #927, opened 2018-12-27
PP-4438 enable hanging tests

georgeracu:
## WHAT
- After fixing a problem with the `StripeGatewayClient` where the client
was hanging, this tests can now be run using `DropwizardJUnitRunner`.
This is at the same time a validation that the fix to the client works
as it is that the Stripe cancelation works

To be merged after PP-4550 has been merged


kbottla (approved):
LGTM!!

PR #933, opened 2019-01-02
bau: Remove NotifyClientFactoryProvider

oswaldquek:
Passing in the NotifyClientFactoryProvider to UserNotificationService just to
get the NotifyClientFactory seems convoluted. Why not `@Provides` a
NotifyClientFactory instead?

By doing this I hope to be able to write a proper integration test for PP-4553
as this commit will make it easier to override the provision of
NotifyClientFactory in the ConnectorModule by extending the ConnectorModule.
Unfortunately we aren't able to write an integration test to test a call to
`/v1/api/notifications/epdq` actually sends a refund as the notify client
requires https and using wiremock with this is too complicated.

It would help greatly if we can mock out the NotifyClientFactory and that is
what this commit will help us to do.

@oswaldquek


georgeracu (approved):
LGTM
Using `@Provides` instead of a provider interface basically has the same functionality.

PR #934, opened 2019-01-02
PP-4570 Add 'requires_3ds' to create gateway account API

kbottla:
## WHAT

Added new field `requires_3ds` to create new Gateway Account endpoint (HTTP method POST, URL /v1/api/accounts). This will allow us to set 3DS to enabled/disabled at the time of creating gateway account and as per service requirement.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResponse.java
@@ -25,10 +25,13 @@ public class GatewayAccountResponse {
 
     @JsonProperty("gateway_account_id")
     private final String gatewayAccountId;
-    
+
+    @JsonProperty("requires_3ds")
+    private final boolean requires3ds;
+
     @JsonProperty("links")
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResponse.java
```diff
@@ -70,6 +74,10 @@ public URI getLocation() {
         return location;
     }
 
+    public boolean isRequires3ds() {
```
georgeracu: Try to name this getter consistently across objects. In GatewayAccountRequest is called `getRequires3ds()`

georgeracu (dismissed):
Overall looks good, just a small comment on being consistent in naming.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResponse.java
kbottla: 👍 Updated to `getRequires3ds`

georgeracu (approved):
👍

PR #935, opened 2019-01-02
Pp 4553 integration test for refund email notifications

oswaldquek:
SendRefundEmailTest tests that email notifications are sent for refunds. One
bad thing about this PR is that ConnectorApp and ConnectorModule have new
protected methods (`getModule` and `getNotifyClientFactory`) respectively. This
is code that enables us to mock out the NotifyClientFactory in the
SendRefundEmailTest. Unfortunately there is no nice way round doing this (it's
not Spring :( ) but it has been deemed essential to have an integration for
this functionality (mocking at unit test level is not enough). The other option
is to spin up a WireMock email notification server but unfortunately the notify
client needs to use https and enabling https in a wiremock and integration test
scenario is hugely not trivial.

@oswaldquek

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceITest.java
@@ -1,23 +1,15 @@
 package uk.gov.pay.connector.it.resources.epdq;
 
-import com.google.common.collect.Lists;
 import com.jayway.restassured.response.ValidatableResponse;
 import junitparams.Parameters;
-import org.apache.http.NameValuePair;
-import org.apache.http.client.utils.URLEncodedUtils;
-import org.apache.http.message.BasicNameValuePair;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import uk.gov.pay.connector.app.ConnectorApp;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
-import uk.gov.pay.connector.gateway.epdq.EpdqSha512SignatureGenerator;
 import uk.gov.pay.connector.it.base.ChargingITestBase;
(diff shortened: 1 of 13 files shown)


src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceITest.java
```diff
@@ -109,24 +105,18 @@ public void shouldHandleARefundNotification() throws Exception {
         String refundExternalId = "999999";
         int refundAmount = 1000;
 
-        String externalChargeId = createNewChargeWithRefund(transactionId, refundExternalId, payIdSub, refundAmount);
-
+        ChargeUtils.ExternalChargeId externalChargeId = createNewChargeWithAccountId(CAPTURED, transactionId, accountId, databaseTestHelper);
+        createNewRefund(REFUND_SUBMITTED, externalChargeId.chargeId, refundExternalId, transactionId + "/" + payIdSub, 1000, databaseTestHelper);
+
```
kbottla, line 110: Could use `refundAmount` variable instead of literal.

src/test/java/uk/gov/pay/connector/junit/DropwizardConfig.java
```diff
@@ -6,6 +6,8 @@
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.List;
+import java.util.Map;
```
kbottla, line 10: Imports unused

kbottla (approved):
Nice work!! 👍🏅 couple of trivial comments, but not too bothered.

I am fine with protected methods in ConnectorModule as they are not meddling with business logic  and it is much better solution than https://github.com/alphagov/pay-connector/pull/924.

oswaldquek:
@kbottla thanks. i'll remove the unused configs in another PR (you'll see why later).

PR #943, opened 2019-01-04
PP-4584: Remove proxy environment variables

rhowe-gds:
Let's just delegate the setting of system properties to the JAVA_OPTS
environment variable.


46bit (approved):
LGTM.

PR #944, opened 2019-01-04
[firebreak] bau: Kill CardExecutorService, use single ExecutorService across connector

oswaldquek:
Experimenting the use of CompletableFuture as a replacement for submitting
async tasks. If the performance tests pass we can be confident this is safe to
deploy.

I did some research on CompletableFuture/ForkJoinPool (see Questions section) below but in the end decided to go with using an ExecutorService as before. This is not only less risky, but the generally correct usage. In general ForkJoin allows one to easily execute divide and conquer jobs, but the ExecutorService is usually used to process many independent requests concurrently, which is our use case. 

There are two changes with the use of ExecutorService:
- It is now a single instance which is used for authorising charges and sending emails
- It is passed as a parameter to the CompetableFuture

This PR also gets rid of the `Pair<ExecutionStatus, T>` idiom used previously by the now deleted CardAuthoriseBaseService.

@oswaldquek

### What's the benefit?
By default using `CompletableFuture.supplyAsync` uses the ForkJoinPool (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html). From it's javadoc: `Especially when setting asyncMode to true in constructors, ForkJoinPools may also be appropriate for use with event-style tasks that are never joined.` We use event-style tasks that are never joined so this fits our use case.

### What is a ForkJoinPool?
From the javadoc: `A ForkJoinPool differs from other kinds of ExecutorService mainly by virtue of employing work-stealing: all threads in the pool attempt to find and execute tasks submitted to the pool and/or created by other active tasks (eventually blocking waiting for work if none exist). This enables efficient processing when most tasks spawn other subtasks (as do most ForkJoinTasks), as well as when many small tasks are submitted to the pool from external clients.`

### What is going to happen to threads in execution when connector shuts down?
We can add a shutdown hook using `Runtime.getRuntime().addShutdownHook` and call the shutdown method (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#shutdown--) and awaitTermination method (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ForkJoinPool.html#awaitTermination-long-java.util.concurrent.TimeUnit-) like we used to do in the deleted `CardExecutorService` class.


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthorisationExecutor.java
```diff
@@ -13,44 +11,65 @@
...
+    private static final int QUEUE_WAIT_WARN_THRESHOLD_MILLIS = 10000;
+    
+    private final Logger logger = LoggerFactory.getLogger(getClass());
     private final MetricRegistry metricRegistry;
+    private final XrayUtils xrayUtils;
 
     @Inject
-    public CardAuthoriseBaseService(CardExecutorService cardExecutorService, Environment environment) {
-        this.cardExecutorService = cardExecutorService;
+    public CardAuthorisationExecutor(Environment environment, XrayUtils xrayUtils) {
```
kbottla: Just out of curiosity, what is going to happen when connector shutdowns ?
Does it wait for the threads to complete execution ? or does it terminate threads immediately ?

`CardExecutorService.java` got shutdown hook which waits for all thread to complete execution. Not sure if same behaviour is expected here

oswaldquek: We can do a shutdown hook for the default ForkJoinPool used by the CompletableFuture but I've decided to revert to using an ExecutorService as before now.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -78,6 +83,14 @@ private JpaPersistModule jpaModule(ConnectorConfiguration configuration) {
 
         return jpaModule;
     }
+    
+    @Provides
+    @Singleton
+    public ExecutorService executorService(ConnectorConfiguration configuration) {
```
kbottla, line 89: Having a single executor service means, operations ( authorisation, send email notifications) share same thread pool. Will it affect performance as no. of threads available (for authorisation - under high load) may be less than they use to be ?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthorisationExecutor.java
```diff
@@ -0,0 +1,133 @@
...
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_UNEXPECTED_ERROR;
+
+/**
+ * CardAuthorisationExecutor executes tasks passed to it in a separate thread. The point of running tasks in a separate thread 	
+ * is that it can keep running after the originating thread has returned to the user. That is the purpose of the 	
+ * .get(config.getTimeoutInSeconds(), TimeUnit.SECONDS). If you look how that is used in the authorise service, it 	
+ * catches the timeout exception and returns to frontend as 'in progress'. Frontend then polls connector until the 	
+ * charge is authorised (by the CES thread), and continues on its merry way.	
+ */
+public class CardAuthorisationExecutor {
```
kbottla, line 41: I think this class is responsible for multiple things.
- Executing task as well as parsing response ( which is sent from this class and is passed back to the same class).
Can the parsing be delegated to utility class?

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
kbottla, line 89: We need to have shutdown hooks (similar to CardExecutorService) so terminating the app will wait for all threads to existing operations. Without this threads will be terminated leaving any existing operations in inconsitent state.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthorisationExecutor.java
```diff
@@ -0,0 +1,133 @@
...
+                                     ExecutorService executorService) {
+        this.metricRegistry = environment.metrics();
+        this.xrayUtils = xrayUtils;
+        this.executorService = executorService;
+        this.config = configuration.getExecutorServiceConfig();
+    }
+    
+    public <T> T executeAuthorise(String chargeId, Supplier<T> f) {
+        final long startTime = System.currentTimeMillis();
+        xrayUtils.beginSegment();
```
kbottla, line 80: segment should be probably be inside async function supplyAsync() so that actual time took to authorise charges can be recorded.
With current logic, `endSegment()` is recorded for time outs too, which means max. time recorded will be `config.getTimeoutInSeconds()`

src/test/java/uk/gov/pay/connector/applepay/AppleAuthoriseServiceTest.java
```diff
@@ -215,19 +207,6 @@ public void doAuthorise_shouldNotStoreProviderSessionId_whenAuthorisationError()
         assertThat(charge.getProviderSessionId(), is(nullValue()));
     }
 
-    @Test
-    public void doAuthorise_shouldThrowAnOperationAlreadyInProgressRuntimeException_whenTimeout() {
-        when(mockExecutorService.execute(any())).thenReturn(Pair.of(IN_PROGRESS, null));
```
kbottla, line 220: We don't have `Pair.of(IN_PROGRESS)` anymore, but don't we need this test to check for `OperationAlreadyInProgressRuntimeException`  raised from CardAuthorisationExecutor or units tests specific to CardAuthorisationExecutor

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -551,28 +536,11 @@ public void doAuthorise_shouldNotProviderSessionId_whenAuthorisationError() {
         assertThat(charge.getProviderSessionId(), is(nullValue()));
     }
 
-    @Test
-    public void doAuthorise_shouldThrowAnOperationAlreadyInProgressRuntimeException_whenTimeout() {
```
kbottla, line 555: We don't have `Pair.of(IN_PROGRESS)` anymore, but don't we need this test to check for `OperationAlreadyInProgressRuntimeException`  raised from CardAuthorisationExecutor or units tests specific to CardAuthorisationExecutor

src/test/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthServiceTest.java
```diff
@@ -182,27 +170,12 @@ public void shouldRespondAuthorisationError() {
         assertThat(response.isSuccessful(), is(false));
     }
 
-    @Test
-    public void authoriseShouldThrowAnOperationAlreadyInProgressRuntimeExceptionWhenTimeout() {
```
kbottla, line 186: We don't have `Pair.of(IN_PROGRESS)` anymore, but don't we need this test to check for `OperationAlreadyInProgressRuntimeException`  raised from CardAuthorisationExecutor or units tests specific to CardAuthorisationExecutor

kbottla (commented):
It is nice attempt 👍 . I have added few comments pertaining to design & performance.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
oswaldquek: oops i forgot to put that in, will do

oswaldquek: i don't know is the short answer. i need to do a performance test run.

oswaldquek:
thanks for taking the time to review this @kbottla. my bad but i think there won't be any time for me to work on this further. the performance environment is kind of bust as well. sorry! i should've tried testing it first before getting you to review it.

PR #945, opened 2019-01-04
PP-4570 Enable 3DS for Stripe gateway accounts

kbottla:
## WHAT

- Enables 3DSecure for all Stripe gateway accounts

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -957,4 +957,10 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="Enable 3DS for all Stripe gateway accounts" author="">
+        <sql>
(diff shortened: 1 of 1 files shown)


oswaldquek:
should there be java code to set `requires_3ds = true` as well for all future new Stripe gateway accounts?

kbottla:
There is code already for this. requires_3ds is configurable through resource endpoint /v1/api/accounts - POST operation. Field can be set to true/false depending on the gateway account need ( currently defaults to 'true when we create Stripe gateway accounts using toolbox). I am going to close this PR as there are no real accounts setup yet and can be a one-off scripts run in test.

kbottla:
Closing as there are no live accounts to run migrations.

PR #947, opened 2019-01-07
PP-4270 Multithreaded capture process

kbottla:
### `Background`

The capture process (captures charges asynchronously in the background) in connector is a single threaded job that takes a number of charges (based on batch size) and captures them one by one. Potential issues with this approach:

1. Multiple connector nodes could potentially end up with overlap of charges being captured.
2. The average response time (live) to capture charges is ~130ms. A batch size of 50 will take ~5 seconds to complete capture. So a single connector capture job scheduled every 10 seconds can only capture ~17k charges/day. Given a high volume of transactions (~1 million/day), capture processes can take weeks to capture charges from a single day and may end up with expiring authorisations.
3. Increasing number of connector nodes doesn't necessarily multiply the number of charges that can be captured as charges can overlap with other connector jobs

### `Solution`

Proposed solution is to introduce a queue (within connector node) and run multi-threaded capture process so that charges can be captured in parallel and quickly. For a batch size of 50 charges and with 10 capture threads we can clear the batch with in ~500ms. It can also likely to mitigate the issue of overlapping charges between connector nodes.

### `Changes`

- CaptureProcessScheduler now schedules multiple threads to capture charges
- Scheduler first loads shared queue (with charges awaiting capture)
- Each thread reads (removes charge from head of the queue - if queue is not empty) a charge from queue to process it further.
        - `ArrayBlockingQueue` is used as a queue which seemed most suitable to use as it is a fixed size queue that can hold charges equivalent to batch size, thread safe and size() method is constant-time operation. Although it is blocking queue, operations used (poll(), isEmpty()) are non blocking. So a thread can exit from operation without being blocked.

**Alternative considered:** 
Considered introducing blocking threads which wait (indefinitely) for charges to be available in queue. But due to the blocking nature all capture related unit tests will become invalid. We need complex test setup to terminate threads / add a mechanism to timeout tests to be able to use queue blocking methods (queue.take()).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
@@ -13,7 +13,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
 public class CaptureProcessScheduler implements Managed {
-    final Logger logger = LoggerFactory.getLogger(CaptureProcessScheduler.class);
+    private final Logger logger = LoggerFactory.getLogger(CaptureProcessScheduler.class);
 
     static final String CAPTURE_PROCESS_SCHEDULER_NAME = "capture-process";
     static final int SCHEDULER_THREADS = 1;
@@ -25,13 +25,14 @@ public class CaptureProcessScheduler implements Managed {
     private long initialDelayInSeconds = INITIAL_DELAY_IN_SECONDS;
     private long randomIntervalMinimumInSeconds = RANDOM_INTERVAL_MINIMUM_IN_SECONDS;
     private long randomIntervalMaximumInSeconds = RANDOM_INTERVAL_MAXIMUM_IN_SECONDS;
+    private int schedulerThreads = SCHEDULER_THREADS;
 
     private final CardCaptureProcess cardCaptureProcess;
     private final ScheduledExecutorService scheduledExecutorService;
     private final XrayUtils xrayUtils;
 
-    public CaptureProcessScheduler(ConnectorConfiguration configuration, 
-                                   Environment environment, 
+    public CaptureProcessScheduler(ConnectorConfiguration configuration,
+                                   Environment environment,
                                    CardCaptureProcess cardCaptureProcess,
                                    XrayUtils xrayUtils) {
         this.cardCaptureProcess = cardCaptureProcess;
@@ -42,12 +43,13 @@ public class CaptureProcessScheduler implements Managed {
             initialDelayInSeconds = captureProcessConfig.getSchedulerInitialDelayInSeconds();
             randomIntervalMinimumInSeconds = captureProcessConfig.getSchedulerRandomIntervalMinimumInSeconds();
             randomIntervalMaximumInSeconds = captureProcessConfig.getSchedulerRandomIntervalMaximumInSeconds();
+            schedulerThreads = captureProcessConfig.getSchedulerThreads();
         }
 
         scheduledExecutorService = environment
                 .lifecycle()
                 .scheduledExecutorService(CAPTURE_PROCESS_SCHEDULER_NAME)
-                .threads(SCHEDULER_THREADS)
+                .threads(schedulerThreads + 1)
                 .build();
     }
 
@@ -56,15 +58,26 @@ public class CaptureProcessScheduler implements Managed {
         logger.info("Scheduling CardCaptureProcess to run every {} seconds (will start in {} seconds)", interval, initialDelayInSeconds);
 
         scheduledExecutorService.scheduleAtFixedRate(() -> {
+            for (int threadNumber = 1; threadNumber <= schedulerThreads; threadNumber++) {
+                final int finalThreadNumber = threadNumber;
+
+                cardCaptureProcess.loadCaptureQueue();
+                scheduleCaptureThreads(finalThreadNumber);
+            }
+        }, initialDelayInSeconds, interval, TimeUnit.SECONDS);
+    }
+
+    private void scheduleCaptureThreads(int finalThreadNumber) {
+        scheduledExecutorService.schedule(() -> {
             try {
                 xrayUtils.beginSegment();
-                cardCaptureProcess.runCapture();
+                cardCaptureProcess.runCapture(finalThreadNumber);
             } catch (Exception e) {
                 logger.error("Unexpected error running capture operations", e);
             } finally {
                 xrayUtils.endSegment();
             }
-        }, initialDelayInSeconds, interval, TimeUnit.SECONDS);
+        }, 0, TimeUnit.SECONDS);
     }
 
     private long randomTimeInterval() {
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
@@ -46,10 +46,14 @@ public class CaptureProcessScheduler implements Managed {
             schedulerThreads = captureProcessConfig.getSchedulerThreads();
         }
 
+        // Total number of threads is schedulerThreads (dedicated threads that captures the charges)
+        //  + main thread that initiates loading and capturing the charges
+        int totalThreads = schedulerThreads + 1;
+        
         scheduledExecutorService = environment
                 .lifecycle()
                 .scheduledExecutorService(CAPTURE_PROCESS_SCHEDULER_NAME)
-                .threads(schedulerThreads + 1)
+                .threads(totalThreads)  
                 .build();
     }
 
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
@@ -58,26 +58,25 @@ public class CaptureProcessScheduler implements Managed {
         logger.info("Scheduling CardCaptureProcess to run every {} seconds (will start in {} seconds)", interval, initialDelayInSeconds);
 
         scheduledExecutorService.scheduleAtFixedRate(() -> {
-            for (int threadNumber = 1; threadNumber <= schedulerThreads; threadNumber++) {
-                final int finalThreadNumber = threadNumber;
-
-                cardCaptureProcess.loadCaptureQueue();
-                scheduleCaptureThreads(finalThreadNumber);
-            }
+            cardCaptureProcess.loadCaptureQueue();
+            scheduleCaptureThreads();
         }, initialDelayInSeconds, interval, TimeUnit.SECONDS);
     }
 
-    private void scheduleCaptureThreads(int finalThreadNumber) {
-        scheduledExecutorService.schedule(() -> {
-            try {
-                xrayUtils.beginSegment();
-                cardCaptureProcess.runCapture(finalThreadNumber);
-            } catch (Exception e) {
-                logger.error("Unexpected error running capture operations", e);
-            } finally {
-                xrayUtils.endSegment();
-            }
-        }, 0, TimeUnit.SECONDS);
+    private void scheduleCaptureThreads() {
+        for (int threadNumber = 1; threadNumber <= schedulerThreads; threadNumber++) {
+            final int finalThreadNumber = threadNumber;
+            scheduledExecutorService.schedule(() -> {
+                try {
+                    xrayUtils.beginSegment();
+                    cardCaptureProcess.runCapture(finalThreadNumber);
+                } catch (Exception e) {
+                    logger.error("Unexpected error running capture operations", e);
+                } finally {
+                    xrayUtils.endSegment();
+                }
+            }, 0, TimeUnit.SECONDS);
+        }
     }
 
     private long randomTimeInterval() {
--- a/src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
@@ -9,23 +9,26 @@ import com.github.tomakehurst.wiremock.junit.WireMockRule;
 import com.google.common.collect.ImmutableMap;
 import org.hamcrest.Matchers;
 import org.junit.Before;
-import org.junit.Rule;
+import org.junit.ClassRule;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.ConnectorApp;
 import uk.gov.pay.connector.it.dao.DatabaseFixtures;
 import uk.gov.pay.connector.it.util.ChargeUtils;
-import uk.gov.pay.connector.paymentprocessor.service.CaptureProcessScheduler;
+import uk.gov.pay.connector.junit.ConfigOverride;
+import uk.gov.pay.connector.junit.DropwizardConfig;
+import uk.gov.pay.connector.junit.DropwizardJUnitRunner;
+import uk.gov.pay.connector.junit.DropwizardTestContext;
+import uk.gov.pay.connector.junit.TestContext;
 import uk.gov.pay.connector.paymentprocessor.service.CardCaptureProcess;
-import uk.gov.pay.connector.rules.GuiceAppWithPostgresRule;
 import uk.gov.pay.connector.rules.WorldpayMockClient;
-import uk.gov.pay.connector.util.PortFactory;
-import uk.gov.pay.connector.util.XrayUtils;
+import uk.gov.pay.connector.util.DatabaseTestHelper;
 
 import java.util.List;
 import java.util.Map;
 
-import static io.dropwizard.testing.ConfigOverride.config;
 import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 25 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
```diff
@@ -42,12 +43,13 @@ public CaptureProcessScheduler(ConnectorConfiguration configuration,
...
             randomIntervalMinimumInSeconds = captureProcessConfig.getSchedulerRandomIntervalMinimumInSeconds();
             randomIntervalMaximumInSeconds = captureProcessConfig.getSchedulerRandomIntervalMaximumInSeconds();
+            schedulerThreads = captureProcessConfig.getSchedulerThreads();
         }
 
         scheduledExecutorService = environment
                 .lifecycle()
                 .scheduledExecutorService(CAPTURE_PROCESS_SCHEDULER_NAME)
-                .threads(SCHEDULER_THREADS)
+                .threads(schedulerThreads + 1)
```
oswaldquek: why is 1 added to schedulerThreads here?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
```diff
@@ -56,15 +58,26 @@ public void start() {
         logger.info("Scheduling CardCaptureProcess to run every {} seconds (will start in {} seconds)", interval, initialDelayInSeconds);
 
         scheduledExecutorService.scheduleAtFixedRate(() -> {
+            for (int threadNumber = 1; threadNumber <= schedulerThreads; threadNumber++) {
+                final int finalThreadNumber = threadNumber;
+
+                cardCaptureProcess.loadCaptureQueue();
```
oswaldquek: we only need to load the capture queue once. could this line go before the for loop?

oswaldquek (commented):
I have two minor comments. 

I think this solution will work. However, it is a little complicated. In my head it would be easier if the only thing that is scheduled is the loading of the capture queue. Then in the background we have a number of workers (the `captureProcessConfig.getSchedulerThreads()`) that are constantly polling this capture queue and capturing charges when the queue is not empty, rather than scheduling when they capture charges. Maybe some input from other team members would be useful to get some consensus.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
kbottla: 👍 good spot. Moved it out of loop.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
kbottla: It is for the main thread (loading + capture) and capture only threads (schedulerThreads)

oswaldquek: hmm then why was the adding of the 1 not there before? will things not work if the 1 is not added? again I approach reading code as a new developer. i'm loathe to this kind of thing unless there's a clear comment.

oswaldquek (changes requested):
ok. given we are likely going to use a real queue at some point, it won't make much difference between my solution and the current implementation, so i'm happy with that.

only thing is i'd like to see an integration test with at least 2 `schedulerThreads` where we it can be verified a charge's status is updated via the capturing process.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
kbottla: We only have one thread before which calls runCapture (which does both loading + capturing charges). 
Things will work without 1 but it will be one less thread available from thread pool for capture threads.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,120 @@
...
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_APPROVED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_SUBMITTED;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_MERCHANT_ID;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_PASSWORD;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_SHA_IN_PASSPHRASE;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_USERNAME;
+
+@RunWith(MockitoJUnitRunner.class)
```
oswaldquek: don't actually need this as there are no `@Mock` annotations on fields

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,120 @@
...
+    protected int port = PortFactory.findFreePort();
+
+    private Appender<ILoggingEvent> mockAppender = mock(Appender.class);
+    private ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
+
+    @Rule
+    public WireMockRule wireMockRule = new WireMockRule(port);
+
+    @Rule
+    public GuiceAppWithPostgresRule app = new GuiceAppWithPostgresRule(
```
oswaldquek: i would recommend using `@RunWith(DropwizardJUnitRunner.class)` to run this test. one of the benefits is you don't need to find a port for the wireMockRule as you get that for free. you can simply do this:
https://github.com/alphagov/pay-connector/blob/56c8ab1670b342db58d0c697ec1000f51ed9cd20/src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java#L85

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,120 @@
...
+            config("worldpay.urls.test", "http://localhost:" + port + "/jsp/merchant/xml/paymentService.jsp"),
+            config("captureProcessConfig.maximumRetries", Integer.toString(CAPTURE_MAX_RETRIES)),
+            config("captureProcessConfig.schedulerThreads", "2"),
+            config("captureProcessConfig.schedulerInitialDelayInSeconds", "0"),
+            config("captureProcessConfig.schedulerRandomIntervalMinimumInSeconds", "1"),
+            config("captureProcessConfig.schedulerRandomIntervalMaximumInSeconds", "1")
```
oswaldquek: if you use `@RunWith(DropwizardJUnitRunner.class)` you can override the configs like in the SendRefundEmailTest:
https://github.com/alphagov/pay-connector/blob/56c8ab1670b342db58d0c697ec1000f51ed9cd20/src/test/java/uk/gov/pay/connector/it/SendRefundEmailTest.java#L46

you could then also ignore the `config("worldpay.urls.test", "http://localhost:" + port + "/jsp/merchant/xml/paymentService.jsp")` part as you get that for free.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,120 @@
...
+    );
+
+    private DatabaseFixtures.TestCharge createTestCharge(String paymentProvider, ChargeStatus chargeStatus) {
```
oswaldquek: it seems this method was copied from `CardCaptureProcessBaseITest`. could we avoid code duplication and put this as a static method in `ChargeUtils.java` please. it won't hurt to do some easy refactoring here.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,120 @@
...
+    public void schedulerShouldStartMultipleThreadsAsPerConfig_AndCaptureCharge() throws InterruptedException {
+
+        DatabaseFixtures.TestCharge testCharge = createTestCharge(PAYMENT_PROVIDER, CAPTURE_APPROVED);
+        new WorldpayMockClient().mockCaptureSuccess();
+
+        CaptureProcessScheduler captureProcessScheduler = new CaptureProcessScheduler(app.getConf(),
+                app.getEnvironment(), app.getInstanceFromGuiceContainer(CardCaptureProcess.class),
+                app.getInstanceFromGuiceContainer(XrayUtils.class));
+
+        captureProcessScheduler.start();
```
oswaldquek: since this is an integration test, we shouldn't have to do call `captureProcessScheduler.start()` here as it should be automatically started when the dropwizard app starts up.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        Thread.sleep(1000L); // Mininum configurable interval for capture scheduler
+
+        // Expected below : For 2 scheduler threads with a charge in CAPTURE_APPROVED state
+        // Thread 1: Capturing: 1 of 1 charges ,  Capturing [1 of 1] [chargeId=...] ,  Capture complete ...
+        // Thread 2: Capture complete ....
+        verify(mockAppender, times(4)).doAppend(loggingEventArgumentCaptor.capture());
```
oswaldquek, line 99: personally i'm happy with the assertion that the charge's status is CAPTURE_SUBMITTED. but if you really want to test the mockAppender, is it possible to verify it got called with the actual logs, e.g. "Capturing: 1 of 1 charges", "Capturing [1 of 1] [chargeId=...]" etc?

oswaldquek (changes requested):

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: 👍

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: My aim is to check if multiple threads are working on capture process (one way which I have done is to check the logging activity by threads) . I don't think we need to check exact logging as the test is if multiple threads are being created by scheduler and capture process is invoked.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: I would prefer lightweight version for testing internal services (which is the initial aim of GuiceAppWithPostgresRule) as opposed to spinning entire app to test this. We will have control over the methods to invoke and check at the same time.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: 👍 Extracted creatingCharge to ChargeUtils. But not done for `CardCaptureProcessBaseITest`.  It would be good to do refactoring in a separate PR.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: We need to call `start()` when using GuiceAppWithPostgresRule

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: okay I have updated test slightly to check partial strings for threads completing capture.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
oswaldquek: It doesn't matter with the `@RunWith(DropwizardJUnitRunner.class)` because the database would only be spun up once for the entire run of the build so you're not really saving anything here. The benefits of using DropwizardJUnitRunner - less code because you don't need to `findFreePort` or configure the `worldpay.urls.test` - are clear.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
oswaldquek: ok. 👍

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: It won't make difference when all tests are run, but difference can be seen when running the test isolated.
`@RunWith(DropwizardJUnitRunner.class)` - spins up jetty server which increases the time to execute test and feedback cycle for developers

Also DropwizardJUnitRunner doesn't seem to be starting capture thread(s) (even on master branch which is single threaded). Not sure at the moment if this is the behaviour expected  and the reason `GuiceAppWithPostgresRule` was introduced

oswaldquek: Hmm well if the capture threads aren't started automatically already suggests there's a problem. The point of an integration test is that one avoids calling things manually, e.g. `captureProcessScheduler.start()` as this should be done automatically when the app is spun up. This is a reason why using `GuiceAppWithPostgresRule` is bad in this case - it's not testing things properly.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
oswaldquek: It wouldn't hurt to do the refactoring for `CardCaptureProcessBaseITest ` in this PR really. It's not that much code.

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
kbottla: It may be a problem with DropwizardJUnitRunner as it doesn't even work with master branch. 
`captureProcessScheduler.start()` is working fine but I think it has problem with getting threads from ThreadPool.

kbottla: okay so the problem was not with acquiring threads, but threads being killed due to exception (DB migrations were incomplete and capture threads being scheduled during jetty startup). Still not convinced to use DropwizardJUnitRunner as we has to wait for server to start (when test is run in isolation). 

Both (DropwizardJUnitRunner, GuiceAppWithPostgresRule - no need to use start for GuiceAppWithPostgresRule but can it without long wait ) need to wait for DB migrations to complete, so the solution is similar except that for DropwizardJUnitRunner need to wait a bit longer for capture process to kick in and could cause flakiness if server startup and migrations take longer than the wait

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,116 @@
...
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_SHA_IN_PASSPHRASE;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_USERNAME;
+import static uk.gov.pay.connector.junit.DropwizardJUnitRunner.WIREMOCK_PORT;
+
+@RunWith(DropwizardJUnitRunner.class)
+@DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml",
+        configOverrides = {
+                @ConfigOverride(key = "logging.level", value = "INFO"),
+                @ConfigOverride(key = "captureProcessConfig.schedulerThreads", value = "2"),
+                @ConfigOverride(key = "captureProcessConfig.schedulerInitialDelayInSeconds", value = "10"), // delay so DB migrations can be complete before capture process
```
oswaldquek, line 50: i guess waiting for the DB to start up is unavoidable :(

src/test/java/uk/gov/pay/connector/it/scheduler/CaptureProcessSchedulerITest.java
```diff
@@ -0,0 +1,116 @@
...
+                @ConfigOverride(key = "captureProcessConfig.schedulerRandomIntervalMinimumInSeconds", value = "5"),
```
oswaldquek, line 51: is there any reason why you chose 5? if i changed `schedulerRandomIntervalMinimumInSeconds` and `schedulerRandomIntervalMaximumInSeconds` to 1 the test fails :(

oswaldquek (approved):

PR #950, opened 2019-01-08
Revert "PP-4584: Manage HTTP proxies via env. vars"

rhowe-gds:
This partially reverts commit e2385cf9abdf45bef8ae7642b65316da8cf2d8be.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
@@ -14,6 +14,7 @@ import org.apache.http.impl.conn.ManagedHttpClientConnectionFactory;
 import org.apache.http.impl.conn.SystemDefaultDnsResolver;
 import org.glassfish.jersey.SslConfigurator;
 import org.glassfish.jersey.apache.connector.ApacheConnectorProvider;
+import org.glassfish.jersey.client.ClientProperties;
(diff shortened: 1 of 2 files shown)


danworth (dismissed):
Minor point about unnecessary optional wrapping which Intellij found.

src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
```diff
@@ -72,24 +77,16 @@ private Client createWithDropwizardClient(PaymentGatewayName gateway, Optional<G
...
-        OperationOverrides overrides = getOverridesFor(operation, gateway);
-        if (overrides != null && overrides.getReadTimeout() != null) {
-            return (int) overrides.getReadTimeout().toMilliseconds();
-        }
-        return getDefaultTimeout();
-    }
-
-    private int getDefaultTimeout() {
-        return (int) conf.getCustomJerseyClient().getReadTimeout().toMilliseconds();
+    private Optional<Duration> getReadTimeout(Optional<GatewayOperation> operation, PaymentGatewayName gateway) {
```
danworth: Won't let this hold up approval but I think the `Optional<GatewayOperation>` as the param is unnecessary since we either pass the value or null with `operation.orElse(null)`, so may as well just be of type GatewayOperation.

danworth (approved):
Looks good.

PR #952, opened 2019-01-09
Refactor ClientFactory

rhowe-gds:
## WHAT
Constructing Jersey clients for connecting to payment gateways is a complex nest of `Optional`s but without really using the features `Optional` provides.

All the checking for whether values are present that we do can be done by `Optional` itself if we use it more carefully.

Also, introduce a new `GatewayOperation`, `ALL`, to use when constructing non-operation-specific clients.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
@@ -45,22 +45,19 @@ public class ClientFactory {
     }
 
     public Client createWithDropwizardClient(PaymentGatewayName gateway, GatewayOperation operation, MetricRegistry metricRegistry) {
-        return createWithDropwizardClient(gateway, Optional.of(operation), metricRegistry);
+        return createWithDropwizardClient(gateway, getReadTimeout(operation, gateway), operation.getConfigKey(), metricRegistry);
     }
 
     public Client createWithDropwizardClient(PaymentGatewayName gateway, MetricRegistry metricRegistry) {
-        return createWithDropwizardClient(gateway, Optional.empty(), metricRegistry);
+        return createWithDropwizardClient(gateway, conf.getCustomJerseyClient().getReadTimeout(), "all", metricRegistry);
     }
 
-    private Client createWithDropwizardClient(PaymentGatewayName gateway, Optional<GatewayOperation> operation, MetricRegistry metricRegistry) {
+    private Client createWithDropwizardClient(PaymentGatewayName gateway, Duration readTimeout, String metricName, MetricRegistry metricRegistry) {
         JerseyClientBuilder defaultClientBuilder = new JerseyClientBuilder(environment)
                 .using(new ApacheConnectorProvider())
                 .using(conf.getClientConfiguration())
-                .withProperty(READ_TIMEOUT, (int) getReadTimeout(operation.orElse(null), gateway).toMilliseconds())
-                .withProperty(CONNECTION_MANAGER, createConnectionManager(
-                        gateway.getName(),
-                        operation.map(GatewayOperation::getConfigKey).orElse("all"),
-                        metricRegistry));
+                .withProperty(READ_TIMEOUT, (int) readTimeout.toMilliseconds())
+                .withProperty(CONNECTION_MANAGER, createConnectionManager(gateway.getName(), metricName, metricRegistry));
 
         if (System.getProperty(PROXY_HOST_PROPERTY) != null && System.getProperty(PROXY_PORT_PROPERTY) != null) {
             defaultClientBuilder.withProperty(ClientProperties.PROXY_URI, format("http://%s:%s",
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
```diff
@@ -45,28 +45,15 @@ public ClientFactory(Environment environment, ConnectorConfiguration conf) {
...
-            defaultClientBuilder
-                    .withProperty(READ_TIMEOUT, getReadTimeoutInMillis(operation.get(), gateway))
-                    .withProperty(CONNECTION_MANAGER, createConnectionManager(gateway.getName(), operation.get().getConfigKey(), metricRegistry));
-        } else {
-            defaultClientBuilder
-                    .withProperty(READ_TIMEOUT, getDefaultTimeout())
-                    .withProperty(CONNECTION_MANAGER, createConnectionManager(gateway.getName(), "all", metricRegistry));
-        }
+                .using(conf.getClientConfiguration())
+                .withProperty(READ_TIMEOUT, (int) getReadTimeout(operation, gateway)
```
simad: 👍 I would encapsulate this `.orElse` in the `getReadTimeout` function, so that it's a bit clearer we are either returning the gateway specific timeout or a default one. That would also allow us to get rid of `Optional<Duration>` and just return a Duration
Alternatively, I would suggest to leave things as they are and name again `conf.getCustomJerseyClient().getReadTimeout()).toMilliseconds()` `defaultTimeout` or something. It makes things a bit easier to follow

src/test/java/uk/gov/pay/connector/gateway/smartpay/BaseSmartpayPaymentProviderTest.java
```diff
@@ -47,7 +48,7 @@
     public void setup() {
         GatewayClientFactory gatewayClientFactory = new GatewayClientFactory(mockClientFactory);
 
-        when(mockClientFactory.createWithDropwizardClient(eq(PaymentGatewayName.SMARTPAY), any(MetricRegistry.class)))
+        when(mockClientFactory.createWithDropwizardClient(eq(PaymentGatewayName.SMARTPAY), GatewayOperation.ALL, any(MetricRegistry.class)))
```
simad: I understand why you created this meta gateway operation. I am not sure it adds much clarity tho -we could just have a constructor which does not take operation at all as a parameter and just instantiates clients using default timeouts. This would just be used when the apps start and the gateway operation factories can use the constructor with operation to do their job.

Up to you to see if you find this clearer - It's just that I am a bit wary of adding enums which do not really map to a "real" thing and are a bit of a syntactic sugar :)

simad (dismissed):
Minor comments/suggestions. The work is great (thanks Russell! :D) but up to you to see if you want to address those here or in another PR or never :)

src/test/java/uk/gov/pay/connector/gateway/smartpay/BaseSmartpayPaymentProviderTest.java
rhowe-gds: I am in the same place you are - it's definitely syntactic sugar, but I'm not convinced that it's particularly ugly. I have no particularly strong feelings about it, which annoys me.

src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
rhowe-gds: Yes, this is much cleaner. Done.

src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
```diff
@@ -82,24 +73,17 @@ private Client createWithDropwizardClient(PaymentGatewayName gateway, Optional<G
...
-        }
-        return getDefaultTimeout();
-    }
-
-    private int getDefaultTimeout() {
-        return (int) conf.getCustomJerseyClient().getReadTimeout().toMilliseconds();
+    private Duration getReadTimeout(GatewayOperation operation, PaymentGatewayName gateway) {
+        return getOverridesFor(operation, gateway)
+                .map(OperationOverrides::getReadTimeout)
+                .orElse(conf.getCustomJerseyClient().getReadTimeout());
```
simad, line 79: just as a note, `.orElse(conf.getCustomJerseyClient().getReadTimeout());` will side effect and be evaluated even if `OperationOverrides::getReadTimeout` returns a value

so I think it’s fine if the config is correct (because if it weren’t, probably the app would’ve :boom: when starting up) but if there’s a chance any of those functions will be null, it might blow up

simad (approved):
LGTM!

src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
alexbishop1: Yeah, use `orElseGet(() -> conf.getCustomJerseyClient().getReadTimeout())` to avoid the side-effect (which works because the argument is a lambda expression that gets invoked if necessary, rather than an immediately evaluated expression).

PR #955, opened 2019-01-09
PP-4602 remove logging of errors from accepted and bad request exceptions

simad:
- There are some cases we don't want to log those, because the behaviour is expected.
  Unfortunately in connector we use ResponseUtils and WebApplicationExceptions instead
  of letting dropwizard help us - this will be hopefully refactored
- Making sure we log errors where they happen rather than letting the exception do it
  automagically


src/main/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResource.java
```diff
@@ -120,6 +121,7 @@ public Response updateChargeStatus(
     }
 
     private Response getInvalidStatusResponse(String chargeId, ChargeStatus newChargeStatus) {
+        logger.warn("charge with id {} cannot be updated to the new status: {}", chargeId, newChargeStatus.getValue());
```
alexbishop1: Does this happen routinely? What’s the cause?

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -67,21 +68,24 @@ public Response authoriseCharge(@PathParam("chargeId") String chargeId, ApplePay
...
 
-    private Response handleAuthResponse(AuthoriseStatus authoriseStatus) {
+    private Response handleAuthResponse(String chargeId, AuthoriseStatus authoriseStatus) {
         if (authoriseStatus.equals(AuthoriseStatus.SUBMITTED)) {
+            logger.info("Charge {}: authorisation was deferred.", chargeId);
             return badRequestResponse("This transaction was deferred.");
         }
 
         if (isAuthorisationDeclined(authoriseStatus)) {
+            logger.info("Charge {}: authorisation was declined.", chargeId);
```
alexbishop1: We should already log a more detailed message elsewhere that says the authorisation was declined and what the status of the charge will be changed from and to as a result

src/main/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResource.java
simad: I couldn't find any occurrence of this in the last 3 days, so I guess it does not happen routinely.

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
simad: true, i see in the logs the status change. I can probably remove all of those logger.info for declined transactions - i guess we should keep the deferred one tho

DanailMinchev (approved):
as confirmed with @simad the comments are addressed and @alexbishop1 is on holiday, so I will approve

src/main/java/uk/gov/pay/connector/applepay/ApplePayService.java
```diff
@@ -62,6 +62,7 @@ private Response handleError(GatewayError error) {
             case GATEWAY_CONNECTION_SOCKET_ERROR:
                 return serviceErrorResponse(error.getMessage());
             default:
+                LOGGER.error("Charge {}: error {}", error.getMessage());
```
rhowe-gds, line 65: Two placeholders, one value?

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -156,6 +159,7 @@ private Response handleError(GatewayError error) {
             case GATEWAY_CONNECTION_SOCKET_ERROR:
                 return serviceErrorResponse(error.getMessage());
             default:
+                logger.error("Charge {}: error {}", error.getMessage());
```
rhowe-gds, line 162: Two placeholders, one value?

PR #959, opened 2019-01-09
PP-4597 Use millisecond precision timestamps in API responses

alexbishop1:
We use `DateTimeFormatter.ISO_INSTANT.withZone(UTC)` to format a lot of the timestamps we put in API responses, which produces strings like `"2015-10-21T07:28:00.123Z"`. The fractional part, if present, will have either three, six or nine digits, depending on the precision available.

When one asks Java 8 for the current time, it always returns the time with millisecond precision, even if the underlying system clock can do better. Java 9 and above don’t have this limitation
and will return higher accuracy if the system clock offers it (see https://bugs.openjdk.java.net/browse/JDK-8068730 for details). On Linux and macOS, this usually means microsecond precision.

Therefore, when `DateTimeFormatter.ISO_INSTANT.withZone(UTC)` is used to format the current time on Java 8, the resulting string has three digits after the decimal point. When it’s used on Java 9 on Linux or macOS, it will have six digits. This means that the timestamps we put into API responses — many of which are the current time or a stored past current time — may vary depending on the Java version.

Eliminate this difference by introducing a `DateTimeFormatter` that always uses millisecond precision. If this `DateTimeFormatter` is used throughout the system, then there will be no change to the timestamps in our API responses when we upgrade to Java 11.

Note that this new `DateTimeFormatter` has one slight behaviour change: if `DateTimeFormatter.ISO_INSTANT.withZone(UTC)` is asked to format a time that’s exactly on the second, it will not include a fractional component (e.g. `"2015-10-21T07:28:00Z"`). The new `DateTimeFormatter` always includes a fractional component (e.g. `"2015-10-21T07:28:00.000Z"`). This actually makes things more consistent, so we’re happy with this change.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/util/DateTimeUtilsTest.java
+++ b/src/test/java/uk/gov/pay/connector/util/DateTimeUtilsTest.java
@@ -2,7 +2,11 @@ package uk.gov.pay.connector.util;
 
 import org.junit.Test;
 
-import java.time.*;
+import java.time.LocalDate;
+import java.time.LocalTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.util.Optional;
 
 import static org.hamcrest.core.Is.is;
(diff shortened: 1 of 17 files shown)


src/test/java/uk/gov/pay/connector/util/DateTimeUtilsTest.java
```diff
@@ -2,7 +2,11 @@
 
 import org.junit.Test;
 
-import java.time.*;
```
simad, line 5: 👏

simad (approved):
I approve of this change with millisecond precision

PR #960, opened 2019-01-09
Silence liquibase noise during tests

rhowe-gds:
Downgrade the logging level of liquibase during test runs. Liquibase logs all
its SQL which clutters the build log.


rhowe-gds:
We already drop the default log level during integration tests to `WARN`, so this PR does nothing

PR #967, opened 2019-01-14
PP-4270 Log exception message for failed captures

kbottla:
## WHAT
Log exception message for failed captures


georgeracu (approved):
LGTM

PR #968, opened 2019-01-14
PP-4437 Stripe - 3DS options

kbottla:
## WHAT

Enable 3DS journey for payments (using Stripe as payment provider) whenever it is possible to do so. Currently, 3DS authentication is only performed if Stripe says it is `required` (based on 3DS Source response from Stripe). This change enables 3DS authentication when it is also `recommended` / `optional`.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeSourcesResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeSourcesResponse.java
@@ -2,9 +2,14 @@ package uk.gov.pay.connector.gateway.stripe.json;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class StripeSourcesResponse {
+
+    private List<String> threeDSecureRequiredOptions = ImmutableList.of("required", "recommended", "optional");
 
     @JsonProperty("id")
     private String id;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeSourcesResponse.java
```diff
@@ -21,7 +26,7 @@ public String getTransactionId() {
     }
 
     public boolean require3ds() {
-        return card != null && "required".equals(card.getThreeDSecure());
+        return card != null && threeDSecureRequiredOptions.contains(card.getThreeDSecure());
```
oswaldquek, line 29: don't we want this to return `true` all the time, unless the card is not null?

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeSourcesResponse.java
```diff
@@ -2,10 +2,15 @@
...
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class StripeSourcesResponse {
 
+    List<String> threeDSecureRequiredOptions = ImmutableList.of("required", "recommended", "optional");
```
oswaldquek: not sure if this is required, but if it is, it should be a private field.

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeSourcesResponse.java
kbottla: This should be `true` only when 3DS is supported (with values specified in list `threeDSecureRequiredOptions`)  and all other times this should be false.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeSourcesResponse.java
kbottla: It is required for list of options. Updated to private

oswaldquek (approved):

PR #972, opened 2019-01-17
PP-4650 Refactoring to prepare for addition of Google Pay.

danworth:
- Moved `applepayment` package into new `wallets` package.
- Rename `ApplePaymentInfo` to `PaymentInfo` and move into `wallets/common/api` package
to be shared with Google Pay.
- Change url for apple pay to `/wallets/apple-pay`.
- Changed name of `ApplePayToken` to `ApplePayAuthRequest` to make it clear
it represents a request which can be validated.

## WHAT
Some of the existing Apple Pay code can be re-used for Google Pay. This is a little refactoring to prepare for a new Google pay end-point and auth request class to following in a separate PR.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
+++ b/src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
@@ -228,8 +228,8 @@ public class ChargingITestBase {
         assertFrontendChargeStatusIs(chargeId, status);
     }
 
-    public static String authoriseChargeUrlForWallet(String chargeId) {
-        return "/v1/frontend/charges/{chargeId}/wallets".replace("{chargeId}", chargeId);
+    public static String authoriseChargeUrlForApplePay(String chargeId) {
+        return "/v1/frontend/charges/{chargeId}/wallets/apple-pay".replace("{chargeId}", chargeId);
     }
     
     public static String authoriseChargeUrlFor(String chargeId) {
--- a/src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
+++ b/src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
@@ -229,7 +229,7 @@ public class ChargingITestBase {
     }
 
     public static String authoriseChargeUrlForApplePay(String chargeId) {
-        return "/v1/frontend/charges/{chargeId}/wallets/apple-pay".replace("{chargeId}", chargeId);
+        return "/v1/frontend/charges/{chargeId}/wallets/apple".replace("{chargeId}", chargeId);
     }
     
     public static String authoriseChargeUrlFor(String chargeId) {
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
@@ -8,7 +8,7 @@ import com.google.inject.assistedinject.FactoryModuleBuilder;
 import com.google.inject.persist.jpa.JpaPersistModule;
 import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.setup.Environment;
-import uk.gov.pay.connector.applepay.ApplePayDecrypter;
+import uk.gov.pay.connector.wallets.applepay.ApplePayDecrypter;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.validator.RequestValidator;
 import uk.gov.pay.connector.gateway.GatewayClientFactory;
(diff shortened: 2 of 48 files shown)


src/main/java/uk/gov/pay/connector/wallets/applepay/WalletType.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.pay.connector.wallets.applepay;
```
georgeracu: I think this class can go in `wallets` package as it contains types of wallets, not types of applepay wallets.

src/main/java/uk/gov/pay/connector/wallets/common/api/PaymentInfo.java
```diff
@@ -1,21 +1,22 @@
-package uk.gov.pay.connector.applepay.api;
+package uk.gov.pay.connector.wallets.common.api;
 
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 import uk.gov.pay.connector.gateway.model.PayersCardType;
 
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
-public class ApplePaymentInfo {
+public class PaymentInfo {
```
georgeracu: I wonder if `WalletPaymentInfo` explains better what this class is intended to used for. We might confuse people with `PaymentInfo` which is generic and could be reused in `card` payments.

src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
```diff
@@ -228,8 +228,8 @@ protected void shouldReturnErrorForAuthorisationDetailsWithMessage(String author
         assertFrontendChargeStatusIs(chargeId, status);
     }
 
-    public static String authoriseChargeUrlForWallet(String chargeId) {
-        return "/v1/frontend/charges/{chargeId}/wallets".replace("{chargeId}", chargeId);
+    public static String authoriseChargeUrlForApplePay(String chargeId) {
+        return "/v1/frontend/charges/{chargeId}/wallets/apple-pay".replace("{chargeId}", chargeId);
```
georgeracu: It really needs the `-pay` part? We will just repeat ourselves when other wallets will be added.

src/main/java/uk/gov/pay/connector/wallets/common/api/PaymentInfo.java
```diff
@@ -1,21 +1,22 @@
-package uk.gov.pay.connector.applepay.api;
+package uk.gov.pay.connector.wallets.common.api;
```
georgeracu: I would argue on two things here. First is the use of `api` as we reuse this object to send it for gateway authorisation, so it goes a few levels down into our stack, doesn't stop at the api edge.
Second is the use of `common`. I believe that everything that is not in any of the specific wallet type packages is considered to be common. I would get rid of the `common` part from this package name.

georgeracu (changes requested):
Overall a good separation of the domain concepts with explicit names. 

The only pain points I found were related to the old problem of naming `common` or `util` things. This usually creates a generic place where we will throw everything that doesn't fit everywhere else. 
In this particular case we can get rid of the `common` package and keep only `api`. I believe classes in the parent package are implied to be `common` to the specialised types. This way `...wallets.api.WalletPaymentInfo` will tell us a few things: 
* when is used in `applepay` then we understand that this is a common class because it is in the parent package; 
* we will force people in the future to name better their common packages, because there is no `common` package. We basically avoid creating a `kitchen sink` in our code;
* we will structure `wallets` as if it wasn't a bag of utilities for the wallet types we use. It will have a clean structure explaining its functionality.

src/main/java/uk/gov/pay/connector/wallets/applepay/WalletType.java
danworth: 👍 good idea

src/main/java/uk/gov/pay/connector/wallets/common/api/PaymentInfo.java
danworth: 👍

src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
danworth: That's a good challenge, I think you're right, I'll remove it.

src/main/java/uk/gov/pay/connector/wallets/common/api/PaymentInfo.java
danworth: Ok, how about putting it into `wallets/model`?

georgeracu (approved):
LGTM 👍

PR #973, opened 2019-01-18
BAU_clean_up_raw_generic_use

danworth:
- `CardExecutorService` needn't be a generic class since the parameter is only
used within `CardExecutorService.execute()`. At present this class is being used
as a raw generic in multiple places which presents compile and IDE
warnings. Also since the type isn't being specified the result of `.execute()`
is being cast which present a `unchecked cast warning`. To simplify the class
and its use, this makes `.execute()` a generic method and removes the generic
typing at the class level.

## WHAT
Little bit of tidy up.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseBaseService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseBaseService.java
@@ -13,7 +13,6 @@ import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeEx
 import uk.gov.pay.connector.gateway.exception.GenericGatewayRuntimeException;
 import uk.gov.pay.connector.gateway.model.GatewayError;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseBaseService.java
```diff
@@ -39,7 +38,7 @@ public CardAuthoriseBaseService(CardExecutorService cardExecutorService, Environ
 
  
     public <T> T executeAuthorise(String chargeId, Supplier<T> authorisationSupplier) {
-        Pair<ExecutionStatus, T> executeResult = (Pair<ExecutionStatus, T>) cardExecutorService.execute(authorisationSupplier);
+        Pair<ExecutionStatus, T> executeResult = cardExecutorService.execute(authorisationSupplier);
```
georgeracu, line 41: Nice removal of casting :)

georgeracu (approved):
Nice improvement 👍

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseBaseService.java
danworth: 👍

PR #977, opened 2019-01-21
Pp 4637: Decrypt tokenization data from Google Pay

oswaldquek:
Decryption follows the instructions as per the guide: https://developers.google.com/pay/api/web/guides/resources/payment-data-cryptography#using-tink.

At this point I've done manual testing and can confirm the guide works. However as the tokenization data was made using my real card number I've declined to put that response in our code repo and therefore the GooglePayDecrypterTest must still be marked as @Ignored for now.


src/main/java/uk/gov/pay/connector/app/GooglePayKeyManagement.java
```diff
@@ -0,0 +1,26 @@
...
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+public class GooglePayKeyManagement {
+
+    @NotNull
+    @JsonProperty("production")
+    private Boolean production;
+
+    @Valid
```
danworth: I believe `@valid` is used to signal cascading validation on the type of the parameter it is annotating (e.g. if a class with its own validations then it signals to apply those validations in a cascading manner), since this is a `String` I'm not sure what this will do here. Perhaps `@NotEmpty` is more appropriate?

src/main/java/uk/gov/pay/connector/wallets/PaymentData.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.pay.connector.wallets;
+
+import java.util.Optional;
+
+public class PaymentData {
+    
+    public final Optional<String> onlinePaymentCryptogram, eciIndicator;
```
danworth: I'm of the understanding that using `Optionals` as types for class members is generally discouraged in Java, something related to them not being serialisable and only having been designed to work as a return type. That said I've never really found something completely concrete do dissuade their use (Effective Java mentions not use as class member type). 

I guess this is necessary since there's no `getter`. I'm ok with the immutable public no getter pattern, but in this case since it means using Optional as a type I'm a little less comfortable since the `getter` would have been providing some useful functionality by wrapping with an Optional and avoiding Optional as a class member type. Any thoughts?

src/main/java/uk/gov/pay/connector/wallets/googlepay/api/EncryptedPaymentData.java
```diff
@@ -0,0 +1,123 @@
+package uk.gov.pay.connector.wallets.googlepay.api;
```
danworth, line 1: 👍 Nice refactoring to pull this out and I think a useful step towards some future refactoring I have in mind to unify the Apple and Google auth request classes.

src/test/java/uk/gov/pay/connector/wallets/googlepay/api/GooglePayAuthRequestTest.java
```diff
@@ -22,9 +22,9 @@
     public void shouldDeserializeFromJsonCorrectly() throws IOException {
         ObjectMapper objectMapper = Jackson.getObjectMapper();
         
-        JsonNode expected = objectMapper.readTree(fixture("googlepay/exampleAuthRequest.json"));
+        JsonNode expected = objectMapper.readTree(fixture("googlepay/example-auth-request.json"));
```
danworth, line 25: 👍 thanks.

src/main/java/uk/gov/pay/connector/wallets/PaymentData.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.pay.connector.wallets;
+
+import java.util.Optional;
+
+public class PaymentData {
```
danworth: Name makes sense when I see how it is used but I wasn't sure what it was on first look and wonder if we can name it something to help differentiate from the `wallet/api/WalletPaymentInfo`? Perhaps `DecryptedPaymentData`?

danworth (changes requested):
Looks good, well done for getting it working! Minor point around naming of `PaymentData` and a validation annotation. Some discussion around the use of Optionals as class member types which I've been led to believe should be avoided but I see why you've done it.

src/main/java/uk/gov/pay/connector/app/GooglePayKeyManagement.java
oswaldquek: yeah you're right. i was copying the code from somewhere else. `@NotNull` will suffice I think.

src/main/java/uk/gov/pay/connector/wallets/PaymentData.java
oswaldquek: I'm not sure about this `wallet/api/WalletPaymentInfo` thing as I can't see any `PaymentData` there. There is a `PaymentData` living under the `AppleDecryptedPaymentData` but the idea is that that will be deleted eventually and both Google and Apple pay will use this class.

src/main/java/uk/gov/pay/connector/wallets/PaymentData.java
oswaldquek: I think this is ok as we're not going to be serialising this class. This class is meant to hold information for populating the `cryptogram` and `eciIndicator` and `tokenNumber` when constructing a payment to Worldpay. See http://support.worldpay.com/support/kb/gg/corporate-gateway-guide/content/mobilewallets/androidpaydecrypted.htm and http://support.worldpay.com/support/kb/gg/corporate-gateway-guide/content/mobilewallets/applepaydecrypted.htm

oswaldquek: And as it is, non-3ds payments don't have the `cryptogram` and `eciIndicator` so they have to be optionals here

src/main/java/uk/gov/pay/connector/app/GooglePayKeyManagement.java
oswaldquek: fixed

src/main/java/uk/gov/pay/connector/wallets/PaymentData.java
danworth: Happy to go with your better judgement, naming things isn't my strongest skill. I feel that `WalletPaymentInfo` can be construed as payment information and my point was only that it isn't immediately apparent the difference between two similarly named classes when 'info' and 'data' are synonyms. We can always look at renaming later on once the implementation has settled down and we know what is left.
```
public class WalletPaymentInfo {
    
    private String lastDigitsCardNumber;
    private String brand;
    private PayersCardType cardType;
    private String cardholderName;
    private String email;
```

src/main/java/uk/gov/pay/connector/wallets/PaymentData.java
danworth: Ok, maybe I need to let go of the "don't use Optionals for class member types". It seems to originate out of a comment from an engineer that worked on the implementation of Optionals stating its design considered their use as a return type only. I can see that their use can be applied beyond their original intended purpose. Since `PaymentData` doesn't implement `Serialisable` itself then this seems safe and I should welcome the future with open arms...

src/main/java/uk/gov/pay/connector/app/GooglePayKeyManagement.java
danworth: 👍

src/main/java/uk/gov/pay/connector/wallets/PaymentData.java
oswaldquek: i've renamed this to the suggested `DecryptedPaymentData`

src/main/java/uk/gov/pay/connector/wallets/DecryptedPaymentData.java
```diff
@@ -0,0 +1,15 @@
+package uk.gov.pay.connector.wallets;
+
+import java.util.Optional;
+
+public class DecryptedPaymentData {
+    
+    public final Optional<String> onlinePaymentCryptogram, eciIndicator;
```
danworth, line 7: public immutable members is worth putting on the standards agenda. Might not be everyones liking but since this has limited scope, easy to revert if necessary and it'll serve as an example I feel ok to approve as-is. Since the force-push lost the conversation about Optionals as class member types I'm just saying that here since that might be worth adding the the agenda too.

danworth (approved):
Worth adding the public immutable members to the standards agenda. Approved as-is for reasons in my comment. 👍

PR #982, opened 2019-01-23
PP-4637: Revert decryption

oswaldquek:
A decision has been taken to not decrypt the google payload as there is already
a key management process in place between google pay and worldpay (and other
supported payment gateways even). We can now simply pass the payload from
google straight through to worldpay.

@oswaldquek


danworth (approved):
On a positive note that's 194 lines of code gone :)

PR #983, opened 2019-01-23
PP-4637: Decryption ADR

oswaldquek:
@oswaldquek

### DIFF
--- a/docs/arch/adr-002-decrypting-google-pay-token-not-required.md
+++ b/docs/arch/adr-002-decrypting-google-pay-token-not-required.md
@@ -1,0 +1,56 @@
+# ADR 002 - Don't decrypt payload from Google Pay
+
+## Context
+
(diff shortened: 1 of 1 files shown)


alexbishop1 (changes requested):
I’m possibly being a bit pernickety here but it feels like quite a lot of the decision is actually in the context part.

In the context bit, we could maybe start with a brief description of the two possible options (decrypting and not decrypting), noting what data we get with each and how the key/certificate requirements differ for each. The deal-breakers list (which is very good) could go either before or after this. Then the decision part could explain why not decrypting is best (in terms of the data we want being available and neither us nor our partner services having to mess with keys).

oswaldquek:
@alexbishop1 Thanks for your input. I don't think you're being pernickety. I've never written an ADR before so your input is welcome, and in this case, I think valid.

alexbishop1:
I’ve just corrected an error in my initial review! I said “decision” at the beginning of the second paragraph when I meant “context”! Sorry!

alexbishop1 (approved):
Awesome summary of the reasoning

PR #998, opened 2019-01-24
bau: Refactor refund testing

oswaldquek:
Code can be made simpler if the refund Id were generated within the
DatabaseTestHelper's addRefund method.

@oswaldquek


danworth (approved):
Looks good, since the refundId was being randomly generated each time, it makes sense to put inside the `DatabaseTestHelper.addRefund` and return it 👍 I assume there's never a need to use a specific refund id when creating it, or least that doesn't exist at present (even if that occurs later on we can just overload the method I guess).

PR #999, opened 2019-01-24
PP-4638 Refactor ApplePay Gateway authorisation for reuse with other …

danworth:
…wallet types.

To prepare for implementing Googlepay authorisation and aid re-use of functionality
originally built for Apple Pay. The intention of this PR is a gateway wallet authorisation interface and implementations that accept a WalletAuthorisationGatewayRequest which can be implemented for any wallet type (the difference being the WalletTemplateData contained within)

- Where necessary remove specific reference to Apple Pay in function and class
names and use interfaces that can be shared by other wallet types.
- `ApplePayAuthorisationGatewayRequest` becomes `WalletAuthorisationGatewayRequest`
and uses type `WalletTemplateData` rather than `ApplePayTemplateData`.

## WHAT
Preparation for implementing GooglePay by making the Gateway authorisation flow originally built for Apple pay a little more general to other wallet types. This is the first part of refactoring and focuses on the Gateway side of the service, the second part which is TBD will look at refactoring the `ApplepayService` and `ApplepayAuthorisationService` to reuse with Googlepay.

## HOW 
Removed reference to Applepay in classes and functions within the Gateway authorisation flow where re-use is expected. Introduced some polymorphism based on more general wallet related interfaces.


src/main/java/uk/gov/pay/connector/gateway/sandbox/applepay/SandboxWalletAuthorisationHandler.java
```diff
@@ -12,10 +12,10 @@
 import static uk.gov.pay.connector.gateway.model.ErrorType.GENERIC_GATEWAY_ERROR;
 import static uk.gov.pay.connector.gateway.model.response.GatewayResponse.GatewayResponseBuilder.responseBuilder;
 
-public class SandboxApplePayAuthorisationHandler implements ApplePayAuthorisationHandler {
+public class SandboxWalletAuthorisationHandler implements WalletAuthorisationHandler {
```
oswaldquek, line 15: doesn't affect this PR but I observe that this class could probably be deleted

oswaldquek (approved):

src/main/java/uk/gov/pay/connector/gateway/sandbox/applepay/SandboxWalletAuthorisationHandler.java
danworth: Thanks I'll try and include in the next refactoring.

PR #1000, opened 2019-01-24
PP-4695, PP-4696: Upgrade httpclient and logback

rhowe-gds:
## WHAT
Upgrade logback from 1.2.2 to 1.2.3

Release notes: https://logback.qos.ch/news.html

Nothing notable

Upgrade httpclient from 4.5.3 to 4.5.6

Release notes: https://archive.apache.org/dist/httpcomponents/httpclient/RELEASE_NOTES-4.5.x.txt

Nothing stands out as super important

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -196,7 +196,7 @@
         <dependency>
             <groupId>org.apache.httpcomponents</groupId>
             <artifactId>httpclient</artifactId>
-            <version>4.5.3</version>
+            <version>4.5.6</version>
         </dependency>
         <dependency>
             <groupId>ch.qos.logback</groupId>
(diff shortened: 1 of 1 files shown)


alexbishop1 (approved):
One of the HttpClient fixes:

```
* [HTTPCLIENT-1889] org.apache.http.client.utils.URLEncodedUtils.parse() 
  should return a new ArrayList when there are no query parameters.
  Contributed by Gary Gregory <ggregory at apache.org>
```

Turns out we use `URLEncodedUtils.parse(…)` in one place (`EpdqNotification`). And we assume the result is always a `List`!

PR #1004, opened 2019-01-25
Pp 4638 refactor apple pay authorisation service original

danworth:
## WHAT
To prepare for implementing Googlepay this PR refactors the existing `ApplePayAuthoriseService` and `ApplePayService`.
- Remove specific reference to apple pay within `ApplePayAuthoriseService` and introduce common interface `WalletAuthorisationData` rather than working with `AppleDecryptedPaymentData`. Rename the service to `WalletAuthoriseService` and move into `wallet` package.
- Bring common code from existing `ApplePayService` into a new abstract class `WalletService` with a single abstract method to convert from the auth request received by the end-point into the new `WalletAuthorisationData` to hand off to the `WalletAuthoriseService`.
- Make existing `ApplePayService` extend `WalletService`, removing duplicated code and place decryption logic into the abstract method providing a concrete implementation.

This should make the implementation of Googlepay a case of creating a `GooglepayService` providing the implementation for the single abstract method to turn the end-point request into a `WalletAuthorisationData` object (no decryption so straight forward).

Since the only difference is the abstract method it feels this could be accomplished via composition and the use of the Strategy Pattern to avoid inheritance, however this provides a simple solution which can be improved upon later, whilst not blocking following stories.


src/main/java/uk/gov/pay/connector/wallets/WalletAuthorisationRequest.java
```diff
@@ -0,0 +1,8 @@
+package uk.gov.pay.connector.wallets;
+
+import uk.gov.pay.connector.wallets.model.WalletPaymentInfo;
+
+public interface WalletAuthorisationRequest {
+    
+    public WalletPaymentInfo getPaymentInfo();
```
oswaldquek: no need for a `public` here

src/main/java/uk/gov/pay/connector/wallets/model/WalletAuthorisationData.java
```diff
@@ -0,0 +1,14 @@
...
+
+import uk.gov.pay.connector.wallets.WalletType;
+
+import java.time.LocalDate;
+
+public interface WalletAuthorisationData {
+    
+    WalletPaymentInfo getPaymentInfo();
+    WalletTemplateData getWalletTemplateData();
+    LocalDate getApplicationExpirationDate();
```
oswaldquek: i'm not sure about this method name. `getCardExpiry` might be better? it's just that i'm not sure what `Application` is which is potentially confusing.

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/wallets/WalletAuthorisationRequest.java
danworth: 👍 thanks, I'll correct.

src/main/java/uk/gov/pay/connector/wallets/model/WalletAuthorisationData.java
danworth: 👍 Yes I agree, I wasn't clear what `Application` was either but kept the existing name in favour of minimising change. That said, since you've raised it too, I'll change it in this PR.

oswaldquek (approved):

danworth:
The Snyk vulnerabilities appear to be related to the branch not being rebased to include recent POM change on master, so all should be ok when merged onto master with correct pom version.

PR #1005, opened 2019-01-25
PP-4512 Make AUTHORISATION 3DS REQUIRED expirable with gateway

mrlumbu:
In certain circumstances a charge that is in status `AUTHORISATION 3DS REQUIRED`
can exist in an authorised state with the gateway (specifically ePDQ - see
https://payments-platform.atlassian.net/browse/PP-4512).

If the charge is stuck in this state then the expiry sweep expires the charge in our
DB, but not with the gateway. This causes user pain, as the authorisation hold
is not released for days/weeks.

This PR adds `AUTHORISATION 3DS REQUIRED` as a 'gateway expirable' status.
In many cases this is likely to fail as the charge won't exist with the gateway
but the job seems to be robust to such failures (the failure is reported
in the HTTP response, but has no other effect).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
@@ -63,6 +63,7 @@ public class ChargeExpiryService {
     private final ChargeEventDao chargeEventDao;
     private final PaymentProviders providers;
     static final List<ChargeStatus> GATEWAY_CANCELLABLE_STATUSES = Arrays.asList(
+            AUTHORISATION_3DS_REQUIRED,
             AUTHORISATION_SUCCESS,
             AWAITING_CAPTURE_REQUEST
     );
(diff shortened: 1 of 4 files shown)


DanailMinchev (dismissed):

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -63,6 +63,7 @@
     private final ChargeEventDao chargeEventDao;
     private final PaymentProviders providers;
     static final List<ChargeStatus> GATEWAY_CANCELLABLE_STATUSES = Arrays.asList(
+            AUTHORISATION_3DS_REQUIRED,
```
alexbishop1, line 66: Maybe this could benefit from a comment explaining briefly why as this status being in the list is quite counter-intuitive

alexbishop1 (changes requested):
Is there not a test for this?

alexbishop1 (approved):

PR #1007, opened 2019-01-27
PP-4637: Remove json-path-assert from pom

rhowe-gds:
The usage of this library was removed in commit
e4af978d0334eb4ab4ae89df57dd6403f3338ba0


rhowe-gds:
This is not true

PR #1015, opened 2019-01-28
PP-4512 Payment discrepancy checker

mrlumbu:
Adds an endpoint that takes a list of charge IDs and for each one checks the status we have recorded in the DB vs the status as recorded by the payment gateway.

At the moment it only implements the checking functionality for ePDQ, as that is the immediate need, but can easily be extended in future.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
@@ -6,6 +6,7 @@ import io.dropwizard.setup.Environment;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.gateway.epdq.model.response.EpdqQueryResponse;
 import uk.gov.pay.connector.wallets.WalletAuthorisationGatewayRequest;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability;
@@ -34,6 +35,7 @@ import uk.gov.pay.connector.gateway.util.ExternalRefundAvailabilityCalculator;
 import uk.gov.pay.connector.gateway.util.GatewayResponseGenerator;
 
 import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.client.Invocation;
 import java.nio.charset.Charset;
 import java.util.Optional;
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/pay/connector/report/resource/DiscrepancyReportResource.java
```diff
@@ -0,0 +1,31 @@
...
+    private final DiscrepancyService discrepancyService;
+    @Inject
+    public DiscrepancyReportResource(DiscrepancyService discrepancyService) {
+        this.discrepancyService = discrepancyService;
+    }
+
+    @GET
+    @Path("/v1/api/reports/discrepancies")
+    @Produces(APPLICATION_JSON)
+    public List<Discrepancy> getPerformanceReport(@QueryParam("chargeIds") @NotEmpty List<String> chargeIds) {
```
heathd: method name?

heathd:
looks good.

I wonder if we want a way to spit out the underlying/raw status from the gateway as well as the mapped status. 

But maybe this is good enough for now.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -130,6 +130,18 @@ public GatewayRefundResponse refund(RefundGatewayRequest request) {
     public GatewayResponse<BaseAuthoriseResponse> authoriseWallet(WalletAuthorisationGatewayRequest request) {
         throw new UnsupportedOperationException("Wallets are not supported for ePDQ");
     }
+    
+    @Override
+    public BaseAuthoriseResponse.AuthoriseStatus queryPaymentStatus(ChargeEntity charge) {
+        Either<GatewayError, GatewayClient.Response> response = authoriseClient.postRequestFor(
```
alexbishop1, line 137: Not sure about this being so tied to authorisation (`BaseAuthoriseResponse`, `authoriseClient` etc.). In theory this could be used for querying a payment in any status.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -130,6 +130,18 @@ public GatewayRefundResponse refund(RefundGatewayRequest request) {
...
+                ROUTE_FOR_QUERY_ORDER,
+                charge.getGatewayAccount(),
+                buildQueryOrderRequestFor(charge)
+        );
+        GatewayResponse<BaseAuthoriseResponse> gatewayResponse = GatewayResponseGenerator.getEpdqGatewayResponse(authoriseClient, response, EpdqAuthorisationResponse.class);
+        return gatewayResponse.getBaseResponse().map(epdqStatus -> ((EpdqAuthorisationResponse) epdqStatus).authoriseStatus()).orElse(BaseAuthoriseResponse.AuthoriseStatus.ERROR);
```
alexbishop1: I think this is too broad. As far as I can tell, we could end up with `AuthoriseStatus.ERROR` in any of the following scenarios:

- payment actually is in an ePDQ status we consider to map to `AuthoriseStatus.ERROR`
- the payment has a post-authorisation status
- connection to ePDQ timed out
- the payment doesn’t exist in ePDQ

alexbishop1 (commented):
I think the general approach is fine but I have some concerns about the specifics of how it talks to ePDQ reusing so much of the authorisation stuff.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
mrlumbu: I've made the response thing more generic. The client is just silly really - there's nothing very authorisey about it.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
mrlumbu: I agree. I have changed this so the response from ePDQ gets serialised into a `ChargeQueryResponse` which has no opinions about what the staus should be - this is then used along with `EpdqStatusMapper` to try and map to a Pay status (this is somewhat imperfect as there are many Pay charge statuses that map to one ePDQ status (e.g. EXPIRED and USER_CANCELLED would both have status 6 in ePDQ), but this at least allows us to see high level mismatches - in particular by looking at whether the external statuses match. I guess an improvement of this would be to look at the events for a charge as well, but I think this is good enough for now.

PR #1018, opened 2019-01-29
PP-4512 Basic Discrepancy resolver

mrlumbu:
This PR adds an endpoint that takes a list of chargeIds and for each one checks the
relative status of the charge with Pay and with the Gateway (currently only possible for ePDQ). For certain kinds of mismatch it will then resolve the discrepancy.

As it stands, the only case in which it will try to resolve a discrepancy is if the Pay status is `EXPIRED` and the gateway status is `AUTHORISED`.

It resolves the discrepancy by 'force cancelling' the charge with the gateway - that is it short circuits all the checking we have put in place to ensure correct status transitions, and does not add anything to the events for a charge. I think this is ok - happy to discuss.

Builds on https://github.com/alphagov/pay-connector/pull/1015, which simply reports possible discrepancies from a provided list of chargeIds. It returns a JSON array of the form:
```
[
    {
        "payStatus": "EXPIRED",
        "gatewayStatus": "AUTHORISATION_SUCCESS",
        "chargeId": "charge1017326793",
        "rawGatewayResponse": "ePDQ query response (PAYID: 3014644340, STATUS: 5)",
        "processed": true,
        "gatewayExternalStatus": "EXTERNAL_SUBMITTED",
        "payExternalStatus": "EXTERNAL_FAILED_EXPIRED"
    },
    {
        "payStatus": "EXPIRED",
        "gatewayStatus": "AUTHORISATION_SUCCESS",
        "chargeId": "charge285518240",
        "rawGatewayResponse": "ePDQ query response (PAYID: 3014644340, STATUS: 5)",
        "processed": true,
        "gatewayExternalStatus": "EXTERNAL_SUBMITTED",
        "payExternalStatus": "EXTERNAL_FAILED_EXPIRED"
    }
]
```
which can be used to figure out whether a charge needs resolving or not.


heathd (approved):
Output looks good

PR #1019, opened 2019-01-29
PP-4512 Fix serialisation issues

mrlumbu:
Fixes a few small problems with the serialisation of
`GatewayStatusComparison`, including some rather silly
null pointer exceptions \o/


alexbishop1 (approved):
🤦‍♂️

PR #1021, opened 2019-01-30
Streams streams streams

rhowe-gds:
## WHAT
Make some code more functional & use the streams API better.

See individual commits for details

## HOW 
Do the tests pass?
Does the code make sense?
Is it actually better?

### DIFF
--- a/src/main/java/uk/gov/pay/connector/healthcheck/resource/HealthCheckResource.java
+++ b/src/main/java/uk/gov/pay/connector/healthcheck/resource/HealthCheckResource.java
@@ -1,7 +1,6 @@
 package uk.gov.pay.connector.healthcheck.resource;
 
 import com.codahale.metrics.health.HealthCheck;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import io.dropwizard.setup.Environment;
@@ -10,18 +9,17 @@ import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
-import java.util.HashMap;
+import java.util.Collection;
 import java.util.Map;
 import java.util.SortedMap;
+import java.util.stream.Collectors;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.Response.status;
-import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.defaultString;
 
 @Path("/")
 public class HealthCheckResource {
-    public static final String HEALTHY = "healthy";
-    public static final String MESSAGE = "message";
 
     private Environment environment;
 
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/report/dao/PerformanceReportDao.java
```diff
@@ -42,7 +43,7 @@ public PerformanceReportEntity aggregateNumberAndValueOfPayments() {
       .getSingleResult();
   }
 
-  public List<GatewayAccountPerformanceReportEntity> aggregateNumberAndValueOfPaymentsByGatewayAccount() {
+  public Stream<GatewayAccountPerformanceReportEntity> aggregateNumberAndValueOfPaymentsByGatewayAccount() {
```
oswaldquek, line 46: noice

oswaldquek (dismissed):

src/main/java/uk/gov/pay/connector/healthcheck/resource/HealthCheckResource.java
```diff
@@ -33,29 +33,22 @@ public HealthCheckResource(Environment environment) {
...
+                                "message", defaultString(healthCheck.getValue().getMessage(), "Healthy"))
+                        )
+                );
 
-        if(healthy) {
-            return Response.ok().entity(response).build();
-        }
-        return status(503).entity(response).build();
-    }
+        boolean allHealthy = results.values().stream().allMatch(HealthCheck.Result::isHealthy);
```
danworth: This could be a private function then in-line it rather than creating a local boolean variable e.g.
```
Response.ResponseBuilder res = allHealthy(results) ? Response.ok() : status(503);

private boolean allHealthy(Map<String, HealthCheck.Result> results) {
        return results.values().stream().allMatch(HealthCheck.Result::isHealthy);
    }
```
I'll stop here since Oz has reviewed the rest. This isn't mandatory, it works and makes sense as-is.

danworth (commented):
Only got so far before Oz approved so I'll leave the one comment I had, no need to include I'll leave it up to you...I've been reading the latest version of Refactoring so I might be a bit too keen :)

src/main/java/uk/gov/pay/connector/healthcheck/resource/HealthCheckResource.java
rhowe-gds: No, I like it

rhowe-gds: Initially I inlined the code into the original `if` and @tlwr looked ill

rhowe-gds: Changed to take a `Collection<HealthCheck.Result>`, so all the knowledge of the structure of `runHealthChecks()`'s result is in `healthCheck()` but we still have some functional decomposition.

danworth: 👍

danworth (approved):
@oswaldquek approved before, this approves the difference of implementing `allHealthy()` 👍

PR #1022, opened 2019-01-30
PP-4758: Reduce error types

oswaldquek:
Mapping error types to charge statuses happens in CardAuthoriseBaseService and
it can be seen the only relevant error types are GENERIC_GATEWAY_ERROR and
GATEWAY_CONNECTION_TIMEOUT_ERROR:
```
    private ChargeStatus mapError(GatewayError gatewayError) {
        switch (gatewayError.getErrorType()) {
            case GENERIC_GATEWAY_ERROR:
                return AUTHORISATION_ERROR;
            case GATEWAY_CONNECTION_TIMEOUT_ERROR:
                return AUTHORISATION_TIMEOUT;
            default:
                return AUTHORISATION_UNEXPECTED_ERROR;
        }
```
All other error types aren't used.

This commit removes the error types we don't care about and will help the
merging of GatewayClient and StripeGatewayClient by virtue of there being less
code to reason about when doing the refactoring.

@oswaldquek

## WHAT

## HOW 
_Steps to test or reproduce:_


src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -83,25 +78,17 @@ public GatewayClient(Client client,
...
             if (pe.getCause() != null) {
-                if (pe.getCause() instanceof UnknownHostException) {
-                    logger.error(format("DNS resolution error for gateway url=%s", gatewayUrl), pe);
-                    return left(unknownHostException("Gateway Url DNS resolution error"));
-                }
                 if (pe.getCause() instanceof SocketTimeoutException) {
                     logger.error(format("Connection timed out error for gateway url=%s", gatewayUrl), pe);
                     return left(gatewayConnectionTimeoutException("Gateway connection timeout error"));
                 }
-                if (pe.getCause() instanceof SocketException) {
```
danworth, line 99: I think this is ok, so long as we're not relying on a particular log message for these kinds of exceptions (e.g. looking for "Socket Exception..." for a metric/warning). I don't think we lose any information since the same information will be logged by the more general "Exception for gateway..." further down.

src/main/java/uk/gov/pay/connector/gateway/model/ErrorType.java
```diff
@@ -2,9 +2,6 @@
 
 public enum ErrorType {
     GENERIC_GATEWAY_ERROR,
-    UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY,
-    MALFORMED_RESPONSE_RECEIVED_FROM_GATEWAY,
-    GATEWAY_URL_DNS_ERROR,
     GATEWAY_CONNECTION_TIMEOUT_ERROR,
-    GATEWAY_CONNECTION_SOCKET_ERROR,
+    OTHER
```
danworth: I wonder if there's a better name than `OTHER` since its quite similar in meaning to `GENERIC` and the `GENERIC_GATEWAY_ERROR` seems to actually be the catch-all "other" code already. Looking at what it is replacing and how they are currently used it seems they relate to connection problems, unknown host, dns and socket errors. If none of those are hit then the catch-all `catch (Exception e)` maps to `GENERIC_GATEWAY_ERROR`. So I'm wondering why we originally made the distinction for  `UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY`, `MALFORMED_RESPONSE_RECEIVED_FROM_GATEWAY`, `GATEWAY_URL_DNS_ERROR` and `GATEWAY_CONNECTION_SOCKET_ERROR` and mapped them to `AUTHORISATION_UNEXPECTED_ERROR` charge status (aren't all errors unexpected?) rather than `AUTHORISATION_ERROR`? I'm not sure if that's important but I would suggest we follow the more descriptive naming convention of the remaining types and include `ERROR` and `GATEWAY` to aid comprehension where it subsequently occurs and maybe replace `OTHER` with something that relates to the connection nature of the codes its replacing....so something like `GATEWAY_CONNECTION_ERROR`?

src/main/java/uk/gov/pay/connector/gateway/model/GatewayError.java
```diff
@@ -27,44 +22,26 @@ public GatewayError(String message, ErrorType errorType) {
...
-        return new GatewayError(msg, GENERIC_GATEWAY_ERROR);
-    }
-
-    public static GatewayError unexpectedStatusCodeFromGateway(String msg) {
-        return new GatewayError(msg, UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY);
-    }
-
-    public static GatewayError malformedResponseReceivedFromGateway(String msg) {
-        return new GatewayError(msg, MALFORMED_RESPONSE_RECEIVED_FROM_GATEWAY);
+    public static GatewayError other(String msg) {
```
danworth: This would need changing if you agreed with my comment about a more descriptive name for the type than `OTHER`.

danworth (changes requested):
I like the intention of this PR to simplify our gateway client code and it looks sensible 👍 👍 👍   I've added a comment discussing the new name of the `ErrorType` which I think we should help differentiate from the remaining catch-all `GENERIC_GATEWAY_ERROR` and relate to the values its replacing. Happy to discuss thoughts.

src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
oswaldquek: yeah we won't lose any information as i'm logging the error message: `logger.error(format("Exception for gateway url=%s, error message: %s", gatewayUrl, pe.getMessage()), pe)` Basically any error gets logged properly.

src/main/java/uk/gov/pay/connector/gateway/model/ErrorType.java
oswaldquek: yes sounds good. i'll change it to `GATEWAY_CONNECTION_ERROR`

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqCaptureHandlerTest.java
```diff
@@ -99,14 +99,14 @@ public void shouldNotCaptureIfPaymentProviderReturnsUnexpectedStatusCode() throw
...
-        Either<GatewayError, GatewayClient.Response> response = left(unexpectedStatusCodeFromGateway("Unexpected HTTP status code 400 from gateway"));
+        Either<GatewayError, GatewayClient.Response> response = left(genericGatewayError("Unexpected HTTP status code 400 from gateway"));
         when(client.postRequestFor(anyString(), any(GatewayAccountEntity.class), any(GatewayOrder.class))).thenReturn(response);
         
         CaptureResponse gatewayResponse = epdqCaptureHandler.capture(buildTestCaptureRequest());
         assertThat(gatewayResponse.isSuccessful(), is(false));
         assertThat(gatewayResponse.getError().isPresent(), is(true));
         assertThat(gatewayResponse.getError().get().getMessage(), is("Unexpected HTTP status code 400 from gateway"));
-        assertThat(gatewayResponse.getError().get().getErrorType(), is(UNEXPECTED_HTTP_STATUS_CODE_FROM_GATEWAY));
+        assertThat(gatewayResponse.getError().get().getErrorType(), is(GENERIC_GATEWAY_ERROR));
```
danworth: Shouldn't the codes being replaced map to `GATEWAY_CONNECTION_ERROR`? I think where some of these are calling `genericGatewayError()` it should be calling the new `gatewayConnectionError()`.

danworth (changes requested):
I noticed in the tests that the codes being replaced should map to the new one, I think its just calling the wrong method to generate it as per my comment.

danworth (approved):
👍 Looks good.

PR #1026, opened 2019-01-31
PP-4597 Replaced String with ZonedDateTime and used custom serializer

georgeracu:
- Part of the movement to make consistent the type used for API responses


gidsg (approved):
Looks good to me *(if you can appease Codacy)* 👍

PR #1027, opened 2019-01-31
PP-4512 Relax conditions for discrepancy resolution

mrlumbu:
Amends the criteria applied to decide whether a charge can be 'force cancelled'
with the gateway. The criteria are now:
- charge has a not finished state with gateway
- charge is finished and not successful with pay
- charge is older than 7 days (this is to avoid accidentally cancelling something
that is temporarily out of sync)

At the moment doing this kind of cancellation may still leave charges in a slightly
weird state. For example if a charge is in `AUTHORISATION_ERROR` with Pay, but
in `AUTHORISATION_SUCCESS` with the gateway (and is older than 7 days etc), then
this change will mean this charge will be allowed to be cancelled, but it
will not result in the states matching, since it makes no change on the Pay
side. Adding something to try and align the Pay state with the gateway state
in this kind of case would be a good next step, but involves a fair bit more work
as we would need to relax all the restrictions around status change etc.

This commit makes some small changes to EPDQ Status Mapper - see
https://support.epdq.co.uk/~/media/kdb/pdf/epdq/en/955815bf-4514-4d75-b683-21034ff5789b/statuses-and-errors.ashx

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapper.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapper.java
@@ -15,14 +15,18 @@ public class EpdqStatusMapper {
     
     static {
         Map<String, ChargeStatus> aMap = new HashMap<>();
-        aMap.put("5", ChargeStatus.AUTHORISATION_SUCCESS);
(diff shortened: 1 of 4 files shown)


kbottla (approved):
LGTM 👍

PR #1029, opened 2019-02-01
PP-4758: Change signature of GatewayClient

oswaldquek:
In an effort to make code simpler, we want to remove some functional idioms
that aren't particularly suited to Java. GatewayClient returns an
`Either<GatewayError, GatewayClient.Response>` which make upstream calling
classes complicated as we end up doing bi-Mapping and the like which is ugly
and confusing. When something goes wrong the java way is to throw exceptions
and catch them, which is what this commit does. GatewayClient now returns a
`GatewayClient.Response` which is a successful response. Unfortunately this is
a massive change as it's used by all the payment providers bar Stripe and there
is no way to reduce this.

With this in place we can deprecate the StripeGatewayClient which will be in
another PR.

Also I'd be keen to hear people's thoughts on returning `Either` vs returning `GatewayClient.Response` and throwing Exceptions. In some ways I like the `Either` thing as it can be neater... throwing Exceptions can make code look clunky.

## HOW 
This is a massive PR - unfortunately there is no way round this as the GatewayClient is used in so many places. In my estimation about half of the PR is making the tests work, and many of these are just making tests throw Exceptions so it hopefully shouldn't be as daunting as it seems.


src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -74,26 +69,28 @@ public GatewayClient(Client client,
...
                     logger.error(format("Connection timed out error for gateway url=%s", gatewayUrl), pe);
-                    return left(gatewayConnectionTimeoutException("Gateway connection timeout error"));
+                    throw new GatewayConnectionTimeoutErrorException("Gateway connection timeout error");
                 }
             }
             logger.error(format("Exception for gateway url=%s, error message: %s", gatewayUrl, pe.getMessage()), pe);
-            return left(genericGatewayError(pe.getMessage()));
+            throw new GenericGatewayErrorException(pe.getMessage());
+        } catch(GatewayConnectionErrorException e) {
+            throw e;
```
mrlumbu, line 89: not sure what the purpose of this is?

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqRefundHandler.java
```diff
@@ -26,14 +29,16 @@ public EpdqRefundHandler(GatewayClient client) {
...
-        if (response.isLeft()) {
-            return GatewayRefundResponse.fromGatewayError(response.left().value());
-        } else {
-            Either<GatewayError, EpdqRefundResponse> unmarshalled = client.unmarshallResponse(response.right().value(), EpdqRefundResponse.class);
-            return fromUnmarshalled(unmarshalled, GatewayRefundResponse.RefundState.PENDING);
+        EpdqRefundResponse unmarshalled;
+        
+        try {
+            GatewayClient.Response response = client.postRequestFor(ROUTE_FOR_MAINTENANCE_ORDER, request.getGatewayAccount(), buildRefundOrder(request));
+            unmarshalled = unmarshallResponse(response, EpdqRefundResponse.class);
```
mrlumbu: is there a reason you don't just return this here?

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -123,54 +134,71 @@ public GatewayRefundResponse refund(RefundGatewayRequest request) {
...
-        GatewayResponse<EpdqQueryResponse> epdqGatewayResponse = GatewayResponseGenerator.getEpdqGatewayResponse(authoriseClient, response, EpdqQueryResponse.class);
-        
-        return epdqGatewayResponse.getBaseResponse().map(EpdqQueryResponse::toChargeQueryResponse).orElseGet(() -> {
-            throw new  WebApplicationException("Things went wrong");
-        });
+        try {
+            GatewayClient.Response response = authoriseClient.postRequestFor(ROUTE_FOR_QUERY_ORDER, charge.getGatewayAccount(), buildQueryOrderRequestFor(charge));
+            GatewayResponse<EpdqQueryResponse> epdqGatewayResponse = getEpdqGatewayResponse(response, EpdqQueryResponse.class);
+            
+            if (!epdqGatewayResponse.getBaseResponse().isPresent()) epdqGatewayResponse.throwGatewayError();
```
mrlumbu: you seem to be deliberately not mapping over `Optional`s - is there a reason for this?

This bit would be fine as
```
return epdqGatewayResponse.getBaseResponse()
                    .map(EpdqQueryResponse::toChargeQueryResponse)
                    .orElseThrow(() -> new WebApplicationException("Things went wrong"));
```
wouldn't it?

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -123,54 +134,71 @@ public GatewayRefundResponse refund(RefundGatewayRequest request) {
...
+                        request.getChargeExternalId(), request.getGatewayAccount().getId(), auth3DResult, authoriseStatus);
+                metricRegistry.counter(format("epdq.authorise-3ds.result.mismatch.account.%s.frontendstatus.%s.gatewaystatus.%s",
+                        request.getGatewayAccount().getGatewayName(),
+                        request.getAuth3DsDetails().getAuth3DsResult(),
+                        authoriseStatus.name()))
+                        .inc();
+                gatewayResponse = reconstructErrorBiasedGatewayResponse(gatewayResponse, authoriseStatus, auth3DResult);
+            }
+            
+            if (!gatewayResponse.getBaseResponse().isPresent()) gatewayResponse.throwGatewayError();
```
mrlumbu, line 185: same comment about mapping optionals

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -203,8 +234,8 @@ public ExternalChargeRefundAvailability getExternalChargeRefundAvailability(Char
         }
     }
 
-    private boolean responseDoesNotMatchWithUserResult(BaseAuthoriseResponse.AuthoriseStatus authoriseStatus, Auth3dsDetails.Auth3dsResult auth3DResult) {
-        boolean respondMatches = (authoriseStatus.equals(BaseAuthoriseResponse.AuthoriseStatus.ERROR) && auth3DResult.equals(Auth3dsDetails.Auth3dsResult.ERROR)) ||
+    private static boolean responseDoesNotMatchWithUserResult(BaseAuthoriseResponse.AuthoriseStatus authoriseStatus, Auth3dsDetails.Auth3dsResult auth3DResult) {
```
mrlumbu: would pefer if you did these kind of small refactorings in a separate PR - they are entirely tangential, and the PR is already pretty big

src/main/java/uk/gov/pay/connector/gateway/GatewayErrors.java
```diff
@@ -0,0 +1,65 @@
...
+        public GenericGatewayErrorException(String message) {
+            super(message);
+        }
+
+        public GatewayError toGatewayError() {
+            return new GatewayError(getMessage(), ErrorType.GENERIC_GATEWAY_ERROR);
+        }
+
+        public ChargeStatus chargeStatus() {
+            return AUTHORISATION_ERROR;
```
mrlumbu: Am not sure how this is going to work - surely the charge status cannot be inferred from the gateway error alone - you need the context of what was being tried. I think it is cleaner for the Exception to just be an exception, and the calling class can figure out stuff about charge statuses with the context it has. Unless I'm missing something?

src/main/java/uk/gov/pay/connector/gateway/GatewayErrors.java
```diff
@@ -0,0 +1,65 @@
...
+import uk.gov.pay.connector.gateway.model.ErrorType;
+import uk.gov.pay.connector.gateway.model.GatewayError;
+
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ERROR;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_TIMEOUT;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_UNEXPECTED_ERROR;
+
+public class GatewayErrors {
+
+    public interface ToGatewayError {
```
mrlumbu: I suggest that you have one abstract class that all the contained Exceptions extend called `GatewayErrorException` or similar
with method `toGatewayError`

Then in all client code they can either rethrow or catch `GatewayErrorException` rather than having to refer to all three if we need more) Exceptions everywhere

I think the ToChargeStatus interface is unnecessary and confusing (see comment below)

So something like 
```
    public abstract class GatewayErrorException extends Exception {
        private GatewayErrorException(String message) {
            super(message);
        }

        public abstract GatewayError toGatewayError();
    }


    public class GenericGatewayErrorException extends GatewayErrorException {

        public GenericGatewayErrorException(String message) {
            super(message);
        }

        public GatewayError toGatewayError() {
            return new GatewayError(getMessage(), ErrorType.GENERIC_GATEWAY_ERROR);
        }

    }

    ...
 ```

File: src/main/java/uk/gov/pay/connector/gateway/GatewayResponseUnmarshaller.java [hunk 8]
```diff
@@ -0,0 +1,26 @@
...
+import uk.gov.pay.connector.gateway.util.XMLUnmarshaller;
+import uk.gov.pay.connector.gateway.util.XMLUnmarshallerException;
+
+import static java.lang.String.format;
+
+public class GatewayResponseUnmarshaller {
+
+    private static final Logger logger = LoggerFactory.getLogger(GatewayResponseUnmarshaller.class);
+    
+    public static <T> T unmarshallResponse(GatewayClient.Response response, Class<T> clazz) throws GatewayConnectionErrorException {
```
mrlumbu: this is not a serious comment, but I really don't like `clazz`

You could call it something like `unmarshallingTarget` or something?

mrlumbu (changes requested):
Overall this looks great - I definitely agree that switching to this way of doing things makes all the client code more readable, and will be a massive improvement when working with this code in the future.

There are some changes I have suggested in my comments. The main things I have suggested are:

- Map over `Optional` when possible. There seem to be cases where you have deliberately chosen not to - I'm not sure  whether this is a deliberate decision on your part, or if this just felt natural in the circumstances, or there's some other reason I've missed. I would prefer that we are consistent in this sort of thing.
- Have all the gateway errors you've introduced extend one abstract class - this will help tidy up all client code and make things cleaner
- Probably don't have a charge status associated with an Exception - I can't see how that will work, and even if it does it's quite confusing
- Theres some smaller things around code style as well, which I don't really have strong feelings about

A final comment - I would have preferred that all refactoring changes had been done in a separate PR, it is quite distracting in what is already a lot of changes. 

Happy to discuss any of this, and as I say, may have missed the reason for some of your decisions, so feel free to push back.

src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
oswaldquek: i wasn't thinking straight... this catch block can be deleted!

src/main/java/uk/gov/pay/connector/gateway/GatewayErrors.java
oswaldquek: You are right - I was overzealous in trying to reduce lines of code, rather than getting the calling class to match on each instance of exception and mapping each to a status which is the right thing to do.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
oswaldquek: the reason i've thrown the gateway error here is that it will be caught in the catch block a few lines below which then sets the authorisationStatus and stringifiedResponse. unfortunately the compiler will complain if we use `.orElseThrow(Supplier)` as the exception thrown from within a Supplier can't be caught in the catch block below.
i also don't think mapping the optional is the right thing to do here because it's not mapping from BaseResponse to another object, it's setting the transactionId, stringifiedResponse and authorisationStatus a few lines down.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqRefundHandler.java
oswaldquek: hmm, I was sure the compiler was complaining before but it appears it's my mistake. will fix.

src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
oswaldquek: oh... actually the purpose is to throw the `GatewayConnectionErrorException` properly, otherwise it will be caught by the `catch (Exception e)` block and rethrown as a `GenericGatewayErrorException`.

src/main/java/uk/gov/pay/connector/gateway/GatewayResponseUnmarshaller.java
oswaldquek: this comes from me probably reading too many hibernate docs in a previous life which had `clazz` all over the place

src/main/java/uk/gov/pay/connector/gateway/GatewayErrors.java
oswaldquek: i've taken onboard your example class here. code is much cleaner and overall it has made this PR reduce the lines of code in connector.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -43,12 +51,41 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
...
+                Optional<BaseAuthoriseResponse> baseResponse = operationResponse.getBaseResponse();
+
+                if (baseResponse.isPresent()) {
+                    chargeStatus = baseResponse.get().authoriseStatus().getMappedChargeStatus();
+                    transactionId = cardAuthoriseBaseService.extractTransactionId(charge.getExternalId(), operationResponse);
+                    sessionIdentifier = operationResponse.getSessionIdentifier();
+                    responseFromPaymentGateway = baseResponse.toString();
+                } else operationResponse.throwGatewayError();
+
+            } catch (GatewayErrorException e) {
```
mrlumbu, line 71: feels like the below block could be extracted into `CardAuthoriseBaseService`? Not a biggie though

src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
mrlumbu: 👍

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
mrlumbu: ok 👍

mrlumbu (approved):
Couple more minor comments, but happy with how this looks now.

🎉

PR #1036, opened 2019-02-05
PP-4512 Check if charge needs expiry with gateway

mrlumbu:
ef8fd47f10a92d9bf23f3223a6ca41714eeccf10 changes the implementation of the logic around charge expiry, so that the 'source of truth' for whether a charge is expirable is contained in the enum `ExpirableChargeStatus`. I think this makes things much clearer, and avoids discrepancies that were possible before, since different services were using there own private variables to figure this out.

53c900dd9c545c3374b9cd06b23bf427d1982d65 changes `ChargeExpiryService` so that, for charges that are in the process of being authorised, it tries to figure out whether the charge exists with the gateway first, and if so, expires it there as well as in our DB. This is obviously an improvement on what we did before, where we guessed if this was the case or not based on our own status only.


src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -82,39 +72,92 @@ public ChargeExpiryService(ChargeDao chargeDao,
...
+                .addAll(getNullSafeList(chargesGroupedByAuthStage.get(POST_AUTHORISATION)))
+                .addAll(getNullSafeList(chargesPartitionedByNeedForExpiryWithGateway.get(Boolean.TRUE)))
+                .build();
+        Pair<Integer, Integer> expireWithCancellationResult = expireChargesWithGateway(toExpireWithGateway);
+
+        return ImmutableMap.of(
+                EXPIRY_SUCCESS, expiredSuccess + expireWithCancellationResult.getLeft(),
+                EXPIRY_FAILED, expireWithCancellationResult.getRight()
+        );
+    }
```
alexbishop1, line 100: I think this would be clearer if it were in the order pre, during, post

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -82,39 +72,92 @@ public ChargeExpiryService(ChargeDao chargeDao,
...
-                EXPIRY_FAILED, expireWithCancellationResult.getRight());
+    private boolean isExpirableWithGateway(ChargeEntity charge) {
+        try {
+            ChargeQueryResponse queryResponse = providers.byName(charge.getPaymentGatewayName())
+                    .queryPaymentStatus(charge);
+            
+            return !queryResponse.getMappedStatus().toExternal().isFinished();
+            
+        } catch (WebApplicationException | UnsupportedOperationException | GatewayErrorException e) {
+            logger.info("Unable to retrieve status for charge {}: {}", charge.getExternalId(), e.getMessage());
```
alexbishop1, line 119: Maybe this should be `warn`?

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -125,6 +125,12 @@ private static GatewayResponse getEpdqGatewayResponse(GatewayClient.Response res
         responseBuilder.withResponse(unmarshallResponse(response, responseClass));
         return responseBuilder.build();
     }
+    
+    private static GatewayResponse getUninterpretedEpdqGatewayResponse(GatewayClient.Response response, Class<? extends BaseResponse> responseClass) throws GatewayConnectionErrorException {
+        GatewayResponse.GatewayResponseBuilder<BaseResponse> responseBuilder = GatewayResponse.GatewayResponseBuilder.responseBuilder();
+        responseBuilder.withResponse(unmarshallResponse(response, responseClass));
+        return responseBuilder.buildUninterpreted();
+    }
```
alexbishop1, line 133: Not sure a static method is the most desirable way to achieve this but I can see the pragmatism

alexbishop1 (approved):
I like the way this makes our expiration logic less hairy

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
mrlumbu: yeah, will look at this in separate PR

PR #1039, opened 2019-02-06
PP-4779: Add/Replace gateway merchant id successfully

oswaldquek:
We could have used the `PATCH /v1/frontend/accounts/{accountId}` method to
update the credentials but operationally curling `PATCH
/v1/frontend/accounts/{accountId}/gateway-merchant-id/{gatewayMerchantId}` manually is
simpler and will be deprecated once the functionality in self service to update
a gateway account with the gateway merchant id is in place.

@oswaldquek @gidsg

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
@@ -170,6 +170,12 @@ public class GatewayAccountEntity extends AbstractVersionedEntity {
         return credentials;
     }
 
+    @JsonProperty("gateway_merchant_id")
+    @JsonView(Views.FrontendView.class)
+    public String getGatewayMerchantId() {
+        return credentials.get("gateway_merchant_id");
+    }
+
     @JsonView(Views.ApiView.class)
     public String getDescription() {
         return description;
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -12,6 +12,7 @@ import java.util.List;
 
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
+import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.FIELD_NOTIFY_API_TOKEN;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.FIELD_NOTIFY_PAYMENT_CONFIRMED_TEMPLATE_ID;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.FIELD_NOTIFY_REFUND_ISSUED_TEMPLATE_ID;
@@ -84,6 +85,12 @@ public class GatewayAccountRequestValidator {
     private void validateGatewayMerchantId(JsonNode payload) {
         throwIfInvalidFieldOperation(payload, ADD_OP, REPLACE_OP);
         throwIfNullFieldValue(payload.get(FIELD_VALUE));
+        throwIfBlankFieldValue(payload.get(FIELD_VALUE));
+    }
+
+    private void throwIfBlankFieldValue(JsonNode valueNode) {
+        if (isBlank(valueNode.asText())) 
+            throw new ValidationException(Collections.singletonList(format("Field [%s] cannot be empty", FIELD_VALUE)));
     }
 
     private void validateNotifySettingsRequest(JsonNode payload) {
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -23,7 +23,9 @@ import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.FIELD_VAL
 public class GatewayAccountRequestValidator {
 
     private static final String REPLACE_OP = "replace";
+    private static final String ADD_OP = "add";
     private static final String REMOVE_OP = "remove";
+    public static final String CREDENTIALS_GATEWAY_MERCHANT_ID = "credentials/gateway_merchant_id";
     public static final String FIELD_ALLOW_WEB_PAYMENTS = "allow_web_payments";
     public static final String FIELD_NOTIFY_SETTINGS = "notify_settings";
     public static final String FIELD_EMAIL_COLLECTION_MODE = "email_collection_mode";
@@ -31,7 +33,9 @@ public class GatewayAccountRequestValidator {
     public static final String FIELD_CORPORATE_DEBIT_CARD_SURCHARGE_AMOUNT = "corporate_debit_card_surcharge_amount";
     public static final String FIELD_CORPORATE_PREPAID_CREDIT_CARD_SURCHARGE_AMOUNT = "corporate_prepaid_credit_card_surcharge_amount";
     public static final String FIELD_CORPORATE_PREPAID_DEBIT_CARD_SURCHARGE_AMOUNT = "corporate_prepaid_debit_card_surcharge_amount";
-    private static final List<String> VALID_PATHS = asList(FIELD_NOTIFY_SETTINGS, 
+    private static final List<String> VALID_PATHS = asList(
+            CREDENTIALS_GATEWAY_MERCHANT_ID,
+            FIELD_NOTIFY_SETTINGS, 
             FIELD_EMAIL_COLLECTION_MODE, 
             FIELD_ALLOW_WEB_PAYMENTS,
             FIELD_CORPORATE_CREDIT_CARD_SURCHARGE_AMOUNT,
@@ -56,6 +60,9 @@ public class GatewayAccountRequestValidator {
             throw new ValidationException(Collections.singletonList(format("Operation [%s] not supported for path [%s]", FIELD_OPERATION, path)));
         
         switch (path) {
+            case CREDENTIALS_GATEWAY_MERCHANT_ID:
+                validateGatewayMerchantId(payload);
+                break;
             case FIELD_NOTIFY_SETTINGS: 
                 validateNotifySettingsRequest(payload);
                 break;
@@ -72,6 +79,11 @@ public class GatewayAccountRequestValidator {
                 validateCorporateCardSurchargePayload(payload);
                 break;
         }
+    }
+
+    private void validateGatewayMerchantId(JsonNode payload) {
+        throwIfInvalidFieldOperation(payload, ADD_OP);
+        throwIfNullFieldValue(payload.get(FIELD_VALUE));
     }
 
     private void validateNotifySettingsRequest(JsonNode payload) {
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -82,7 +82,7 @@ public class GatewayAccountRequestValidator {
     }
 
     private void validateGatewayMerchantId(JsonNode payload) {
-        throwIfInvalidFieldOperation(payload, ADD_OP);
+        throwIfInvalidFieldOperation(payload, ADD_OP, REPLACE_OP);
         throwIfNullFieldValue(payload.get(FIELD_VALUE));
     }
 
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -9,7 +9,6 @@ import uk.gov.pay.connector.gatewayaccount.model.EmailCollectionMode;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.stream.Collectors;
 
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
@@ -251,15 +251,17 @@ public class GatewayAccountResource {
             gatewayAccount.setCredentials(updatedCredentials);
             return Response.ok().build();
         }).orElseGet(() -> notFoundResponse(format("The gateway account id '%s' does not exist", gatewayAccountId)));
-    } 
-    
+    }
+
     @GET
     @Path("/v1/frontend/accounts/{accountId}/gateway-merchant-id")
     @Produces(APPLICATION_JSON)
     public Response getGatewayMerchantId(@PathParam("accountId") Long gatewayAccountId) {
         return gatewayDao.findById(gatewayAccountId).map(gatewayAccount -> {
-            String gatewayMerchantId = gatewayAccount.getCredentials().get("gateway_merchant_id");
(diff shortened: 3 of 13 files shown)


src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
```diff
@@ -139,6 +139,21 @@ public void updateAndGetGatewayAccountWithWorldpayMerchantId() {
                 .then()
                 .statusCode(HttpStatus.SC_OK)
                 .body("gateway_merchant_id", is("1234abc"));
+
+        givenSetup().accept(JSON)
```
oswaldquek: @gidsg I think this block can be deleted?

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -238,6 +238,31 @@ public Response patchGatewayAccount(@PathParam("accountId") Long gatewayAccountI
                 .orElseGet(() -> Response.status(NOT_FOUND).build());
     }
 
+    //Deprecate when self service has functionality to update the gateway merchant id
+    @PATCH
+    @Path("/v1/frontend/accounts/{accountId}/gateway-merchant-id/{gatewayMerchantId}")
```
kbottla: Can we post 'gatewayMerchantId' in body rather than pass in URL ?
Requests are logged and we will log merchant Id if it is passed in URL.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -238,6 +238,33 @@ public Response patchGatewayAccount(@PathParam("accountId") Long gatewayAccountI
...
+    @Consumes(APPLICATION_JSON)
+    @Transactional
+    public Response updateGatewayAccountCredentialsWithGatewayMerchantId(@PathParam("accountId") Long gatewayAccountId, @PathParam("gatewayMerchantId") String gatewayMerchantId) {
+        return gatewayDao.findById(gatewayAccountId).map(gatewayAccount -> {
+            Map<String, String> credentials = gatewayAccount.getCredentials();
+            Map<String, String> updatedCredentials = new HashMap<>(credentials);
+            updatedCredentials.put("gateway_merchant_id", gatewayMerchantId);
```
kbottla: > Map<String, String> credentials = gatewayAccount.getCredentials();
            Map<String, String> updatedCredentials = new HashMap<>(credentials);
            updatedCredentials.put("gateway_merchant_id", gatewayMerchantId);

Is there any reason a new HashMap is created here instead of something like
```
gatewayAccount
.getCredentials()
.put("gateway_merchant_id", gatewayMerchantId);
```

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -238,6 +238,33 @@ public Response patchGatewayAccount(@PathParam("accountId") Long gatewayAccountI
                 .orElseGet(() -> Response.status(NOT_FOUND).build());
     }
 
+    //Deprecate when self service has functionality to update the gateway merchant id
+    @PATCH
```
kbottla: I suppose there is going to be some functionality in toolbox to use this endpoint ?

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -238,6 +238,33 @@ public Response patchGatewayAccount(@PathParam("accountId") Long gatewayAccountI
                 .orElseGet(() -> Response.status(NOT_FOUND).build());
     }
 
+    //Deprecate when self service has functionality to update the gateway merchant id
```
kbottla: Is there any story to update self service ? Would be good to add reference here so someone can find and deprecate/remove this endpoint

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -238,6 +238,33 @@ public Response patchGatewayAccount(@PathParam("accountId") Long gatewayAccountI
...
+            gatewayAccount.setCredentials(updatedCredentials);
+            return Response.ok().build();
+        }).orElseGet(() -> notFoundResponse(format("The gateway account id '%s' does not exist", gatewayAccountId)));
+    }
+
+    @GET
+    @Path("/v1/frontend/accounts/{accountId}/gateway-merchant-id")
+    @Produces(APPLICATION_JSON)
+    public Response getGatewayMerchantId(@PathParam("accountId") Long gatewayAccountId) {
+        return gatewayDao.findById(gatewayAccountId).map(gatewayAccount -> {
```
kbottla: This went back to old ways of connector code with all functionality (transactional boundary, db updates, error handling.. ) in this endpoint. Given that we may remove endpoint (and with all other endpoints in this resource following same approach), don't see any issue to keep this way. But if this is going to be stay for a while we will need to refactor code into appropriate service.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
```diff
@@ -121,6 +122,51 @@ String getMerchantId() {
...
+
+        givenSetup()
+                .get(ACCOUNTS_FRONTEND_URL + accountId + "/gateway-merchant-id")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .body("gateway_merchant_id", is("1234abc"));
+    }
+
+    @Test
+    public void getGatewayAccountWithInvalidAccount(){
```
kbottla: Can we name it similar to other tests method names `shouldReturn404IfGatewayAccountDoesNotExist` and to be descriptive

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
```diff
@@ -121,6 +122,51 @@ String getMerchantId() {
...
+        givenSetup()
+                .get(ACCOUNTS_FRONTEND_URL + "99999999" + "/gateway-merchant-id")
+                .then()
+                .statusCode(HttpStatus.SC_NOT_FOUND)
+                .body("message", is("The gateway account id '99999999' does not exist"));
+    }
+
+    @Test
+    public void getGatewayAccountWithoutMerchantId(){
```
kbottla: Same as above for naming..

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
kbottla: Path could be `/v1/accounts/{accountId}/gateway-merchant-id/{gatewayMerchantId}` if we are not going to invoke it through frontend/selfservice

kbottla (changes requested):

oswaldquek:
Do not merge. I think the requirements around this story need to be slightly updated.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: yeah we tried that. it doesn't work. i think it's something to do with JPA seeing the same object (Map) reference and thinks it doesn't need to update the credentials.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: I will remove this endpoint. I just spoke to Jon and it will be easier to return the gateway merchant id in the call to `/v1/frontend/charges/{chargeId}` so frontend doesn't have to make an extra api call.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: yup all good points. will fix these.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: toolbox and selfservice will use https://github.com/alphagov/pay-connector/blob/67e8a2378cb4152887e235b0b39f870fc5a45e53/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java#L232

i created this method to make it easier for me to manually update the gateway account with the gateway merchant id using `curl`. this is only temporary and when selfservice is updated to allow services to to update their account with their gateway merchant id this endpoint will be deleted.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: yes in here https://payments-platform.atlassian.net/browse/PP-4645. it's not prioritised yet though. there will probably be a few stories relating to self service.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: i fixed this by deleting the path ;)
`PATCH /v1/api/accounts/accountId` will be used instead

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: fixed

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: fixed

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
oswaldquek: this test is deleted

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
oswaldquek: this test is deleted

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -170,6 +170,12 @@ public String getGatewayName() {
         return credentials;
     }
 
+    @JsonProperty("gateway_merchant_id")
+    @JsonView(Views.FrontendView.class)
+    public String getGatewayMerchantId() {
+        return credentials.get("gateway_merchant_id");
```
gidsg, line 176: does this handle the possible null case?

oswaldquek: it doesn't matter as we usually return "null" if there's nothing. for example, 
```
    "gateway_account": {
        "version": 3,
        "requires3ds": false,
        "allowWebPayments": false,
        "notifySettings": null,
        "live": false,
        "gateway_account_id": 1,
        "payment_provider": "worldpay",
        "type": "test",
        "service_name": null,
        "analytics_id": null,

```

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
```diff
@@ -75,6 +81,11 @@ public void validatePatchRequest(JsonNode payload){
         }
     }
 
+    private void validateGatewayMerchantId(JsonNode payload) {
+        throwIfInvalidFieldOperation(payload, ADD_OP);
```
kbottla: I think we need `replace` op too for updating existing

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
```diff
@@ -62,13 +63,20 @@ public GatewayAccountResponse createGatewayAccount(GatewayAccountRequest gateway
...
     }
 
 
     private final Map<String, BiConsumer<PatchRequest, GatewayAccountEntity>> attributeUpdater =
             new HashMap<String, BiConsumer<PatchRequest, GatewayAccountEntity>>() {{
+                put(CREDENTIALS_GATEWAY_MERCHANT_ID,
+                        (gatewayAccountRequest, gatewayAccountEntity) -> {
+                            Map<String, String> credentials = gatewayAccountEntity.getCredentials();
+                            Map<String, String> updatedCredentials = new HashMap<>(credentials);
+                            updatedCredentials.put("gateway_merchant_id", gatewayAccountRequest.valueAsString());
```
kbottla, line 76: We need to check if gateway_merchant_id exists and update - if we allow `replace` op

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
```diff
@@ -57,6 +60,9 @@ public void validatePatchRequest(JsonNode payload){
             throw new ValidationException(Collections.singletonList(format("Operation [%s] not supported for path [%s]", FIELD_OPERATION, path)));
         
         switch (path) {
+            case CREDENTIALS_GATEWAY_MERCHANT_ID:
```
kbottla, line 64: I would like to see tests for this in `GatewayAccountRequestValidatorTest`

kbottla (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
oswaldquek: do we really need tests in `GatewayAccountRequestValidatorTest` as the positive test is covered by `GatewayAccountFrontendResourceITest. updateGatewayAccountWithGatewayMerchantId`? also, looking at `GatewayAccountRequestValidatorTest` there are only two positive tests `shouldNotThrow_whenAllValidationsPassed` and `shouldNotThrow_whenPathIsValidEmailCollectionMode` which suggests it's overkill to test everything at this unit test level as the rest are tested in the integration tests.

i'm also not convinced about testing for the negative as the validation of the gateway merchant id uses `throwIfInvalidFieldOperation` and `throwIfNullFieldValue` which are already tested.

kbottla: I am happy about the positive test as it is covered in Integration test but we don't cover negative cases anywhere.

Although `throwIfInvalidFieldOperation` and `throwIfNullFieldValue` are tested, the method `validateGatewayMerchantId` itself is not. It means we can change it to anything without detected.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
oswaldquek: I've created a test in `GatewayMerchantIdITest.replaceGatewayMerchantId`

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
oswaldquek: i've created a negative test in `GatewayAccountRequestValidatorTest.shouldThrowForInvalidMerchantGatewayIdPatchRequest`

kbottla: 👍

kbottla (approved):
LGTM! You might want to update PR description since it doesn't reflect current state of changes

PR #1041, opened 2019-02-08
PP-4790 Bump libraries

georgeracu:
## WHAT
- Added pay-java-commons with fix for NPE in XRaySessionFilter
- Not all the libraries have been upgraded as Liquibase and JDBC driver are throwing exceptions locally for missing extensions that need to be installed via db migrations. These should be fixed in a future PR. 

Edit: I added `postgresql` to this commit, as the error is from Liquibase.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -7,13 +7,17 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <dropwizard.version>1.3.8</dropwizard.version>
+        <liquibase.version>3.3.5</liquibase.version>
         <wiremock.version>2.21.0</wiremock.version>
-        <eclipselink.version>2.7.3</eclipselink.version>
-        <guice.version>4.1.0</guice.version>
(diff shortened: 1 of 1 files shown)


heathd:
when you say 'jdbc driver' I assume your referring to `org.postgresql:postgresql` ?

can you elaborate on the errors? are you sure the postgresql library causes this? I know that there was previously an error with upgrading liquibase due to the way it handles checksums on csv fixtures (see commit 82368cebd88e33734f0a75646eb728638e1acf4f)

georgeracu:
Yes, I was checking both `postgresql` and `liquibase` and the exception comes from `liquibase`. It matches with the description from the commit specified. This is a known bug in Liquibase core https://liquibase.jira.com/browse/CORE-3287

heathd (dismissed):
would be good to describe the version number changes in the commit message as well eg.

jooq 3.11.9->3.11.9

etc

but that's not a blocker

rhowe-gds (approved):

PR #1042, opened 2019-02-08
PP-4802 Support including reference in email custom paragraph

alexbishop1:
When sending a payment confirmation email, make it so that every instance of “$reference” in the custom paragraph is replaced by the actual payment reference.

For example, if the custom paragraph is “Have your reference number $reference to hand when you call” and the reference for the payment is “123”, the final email will contain “Have your reference number 123 to hand when you call”.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/usernotification/service/UserNotificationServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/usernotification/service/UserNotificationServiceTest.java
@@ -13,6 +13,7 @@ import org.mockito.junit.MockitoJUnitRunner;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.app.ExecutorServiceConfig;
 import uk.gov.pay.connector.app.NotifyConfiguration;
+import uk.gov.pay.connector.charge.model.ServicePaymentReference;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.model.domain.ChargeEntityFixture;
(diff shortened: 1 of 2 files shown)


46bit:
The build may fail because I had to reboot CI. Re-run it if it goes red.

kbottla (dismissed):
LGTM!

src/test/java/uk/gov/pay/connector/usernotification/service/UserNotificationServiceTest.java
```diff
@@ -114,6 +115,43 @@ public void shouldSendPaymentConfirmationEmailIfEmailNotifyIsEnabled() throws Ex
...
+        personalisation.put("description", "This is a description");
+        personalisation.put("serviceName", "MyService");
+        personalisation.put("customParagraph", "^ Here’s the ref: 123$000. Here it is again: abc123$000def. And to end: 123$000");
+        personalisation.put("amount", "5.00");
+        personalisation.put("corporateCardSurcharge", "");
+
+        when(mockNotifyClient.sendEmail(mockNotifyConfiguration.getEmailTemplateId(),
+                charge.getEmail(),
+                personalisation,
+                null)).thenReturn(mockNotificationCreatedResponse);
```
stephencdaly, line 154: Would it be clearer to use an argument captor for the personalisation here and then do assertions on that?

alexbishop1: Yes, I think it would.

I’ll add another commit to change this test and the others that use the same pattern to use argument captors.

alexbishop1: Decided not to do this in the end: since the method actually returns something (is non-`void`) we’d still have to stub it and generally we don’t want to stub and verify the same call.

alexbishop1:
As discussed, I didn’t make the change to use argument captors but I did move a load of repeated stubbing into the `@Before` method, which reduces the length of this test class substantially.

stephencdaly (approved):
Looks good!

PR #1044, opened 2019-02-11
PP-4560 Add gateway_accounts_stripe_setup connector database table

alexbishop1:
It’s OK to add the index non-concurrently because the table won’t have any rows yet.


stephencdaly (dismissed):
Looks good 🏆

stephencdaly (approved):
Even better 🥇

PR #1045, opened 2019-02-11
PP-4779: Add flags for Google/Apple Pay

oswaldquek:
We want the option of allowing services to turn Google/Apple Pay on/off
independently.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -957,4 +957,32 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add allow_apple_pay and allow_google_pay to gateway_accounts table" author="">
+        <addColumn tableName="gateway_accounts">
+            <column name="allow_apple_pay" type="boolean">
+            </column>
+        </addColumn>
+        <addColumn tableName="gateway_accounts">
+            <column name="allow_google_pay" type="boolean">
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="add default to allow_apple_pay and allow_google_pay" author="">
+        <addDefaultValue tableName="gateway_accounts" columnName="allow_apple_pay" defaultValue="false"/>
+        <addDefaultValue tableName="gateway_accounts" columnName="allow_google_pay" defaultValue="false"/>
+    </changeSet>
+
+    <changeSet id="update allow_apple_pay with default value" author="">
+        <update tableName="gateway_accounts">
+            <column name="allow_apple_pay" value="false"/>
+        </update>
+    </changeSet>
+
+    <changeSet id="update allow_google_pay with default value" author="">
+        <update tableName="gateway_accounts">
+            <column name="allow_google_pay" value="false"/>
+        </update>
+    </changeSet>
+
 </databaseChangeLog>


src/main/resources/migrations.xml
```diff
@@ -957,4 +957,15 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add allow_apple_pay and allow_google_pay to gateway_accounts table" author="">
```
kbottla, line 960: We need update scripts too or set defaults when adding new columns

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -110,6 +110,12 @@ public GatewayAccountEntity() {
     @Column(name = "allow_web_payments")
     private boolean allowWebPayments;
 
+    @Column(name = "allow_google_pay")
```
kbottla: Not exactly sure if this works as fields won't be available in environments without migration.

kbottla (changes requested):

src/main/resources/migrations.xml
```diff
@@ -957,4 +957,31 @@
...
+
+    <changeSet id="update allow_apple_pay and allow_google_pay with default values" author="">
+        <update tableName="gateway_accounts">
+            <column name="allow_apple_pay" value="false"/>
+            <where>allow_apple_pay IS NULL</where>
+        </update>
+        <update tableName="gateway_accounts">
+            <column name="allow_google_pay" value="false"/>
+            <where>allow_google_pay IS NULL</where>
+        </update>
```
alexbishop1, line 985: Probably safer to break this into two different changesets

alexbishop1 (changes requested):
One small change request

src/main/resources/migrations.xml
oswaldquek: done

src/main/resources/migrations.xml
```diff
@@ -957,4 +957,34 @@
...
+
+    <changeSet id="add default to allow_apple_pay and allow_google_pay" author="">
+        <addDefaultValue tableName="gateway_accounts" columnName="allow_apple_pay" defaultValue="false"/>
+        <addDefaultValue tableName="gateway_accounts" columnName="allow_google_pay" defaultValue="false"/>
+    </changeSet>
+
+    <changeSet id="update allow_apple_pay with default value" author="">
+        <update tableName="gateway_accounts">
+            <column name="allow_apple_pay" value="false"/>
+            <where>allow_apple_pay IS NULL</where>
```
kbottla: We don't need `IS NULL` condition as it is new column. Query is usually faster without additional clauses

src/main/resources/migrations.xml
```diff
@@ -957,4 +957,34 @@
...
+        </update>
+    </changeSet>
+
+    <changeSet id="update allow_google_pay with default value" author="">
+        <update tableName="gateway_accounts">
+            <column name="allow_google_pay" value="false"/>
+            <where>allow_google_pay IS NULL</where>
```
kbottla: Same as above for `IS NULL`

alexbishop1 (approved):

kbottla (approved):

PR #1049, opened 2019-02-12
PP-4638 Add email to charge when authorising wallet payment.

danworth:
- Add email from the wallet authorisation request to the charge post authorisation.


kbottla (commented):
Looks good to me. Can we also have a check in integration test too similar to https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/CardResourceAuthoriseITest.java#L94?

kbottla (approved):
As discussed (with @danworth ), WalletAuthoriseServiceTest covers if the charge data is set post persist call. Checking authorisation status and email assures that the details are persisted in database. I am happy with the test coverage related to googlepay endpoint

PR #1050, opened 2019-02-12
PP-4560 Make GatewayAccountResource not depend directly on DAO

alexbishop1:
Make the GatewayAccountResource not depend on the GatewyAccountDao directly and instead use the GatewayAccountService (which delegates to the GatewayAccountDao).

GatewayAccountResource no longer has GatewayAccountDao injected.

with @stephencdaly


mrlumbu (approved):
Looks good. I strongly approve of using services rather than DAOs. 
There's also a bit of sensible renaming which is 👍

PR #1051, opened 2019-02-12
PP-4560 Add GET /v1/api/accounts/ACCOUNT_ID/stripe-setup

alexbishop1:
Add GET endpoint to retrieve the bank account, responsible person and organisation details completed flags for Stripe gateway accounts in JSON format:

```json
{
  "bank_account": true,
  "responsible_person": false,
  "organisation_details": true
}
```

If the account does not exist or is not a Stripe account, return a 404.

with @stephencdaly

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountStripeSetup.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountStripeSetup.java
@@ -1,0 +1,33 @@
+package uk.gov.pay.connector.gatewayaccount.model;
+
+public class GatewayAccountStripeSetup {
+
+    private boolean bankDetailsCompleted = false;
+    private boolean responsiblePersonCompleted = false;
+    private boolean organisationDetailsCompleted = false;
+    
+    public boolean isBankDetailsCompleted() {
+        return bankDetailsCompleted;
+    }
+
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountStripeSetup.java
```diff
@@ -0,0 +1,45 @@
...
+public class GatewayAccountStripeSetup {
+
+    @JsonProperty("bank_account_details")
+    private boolean bankDetailsCompleted = false;
+
+    @JsonProperty("responsible_person")
+    private boolean responsiblePersonCompleted = false;
+
+    @JsonProperty("organisation_vat_number_company_number")
+    private boolean organisationDetailsCompleted = false;
```
DanailMinchev, line 19: Should we rename this to organisationVatNumberCompanyNumber? Or probably it is more clear for our internal system that `organisation_vat_number_company_number` is used as `organisationDetailsCompleted`?

alexbishop1: We’ve resolved this from the other side by renaming `"organisation_vat_number_company_number"` to `"organisation_details"` in the JSON.

DanailMinchev (approved):
Awesome, much clearer now 👍

PR #1052, opened 2019-02-12
PP-4779 extra tests and edge case handling

gidsg:
## WHAT
Extra test and functionality for PP-4779

Given a charge exists in our system
And the account associated with the charge has no gateway merchant id
When I call
GET /v1/frontend/charges/{chargeId}
Then the gateway_merchant_id is null

Given I have a worldpay merchant id for a service
And the service exists in our system
And the service has no gateway account credentials
When I PATCH the worldpay merchant id
Then a 400 BAD REQUEST is returned

## HOW 
See user stories above


gidsg:
@oswaldquek I think this is ready for review now. I believe the Codacy warnings are a false-positive as `rest-assured` wraps the assertions in the tests.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MerchantIdWithoutCredentialsException.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.pay.connector.gatewayaccount.exception;
+
+import uk.gov.pay.connector.util.ResponseUtil;
+
+import javax.ws.rs.BadRequestException;
+
+public class MerchantIdWithoutCredentialsException extends BadRequestException {
```
oswaldquek, line 7: normally we do `implements ExceptionMapper` (see ValidationExceptionMapper and UnsupportedOperationExceptionMapper for example) but it seems InvalidKeyException extends BadRequestException too so there doesn't seem to be any standard for this. (This isn't a request to change it, just an observation.)

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceAllowWebPaymentsITest.java
```diff
@@ -79,6 +79,20 @@ public void assertGooglePayPermission(boolean setAllowGooglePayFlag, boolean set
...
+    public void assertBadRequestResponseIfPatchingMerchantIdWithNoGatewayAccountCredentials() throws JsonProcessingException {
+        String accountIdWithoutGatewayAccountCredentials = extractGatewayAccountId(createAGatewayAccountFor(testContext.getPort(), "worldpay"));
+        String payload = new ObjectMapper().writeValueAsString(ImmutableMap.of("op", "add",
+                "path", "credentials/gateway_merchant_id",
+                "value", "1234abc"));
+        given().port(testContext.getPort()).contentType(JSON)
+                .body(payload)
+                .patch("/v1/api/accounts/" + accountIdWithoutGatewayAccountCredentials)
+                .then()
+                .assertThat()
```
oswaldquek: you don't need this `assertThat` line

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceAllowWebPaymentsITest.java
```diff
@@ -79,6 +79,20 @@ public void assertGooglePayPermission(boolean setAllowGooglePayFlag, boolean set
...
+                .statusCode(HttpStatus.SC_BAD_REQUEST);
```
oswaldquek, line 95: could you also assert the error message `"Account Credentials are required to set a Gateway Merchant ID"`? an easy way of printing the actual returned json body is to use `.log().body()` after the `.then()` part. 

i think the assertion will be `.path("message", is("Account Credentials are required to set a Gateway Merchant ID"))`

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MerchantIdWithoutCredentialsException.java
gidsg: OK, I took this style from https://github.com/alphagov/pay-connector/blob/a4f6a67778691ff82fc907d2b7830dfdc2aa8ff6/src/main/java/uk/gov/pay/connector/wallets/applepay/InvalidKeyException.java

oswaldquek (approved):

PR #1053, opened 2019-02-12
PP-4791 Add length validation

georgeracu:
## WHAT
- Added length validation for address fields and cardholder. This
validation is valid for the values that we receive in the auth request
- Added annotation type definition for the custom constraint validator
    for AuthCardDetails. This will work with annotation @Valid, instead of
    calling the validator by hand every time we try to map AuthCardDetails
    from external payload
    - Constraint violations are now intercepted by javax and returned into
    an array called `message` which is the default format. This requires a
    small change in the way we assert in tests, as we need to assert that
    the value is present in the array.
    - Small code style changes: in-lined local variables, made default constructor
    private to signal that this class should not be instantiated, removed
    exception declaration from method signature

Behaviour noticed on frontend:
- When `connector` returns status 400, frontend reloads the Card Details page with empty fields. The payment can be resumed and competed successfully.


heathd:
This looks good to me. We still need to improve how frontend reports the validation errors, but we'll raise a separate story for that.

heathd (approved):

PR #1059, opened 2019-02-13
PP-4512 Fix NPE when getting mapped status

mrlumbu:
ePDQ can respond with a parseable respnse with an unknown status.
Rather than pass nulls up the stack, I have made the status mapper
throw an error, which at the moment is caught and dealt with correctly by ChargeExpiryService. The error handling and passing can be
improved, but this commit ensures Expiry sweep works in all cases.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/report/model/GatewayStatusComparison.java
+++ b/src/main/java/uk/gov/pay/connector/report/model/GatewayStatusComparison.java
@@ -29,7 +29,8 @@ public final class GatewayStatusComparison {
     private GatewayStatusComparison(ChargeEntity charge, ChargeQueryResponse gatewayQueryResponse) {
         chargeId = charge.getExternalId();
         payStatus = ChargeStatus.fromString(charge.getStatus());
-        gatewayStatus = gatewayQueryResponse.getMappedStatus();
+        gatewayStatus = gatewayQueryResponse.getMappedStatus().isPresent() ?
+                gatewayQueryResponse.getMappedStatus().get() : null;
         rawGatewayResponse = gatewayQueryResponse.getRawGatewayResponse();
         this.charge = charge;
     }
--- a/src/main/java/uk/gov/pay/connector/report/model/GatewayStatusComparison.java
+++ b/src/main/java/uk/gov/pay/connector/report/model/GatewayStatusComparison.java
@@ -29,8 +29,7 @@ public final class GatewayStatusComparison {
     private GatewayStatusComparison(ChargeEntity charge, ChargeQueryResponse gatewayQueryResponse) {
         chargeId = charge.getExternalId();
         payStatus = ChargeStatus.fromString(charge.getStatus());
-        gatewayStatus = gatewayQueryResponse.getMappedStatus().isPresent() ?
-                gatewayQueryResponse.getMappedStatus().get() : null;
+        gatewayStatus = gatewayQueryResponse.getMappedStatus();
         rawGatewayResponse = gatewayQueryResponse.getRawGatewayResponse();
         this.charge = charge;
     }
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/report/model/GatewayStatusComparison.java
```diff
@@ -29,7 +29,8 @@ public static GatewayStatusComparison getEmpty(ChargeEntity charge) {
     private GatewayStatusComparison(ChargeEntity charge, ChargeQueryResponse gatewayQueryResponse) {
         chargeId = charge.getExternalId();
         payStatus = ChargeStatus.fromString(charge.getStatus());
-        gatewayStatus = gatewayQueryResponse.getMappedStatus();
+        gatewayStatus = gatewayQueryResponse.getMappedStatus().isPresent() ?
```
danworth: Should `getGatewayStatus` a little further down return `Optional<ChargeStatus>` since it is called by `canBeCancelled()` within `DiscrepancyService` as:
`!gatewayStatusComparison.getGatewayStatus().toExternal().isFinished() &&`
which has the potential to NPE if `getGatewayStatus()` returns the possible `null`?

mrlumbu: hmm, yeah. Probably needs a bit more thought

src/test/java/uk/gov/pay/connector/it/resources/ChargeExpiryResourceEpdqITest.java
```diff
@@ -57,6 +57,30 @@ public void shouldExpireWithGatewayIfExistsInCancellableStateWithGatewayEvenIfCh
         verifyPostToPath(String.format("/epdq/%s", ROUTE_FOR_QUERY_ORDER));
     }
 
+    @Test
+    public void shouldHandleCaseWhereEpdqRespondsWithUnknownStatus() {
```
danworth, line 61: 👍

src/test/java/uk/gov/pay/connector/charge/service/EpdqChargeExpiryServiceTest.java
```diff
@@ -0,0 +1,4 @@
+package uk.gov.pay.connector.charge.service;
+
+public class EpdqChargeExpiryServiceTest {
```
danworth: This is empty and unused, I assume its no longer needed and should not be committed, sorry if I'm missing something.

danworth (changes requested):
I think there's a test file that can be removed (probably a left over from the first approach). 

My only musing is that using the unchecked IllegalArgumentException where checked exceptions are usually appropriate for predictable, but unpreventable errors that are reasonable to recover from. The unchecked doesn't force the user of `map` to consider the possible exception result (no `throws` on the method) but then its internal and perhaps we'd expect them to at least look at the method before using it :)
There's lot of debate over when and which to use so I'll leave it to your better judgement and I'm happy with which ever you decide (I think I'm mentioning it hear as I'm interested in your opinion, perhaps later over a cup of tea).

mrlumbu:
Yes, I did consider making it a checked exception, and maybe it should be, although it is worth reading Effective Java on this point - he makes quite a good case for using checked exceptions very rarely. I think a better thing to do would be to handle the `IllegalArgumentException` within `EpdqPaymentProvider`, and wrap it into one of the errors it throws. I would quite like to get this fix out then refactor if that's ok, as this is fairly important to fix.

mrlumbu:
And yes, we should definitely discuss in more detail over some 🍵 
😃

danworth (approved):
Looks good.

PR #1060, opened 2019-02-13
PP-4560 s/GatewayAccountStripeSetup/StripeAccountSetup/g

alexbishop1:
Because `GatewayAccountStripeSetupTaskEntity` is way too long.

with @stephencdaly


DanailMinchev (approved):
LGTM 👍

PR #1063, opened 2019-02-14
PP-4560 Make PatchRequest less gateway account-specific

alexbishop1:
Rename `PatchRequest` to `GatewayAccountPatchRequest`.

Move constants for `"op"`, `"path"` and `"value"` from `GatewayAccount` to new `JsonPatchKeys` interface.

Move stringly-typed `"op"` values (`"add"`, `"replace"`, `"remove"`) to new `JsonPatchOp` enum.

with @stephencdaly


DanailMinchev (approved):
LGTM 👍

PR #1066, opened 2019-02-14
PP-4560 Validation for Stripe setup JSON Patch payload

alexbishop1:
Add validation for JSON Patch payload used to set the Stripe account setup bank details, responsible person and organisation details flags.

with @stephencdaly


DanailMinchev (approved):
LGTM 👍

PR #1067, opened 2019-02-14
PP-4779 add format validation for worldpay merchant ID

gidsg:
## WHAT
add format validation for worldpay merchant ID

## HOW 
PATCH'ing an invalidly formatted merchant ID should cause a 400 response

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -106,7 +106,7 @@ public class GatewayAccountRequestValidator {
     }
 
     private void throwIfNotValidMerchantId(JsonNode valueNode, String field) {
-        boolean isValidWorldpayMerchantId = valueNode.textValue().matches("[0-9a-f]{15,50}");
+        boolean isValidWorldpayMerchantId = valueNode.textValue().matches("[0-9a-f]{15}");
         if(!isValidWorldpayMerchantId) {
             throw new ValidationException(Collections.singletonList(format("Field [%s] value [%s] does not match that expected for a Worldpay Merchant ID; characters should be within range [0-9a-f]", field, valueNode.asText())));
         }
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -92,7 +92,7 @@ public class GatewayAccountRequestValidator {
                 break;
         }
     }
- 
+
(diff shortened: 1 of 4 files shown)


gidsg:
Closing as we don't have confirmation of the format from WorldPay and length constraints do not apply as we are persisting this as JSON.

gidsg:
@danworth are you able to review this now we have confirmation from WorldPay that this is the correct format?

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
```diff
@@ -93,17 +93,25 @@ public void validatePatchRequest(JsonNode payload){
...
 
     private void throwIfBlankFieldValue(JsonNode valueNode) {
         if (isBlank(valueNode.asText())) 
             throw new ValidationException(Collections.singletonList(format("Field [%s] cannot be empty", FIELD_VALUE)));
     }
 
+    private void throwIfNotValidMerchantId(JsonNode valueNode, String field) {
+        boolean isValidWorldpayMerchantId = valueNode.textValue().matches("[0-9a-f]{15}");
+        if(!isValidWorldpayMerchantId) {
+            throw new ValidationException(Collections.singletonList(format("Field [%s] value [%s] does not match that expected for a Worldpay Merchant ID; characters should be within range [0-9a-f]", field, valueNode.asText())));
```
danworth: Should we add that it must be 15 characters long too the more info we can provide the easier it might make for someone to realise their error.

danworth (changes requested):
Minor point about whether the validation message should include something around the expected value should be 15 characters long. 

The other thing I'm wondering is whether this validation should actually/also be done in Selfservice. If only here then we'll need some way of the error information getting back to the user so they can correct which leads me to think that validation closer to the user/source of input might be more appropriate/ also necessary?

src/test/java/uk/gov/pay/connector/it/resources/GatewayMerchantIdITest.java
```diff
@@ -33,16 +33,16 @@
     @Test
     public void replaceGatewayMerchantId() throws Exception {
         String accountId = setUpAccount();
-        addGatewayMerchantIdToAccount(accountId, "1234abc");
-        replaceGatewayMerchantId(accountId, "abc1234");
-        assertGatewayMerchantIdOnCharge(accountId, "abc1234");
+        addGatewayMerchantIdToAccount(accountId, "4d6ea2a35ab089c");
+        replaceGatewayMerchantId(accountId, "4d6ea2a35ab089c");
```
danworth: Sorry for being picky, but should we use a different `gatewayMerchantId` for the replace to make sure the replace actually worked (this test could pass if the add succeeds yet the replace is ineffective).

danworth (approved):
Looks good.

PR #1068, opened 2019-02-14
PP-4832: Upgrade postgres for tests to 9.6.12

rhowe-gds:


rhowe-gds:
Need to update pay-java-commons as well

danworth (approved):

PR #1070, opened 2019-02-15
Make PaymentGatewayStateTransitions fully static

rhowe-gds:
❗️This PR contains an opinion

This singleton class only contains fixed data and takes no inputs. We can make
it fully static and simplify both callers and tests.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/command/StateTransitionGraphVizRenderer.java
+++ b/src/main/java/uk/gov/pay/connector/command/StateTransitionGraphVizRenderer.java
@@ -11,12 +11,6 @@ import java.util.StringJoiner;
  * Render a graph - either install graphviz or use http://www.webgraphviz.com/
  */
 public class StateTransitionGraphVizRenderer {
-    private final PaymentGatewayStateTransitions transitions;
-
-    public StateTransitionGraphVizRenderer(PaymentGatewayStateTransitions transitions) {
-        this.transitions = transitions;
-    }
-
     public String toString() {
         StringJoiner s = new StringJoiner("\n", "", "\n");
         s.add("digraph PaymentGatewayStateTransitions {");
@@ -24,7 +18,7 @@ public class StateTransitionGraphVizRenderer {
         s.add("rankdir=TB overlap=false splines=true");
         s.add("node [style=filled shape=box color=\"0.650 0.200 1.000\"]");
 
-        transitions.allTransitions().forEach(edge ->
+        PaymentGatewayStateTransitions.allTransitions().forEach(edge ->
             s.add(formatEdge(edge))
         );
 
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/command/StateTransitionGraphVizRenderer.java
```diff
@@ -11,20 +11,14 @@
  * Render a graph - either install graphviz or use http://www.webgraphviz.com/
  */
 public class StateTransitionGraphVizRenderer {
-    private final PaymentGatewayStateTransitions transitions;
```
danworth, line 14: This might be point for consideration. Moving `PaymentGatewayStateTransitions` to be used in a static manner means we're no longer passing it as an object and we'll lock `StateTransitionsGraphVizRenderer` to only work with `PaymentGatewayStateTransitions` (rather than keeping composition pattern which would permit future reuse with other types of state graphs). I guess that depends on how we see the use of `StateTransitionsGraphVizRenderer` evolving (or not as the case may be).

alexbishop1: A long, long time ago, there was actually an interface for the state transitions, so in theory there could have been multiple implementations. In practice, of course, there was only one. Having an interface with just one implementation is usually a bit redundant, so at some point they were merged and the result became a singleton. It’s possible we may end decide to have multiple implementations again in future (perhaps for different types of payments e.g. delayed capture or zero pence payments) but until then I’m not too worried about the Graphviz renderer (which is really just a tool to aid development) being tied quite closely to a specific set of state transitions.

alexbishop1 (commented):
I think this change introduces advantages and disadvantages, but neither are particularly compelling.

The main advantage is that classes that want to use `PaymentGatewayStateTransitions` don’t need an instance of it. However, we use dependency injection already, so this doesn’t save us a huge amount.

The main disadvantage is that we can’t replace `PaymentGatewayStateTransitions` with another implementation (e.g. a mock) for testing. This couples our tests to the exact transitions, even where they’re not especially relevant. However, I don’t think this is a huge deal: it’s pretty easy to use an appropriate real transition in a test and this might actually improve confidence in the code because it could detect failures caused by accidentally using invalid transitions. As far as I can tell, we haven’t ever had a need to mock `PaymentGatewayStateTransitions`.

On balance, I think I’m slightly against this change, simply because using classes full of static methods isn’t very idiomatic Java.

src/main/java/uk/gov/pay/connector/command/StateTransitionGraphVizRenderer.java
rhowe-gds: My viewpoint is that this is effectively a .yaml file or a .dot file implemented in Java and as such is effectively static data with a few helper functions. Maybe a better thing would be to separate data and code more fully (i.e. go in the opposite direction to what this PR is attempting)

rhowe-gds:
Reviewers are persuasive!

PR #1072, opened 2019-02-15
PP-4560 Make JsonPatch.valueAs… methods not return null

alexbishop1:
Make `JsonPatch.valueAsString()` and `JsonPatch.valueAsLong()` throw exceptions rather than return null if the node cannot be converted to the appropriate type.

Remove `JsonNode.valueAsList()` because it’s unused.

This change is safe to make because the validation in `GatewayAccountRequestValidator` (which everything that is used to construct a `JsonPatchRequest` goes through) ensures that the nodes
of appropriate types as described below.

`valueAsString()`:

- `"credentials/gateway_merchant_id"` — checks not null
- `"allow_google_pay"` — checks `"true"` or `"false"`
- `"allow_apple_pay"` — checks `"true"` or `"false"`
- `"allow_web_payments"` — checks `"true"` or `"false"`
- `"email_collection_mode"` — checks if `EmailCollectionMode` variant

`valueAsLong()`:

- `"corporate_credit_card_surcharge_amount"` — checks +ve number
- `"corporate_debit_card_surcharge_amount"` — checks +ve number
- `"corporate_prepaid_credit_card_surcharge_amount"` — checks +ve num
- `"corporate_prepaid_debit_card_surcharge_amount"` — checks +ve num

with @stephencdaly

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/api/jsonpatch/JsonPatchRequest.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/api/jsonpatch/JsonPatchRequest.java
@@ -5,12 +5,9 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import java.io.IOException;
-import java.util.List;
 import java.util.Map;
 
-import static com.google.common.collect.Lists.newArrayList;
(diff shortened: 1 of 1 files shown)


SandorArpa (approved):
LOKTM

PR #1078, opened 2019-02-18
PP-4860: Log successful/unsuccessful Google payments

oswaldquek:
At present we aren't able to introduce smoke tests for Google payments, so we
need to introduce some monitoring. These log messages are inteded to be
searched for by our logging provider so as to build up some form of metric over
time.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
@@ -22,6 +22,8 @@ import uk.gov.pay.connector.wallets.model.WalletAuthorisationData;
 import java.time.format.DateTimeFormatter;
 import java.util.Optional;
 
+import static java.lang.String.format;
+
 public class WalletAuthoriseService {
     private static final DateTimeFormatter EXPIRY_DATE_FORMAT = DateTimeFormatter.ofPattern("MM/yy");
     private final CardAuthoriseBaseService cardAuthoriseBaseService;
@@ -49,23 +51,29 @@ public class WalletAuthoriseService {
             Optional<String> sessionIdentifier = Optional.empty();
             ChargeStatus chargeStatus = null;
             String responseFromPaymentGateway = null;
-
+            String requestStatus = "failure";
+            
             try {
                 operationResponse = authorise(charge, walletAuthorisationData);
                 Optional<BaseAuthoriseResponse> baseResponse = operationResponse.getBaseResponse();
-
+                
                 if (baseResponse.isPresent()) {
+                    requestStatus = "success";
                     chargeStatus = baseResponse.get().authoriseStatus().getMappedChargeStatus();
                     transactionId = cardAuthoriseBaseService.extractTransactionId(charge.getExternalId(), operationResponse);
                     sessionIdentifier = operationResponse.getSessionIdentifier();
                     responseFromPaymentGateway = baseResponse.toString();
-                } else operationResponse.throwGatewayError();
+                } else {
+                    operationResponse.throwGatewayError();
+                }
 
             } catch (GatewayErrorException e) {
                 chargeStatus = CardAuthoriseBaseService.mapFromGatewayErrorException(e);
                 responseFromPaymentGateway = e.getMessage();
                 operationResponse = GatewayResponse.GatewayResponseBuilder.responseBuilder().withGatewayError(e.toGatewayError()).build();
             }
+                
+            logMetrics(charge, operationResponse, requestStatus, walletAuthorisationData.getWalletType());
 
             processGatewayAuthorisationResponse(
                     charge.getExternalId(),
@@ -78,6 +86,21 @@ public class WalletAuthoriseService {
 
             return operationResponse;
         });
+    }
+
+    private void logMetrics(ChargeEntity chargeEntity,
+                            GatewayResponse<BaseAuthoriseResponse> operationResponse,
+                            String successOrFailure,
+                            WalletType walletType) {
+        
+        logger.info("{} authorisation {} - charge_external_id={}, payment provider response={}", 
+                walletType.toString(), successOrFailure, chargeEntity.getExternalId(), operationResponse.toString());
+        metricRegistry.counter(format("gateway-operations.%s.%s.%s.authorise.%s.result.%s", 
+                chargeEntity.getGatewayAccount().getGatewayName(), 
+                chargeEntity.getGatewayAccount().getType(),
+                chargeEntity.getGatewayAccount().getId(),
+                walletType.equals(WalletType.GOOGLE_PAY)? "google-pay" : "apple-pay",
+                successOrFailure));
     }
 
     @Transactional
--- a/src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
@@ -1,16 +1,22 @@
 package uk.gov.pay.connector.wallets;
 
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.LoggingEvent;
+import ch.qos.logback.core.Appender;
 import com.amazonaws.util.json.Jackson;
 import com.codahale.metrics.Counter;
-import com.codahale.metrics.MetricRegistry;
 import com.google.common.collect.ImmutableMap;
 import io.dropwizard.setup.Environment;
 import org.apache.commons.lang3.tuple.Pair;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
+import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -68,6 +68,10 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
                     responseFromPaymentGateway = baseResponse.toString();
                 } else operationResponse.throwGatewayError();
```
danworth: Doesn't prevent approval, and nothing you've done, but I think enclosing the `else` condition in `{}` makes it a little clearer what is and isn't part of the else (more when viewing the github generated diff, originally looked like the new log line was part of the `else`.
```
else {
    operationResponse.throwGatewayError();
}
```

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -68,6 +68,10 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
...
 
+                // This is logging intended to be searched for by our logging provider
+                logger.info("Google pay authorisation success - charge_external_id={}, payment provider response={}",
```
danworth: Since this is shared for wallets I think `Google` should be replaced with reference to `walletAuthorisationData.getWalletType()`

danworth (changes requested):
Looks good, only thing I believe needs changing is that this service is intended for all wallet types so need to replace the hard-coded `Google` with `walletAuthorisationData.getWalletType()`. Couple of other things if you're inclined to agree.

danworth:
Sorry Oz, I see that you've requested review from @heathd  so I'll defer to his better judgement, feel free to dismiss mine :)

oswaldquek:
@danworth @heathd this is ready for review. Thanks.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -80,6 +90,20 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
...
 
+    private void logMetrics(String chargeId,
+                            GatewayResponse<BaseAuthoriseResponse> operationResponse,
+                            boolean isSuccessful,
+                            String gatewayAccount,
+                            PaymentGatewayName paymentProvider, 
+                            WalletType walletType) {
+        
+        String webPaymentType = walletType.equals(WalletType.GOOGLE_PAY)? "google-pay" : "apple-pay";
+        String successOrUnsuccessful = isSuccessful? "success" : "unsuccessful";
```
danworth: Perhaps use `failure` instead of `unsuccessful` since someone might inadvertently capture the `unsuccessful` cases if searching for `success` with regex that's a little too lose, e.g. using `success` rather than something like `\bsuccess\b`.

Could consider creating an Enum for this kind of thing rather than a String to aid consistency across logs but not entirely necessary.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -80,6 +90,20 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
...
     }
 
+    private void logMetrics(String chargeId,
+                            GatewayResponse<BaseAuthoriseResponse> operationResponse,
+                            boolean isSuccessful,
+                            String gatewayAccount,
+                            PaymentGatewayName paymentProvider, 
+                            WalletType walletType) {
+        
+        String webPaymentType = walletType.equals(WalletType.GOOGLE_PAY)? "google-pay" : "apple-pay";
```
danworth: We already have log entries for wallets that use `WalletType.toString()` (`WalletService` line 32) and I think we should be consistent across all logs to help debugging. Perhaps this should use `WalletType.toString()` for the log entry.

For the metric name this works fine for now but should a should a third type be introduced, or we use a wallet metric elsewhere and want to ensure consistent naming it might be prudent to perhaps add a 'getMetricName()' to the `WalletType` enum?

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -59,12 +62,19 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
...
+                }
+                
+                logMetrics(chargeId, operationResponse, true, charge.getGatewayAccount().getGatewayName(), 
+                        charge.getPaymentGatewayName(), walletAuthorisationData.getWalletType());
 
             } catch (GatewayErrorException e) {
                 chargeStatus = CardAuthoriseBaseService.mapFromGatewayErrorException(e);
                 responseFromPaymentGateway = e.getMessage();
                 operationResponse = GatewayResponse.GatewayResponseBuilder.responseBuilder().withGatewayError(e.toGatewayError()).build();
+                logMetrics(chargeId, operationResponse, false, charge.getGatewayAccount().getGatewayName(),
```
danworth: Minor and only for consideration but could pass the `charge` rather than deconstructing some of its values into the method params to reduce the number of params.

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
```diff
@@ -88,9 +96,10 @@
                             anApplePayPaymentInfo().build())
                     .build();
 
+    private Appender<ILoggingEvent> mockAppender;
+
     @Before
-    public void setUpAppleAuthorisationService() {
-        mockMetricRegistry = mock(MetricRegistry.class);
+    public void setUp() {
```
danworth, line 102: 👍

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
```diff
@@ -107,6 +116,28 @@ public void setUpAppleAuthorisationService() {
...
+
+    private void setUpLogging() {
+        Logger root = (Logger) LoggerFactory.getLogger(WalletAuthoriseService.class);
+        mockAppender = mock(Appender.class);
+        root.setLevel(Level.INFO);
+        root.addAppender(mockAppender);
+    }
+
+    @Test
+    public void verifyLogging() throws Exception {
```
danworth, line 131: Nice to have the test for the logging.

danworth (changes requested):
Looks good, few minor things about consistency with wallet logging elsewhere and whether to use `failure` rather than `unsuccessful`.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
oswaldquek: Good point about searching the logs. Will change. Not sure about the Enum thing as this is just a small private method...

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
oswaldquek: fixed

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
oswaldquek: I've used `WalletType.toString()` in the log now.
I think metrics and the `WalletType` are two very separate things so am not too keen on a `getMetricName` method on the `WalletType`.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
danworth: That's ok 👍, FYI my intention wasn't for the benefit of just this simple private method but something that could be used whenever we're signalling success/failure in a log message to help consistency and avoid variations of `fail | failure | unsuccessful | not successful | etc..`, it would be something to point to and use consistently throughout the project. Happy as-is if you prefer.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
danworth: Sure, I guess we can't use `WalletType.toString()` for the metric name since it doesn't conform to the metric naming convention of using `-`? Not a huge problem since anyone introducing new metrics for wallet payments should look at the existing metrics and use the same name.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -59,12 +62,17 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
...
+                    operationResponse.throwGatewayError();
+                }
+                
+                logMetrics(charge, operationResponse, true, walletAuthorisationData.getWalletType());
 
             } catch (GatewayErrorException e) {
                 chargeStatus = CardAuthoriseBaseService.mapFromGatewayErrorException(e);
                 responseFromPaymentGateway = e.getMessage();
                 operationResponse = GatewayResponse.GatewayResponseBuilder.responseBuilder().withGatewayError(e.toGatewayError()).build();
+                logMetrics(charge, operationResponse, false, walletAuthorisationData.getWalletType());
```
danworth: Very very minor and no need to change if you don't wish. It might be more expressive for the param to be a String (or better an enum) of `success` or `failure` rather than a boolean which is later converted within the method, on first look is isn't clear what the value represents.

danworth (dismissed):
Looks good, thanks for considering my suggestions.

heathd (dismissed):
Looks good to me

heathd:
one other thought.. it's good to log gateway account ID but I think we should do that by adding it via MDC to every log line

oswaldquek:
@heathd i can't see how we're logging gateway account id via MDC in the code?

heathd (approved):
looks good

PR #1079, opened 2019-02-18
PP-4758: Goodbye to the session identifier bifunction

oswaldquek:
All the session identifier bifunction does is add a cookie to the http request. This is only useful for worldpay auth requests. Removing the `BiFunction<GatewayOrder, Builder, Builder> sessionIdentifier` field in GatewayClient will make it easier to deprecate StripeGatewayClient.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
@@ -13,13 +13,15 @@ import javax.ws.rs.ProcessingException;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.Invocation.Builder;
+import java.net.HttpCookie;
 import java.net.SocketTimeoutException;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-import java.util.function.BiFunction;
 
 import static java.lang.String.format;
+import static java.util.Collections.emptyList;
 import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
 import static javax.ws.rs.core.Response.Status.OK;
 import static uk.gov.pay.connector.gateway.util.AuthUtil.encode;
@@ -32,19 +34,22 @@ public class GatewayClient {
     private final Client client;
     private final Map<String, String> gatewayUrlMap;
     private final MetricRegistry metricRegistry;
-    private final BiFunction<GatewayOrder, Builder, Builder> sessionIdentifier;
 
-    public GatewayClient(Client client, 
-                         Map<String, String> gatewayUrlMap,
-                         BiFunction<GatewayOrder, Builder, Builder> sessionIdentifier, 
-                         MetricRegistry metricRegistry) {
+    public GatewayClient(Client client, Map<String, String> gatewayUrlMap, MetricRegistry metricRegistry) {
         this.gatewayUrlMap = gatewayUrlMap;
         this.client = client;
         this.metricRegistry = metricRegistry;
-        this.sessionIdentifier = sessionIdentifier;
     }
 
     public GatewayClient.Response postRequestFor(String route, GatewayAccountEntity account, GatewayOrder request) 
+            throws GenericGatewayErrorException, GatewayConnectionErrorException, GatewayConnectionTimeoutErrorException {
+        return postRequestFor(route, account, request, emptyList());
+    }
+    
+    public GatewayClient.Response postRequestFor(String route,
+                                                 GatewayAccountEntity account,
+                                                 GatewayOrder request,
+                                                 List<HttpCookie> cookies) 
             throws GenericGatewayErrorException, GatewayConnectionTimeoutErrorException, GatewayConnectionErrorException {
         String metricsPrefix = format("gateway-operations.%s.%s.%s", account.getGatewayName(), account.getType(), request.getOrderRequestType());
         javax.ws.rs.core.Response response = null;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -32,19 +34,22 @@
...
 
     public GatewayClient.Response postRequestFor(String route, GatewayAccountEntity account, GatewayOrder request) 
+            throws GenericGatewayErrorException, GatewayConnectionErrorException, GatewayConnectionTimeoutErrorException {
+        return postRequestFor(route, account, request, emptyList());
+    }
+    
+    public GatewayClient.Response postRequestFor(String route,
+                                                 GatewayAccountEntity account,
+                                                 GatewayOrder request,
+                                                 List<HttpCookie> cookies)
```
alexbishop1, line 52: Since there’s only ever a single cookie, do we really need a list?

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -85,14 +85,17 @@ public ChargeQueryResponse queryPaymentStatus(ChargeEntity charge) {
...
         return getWorldpayGatewayResponse(response);
     }
 
     @Override
     public Gateway3DSAuthorisationResponse authorise3dsResponse(Auth3dsResponseGatewayRequest request) {
         try {
-            GatewayClient.Response response = authoriseClient.postRequestFor(null, request.getGatewayAccount(), build3dsResponseAuthOrder(request));
+            GatewayOrder gatewayOrder = build3dsResponseAuthOrder(request);
+            GatewayClient.Response response = authoriseClient.postRequestFor(null, request.getGatewayAccount(), gatewayOrder,
+                    ImmutableList.of(new HttpCookie(WORLDPAY_MACHINE_COOKIE_NAME, gatewayOrder.getProviderSessionId().get())));
```
alexbishop1: Hmm, this is going to blow up if there isn’t a provider session ID for some reason. In practice, there should be, but I think I would prefer it the absence of a value meant the cookie wasn’t sent at all. This will probably still result in the authorisation failing but it will be handled by our existing code authorisation errors, which is fairly graceful and cleans things up.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -178,24 +179,28 @@ public void shouldIncludePaResponseIn3dsSecondOrder() throws Exception {
 
         Map<String, String> credentialsMap = ImmutableMap.of("merchant_id", "MERCHANTCODE");
         when(mockGatewayAccountEntity.getCredentials()).thenReturn(credentialsMap);
-        when(mockGatewayClient.postRequestFor(isNull(), any(GatewayAccountEntity.class), any(GatewayOrder.class)))
+        when(mockGatewayClient.postRequestFor(isNull(), any(GatewayAccountEntity.class), any(GatewayOrder.class), any(List.class)))
```
alexbishop1: I think `anyList()` is better because generics

alexbishop1 (changes requested):
This is a good changes that removes a useless abstraction (where we pretend there’s a so-generic-it’s-meaningless concept of a provider session ID, which is actually always a Worldpay cookie called “machine”) and makes both our code for Worldpay and the other gateways (which should have never had to know about this in the first place) simpler.

I have a couple of concerns about the implementation but the overall approach is great.

src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
mrlumbu: I think it makes sense to be able to pass a list of cookies into an HttpClient. For stripe we have to send an authorisation header (which again is only one, sure).

oswaldquek: i don't think it hurts to make this a generic case of passing in a list of cookies in this case

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: well, the `build3dsResponseAuthOrder` is a private method that sets a provider session id... maybe a better way would be to 
```
String providerSessionId = request.getProviderSessionId().orElse("")
GatewayClient.Response response = authoriseClient.postRequestFor(
    null, 
    request.getGatewayAccount(), 
    build3dsResponseAuthOrder(request),
    ImmutableList.of(new HttpCookie(WORLDPAY_MACHINE_COOKIE_NAME, providerSessionId)));
```
?

alexbishop1: That would still send a blank cookie if there is no provider session ID. I know from bitter past experience that blank cookies can behave very strangely. I think I’d prefer if it didn’t send the cookie at all. Something like this:

```java
List<HttpCookie> cookies = request.getProviderSessionId()
        .map(providerSessionId -> Collections.singletonList(new HttpCookie(WORLDPAY_MACHINE_COOKIE_NAME, providerSessionId)))
        .orElse(Collections.emptyList());
GatewayClient.Response response = authoriseClient.postRequestFor( null, request.getGatewayAccount(), build3dsResponseAuthOrder(request), cookies);
```

oswaldquek: I'll go with your suggestion 👍 
Although we're currently sending blank cookies!

oswaldquek:
thanks for the review @alexbishop1, i believe i've addressed all the issues.

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
alexbishop1: Really? We should stop doing that!

PR #1083, opened 2019-02-19
BAU test capture queue behaviour when DB is restarted

SandorArpa:
Testing the capture process scheduler


SandorArpa:
Closed as testing is finished

PR #1086, opened 2019-02-20
PP-4750: Have PerformanceReportDao return a Stream of results

rhowe-gds:
`PerformanceReportResource`'s transforms of the data returned from this DAO can
be much nicer if it's given a Stream to work with. Refactor the DAO to make
consumers' lives easier.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/report/dao/PerformanceReportDao.java
+++ b/src/main/java/uk/gov/pay/connector/report/dao/PerformanceReportDao.java
@@ -11,6 +11,7 @@ import javax.persistence.EntityManager;
 import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.List;
+import java.util.stream.Stream;
 
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity.Type.LIVE;
@@ -42,7 +43,7 @@ public class PerformanceReportDao extends JpaDao<PerformanceReportEntity> {
       .getSingleResult();
   }
 
-  public List<GatewayAccountPerformanceReportEntity> aggregateNumberAndValueOfPaymentsByGatewayAccount() {
+  public Stream<GatewayAccountPerformanceReportEntity> aggregateNumberAndValueOfPaymentsByGatewayAccount() {
     return entityManager
       .get()
       .createQuery(
@@ -61,10 +62,11 @@ public class PerformanceReportDao extends JpaDao<PerformanceReportEntity> {
         + " AND   g.type = :type"
         + " GROUP BY g.id"
         + " ORDER BY g.id ASC"
+              , GatewayAccountPerformanceReportEntity.class
       )
       .setParameter("status", CAPTURED.toString())
       .setParameter("type", LIVE)
-      .getResultList();
+      .getResultStream();
   }
 
   public PerformanceReportEntity aggregateNumberAndValueOfPaymentsForAGivenDay(ZonedDateTime date) {
--- a/src/main/java/uk/gov/pay/connector/report/resource/PerformanceReportResource.java
+++ b/src/main/java/uk/gov/pay/connector/report/resource/PerformanceReportResource.java
@@ -11,6 +11,8 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.QueryParam;
 import javax.ws.rs.core.Response;
 import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.MediaType.TEXT_PLAIN;
@@ -67,22 +69,16 @@ public class PerformanceReportResource {
     @Path("/v1/api/reports/gateway-account-performance-report")
     @Produces(APPLICATION_JSON)
     public Response getGatewayAccountPerformanceReport() {
-        HashMap<String, ImmutableMap<String, Object>> response = new HashMap<>();
-
-         performanceReportDao
-           .aggregateNumberAndValueOfPaymentsByGatewayAccount()
-           .forEach(
-             performance -> response.put(
-               performance.getGatewayAccountId().toString(),
-               ImmutableMap.of(
-                 "total_volume", performance.getTotalVolume(),
-                 "total_amount", performance.getTotalAmount(),
-                 "average_amount", performance.getAverageAmount(),
-                 "min_amount", performance.getMinAmount(),
-                 "max_amount", performance.getMaxAmount()
-               )
-             )
-           );
+         Map<String, Map<String, Object>> response = performanceReportDao.aggregateNumberAndValueOfPaymentsByGatewayAccount()
+                 .collect(Collectors.toMap(
+                         performance -> performance.getGatewayAccountId().toString(),
+                         performance -> ImmutableMap.of(
+                                 "total_volume", performance.getTotalVolume(),
+                                 "total_amount", performance.getTotalAmount(),
+                                 "average_amount", performance.getAverageAmount(),
+                                 "min_amount", performance.getMinAmount(),
+                                 "max_amount", performance.getMaxAmount()
+                 )));
 
          return ok().entity(response).build();
     }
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/pay/connector/report/dao/PerformanceReportDao.java
```diff
@@ -42,7 +43,7 @@ public PerformanceReportEntity aggregateNumberAndValueOfPayments() {
       .getSingleResult();
   }
 
-  public List<GatewayAccountPerformanceReportEntity> aggregateNumberAndValueOfPaymentsByGatewayAccount() {
+  public Stream<GatewayAccountPerformanceReportEntity> aggregateNumberAndValueOfPaymentsByGatewayAccount() {
     return entityManager
       .get()
       .createQuery(
```
danworth, line 49: You can use an overloaded method of `createQuery` to return a typedQuery by passing in the type as the second argument. Nothing that you've added but nice to fix it up whilst you're there. This will remove the unchecked cast assignment warning. e.g.
```
.createQuery(
        "SELECT new uk.gov.pay.connector.report.model.domain.GatewayAccountPerformanceReportEntity("
        + "   COALESCE(COUNT(c.amount), 0),"
        + "   COALESCE(SUM(c.amount),   0),"
        + "   COALESCE(AVG(c.amount),   0),"
        + "   COALESCE(MIN(c.amount),   0),"
        + "   COALESCE(MAX(c.amount),   0),"
        + "   g.id"
        + " )"
        + " FROM ChargeEntity c"
        + " JOIN GatewayAccountEntity g"
        + " ON c.gatewayAccount.id = g.id"
        + " WHERE c.status = :status"
        + " AND   g.type = :type"
        + " GROUP BY g.id"
        + " ORDER BY g.id ASC"
      , GatewayAccountPerformanceReportEntity.class)
```

danworth (changes requested):
Interesting PR! I agree returning the Stream makes processing it simpler. Interestingly Eclipse Link doesn't provide a more sophisticated implementation of `getResultStream` (unlike Hibernate) so there's no performance benefit since it'll load all results and just convert the collection to a stream. 
```
default Stream<X> getResultStream() {
        return getResultList().stream();
    }
```
Since this is an internal API and we don't need to accommodate users who might wish to iterate with something like a for-each-loop (Stream doesn't extend `Iterable` but going Collection -> Stream is simple) then this looks good; courtesy of Effective Java :)

I've requested a minor change to use the version of `createQuery` that returns a typed query to avoid the unchecked assignment warning and results in the above paramaterized version of `getResultStream` rather than the current raw Stream version.

Might be an interesting one for @alexbishop1 's coding standards, when to return Stream v Collection...

src/main/java/uk/gov/pay/connector/report/resource/PerformanceReportResource.java
```diff
@@ -67,22 +69,16 @@ public Response getDailyPerformanceReport(@QueryParam("date") String rawDate) {
...
+         Map<String, Map<String, Object>> response = performanceReportDao.aggregateNumberAndValueOfPaymentsByGatewayAccount()
+                 .collect(Collectors.toMap(
+                         performance -> performance.getGatewayAccountId().toString(),
+                         performance -> ImmutableMap.of(
+                                 "total_volume", performance.getTotalVolume(),
+                                 "total_amount", performance.getTotalAmount(),
+                                 "average_amount", performance.getAverageAmount(),
+                                 "min_amount", performance.getMinAmount(),
+                                 "max_amount", performance.getMaxAmount()
+                 )));
```
alexbishop1, line 81: These changes are nice but could also be achieved without changing `PerformanceReportDao` from returning `List`s to returning `Stream`s by adding `.stream()` just before the `.collect(…)`. Of course, if what `PerformanceReportDao` is always used as a `Stream`, changing the return types makes sense.

src/test/java/uk/gov/pay/connector/report/dao/PerformanceReportDaoITest.java
```diff
@@ -75,14 +75,21 @@ public void shouldAggregateNumberAndValueOfPaymentsForGatewayAccount() {
...
-        List<GatewayAccountPerformanceReportEntity> gatewayAccountPerformanceReportEntities = performanceReportDao.aggregateNumberAndValueOfPaymentsByGatewayAccount();
-        assertThat(gatewayAccountPerformanceReportEntities.size(), is(2));
-        assertThat(gatewayAccountPerformanceReportEntities.get(0).getAverageAmount(), is(closeTo(new BigDecimal("6"), ZERO)));
-        assertThat(gatewayAccountPerformanceReportEntities.get(0).getTotalAmount(), is(closeTo(new BigDecimal("12"), ZERO)));
-        assertThat(gatewayAccountPerformanceReportEntities.get(0).getTotalVolume(), is(2L));
-        assertThat(gatewayAccountPerformanceReportEntities.get(1).getAverageAmount(), is(closeTo(new BigDecimal("5"), ZERO)));
-        assertThat(gatewayAccountPerformanceReportEntities.get(1).getTotalAmount(), is(closeTo(new BigDecimal("15"), ZERO)));
-        assertThat(gatewayAccountPerformanceReportEntities.get(1).getTotalVolume(), is(3L));
+
+        Object[] gatewayAccountPerformanceReportEntities = performanceReportDao.aggregateNumberAndValueOfPaymentsByGatewayAccount().toArray();
```
alexbishop1: This might be an argument to not change the return types in `PerformanceReportDao`. 😀

Rather than converting to an array, I’d recommend `.collect(…)`ing to a `List`: `List`s are generally preferable and it will avoid all the casting later.

src/test/java/uk/gov/pay/connector/resources/GatewayAccountPerformanceReportResourceTest.java
```diff
@@ -62,7 +62,7 @@ public void setUp() {
     @Test
     public void noTransactionsPerformanceReportEntitySerializesCorrectly() {
         given(mockPerformanceReportDao.aggregateNumberAndValueOfPaymentsByGatewayAccount())
-                .willReturn(noTransactionsPerformanceReportEntity);
+                .willReturn(noTransactionsPerformanceReportEntity.stream());
```
alexbishop1: Can simply change the mocks to be mock `Stream`s rather than mock `List`s.

src/test/java/uk/gov/pay/connector/resources/GatewayAccountPerformanceReportResourceTest.java
```diff
@@ -27,10 +26,7 @@
     @Mock
     private PerformanceReportDao mockPerformanceReportDao;
 
-    @Mock
-    private List<GatewayAccountPerformanceReportEntity> noTransactionsPerformanceReportEntity;
-    @Mock
-    private List<GatewayAccountPerformanceReportEntity> someTransactionsPerformanceReportEntity;
+    private Stream<GatewayAccountPerformanceReportEntity> someTransactionsPerformanceReportEntity;
```
rhowe-gds, line 29: This was an abuse of `Mock`

src/test/java/uk/gov/pay/connector/resources/GatewayAccountPerformanceReportResourceTest.java
rhowe-gds: Turns out we can do even better than that

src/test/java/uk/gov/pay/connector/resources/GatewayAccountPerformanceReportResourceTest.java
```diff
@@ -42,25 +38,22 @@
...
-
-        someTransactionsPerformanceReportEntity.add(mockedGatewayAccountPerformanceReport);
+        ));
 
         resource = new PerformanceReportResource(mockPerformanceReportDao);
     }
 
     @Test
     public void noTransactionsPerformanceReportEntitySerializesCorrectly() {
+        Stream noTransactionsPerformanceReportEntity
```
danworth: Sorry to be picky can you add the paramaterization please, otherwise there's an unchecked assignment warning when using the raw type (even though ultimately its `Stream.empty()`) e.g.
`Stream<GatewayAccountPerformanceReportEntity>`

src/test/java/uk/gov/pay/connector/resources/GatewayAccountPerformanceReportResourceTest.java
danworth: 👍

danworth (changes requested):
Sorry, tiny point about avoiding the use of raw Generics, if you don't mind too much.

danworth (approved):
LGTM 👍

PR #1087, opened 2019-02-20
PP-4750: Support larger quantities of transactions in the summary report

rhowe-gds:
`summaryStatistics()` returns longs for the count of elements in a stream.
Rather than cast this to an int, we may as well return a `long` from our API.


rhowe-gds:
Not worth the hassle

PR #1088, opened 2019-02-20
PP-4868 add a try-catch block to avoid the scheduler blowing up

SandorArpa:
## WHAT
- add a try/catch block around the capture charges scheduler process to avoid the scheduler blowing up by an unchecked exception when the DB is not available

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
@@ -62,8 +62,12 @@ public class CaptureProcessScheduler implements Managed {
         logger.info("Scheduling CardCaptureProcess to run every {} seconds (will start in {} seconds)", interval, initialDelayInSeconds);
 
         scheduledExecutorService.scheduleAtFixedRate(() -> {
-            cardCaptureProcess.loadCaptureQueue();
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
```diff
@@ -62,8 +62,14 @@ public void start() {
         logger.info("Scheduling CardCaptureProcess to run every {} seconds (will start in {} seconds)", interval, initialDelayInSeconds);
 
         scheduledExecutorService.scheduleAtFixedRate(() -> {
-            cardCaptureProcess.loadCaptureQueue();
-            scheduleCaptureThreads();
+            // try-catch block was added to avoid the scheduler blowing up
```
mrlumbu: don't think we need a comment her e- it's fairly self explanatory, and if people want to understand why, they can read the commit

mrlumbu (changes requested):

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CaptureProcessScheduler.java
SandorArpa: 👍

mrlumbu (approved):

PR #1091, opened 2019-02-21
PP-4856 Move and tweak ChargeQueryResponse

mrlumbu:
ChargeQueryResponse is gateway agnostic so move to root of gateway package
Make getMappedStatus return an Optional, to acknowledge fact that we cannot
reliably mapped returned status to one of our statuses.


src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapper.java
```diff
@@ -37,7 +37,7 @@ public static ChargeStatus map(String epdqStatus) {
         if (epdqStatusMap.containsKey(epdqStatus)) {
             return epdqStatusMap.get(epdqStatus);
         }
-
-        throw new IllegalArgumentException(String.format("Unknown ePDQ status %s", epdqStatus));
+        
+        return null;
```
oswaldquek, line 41: can this return an Optional?

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqStatusMapper.java
mrlumbu: yes. I was gonna change this in the next PR if that's ok. At the moment the only place this method is used is in the ChargeQueryResponse, which wraps it in an Optional anyway

oswaldquek (approved):

PR #1092, opened 2019-02-21
PP-4806 Correct state name for Frontend wallet contract test.

danworth:
Correct the state name for the Frontend wallet contract test to match the updated value now being used by Frontend. This new value signals that the state should use an account set-up for sandbox.

@kbottla

### DIFF
--- a/src/test/java/uk/gov/pay/connector/pact/WalletApiContractTest.java
+++ b/src/test/java/uk/gov/pay/connector/pact/WalletApiContractTest.java
@@ -65,7 +65,7 @@ public class WalletApiContractTest {
         }
     }
 
-    @State("a charge exists with id testChargeId and is in state ENTERING_CARD_DETAILS.")
+    @State("a sandbox account exists with a charge with id testChargeId that is in state ENTERING_CARD_DETAILS.")
     public void aChangeExistsAwaitingAuthorisation(Map<String, String> params) {
         long gatewayAccountId = 666L;
         setUpGatewayAccount(gatewayAccountId);


src/test/java/uk/gov/pay/connector/pact/WalletApiContractTest.java
```diff
@@ -65,7 +65,7 @@ private void setUpGatewayAccount(long accountId) {
         }
     }
 
-    @State("a charge exists with id testChargeId and is in state ENTERING_CARD_DETAILS.")
+    @State("a sandbox account exits with a charge with id testChargeId that is in state ENTERING_CARD_DETAILS.")
```
mrlumbu: `exits` or `exists`?

danworth: lol I think we both know which one I mean. Fixed it up, thanks 👍

mrlumbu: I thought you might be matching a pre-existing typo in frontend pact, so I wasn't being totally facetious 😄

mrlumbu (approved):

src/test/java/uk/gov/pay/connector/pact/WalletApiContractTest.java
danworth: I was unwittingly matching my own pre-existing typo in frontend that I'd made five minutes before 🤦‍♂️

PR #1097, opened 2019-02-22
PP-4882 Refactor GatewayAccountDao to not return GatwayAccountResourceDTO

danworth:
GatewayAccountDao is a JPA dao paramaterized to type GatewayAccountEntity. This
changes the `list` and `listAll` methods to return `GatewayAccountEntity` and
remove the concern of converting `GatewayAccountEntity` to a `GatewayAccountResponseDTO`.
This is in the interest of maintaining a sensible separation of concerns, to
permit future refactoring of the numerous Gateway Account return type objects and move
towards a common means of fetching Gateway Accounts from the database and converting
to the required return type object (at the time of this change there are numerous ways
which causes confusion to new team members, especially when adding a new attribute
to Gateway Account.
- Change `list` and `listAll` methods to return a list of `GatewayAccountEntity`
- Add `.fromEntity` static method to `GatewayAccountResourceDTO`.
- Move responsibility for converting between the entity and DTO into the
`GatewayAccountService`.
- Update test and add new test for `.fromEntity` method.

## WHAT
Whilst helping Gideon with a story to introduce a new attribute to the Gateway Account model it was evident that the current numerous ways of fetching a gateway account and converting to the various return types is confusing. Broadly speaking at present we either (cannot guarantee this is exhaustive):
1. Use the dao to query and directly convert to `GatewayAccountResponseDTO` - which this PR addresses.
2. Use the dao to fetch a `GatewayAccountEntity` and return it.
3. As above but remove the password from the credentials.
4. Use the dao to fetch a `GatewayAccountEntity` then call `withoutCredtentials()` to convert to a `Map` before returning it (seems this should be a sub-type of its own).
5. Upon patching a gateway account it returns a `GatewayAccount` type.
6. The gateway account can also be retrieved on a `FrontendChargeResponse` as a `GatewayAccountEntity`.

Most of the confusion was centred on not knowing why the different types are returned (`GatewayAccount`, `GatwayAccountResourceDTO`, `GatewayAccountEntity`, `GatewayAccountEntity.withoutCredentials`)  exist and which ones to add a new attribute to, along with the process of fetching, transforming and returning them not being clear.

## HOW 
This PR removes the concern from the `GatewayAccountDao` of converting between `GatewayAccountEntity` to a `GatewayAccountResourceDTO` as a stepping stone towards creating a more common and understandable process surrounding gateway accounts. Following this PR if approved I would like to look at the various return types for a gateway account and understand how it can be simplified and in the case of `public Map<String, Object> withoutCredentials()` look to make this a proper sub-type, e.g. `GatewayAccountWithoutCredentials`.

Please let me know your thoughts both on this specific PR and whether you're in agreement that process/types around GatewayAccount should be made clearer. If you feel its ok as-is and this is just a result of us lacking experience with this area of the code-base then that's ok, please let me know 👍

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTOTest.java
+++ b/src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTOTest.java
@@ -1,0 +1,46 @@
+package uk.gov.pay.connector.gatewayaccount.model;
+
+import org.junit.Test;
+
+import java.util.Collections;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+
+
+public class GatewayAccountResourceDTOTest {
+
+    @Test
+    public void fromEntity() {
+        GatewayAccountEntity entity = new GatewayAccountEntity();
+        entity.setId(100L);
+        entity.setGatewayName("testGatewayName");
+        entity.setType(GatewayAccountEntity.Type.fromString("test"));
+        entity.setDescription("aDescription");
+        entity.setServiceName("aServiceName");
+        entity.setAnalyticsId("123");
+        entity.setCorporateDebitCardSurchargeAmount(200L);
+        entity.setCorporateCreditCardSurchargeAmount(300L);
+        entity.setCorporatePrepaidCreditCardSurchargeAmount(400L);
+        entity.setCorporatePrepaidDebitCardSurchargeAmount(500L);
+        entity.setAllowWebPayments(true);
+        entity.setAllowApplePay(false);
+        entity.setAllowGooglePay(true);
+        entity.setCredentials(Collections.emptyMap());
+        
+        GatewayAccountResourceDTO dto = GatewayAccountResourceDTO.fromEntity(entity);
+        assertThat(dto.getAccountId(), is(entity.getId()));
+        assertThat(dto.getPaymentProvider(), is(entity.getGatewayName()));
+        assertThat(dto.getType(), is(entity.getType()));
+        assertThat(dto.getDescription(), is(entity.getDescription()));
+        assertThat(dto.getServiceName(), is(entity.getServiceName()));
+        assertThat(dto.getAnalyticsId(), is(entity.getAnalyticsId()));
+        assertThat(dto.getCorporateCreditCardSurchargeAmount(), is(entity.getCorporateNonPrepaidCreditCardSurchargeAmount()));
+        assertThat(dto.getCorporateDebitCardSurchargeAmount(), is(entity.getCorporateNonPrepaidDebitCardSurchargeAmount()));
+        assertThat(dto.getCorporatePrepaidCreditCardSurchargeAmount(), is(entity.getCorporatePrepaidCreditCardSurchargeAmount()));
+        assertThat(dto.getCorporatePrepaidDebitCardSurchargeAmount(), is(entity.getCorporatePrepaidDebitCardSurchargeAmount()));
+        assertThat(dto.isAllowWebPayments(), is(entity.isAllowWebPayments()));
+        assertThat(dto.isAllowApplePay(), is(entity.isAllowApplePay()));
+        assertThat(dto.isAllowGooglePay(), is(entity.isAllowGooglePay()));
+    }
+}
--- a/src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
@@ -24,7 +24,7 @@ import java.util.Map;
 import java.util.Optional;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.core.Is.is;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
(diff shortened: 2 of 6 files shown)


danworth:
@alexbishop1 I think you understand our Gateway Account processes as well as anyone, would you mind taking a look if only for your opinion and words of wisdom rather than a detailed review please?

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTOTest.java
```diff
@@ -0,0 +1,36 @@
+package uk.gov.pay.connector.gatewayaccount.model;
+
+import org.junit.Test;
+
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.*;
```
alexbishop1: Wildcard imports are the devil’s work

danworth:
Please hold-off on a detailed review, since @gidsg  work will conflict with this and I'll have to rebase with his changes.

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
```diff
@@ -62,22 +64,26 @@ public void shouldGetGatewayAccount() {
...
     public void shouldGetAllGatewayAccounts() {
-        when(mockGatewayAccountDao.listAll()).thenReturn(Arrays.asList(mockGatewayAccountResourceDto1, mockGatewayAccountResourceDto2));
+        when(getMockGatewayAccountEntity1.getType()).thenReturn("test");
+        when(getMockGatewayAccountEntity2.getType()).thenReturn("test");
+        when(mockGatewayAccountDao.listAll()).thenReturn(Arrays.asList(getMockGatewayAccountEntity1, getMockGatewayAccountEntity2));
 
         List<GatewayAccountResourceDTO> gatewayAccounts = gatewayAccountService.getAllGatewayAccounts();
 
-        assertThat(gatewayAccounts, contains(mockGatewayAccountResourceDto1, mockGatewayAccountResourceDto2));
+        assertThat(gatewayAccounts, hasSize(2));
```
alexbishop1, line 74: In this test (and the other one you modified), it would be good to check that the `GatewayAccountResourceDTO` objects that are returned have been converted from the correct `GatewayAccountEntity` objects. Two ways I think of to this:

1. Stub `getServiceName()` on each mock `GatewayAccountEntity` with a unique name so we can verify that `getServiceName()` on the matching `GatewayAccountResourceDTO` object returns the same value.

1. Move the logic for converting from a `GatewayAccountEntity` to `GatewayAccountResourceDTO` out from a static method on `GatewayAccountResourceDTO` into a separate converter class. If this converter is itself well unit-tested, then we can just stub the call to a mock converter in these tests:

```java
when(mockGatewayAccountEntityToGatewayAccountResourceDTOConverter.convert(mockGatewayAccountEntity1).thenReturn(mockGatewayAccountResourceDTO1);
```

Having separate converters also feels a bit cleaner to me from a separation of concerns point of view (why should `GatewayAccountResourceDTO` even know `GatewayAccountEntity` exists?) but it does mean the converter has to be injected in various places and it isn’t generally the approach we use in connector.

Just because it’s Friday, I’m going to take a moment to wistfully wish we were using Rust, which has in-built support for conversions with its `From` and `Into` traits.

Anyway, I’d be happy if you just did the first one.

alexbishop1 (changes requested):
This is a good change.

You’re right that we have far too many different ways of representing gateway accounts (and to add insult to injury, some of them aren’t even proper types: they’re `Map`s!).

While it is often appropriate to have multiple objects to represent the same thing in different parts of the application (e.g. a separate type for a database entity and a JSON response object, which — because we like to screw things up in opposing ways — we notably don’t have in many places: witness the number of classes that have both JPA annotations _and_ Jackson annotations!), the proliferation of gateway account classes is too much. And the layering is all wrong: we certainly shouldn’t be returning REST DTOs from DAOs.

I like that this PR doesn’t try to fix the world and just makes a small, modest improvement. It’s good to do this incrementally to make the changes easier to follow and reduce the risk of introducing regressions.

I’ve left a couple of small comments about the specifics but I have no problems at all with the overall idea and the direction it takes us in.

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTOTest.java
danworth: Thanks, strange though I already have `count to use wildcard import = 99` so not sure why Intelij stuck it in there. I'll remove 👍

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
danworth: Yes this is a good idea, I've gone with option 1 to avoid the injection and its a little more aligned with other patterns elsewhere (although option 2 is probably the 'correct' way wrt to separation of concerns).

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTOTest.java
alexbishop1: I believe it’s because IntelliJ IDEA automatically adds the wildcard import when you tell it to generate a skeletal test class. You then have to optimise imports to have it replaced or removed.

danworth:
I've rebased with Gideon's changes and addressed Alex's comments so its ready to review whenever someone has time please?

alexbishop1 (approved):

PR #1099, opened 2019-02-22
PP-4856 Implement `queryPaymentStatus` for Worldpay

mrlumbu:
Adds ability to query worldpay for a payment status. This will have an affect both on ChargeExpiryService and on DiscrepancyService. 
The main change is in `WorldpayPaymentProvider`, and I have added tests for this.
I have also added  second overloaded method in `WorldpayGatewayResponseGenerator` that deserialises a worldpay response to an object of a type passed in. As a bonus it doesn't use raw types \o/

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
@@ -13,6 +13,7 @@ import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.chargeevent.dao.ChargeEventDao;
+import uk.gov.pay.connector.gateway.ChargeQueryResponse;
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -231,4 +238,30 @@ public void shouldSweepAndExpireCharges() throws Exception {
         assertThat(chargeEntityAwaitingCapture.getStatus(), is(ChargeStatus.EXPIRED.getValue()));
         assertThat(chargeEntityAuthorisationSuccess.getStatus(), is(ChargeStatus.EXPIRED.getValue()));
     }
+
+    @Test
+    public void shouldCancelChargeWithGatewayWhenChargeInPreAuthorisedStateAndExistsWithGateway() throws Exception {
```
oswaldquek, line 243: This seems like it's testing the same things as `ChargeExpiryResourceITest. shouldExpireWithGatewayIfExistsInCancellableStateWithGatewayEvenIfChargeIsPreAuthorisation`.

oswaldquek (approved):
Personally I think the integration test is enough but the existence of the unit test isn't a blocker for approving this PR.

PR #1100, opened 2019-02-22
PP-4819: add google pay and apple pay booleans to /frontend/accounts endpoints

gidsg:
## WHAT
Exposes `allow_apple_pay` and `allow_google_pay` in the `/v1/frontend/accounts` endpoints.
This is required to allow selfservice to GET the status of these booleans.

## HOW 
`/v1/frontend/accounts/{accountId}` should include `allow_apple_pay` and `allow_google_pay` booleans

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
@@ -70,7 +70,9 @@ public class GatewayAccountFrontendResourceITest extends GatewayAccountResourceT
                 .body("analytics_id", is(nullValue()))
                 .body("service_name", is(gatewayAccountPayload.getServiceName()))
                 .body("corporate_credit_card_surcharge_amount", is(250))
-                .body("corporate_debit_card_surcharge_amount", is(50));
+                .body("corporate_debit_card_surcharge_amount", is(50))
+                .body("allow_apple_pay", is(false))
+                .body("allow_google_pay", is(false));
     }
 
     @Test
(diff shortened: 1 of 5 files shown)


oswaldquek (approved):

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
```diff
@@ -124,7 +126,9 @@ public void shouldFilterGetGatewayAccountForExistingAccount() {
                 .body("accounts[0].service_name", is(gatewayAccountPayload.getServiceName()))
                 .body("accounts[0].corporate_credit_card_surcharge_amount", is(250))
                 .body("accounts[0].corporate_debit_card_surcharge_amount", is(50))
-                .body("accounts[0].allow_web_payments", is(false));
+                .body("accounts[0].allow_web_payments", is(false))
+                .body("accounts[0].allow_apple_pay", is(false))
+                .body("accounts[0].allow_google_pay", is(false));
```
danworth, line 131: I could be wrong here but worth validating with you just in case...

In this integration test all the newly added web payment related values are checked against `false` which is the default value for the `boolean` type used by `GatewayAccountEntity`. Unless I'm mistaken this could still pass if there's something wrong in setting these fields between fetching the db entry and returning it (e.g. if somehow it isn't set during some transformation between types and we're unwittingly checking & passing against the default value). Perhaps it would be prudent to use the `databaseTestHelper` to set one of them to `true` and assert that it actually returns as such? I think I'm conscious of this because I'm looking at refactoring the various return types and it would be useful to make sure any conversion between types doesn't introduce an issue by omitting a setting or miss-matching fields...

What do you think?

gidsg: Yes, we could do that, I will look at it in a separate PR.

PR #1102, opened 2019-02-22
PP-4819: test gateway account reads allow_apple_pay true boolean

gidsg:
## WHAT
Follow up to #1100 add test suggested in https://github.com/alphagov/pay-connector/pull/1100#discussion_r259318880

## HOW 
Test should pass


danworth (approved):
LGTM!

PR #1109, opened 2019-02-26
PP-4898 Remove withoutCredentials method on GatewayAccountEntity

danworth:
Remove the use of `GatewayAccountEntity.withoutCredentials()` within the end-point of `GET v1/api/accounts/{accountId}` in favour of the existing `GatewayAccountResourceDTO` to rationalise the number of return type objects and processes surrounding gateway accounts in Connector (also because its now a proper type).
- Add "email_notifications", "email_collection_mode" and "toggle_3ds" to `GatewayAccountResourceDTO`
so that this type returns the same information as the `Map` produced by `withoutCredentials`.
- This changes the type of the values contained within the response since `withoutCredentials`
was converting each type to String instead of using `boolean` for example.
- Where necessary update tests to assert on type boolean rather than expecting the String equivalent e.g. `"true"`.
- Update CardResource to use `GatewayAccountResourceDTO.fromEntity()` in place
of calling `withoutCredentials()`.

## WHAT
`GET v1/api/accounts/{accountId}` appears to only be used by the `pay-toolbox` within 
```
pay-toolbox//app/lib/pay-request/api_utils/connector.js:9:    return axiosInstance.get("/v1/api/accounts/${id}").then(utilExtractData)
```
SelfService uses the `v1/frontend/accounts/{accountId}` to GET accounts and only PATCHES to `v1/api/accounts/{accountId}`. Therefore I am reasonably confident that the change to returning boolean as boolean rather than wrapping in a String is safe.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
@@ -98,7 +98,7 @@ public class GatewayAccountResourceITest extends GatewayAccountResourceTestBase 
                 .get(ACCOUNTS_API_URL + gatewayAccountId)
                 .then()
                 .statusCode(200)
-                .body("toggle_3ds", is("true"));
+                .body("toggle_3ds", is(true));
     }
 
     @Test
@@ -151,7 +151,7 @@ public class GatewayAccountResourceITest extends GatewayAccountResourceTestBase 
                 .then()
                 .statusCode(200)
                 .body("payment_provider", is("sandbox"))
-                .body("gateway_account_id", is(String.valueOf(defaultTestAccount.getAccountId())))
+                .body("gateway_account_id", is(Math.toIntExact(defaultTestAccount.getAccountId())))
                 .body("type", is(TEST.toString()))
                 .body("description", is("a description"))
                 .body("analytics_id", is("an analytics id"))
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
```diff
@@ -151,7 +151,7 @@ public void shouldReturnAccountInformationForGetAccountById() {
                 .then()
                 .statusCode(200)
                 .body("payment_provider", is("sandbox"))
-                .body("gateway_account_id", is(String.valueOf(defaultTestAccount.getAccountId())))
+                .body("gateway_account_id", is(Math.toIntExact(defaultTestAccount.getAccountId())))
```
alexbishop1, line 154: Does doing `body("gateway_account_id", is(defaultTestAccount.getAccountId()))` not work for some reason?

alexbishop1 (approved):
Awesome consolidation!

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITest.java
danworth: Yeah, its a `long` so the "expected" value comes out at `123L` rather than just `123` (making it an int ensures it's just `123`). I looked at converting the "actual" value, i.e.  "gateway_account_id" to a Long but there doesn't seem to be a way to do it with arguments to the `.body` method of a ValidatableReponse.  I won't merge until tomorrow morning now so if there's a better way feel free to suggest it.

PR #1112, opened 2019-02-27
PP-4869 Change Apple Pay secret names to match names used by Apple.

danworth:
## WHAT YOU DID
We've decided to re-name the env variables relating to apply pay secrets so they match the naming convention used by Apple to avoid confusion when the time comes to renew them.


sfount (approved):

oswaldquek:
@danworth does there need to be a corresponding change in pay-infra?

danworth:
@oswaldquek yes, there is in this PR https://github.com/alphagov/pay-infra/pull/2011#event-2161430571.

PR #1114, opened 2019-02-28
PP-4870 Remove DB column `allow_web_payments`

kbottla:
## WHAT YOU DID

 - Drop column from gateway_accounts table as it is no longer used


alexbishop1 (approved):
👋

kbottla:
Thanks @alexbishop1 😄

PR #1115, opened 2019-02-28
PP 4910 return wallet type in charge response

danworth:
## WHAT YOU DID
Add wallet type to `ChargeResponse`
    - Add wallet type to `ChargeResponse` and `AbstractChargeResponseBuilder`.
    - Set wallet type when creating the charge response via `v1/api/accounts/{accountId}
    /charges/{chargeId}` and `v1/frontend/charges/{chargeId}`
    - Update tests

 Minor refactoring to simplify charge response and associated builder pattern.
 - Pass the charge response builder into the charge response constructors rather than deconstructing
 it which leads to complicated constructors prone to human error
 and duplication. This also simplifies adding new attributes to the various charge response
 types (which is why it was done prior to implementing the change above).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
@@ -248,9 +248,7 @@ public class ChargeResponse {
 
         @Override
         public ChargeResponse build() {
-            return new ChargeResponse(chargeId, amount, state, cardBrand, gatewayTransactionId, returnUrl, email,
-                    description, reference, providerName, createdDate, links, refundSummary, settlementSummary,
-                    cardDetails, auth3dsData, language, delayedCapture, corporateCardSurcharge, totalAmount);
+            return new ChargeResponse(this);
         }
     }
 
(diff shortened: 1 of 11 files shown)


danworth:
Note on Codacy prompt that `Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes.` which I agree with, however the existing format of the class does not follow this. I'm happy to fix it for the new field as well as existing but perhaps after this PR since this kind of re-structuring will make understanding the new changes much harder.

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -438,33 +441,36 @@ public boolean equals(Object o) {
...
+                Objects.equals(getAmount(), that.getAmount()) &&
+                Objects.equals(getState(), that.getState()) &&
+                Objects.equals(getCardBrand(), that.getCardBrand()) &&
+                Objects.equals(getGatewayTransactionId(), that.getGatewayTransactionId()) &&
+                Objects.equals(getReturnUrl(), that.getReturnUrl()) &&
+                Objects.equals(getEmail(), that.getEmail()) &&
+                Objects.equals(getDescription(), that.getDescription()) &&
+                Objects.equals(getReference(), that.getReference()) &&
+                Objects.equals(getProviderName(), that.getProviderName()) &&
                 Objects.equals(createdDate, that.createdDate) &&
```
kbottla: Is there any benefit to use getters instead of fields as they were before and `createdDate` not using getters. Using fields may improve readability IMHO

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -438,33 +441,36 @@ public boolean equals(Object o) {
...
+                Objects.equals(getTotalAmount(), that.getTotalAmount()) &&
+                getWalletType() == that.getWalletType();
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(dataLinks, chargeId, amount, state, cardBrand, gatewayTransactionId, returnUrl, email,
-                description, reference, providerName, createdDate, refundSummary, settlementSummary, auth3dsData,
-                cardDetails, language, delayedCapture, corporateCardSurcharge, totalAmount);
+        return Objects.hash(getDataLinks(), getChargeId(), getAmount(), getState(), getCardBrand(),
```
kbottla: Same as above for getters

kbottla (commented):
Looks good overall and nice refactoring using Builder. Just added couple of comments.

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
danworth: Good point and no I can't find anything concrete about an advantage of using getters. I must have `use getters` selected for auto generating hash and equals. I'll remove the use of getters since I agree its easier to read without.

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
danworth: 👍 fixed as above.

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -455,16 +458,17 @@ public boolean equals(Object o) {
                 Objects.equals(settlementSummary, that.settlementSummary) &&
                 Objects.equals(auth3dsData, that.auth3dsData) &&
                 Objects.equals(cardDetails, that.cardDetails) &&
+                language == that.language &&
```
kbottla, line 461: We could use Objects.equals(language, that.language) for consistency, but won't block on it.

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -455,16 +458,17 @@ public boolean equals(Object o) {
...
                 Objects.equals(corporateCardSurcharge, that.corporateCardSurcharge) &&
                 Objects.equals(totalAmount, that.totalAmount) &&
-                language == that.language;
+                walletType == that.walletType;
```
kbottla, line 464: Same as language for consistency

kbottla (approved):
LGTM!

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
kbottla: 👍

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
kbottla: 👍

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
danworth: Consistency is a good point. Intellij uses `==` for primitives to avoid unnecessary boxing and for enums I assume because it's equals is simply `return this==other;` and there is only ever one instance of each enum constant.

PR #1118, opened 2019-03-01
PP-4758: Add headers to GatewayClient

oswaldquek:
At the moment GatewayClient automatically sets credentials when making external
calls to payment providers:
https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java#L58-L60.
This will not work for Stripe as credentials are got from the
StripeGatewayConfig, and not the credentials in a GatewayAccountEntity (see
https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java#L281
as an example).

To be able to get rid of StripeGatewayClient, this commit makes GatewayClient
more generic by making it accept a Map of headers, and it is the responsibility
of callers to pass in an AUTHORIZATION header.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


kbottla (approved):
LGTM!

PR #1123, opened 2019-03-04
PP-4937 Add Stripe Account endpoint

DanailMinchev:
## WHAT YOU DID

Add `/v1/api/accounts/{accountId}/stripe-account` endpoint which can be used to query the `stripe_account_id` stored in `GatewayAccount` credentials.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeAccountResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeAccountResponse.java
@@ -1,0 +1,22 @@
+package uk.gov.pay.connector.gatewayaccount.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class StripeAccountResponse {
+
+    @JsonProperty("stripe_account_id")
+    private final String stripeAccountId;
+
+    public StripeAccountResponse(String stripeAccountId) {
+        this.stripeAccountId = Objects.requireNonNull(stripeAccountId);
+    }
+
+    public String getStripeAccountId() {
+        return stripeAccountId;
+    }
+
+}
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
@@ -1,0 +1,16 @@
+package uk.gov.pay.connector.gatewayaccount.service;
+
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.model.StripeAccountResponse;
+
+import java.util.Optional;
+
+public class StripeAccountService {
+
+    public Optional<StripeAccountResponse> buildStripeAccountResponse(GatewayAccountEntity gatewayAccountEntity) {
+        return Optional.ofNullable(gatewayAccountEntity.getCredentials())
+                .map(credentials -> credentials.get("stripe_account_id"))
+                .map(StripeAccountResponse::new);
+    }
+
+}
--- a/src/test/java/uk/gov/pay/connector/it/resources/StripeAccountResourceITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/StripeAccountResourceITest.java
@@ -1,0 +1,71 @@
+package uk.gov.pay.connector.it.resources;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import uk.gov.pay.connector.app.ConnectorApp;
+import uk.gov.pay.connector.it.dao.DatabaseFixtures;
+import uk.gov.pay.connector.junit.DropwizardConfig;
+import uk.gov.pay.connector.junit.DropwizardJUnitRunner;
+
+import java.util.Collections;
+
+import static org.hamcrest.Matchers.is;
+
+@RunWith(DropwizardJUnitRunner.class)
+@DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml")
+public class StripeAccountResourceITest extends GatewayAccountResourceTestBase {
+
+    private static final String STRIPE_ACCOUNT_ID = "acct_123example123";
+
+    @Test
+    public void getStripeAccountReturnsSuccessfulResponse() {
+        DatabaseFixtures.TestAccount testAccount = databaseFixtures
+                .aTestAccount()
+                .withPaymentProvider("stripe")
+                .withCredentials(Collections.singletonMap("stripe_account_id", STRIPE_ACCOUNT_ID))
+                .insert();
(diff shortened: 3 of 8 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeAccountResponse.java
```diff
@@ -0,0 +1,20 @@
+package uk.gov.pay.connector.gatewayaccount.model;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class StripeAccountResponse {
```
alexbishop1, line 9: I’d be very tempted to make this pseudo-immutable:

```java
public class StripeAccountResponse {

  @JsonProperty("stripe_account_id")
  private final String stripeAccountId;

  public StripeAccountResponse(String stripeAccountId) {
    this.stripeAccountId = Objects.requireNonNull(stripeAccountId)
  }

}
```

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
```diff
@@ -0,0 +1,24 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.model.StripeAccountResponse;
+
+import java.util.Optional;
+
+public class StripeAccountService {
+
+    public Optional<StripeAccountResponse> newStripeAccountResponse(GatewayAccountEntity gatewayAccountEntity) {
```
alexbishop1: Not convinced by the name… `getStripeAccountResponse`, maybe?

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
```diff
@@ -0,0 +1,24 @@
...
+        if (gatewayAccountEntity.getCredentials() == null ||
```
alexbishop1: Can `getCredentials()` ever actually return `null`?

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
```diff
@@ -0,0 +1,24 @@
...
+                gatewayAccountEntity.getCredentials().isEmpty() ||
```
alexbishop1: We don’t need the `isEmpty()` check; `get("stripe_account_id")` will return `null`, which the code handles

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
```diff
@@ -0,0 +1,24 @@
...
+                StringUtils.isEmpty(gatewayAccountEntity.getCredentials().get("stripe_account_id"))) {
```
alexbishop1: I wouldn’t worry about blank strings, just `null`

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
```diff
@@ -0,0 +1,24 @@
...
+            return Optional.empty();
+        }
+
+        StripeAccountResponse stripeAccountResponse = new StripeAccountResponse();
+        stripeAccountResponse.setStripeAccountId(gatewayAccountEntity.getCredentials().get("stripe_account_id"));
```
alexbishop1: We can probably avoid doing `gatewayAccountEntity.getCredentials().get("stripe_account_id"))` twice.

If my other suggestions are implemented, we could do something like this:

```java
return Optional.ofNullable(gatewayAccountEntity.getCredentials())
        .map(credentials -> credentials.get("stripe_account_id"))
        .map(StripeAccountResponse::new);
```

src/test/java/uk/gov/pay/connector/it/resources/StripeAccountResourceITest.java
```diff
@@ -0,0 +1,71 @@
...
+                .get("/v1/api/accounts/" + testAccount.getAccountId() + "/stripe-account")
+                .then()
+                .statusCode(200)
+                .body("stripe_account_id", is(STRIPE_ACCOUNT_ID));
+    }
+
+    @Test
+    public void getStripeAccountReturnsNotFoundResponseWhenGatewayAccountNotFound() {
+        givenSetup()
+                .get("/v1/api/accounts/1337/stripe-account")
```
alexbishop1, line 38: 😀

src/test/java/uk/gov/pay/connector/it/resources/StripeAccountResourceITest.java
```diff
@@ -0,0 +1,71 @@
...
+                .insert();
+
+        givenSetup()
+                .get("/v1/api/accounts/" + testAccount.getAccountId() + "/stripe-account")
+                .then()
+                .statusCode(404);
+    }
+
+    @Test
+    public void getStripeAccountReturnsNotFoundResponseWhenGatewayAccountCredentialsAreEmpty() {
```
alexbishop1, line 58: This is covered by the `StripeAccountServiceTest`: I wouldn’t retest it here

alexbishop1 (dismissed):
This is good. I’ve left a few suggestions but nothing big enough to block approval.

src/test/java/uk/gov/pay/connector/it/resources/StripeAccountResourceITest.java
alexbishop1: Actually, I’ve changed my mind about this; I now see that this test is the only test that checks what the resource does if the `StripeAccountService` returns an empty `Optional`. Keep it.

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
DanailMinchev: We discussed to use `buildStripeAccountResponse`, thanks

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
DanailMinchev: I am not sure, the field of the class is `null`, that is why I decided to check for `null`s, as discussed I will check for `null`s

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
DanailMinchev: Good point!

src/main/java/uk/gov/pay/connector/gatewayaccount/service/StripeAccountService.java
DanailMinchev: Beautiful! 👍

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #1124, opened 2019-03-05
PP-4937 Update Stripe Account credentials Pact fixture

DanailMinchev:
## WHAT YOU DID

- Update Stripe Gateway Account Pact fixture to have credentials which will used by connector client in Selfservice


cobainc0 (approved):
👍 LGTM

PR #1126, opened 2019-03-05
PP-4910 Add wallet type to Transaction

danworth:
- Add wallet type to `Transaction`, `TransactionDao` and associated tests. This
will return `wallet_type` in the response to `v2/api/accounts/{accountsId}/charges`
when the charge is a wallet payment.

N.B. `databaseTestHelper.addCharge()` warrants refactoring to remove the many overloaded methods with vast lists of arguments. I'll do this afterwards rather than mix with this PR and hold up progress (thus this makes do with existing patterns).


gidsg (approved):
💸 👍

PR #1127, opened 2019-03-05
PP-4653: Card holder name and email validation

oswaldquek:
These must be validated before being passed to worldpay, lest we end up in a
situation where worldpay processes the payment, but connector throws an
error due to the card holder name and/or being over 255 characters limit as
specified in the database.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
@@ -62,7 +62,8 @@ public class CardResource {
     @Path("/v1/frontend/charges/{chargeId}/wallets/apple")
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
-    public Response authoriseCharge(@PathParam("chargeId") String chargeId, ApplePayAuthRequest applePayAuthRequest) {
+    public Response authoriseCharge(@PathParam("chargeId") String chargeId, 
+                                    @NotNull @Valid ApplePayAuthRequest applePayAuthRequest) {
         logger.info("Received encrypted payload for charge with id {} ", chargeId);
         return applePayService.authorise(chargeId, applePayAuthRequest);
     }
--- a/src/main/java/uk/gov/pay/connector/wallets/model/WalletPaymentInfo.java
+++ b/src/main/java/uk/gov/pay/connector/wallets/model/WalletPaymentInfo.java
@@ -2,6 +2,7 @@ package uk.gov.pay.connector.wallets.model;
 
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import org.hibernate.validator.constraints.Length;
 import uk.gov.pay.connector.gateway.model.PayersCardType;
 
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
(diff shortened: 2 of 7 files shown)


src/test/java/uk/gov/pay/connector/it/resources/CardResourceAuthoriseApplePayITest.java
```diff
@@ -0,0 +1,82 @@
+package uk.gov.pay.connector.it.resources;
+
+import org.apache.http.HttpStatus;
```
danworth, line 8: I like these two separate test files for google and apple, makes it much easier to see what is covered and where rather than being buried amongst the regular card IT flow 👍

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -62,7 +62,8 @@ public CardResource(CardAuthoriseService cardAuthoriseService, Card3dsResponseAu
     @Path("/v1/frontend/charges/{chargeId}/wallets/apple")
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
-    public Response authoriseCharge(@PathParam("chargeId") String chargeId, ApplePayAuthRequest applePayAuthRequest) {
+    public Response authoriseCharge(@PathParam("chargeId") String chargeId, 
+                                    @NotNull @Valid ApplePayAuthRequest applePayAuthRequest) {
```
danworth, line 66: 👍

src/main/java/uk/gov/pay/connector/wallets/model/WalletPaymentInfo.java
```diff
@@ -10,7 +11,9 @@
     private String lastDigitsCardNumber;
     private String brand;
     private PayersCardType cardType;
+    @Length(max = 255, message = "Card holder name must be a maximum of 255 chars")
```
danworth, line 14: Only mentioning this out of interest, looks fine as-is, you can reference the values with template strings like `{max}` e.g. `"Card holder name must be a maximum of {max} chars"` but probably more descriptive as-is though 👍

src/main/java/uk/gov/pay/connector/wallets/model/WalletPaymentInfo.java
```diff
@@ -10,7 +11,9 @@
     private String lastDigitsCardNumber;
     private String brand;
     private PayersCardType cardType;
+    @Length(max = 255, message = "Card holder name must be a maximum of 255 chars") 
     private String cardholderName;
+    @Length(max = 255, message = "Email must be a maximum of 255 chars")
```
danworth: Strangely in the db I see email is 254
`email  character varying(254)`
whilst `cardholder_name character varying(255)`. Maybe this should be constrained to 254?

danworth (changes requested):
Looks very good, especially the refactoring to separate out the tests. Thanks for adding the validation to `wallets/apple` end-point, I had that on my mind to do but never did 😞 Query about email being `character varying(254)` in the charges table rather than 255 like the others and whether the validation should be 254 rather than 255?

src/main/java/uk/gov/pay/connector/wallets/model/WalletPaymentInfo.java
oswaldquek: good spot

danworth (dismissed):
Looks good.

danworth (approved):
👍

PR #1129, opened 2019-03-06
PP-4856 Fix worldpay inqury call

mrlumbu:
Previous merge did not compile due to changes in the GatewayClient.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -94,7 +94,12 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
 
     @Override
     public ChargeQueryResponse queryPaymentStatus(ChargeEntity charge)  throws GatewayErrorException {
-        GatewayClient.Response response = inquiryClient.postRequestFor(gatewayUrlMap.get(charge.getGatewayAccount().getType()), charge.getGatewayAccount(), buildQuery(charge));
+        GatewayClient.Response response = inquiryClient.postRequestFor(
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -94,7 +94,12 @@ public PaymentGatewayName getPaymentGatewayName() {
 
     @Override
     public ChargeQueryResponse queryPaymentStatus(ChargeEntity charge)  throws GatewayErrorException {
-        GatewayClient.Response response = inquiryClient.postRequestFor(gatewayUrlMap.get(charge.getGatewayAccount().getType()), charge.getGatewayAccount(), buildQuery(charge));
+        GatewayClient.Response response = inquiryClient.postRequestFor(
```
DanailMinchev, line 97: Do we need some tests for that or it is covered by `shouldCancelChargeWithGatewayWhenChargeInPreAuthorisedStateAndExistsWithGateway` test instead?

DanailMinchev (approved):
Approved to fix the red build with question about tests

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
mrlumbu: there are lots of tests for this. It only snuck in because we don't force branches to be up to date before merging

DanailMinchev: Sorry, I was unclear, I meant unit tests for `queryPaymentStatus` method, looks like we have some integration tests as mentioned in the first comment, do we need unit tests for this method or it's fine to have the integration tests only? Thanks

PR #1139, opened 2019-03-11
PP-4896: Add more logging

oswaldquek:
To better understand why Apple Pay payments fail.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/GatewayErrorException.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/GatewayErrorException.java
@@ -35,8 +35,20 @@ public abstract class GatewayErrorException extends Exception {
 
     public static class GatewayConnectionErrorException extends GatewayErrorException {
 
+        private final String responseFromGateway;
+
+        public GatewayConnectionErrorException(String message, String responseFromGateway) {
+            super(message);
+            this.responseFromGateway = responseFromGateway;
+        }
+
         public GatewayConnectionErrorException(String message) {
             super(message);
+            this.responseFromGateway = "null";
+        }
+
+        public String getResponseFromGateway() {
+            return responseFromGateway;
         }
 
         public GatewayError toGatewayError() {
--- a/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
@@ -10,6 +10,7 @@ import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.gateway.GatewayErrorException;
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -68,6 +69,13 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
                 }
 
             } catch (GatewayErrorException e) {
+                
+                logger.error("Error occurred authorising charge. Charge external id: {}; message: {}", charge.getExternalId(), e.getMessage());
+                
+                if (e instanceof GatewayConnectionErrorException) {
```
danworth, line 75: Codacy doesn't like the use of `instanceof` within exception handling and I think that was one of the rules we thought was worth keeping. Maybe its more appropriate to have two catch clauses for each type of exception?

oswaldquek: We should ignore Codacy in this case. Having two catch clauses means we'll have to set the `chargeStatus`, `responseFromPaymentGateway`, and `operationResponse` all over again.
Having this `instanceof` statement here is quite nice because we get the `"Error occurred authorising charge. ` log, straightaway followed by `"Response from gateway:` if its available. It's nicer to read here in my opinion.

danworth: OK.

src/main/java/uk/gov/pay/connector/gateway/GatewayErrorException.java
```diff
@@ -35,8 +35,20 @@ public GatewayError toGatewayError() {
...
+        private final String responseFromGateway;
+
+        public GatewayConnectionErrorException(String message, String responseFromGateway) {
+            super(message);
+            this.responseFromGateway = responseFromGateway;
+        }
+
         public GatewayConnectionErrorException(String message) {
             super(message);
+            this.responseFromGateway = "null";
```
danworth, line 47: Merely suggestion but might be better to use a more descriptive default like `response from gateway not captured`? You could also just call the other constructor with `this(message, "response from gateway not captured")`; to save duplicating constructor logic.

danworth (approved):
Looks good, comment for your consideration but not blocking.

src/main/java/uk/gov/pay/connector/gateway/GatewayErrorException.java
oswaldquek: Yeah I think this needs a bit of refactor in the future. There are places like https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayResponseUnmarshaller.java#L23 where a different Exception should be thrown and having a `"response from gateway not captured"` wouldn't make sense currently. I do want to get logging going to see what's happening with Apple Pay now though, and would rather not do this refactor just now.

danworth: Yeah that makes sense. I figured you were hampered a little by its existing use, so agree lets get the logging in there and chalk it up to refactor later.

PR #1142, opened 2019-03-11
PP-4926 deprecate service name and service name patch endpoint

SandorArpa:
## WHAT YOU DID
- service name is before the times of the concept of service was introduced in adminusers. We no longer use this field and keeping it in sync with the source of truth (adminusers) is a costly and messy operation.

## How to test

This PR only adds `@Deprecated` tag to the code


DanailMinchev (approved):
👍

PR #1143, opened 2019-03-12
PP-4735 Context Only / Relative URI for pagination links

kbottla:
## WHAT YOU DID
- Changes full URLs for pagination to relative links for endpoint `/v2/api/accounts` ( which is used by selfservice to download transactions ) as full URLs are generated with incorrect port number
- Selfservice can resolve relative URLs as it has knowledge of connector URL (https://github.com/alphagov/pay-selfservice/pull/1265)

_**Note:**_
- All pagination links generated by connector are using `UriInfo` which means links are all generated with incorrect port number. But URLs are probably not actively used or transformed (publicapi transforms URLs to publicapi's baseURL and selfservice web pages uses baseUrl to start journey to connector except while downloading transactions)

## How to test
- All tests should pass
- Any suggestions to how the solution can be better?

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/service/PaginationResponseBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/common/service/PaginationResponseBuilder.java
@@ -1,27 +1,34 @@
 package uk.gov.pay.connector.common.service;
 
 import black.door.hate.HalRepresentation;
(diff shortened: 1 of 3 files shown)


oswaldquek (approved):
LGTM

PR #1152, opened 2019-03-18
PP-4758: Differentiate between 4xx and 5xx exception

oswaldquek:
Callers of StripeGatewayClient handle 4xx and 5xx differently by virtue of
StripeGatewayClient throwing separate exceptions for each. This commit
introduces GatewayErrorException.DownstreamErrorException (for 4xx status
codes) and GatewayErrorException.ClientErrorException (for 5xx status codes) to the GatewayClient to
make deletion of StripeGatewayClient easier later. It's also probably a
good idea so we can tell whether it's connector making a bad request or a
downstream error from a payment provider.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
@@ -4,7 +4,8 @@ import com.codahale.metrics.MetricRegistry;
 import com.google.common.base.Stopwatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import uk.gov.pay.connector.gateway.GatewayErrorException.GatewayConnectionErrorException;
+import uk.gov.pay.connector.gateway.GatewayErrorException.ClientErrorException;
+import uk.gov.pay.connector.gateway.GatewayErrorException.DownstreamErrorException;
 import uk.gov.pay.connector.gateway.GatewayErrorException.GatewayConnectionTimeoutErrorException;
 import uk.gov.pay.connector.gateway.GatewayErrorException.GenericGatewayErrorException;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
@@ -13,6 +14,7 @@ import javax.ws.rs.ProcessingException;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.Invocation.Builder;
+import javax.ws.rs.core.Response.Status.Family;
 import java.net.HttpCookie;
 import java.net.SocketTimeoutException;
 import java.net.URI;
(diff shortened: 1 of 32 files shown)


src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -36,38 +38,31 @@ public GatewayClient(Client client, MetricRegistry metricRegistry) {
...
+    public GatewayClient.Response postRequestFor(URI url,
+                                                 GatewayAccountEntity account,
+                                                 GatewayOrder request,
+                                                 List<HttpCookie> cookies,
+                                                 Map<String, String> headers)
+            throws GenericGatewayErrorException, GatewayConnectionTimeoutErrorException, ClientErrorException, DownstreamErrorException {
+
         String metricsPrefix = format("gateway-operations.%s.%s.%s", account.getGatewayName(), account.getType(), request.getOrderRequestType());
         javax.ws.rs.core.Response response = null;
+        Response gatewayResponse = null;
```
kbottla, line 55: This could just be ` Response gatewayResponse;`

src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -91,12 +84,25 @@ public GatewayClient(Client client, MetricRegistry metricRegistry) {
                 response.close();
             }
         }
+
+        int statusCode = response.getStatus();
```
kbottla, line 88: Can we use `gatewayResponse` response here. Finally block has condition `response!=null` which indicates this can be too for any edge cases ?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -132,7 +133,10 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
             StripeGatewayClientResponse response = e.getResponse();
 
             if (response.getStatus() == HttpStatus.SC_UNAUTHORIZED) {
-                return unexpectedGatewayResponse(request.getChargeExternalId(), responseBuilder, e.getMessage(), HttpStatus.SC_UNAUTHORIZED);
+                logger.error("Authorisation failed for charge {}. Reason: {}. Status code from Stripe: {}.",
```
kbottla, line 136: Is there any reason we don't use method `unexpectedGatewayResponse ` anymore ? It looks clean to have it in a method

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -143,7 +147,10 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
             return responseBuilder.withGatewayError(gatewayError).build();
 
         } catch (DownstreamException e) {
-            return unexpectedGatewayResponse(request.getChargeExternalId(), responseBuilder, e.getMessage(), e.getStatusCode());
+            logger.error("Authorisation failed for charge {}. Reason: {}. Status code from Stripe: {}.",
```
kbottla, line 150: Same as above for method `unexpectedGatewayResponse`

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCancelHandler.java
```diff
@@ -26,7 +26,7 @@
 import static java.util.Collections.singletonList;
 import static java.util.UUID.randomUUID;
 import static javax.ws.rs.core.MediaType.APPLICATION_FORM_URLENCODED_TYPE;
-import static uk.gov.pay.connector.gateway.model.GatewayError.gatewayConnectionError;
+import static uk.gov.pay.connector.gateway.model.GatewayError.gatewayDownstreamError;
 import static uk.gov.pay.connector.gateway.model.GatewayError.genericGatewayError;
```
kbottla, line 30: For `GatewayClientException` across Stripe functionality we could probably use `gatewayClientError` for consistency

kbottla (changes requested):
Looks good to me for the changes and a few minor comments.There are a few unused imports as per Codacy.

As far as I understand we don't handle 4xx & 5xx errors separately up the chain, we may not be gaining much benefit from handling errors separately and increasing the cases to handle across payment providers. Tests have gone a little inconsistent too as 4xx is tested for some payment providers and 5xx in other. I think we need tests for both 4xx & 5xx so we can be sure each payment provider handles both.

StripeGatewayClient is handling 4xx separately as it initially returned `Response` object and 4xx responses are structured from Stripe. So we want to handle 4xx responses to cast to object for logging. But this is not the case anymore as StripeGatewayClient now returns response as String. So it can just be one exception for both (4xx & 5xx) for Stripe too.

Unless we have different goals & needs for 4xx & 5xx, I feel a single exception would serve the purpose.

oswaldquek:
@kbottla Yeah I think you're right, both 4xx and 5xx exceptions can just be one exception for both. I think it would be more prudent for me to make the existing GatewayConnectionErrorException (https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayErrorException.java#L36) contain the status code and response entity. Then it will be easier to get rid of the StripeGatewayClient after that. I will do this in another PR.

PR #1153, opened 2019-03-18
bau: Temporarily log apple payload

oswaldquek:
Unfortunately since the keys were rotated on 15/3/2019 apple pay no longer
works. I need a payload to test this locally.


alexbishop1 (approved):
Please announce a suitable warning before this gets merged and deployed to test so that people are aware of the implications if they make test payments. Make another announcement when the logging is removed.

And don’t let it get anywhere near production!

oswaldquek:
@alexbishop1 unfort connector builds are also broken at the moment so i'll hold off merging this for now

PR #1158, opened 2019-03-19
PP-4992: Add refunds gateway ID, processing column

sfount:
* Add `gateway_transaction_id` to bring this entity in line with charges
(and upcoming fees)
* Index on `gatway_transaction_id` as this value should be searchable
for payout reconiciliation
* Add `gateway_processing_details` column to allow unkown gateway
details to be recorded (metatdata about gateway transactions, process
specific timestamps)
* Gateway processing details is JSONB to allow unkown structure to be
used, indexes should be addded in the future for any commonly used
attributes

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1020,9 +1020,12 @@
                 <constraints nullable="true" unique="true" />
             </column>
         </addColumn>
-        <createIndex tableName="refunds" indexName="idx_refunds_gateway_transaction_id" unique="true">
-            <column name="gateway_transaction_id" type="text" />
-        </createIndex>
+    </changeSet>
+    
+    <changeSet id="add index concurrently to new refunds gateway_transaction_id column" runInTransaction="false" author="">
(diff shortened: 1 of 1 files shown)


mrlumbu (changes requested):
Can you put each change in a separate changeset - see https://www.liquibase.org/bestpractices.html
```
As far as possible, Avoid multiple changes per changeset to avoid failed autocommit statements that can leave the database in an unexpected state.
```

mrlumbu (approved):

mrlumbu (changes requested):
Just remembered you need to create the index in a separate thing in this case, because index creation will lock the table. So need to do something like:
```
<changeSet id="add index to charges.last_digits_card_number" runInTransaction="false" author="">
        <sql>
            CREATE INDEX CONCURRENTLY idx_last_digits_card_number_idx ON charges(last_digits_card_number);
        </sql>
    </changeSet>
```

mrlumbu (approved):

PR #1159, opened 2019-03-19
PP-4992: Expose gateway transaction ID for refunds

sfount:
* Links `RefundEntity` & `RefundsHistory` to new `gatway_transaction_id` column
* Doesn't expose details column, this should be updated given a use case

Depends on https://github.com/alphagov/pay-connector/pull/1158

- [x] Update any relevant tests 
- [x] Update any other entities that needs this information

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/dao/RefundDao.java
+++ b/src/main/java/uk/gov/pay/connector/refund/dao/RefundDao.java
@@ -75,7 +75,7 @@ public class RefundDao extends JpaDao<RefundEntity> {
 
     public List<RefundHistory> searchHistoryByChargeId(Long chargeId) {
 
-        String query = "SELECT id, external_id, amount, status, charge_id, created_date, version, reference, history_start_date, history_end_date, user_external_id " +
+        String query = "SELECT id, external_id, amount, status, charge_id, created_date, version, reference, history_start_date, history_end_date, user_external_id, gateway_transaction_id " +
                 "FROM refunds_history r " +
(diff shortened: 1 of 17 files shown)


sfount:
Note Codacy rule flagged here is going to be ignored in the future, it doesn't seem to actually line up with repository uses: 
> Use explicit scoping instead of the default package private level

kbottla (commented):
Looks good to me. We would also need to include `gateway_transaction_id` when inserting RefundsHistory (https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundEntity.java#L37) but in a separate PR.

Also PR needs a rebase!

kbottla:
Having looked at the code again, looks like `HistoryCustomiser` is taking care of adding `gateway_transaction_id` into RefundsHistory. We still need to decide if we need to add gateway_transaction_id for select on RefundsHistory (Don't need it if impacts any API) https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundEntity.java#L37

src/main/resources/migrations.xml
```diff
@@ -1086,4 +1086,12 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add gateway_transaction_id column to refunds_history table" author="">
```
kbottla: Can we please have migrations first as separate PR ? Refunds may fail if changes are deployed but DB migrations are done afterwards

mrlumbu (approved):

PR #1161, opened 2019-03-19
PP-5003: Add fee table schema migration

sfount:
* Adds initial columns required for modelling fees
* Indexing gateway processing details will follow when we know what keys
are required

Fees are modelled separately to reflect the second account debit transaction that is required on the Stripe gateway. For this use case they are 1:1 for a charge which is enforced in https://github.com/alphagov/pay-connector/pull/1161/commits/87c575f81a9dffc417c4c30b721442f282ea9b67. If multiple fees for a given charge become a use case in the future this constraint can be updated along with the "view" code for charges and their fees.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1024,11 +1024,17 @@
                 <constraints foreignKeyName="fk__charge_fees"
                              referencedTableName="charges"
                              referencedColumnNames="id"
+                             unique="true"
                              nullable="false" />
             </column>
-            <column name="fee_due" type="bigint" />
-            <column name="fee_collected" type="bigint" />
+            <column name="amount_due" type="bigint">
+                <constraints nullable="false" />
+            </column>
+            <column name="amount_collected" type="bigint">
+                <constraints nullable="false" />
+            </column>
             <column name="created_date" type="timestamp without timezone" />
+            <column name="collected_date" type="timestamp without timezone" />
             <column name="gateway_transaction_id"  type="text">
                 <constraints nullable="true" />
             </column>
@@ -1036,7 +1042,7 @@
         </createTable>
     </changeSet>
     
-    <changeSet id="index columns for fee table" author="">
+    <changeSet id="index gateway_transaction_id, external_id and charge_id columns for fee table" author="">
         <createIndex tableName="fees" indexName="idx_fees_gateway_transaction_id">
             <column name="gateway_transaction_id" />
         </createIndex>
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1013,5 +1013,33 @@
...
+            <column name="external_id" type="char(26)" />
+            <column name="charge_id" type="bigserial">
+                <constraints foreignKeyName="fk__charge_fees"
+                             referencedTableName="charges"
+                             referencedColumnNames="id"
+                             nullable="false" />
+            </column>
+            <column name="fee_due" type="bigint" />
+            <column name="fee_collected" type="bigint" />
+            <column name="created" type="timestamp without timezone" />
```
mrlumbu: can this be `created_date` for consistency

src/main/resources/migrations.xml
```diff
@@ -1013,5 +1013,33 @@
...
+            <column name="gateway_transaction_id"  type="text">
+                <constraints nullable="true" />
+            </column>
+            <column name="gateway_processing_details" type="jsonb" />
+        </createTable>
+    </changeSet>
+    
+    <changeSet id="index gateway_transaction_id for fee table" author="">
+        <createIndex tableName="fees" indexName="itx_fees_gateway_transaction_id">
```
mrlumbu: `s/itx/idx/` ?

mrlumbu (changes requested):

src/main/resources/migrations.xml
sfount: 👍 will line this up with other entities

src/main/resources/migrations.xml
```diff
@@ -1013,5 +1013,33 @@
...
+            </column>
+            <column name="gateway_processing_details" type="jsonb" />
+        </createTable>
+    </changeSet>
+    
+    <changeSet id="index gateway_transaction_id for fee table" author="">
+        <createIndex tableName="fees" indexName="idx_fees_gateway_transaction_id">
+            <column name="gateway_transaction_id"></column>
+        </createIndex>
+    </changeSet>
```
mrlumbu, line 1057: probs need index on charge_id and external_id

sfount: I've added these indexes to the same `changeSet`

sfount:
EDIT Update fee constraints given: 
> Fees are modelled separately to reflect the second account debit transaction that is required on the Stripe gateway. For this use case they are 1:1 for a charge which is enforced in 87c575f. If multiple fees for a given charge become a use case in the future this constraint can be updated along with the "view" code for charges and their fees.

mrlumbu (approved):

PR #1162, opened 2019-03-19
Linter fixes to appease lgtm.com

rhowe-gds:
## WHAT YOU DID
Avoid unnecessary boxing of primitive types
Make the parameters passed to loggers match the number of format placeholders

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
@@ -121,9 +121,11 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
         Query query = entityManager.get().createQuery(cq);
 
         if (params.getPage() != null && params.getDisplaySize() != null) {
-            Long firstResult = (params.getPage() - 1) * params.getDisplaySize(); // page coming from params is 1 based, so -1
-            query.setFirstResult(firstResult.intValue());
-            query.setMaxResults(params.getDisplaySize().intValue());
+            long displaySize = params.getDisplaySize();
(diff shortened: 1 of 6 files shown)


danworth (approved):
Looks good, I think we could fix up `SearchParams` to use `Integer` than `Long` but that isn't something you've introduced.

src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
```diff
@@ -121,9 +121,11 @@ public ChargeDao(final Provider<EntityManager> entityManager) {
         Query query = entityManager.get().createQuery(cq);
 
         if (params.getPage() != null && params.getDisplaySize() != null) {
-            Long firstResult = (params.getPage() - 1) * params.getDisplaySize(); // page coming from params is 1 based, so -1
-            query.setFirstResult(firstResult.intValue());
-            query.setMaxResults(params.getDisplaySize().intValue());
+            long displaySize = params.getDisplaySize();
```
danworth, line 124: Its unfortunate that `getDisplaySize` and `getPage` return `Long` which is then cast to `int` since  I can't foresee there being more than 2,147,483,647 results nor anyone wishing to view that many on one page. Also if there ever was we'll get weird results from overflow when casting to int...
```
jshell> long aLong = Integer.MAX_VALUE;
aLong ==> 2147483647

jshell> (int) ++aLong;
$9 ==> -2147483648
```
Nothing that you've introduced so perhaps a little refactoring for later on...

rhowe-gds: Yes, this is indeed unfortunate

PR #1166, opened 2019-03-21
PP-5010 Add fee entity

mrlumbu:
Adds FeeEntity, and mapping to ChargeEntity.

At present the only way we want to expose the fee is as part of the
charge, so this commit adds a `getFeeAmount` method to the charge.

Passes through fee from ChargeEntity to ChargesFrontendResource and
ChargeApiResource.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -7,7 +7,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.commons.model.SupportedLanguageJpaConverter;
-import uk.gov.pay.connector.wallets.WalletType;
 import uk.gov.pay.connector.charge.model.CardDetailsEntity;
 import uk.gov.pay.connector.charge.model.ServicePaymentReference;
 import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
@@ -19,6 +18,7 @@ import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.refund.model.domain.RefundEntity;
 import uk.gov.pay.connector.util.RandomIdGenerator;
+import uk.gov.pay.connector.wallets.WalletType;
 
 import javax.persistence.Access;
 import javax.persistence.AccessType;
@@ -35,6 +35,7 @@ import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
+import javax.persistence.OneToOne;
 import javax.persistence.OrderBy;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
@@ -42,7 +43,6 @@ import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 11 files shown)


sfount (approved):
Looks good from a functionality perspective -- code looks good but would be good for some more familiar with Connector/ our Java code to approve this too. 

👍  Testing the fee collected is reflected as correct fee amount

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -95,6 +96,9 @@
     @JoinColumn(name = "gateway_account_id", updatable = false)
     private GatewayAccountEntity gatewayAccount;
 
+    @OneToMany(mappedBy = "chargeEntity", fetch = FetchType.EAGER)
+    private List<FeeEntity> fees = Collections.emptyList();
```
oswaldquek: Given the number of fees for a given charge is guaranteed to be 0 or 1, is there any reason for using `@OneToMany` and a `List` of fees here? Or is this for some future functionality?

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -1536,6 +1536,26 @@ public void findByIdAndLimit() {
...
+                .insert();
+
+        assertThat(chargeDao
+                .findById(defaultTestCharge.getChargeId())
+                .flatMap(ChargeEntity::getFeeAmount)
+                .get(), 
+                is(10L)
+        );
+    }
+
```
oswaldquek, line 1558: i'm somewhat uncomfortable that this test is in excess of 1558 lines, and would look to splitting this out by functionality

sfount (changes requested):
It might be a good idea to test the search resources/ API endpoints that fetch multiple charges at once, just to make sure the entity is correctly propagated and there is no custom logic that will have to be updated.

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
mrlumbu: For some reason I thought this wasn't possible without specifying fee_id on charge, but turns out it is 👍

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
mrlumbu: I agree, but would rather not do as part of this PR, as it will just muddy the waters. I can do as a separate PR 👍

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -325,4 +329,12 @@ public void setDelayedCapture(boolean delayedCapture) {
     public void setCorporateSurcharge(Long corporateSurcharge) {
         this.corporateSurcharge = corporateSurcharge;
     }
+
+    public Optional<Long> getFeeAmount() {
+        if (fee != null) {
```
oswaldquek: `Optional.ofNullable` could be used?

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
oswaldquek: i agree

oswaldquek (changes requested):
Just a comment on the use of Optional

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
mrlumbu: Not sure you can - it's the fee that might be null, not the amount? Or maybe optional is cleverer than I thought.

Having checked, it's not

mrlumbu:
> It might be a good idea to test the search resources/ API endpoints that fetch multiple charges at once, just to make sure the entity is correctly propagated and there is no custom logic that will have to be updated.

I would like to do this in a separate PR

sfount (dismissed):

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
oswaldquek: To be honest, I misread it and missed the `fee.getAmountCollected()` bit. but one can still use "ofNullable" like `return Optional.ofNullable(fee).map(f -> f.getAmountCollected())`

oswaldquek: or `Optional.ofNullable(fee).map(FeeEntity::getAmountCollected)` even

mrlumbu: yeah, true, that is nicer

oswaldquek (approved):

PR #1169, opened 2019-03-21
PP-4864: Use the default Java truststore for TLS

rhowe-gds:
Note for reviewers: Does this change the TLS settings (TLS versions, ciphersuites etc) of our HTTP clients?

Prior to this, we ignored the default Java truststore and built our own using
the certificates from the directory specified by the CERTS_PATH environment
variable.

We can instead rely on the default Java truststore, although there are cases
where we need to add additional CA certificates:

* Amazon RDS has its own CA certificate. We add this to the default Java
  truststore when building our base container image.
* During our end to end tests, we supply custom certificates in order to run
  TLS in our docker-compose environment.

Handle these custom certificates in the existing way, using CERTS_PATH to point
to them and have the shell wrapper add them to the default Java truststore
prior to starting the application.

Since this is a rather slow operation, taking about 0.5s per certificate, have
some logic to avoid adding certificates to the truststore if they're already in
there, matching on their certificate fingerprint. We need to add openssl to the
container in order to extract that data.

By default, the PostgreSQL JDBC driver does not use the default Java truststore
to validate server certificates. Instead, it defaults to looking for
${user.home}/.postgresql/root.crt

Rather than supply this file specifically, we can use the provided
org.postgresql.ssl.DefaultJavaSSLFactory which will consult the default Java
truststore instead.


kbottla (approved):
LGTM! but left a comment on runtime exception of SSL factory

src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
```diff
@@ -87,21 +87,24 @@ private Duration getReadTimeout(GatewayOperation operation, PaymentGatewayName g
...
+        SSLConnectionSocketFactory sslConnectionSocketFactory;
+        try {
+            sslConnectionSocketFactory = new SSLConnectionSocketFactory(
+                    SSLContext.getDefault(),
+                    new String[]{"TLSv1.2"},
+                    null,
+                    (HostnameVerifier) null
+            );
+        } catch (NoSuchAlgorithmException e) {
+            throw new RuntimeException("Unable to create SSL connection socket factory", e);
```
kbottla, line 100: This slightly worries me as we will not probably know if this works until we hit smoke tests as connector doesn't have a way to test this / fail to start if can't create factory

rhowe-gds: Yes, I agree, however I couldn't think of a better way to do it. I think the smoketests are maybe good enough to catch this being a problem? We're basically falling back to defaults here mostly so the main failure scenario I can foresee is a change of Java runtime meaning that `TLSv1.2` becomes unavailable.
Rare enough, I reckon.

PR #1170, opened 2019-03-25
PP-5010 Add fee to v2 api

mrlumbu:
The transaction based search uses a different mechanism to build
results. This commit adds fees to transaction reporting.


sfount (approved):
Lines up with story requirements 👍 

Supports the fee amount for queries using the `Transaction` entity, related to https://github.com/alphagov/pay-connector/pull/1166

PR #1171, opened 2019-03-25
PP-4994: Use Stripe charge `transfer_data` API

sfount:
Stripe's charge creation API has been updated with no direct mention of
the `destination[amount]` syntax that we use for creating charges.

Update the params to use the `transfer_data[destination]` syntax and
explicitly set `on_behalf_of` to specify the settlement merchant to use
for the transaction, this is no longer implicit.

The responses from Stripe for contractual tests may need to be updated
to reflect the new attributes that are now set.

## WHAT YOU DID
* Update charge creation params

## How to test
* Make sure payments still work in exactly the same way

This should be carefully tested from both our end (connector DB) through to Stripe's end (Stripe dashboard).


kbottla (commented):
LGTM and shouldn't be different as per https://stripe.com/docs/connect/destination-charges?#differences

kbottla (approved):

sfount:
Merging this API bump, we should be keeping an eye on Stripe transactions in tests and production to make sure everything is running BAU.

PR #1173, opened 2019-03-26
PP-4994: Update refund process for fee billing, 

sfount:
* Remove `refund_application_fee` as we do not take application fees
during the capture process
* Add the relevant Connect account to the refund metadata, this will
help Support follow the payment trail if required

Required: 
- [ ] fill in the `gatewayTransactionId` when a refund is created with the Stripe gateway

Depends on https://github.com/alphagov/pay-connector/pull/1159


sfount:
Superseded by https://github.com/alphagov/pay-connector/pull/1199 -- refund process updated for net transfer method.

PR #1174, opened 2019-03-27
PP-4993 Add Stripe Fee Processor

mrlumbu:
Adds fee processor interface, and implementation for Stripe.

The fee processor does 2 things:
given a total amount and a baseFee, it calculates a total fee due for a charge.
I haven't tried to make this very clever or generic. If we need to extend in 
future it is likely to be in ways we cannot foresee.

recoups a fee: it can talk to a gateway to recoup a fee for a charge.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
@@ -18,6 +18,10 @@ public class StripeGatewayConfig extends Configuration {
     @Valid
     @NotNull
     private StripeWebhookSigningSecrets webhookSigningSecrets;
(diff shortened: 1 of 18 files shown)


src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
```diff
@@ -18,6 +18,10 @@
     @Valid
     @NotNull
     private StripeWebhookSigningSecrets webhookSigningSecrets;
+    
+    @Valid
+    @NotNull
+    private Double feePercentage;
```
heathd, line 24: would it be better to use a precise representation of this rather than an imprecise float?

PR #1175, opened 2019-03-28
PP-5042 Make cancelling use smaller, tighter transactions

alexbishop1:
When cancelling a charge, don’t do the entire operation, including sending a cancellation message to the gateway, inside a single transaction because this gives lots of time for something else to come in and change the charge in the mean time. Instead, have transactions tightly around just the database updates and do the communication with the gateway outside a transaction. This means we will atomically change the status to a `READY` state, then communicate with the gateway, then atomically change the status to a terminal cancellation state. If anything else tries to change the charge while we’re talking to the gateway, it will observe the charge in a `READY` state and know to not try to do that.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
@@ -5,6 +5,7 @@ import com.google.inject.persist.Transactional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
+import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.chargeevent.dao.ChargeEventDao;
@@ -53,7 +54,6 @@ public class ChargeCancelService {
         this.providers = providers;
     }
 
-    @Transactional
     public Optional<ChargeEntity> doSystemCancel(String chargeId, Long accountId) {
         return chargeDao.findByExternalIdAndGatewayAccount(chargeId, accountId)
                 .map(chargeEntity -> {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -127,20 +124,20 @@ private void nonGatewayCancel(ChargeEntity chargeEntity, StatusFlow statusFlow)
...
-        chargeEventDao.persistChargeEventOf(chargeEntity);
+
+        setChargeStatusTo(chargeEntity, completeStatus);
     }
 
     private boolean gatewayIsNotAwareOfCharge(ChargeEntity chargeEntity) {
         return nonGatewayStatuses.contains(ChargeStatus.fromString(chargeEntity.getStatus()));
     }
 
+    @Transactional
```
kbottla: This method should be public for \@Transactional annotation to work as per https://github.com/google/guice/wiki/Transactions

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -154,6 +151,12 @@ private void prepareForTerminate(ChargeEntity chargeEntity, StatusFlow statusFlo
                 chargeEntity.getGatewayAccount().getType(),
                 newStatus);
 
+        setChargeStatusTo(chargeEntity, newStatus);
+    }
+
+    @Transactional
+    private void setChargeStatusTo(ChargeEntity chargeEntity, ChargeStatus chargeStatus) {
```
kbottla: Same as above for public method

kbottla (changes requested):

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
alexbishop1: Oops! Will fix

kbottla (dismissed):
LGTM 👍

alexbishop1:
End-to-end tests alerted us that this change wasn’t persisting the statuses correctly. Fixed that by making the code retrieve the `ChargeEntity` from the DAO between transactions (which required unit test changes).

kbottla (approved):
LGTM!

PR #1176, opened 2019-03-28
BAU - Fix log message for Stripe notification

kbottla:
## WHAT YOU DID
- Updated log message that incorrectly logs current charge status as both current and new.


danworth (approved):
good spot 👍

PR #1177, opened 2019-03-29
PP-5050 Migrate to Java 11

kbottla:
## WHAT YOU DID
- Java 11 everywhere (container base image, Jenkins, pom )

## How to test
- CI should pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -445,8 +445,7 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.0</version>
                 <configuration>
-                    <source>11</source>
-                    <target>11</target>
+                    <release>11</release>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -445,8 +445,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.8.0</version>
                 <configuration>
-                    <source>1.8</source>
-                    <target>1.8</target>
+                    <source>11</source>
```
danworth: Is there a reason why we're not using the `release` flag here instead?

kbottla: No, updated to use <release> tag

danworth (approved):
👍

PR #1180, opened 2019-04-01
PP-4758: Add status code to GatewayErrorException

oswaldquek:
The end goal is to delete StripeGatewayClient. Classes that use
StripeGatwayClient rely on the status code field of the
`uk.gov.pay.connector.gateway.stripe.GatewayClientException` thrown to
determine how to respond to the original API call. This commit therefore adds
status code to the now-renamed GatewayErrorException so we can deprecate
StripeGatewayClient while preserving functionality. Note that
`uk.gov.pay.connector.gateway.GatewayException.GatewayErrorException` will
represent 4xx and 5xx errors and there is no need to have separate classes
representing either like it's being done with StripeGatewayClient.

I also took the opportunity to rename the
`uk.gov.pay.connector.gateway.GatewayException` class and subclasses to
something more succint. I really disliked the old naming of
'GatewayConnectionErrorException' as it was never about an issue with
connecting to the PSP.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/GatewayException.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/GatewayException.java
@@ -3,17 +3,19 @@ package uk.gov.pay.connector.gateway;
 import uk.gov.pay.connector.gateway.model.ErrorType;
 import uk.gov.pay.connector.gateway.model.GatewayError;
 
-public abstract class GatewayErrorException extends Exception {
+import java.util.Optional;
 
-    private GatewayErrorException(String message) {
+public abstract class GatewayException extends Exception {
+
+    private GatewayException(String message) {
         super(message);
     }
 
     public abstract GatewayError toGatewayError();
 
-    public static class GenericGatewayErrorException extends GatewayErrorException {
+    public static class GenericGatewayException extends GatewayException {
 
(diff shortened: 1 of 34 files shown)


src/main/java/uk/gov/pay/connector/gateway/GatewayException.java
```diff
@@ -33,18 +33,25 @@ public GatewayError toGatewayError() {
...
             super(message);
             this.responseFromGateway = responseFromGateway;
+            this.status = status;
         }
 
-        public GatewayConnectionErrorException(String message) {
+        public GatewayErrorException(String message) {
             super(message);
             this.responseFromGateway = "null";
+            this.status = 0;
```
alexbishop1: Rather than defaulting to zero, would it be possible to make `getStatus()` return an `Optional<Integer>`?

alexbishop1 (commented):
Good renaming but I would have found this easier to follow if each renaming was in a separate commit, with an additional commit to add the status code to `GatewayErrorException`. One question below.

src/main/java/uk/gov/pay/connector/gateway/GatewayException.java
oswaldquek: fixed

alexbishop1 (approved):

PR #1184, opened 2019-04-03
Pp 5022 Log payment processing certificate time to expire

oswaldquek:
Also throw an exception on startup if the private key and/or certificate aren't valid. It makes sense to do so.


src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypter.java
```diff
@@ -30,37 +30,44 @@
...
-        this.publicCertificate = BASE64_DECODER.decode(applePayConfig.getPublicCertificate());
         this.objectMapper = objectMapper;
+        ApplePayConfig applePayConfig = configuration.getWorldpayConfig().getApplePayConfig();
+        try {
+            privateKey = generatePrivateKey(BASE64_DECODER.decode(applePayConfig.getPrivateKey()));
+            certificate = generateCertificate(BASE64_DECODER.decode(applePayConfig.getPublicCertificate()));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        long daysToExpiry = TimeUnit.MILLISECONDS.toDays(certificate.getNotAfter().getTime() - new Date().getTime());
```
alexbishop1: This might be a bit more readable:

```java
long daysToExpiry = ChronoUnit.DAYS.between(Instant.now(), certificate.getNotAfter().toInstant());
```

(Not checked if it behaves exactly the same for partial days but it’ll be close enough for our purposes as we’re always going renew the certificate several days before.)

alexbishop1 (changes requested):
Looks good — one small suggestion

src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypter.java
oswaldquek: fixed

src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypter.java
```diff
@@ -29,38 +29,47 @@
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.security.spec.X509EncodedKeySpec;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Base64;
+import java.util.Date;
+import java.util.concurrent.TimeUnit;
```
alexbishop1: These two imports aren’t actually used in the latest revision of this code

alexbishop1 (changes requested):
Thanks for making the change. Please remove a couple of unused imports and then this is good to go.

src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypter.java
alexbishop1: Might actually be nice to statically import `ChronoUnit.DAYS` so this code becomes:

```java
long daysToExpiry = DAYS.between(Instant.now(), certificate.getNotAfter().toInstant());
```

alexbishop1 (approved):

PR #1188, opened 2019-04-08
Remove dependency-check plugin from POM

alexbishop1:
We only ever run this manually, which we can do without it being in the POM:

```
$ mvn org.owasp:dependency-check-maven:check
```


rhowe-gds:
`mvn -DsuppressionFile=project-suppression.xml org.owasp:dependency-check-maven:check`

rhowe-gds (approved):

PR #1207, opened 2019-04-16
PP-4884 Add externalMetadata to the ChargeEntity

danworth:
- add externalMetadata to `ChargeEntity` and `ChargeEntityFixture`.
- add tests in `ChargeDaoITest` to confirm persistance and retrieval of
externalMetadata.

## WHAT YOU DID
Splitting this work up into a few PRs after it snow balled. This is to add the persistence layer. Subsequent PRs to follow to add to service and Api layer.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.charge.model.domain;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -220,6 +227,10 @@ public String getProviderSessionId() {
         return providerSessionId;
     }
 
+    public JsonNode getExternalMetadata() {
```
alexbishop1, line 230: Maybe make this return an `Optional<JsonNode>`? Seems we do that for some other getters in this class.

alexbishop1 (changes requested):
Looks good — one suggestion (that I’m not wedded too).

On a wider point — are we certain about using `JsonNode` to represent the metadata? Theoretically, a `JsonNode` can be basically any JSON tree, whereas our metadata is restricted to just key-value pairs. If we had a specific type for this (probably wrapping a `Map`), we could have a nice API that makes it clear that it’s only ever going to be key-value pairs.

(Also, I’m a bit worried about what happens if someone goes into the database directly and modifies the external metadata for a payment. If it just gets deserialised into a `JsonNode`, it could potentially get out of connector and only blow up when some other service that assumes it’s just key-value pairs tries to use it.)

danworth:
After speaking with Alex, its agreed that we'll create a Type that stores the value as a Map. This is probably the safest approach and most strongly typed (lacks the future flexibility of using a JsonNode but we might never need it...). Thanks Alex.

PR #1209, opened 2019-04-16
PP-5081 add logic to ChargeService to handle PrefilledCardHolderDetails

SandorArpa:
## WHAT YOU DID
- add logic to transform PrefilledCardHolderDetails to ChargeEntity if present
- add relevant tests

with @nimalank7


kbottla (approved):
Looks good to me 👍. 
Don't we need test in `ChargesApiResourceITest` but happy to be in separate PR if necessary ?

PR #1210, opened 2019-04-16
PP-4884 Add externalMetadata to ChargeEntity

danworth:
- Create `ExternalMetadata` type
- Add externalMetadata to `ChargeEntity` and `ChargeEntityFixture`
- Add test to `ChargeDaoITest` to confirm persistence and retrieval of externalMetadata.
- Add attributeConverter and custom exception for converting between `ExternalMetadata`
and `PGobject` as required by postgres for `json-b` column type.

## WHAT YOU DID
As discussed with @alexbishop1 we've decided to use a custom Type to store the external metadata rather than `JsonNode` for increased type safety. This introduces the type and uses it to add externalMetadata to the ChargeEntity (which I could see migrating to `pay-java-commons` if and when required to do so).

Validation logic of the meta data (e.g. the `Map` inside ExternalMetadata) will follow in the next step of adding it to the service and Api layer).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
@@ -1,0 +1,16 @@
+package uk.gov.pay.connector.charge.model;
+
+import java.util.Map;
+
+public class ExternalMetadata {
+
+    private final Map<String, Object> metadata;
+
+    public ExternalMetadata(Map<String, Object> metadata) {
+        this.metadata = Map.copyOf(metadata);
+    }
+
+    public Map<String, Object> getMetadata() {
+        return metadata;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -8,7 +8,9 @@ import org.slf4j.LoggerFactory;
 import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.commons.model.SupportedLanguageJpaConverter;
 import uk.gov.pay.connector.charge.model.CardDetailsEntity;
+import uk.gov.pay.connector.charge.model.ExternalMetadata;
 import uk.gov.pay.connector.charge.model.ServicePaymentReference;
+import uk.gov.pay.connector.charge.util.ExternalMetadataConverter;
 import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
 import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
 import uk.gov.pay.connector.common.model.api.ExternalChargeState;
@@ -132,21 +134,25 @@ public class ChargeEntity extends AbstractVersionedEntity {
     @Column(name = "wallet")
     @Enumerated(EnumType.STRING)
     private WalletType walletType;
-    
+
+    @Column(name = "external_metadata", columnDefinition = "jsonb")
+    @Convert(converter = ExternalMetadataConverter.class)
+    private ExternalMetadata externalMetadata;
+
     public ChargeEntity() {
         //for jpa
     }
 
     public ChargeEntity(Long amount, String returnUrl, String description, ServicePaymentReference reference,
                         GatewayAccountEntity gatewayAccount, String email, SupportedLanguage language,
-                        boolean delayedCapture) {
-        this(amount, CREATED, returnUrl, description, reference, gatewayAccount, email, ZonedDateTime.now(ZoneId.of("UTC")), language, delayedCapture);
+                        boolean delayedCapture, ExternalMetadata externalMetadata) {
+        this(amount, CREATED, returnUrl, description, reference, gatewayAccount, email, ZonedDateTime.now(ZoneId.of("UTC")), language, delayedCapture, externalMetadata);
     }
 
     // Only the ChargeEntityFixture should directly call this constructor
     public ChargeEntity(Long amount, ChargeStatus status, String returnUrl, String description, ServicePaymentReference reference,
                         GatewayAccountEntity gatewayAccount, String email, ZonedDateTime createdDate, SupportedLanguage language,
-                        boolean delayedCapture) {
+                        boolean delayedCapture, ExternalMetadata externalMetadata) {
         this.amount = amount;
         this.status = status.getValue();
         this.returnUrl = returnUrl;
@@ -158,6 +164,7 @@ public class ChargeEntity extends AbstractVersionedEntity {
         this.email = email;
         this.language = language;
         this.delayedCapture = delayedCapture;
+        this.externalMetadata = externalMetadata;
     }
 
     public Long getId() {
@@ -218,6 +225,10 @@ public class ChargeEntity extends AbstractVersionedEntity {
 
     public String getProviderSessionId() {
         return providerSessionId;
+    }
+
+    public Optional<ExternalMetadata> getExternalMetadata() {
+        return Optional.ofNullable(externalMetadata);
     }
 
     public void setExternalId(String externalId) {
--- a/src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataConverter.java
+++ b/src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataConverter.java
@@ -1,0 +1,51 @@
+package uk.gov.pay.connector.charge.util;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.postgresql.util.PGobject;
+import uk.gov.pay.connector.charge.exception.ExternalMetadataConverterException;
+import uk.gov.pay.connector.charge.model.ExternalMetadata;
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Converter;
+import java.io.IOException;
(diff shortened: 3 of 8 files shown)


src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.pay.connector.charge.model;
+
+import java.util.Map;
+
+public class ExternalMetadata {
+
+    private Map<String, Object> metadata;
```
alexbishop1: OK, bear with me here… is it worth pondering maybe using a more specific type than `Object`. Since we support strings, booleans and numbers, we could have an `ExternalMetadataValue` interface:

```java
interface ExternalMetadataValue<T> {

    T getValue();

}
```

… with implementations like `StringExternalMetadataValue implements ExternalMetadataValue<String>` and `BooleanExternalMetadataValue implements ExternalMetadataValue<Boolean>`.

This would introduce the overhead of a wrapper around the actual value but I don’t think this would be too bad because the only things that are ever going to extract the value are the code that builds the response (and hopefully there’s some Jackson annotation magic we use to hide that away) and tests.

alexbishop1: Also, can `metadata` be `final`?

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
```diff
@@ -0,0 +1,16 @@
...
+
+    public ExternalMetadata(Map<String, Object> metadata) {
+        this.metadata = metadata;
```
alexbishop1: If this were `this.metadata = Map.copyOf(metadata)`, the knowledge that every `ExternalMetadata` instance definitely has an internal map and that map is an [unmodifiable map](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#unmodifiable) would make me feel warm and fuzzy inside.

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -220,6 +227,10 @@ public String getProviderSessionId() {
         return providerSessionId;
     }
 
+    public Optional<ExternalMetadata> getExternalMetadata() {
+        return Optional.ofNullable(externalMetadata);
```
alexbishop1, line 231: Now that `ExternalMetadata` wraps a map, I wonder if it might be better for this method to always return a non-null `ExternalMetadata` object (with it wrapping an empty map if there’s no metadata). Not sure about this though — probably depends a lot on how callers of this method are going to use it.

src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataConverter.java
```diff
@@ -0,0 +1,47 @@
...
+@Converter
+public class ExternalMetadataConverter implements AttributeConverter<ExternalMetadata, PGobject> {
+    
+    private final ObjectMapper mapper = new ObjectMapper();
+    
+    @Override
+    public PGobject convertToDatabaseColumn(ExternalMetadata externalMetadata) {
+        PGobject pgObject = new PGobject();
+        pgObject.setType("jsonb");
+        var metaData = externalMetadata == null ? null : externalMetadata.getMetadata();
```
alexbishop1: I think in this case declaring `metaData` (which shouldn’t have a capital D!) is a `Map` rather than using `var` might be clearer

src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataConverter.java
```diff
@@ -0,0 +1,47 @@
...
+        try {
+            pgObject.setValue(mapper.writeValueAsString(metaData));
+        } catch (JsonProcessingException | SQLException e) {
+            throw new ExternalMetadataConverterException("Failed to serialise externalMetadata");
```
alexbishop1, line 33: It might be useful to include the unserialisable (which my spell checker wants to change to “unrealisable” — that got existential quickly!) metadata in the exception message but we have to balance that against the probability that the metadata may contain sensitive information (e.g. personally-identifiable information), which we don’t want to end up in our logs.

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -957,6 +961,31 @@ public void shouldCreateANewChargeWithProviderSessionId() {
...
 
+    @Test
+    public void shouldCreateANewChargeWithExternalMetadata() throws IOException {
+        GatewayAccountEntity gatewayAccount = new GatewayAccountEntity(defaultTestAccount.getPaymentProvider(), new HashMap<>(), TEST);
+        gatewayAccount.setId(defaultTestAccount.getAccountId());
+
+        ExternalMetadata expectedExternalMetadata = new ExternalMetadata(
+                Map.of("key1", "String1",
+                        "key2", 123,
+                        "key3", true));
```
alexbishop1, line 972: I still find it hard to believe we lived for so long without the succinct elegance of `Map.of(…)`

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
```diff
@@ -957,6 +961,31 @@ public void shouldCreateANewChargeWithProviderSessionId() {
...
+
+        ChargeEntity chargeEntity = aValidChargeEntity()
+                .withId(null)
```
alexbishop1: Why does the ID have to be null?

alexbishop1 (changes requested):
Looks good — left a few suggestions to consider

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
danworth: yes it can be, good spot. I'll come back to the discussion about a type rather than Object....

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
danworth: Ha, I did think about doing this and then I thought that I/people would think I was being paranoid taking a defensive copy. But, yes I'll gladly do it!

src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataConverter.java
danworth: ok, perhaps just because I was familiar with it being a map I thought `var` appropriate and exciting...but yes, I'm in favour of using `var` only when the type doesn't add any extra clarity, seems here it does. Changed.

src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataConverter.java
danworth: Yes I thought we wanted to be careful about what we log. I'll hold off on this for now.

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
danworth: Ha yes!

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoITest.java
danworth: Shamefully my answer is, because all the other tests are like that...I've removed it and the test passes.

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
danworth: I agree there's a benefit to a custom type for the value, and as you say it introduces the overhead of wrapping existing types. Can we come back to this after I've got the implementation working? The validation I'm working on will for the time being offer some defence of other values being used.

danworth (commented):
Thanks Alex there's some great points in there and its a better implementation for it. I'll push the changes shortly.

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
danworth: Since we don't wish to return `metadata` if it isn't set then perhaps using Optional will make that simpler when we come to add this to the API layer. I'll hold off for now if that's ok.

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
alexbishop1: Yes, that’s fine

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
alexbishop1: Defensive copies make me feel warm and fuzzy inside

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
alexbishop1: Yes, I agree

src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataConverter.java
alexbishop1: Yeah, we can always revisit if it turns out we actually get serialisation problems

alexbishop1 (approved):

PR #1214, opened 2019-04-17
PP-5108 Include allow_zero_amount in get gateway account response

stephencdaly:
Co-Authored-By Yancoba Thompson <yancoba.thompson@digital.cabinet-office.gov.uk>

### Logging

### Documentation

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -110,24 +110,47 @@ Content-Type: application/json
     "corporate_credit_card_surcharge_amount": 250,
     "corporate_debit_card_surcharge_amount": 50
     "corporate_prepaid_credit_card_surcharge_amount": 250,
-    "corporate_prepaid_debit_card_surcharge_amount": 50
+    "corporate_prepaid_debit_card_surcharge_amount": 50,
+    "allow_apple_pay": false,
+    "allow_google_pay": false,
+    "allow_zero_amount": false,
+    "email_collection_mode": "MANDATORY",
+    "toggle_3ds": false,
+    "email_notifications": {
+        "REFUND_ISSUED": {
+            "version": 1,
+            "enabled": true,
+            "template_body": null
+        },
+        "PAYMENT_CONFIRMED": {
+            "version": 1,
+            "enabled": true,
+            "template_body": null
+        }
+    }
 }
 ```
 
 #### Response field description
 
-| Field                                              | always present   | Description                                                                                       |
-| -------------------------------------------------- | ---------------- | ------------------------------------------------------------------------------------------------- |
-| `gateway_account_id`                               | X                | The account Id                                                                                    |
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -42,41 +42,7 @@
...
-                    return typeEnum;
-                }
-            }
-            throw new IllegalArgumentException("gateway account type has to be one of (test, live)");
-        }
-    }
-
-    public GatewayAccountEntity() {
-    }
-
```
stephencdaly, line 79: moved within file to please codacy

docs/api_specification.md
```diff
@@ -110,24 +110,47 @@ Content-Type: application/json
     "corporate_credit_card_surcharge_amount": 250,
     "corporate_debit_card_surcharge_amount": 50
     "corporate_prepaid_credit_card_surcharge_amount": 250,
-    "corporate_prepaid_debit_card_surcharge_amount": 50
+    "corporate_prepaid_debit_card_surcharge_amount": 50,
+    "allow_apple_pay": false,
```
danworth, line 114: Thanks for updating the docs! 👍

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
```diff
@@ -73,7 +73,8 @@ public void shouldGetGatewayAccountForExistingAccount() {
                 .body("corporate_credit_card_surcharge_amount", is(250))
                 .body("corporate_debit_card_surcharge_amount", is(50))
                 .body("allow_apple_pay", is(true))
-                .body("allow_google_pay", is(false));
+                .body("allow_google_pay", is(false))
```
danworth, line 77: I'd be inclined to test the positive (non-default) case in at least one of the ITests since the default might show success even if there's a problem setting a provided non-default value (if that makes sense).

danworth (changes requested):
Looks good, there a tiny point about perhaps it being beneficial to test the non-default case in at least one of the  ITests (e.g setting `allow_zero_payments` to true). I'm surprised that we're returning the `GatewayAccounEntity` in the response as we're usually quite strict at separating the layers, but on the plus side there's fewer classes to change and keep in sync when adding a new attributes so.... 🤷‍♂️

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceITest.java
stephencdaly: Good point. Made that change now

danworth (approved):
Looks good 👍

PR #1216, opened 2019-04-17
PP-4884 Add validation for ExternalMetadata

danworth:
## WHAT YOU DID
- Add validation to `ExternalMetadata` based on composing validations which gives us a fine grain control and flexibility of messages/ limits.
- Create custom validations for map key length, map value length and map value types.
- I would like to add this as-is (subject to review comments of course) to progress the story and unblock future work. I will revisit with a view of increasing the sophistication of the error messages to to reference the keys when a value is invalid.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/util/MapValueTypesValidator.java
+++ b/src/main/java/uk/gov/pay/connector/util/MapValueTypesValidator.java
@@ -1,0 +1,33 @@
+package uk.gov.pay.connector.util;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+public class MapValueTypesValidator implements ConstraintValidator<MapValueTypes, Map<String, Object>> {
+
+    private Set<Class> allowedTypes;
+
+    @Override
+    public void initialize(MapValueTypes constraintAnnotation) {
+        allowedTypes = Set.of(constraintAnnotation.types());
+    }
+
+    @Override
+    public boolean isValid(Map<String, Object> theMap, ConstraintValidatorContext context) {
+        if (theMap == null) {
+            return true;
+        }
+
+        return theMap.values().stream()
+                .filter(Objects::nonNull)
+                .map(Object::getClass)
+                .allMatch(this::validType);
+    }
+
+    private boolean validType(Class clazz) {
+        return allowedTypes.stream().anyMatch(permittedClass -> permittedClass.isAssignableFrom(clazz));
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/util/ValidExternalMetadata.java
+++ b/src/main/java/uk/gov/pay/connector/util/ValidExternalMetadata.java
@@ -1,0 +1,34 @@
+package uk.gov.pay.connector.util;
+
+import javax.validation.Constraint;
+import javax.validation.Payload;
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({FIELD, PARAMETER})
+@Retention(RUNTIME)
+@Constraint(validatedBy = {})
+@NotNull(message = "metadata must not be null")
+@Size(max = 10, message = "metadata cannot have more than {max} key-value pairs")
+@MapKeyLength(max = 30, min = 1, message = "metadata keys must be between {min} and {max} characters long")
+@MapValueTypes(types = {String.class, Number.class, Boolean.class}, message = "metadata values must be of type String, Boolean or Number")
+@MapValueLength(max = 50, min = 0, message = "metadata values must be no greater than {max} characters long")
+public @interface ValidExternalMetadata {
+    String message() default "Invalid metadata";
+
+    Class<?>[] groups() default {};
+
+    Class<? extends Payload>[] payload() default {};
+
+    @Target({FIELD, PARAMETER})
+    @Retention(RUNTIME)
+    @interface List {
+        ValidExternalMetadata[] value();
+    }
+}
--- a/src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
@@ -1,0 +1,130 @@
+package uk.gov.pay.connector.charge.model;
+
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.Test;
+
+import javax.validation.ConstraintViolation;
+import javax.validation.Validation;
+import javax.validation.Validator;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+
+public class ExternalMetadataTest {
+
+    private final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    private final String TOO_LONG_VALUE = "This value is over fifty characters long and is invalid!";
+    private final String TOO_LONG_KEY = "This key is over thirty characters long and is invalid!";
+
+    @Test
+    public void shouldPassValidation() {
+        Map<String, Object> metadata = Map.of(
+                "key1", "string",
+                "key2", true,
+                "key3", 123,
+                "key4", 1.234,
+                "key5", ""
+        );
+        ExternalMetadata validExternalMetadata = new ExternalMetadata(metadata);
+
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(validExternalMetadata);
+
+        assertThat(violations.size(), is(0));
+    }
+
+    @Test
+    public void shouldFailValidationWithMoreThan10Keys() {
+        Map<String, Object> metadata = IntStream.rangeClosed(1, 11).boxed()
+                .collect(Collectors.toUnmodifiableMap(i -> "key" + i, i-> "value" + i));
+        ExternalMetadata invalidExternalMetadata = new ExternalMetadata(metadata);
+
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);
+
+        assertThat(violations.size(), is(1));
+        assertThat(violations.iterator().next().getMessage(), is("metadata cannot have more than 10 key-value pairs"));
+    }
+
+    @Test
+    public void shouldFailValidationForKeysToLong() {
+        Map<String, Object> metadata = Map.of(TOO_LONG_KEY, "string");
+        ExternalMetadata invalidExternalMetadata = new ExternalMetadata(metadata);
+
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);
+
+        assertThat(violations.size(), is(1));
+        assertThat(violations.iterator().next().getMessage(), is("metadata keys must be between 1 and 30 characters long"));
+    }
+
+    @Test
+    public void shouldFailValidationForEmptyKey() {
+        Map<String, Object> metadata = Map.of("", "string");
+        ExternalMetadata invalidExternalMetadata = new ExternalMetadata(metadata);
+
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);
+
+        assertThat(violations.size(), is(1));
+        assertThat(violations.iterator().next().getMessage(), is("metadata keys must be between 1 and 30 characters long"));
+    }
+
+    @Test
+    public void shouldFailValidationForValueToLong() {
+        Map<String, Object> metadata = Map.of("key1", TOO_LONG_VALUE);
+        ExternalMetadata invalidExternalMetadata = new ExternalMetadata(metadata);
+
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);
+
+        assertThat(violations.size(), is(1));
+        assertThat(violations.iterator().next().getMessage(), is("metadata values must be no greater than 50 characters long"));
+    }
+
+    @Test
+    public void shouldFailValidationWhenValueIsObject() {
+        Map<String, Object> metadata = Map.of("key1", mapper.createObjectNode());
+        ExternalMetadata invalidExternalMetadata = new ExternalMetadata(metadata);
+
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);
+
+        assertThat(violations.size(), is(1));
+        assertThat(violations.iterator().next().getMessage(), is("metadata values must be of type String, Boolean or Number"));
+    }
+
+    @Test
+    public void shouldFailValidationWhenValueIsArray() {
+        Map<String, Object> metadata = Map.of("key1", mapper.createArrayNode());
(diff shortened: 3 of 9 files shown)


src/main/java/uk/gov/pay/connector/util/MapKeySize.java
```diff
@@ -0,0 +1,32 @@
...
+@Retention(RUNTIME)
+@Constraint(validatedBy = MapKeySizeValidator.class)
+public @interface MapKeySize {
+    String message() default "keys must be between {min} and {max}";
+
+    Class<?>[] groups() default { };
+
+    Class<? extends Payload>[] payload() default { };
+
+    int min() default 0;
```
alexbishop1: Isn’t the minimum one character for keys?

src/main/java/uk/gov/pay/connector/util/MapKeySizeValidator.java
```diff
@@ -0,0 +1,30 @@
...
+    @Override
+    public boolean isValid(Map<String, Object> theMap, ConstraintValidatorContext context) {
+        if (theMap == null) {
+            return true;
+        }
+
+        return theMap.keySet().stream()
+                .filter(key -> key.length() < min || key.length() > max)
+                .collect(Collectors.toList())
+                .size() == 0;
```
alexbishop1: I’m not joking — this is literally the second time in two days I’ve seen some code that filters a stream, collects to a list and then checks the size of the list. So take comfort in the fact that you’re in good company when I say there’s an easier way to do this:

```java
return theMap.keySet().stream().noneMatch(key -> key.length() < min || key.length() > max);
```

src/main/java/uk/gov/pay/connector/util/MapValueAsStringSizeValidator.java
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    @Override
+    public boolean isValid(Map<String, Object> theMap, ConstraintValidatorContext context) {
+        if (theMap == null) {
+            return true;
+        }
+
+        return theMap.values().stream()
+                .filter(Objects::nonNull)
```
alexbishop1: This will essentially remove null values when validating rather than their presence causing an error, which I don’t think is what we want

src/main/java/uk/gov/pay/connector/util/MapValueAsStringSizeValidator.java
```diff
@@ -0,0 +1,33 @@
...
+                .map(Object::toString)
+                .filter(value -> value.length() < min || value.length() > max)
+                .collect(Collectors.toList())
+                .size() == 0;
```
alexbishop1: Another `noneMatch(…)` opportunity

src/main/java/uk/gov/pay/connector/util/MapValueTypesValidator.java
```diff
@@ -0,0 +1,29 @@
...
+    @Override
+    public boolean isValid(Map<String, Object> theMap, ConstraintValidatorContext context) {
+        if (theMap == null) {
+            return true;
+        }
+
+        return theMap.values().stream()
+                .filter(value -> value == null || !allowedTypes.contains(value.getClass()))
+                .collect(Collectors.toList())
+                .size() == 0;
```
alexbishop1: Another fine `noneMatch(…)` opportunity

src/main/java/uk/gov/pay/connector/util/ValidExternalMetadata.java
```diff
@@ -0,0 +1,34 @@
...
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target({FIELD, PARAMETER})
+@Retention(RUNTIME)
+@Constraint(validatedBy = {})
+@NotNull(message = "metadata must not be null")
+@Size(max = 10, message = "metadata cannot have more than {max} key-value pairs")
+@MapKeySize(max = 30, min = 1, message = "metadata keys must be between {min} and {max} characters long")
+@MapValueTypes(types = {String.class, Integer.class, Boolean.class}, message = "metadata values must be of type String, Boolean or Integer")
```
alexbishop1: Didn’t we say any JSON number rather than just integers?

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
```diff
@@ -0,0 +1,109 @@
...
+    private static Validator validator;
+    private final ObjectMapper mapper = new ObjectMapper();
+
+    private final String TOO_LONG_VALUE = "This value is over fifty characters long and is invalid";
+    private final String TOO_LONG_KEY = "This key is over thirty characters long and is invalid";
+
+    @BeforeClass
+    public static void setUp() {
+        ValidatorFactory vf = Validation.buildDefaultValidatorFactory();
+        validator = vf.getValidator();
```
alexbishop1: Can this be inlined into the `validator` field declaration?

```java
private final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
```

This will re-initialise `validator` for every test but I assume it’s not too expensive to create?

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
```diff
@@ -0,0 +1,109 @@
...
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(validExternalMetadata);
+        assertThat(violations.size(), is(0));
+    }
+
+    @Test
+    public void shouldFailValidationWithMoreThan10Keys() {
+        Map<String, Object> metadata = new HashMap<>();
+        for (int i = 0; i < 11; i++) {
+            metadata.put("key" + i, "value" + i);
+        }
```
alexbishop1: An alternative way to initialise this map could be:

```java
Map<String, Object> metadata = IntStream.rangeClosed(1, 11).boxed()
        .collect(Collectors.toUnmodifiableMap(i -> "key" + i, i -> "value" + i));
```

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
```diff
@@ -0,0 +1,109 @@
...
+        assertThat(violations.size(), is(1));
+        assertThat(violations.iterator().next().getMessage(), is("metadata values must be no greater than 50 characters long"));
+    }
+
+    @Test
+    public void shouldFailValidationInvalidValueTypes() {
+        Map<String, Object> metadata = Map.of(
+                "key1", "string",
+                "key2", mapper.createArrayNode(),
+                "key3", mapper.createObjectNode()
```
alexbishop1: I think I’d prefer if this was two tests (one for array node and one for object node). Right now, it will pass even if the code were only rejecting array nodes and allowing object nodes, for example.

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
```diff
@@ -0,0 +1,109 @@
...
+                "metadata values must be of type String, Boolean or Integer",
+                "metadata values must be no greater than 50 characters long",
+                "metadata keys must be between 1 and 30 characters long");
+
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);
+        assertThat(violations.size(), is(3));
+        assertThat(expectedErrorMessages.contains(violations.iterator().next().getMessage()), is(true));
+        assertThat(expectedErrorMessages.contains(violations.iterator().next().getMessage()), is(true));
+        assertThat(expectedErrorMessages.contains(violations.iterator().next().getMessage()), is(true));
+    }
```
alexbishop1, line 129: Can we have some tests for some other cases?

- Null values (should pass)
- Empty string keys (should fail)
- Empty string values (should pass)

alexbishop1 (changes requested):
Looks good — a few things to look at

src/main/java/uk/gov/pay/connector/util/MapValueAsStringSizeValidator.java
alexbishop1: Actually, I’ve now realised that since we allow null values, this will actually do the right thing

src/main/java/uk/gov/pay/connector/util/MapKeySize.java
danworth: Its possible to use an empty string as a key in a map. Since this validation could be used elsewhere I thought best to set the default to 0 to permit "" as a key. When we use this annotation to compose `ValidExternalMetadata` I set the `min=1` since for our purposes we're not allowing an empty string as a key.

src/main/java/uk/gov/pay/connector/util/MapKeySizeValidator.java
danworth: Thank you! I will use `noneMatch` and spread the word.

src/main/java/uk/gov/pay/connector/util/MapValueAsStringSizeValidator.java
danworth: yep. There's a point of best practice with bean validation that nulls shouldn't be considered invalid unless the check is specifically designed to treat them as such. This allows us to create compound annotations by re-using others and where necessary using something like `@NotNull` to catch the null case. This also allows it to be used for the instance of "I don't care if its null, but if it is present then it must be..."

src/main/java/uk/gov/pay/connector/util/MapValueAsStringSizeValidator.java
danworth: done, thanks!

src/main/java/uk/gov/pay/connector/util/MapValueTypesValidator.java
danworth: Done :)

src/main/java/uk/gov/pay/connector/util/ValidExternalMetadata.java
danworth: Yes, I've adapted this now to check if the type of the value is of a subtype to those permitted. So I've exchanged `Integer` for `Number`.

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
danworth: I'm happy to adopt new ways and this is more succinct without the boilerplate `for` loop. I've added it. Thanks.

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
danworth: Ok, done.

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
danworth: Interestingly, since we added the defensive copy with `Map.copyOf` it will not permit null values. So it isn't possible to create an instance of `ExternalMetadata` with null values. This will have to be caught as part of the deserialisation which will follow.

I've added empty string values (pass) and empty string keys (fail).

danworth: As demonstrated with `JShell`
```
jshell> myMap.put("key1", null)
$6 ==> null

jshell> var copyMap = Map.copyOf(myMap);
|  Exception java.lang.NullPointerException
|        at Objects.requireNonNull (Objects.java:221)
|        at ImmutableCollections$Map1.<init> (ImmutableCollections.java:732)
|        at Map.ofEntries (Map.java:1611)
|        at Map.copyOf (Map.java:1684)
|        at do_it$Aux (#7:1)
|        at (#7:1)
```

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
danworth: Ok, I'm on the fence for this but I've adopted it for the benefit of brevity (it feels a little wasteful to keep initialising it thought).

src/main/java/uk/gov/pay/connector/util/MapKeySize.java
alexbishop1: Ah, I see

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
alexbishop1: Of course. And we do want to disallow null values (I was getting confused with empty string values, which we do want to allow).

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Test
+    public void shouldFailValidationForKeysToLong() {
+
+        Map<String, Object> metadata = Map.of(TOO_LONG_KEY, "string");
+        ExternalMetadata invalidExternalMetadata = new ExternalMetadata(metadata);
+        Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);
+        assertThat(violations.size(), is(1));
+        assertThat(violations.iterator().next().getMessage(), is("metadata keys must be between 1 and 30 characters long"));
+    }
```
alexbishop1, line 64: This is just a suggestion but I quite like to put blank lines in tests so that there’s the setup/stubbing, then a blank line, then the actual call that you’re trying to test, then a blank line, then the assertions and verifications. I find it makes it easier to see what the test is testing at a glance and makes it a bit simpler to understand the differences between each test. Like this:

```java
@Test
public void shouldFailValidationForKeysToLong() {
    Map<String, Object> metadata = Map.of(TOO_LONG_KEY, "string");
    ExternalMetadata invalidExternalMetadata = new ExternalMetadata(metadata);

    Set<ConstraintViolation<ExternalMetadata>> violations = validator.validate(invalidExternalMetadata);

    assertThat(violations.size(), is(1));
    assertThat(violations.iterator().next().getMessage(), is("metadata keys must be between 1 and 30 characters long"));
}
```

alexbishop1 (changes requested):
This looks good. One thing (which I really should have mentioned in the first review, sorry) is that a lot of the new stuff for string validation uses the term “size” but the usual term for strings is “length”. Could you change it please? I know I’m being a bit picky but I think it would aid understanding.

danworth:
The reason I used `size` was to be consistent with the terms used by the Validation api which uses
`@Size` https://docs.oracle.com/javaee/7/api/javax/validation/constraints/Size.html. However, since the types we're validating are always Strings (for now), it might make more sense to refer to it as Length. I'll change it.

alexbishop1:
I didn’t realise “size” is an existing convention so I’m less bothered about it now.

src/test/java/uk/gov/pay/connector/charge/model/ExternalMetadataTest.java
danworth: I like this format, so I've updated to match.

src/main/java/uk/gov/pay/connector/util/MapValueTypesValidator.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public boolean isValid(Map<String, Object> theMap, ConstraintValidatorContext context) {
+        if (theMap == null) {
+            return true;
+        }
+
+        return theMap.values().stream()
+                .filter(Objects::nonNull)
+                .noneMatch(value -> !validType(value.getClass()));
```
alexbishop1: Would `.allMatch(value -> validType(value.getClass()))` work here too?

alexbishop1: Or maybe even make the whole thing into this:

 ```java
return theMap.values().stream()
        .filter(Objects::nonNull)
        .map(Object::getClass)
        .allMatch(this::validType);
```

alexbishop1 (dismissed):
Looks good — one final suggestion but it’s fairly minor

alexbishop1 (approved):

PR #1217, opened 2019-04-17
PP-5108 Allow update of allow_zero_amount flag

stephencdaly:
Add valid path to gateway account patch request which updates the allow_zero_amount flag on a gateway account.

Co-Authored-By Yancoba Thompson <yancoba.thompson@digital.cabinet-office.gov.uk>

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/api/jsonpatch/JsonPatchRequest.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/api/jsonpatch/JsonPatchRequest.java
@@ -28,10 +28,7 @@ public class JsonPatchRequest {
     }
 
     public String valueAsString() {
-        if (value != null && value.isTextual()) {
-            return value.asText();
-        }
-        throw new JsonNodeNotCorrectTypeException("JSON node " + value + " is not of type string");
+        return value.asText();
     }
     
     public long valueAsLong() {
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -36,6 +36,8 @@ public class GatewayAccountRequestValidator {
     public static final String FIELD_CORPORATE_DEBIT_CARD_SURCHARGE_AMOUNT = "corporate_debit_card_surcharge_amount";
     public static final String FIELD_CORPORATE_PREPAID_CREDIT_CARD_SURCHARGE_AMOUNT = "corporate_prepaid_credit_card_surcharge_amount";
     public static final String FIELD_CORPORATE_PREPAID_DEBIT_CARD_SURCHARGE_AMOUNT = "corporate_prepaid_debit_card_surcharge_amount";
+    public static final String FIELD_ALLOW_ZERO_AMOUNT = "allow_zero_amount";
+    
     private static final List<String> VALID_PATHS = asList(
             CREDENTIALS_GATEWAY_MERCHANT_ID,
             FIELD_NOTIFY_SETTINGS, 
@@ -45,7 +47,8 @@ public class GatewayAccountRequestValidator {
             FIELD_CORPORATE_CREDIT_CARD_SURCHARGE_AMOUNT,
             FIELD_CORPORATE_DEBIT_CARD_SURCHARGE_AMOUNT,
             FIELD_CORPORATE_PREPAID_CREDIT_CARD_SURCHARGE_AMOUNT,
-            FIELD_CORPORATE_PREPAID_DEBIT_CARD_SURCHARGE_AMOUNT);
+            FIELD_CORPORATE_PREPAID_DEBIT_CARD_SURCHARGE_AMOUNT,
+            FIELD_ALLOW_ZERO_AMOUNT);
     
     private final RequestValidator requestValidator;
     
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/common/model/api/jsonpatch/JsonPatchRequest.java
```diff
@@ -28,10 +28,7 @@ public String getPath() {
     }
 
     public String valueAsString() {
-        if (value != null && value.isTextual()) {
-            return value.asText();
-        }
-        throw new JsonNodeNotCorrectTypeException("JSON node " + value + " is not of type string");
+        return value.asText();
```
stephencdaly, line 31: Removed this validation as the existing patch requests for updating boolean values on gateway accounts (allow_google_pay and allow_apply_pay) allow string representations of booleans (and selfservice sends a string!) and so use this method. 
They probably shouldn't but we need to keep backwards compatibility for now until that's changed and allow both text and boolean values for now.

alexbishop1: I don’t think I follow why this is necessary — can’t `GatewayAccountService` call `valueAsBoolean()` for `allow_zero_amount` (but continue calling `valueAsString()` for `allow_google_pay` and `allow_apple_pay`?

This change would also mean that one could call `valueAsString()` on a numeric value and get its stringification, which is more loose than I think we want.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
```diff
@@ -131,7 +137,7 @@ private void validateEmailCollectionMode(JsonNode payload) {
         }
     }
 
-    private void validateAllowWebPayment(JsonNode payload, String field) {
+    private void validateReplaceBooleanValue(JsonNode payload, String field) {
```
alexbishop1, line 140: I think I’d prefer if we didn’t try to use the same validation for `allow_zero_amount` (which will be a boolean) and `allow_google_pay` and `allow_apple_pay` (which will be a string of either `"true"` or `"false"`). I think we’d be better off with two separate methods (with a view to fixing the abomination that is booleans-as-strings for Web payments config in the future).

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
```diff
@@ -65,31 +72,6 @@ public void shouldThrowWhenWebPaymentRequestsAreInvalid(String op, String path,
...
-        
-        JsonNode jsonNode = new ObjectMapper().valueToTree(patch);   
-        try{
-            validator.validatePatchRequest(jsonNode);
-            fail( "Expected ValidationException" );
-        } catch (ValidationException e) {
-            assertThat(e.getErrors().size(), is(1));
-            assertThat(e.getErrors(), hasItems(expectedErrorMessage));
-        }
-    }
```
alexbishop1, line 92: I see this has been removed because it’s now handled by the expanded and renamed `shouldThrowWhenRequestsAreInvalid` test.

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
```diff
@@ -211,7 +211,22 @@ public void shouldNotAllowDigitalWalletForUnsupportedGateways() {
...
+        JsonPatchRequest request = JsonPatchRequest.from(new ObjectMapper().valueToTree(ImmutableMap.of("op", "replace",
+                "path", "allow_zero_amount",
+                "value", true)));
+        GatewayAccountEntity entity = mock(GatewayAccountEntity.class);
+
+        when(mockGatewayAccountDao.findById(gatewayAccountId)).thenReturn(Optional.of(entity));
+        Optional<GatewayAccount> optionalGatewayAccount = gatewayAccountService.doPatch(gatewayAccountId, request);
+        assertThat(optionalGatewayAccount.isPresent(), is(true));
+        verify(entity).setAllowZeroAmount(true);
+        verify(mockGatewayAccountDao).merge(entity);
```
alexbishop1, line 230: Since there’s only two possible values, I don’t think it would be too much of a stretch to also have a test for setting it to `false` (with the gateway account having it set to `true` at the beginning so we know it’s actually changing something).

alexbishop1 (changes requested):
Looks good in general but I have a few qualms about how it validates

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
cobainc0: 👍 done

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
cobainc0: As discussed, we will keep this for now with a view to updating `allow_google_pay` and `allow_apple_pay` to booleans rather than Strings.

src/main/java/uk/gov/pay/connector/common/model/api/jsonpatch/JsonPatchRequest.java
alexbishop1: As discussed, we’ll have this go in as it is, then change `allow_google_pay` and `allow_apple_pay` to use real (not stringified!) booleans, then change this back

alexbishop1 (approved):

PR #1221, opened 2019-04-18
PP-4884 Add ExernalMetadata to create charge request

danworth:
- Add `ExternalMetadata` to `ChargeCreateRequest`
- Add `ExternalMetadata` to `ChargeService`
- Add test to `ChargeServiceTest`
- Add `ChargesApiCreateResourceITest` tests for externalMetadata.
- Add `ExternalMetadataDeserialiser`
- Add validation for map values being null.

## WHAT YOU DID
This does not include returning the metadata in the response which is covered by PP-4885. The ITest therefore checks that is has been persisted to the database and stops short of checking it is present in the successful response.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
@@ -2,18 +2,20 @@ package uk.gov.pay.connector.charge.model;
 
 import uk.gov.pay.connector.util.ValidExternalMetadata;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
 public class ExternalMetadata {
-    
+
     @ValidExternalMetadata
     private final Map<String, Object> metadata;
 
     public ExternalMetadata(Map<String, Object> metadata) {
-        this.metadata = Map.copyOf(metadata);
+        this.metadata = new HashMap<>(metadata);
     }
 
     public Map<String, Object> getMetadata() {
-        return metadata;
+        return Collections.unmodifiableMap(metadata);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -129,26 +129,24 @@ public class ChargeService {
                     chargeRequest.getEmail(),
                     language,
                     chargeRequest.isDelayedCapture(),
-                    null);
+                    chargeRequest.getExternalMetadata().orElse(null));
             
-            if (chargeRequest.getPrefilledCardHolderDetails().isPresent()) {
-                PrefilledCardHolderDetails cardHolderDetails = chargeRequest.getPrefilledCardHolderDetails().get();
-                CardDetailsEntity cardDetailsEntity = new CardDetailsEntity();
-                if (cardHolderDetails.getCardHolderName().isPresent()) {
-                    cardDetailsEntity.setCardHolderName(cardHolderDetails.getCardHolderName().get());
-                }
-                if (cardHolderDetails.getAddress().isPresent()) {
-                    AddressEntity addressEntity = new AddressEntity(cardHolderDetails.getAddress().get());
-                    cardDetailsEntity.setBillingAddress(addressEntity);
-                }
-                chargeEntity.setCardDetails(cardDetailsEntity);
-            }
-            
+            chargeRequest.getPrefilledCardHolderDetails()
+                    .map(this::createCardDetailsEntity)
+                    .ifPresent(chargeEntity::setCardDetails);
+
             chargeDao.persist(chargeEntity);
 
             chargeEventDao.persistChargeEventOf(chargeEntity);
-            return Optional.of(populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, chargeEntity).build());
-        }).orElseGet(Optional::empty);
+            return populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, chargeEntity).build();
+        });
+    }
+
+    private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails prefilledCardHolderDetails) {
+        CardDetailsEntity cardDetailsEntity = new CardDetailsEntity();
+        prefilledCardHolderDetails.getCardHolderName().ifPresent(cardDetailsEntity::setCardHolderName);
+        prefilledCardHolderDetails.getAddress().map(AddressEntity::new).ifPresent(cardDetailsEntity::setBillingAddress);
+        return cardDetailsEntity;
     }
 
     @Transactional
--- a/src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataDeserialiser.java
+++ b/src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataDeserialiser.java
@@ -1,0 +1,25 @@
+package uk.gov.pay.connector.charge.util;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import uk.gov.pay.connector.charge.model.ExternalMetadata;
+
+import java.io.IOException;
+import java.util.Map;
+
+public class ExternalMetadataDeserialiser extends JsonDeserializer<ExternalMetadata> {
+
+    @Override
+    public ExternalMetadata deserialize(final JsonParser jsonParser, final DeserializationContext ctxt) throws IOException {
+        Map<String, Object> metadata = jsonParser.getCodec().readValue(jsonParser, new TypeReference<Map<String, Object>>() {});
+        if (metadata != null) {
+            return new ExternalMetadata(metadata);
+        }
+
+        assert false : "This should never be invoked since we currently do no deserialize null values.";
+        throw new JsonMappingException(jsonParser, "metadata cannot be null");
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
@@ -1,0 +1,25 @@
+package uk.gov.pay.connector.util;
+
+import com.fasterxml.jackson.databind.JsonMappingException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.annotation.Priority;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import java.util.Map;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Priority(1)
+public class JsonMappingExceptionMapper implements ExceptionMapper<JsonMappingException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(JsonMappingExceptionMapper.class);
+
+    @Override
+    public Response toResponse(JsonMappingException exception) {
+        LOGGER.error(exception.getMessage());
+        Map<String, String> entity = Map.of("message", exception.getMessage());
+        return Response.status(400).entity(entity).type(APPLICATION_JSON).build();
+    }
+}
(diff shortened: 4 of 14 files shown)


danworth:
The failing Codacy check is for `JUnit tests should include assert() or fail()` which isn't always appropriate for us when we're testing with a `RestAssured.ValidatableResponse`

alexbishop1:
Yeah, we need to update the Codacy rules to better match our circumstances. They’re basically just the defaults right now.

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
```diff
@@ -5,12 +5,12 @@
...
 
 public class ExternalMetadata {
-    
+
     @ValidExternalMetadata
     private final Map<String, Object> metadata;
 
     public ExternalMetadata(Map<String, Object> metadata) {
-        this.metadata = Map.copyOf(metadata);
+        this.metadata = metadata;
```
alexbishop1: Were we not going to move the `Map.copyOf(…)` to the getter?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -129,7 +129,7 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
                     chargeRequest.getEmail(),
                     language,
                     chargeRequest.isDelayedCapture(),
-                    null);
+                    chargeRequest.getExternalMetadata().orElse(null));
```
alexbishop1, line 132: If it’s not too much bother, I think it might be good to have two constructors: one with external metadata and one without)

src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataDeserialiser.java
```diff
@@ -0,0 +1,19 @@
...
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import uk.gov.pay.connector.charge.model.ExternalMetadata;
+
+import java.io.IOException;
+import java.util.Map;
+
+public class ExternalMetadataDeserialiser extends JsonDeserializer<ExternalMetadata> {
+
+    @Override
+    public ExternalMetadata deserialize(final JsonParser p, final DeserializationContext ctxt) throws IOException {
```
alexbishop1: Can we use something a bit more descriptive than `p`?

https://www.youtube.com/watch?v=wi-H6ohY37k

src/main/java/uk/gov/pay/connector/util/ValidExternalMetadata.java
```diff
@@ -14,11 +14,12 @@
...
-@Size(max = 10, message = "metadata cannot have more than {max} key-value pairs")
-@MapKeyLength(max = 30, min = 1, message = "metadata keys must be between {min} and {max} characters long")
-@MapValueTypes(types = {String.class, Number.class, Boolean.class}, message = "metadata values must be of type String, Boolean or Number")
-@MapValueLength(max = 50, min = 0, message = "metadata values must be no greater than {max} characters long")
+@NotNull(message = "Field [metadata] value must not be null")
+@Size(max = 10, message = "Field [metadata] cannot have more than {max} key-value pairs")
+@MapKeyLength(max = 30, min = 1, message = "Field [metadata] keys must be between {min} and {max} characters long")
+@MapValueTypes(types = {String.class, Number.class, Boolean.class}, message = "Field [metadata] values must be of type String, Boolean or Number")
+@MapValueLength(max = 50, message = "Field [metadata] values must be no greater than {max} characters long")
+@MapValueNotNull(message = "Field [metadata] values must not be null")
```
alexbishop1: This is identical to the `@NotNull` error message. Maybe use `Field [metadata] must not be null` for that one and `Field [metadata] must not have null values` for this one?

alexbishop1 (changes requested):
Looks good — just a few small things

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
danworth: I feel that would make it more complicated for not a huge gain? We would have to check whether getExternalMetadata is present before then selecting the correct version? I think the optional.orElse make its easy to see what's going on and it saves having two very similar constructors and keeps it more succinct.

src/main/java/uk/gov/pay/connector/charge/util/ExternalMetadataDeserialiser.java
danworth: yes, good idea.

src/main/java/uk/gov/pay/connector/util/ValidExternalMetadata.java
danworth: Not quite identical, one has `value` the other `values` but I see your point. I'll update, thanks 👍

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
alexbishop1: As discussed, maybe it’s probably not worth having `Map.copyOf(…)` in the getter.

It makes me sad that we can’t make a private unmodifiable map from the source map because that would disallow null values and a pre-validation `ExternalMetadata` might have null values (it’s a real shame that the validation works this way because it makes it harder to have types that maintain their invariants — particularly in this case where the fact we’re converting a map into an `ExternalMetadata` means the `ExternalMetadata` doesn’t have the annotations on it so it’s not obvious what the constraints are).

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Using a separate constructor would make it clear that a charge can be created without metadata. Though the fact that the getter returns an optional hopefully makes this clear.

I suppose the other alternative would be to have a `ChargeEntityBuilder`, which could be used like this:

```java
ChargeEntityBuilder chargeEntityBuilder = new ChargeEntityBuilder(
        chargeRequest.getAmount(),
        chargeRequest.getReturnUrl(),
        chargeRequest.getDescription(),
        ServicePaymentReference.of(chargeRequest.getReference()),
        gatewayAccount,
        chargeRequest.getEmail(),
        language,
        chargeRequest.isDelayedCapture());

chargeRequest.getExternalMetadata().ifPresent(ChargeEntityBuilder::withExternalMetadata);

ChargeEntity chargeEntity = chargeEntityBuilder.build();
```

alexbishop1 (dismissed):
Still a couple of things I’m a little sad about but following our discussion I’m happy to approve

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
```diff
@@ -2,18 +2,21 @@
...
 
 public class ExternalMetadata {
-    
+
     @ValidExternalMetadata
     private final Map<String, Object> metadata;
 
     public ExternalMetadata(Map<String, Object> metadata) {
-        this.metadata = Map.copyOf(metadata);
+        this.metadata = new HashMap<>(Objects.requireNonNull(metadata));
```
alexbishop1: `Objects.requireNonNull(metadata)` will just throw a `NullPointerException` if `metadata` is null, which is the same thing that will happen if the `HashMap` constructor is invoked with a null `metadata`, so I think this can just be `new HashMap(metadata)`

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
```diff
@@ -2,18 +2,21 @@
...
     }
 
     public Map<String, Object> getMetadata() {
-        return metadata;
+        return Collections.unmodifiableMap(metadata);
```
alexbishop1, line 19: I think I’d prefer it if this was `Map.copyOf(metadata)` so anyone can tell at a glance that it’s an [unmodifiable map](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Map.html#unmodifiable) (not to be confused with an [unmodifiable view collection](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Collection.html#unmodview)!) — in particular, that it will never have null values. (I recognise this will be slightly less efficient but I don’t think it’s going to make a noticeable difference for small maps we’d be dealing with.)

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: As discussed in person, we pinky-swear to add a builder if we add any more optional fields to `ChargeEntity` but will leave it for now

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -129,26 +129,24 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
...
-        }).orElseGet(Optional::empty);
+            return populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, chargeEntity).build();
+        });
+    }
+
+    private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails prefilledCardHolderDetails) {
+        CardDetailsEntity cardDetailsEntity = new CardDetailsEntity();
+        prefilledCardHolderDetails.getCardHolderName().ifPresent(cardDetailsEntity::setCardHolderName);
+        prefilledCardHolderDetails.getAddress().map(AddressEntity::new).ifPresent(cardDetailsEntity::setBillingAddress);
+        return cardDetailsEntity;
```
alexbishop1, line 149: This ended up much nicer!

src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Priority(1)
+public class JsonMappingExceptionMapper implements ExceptionMapper<JsonMappingException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(JsonMappingExceptionMapper.class);
+
+    @Override
+    public Response toResponse(JsonMappingException exception) {
+        LOGGER.error(exception.getMessage());
+        ImmutableMap<String, String> entity = ImmutableMap.of("message", exception.getMessage());
```
alexbishop1: Can just use `Map.of(…)` here

alexbishop1 (changes requested):
This is good. Just a few really small things.

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
danworth: Fair point, I've removed.

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
danworth: I felt that using `Collections.unmodifiableMap` had the benefit (relatively speaking) that it permits `null` values, so we don't end up in a position where we've accepted something through the constructor which we could never return, which makes me uneasy.

src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
danworth: Good catch

src/main/java/uk/gov/pay/connector/charge/model/ExternalMetadata.java
alexbishop1: Yeah, I can see your point. In an ideal world, it would be cool if we could convert the object into another object after validating. But we don’t live in an ideal world. So I’m happy with `Collections.unmodifiableMap(…)`.

It’s a shame it’s a bit non-obvious that we have to allow the class to be instantiated with null values but it will never actually have them because of the validation.

danworth: Yes agreed. It seems the `@Valid` within a REST API relies on Jackson having deseriliased it and then validating it so we must permit invalid values through 😞 . I tried moving the validation to within the custom deserializer and validate based on calling the constructor having annotated the `metadata` parameter with `@ValidExternalMetadata`; this can be done manually with `ExecutableValidator.validateConstructorParameters()`. This validated it prior to instantiating the object and caught null values then throwing a `ConstraintViolationException`, but unfortunately Jackson catches any exception during marshalling and wraps in a `JsonProcessingException`. It also felt that moving the validation logic into the deserialisation wouldn't be obvious to someone in the future since `@Valid` would not invoke it (or if the validation was on both the constructor parameter and the class attribute, it would be called twice....)

alexbishop1 (approved):

danworth:
The Codacy prompts are based on needing `assert()` and no `_` in method names, both of which are not appropriate here. I'm merging as-is.

PR #1222, opened 2019-04-18
PP-5092 Payment amount representation

sfount:
Adds `net_amount` value to `ChargeResponse` - ensuring this value is only calculated once and then used by all consumers (public API, self service, CSV export). 

For now this is calculated at the service level, we might want to consider storing it as a calculated column in the future to remove the calculations on every entity. 

[ ] Confirm that this considers all API end points re: charges - currently only updates /v1/
[ ] Confirm that the tests cover all uses of the charge entity


src/main/java/uk/gov/pay/connector/charge/util/FeeNetCalculator.java
```diff
@@ -0,0 +1,19 @@
...
+    private FeeNetCalculator() {
+    }
+   
+    // @TODO(sfount) this essentially rebuilds the `total_amount` column that is calculated at the service level, if this
+    //               was stored as part of the entity (calculation column?) the same value could be used
+    public static Long getNetAmountFor(ChargeEntity charge) {  
+        return charge.getAmount() + charge.getCorporateSurcharge().orElse(0L) - charge.getFeeAmount().orElse(0L);
+    }
+    
+    public static Long getNetAmountFor(Transaction transaction) {
```
kbottla: Do we this method ? It appears unused

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -404,6 +404,37 @@ public void shouldFindChargeForChargeId_withCorporateSurcharge() {
         assertThat(chargeResponse.getTotalAmount(), is(totalAmount));
         assertThat(chargeResponse.getRefundSummary().getAmountAvailable(), is(totalAmount));
     }
+   
+    @Test
+    public void shouldFindChargeForChargeId_withFee() {
```
kbottla, line 409: Should it also say `calculateNetAmount` ?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceITest.java
```diff
@@ -310,7 +311,29 @@ public void shouldReturnFeeIfItExists() {
                 .contentType(JSON)
                 .body("fee", is(100));
     }
-
+    
+    @Test
+    public void shouldReturnNetAmountIfFeeExists() {
```
kbottla, line 316: Can we also have test for `/v2/api/accounts/{accountId}/charges` ?

kbottla (commented):
Looks good. Added few minor comments

kbottla (dismissed):
LGTM!

src/main/java/uk/gov/pay/connector/charge/model/domain/Transaction.java
```diff
@@ -266,13 +266,13 @@ public String getUserExternalId() {
     public WalletType getWalletType() {
         return walletType;
     }
-    
-    public Optional<Long> getFeeAmount() {
-        return Optional.ofNullable(feeAmount);
+
+    @Override
+    public Optional<Long> getCorporateSurcharge() {
```
kbottla: There is another method `getCorporateCardSurcharge` with same definition. Can we get rid of old method to avoid confusing usage ?

kbottla (commented):
Looks good. Added a minor comment.

src/main/java/uk/gov/pay/connector/charge/model/domain/Transaction.java
mrlumbu: ah, good spot

kbottla (approved):

PR #1227, opened 2019-04-23
PP-5108 Refactor validation for gateway account PATCH requests

cobainc0:
## WHAT YOU DID
 - refactor existing validation to validate `allow_apple_pay`, `allow_google_pay` and
 `allow_zero_amount` as booleans and stop converting to Strings
 - update relevant tests

Solo: @cobainc0

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
@@ -77,13 +77,13 @@ public class GatewayAccountRequestValidator {
                 validateEmailCollectionMode(payload);
                 break;
             case FIELD_ALLOW_GOOGLE_PAY:
-                validateReplaceBooleanValue(payload, FIELD_ALLOW_GOOGLE_PAY);
+                validateReplaceBooleanValue(payload);
                 break;
             case FIELD_ALLOW_APPLE_PAY:
-                validateReplaceBooleanValue(payload, FIELD_ALLOW_APPLE_PAY);
+                validateReplaceBooleanValue(payload);
                 break;
             case FIELD_ALLOW_ZERO_AMOUNT:
-                validateReplaceBooleanValue(payload, FIELD_ALLOW_ZERO_AMOUNT);
+                validateReplaceBooleanValue(payload);
                 break;
             case FIELD_CORPORATE_CREDIT_CARD_SURCHARGE_AMOUNT: 
             case FIELD_CORPORATE_DEBIT_CARD_SURCHARGE_AMOUNT:
@@ -137,13 +137,10 @@ public class GatewayAccountRequestValidator {
         }
     }
 
-    private void validateReplaceBooleanValue(JsonNode payload, String field) {
+    private void validateReplaceBooleanValue(JsonNode payload) {
         throwIfInvalidFieldOperation(payload, REPLACE);
         throwIfNullFieldValue(payload.get(FIELD_VALUE));
-        String booleanString = payload.get(FIELD_VALUE).asText().toLowerCase();
-        if (!booleanString.equals("false") && !booleanString.equals("true")) {
-            throw new ValidationException(Collections.singletonList(format("Value [%s] is not valid for [%s]", booleanString, field)));
-        }
+        throwIfNotBoolean(payload);
     }
     
     private void validateCorporateCardSurchargePayload(JsonNode payload) {
@@ -169,6 +166,12 @@ public class GatewayAccountRequestValidator {
     private void throwIfNotNumber(JsonNode payload) {
         if (!payload.get(FIELD_VALUE).isNumber()) {
             throw new ValidationException(Collections.singletonList(format("Value [%s] is not valid for path [%s]", payload.get(FIELD_VALUE).asText(), payload.get(FIELD_OPERATION_PATH).asText())));
+        }
+    }
+
+    private void throwIfNotBoolean(JsonNode payload) {
+        if (payload.get(FIELD_VALUE) != null && !payload.get(FIELD_VALUE).isBoolean()) {
+            throw new ValidationException(Collections.singletonList(format("Value [%s] must be of type boolean for path [%s]", payload.get(FIELD_VALUE).asText(), payload.get(FIELD_OPERATION_PATH).asText())));
         }
     }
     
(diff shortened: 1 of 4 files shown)


oswaldquek:
looks good but the build is failing

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
```diff
@@ -137,13 +137,10 @@ private void validateEmailCollectionMode(JsonNode payload) {
...
 
-    private void validateReplaceBooleanValue(JsonNode payload, String field) {
+    private void validateReplaceBooleanValue(JsonNode payload) {
         throwIfInvalidFieldOperation(payload, REPLACE);
         throwIfNullFieldValue(payload.get(FIELD_VALUE));
-        String booleanString = payload.get(FIELD_VALUE).asText().toLowerCase();
-        if (!booleanString.equals("false") && !booleanString.equals("true")) {
-            throw new ValidationException(Collections.singletonList(format("Value [%s] is not valid for [%s]", booleanString, field)));
-        }
+        throwIfNotBoolean(payload);
```
danworth, line 143: This is much tidier and easier to read 👍

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
```diff
@@ -171,6 +168,12 @@ private void throwIfNotNumber(JsonNode payload) {
             throw new ValidationException(Collections.singletonList(format("Value [%s] is not valid for path [%s]", payload.get(FIELD_VALUE).asText(), payload.get(FIELD_OPERATION_PATH).asText())));
         }
     }
+
+    private void throwIfNotBoolean(JsonNode payload) {
+        if (!payload.get(FIELD_VALUE).isBoolean()) {
+            throw new ValidationException(Collections.singletonList(format("Value [%s] is not valid for path [%s]", payload.get(FIELD_VALUE).asText(), payload.get(FIELD_OPERATION_PATH).asText())));
```
danworth: Should this mention why it isn't valid, something like `Value [%s] must be of type boolean for path [%s]'

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
```diff
@@ -171,6 +168,12 @@ private void throwIfNotNumber(JsonNode payload) {
             throw new ValidationException(Collections.singletonList(format("Value [%s] is not valid for path [%s]", payload.get(FIELD_VALUE).asText(), payload.get(FIELD_OPERATION_PATH).asText())));
         }
     }
+
+    private void throwIfNotBoolean(JsonNode payload) {
+        if (!payload.get(FIELD_VALUE).isBoolean()) {
```
danworth: I can see that within `validateReplaceBooleanValue` you're previously checking for null value in the `throwIfNullFieldValue` but to make this more reusable I think you should check for null and deal with it explicitly, otherwise this will throw an NPE should someone re-use it carelessly. The guidelines for validation suggest treating `null` values as passing the check, so that validation methods can be stacked together, e.g. "I don't care if its null but if its non-null then it must be a boolean and If `null` values are to be invalid then a specific not null validation should be added, as you've done here within `validateReplaceBooleanValue`. I would suggest something like...
```
if (payload.get(FIELD_VALUE) != null && !payload.get(FIELD_VALUE).isBoolean()) {
```

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceValidationTest.java
```diff
@@ -195,4 +195,20 @@ public void shouldReturn400_whenCorporateDebitCardSurchargeAmountValueIsNull() {
         String errorMessage = response.readEntity(JsonNode.class).get("errors").get(0).textValue();
         assertThat(errorMessage, is("Field [value] is required"));
     }
+
+    @Test
+    public void shouldReturn400_whenAllowZeroAmountValueIsNull() {
```
danworth, line 200: Might be worth adding a test case that "true"/"false" e.g. a boolean as String returns a validation error as this is the crux of the change.

danworth (dismissed):
Thank you for taking the time to fix this up!! 👍 👍 👍 

I've made a couple of recommendations but I think this looks good, especially refactoring isBoolean into it's own method.

danworth:
Sorry @oswaldquek I dived straight in and didn't notice you'd self-assigned it.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceValidationTest.java
cobainc0: Thanks. Good point!

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
cobainc0: Will update this as its more valuable than ` Value [%s] is not valid for path [%s]`

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidator.java
cobainc0: as discussed this has been updated! 👍

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
```diff
@@ -183,7 +183,7 @@ public void shouldIgnoreEmptyOrMissingValue_whenRemoveOperation() {
                         FIELD_VALUE, ImmutableMap.of(FIELD_NOTIFY_API_TOKEN, "")));
         validator.validatePatchRequest(jsonNode);
     }
-    
+
```
danworth, line 186: Appreciate you cleaning up the formatting 👍

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceValidationTest.java
```diff
@@ -195,4 +195,36 @@ public void shouldReturn400_whenCorporateDebitCardSurchargeAmountValueIsNull() {
...
+                .target("/v1/api/accounts/12")
+                .request()
+                .method("PATCH", Entity.json(jsonNode));
+        assertThat(response.getStatus(), is(400));
+        String errorMessage = response.readEntity(JsonNode.class).get("errors").get(0).textValue();
+        assertThat(errorMessage, is("Field [value] is required"));
+    }
+
+    @Test
+    public void shouldReturn400_whenAllowZeroAmountValueIsNotBooleanValue() {
```
danworth, line 216: Good to know it works 👍

danworth (approved):
Looks good 🥇

PR #1231, opened 2019-04-24
Pp 4885 add metadata to ChargeResponse

danworth:
## WHAT YOU DID
- First commit restructures `ChargeResponse` to keep members, constructor and methods of the parent class at the top, with static classes and builder beneath.
- Second commit add `metadata` to the `ChargeResponse` including tests.


danworth:
Codacy is complaining about lack of `assert` which isn't appropriate here. Also doesn't like the many parameters of the `addCharge` within `DatabaseTestHelper` but that would require refactoring the approach which is beyond the scope of this PR.

danworth:
I will move `ExternalMetadataSerializer` to java-commons after this PR is merged in a separate PR.

stephencdaly (approved):
Looks good 👍

PR #1233, opened 2019-04-25
Pp 4884 handle non object metadata

danworth:
## WHAT YOU DID
If `metadata` isn't an object then return a status of 400 and a helpful exception message.

The exception that is thrown from `ExternalMetadataDeserilizer` (in `pay-java-commons`) is always caught within a Jackson library and the message is appended with some path information. This isn't useful to the client so the `JsonMappingExceptionMapper` removes it to return a cleaner message.

The first commit bumps `pay-java-commons` to include the necessary change to the `ExternalMetadataDeserializer`.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
@@ -19,7 +19,15 @@ public class JsonMappingExceptionMapper implements ExceptionMapper<JsonMappingEx
     @Override
     public Response toResponse(JsonMappingException exception) {
         LOGGER.error(exception.getMessage());
-        Map<String, String> entity = Map.of("message", exception.getMessage());
+        Map<String, String> entity = Map.of("message", sanitiseExceptionMessage(exception.getMessage()));
         return Response.status(400).entity(entity).type(APPLICATION_JSON).build();
+    }
+
+    private String sanitiseExceptionMessage(String message) {
+        if (message.contains("Field [metadata] must be an object of JSON key-value pairs")) {
+            return "Field [metadata] must be an object of JSON key-value pairs";
+        }
+
+        return message;
     }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
         <eclipselink.version>2.7.4</eclipselink.version>
         <guice.version>4.2.2</guice.version>
         <jackson.version>2.9.8</jackson.version>
-        <pay-java-commons.version>1.0.20190424113839</pay-java-commons.version>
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
```diff
@@ -19,7 +19,19 @@
     @Override
     public Response toResponse(JsonMappingException exception) {
         LOGGER.error(exception.getMessage());
-        Map<String, String> entity = Map.of("message", exception.getMessage());
+        Map<String, String> entity = Map.of("message", processExceptionMessage(exception.getMessage()));
         return Response.status(400).entity(entity).type(APPLICATION_JSON).build();
     }
+
+    private String processExceptionMessage(String message) {
```
alexbishop1: Can we get a more descriptive name? `condenseExceptionMessage` maybe?

src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
```diff
@@ -19,7 +19,19 @@
...
+        if (message == null) {
```
alexbishop1: If `message` is null then it’s going to blow up when we try to put it in the unmodifiable map…

src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
```diff
@@ -19,7 +19,19 @@
...
+            return message;
+        }
+
+        if (message.contains("Field [metadata] must be an object of JSON key-value pairs")) {
+            return "Field [metadata] must be an object of JSON key-value pairs";
+        }
```
alexbishop1, line 29: Is the stuff we want to remove at the beginning and end of the message? Could we use substring operations to extract the bit we want? It might make what’s happening a bit clearer.

danworth:
Codacy is complaining about the lack of `assert()` which isn't valid in this instance as we use ValidatableResponses for ITests.

alexbishop1 (changes requested):
Looks pretty good — couple of comments

src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
danworth: Yes, ok, how about sanitiseExceptionMessage.

src/main/java/uk/gov/pay/connector/util/JsonMappingExceptionMapper.java
danworth: As discussed with the more meaningful method name, hopefully this is acceptable and most simple.

alexbishop1 (approved):

PR #1235, opened 2019-04-25
PP-4885 Add metadata to TransactionResponse

danworth:
- Added `ExternalMetadata` to Transaction (entity).
- Added test to `TransactionDaoITest` to confirm return of metadata.
- Added test to `ChargesApiV2ResourceITest` to assert return of metadata.

## WHAT YOU DID
This will add `metadata` to response from GET v2/api/account/{accountid}/charges

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/dao/TransactionDao.java
+++ b/src/main/java/uk/gov/pay/connector/charge/dao/TransactionDao.java
@@ -76,7 +76,8 @@ public class TransactionDao {
                         field("delayed_capture"),
                         field("corporate_surcharge"),
                         field("wallet"),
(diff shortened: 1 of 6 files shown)


danworth:
Codacy is just complaining about the lack of `assert()` which isn't appropriate. Maybe we should turn that rule off @alexbishop1 ?

stephencdaly (approved):
Looks good

alexbishop1:
You’re right, we don’t want that particular Codacy check.

It’s my intention to get Codacy configured appropriately using config files checked in to each repo but this involves starting from scratch. For now, I’ve just disabled directly in Codacy.

PR #1236, opened 2019-04-25
PP-5108 Add provider state for wallet payments

cobainc0:
## WHAT YOU DID
 - wallet payments are currently only available with Worldpay
 so this state needed to be added for future pact testing.

Solo: @cobainc0

### DIFF
--- a/src/test/java/uk/gov/pay/connector/pact/TransactionsApiContractTest.java
+++ b/src/test/java/uk/gov/pay/connector/pact/TransactionsApiContractTest.java
@@ -160,6 +160,17 @@ public class TransactionsApiContractTest {
                 .insert();
     }
 
+    @State("a gateway account supporting digital wallet with external id 666 exists in the database")
+    public void anAccountExists() {
+        String aDigitalWalletSupportedPaymentProvider = "worldpay";
+        DatabaseFixtures
+                .withDatabaseTestHelper(dbHelper)
+                .aTestAccount()
+                .withAccountId(666L)
+                .withPaymentProvider(aDigitalWalletSupportedPaymentProvider)
+                .insert();
+    }
+
     @State({"default", "Card types exist in the database"})
     public void defaultCase() {
     }


danworth:
Hmm is this test failing at present and hence the change? I think it should work against Sandbox (sorry, when you asked in Slack I thought you meant for the real providers).

danworth:
If you're creating a pact test related to wallet payments then check out `WalletApiContractTest` in connector, that might help. Happy to help if I can.

cobainc0:
> Hmm is this test failing at present and hence the change? I think it should work against Sandbox (sorry, when you asked in Slack I thought you meant for the real providers).

Unfortunately, it doesn't play nice/work with sandbox, hence the change.

cobainc0:
> If you're creating a pact test related to wallet payments then check out `WalletApiContractTest` in connector, that might help. Happy to help if I can.

Since this at the gateway account level i'll leave the state in `TransactionsApiContractTest` where similar states live until we have made a decision on how to organise and split out this file.

src/test/java/uk/gov/pay/connector/pact/TransactionsApiContractTest.java
```diff
@@ -160,6 +160,17 @@ public void stripeAccountExists() {
...
     }
 
+    @State("a gateway account supporting digital wallet with external id 666 exists in the database")
+    public void anAccountExists() {
+        String aDigitalWalletSupportedPaymentProvider = "worldpay";
+        DatabaseFixtures
+                .withDatabaseTestHelper(dbHelper)
+                .aTestAccount()
+                .withAccountId(666L)
+                .withPaymentProvider(aDigitalWalletSupportedPaymentProvider)
```
danworth, line 170: I like that you've made it clear why its "worldpay" with the descriptive variable name 👍 This helps anyone looking at this state/test in the future.

danworth (approved):
I'm not sure this is the right file for it since it relates to PATCH'ing a gateway account but since you mention that you're discussing a more appropriate place and will move it in the future then I don't wish to hold up your progress. Please can you keep it on your to-do list though to move once the decision is made?

PR #1241, opened 2019-04-26
PP-4993 Backwards compatible stripe capture

mrlumbu:
In order to start taking fees, we need to move to a new way of capturing charges. Instead of using Stripe's destination attribute to automatically route a charge to a Connect account, we will first authorise and capture the charge on the platform account, and then transfer across the correct amount in a second call to Stripe.

In order to get this out without breaking in flight transactions we have to be careful. Whether or not a charge is a 'destination charge'  is determined at the point of authorisation, but a capture can happen some minutes (or in extreme cases hours or days) later. Therefore we need to make sure that at all times, the method of authorisation and the method of capture are compatible with each other.

This PR is the first step. It only changes the Stripe capture functionality.It is backwards compatible with the 'old' (i.e. destination) authorisation method, as it first checks whether the captured charge needs to be transferred before transferring it, so as it stands the new code path should not be used.I have nevertheless included tests for the new code path, as this seems the natural place to put them.


kbottla (commented):
Looks good 👌. Codacy highlighted a few unused imports and variable scoping

mrlumbu:
thanks @kbottla, have resolved those Codacy issues

kbottla (approved):

PR #1242, opened 2019-04-26
bau: Gateway acc id does not need to be specified...

oswaldquek:
...in the post body when creating a charge.


alexbishop1 (approved):
True dat

PR #1243, opened 2019-04-26
PP-5142 Unify error responses produced by exception mappers

stephencdaly:
- Add ErrorResponse model to unify the contents of error responses. Include in this an ErrorIdentifier, which will always be "GENERIC" for now.
- All error responses created by exception mappers now include a "message" field which is an array of strings.
- Error responses produced by ResponseUtils.java will be modified in a later commit.

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/api/ErrorResponse.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/api/ErrorResponse.java
@@ -1,0 +1,43 @@
+package uk.gov.pay.connector.common.model.api;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.List;
+
+import static com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+@JsonInclude(Include.NON_NULL)
+public class ErrorResponse {
+    
+    @JsonProperty("error_identifier")
+    private final ErrorIdentifier identifier;
+    
+    @JsonProperty("message")
+    private final List<String> messages;
+    
+    @JsonProperty("reason")
+    private final String reason;
+    
+    public ErrorResponse(ErrorIdentifier identifier, List<String> messages, String reason) {
+        this.identifier = identifier;
+        this.messages = messages;
+        this.reason = reason;
+    }
+    
+    public ErrorResponse(ErrorIdentifier identifier, List<String> messages) {
+        this(identifier, messages, null);
+    }
+
+    public ErrorIdentifier getIdentifier() {
+        return identifier;
+    }
+
+    public List<String> getMessages() {
+        return messages;
+    }
+
+    public String getReason() {
+        return reason;
+    }
+}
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/connector/common/model/api/ErrorResponse.java
```diff
@@ -0,0 +1,43 @@
...
+@JsonInclude(Include.NON_NULL)
+public class ErrorResponse {
+    
+    @JsonProperty("error_identifier")
+    private final ErrorIdentifier identifier;
+    
+    @JsonProperty("message")
+    private final List<String> messages;
+    
+    @JsonProperty("reason")
```
oswaldquek, line 19: What's the purpose of this `reason` json field? https://payments-platform.atlassian.net/browse/PP-5142 specifies the body of the response to be 
```
{
  "error_identifier": "zero_pence_transaction_not_allowed",
  "message": ["Zero pence transactions not enabled for this gateway account"]
}
``` 
and the RFC at https://docs.google.com/document/d/16PAyL7fqy2CpcTsQ-OB2rsxS99Z3ERbbqSrsvYDXwzI/edit#heading=h.sjuhspjyssr1 doesn't mention it either.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceValidationTest.java
```diff
@@ -115,7 +115,7 @@ public void shouldReturn400_whenCorporatePrepaidCreditCardSurchargeOperationIsMi
                 .request()
                 .method("PATCH", Entity.json(jsonNode));
         assertThat(response.getStatus(), is(400));
-        String errorMessage = response.readEntity(JsonNode.class).get("errors").get(0).textValue();
+        String errorMessage = response.readEntity(JsonNode.class).get("message").get(0).textValue();
```
oswaldquek: I know it's not directly related to the story but could the test assert the "message" field is an array of size 1 too? Just to be more rigorous.

oswaldquek (changes requested):
Looks good. Some comments.

src/main/java/uk/gov/pay/connector/common/model/api/ErrorResponse.java
stephencdaly: As we discussed at the story writing, there is an existing "reason" field that is already included for a handful of errors (see ResponseUtils.java `private static Response responseWith(Status status, String message, String reason) {`)

Not planning on using this any further than it is already, but I want all the error responses to use this object so I'm adding the "reason" field for this purpose.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceValidationTest.java
stephencdaly: Yes, might as well - I'll update the tests

src/main/java/uk/gov/pay/connector/common/model/api/ErrorResponse.java
oswaldquek: 👍

stephencdaly:
I've updated the GatewayAccountResourceValidationTests to check the size of the "message" array. Also added in assertions on the "error_identifier" while I was at it.

oswaldquek (approved):

PR #1249, opened 2019-04-30
PP-4993 Backwards compatible Stripe refund

mrlumbu:
This PR switches Stripe refund to be a two step process. Rather than
using `reverse_transfer: true` to tell Stripe to reverse the transfer
associated with a charge, we now refund the charge without that option, then
make a second call to Stripe to transfer the refund amount from the
Connect account to the Platform account.
The reason for this is that this new method is compatible with both our current
way of authorising and capturing charges, and with the new way that we
are introducing in order to take fees.


kbottla (approved):
Looks good to me 👌 and refunds works as expected when tested with Stripe 

As discussed, we don't currently have a mechanism to recover, if a refund succeeds and a transfer fails. Given the low-to-none percentage of refunds and probability of transfers failing we can have  log alerts for any such scenarios and a decision to be taken accordingly.

PR #1252, opened 2019-04-30
PP-4887: Pact provider test

oswaldquek:
This commit helps the new publicapi-connector pact test in
https://github.com/alphagov/pay-publicapi/pull/442 to work.

There is certainly more refactor work that can be done with the builders but it's best to put that in another PR.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/util/AddChargeParams.java
+++ b/src/test/java/uk/gov/pay/connector/util/AddChargeParams.java
@@ -1,0 +1,220 @@
+package uk.gov.pay.connector.util;
+
+import uk.gov.pay.commons.model.SupportedLanguage;
+import uk.gov.pay.commons.model.charge.ExternalMetadata;
+import uk.gov.pay.connector.charge.model.ServicePaymentReference;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+
+import java.time.ZonedDateTime;
+
+public class AddChargeParams {
+    private final Long chargeId;
+    private final String externalChargeId;
+    private final String gatewayAccountId;
+    private final long amount;
+    private final ChargeStatus status;
+    private final String returnUrl;
+    private final String transactionId;
+    private final String description;
+    private final ServicePaymentReference reference;
+    private final ZonedDateTime createdDate;
+    private final long version;
+    private final String email;
+    private final SupportedLanguage language;
+    private final boolean delayedCapture;
+    private final Long corporateSurcharge;
+    private final ExternalMetadata externalMetadata;
+
+    private AddChargeParams(AddChargeParamsBuilder builder) {
+        chargeId = builder.chargeId;
+        externalChargeId = builder.externalChargeId;
+        amount = builder.amount;
+        gatewayAccountId = builder.gatewayAccountId;
+        status = builder.status;
+        returnUrl = builder.returnUrl;
+        transactionId = builder.transactionId;
+        description = builder.description;
+        reference = builder.reference;
+        createdDate = builder.createdDate;
+        version = builder.version;
+        email = builder.email;
+        language = builder.language;
+        delayedCapture = builder.delayedCapture;
+        corporateSurcharge = builder.corporateSurcharge;
+        externalMetadata = builder.externalMetadata;
+    }
+
+    public Long getChargeId() {
+        return chargeId;
+    }
+
+    public String getExternalChargeId() {
+        return externalChargeId;
+    }
+
+    public String getGatewayAccountId() {
+        return gatewayAccountId;
+    }
+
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/util/AddChargeParams.java
```diff
@@ -0,0 +1,218 @@
...
+            return this;
+        }
+
+        public AddChargeParamsBuilder withExternalMetadata(ExternalMetadata externalMetadata) {
+            this.externalMetadata = externalMetadata;
+            return this;
+        }
+
+        public AddChargeParams build() {
+            AddChargeParams addChargeParams = new AddChargeParams();
```
danworth: Slightly different approach to the last one where the class to be built had a constructor that took the builder. Out of interest any particular reason to choose this approach this time?

src/test/java/uk/gov/pay/connector/util/DatabaseTestHelper.java
```diff
@@ -210,6 +204,59 @@ private void addCharge(
...
+                        addChargeParams.getEmail(),
+                        addChargeParams.getLanguage().toString(),
+                        addChargeParams.isDelayedCapture(),
+                        addChargeParams.getCorporateSurcharge(),
+                        jsonMetadata
+                )
+        );
+    }
+    
     private void addCharge(
```
danworth, line 264: I think it would be good to mark this original method with `@Deprecated` so that people know to build future new parameters into the `AddChargeParams` and into `addCharge(AddChargeParams addChargeParams)` until the refactoring to remove it is complete (might help people seeing this understand why we have these two similar methods and avoid spending time maintaining the SQL/insert in this older method.

src/test/java/uk/gov/pay/connector/util/AddChargeParams.java
```diff
@@ -0,0 +1,218 @@
+package uk.gov.pay.connector.util;
+
+import uk.gov.pay.commons.model.SupportedLanguage;
+import uk.gov.pay.commons.model.charge.ExternalMetadata;
+import uk.gov.pay.connector.charge.model.ServicePaymentReference;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+
+import java.time.ZonedDateTime;
+
+public class AddChargeParams {
```
danworth, line 10: Another builder, thanks 👍

danworth (changes requested):
Looks good, thanks for taking the time to introduce the builder pattern to `addCharge`. It's a minor point but I feel the original `addCharge` method that results in inserting to the DB should be marked with `@Deprecated` and a helpful message so that when people add new parameters to a charge, which is rather often, they'll know to add to the new builder and associated method and we won't have to maintain both means of doing this.

src/test/java/uk/gov/pay/connector/util/AddChargeParams.java
oswaldquek: Oh I got intellij to generate the builder and this is what got generated. I think it's good have a private constructor for the `AddChargeParams` and make the fields final so it can't be constructed by anything else. I'll make the change 👍

src/test/java/uk/gov/pay/connector/util/DatabaseTestHelper.java
oswaldquek: good shout, done. i have a PR to get rid of this method very shortly!

src/test/java/uk/gov/pay/connector/util/AddChargeParams.java
```diff
@@ -0,0 +1,220 @@
...
+    private final ServicePaymentReference reference;
+    private final ZonedDateTime createdDate;
+    private final long version;
+    private final String email;
+    private final SupportedLanguage language;
+    private final boolean delayedCapture;
+    private final Long corporateSurcharge;
+    private final ExternalMetadata externalMetadata;
+
+    private AddChargeParams(AddChargeParamsBuilder builder) {
```
danworth, line 28: 👍

src/test/java/uk/gov/pay/connector/util/DatabaseTestHelper.java
```diff
@@ -210,6 +204,63 @@ private void addCharge(
...
+                        addChargeParams.getLanguage().toString(),
+                        addChargeParams.isDelayedCapture(),
+                        addChargeParams.getCorporateSurcharge(),
+                        jsonMetadata
+                )
+        );
+    }
+    
+    /*
+        Use addCharge(AddChargeParams addChargeParams) instead.
```
danworth, line 261: 💯

danworth (approved):
Approved changes since last review, looks good!

PR #1254, opened 2019-05-01
PP-4993 Authorise stripe charges without setting destination

mrlumbu:
In order to start taking fees via the method of doing a net transfer
from our Stripe platform account to the relevant Stripe connect
account, we need to stop telling Stripe to automatically transfer
a charge to a connect account on capture. This PR does that.

Whilst this change is small, it will have a quite a big effect, as
it will activate the code path here
https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java#L50

which will mean captures will now be processed by a two stage process.
Therefore, to test this PR I recommend testing locally against real Stripe
gateway and ensuring that authorisation, capture, and refund all behave as expected.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
@@ -365,7 +365,6 @@ public class StripePaymentProvider implements PaymentProvider {
         String stripeAccountId = getStripeAccountId(externalId, gatewayAccount);
 
         params.add(new BasicNameValuePair("on_behalf_of", stripeAccountId));
-        params.add(new BasicNameValuePair("transfer_data[destination]", stripeAccountId));
         return URLEncodedUtils.format(params, UTF_8);
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -365,7 +365,6 @@ private String buildAuthorisePayload(String sourceId, String externalId, String
         String stripeAccountId = getStripeAccountId(externalId, gatewayAccount);
 
         params.add(new BasicNameValuePair("on_behalf_of", stripeAccountId));
-        params.add(new BasicNameValuePair("transfer_data[destination]", stripeAccountId));
```
kbottla, line 368: Don't we need to transfer funds if we remove this? 

Stripe docs says funds are transferred when `transfer_data[destination]` is set. If not they may just be on platform account

sfount: My understanding is that removing `transfer_data[destination]` will have authorisations take a second path in Connector -- merging this PR will start using that new path.

kbottla (approved):
Looks good and payment amount is now transferred to connect account when a charge is captured

PR #1255, opened 2019-05-01
bau: Use the builder pattern for DatabaseTestHelper

oswaldquek:
This follows from https://github.com/alphagov/pay-connector/pull/1252


stephencdaly (dismissed):
Great improvement, nice work.

My only comment would be could we use defaults more for some of the fields where we set a hardcoded value in all the usages but don't really care about the value of, such as reference, description, return_url and language etc in most cases? It would reduce the number of lines by a fair bit.

This is partly preference though, and would take more time, so I'm happy with it as is.

oswaldquek:
@stephencdaly yeah I agree its sensible to use defaults where possible. In fact I already do that for created date and version. I'll amend the PR.

stephencdaly (approved):

PR #1257, opened 2019-05-01
PP-4993 Take fees for stripe payments

mrlumbu:
This commit alters the stripe capture method to do the following:
- capture the charge
- examine the charge to see if it needs to be transferred (see previous commits)
- if it does, look at the returned Stripe charge and extract the fee Stripe
has charged for the transaction
- use this fee to calculate the total fee we need to charge - there is an additional
amount currently controlled by `STRIPE_TRANSACTION_FEE_PERCENTAGE`
- transfer the total amount - calculated fee to the connect account

This fee is returned up to the capture service as an optional attribute of the
CaptureResponse, where it is persisted if it exists.

`COLLECT_FEE_FEATURE_FLAG` needs to be set to true for this to actually start working. This flag defaults to false.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
@@ -18,6 +18,8 @@ import uk.gov.pay.connector.gateway.stripe.request.StripeTransferOutRequest;
 import uk.gov.pay.connector.gateway.stripe.response.StripeCaptureResponse;
 import uk.gov.pay.connector.util.JsonObjectMapper;
 
+import java.util.Optional;
+
 import static javax.ws.rs.core.Response.Status.Family.CLIENT_ERROR;
 import static javax.ws.rs.core.Response.Status.Family.SERVER_ERROR;
 import static uk.gov.pay.connector.gateway.CaptureResponse.ChargeState.COMPLETE;
@@ -45,13 +47,22 @@ public class StripeCaptureHandler implements CaptureHandler {
         String transactionId = request.getTransactionId();
 
         try {
-            StripeCharge stripeCaptureResponse = captureCharge(request);
+            StripeCharge stripeCaptureResponse = captureWithPlatform(request);
+            Optional<Long> maybeFee = Optional.empty();
 
             if (stripeCaptureResponse.isPlatformCharge()) {
-                transferToConnectAccount(request);
+                maybeFee = calculateProcessingFee(request.getAmount(), stripeCaptureResponse);
+                Long netTransferAmount = maybeFee
+                        .map(fee -> calculateNetTransferAmount(request.getAmount(), fee))
+                        .orElse(request.getAmount());
+                transferToConnectAccount(request, netTransferAmount);
             }
 
-            return fromBaseCaptureResponse(new StripeCaptureResponse(stripeCaptureResponse.getId()), COMPLETE);
+            return new CaptureResponse(
+                    stripeCaptureResponse.getId(),
+                    COMPLETE,
+                    maybeFee.orElse(null)
+            );
         } catch (GatewayErrorException e) {
 
             if (e.getFamily() == CLIENT_ERROR) {
@@ -80,7 +91,7 @@ public class StripeCaptureHandler implements CaptureHandler {
         }
     }
 
-    private StripeCharge captureCharge(CaptureGatewayRequest request) throws GatewayException.GenericGatewayException, GatewayErrorException, GatewayException.GatewayConnectionTimeoutException {
+    private StripeCharge captureWithPlatform(CaptureGatewayRequest request) throws GatewayException.GenericGatewayException, GatewayErrorException, GatewayException.GatewayConnectionTimeoutException {
         String captureResponse = client.postRequestFor(StripeCaptureRequest.of(request, stripeGatewayConfig)).getEntity();
         StripeCharge stripeCaptureResponse = jsonObjectMapper.getObject(captureResponse, StripeCharge.class);
         logger.info("Captured charge id {} with platform account - stripe capture id {}",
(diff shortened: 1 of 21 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -102,4 +113,17 @@ private void transferToConnectAccount(CaptureGatewayRequest request) throws Gate
                 stripeTransferResponse.getStripeTransferGroup()
         );
     }
+
+    private Optional<Long> calculateFee(Long grossChargeAmount, StripeCharge stripeCaptureResponse) {
+        if (stripeGatewayConfig.isCollectFee()) {
+            Double additionalFee = Math.ceil((stripeGatewayConfig.getFeePercentage()/100) * grossChargeAmount);
+            return stripeCaptureResponse.getFee().map(stripeFee -> stripeFee + additionalFee.longValue());
```
kbottla: Will it ever be possible to have a charge with out fee when feature is turned on ?

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -102,4 +113,17 @@ private void transferToConnectAccount(CaptureGatewayRequest request) throws Gate
                 stripeTransferResponse.getStripeTransferGroup()
         );
     }
+
+    private Optional<Long> calculateFee(Long grossChargeAmount, StripeCharge stripeCaptureResponse) {
+        if (stripeGatewayConfig.isCollectFee()) {
+            Double additionalFee = Math.ceil((stripeGatewayConfig.getFeePercentage()/100) * grossChargeAmount);
```
sfount: `additionalFee` here is potentially vague? It might read more clearly to name this something along the lines of: 

`applicationFee` (this is what the concept is referred to on Stripe's end)
`platformFee` 
`stripePlatformApplicationFee`

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -45,13 +47,22 @@ public CaptureResponse capture(CaptureGatewayRequest request) {
         String transactionId = request.getTransactionId();
 
         try {
-            StripeCharge stripeCaptureResponse = captureCharge(request);
+            StripeCharge stripeCaptureResponse = captureWithPlatform(request);
+            Optional<Long> maybeFee = Optional.empty();
 
             if (stripeCaptureResponse.isPlatformCharge()) {
-                transferToConnectAccount(request);
+                maybeFee = calculateFee(request.getAmount(), stripeCaptureResponse);
```
kbottla: Can we name it to something like `calculateConnectFee` or similar ?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -88,29 +93,37 @@ private CaptureResponse capture(ChargeEntity chargeEntity) {
...
+    public void processGatewayCaptureResponse(String chargeId, String oldStatus, CaptureResponse captureResponse) {
+
+        ChargeStatus nextStatus = determineNextStatus(captureResponse);
+        checkTransactionId(chargeId, captureResponse);
 
-        ChargeStatus nextStatus = determineNextStatus(operationResponse);
-        checkTransactionId(chargeId, operationResponse);
 
         ChargeEntity charge = chargeService.updateChargePostCapture(chargeId, nextStatus);
+        captureResponse.getFee().ifPresent(fee -> persistFee(charge, fee));
```
kbottla, line 103: is the `addtional fee` amount (STRIPE_TRANSACTION_FEE_PERCENTAGE) included in this ?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -88,29 +93,37 @@ private CaptureResponse capture(ChargeEntity chargeEntity) {
...
                 charge.getGatewayAccount().getId(), nextStatus.toString())).inc();
 
-        if (operationResponse.isSuccessful()) {
+        if (captureResponse.isSuccessful()) {
             userNotificationService.sendPaymentConfirmedEmail(charge);
         }
     }
 
+    @Transactional
+    private void persistFee(ChargeEntity charge, Long feeAmount) {
```
kbottla: This should be `public` method as `@Transactional` won't work with private methods

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
mrlumbu: not sure if `calculateConnectFee` correctly describes the fee? It is a fee that includes the stripe processing fee and the connect fee that we pay stripe. So I just left it as fee

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
mrlumbu: yes. that was included in the calculation of the fee in stripe capture

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
kbottla: may be `calculateStripeFee` ? so it won't be confusing with card fee

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
mrlumbu: `calculateProcessingFee`?

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
mrlumbu: `calculateProcessingFee`?

kbottla: 👍

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
mrlumbu: No, but to get the fee from a StripeCharge object, it needs to have it's `balance_transaction` attribute expanded. If it doesn't, you can't get the fee. Since it is hard to guarantee that in all contexts the `balance_transaction` will be expanded, I have defensively allowed for it not being available, which is why you end up with an optional fee.

mrlumbu:
@sfount @kbottla thanks, I have addressed your comments

src/main/resources/config/config.yaml
```diff
@@ -83,6 +83,8 @@ stripe:
     test: ${GDS_CONNECTOR_STRIPE_WEBHOOK_SIGN_SECRET}
     live: ${GDS_CONNECTOR_STRIPE_WEBHOOK_LIVE_SIGN_SECRET}
   platformAccountId: ${STRIPE_PLATFORM_ACCOUNT_ID}
+  feePercentage: ${STRIPE_TRANSACTION_FEE_PERCENTAGE}
```
kbottla, line 86: We will need default here as container won't start up due to `Cannot deserialize value of type java.lang.Double from String "${STRIPE_TRANSACTION_FEE_PERCENTAGE}": not a valid Double value`

mrlumbu: or you need to set the env var? I'm not sure about setting a default value here - kind of publicising our fee structuring, and it's a bit of a weird magic place for it to live.

kbottla: 👍 May be a small PR for infra & scripts then.

kbottla (dismissed):
LGTM! but probably need to update scripts with new environment before merging this

sfount (dismissed):

kbottla (approved):

PR #1258, opened 2019-05-02
PP-5142 Finish using ErrorResponse in ResponseUtil

stephencdaly:
There was one last method which constructed a response using a List object rather than a typed List. Modify calling code so that it always passes a List<String> which allows us to use an ErrorResponse
to construct the response body.

Also renamed some private methods in ResponseUtil.


oswaldquek (dismissed):

cobainc0 (approved):
👍 LGTM

PR #1261, opened 2019-05-02
PP-5080 Added provider state for fee and net_amount pact

kbottla:
## WHAT YOU DID

- Adds a new pact state for new fields `net_amount` and `fee` between connector and publicapi


SandorArpa (approved):
LOKTM

PR #1256, opened 2019-05-01
PP-5142 Use ErrorResponse to construct responses created by ResponseUtil

stephencdaly:
- Modify ResponseUtil so that most of the methods that construct an error response use the ErrorResponse model.
- Modify tests to expect a "messages" array and to assert on the "error_identifier" - which has a value of "GENERIC" for all responses currently.
- Not all responses have been converted to use ErrorResponse yet.

PR #1264, opened 2019-05-02
PP-5142 Use ErrorResponse to construct responses created by ResponseUtil

stephencdaly:
- Modify ResponseUtil so that most of the methods that construct an error response use the ErrorResponse model. Modify tests to expect a "messages" array and to assert on the "error_identifier", which has a value of "GENERIC" for all responses currently.
- Not all responses have been converted to use ErrorResponse yet.

This pull request is the same changes as were in https://github.com/alphagov/pay-connector/pull/1256 but that PR had to be reverted after it was merged as it broke end-to-end tests. End-to-end was fixed by a change in Public API


stephencdaly:
End to end tests were run on this and passed

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #1269, opened 2019-05-03
BAU Fix dropwizard config overrides leaking to other apps

stephencdaly:
- The way that ConfigOverrides work in DropwizardTestSupport is to set
  System properties with a ".dw" prefix and later read them when
  starting the app. These are only removed from the System properties
  when the app is stopped. However, we don't use DropwizardTestSupport
  in the recommended way and instead keep Dropwizard apps running
  between tests. To fix this, we need to manually clear the config
  overrides from the System properties immediately after starting the
  app.
- Also fix the check in DropwizardTestApplications which checks if the
  ConfigOverrides differ for a test class from the ones used to start
  the previous app. This didn't take into account the fact that multiple
  Dropwizard apps could be running, so shouldn't check on the key of the
  new app, and instead just look at the config overrides.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/junit/DropwizardJUnitRunner.java
+++ b/src/test/java/uk/gov/pay/connector/junit/DropwizardJUnitRunner.java
@@ -58,30 +58,35 @@ public final class DropwizardJUnitRunner extends JUnitParamsRunner {
     protected Statement classBlock(final RunNotifier notifier) {
         DropwizardConfig dropwizardConfigAnnotation = dropwizardConfigAnnotation();
         List<ConfigOverride> configOverride = newArrayList();
-        
+
         if (dropwizardConfigAnnotation.withDockerPostgres()) {
             getOrCreate();
             configOverride.add(config("database.url", getDbUri()));
             configOverride.add(config("database.user", getDbUsername()));
             configOverride.add(config("database.password", getDbPassword()));
         }
-        
-        if (dropwizardConfigAnnotation.configOverrides().length > 0) 
+
+        if (dropwizardConfigAnnotation.configOverrides().length > 0) {
             Arrays.stream(dropwizardConfigAnnotation.configOverrides()).forEach(c -> configOverride.add(config(c.key(), c.value())));
-        
+        }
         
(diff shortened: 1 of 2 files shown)


alexbishop1 (dismissed):

src/test/java/uk/gov/pay/connector/junit/DropwizardJUnitRunner.java
```diff
@@ -79,6 +79,9 @@ protected Statement classBlock(final RunNotifier notifier) {
         if (dropwizardConfigAnnotation.withDockerPostgres() && createdApp.isPresent()) {
             try {
                 createdApp.get().getApplication().run("db", "migrate", resourceFilePath(dropwizardConfigAnnotation.config()));
+                // DropwizardTestSupport adds the config overrides to the System properties. Clear these up immediately
+                // otherwise all subsequent DropWizard apps started by tests will use these until the app is stopped
+                DropwizardTestApplications.removeConfigOverridesFromSystemProperties();
             } catch (Exception e) {
                 throw new DropwizardJUnitRunnerException(e);
```
rhowe-gds: Should we also clean up in the case where an exception is thrown, i.e. move to a `finally` block?

stephencdaly: yes, good spot

src/test/java/uk/gov/pay/connector/junit/DropwizardJUnitRunner.java
```diff
@@ -81,6 +81,10 @@ protected Statement classBlock(final RunNotifier notifier) {
                 createdApp.get().getApplication().run("db", "migrate", resourceFilePath(dropwizardConfigAnnotation.config()));
             } catch (Exception e) {
                 throw new DropwizardJUnitRunnerException(e);
+            } finally {
+                // DropwizardTestSupport adds the config overrides to the System properties. Clear these up immediately
+                // otherwise all subsequent DropWizard apps started by tests will use these until the app is stopped
+                DropwizardTestApplications.removeConfigOverridesFromSystemProperties();
```
rhowe-gds: I'm not sure even this is watertight to be honest - what if the condition in the `if` isn't true? Maybe move the `try {` two lines up?

rhowe-gds (approved):

PR #1275, opened 2019-05-07
PP-5142 Add error_identifiers for create refund error responses

stephencdaly:
Currently, in public api we make some presumptions about the error responses based on the HTTP status code in order to return a meaningful error response to the consumer. Adding in an error_identifier for these cases will allow us to return the errors without making these presumptions.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/api/ErrorIdentifier.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/api/ErrorIdentifier.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.common.model.api;
 
 public enum ErrorIdentifier {
-    GENERIC
+    GENERIC,
+    REFUND_NOT_AVAILABLE,
+    REFUND_AMOUNT_AVAILABLE_MISMATCH
 }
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/common/model/api/ErrorIdentifier.java
```diff
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.common.model.api;
 
 public enum ErrorIdentifier {
-    GENERIC
+    GENERIC,
+    REFUND_NOT_AVAILABLE,
+    REFUND_AMOUNT_AVAILABLE_MISMATCH
```
alexbishop1, line 6: Isn’t this in Pay Java commons now?

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/common/model/api/ErrorIdentifier.java
stephencdaly: Yes but this is a commit i did a while ago, next PR will use java commons

PR #1279, opened 2019-05-08
PP-5190 Treat Stripe auth failures as `AUTHORISATION_REJECTED`

kbottla:
## WHAT
Currently when a charge fails authorisation (card cvc check fails, insufficient funds, and so on),
these are treated as authorisation errors (AUTHORISATION_ERROR) and frontend displays standard
error page with message "We’re experiencing technical problems", in which case users may see Pay's platform as having issue and may retry using same card.

As the errors are authorisation failures, categorising them to AUTHORISATION_REJECTED would be sensible and frontend can show 'payment declined' page.


SandorArpa (approved):
LOKTM

PR #1280, opened 2019-05-09
PP-4886: Set metadata to null if empty

oswaldquek:
If a payment was created with empty metadata
```
"metadata":{}
```

The payment shoud be treated as if there were no metadata.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
@@ -36,6 +36,7 @@ import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.text.MatchesPattern.matchesPattern;
+import static org.junit.Assert.assertNull;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CREATED;
 import static uk.gov.pay.connector.matcher.ResponseContainsLinkMatcher.containsLink;
 import static uk.gov.pay.connector.matcher.ZoneDateTimeAsStringWithinMatcher.isWithin;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
```diff
@@ -522,6 +526,35 @@ public void shouldReturn201WithBillingAddresssWhenPrefilledCardHolderDetailsFiel
                 .body("card_details." + JSON_CARDHOLDER_NAME_KEY, is(cardholderName))
                 .body("containsKey('billing_address')", is(false));
     }
+    
+    @Test
+    public void shouldReturnChargeWithNoMetadataField_whenCreatedWithEmptyMetadata() {
+        String reference = UUID.randomUUID().toString();
```
alexbishop1: Why are we not just using a fixed value here?

alexbishop1 (commented):
How feasible would it be to have a test that ensures we actually don’t write an empty object to the database?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
oswaldquek: This is just for convenience to perform "searching" in the test below:
```        connectorRestApiClient
                .withQueryParam("reference", reference)
                .getChargesV1()
```

oswaldquek:
@alexbishop1 I've added a test by getting the charge directly from the databaseTestHelper. Hope that works for you.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
alexbishop1: Yeah, but why can’t it be just be `String reference = "my-awesome-reference"`?

oswaldquek: ok, i give in. fixed.

alexbishop1 (approved):

PR #1295, opened 2019-05-13
PP-5168 Feature flag for Capture process using SQS

kbottla:


mrlumbu (approved):

heathd:
maybe document that it will control the behaviour when reading from the queue, rather than writing?

kbottla:
> maybe document that it will control the behaviour when reading from the queue, rather than writing?

@heathd We may need to use it when writing to the queue too i.e., when writing to SQS, ignore any exceptions if feature flag is not enabled. When enabled, abort the payment and return 500 for exceptions.

PR #1297, opened 2019-05-13
PP-5143 Allow zero amount charges if gateway account allows it

alexbishop1:
If a gateway account is has `allow_zero_amount` set to `true`, allow a payment to be created with an amount of 0. If `allow_zero_amount` is `false`, return a 422 response with a new error:

```json
{
    "error_identifier": "ZERO_AMOUNT_NOT_ALLOWED",
    "message": ["Zero amount charges are not enabled for this gateway account"]
}
```

Details:

- Change the validation for amount from being `@Min(value = 1)` to `@Min(value = 0)`
- Make amount in `ChargeCreateRequest` a boxed type so that if it’s missing in the request it doesn’t default to 0 and then pass validation — this changes the error message returned if the amount is missing to explicitly complain it’s null (rather than just between the accepted values) but this is actually more accurate
- Update the amount out of range validation error message so it describes the new minimum of 0 rather than 1
- Put a check in `ChargeService` that throws a `ZeroAmountNotAllowedForGatewayAccountException` if an amount of 0 is provided but the gateway account has `allow_zero_amount` set to `false`
- Add an exception mapper to map the `ZeroAmountNotAllowedForGatewayAccountException` to the appropriate response

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
@@ -25,6 +25,7 @@ import org.glassfish.jersey.server.filter.RolesAllowedDynamicFeature;
 import uk.gov.pay.commons.utils.logging.LoggingFilter;
 import uk.gov.pay.commons.utils.xray.Xray;
 import uk.gov.pay.connector.cardtype.resource.CardTypesResource;
+import uk.gov.pay.connector.charge.exception.ZeroAmountNotAllowedForGatewayAccountExceptionMapper;
 import uk.gov.pay.connector.charge.resource.ChargesApiResource;
 import uk.gov.pay.connector.charge.resource.ChargesFrontendResource;
 import uk.gov.pay.connector.chargeevent.resource.ChargeEventsResource;
@@ -100,11 +101,12 @@ public class ConnectorApp extends Application<ConnectorConfiguration> {
         environment.jersey().register(new ConstraintViolationExceptionMapper());
         environment.jersey().register(new ValidationExceptionMapper());
         environment.jersey().register(new UnsupportedOperationExceptionMapper());
-        environment.jersey().register(new LoggingExceptionMapper<Throwable>() {
-        });
+        environment.jersey().register(new LoggingExceptionMapper<>() {});
         environment.jersey().register(new JsonProcessingExceptionMapper());
         environment.jersey().register(new EarlyEofExceptionMapper());
         environment.jersey().register(new JsonMappingExceptionMapper());
+        environment.jersey().register(new JsonMappingExceptionMapper());
+        environment.jersey().register(new ZeroAmountNotAllowedForGatewayAccountExceptionMapper());
 
         environment.jersey().register(injector.getInstance(GatewayAccountResource.class));
         environment.jersey().register(injector.getInstance(StripeAccountSetupResource.class));
--- a/src/main/java/uk/gov/pay/connector/charge/exception/ZeroAmountNotAllowedForGatewayAccountExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/connector/charge/exception/ZeroAmountNotAllowedForGatewayAccountExceptionMapper.java
@@ -1,0 +1,32 @@
+package uk.gov.pay.connector.charge.exception;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.commons.model.ErrorIdentifier;
+import uk.gov.pay.connector.common.model.api.ErrorResponse;
+
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import java.util.List;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+public class ZeroAmountNotAllowedForGatewayAccountExceptionMapper implements ExceptionMapper<ZeroAmountNotAllowedForGatewayAccountException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ZeroAmountNotAllowedForGatewayAccountExceptionMapper.class);
+
+    private static final String RESPONSE_ERROR_MESSAGE = "Zero amount charges are not enabled for this gateway account";
+
+    @Override
+    public Response toResponse(ZeroAmountNotAllowedForGatewayAccountException exception) {
+        LOGGER.info(exception.getMessage());
+
+        ErrorResponse errorResponse = new ErrorResponse(ErrorIdentifier.ZERO_AMOUNT_NOT_ALLOWED, List.of(RESPONSE_ERROR_MESSAGE));
+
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -100,11 +101,13 @@ public void run(ConnectorConfiguration configuration, Environment environment) {
         environment.jersey().register(new ConstraintViolationExceptionMapper());
         environment.jersey().register(new ValidationExceptionMapper());
         environment.jersey().register(new UnsupportedOperationExceptionMapper());
-        environment.jersey().register(new LoggingExceptionMapper<Throwable>() {
+        environment.jersey().register(new LoggingExceptionMapper<>() {
```
oswaldquek: can lines 104-105 be collapsed into a single line please?

src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
```diff
@@ -19,10 +19,10 @@
 public class ChargeCreateRequest {
 
     @NotNull(message = "Field [amount] cannot be null")
-    @Min(value = 1, message = "Field [amount] can be between 1 and 10_000_000")
-    @Max(value = 10_000_000, message = "Field [amount] can be between 1 and 10_000_000")
+    @Min(value = 0, message = "Field [amount] can be between 0 and 10_000_000")
+    @Max(value = 10_000_000, message = "Field [amount] can be between 0 and 10_000_000")
     @JsonProperty("amount")
-    private long amount;
+    private Long amount; // Long not long so missing doesn’t get deserialised to 0
```
oswaldquek: If we have a test for a missing amount, I don't think we need this comment here.

src/test/java/uk/gov/pay/connector/charge/exception/ZeroAmountNotAllowedForGatewayAccountExceptionMapperTest.java
```diff
@@ -0,0 +1,31 @@
...
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+import static uk.gov.pay.connector.charge.exception.ZeroAmountNotAllowedForGatewayAccountExceptionMapper.RESPONSE_ERROR_MESSAGE;
+
+public class ZeroAmountNotAllowedForGatewayAccountExceptionMapperTest {
+
+    private final ZeroAmountNotAllowedForGatewayAccountExceptionMapper mapper = new ZeroAmountNotAllowedForGatewayAccountExceptionMapper();
+
+    @Test
+    public void toResponseReturns422JsonWithZeroAmountNotAllowedErrorIdentifierAndMessage() {
```
oswaldquek: this seems like it's already covered in ChargeServiceTest.shouldReturn422WhenAmountIsZeroIfAccountDoesNotAllowIt

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
```diff
@@ -245,8 +245,29 @@ public void shouldReturn404WhenCreatingChargeAccountIdIsNonNumeric() {
     }
 
     @Test
-    public void shouldReturn422WhenAmountIsLessThanMinAmount() {
+    public void shouldReturn422WhenAmountIsZeroIfAccountDoesNotAllowIt() {
+        String expectedReference = "Test reference";
+        String expectedDescription = "Test description";
```
oswaldquek, line 250: can `expectedReference` and `expectedDescription` not be inlined? i don't see the value of declaring them separately as they're only used once in lines 253 and 254.

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
alexbishop1: Sure

src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
alexbishop1: We do have a test that will fail — I thought it was worth noting because the behaviour is rather surprising but I can remove it if you think it’s unnecessary

src/test/java/uk/gov/pay/connector/charge/exception/ZeroAmountNotAllowedForGatewayAccountExceptionMapperTest.java
alexbishop1: It’s `ChargesApiCreateResourceITest.shouldReturn422WhenAmountIsZeroIfAccountDoesNotAllowIt`, which is an integration test rather than a unit test. I was in two minds about the value of adding this unit test though, so I can remove it.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiCreateResourceITest.java
alexbishop1: They can, but all the tests follow this format. How about I leave it as it is for now and create a follow-up PR to make the change across all the tests in the class?

src/main/java/uk/gov/pay/connector/charge/exception/ZeroAmountNotAllowedForGatewayAccountExceptionMapper.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.ws.rs.ext.ExceptionMapper;
+import java.util.List;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+public class ZeroAmountNotAllowedForGatewayAccountExceptionMapper implements ExceptionMapper<ZeroAmountNotAllowedForGatewayAccountException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ZeroAmountNotAllowedForGatewayAccountExceptionMapper.class);
+
+    static final String RESPONSE_ERROR_MESSAGE = "Zero amount charges are not enabled for this gateway account";
```
oswaldquek: Codacy flagged this. Should this be private?

oswaldquek (changes requested):
Thanks for addressing the comments. Just one more which was flagged by Codacy.

src/main/java/uk/gov/pay/connector/charge/exception/ZeroAmountNotAllowedForGatewayAccountExceptionMapper.java
alexbishop1: This is actually something where Codacy and IntelliJ give conflicting advice and in previous discussions we’ve said we should reconfigure Codacy. In any case, this constant used to be used by the test but since we don’t have a specific test for this class now, it can be made private.

oswaldquek (approved):

PR #1304, opened 2019-05-14
PP-5166 Basic read and write with SQS

kbottla:
## WHAT YOU DID
- Adds a service to send message to SQS and reads messages from SQS

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/SqsQueueReceiveResponse.java
+++ b/src/main/java/uk/gov/pay/connector/queue/SqsQueueReceiveResponse.java
@@ -1,0 +1,41 @@
+package uk.gov.pay.connector.queue;
+
+import com.amazonaws.services.sqs.model.ReceiveMessageResult;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class SqsQueueReceiveResponse {
+
+    private String messageId;
+    private String receiptHandle;
+    private String messageBody;
+
+    private SqsQueueReceiveResponse(String messageId, String receiptHandle, String messageBody) {
+        this.messageId = messageId;
+        this.receiptHandle = receiptHandle;
+        this.messageBody = messageBody;
+    }
+
+    public static List<SqsQueueReceiveResponse> of(ReceiveMessageResult receiveMessageResult) {
+
+        List<SqsQueueReceiveResponse> sqsQueueReceiveResponse = receiveMessageResult.getMessages()
+                .stream()
+                .map(c -> new SqsQueueReceiveResponse(c.getMessageId(), c.getReceiptHandle(), c.getBody()))
+                .collect(Collectors.toList());
+
+        return sqsQueueReceiveResponse;
+    }
+
+    public String getMessageId() {
+        return messageId;
+    }
+
+    public String getReceiptHandle() {
+        return receiptHandle;
+    }
+
+    public String getMessageBody() {
+        return messageBody;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/queue/SqsQueueService.java
+++ b/src/main/java/uk/gov/pay/connector/queue/SqsQueueService.java
@@ -30,11 +30,11 @@ public class SqsQueueService {
         }
     }
 
-    public List<SqsQueueReceiveResponse> receiveMessages(String queueUrl) {
+    public List<QueueMessage> receiveMessages(String queueUrl) {
         try {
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/pay/connector/queue/SqsQueueService.java
```diff
@@ -0,0 +1,44 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    private AmazonSQS sqsClient;
+
+    public SqsQueueService(AmazonSQS sqsClient) {
+        this.sqsClient = sqsClient;
+    }
+
+    public void sendMessage(String queueUrl, String messageBody) {
```
mrlumbu: not sure why we're passing the queue url into `sendMessage` - surely that is config that just needs to be injected?

src/main/java/uk/gov/pay/connector/queue/SqsQueueReceiveResponse.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class SqsQueueReceiveResponse {
+
+    private String messageId;
+    private String receiptHandle;
+    private String messageBody;
+
+    private SqsQueueReceiveResponse(String messageId, String receiptHandle, String messageBody) {
```
mrlumbu: this name wasn't clear to me on first reading - response made me think it was an http thing, but really it's just a pojo to wrap a message right? In fact I'm not sure what we're gaining here, as we're just swapping one Sqs specific class for another (`sqs.model.Message` for `SqsQueueReceiveResponse`)

src/test/java/uk/gov/pay/connector/queue/SqsQueueServiceTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SqsQueueServiceTest {
+
+    private static final String CAPTURE_QUEUE_URL = "http://capture-url";
```
mrlumbu: the implementation of the queue service is totally agnostic of what it is for, so the test should probably also be generic - i.e it shouldn't be talking about specific things like capture etc

mrlumbu (commented):
I've made some comments - they're not that precise so happy to discuss them. I'm also interested in how you intend to expose this. i.e. Nothing else should know it's interacting with an sqs queue. But you may have already thought about how that's going to work

src/main/java/uk/gov/pay/connector/queue/SqsQueueService.java
kbottla: Thinking to make it reusable as we are going to use more queue(s) for reporting. The same method `sendMessage` can be used everywhere and each process (Capture, Report related) can pass its own queue URL.

src/test/java/uk/gov/pay/connector/queue/SqsQueueServiceTest.java
kbottla: 👍 Updated

src/main/java/uk/gov/pay/connector/queue/SqsQueueReceiveResponse.java
kbottla: Agree! after a bit of thinking renamed it to `QueueMessage`. If you have any other suggestions happy to take them into account.

It is almost a swap with SQS specific class so to avoid passing SQS related objects (`ReceiveMessageResult`) throughout the code base.

kbottla:
Thanks @mrlumbu . I have addressed your comments

>  I'm also interested in how you intend to expose this. i.e. Nothing else should know it's interacting with an sqs queue.

I envisage having the below structure when it is completed

**Writing to the queue**
`CardCaptureService -> CaptureQueue.sendForCapture(charge) -> SqsQueueService.sendMessage()`

**Reading from the queue**
`CardCaptureProcess -> CaptureQueue.getChargesToCapture()  gets List<QueueMessage> and converts them as required by CardCaptureProcess -> SqsQueueService.receiveMessages() returns CardCaptureProcess`

Happy to discuss them further..

mrlumbu (approved):
Looks good! We'll figure out how this gets wired in as we go

PR #1309, opened 2019-05-15
PP-5159 Authorise request with idempotency key

mrlumbu:
We are going to remove the locking for Stripe authorisation, so in order to
manage the case where we may process two authorise requests for same
charge simultaneously, we need to send an idempotency key to Stripe.
This commit does that by making use of existing StripeRequest base class


kbottla (approved):
LGTM!

PR #1312, opened 2019-05-16
[DO NOT MERGE] PP-5166 write charge to queue

kbottla:
## WHAT YOU DID

Changes to
- Write charges to Capture queue
- Wiremock config for SQS queue
- Units/Integration tests for captureUsingSQS feature flag enabled
- and some fixes to existing tests that implicitly required capture queue

## How to test
- CI should pass
- Should start send charges to capture queue when deployed
- Any failures to write to SQS should be ignored

### DIFF
--- a/src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
@@ -36,9 +36,11 @@ import uk.gov.pay.connector.gateway.CaptureResponse;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseCaptureResponse;
 import uk.gov.pay.connector.queue.CaptureQueue;
+import uk.gov.pay.connector.queue.QueueException;
 import uk.gov.pay.connector.usernotification.service.UserNotificationService;
 
 import javax.persistence.OptimisticLockException;
+import javax.ws.rs.WebApplicationException;
 import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.Optional;
@@ -52,6 +54,7 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.anyString;
 import static org.mockito.Mockito.argThat;
 import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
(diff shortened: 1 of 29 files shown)


heathd:
you've moved the `@Transactional` annotations around and I'd like to see some explanation in the commit messages as to the rationale/reasoning behind this change. 

I'd also like to see a bit more narrative explanation in the "Wiremock for SQS" commit -- what are we mocking? Why? How?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
```diff
@@ -90,6 +91,8 @@
     private Appender<ILoggingEvent> mockAppender;
     @Mock
     private CaptureQueue mockCaptureQueue;
+    @Mock
+    private CaptureProcessConfig mockCaptureProcessConfig;
```
heathd, line 98: given that this is just a vanilla value object, does it need to be mocked?

heathd: (not sure what the standard pattern is around this)

oswaldquek:
Is wiremock appropriate for stubbing out an sqs queue? Sqs queue urls are http so that's ok but my question would be around a full integration test in the future where some component puts a message in the queue, and that message should be picked up from another component listening on the queue.

heathd:
Have discussed further with Krishna, plan is to:

1. use the docker container for the `CardResourceCaptureWithSqsQueueITest`
2. use wiremock for all other integration tests
3. control which mock is configured by using an new parameter on the `@DropwizardConfig` annotation
4. review if it’s possible to remove the wiremock setup from `captureChargeUrlWithSqsMockFor`, so that the `Arrange` and `Act` steps of the `it` tests remain separate
5. don’t update status to CAPTURE_ERROR if the write to the SQS queue fails. We should just return an error to the user saying that the operation failed and not modify the charge state.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
kbottla: Agree we don't need mocking for vanilla object. But for configuration classes we don't use setters and hence mocking is required.

kbottla:
All have been addressed in separate PRs

**1. use the docker container for the `CardResourceCaptureWithSqsQueueITest`**
Using docker container https://github.com/alphagov/pay-connector/blob/b88a7267397852d73350ae89bb3a7bc3cf0e49f2/src/test/java/uk/gov/pay/connector/it/resources/CardResourceCaptureWithSqsQueueIT.java#L39
**2. use wiremock for all other integration tests**
  https://github.com/alphagov/pay-connector/commit/84860a0eb8bd364fd280b4a6440f4792e9bfc6ff#diff-f6c4a8a691668117480ef80714262f0aR56
**3. control which mock is configured by using an new parameter on the `@DropwizardConfig` annotation**
New flag is available to control docker SQS container https://github.com/alphagov/pay-connector/commit/3bafc33278be28b8dee621710de5dced3cc9b397#diff-a6ad3e1de4f7b424f9b2dda42ebab721R64
**4. review if it’s possible to remove the wiremock setup from `captureChargeUrlWithSqsMockFor`, so that the `Arrange` and `Act` steps of the `it` tests remain separate**
Not required any more as SQS is used for integration tests and wiremock for SQS is used for contract tests
**5. don’t update status to CAPTURE_ERROR if the write to the SQS queue fails. We should just return an error to the user saying that the operation failed and not modify the charge state.**
Not updating charge status on failing to write to SQS https://github.com/alphagov/pay-connector/blob/325a69cef8b6c54e632d25bcfad0f65bb3cbcc6e/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java#L144

PR #1315, opened 2019-05-17
PP-5166 write charge to queue with docker container for ITests

kbottla:
## WHAT YOU DID

Changes to
- Write charges to Capture queue
- SQS Docker rule for integration tests

_Pending tests_


kbottla:
Will raise a separate PR with only SQSDocker rule

PR #1316, opened 2019-05-17
PP-5167 Managed message receiver

sfount:
Basic life cycle for receiving messages; will be updated to use `CaptureQueue` abstraction with the introduction of https://github.com/alphagov/pay-connector/pull/1312

Requirements: 
- [x] tests for deleting messages in the SQSQueueService 
- [x] unit tests for receiver thread responding and logging

Using locally: 
* `pay local up --cluster card --local connector`
* Add things to queue: `aws --endpoint-url http://localhost:9324 sqs send-message --queue-url http://localhost:9324/queue/pay_capture_queue --message-body "capture first message"`
* After configured delay, connector will consume the message


pom.xml
```diff
@@ -42,6 +42,12 @@
             <groupId>uk.gov.pay</groupId>
             <artifactId>utils</artifactId>
             <version>${pay-java-commons.version}</version>
+            <exclusions>
```
sfount: This can be removed - it should have been addressed in https://github.com/alphagov/pay-connector/pull/1317/files

kbottla: Addressed here https://github.com/alphagov/pay-connector/pull/1321 but this PR is failing on CI for some reason but works fine locally

src/main/java/uk/gov/pay/connector/queue/managed/SQSMessageReceiver.java
```diff
@@ -0,0 +1,68 @@
...
+        scheduledExecutorService = environment
+                .lifecycle()
+                .scheduledExecutorService(SQS_MESSAGE_RECEIVER_NAME)
+                .threads(TOTAL_MESSAGE_RECEIVER_THREADS)
+                .build();
+    }
+    
+    @Override
+    public void start() { 
+        int INITIAL_DELAY = 1;
```
mrlumbu: `SCREAMING_SNAKE_CASE` is conventionally reserved for static final fields. I understand (and applaud) the intent to make this readable by using named variables, but please use camel case

src/main/java/uk/gov/pay/connector/queue/managed/SQSMessageReceiver.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.queue.CaptureQueue;
+import uk.gov.pay.connector.queue.QueueException;
+
+import javax.inject.Inject;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public class SQSMessageReceiver implements Managed {
```
mrlumbu: I don't think this class has anything to do with SQS, so should not refer to it in its name. Either `QueueMessageReciever` or `CaptureQueueMessageReciever` would be fine, depending on how we envisage this classes responsibilities growing in future (we don't have to decide now, so probably `CaptureQueueMessageReciever`?)

src/main/java/uk/gov/pay/connector/queue/managed/SQSMessageReceiver.java
mrlumbu: also, ideally these values would be fetched from config, and set by env var with sensible defaults. Happy for that to be in a separate PR though

src/main/java/uk/gov/pay/connector/queue/sqs/SqsQueueService.java
```diff
@@ -35,15 +34,29 @@ public QueueMessage sendMessage(String queueUrl, String messageBody) throws Queu
         }
     }
 
-    public List<QueueMessage> receiveMessages(String queueUrl) throws QueueException {
+    public List<QueueMessage> receiveMessages(String queueUrl, String messageAttributeName) throws QueueException {
         try {
-            ReceiveMessageResult receiveMessageResult = sqsClient.receiveMessage(queueUrl);
+            Integer MAX_MESSAGES = 10;
```
mrlumbu: I think this should either be in config, or just inline in the method call (it is clear what it is from name of method)

src/main/java/uk/gov/pay/connector/queue/sqs/SqsQueueService.java
```diff
@@ -35,15 +34,29 @@ public QueueMessage sendMessage(String queueUrl, String messageBody) throws Queu
...
+            
+            ReceiveMessageRequest receiveMessageRequest = new ReceiveMessageRequest(queueUrl);
+            receiveMessageRequest.setMaxNumberOfMessages(MAX_MESSAGES);
+
+            ReceiveMessageResult receiveMessageResult = sqsClient.receiveMessage(receiveMessageRequest.withMessageAttributeNames(messageAttributeName));
```
mrlumbu:
```
ReceiveMessageRequest receiveMessageRequest = new ReceiveMessageRequest(queueUrl)
                    .withMaxNumberOfMessages(MAX_MESSAGES)
                    .withMessageAttributeNames(messageAttributeName);

            ReceiveMessageResult receiveMessageResult = sqsClient.receiveMessage(receiveMessageRequest);
```
?

src/main/java/uk/gov/pay/connector/queue/sqs/SqsQueueService.java
```diff
@@ -35,15 +34,29 @@ public QueueMessage sendMessage(String queueUrl, String messageBody) throws Queu
...
 
             return QueueMessage.of(receiveMessageResult);
         } catch (AmazonSQSException | UnsupportedOperationException e) {
             logger.error("Failed to receive messages from SQS queue - {}", e.getMessage());
             throw new QueueException(e.getMessage());
         }
     }
-
+    
+    public DeleteMessageResult deleteMessage(String queueUrl, QueueMessage message) throws QueueException {
```
mrlumbu: Slightly cleaner if this just took a `messageReceiptHandle`? Since that's all that's needed

src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
```diff
@@ -32,6 +43,16 @@ public void sendForCapture(String externalId) throws QueueException {
...
         logger.info("Charge [{}] added to capture queue. Message ID [{}]", externalId, queueMessage.getMessageId());
     }
-
-
+    
+    public void receiveCaptureMessages() throws QueueException {
+        List<QueueMessage> captureMessages = sqsQueueService.receiveMessages(this.captureQueueUrl, CAPTURE_MESSAGE_ATTRIBUTE_NAME);
+        for (QueueMessage message: captureMessages) {
+            try {
+                cardCaptureMessageProcess.runCapture(message);
```
mrlumbu: I know you discussed this, but this seems like it's doing too much given the method name - I would expect the caller of `receiveCaptureMessages` to take the messages and do sensible things with them

src/main/java/uk/gov/pay/connector/queue/managed/SQSMessageReceiver.java
```diff
@@ -0,0 +1,68 @@
...
+    
+    private final String SQS_MESSAGE_RECEIVER_NAME = "sqs-message-receiver";
+    private final int TOTAL_MESSAGE_RECEIVER_THREADS = 1;
```
mrlumbu: as with above, this should be in config (ideally), or `private static final`

src/test/java/uk/gov/pay/connector/queue/managed/SQSMessageReceiverTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+import uk.gov.pay.connector.queue.CaptureQueue;
+
+import java.util.concurrent.ScheduledExecutorService;
+
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class SQSMessageReceiverTest {
```
mrlumbu: I salute your diligence but not sure we need these tests. What are they guarding against?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureMessageProcess.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.connector.paymentprocessor.service;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.queue.QueueMessage;
+
+import javax.inject.Inject;
+
+// @TODO(sfount) replace `CardCaptureProcess` when feature flag is switched
+public class CardCaptureMessageProcess {
```
mrlumbu, line 13: Not sure why this is a `Process`? Also not sure if we need this over just using `ChargeCaptureService`?

mrlumbu (changes requested):
Looks like a good start. I have made some suggestions around naming that I think clarify what's going on (at least for me). I also have some questions about the level at which things are happening - e.g. the CaptureQueue doing the capture itself. Might be good to talk this through

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureMessageProcess.java
sfount: This will replace the current `CardCaptureProcess` when the feature flag is fully enabled, this is a parallel class with similar interface. Moved receive + process logic into this class to reduce complexity in `CaptureQueue`

src/main/java/uk/gov/pay/connector/queue/sqs/SqsQueueService.java
sfount: 👍

src/main/java/uk/gov/pay/connector/queue/sqs/SqsQueueService.java
sfount: 👍

src/test/java/uk/gov/pay/connector/queue/managed/SQSMessageReceiverTest.java
sfount: This just tests that the correct number of processes is spun up by the scheduler - I think it will become more useful when there is more custom behaviour (rather than just one thread running)

src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
sfount: Moved receive + process logic, ref: https://github.com/alphagov/pay-connector/pull/1316#discussion_r285974600

src/main/java/uk/gov/pay/connector/queue/sqs/SqsQueueService.java
```diff
@@ -1,15 +1,14 @@
 package uk.gov.pay.connector.queue.sqs;
 
 import com.amazonaws.services.sqs.AmazonSQS;
-import com.amazonaws.services.sqs.model.AmazonSQSException;
-import com.amazonaws.services.sqs.model.ReceiveMessageResult;
-import com.amazonaws.services.sqs.model.SendMessageResult;
+import com.amazonaws.services.sqs.model.*;
```
mrlumbu: we try to avoid wildcard imports to keep @alexbishop1 happy. But can be done in next PR

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureMessageProcess.java
```diff
@@ -0,0 +1,40 @@
...
+    private final CaptureQueue captureQueue;
+
+    @Inject
+    public CardCaptureMessageProcess(CaptureQueue captureQueue) { 
+        this.captureQueue = captureQueue;
+    }
+    
+    public void handleCaptureMessages() throws QueueException { 
+        List<QueueMessage> captureMessages = captureQueue.receiveCaptureMessages();     
+        for (QueueMessage message: captureMessages) {
```
mrlumbu, line 25: might be nice to `stream` this, but can be done later

mrlumbu (dismissed):
👍Nice one

mrlumbu (approved):

PR #1319, opened 2019-05-17
Separate integration tests from unit tests

rhowe-gds:
Integration tests are slow. They involve starting a PostgreSQL container and
instantiating Dropwizard.

Split them from the unit tests and run the unit tests with the surefire plugin
instead.

Rename integration tests to match the default include/exclude rules for the
surefire and failsafe maven plugins. By default, the failsafe plugin will
include any tests matching IT*.java, *IT.java and *ITCase.java. This was not
the naming scheme our integration tests had, so fix that.

Since unit tests are so quick, we no longer need to skip them when running
local builds.

maven's 'package' target doesn't include running integration tests, so switch
to using the 'verify' target during CI builds, which is what we should've been
doing anyway.


rhowe-gds:
This classifies some of the tests in the `it` package(s) as unit tests.. I don't know if that's a deal-breaker?

alexbishop1 (approved):

PR #1321, opened 2019-05-20
PP-5166 SQS container for tests

kbottla:
## WHAT

- Adds support to start SQS container (using org.testcontainers) for tests.
  Creating Dropwizard annotation with flag `withDockerSQS` set to true starts SQS test container before the tests

      @DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml",
        configOverrides = {@ConfigOverride(key = "captureProcessConfig.captureUsingSQS", value = "true")},
        withDockerSQS = true
      )
- Excluded `aws-xray-recorder-sdk-aws-sdk-instrumentor` as library automatically instruments all SDK clients (SQS Client currently).
  Required clients can still be instrumented manually https://docs.aws.amazon.com/xray/latest/devguide/scorekeep-sdkclients.html

## HOW

- Spins up SQS container for tests but not used yet. CI builds should continue to pass

### DIFF
--- a/src/test/java/uk/gov/pay/connector/junit/SqsTestDocker.java
+++ b/src/test/java/uk/gov/pay/connector/junit/SqsTestDocker.java
@@ -1,0 +1,118 @@
+package uk.gov.pay.connector.junit;
+
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.client.builder.AwsClientBuilder;
+import com.amazonaws.services.sqs.AmazonSQS;
+import com.amazonaws.services.sqs.AmazonSQSClientBuilder;
+import com.amazonaws.services.sqs.model.ListQueuesResult;
+import com.google.common.base.Stopwatch;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.GenericContainer;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Start a Generic Test container with SQS docker image and creates queues if required.
+ * No need to call stop() as container will be stopped after executing the tests
+ */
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/junit/SqsTestDocker.java
```diff
@@ -0,0 +1,118 @@
...
+            logger.error("Exception initialising SQS Container - {}", e.getMessage());
+            throw new SqsTestDockerException(e);
+        }
+    }
+
+    private static void createContainer() {
+        if (sqsContainer == null) {
+            // wait at least 3 minutes for container to startup to avoid timeout exception. Default is 60 seconds
+            sqsContainer = new GenericContainer("roribio16/alpine-sqs")
+                    .withStartupTimeout(Duration.ofMinutes(3));
```
jankowiakmaria: Possibly the better option (not requiring `waitForSQSContainerToStart`) would be to use different waiting strategy (reference: https://www.testcontainers.org/features/startup_and_waits/): 
```java
.withExposedPorts(80)
.waitingFor(Wait.forHttp("/"));
```
withe the port exposed by sqs and end the endpoint pointing to the queue we're interested in (https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueAttributes.html)

sfount:
Following @jankowiakmaria comments: https://github.com/alphagov/pay-connector/pull/1321/commits/89ce1b2742c6f15e3fef0e13ed8dd545da65ed11 makes use of the SQS `/?Action=GetQueueUrl&QueueName=` action to greatly simplify waiting on queue availability.

heathd (approved):

heathd:
Looks fine

It's a shame to completely disable the XRay instrumentation as that would probably be useful in production, but let's roll with this for now to get things working. Maybe we can figure out in another PR  a way to disable it just for tests?

PR #1324, opened 2019-05-21
[WIP] Pp 5237 Expire charges stuck in Auth Error state

oswaldquek:
When making 3ds Authorisation requests, the EPDQ server sometimes responds with
a 503 status, which causes the charge to be transitioned to AUTHORISATION ERROR
in connector. It appears that despite the 503, the charge will still be in the
AUTHORISED SUCCESS (or 5-Authorised) state. This makes sense as it was already
authorised in the first authorisation request before the 3ds one. This commit
therefore adds the AUTHORISED ERROR state as one of the statuses upon which the
charge should be cancelled.

I had considered using the DiscrepancyService to determine if there is a
discrepancy before cancelling the charge; however looking at EpdqStatusMapper
and WorldpayStatus classes these PSPs never have a state that is equivalent to
our AUTHORISATION ERROR status so I surmise it is always safe to call cancel on
a charge that has errored.

Add test using epdq as a PSP to spice things up.


alexbishop1:
It’s great that you’ve identified this problem. However, I’m concerned about proposed solution. Two reasons:

1. I think this will mean that all `AUTHORISATION ERROR` payments will eventually become `EXPIRED`, which doesn’t seem desirable as it will hide errors from us and misrepresent what happened to services
2. Technically, `AUTHORISATION ERROR` is a terminal status i.e. it shouldn’t change to something else afterwards (though this is arguably not a very serious problem)

I’m not exactly sure what the solution is. The best I can think of would be something along the lines of making a payment that’s in `AUTHORISATION 3DS READY` transition to a new status, say `AUTHORISATION 3D ERROR`, if there’s an error. This new status would be expirable but rather than moving to `EXPIRED`, it would become another new status called `AUTHORISATION ERROR EXPIRED` or something. Both these two new statuses would map to the existing external error status. This way, we could still cancel with the gateway but we won’t be losing the fact that there was an error.

oswaldquek:
@alexbishop1 Thanks for that. Sounds like a plan.

PR #1325, opened 2019-05-21
PP-5166 Add charge to queue

mrlumbu:
- Sends charge to queue during capture approve action
- Unit tests for adding charges to queue and exception cases
- Integration tests with SQS docker container
- Contract tests uses Wiremock to stub SQS calls - If an SQS container to be used
   `DropwizardAppWithPostgresRule` should spin new SQS docker container
- Moved `markChargeAsEligibleForCapture` from ChargeService to `CardCaptureService`
  as business logic is close to CardCaptureService and adding charge to queue is an
  intermediate step in marking charge as eligible

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
@@ -5,24 +5,32 @@ import com.google.inject.persist.Transactional;
 import io.dropwizard.setup.Environment;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.model.domain.FeeEntity;
 import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
+import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
+import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
 import uk.gov.pay.connector.fee.dao.FeeDao;
 import uk.gov.pay.connector.gateway.CaptureResponse;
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -124,6 +151,19 @@ public void persistFee(ChargeEntity charge, Long feeAmount) {
         feeDao.persist(fee);
     }
 
+    private void addChargeToCaptureQueue(String externalId) {
+        if (captureUsingSQS) {
+            try {
+                captureQueue.sendForCapture(externalId);
```
sfount, line 157: With: @jankowiakmaria. It looks like `CaptureQueue` could be the one that knows what to do with the charge -- this was if the message format changes or any further details need to be send this interface wouldn't have to change. 

Could be done in another PR but passing the charge entity into `sendForCapture` might decouple these pieces well.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -124,6 +151,19 @@ public void persistFee(ChargeEntity charge, Long feeAmount) {
...
+            } catch (QueueException e) {
+                logger.error("Exception sending charge [{}] to capture queue", externalId);
+                throw new WebApplicationException(format(
```
sfount, line 160: Unsure if this should be a `WebApplicationException`, seems like it will only be performed on separate background process threads. This could just be an understanding of Connector gubbins.

sfount (approved):
Code LGTM 👍  
* minor comments about the interface for send to capture, these could either be done here or in a follow up PR

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
kbottla: This is invoked through resource endpoint (`v1/frontend/charges/{chargeId}/capture`) when a user confirm payment / when service approve payment for capture

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
kbottla: 👍 will address in a separate PR

PR #1326, opened 2019-05-21
PP-5170 Process charge capture from sqs message

sfount:
* call capture charge on capture queue messages 
* models messages with Jackson library

Co-authored-by: @jankowiakmaria

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
+++ b/src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
@@ -5,16 +5,13 @@ import com.google.gson.GsonBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
-import uk.gov.pay.connector.queue.sqs.CaptureCharge;
 import uk.gov.pay.connector.queue.sqs.SqsQueueService;
-import uk.gov.pay.connector.queue.sqs.ChargeCaptureMessage;
 import uk.gov.pay.connector.util.JsonObjectMapper;
 
 import javax.inject.Inject;
 import javax.ws.rs.WebApplicationException;
 import java.util.List;
 import java.util.Objects;
-import java.util.function.Function;
 import java.util.stream.Collectors;
 
 public class CaptureQueue {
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
```diff
@@ -39,11 +49,31 @@ public void sendForCapture(String externalId) throws QueueException {
...
+            try {
+                CaptureCharge captureCharge = jsonObjectMapper
+                        .getObject(qm.getMessageBody(), CaptureCharge.class);
+                return ChargeCaptureMessage.of(captureCharge, qm);
+            } catch (WebApplicationException e) {
+                logger.warn("Error parsing the charge capture message from queue [{}]", e.getMessage());
+                return null;
+            }
+        };
+    }
```
mrlumbu, line 69:
```
public List<ChargeCaptureMessage> retrieveChargesForCapture() throws QueueException {
        List<QueueMessage> queueMessages = sqsQueueService
                .receiveMessages(this.captureQueueUrl, CAPTURE_MESSAGE_ATTRIBUTE_NAME);

        return queueMessages
                .stream()
                .map(this::toChargeCaptureMessage)
                .filter(Objects::nonNull)
    }

    private ChargeCaptureMessage toChargeCaptureMessage(QueueMessage qm) {
        try {
            CaptureCharge captureCharge = jsonObjectMapper
                    .getObject(qm.getMessageBody(), CaptureCharge.class);
            return ChargeCaptureMessage.of(captureCharge, qm);
        } catch (WebApplicationException e) {
            logger.warn("Error parsing the charge capture message from queue [{}]", e.getMessage());
            return null;
        }
    }
```
is cleaner

src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
```diff
@@ -39,11 +49,31 @@ public void sendForCapture(String externalId) throws QueueException {
...
+                .filter(Objects::nonNull)
+                .collect(Collectors.toList());
     }
-    
-    public void markMessageAsProcessed(QueueMessage message) throws QueueException {    
+
+    private Function<QueueMessage, ChargeCaptureMessage> getChargeCaptureMessageFunction() {
+        return qm -> {
+            try {
+                CaptureCharge captureCharge = jsonObjectMapper
```
mrlumbu: it occurs to me that this mapping from string to object would more nicely done as a static factory in CaptureCharge class, but that's a minor thing

src/main/java/uk/gov/pay/connector/queue/sqs/ChargeCaptureMessage.java
```diff
@@ -0,0 +1,26 @@
+package uk.gov.pay.connector.queue.sqs;
```
mrlumbu: don't think this is sqs specific. I'm quite keen to try and separate interface and implementation for this queue stuff. Finessing that can wait, but we should at least aim to keep implementation agnostic classes out of the implementation package

mrlumbu (changes requested):
Looks good. Few minor points

src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
sfount: Agreed! Great suggestion

src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
sfount: Did consider this - can be improved in a further PR I think

mrlumbu (approved):

PR #1329, opened 2019-05-22
zd3689082: Increase cancel timeout

oswaldquek:
There are a significant number of cancel requests that take longer than 2
seconds to complete. This is causing many cancel operations to not actually
cancel with ePDQ at all.


heathd (approved):

heathd:
confirmed from metrics that cancel can time out

PR #1331, opened 2019-05-22
zd3689082: Log charge id when making POST requests

oswaldquek:
We got quite a lot of logs in March like

`Connection timed out error for gateway url=...`
`Gateway returned unexpected status code: 503...`

These come from https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java.

It'd be useful to tie the chargeId to these logs so we can have another level
of confirmation what a charge's status should be from the calls that were made.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
+import uk.gov.pay.connector.app.ChargeIdMDCRequired;
 import uk.gov.pay.connector.charge.service.ChargeCancelService;
 import uk.gov.pay.connector.gateway.model.Auth3dsDetails;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
@@ -37,6 +38,7 @@ import static uk.gov.pay.connector.util.ResponseUtil.badRequestResponse;
 import static uk.gov.pay.connector.util.ResponseUtil.serviceErrorResponse;
 
 @Path("/")
+@ChargeIdMDCRequired
 public class CardResource {
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
@@ -65,7 +67,6 @@ public class CardResource {
     @Produces(APPLICATION_JSON)
     public Response authoriseCharge(@PathParam("chargeId") String chargeId, 
                                     @NotNull @Valid ApplePayAuthRequest applePayAuthRequest) {
-        MDC.put("chargeId", chargeId);
         logger.info("Received encrypted payload for charge with id {} ", chargeId);
         return applePayService.authorise(chargeId, applePayAuthRequest);
     }
@@ -76,7 +77,6 @@ public class CardResource {
     @Produces(APPLICATION_JSON)
     public Response authoriseCharge(@PathParam("chargeId") String chargeId,
                                     @NotNull @Valid GooglePayAuthRequest googlePayAuthRequest) {
-        MDC.put("chargeId", chargeId);
         logger.info("Received encrypted payload for charge with id {} ", chargeId);
         return googlePayService.authorise(chargeId, googlePayAuthRequest);
     }
@@ -87,7 +87,6 @@ public class CardResource {
     @Produces(APPLICATION_JSON)
     public Response authoriseCharge(@PathParam("chargeId") String chargeId,
                                     @Valid AuthCardDetails authCardDetails) {
-        MDC.put("chargeId", chargeId);
         AuthorisationResponse response = cardAuthoriseService.doAuthorise(chargeId, authCardDetails);
 
         return response.getGatewayError().map(error -> handleError(chargeId, error))
@@ -113,7 +112,6 @@ public class CardResource {
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
     public Response authorise3dsCharge(@PathParam("chargeId") String chargeId, Auth3dsDetails auth3DsDetails) {
-        MDC.put("chargeId", chargeId);
         Gateway3DSAuthorisationResponse response = card3dsResponseAuthService.process3DSecureAuthorisation(chargeId, auth3DsDetails);
         return response.isDeclined() ? badRequestResponse("This transaction was declined.") : handleGateway3DSAuthoriseResponse(response);
     }
@@ -134,7 +132,6 @@ public class CardResource {
     @Consumes(APPLICATION_JSON)
     @Produces(APPLICATION_JSON)
     public Response captureCharge(@PathParam("chargeId") String chargeId) {
-        MDC.put("chargeId", chargeId);
         cardCaptureService.markChargeAsEligibleForCapture(chargeId);
         return ResponseUtil.noContentResponse();
     }
@@ -154,7 +151,6 @@ public class CardResource {
     @Path("/v1/api/accounts/{accountId}/charges/{chargeId}/cancel")
     @Produces(APPLICATION_JSON)
     public Response cancelCharge(@PathParam("accountId") Long accountId, @PathParam("chargeId") String chargeId) {
-        MDC.put("chargeId", chargeId);
         return chargeCancelService.doSystemCancel(chargeId, accountId)
                 .map(chargeEntity -> Response.noContent().build())
                 .orElseGet(() -> ResponseUtil.responseWithChargeNotFound(chargeId));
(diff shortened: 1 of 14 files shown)


alexbishop1:
Most of the time, we should output a representation of the gateway response (see https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java#L91 and https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java#L79 and a few similar lines) but it does require us to know what to look for in the response and I’m not sure this is consistently done for every gateway (I’m not sure Stripe responses get logged like that) and possibly not for Apple Pay and Google Pay. Logging the whole response seems fine but we have to be careful we don’t accidentally log personally-identifiable information.

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -3,6 +3,7 @@
 import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
```
rhowe-gds: `pay-java-commons` uses jboss's MDC. I don't know if mixing them is a good idea?
We should probably switch to using the slf4j one? I don't know what if any difference there is

rhowe-gds:
If you add something to the MDC my understanding is that it will persist for the lifetime of the thread. This presumably means that following requests may inherit a `chargeId` of a previous request unless we guarantee that every request to connector sets the `chargeId` key in the MDC

We may be setting ourselves up for a world of pain here - using the power of Splunk we should be able to tie things together by request ID instead?

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
oswaldquek: Do you mean switching the one in `pay-java-commons`? I don't think there's a problem with this line as we do a similar thing at https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureProcess.java#L53

oswaldquek:
@rhowe-gds I think you're right about following requests inheriting a `chargeId`. Looking at https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureProcess.java#L94 it looks like the requestID is removed. This means I'll have to do a similar thing but it means writing `try...finally`s everywhere :(

oswaldquek:
@rhowe-gds I don't think we can use the requestID.... it's already being logged and I can't tell what external requests to PSPs relate to which charge ids.

oswaldquek:
@rhowe-gds maybe a better solution is to to `MDC.put("chargeId", chargeId)` in https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java#L56 instead? Then we can remove it in the finally block?

oswaldquek:
@alexbishop1 you're right. https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java#L85 already logs the raw response from the PSP. i've removed the code i've added to log the response from the server.

rhowe-gds:
> @rhowe-gds maybe a better solution is to to `MDC.put("chargeId", chargeId)` in https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java#L56 instead? Then we can remove it in the finally block?

Seems fair. You only need to use the MDC if you want a shorthand way of passing the chargeId around though - if the logs you want the chargeId to be associated with are logged by that method we can just include it in the log message and the MDC doesn't need to know about it at all.

heathd:
> We may be setting ourselves up for a world of pain here - using the power of Splunk we should be able to tie things together by request ID instead?

I think that's a bit optimistic unfortunately. When you're trying to look at patterns of behaviour, tying things to context via the requestID is challenging. ChargeID is so ubiquitous that I think there's good reason to have it on every log line where it's appropriate.

In the ideal world that would mean:

- for HTTP request we should set it at the start of the HTTP request and clear it at the end
- for background threads (as is the case here) we should set it when the thread starts processing a charge and clear it when done

As Russell said there the MDC is a thread local so you do need to be careful to clear it, and I think using a try/finally block is the only reliable way to do that. 

For the case of this PR, hooking into a try/finally block in the gateway client seems reasonable

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
heathd: I suspect the use of Jboss MDC in pay-java-commons is a mistake

alexbishop1:
@oswaldquek Well, it’s not quite the raw response (it’s the stringification of the Java object we deserialise the response into) but it’s usually pretty good (we’ve improved it a few times to extract more relevant information). For ePDQ payments, it gets the status and the error messages, which is the main thing.

oswaldquek:
@alexbishop1 oh, I was wrong.  https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java#L85 DOESN'T log the raw response from the PSP. it's because our error handling code is ridiculously hard to read 💢

oswaldquek:
AOP sounds good for this type of situation. Sadly it doesn't work out of the box with dropwizard 😢 
I'll investigate doing this in a filter.

oswaldquek:
@alexbishop1 @heathd @rhowe-gds I've used the MDC to log the request id now. There are a few new commits but the meat of it is at https://github.com/alphagov/pay-connector/pull/1331/commits/3dff7b9089efa5bede7174cc44c41d8660403c42. 

Unfortunately, testing these things is impossible in my opinion. You can satisfy yourselves that this works by running https://github.com/alphagov/pay-connector/blob/d74c53b539489cf0aa5a9a4542aacc940f8681ae/src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java#L273 in Intellij. You should see something like `[2019-05-23 16:01:53.379] [thread=CardExecutorService-0] [level=ERROR] [logger=u.g.p.c.g.GatewayClient] [requestID=(none)] [chargeID=charge-1186667931] - Exception for gateway url=http://localhost:57596/jsp/merchant/xml/paymentService.jsp, error message: java.net.SocketException: Connection reset` logged.

heathd (dismissed):
A couple of minor comments which could be addressed in a later PR 

Looks good, thanks!

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java
```diff
@@ -109,6 +109,7 @@ public ExecutorService getExecutor() {
                 return task.call();
             } finally {
                 xrayUtils.endSegment();
+                MDC.clear();
```
heathd, line 112: ok this may be a bit picky but I think:
- the try{ should enclose line 100 `setContextMap` 
- possibly we should store the MDC before setting and then restore it?

oswaldquek: hmm does it matter? we have this logged on line 103: `logger.debug("Card operation task spent {} ms in queue", totalWaitTime);` do we not want to log the chargeId there too?

oswaldquek: alternatively we could move the `try {` to line 100, i.e. before the `MDC.setContextMap(mdcContextMap);`

heathd: I was thinking of:
```java
    // accepts a supplier function and executed that in a separate Thread of its own.
    // returns a Pair of the execution status and the return type
    public <T> Pair<ExecutionStatus, T> execute(Supplier<T> callable) {
        Callable<T> task = callable::get;
        Map<String, String> mdcContextMap = Optional.ofNullable(MDC.getCopyOfContextMap()).orElse(Map.of());
        final long startTime = System.currentTimeMillis();

        Future<T> futureObject = executor.submit(() -> {
            try {
              MDC.setContextMap(mdcContextMap);
              xrayUtils.beginSegment();
              long totalWaitTime = System.currentTimeMillis() - startTime;
              logger.debug("Card operation task spent {} ms in queue", totalWaitTime);
              if (totalWaitTime > QUEUE_WAIT_WARN_THRESHOLD_MILLIS) {
                  logger.warn("CardExecutor Service delay - queue_wait_time={}", totalWaitTime);
              }
              metricRegistry.histogram("card-executor.delay").update(totalWaitTime);
              return task.call();
            } finally {
                xrayUtils.endSegment();
                MDC.clear();
            }
        });
```

heathd (approved):
replied, but I don't think it's blocking

PR #1332, opened 2019-05-22
PP-5167 Capture queue - add max wait time for long polling

sfount:
In local environments the client must specify the wait time for a message request to enable long polling, this behaviour should be verified for a configured SQS instance. PR tests "long polling" attribute.


mrlumbu (approved):
We should get this stuff in config soonish, but ok for now

sfount:
Environment configs for receiving behaviour to follow: 
* number of messages to receive at once . 
* how long to poll for messages 
* how long to delay the thread after polling the queue
* the number of threads required to poll queues

PR #1333, opened 2019-05-22
PP-5166 Change `sendMessage` signature

kbottla:
## WHAT YOU DID
- CaptureQueue.sendMessage now requires entire charge as parameter. Any changes to the message format can be dealt by CaptureQueue without a need for new attributes from other services/change in signature

## How to test

- CI should pass

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
@@ -92,7 +92,7 @@ public class CardCaptureService {
             return chargeService.updateChargeStatus(charge, AWAITING_CAPTURE_REQUEST);
 
         } else {
-            addChargeToCaptureQueue(externalId);
+            addChargeToCaptureQueue(charge);
             try {
                 return chargeService.updateChargeStatus(charge, CAPTURE_APPROVED);
             } catch (InvalidStateTransitionException e) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -110,7 +110,8 @@ void markChargeAsCaptureError(String chargeId) {
 
     @Transactional
     public ChargeEntity markChargeAsCaptureApproved(String externalId) {
-        addChargeToCaptureQueue(externalId);
+        ChargeEntity charge = chargeService.findChargeById(externalId);
```
jankowiakmaria, line 113: It might be better to put it in `addChargeToCaptureQueue` method. This method doesn't use it for anything other than passing to the private method in the same class.
(Nothing major - just a comment)

jankowiakmaria: @kbottla explained to me that it makes sense (at least for now) to do it this way. In `markChargeAsEligibleForCapture` we already retrieve the charge so if we passed id it would have caused calling `chargeService.findChargeById(externalId)` twice.

jankowiakmaria (approved):

sfount (approved):
Code LGTM 👍. Addresses comments from https://github.com/alphagov/pay-connector/pull/1325

PR #1336, opened 2019-05-23
PP-4060 clean up expired tokens sweep database migration

nimalank7:
## WHAT YOU DID

This PR adds a: 

1. **created_date** column with data type 'timestamp without timezone' to the Tokens table in Connector's database with a default value of now().

2. **created_date** field in TokenEntity. The TokenEntity constructor fills this out with the created_date field from ChargeEntity

### DIFF
--- a/src/main/java/uk/gov/pay/connector/token/model/domain/TokenEntity.java
+++ b/src/main/java/uk/gov/pay/connector/token/model/domain/TokenEntity.java
@@ -28,7 +28,7 @@ public class TokenEntity extends AbstractVersionedEntity {
 
     @Column(name = "secure_redirect_token")
     private String token;
-
+    
     @ManyToOne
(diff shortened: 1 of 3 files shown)


alexbishop1:
Could you split the database migrations and the code changes into two separate PRs please?

In staging and production, we have to run database migrations manually. So if both the database migrations and the code changes went out at the same time, the code would have to handle both when the migration hasn’t yet been run and when it has. This is very hard to get right so, as a rule, we have database migrations in their own releases without any code changes. Once the migrations have been run in all environments, we can release code changes that rely on them having happened.

nimalank7:
@alexbishop1 I've removed the code leaving just the migrations.xml

heathd:
it would be good to should squash/rebase the commit history so that there's just one commit adding the migration.

heathd (approved):
Great! 💯

PR #1339, opened 2019-05-24
PP-5280 Fix SQSClient for non-standard SQS endpoint

kbottla:
## WHAT YOU DID
- AWS SQS SDK defines non-standard endpoint as a custom SQS url (http://own-sqs-url).
  While endpoint is not needed in EC2 environment (SQS SDK will determine based on region),
  it is required (along with credentials)  while spinning up SQS equivalent containers for development/CI.

  The change includes initiating SQSClient with `endpoint and also credentials` required for development and CI

- Renamed `CardResourceCaptureWithSqsQueueITest` -> CardResourceCaptureWithSqsQueueIT (to failsafe plug way for integration tests)

## How to test

- CI should continue to pass
- Should pass locally with the below command even when `.aws` folder (with credentials) doesn't exists in users' home directory
   ` mvn verify -Dit.test=uk.gov.pay.connector.it.resources.CardResourceCaptureWithSqsQueueIT`
### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.app;
 
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
 import com.amazonaws.client.builder.AwsClientBuilder;
 import com.amazonaws.services.sqs.AmazonSQS;
 import com.amazonaws.services.sqs.AmazonSQSClientBuilder;
@@ -28,8 +30,6 @@ import uk.gov.pay.connector.wallets.applepay.ApplePayDecrypter;
 import java.util.Properties;
 import java.util.Queue;
 import java.util.concurrent.ArrayBlockingQueue;
-
-import static org.apache.commons.lang3.StringUtils.isEmpty;
 
 public class ConnectorModule extends AbstractModule {
     final ConnectorConfiguration configuration;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -129,18 +129,27 @@ protected NotifyClientFactory getNotifyClientFactory(ConnectorConfiguration conn
...
-        if (isEmpty(connectorConfiguration.getSqsConfig().getEndpoint())) {
-            return AmazonSQSClientBuilder.standard()
-                    .withRegion(connectorConfiguration.getSqsConfig().getRegion())
-                    .build();
-        } else {
-            return AmazonSQSClientBuilder.standard()
+        AmazonSQSClientBuilder clientBuilder = AmazonSQSClientBuilder
+                .standard();
+
+        if (connectorConfiguration.getSqsConfig().isNonStandardServiceEndpoint()) {
```
mrlumbu, line 135: do you think we should do some validation of the config to ensure that if `isNonStandardServiceEndpoint` is true that an endpoint is set?

mrlumbu (changes requested):
One suggestion

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
kbottla: Integration tests are set with `isNonStandardServiceEndpoint` to true as its sole purpose is to support local development and CI. May be that is sufficient for this ?

If we want, is there a way to add validation to config fields based on the values of other ?

mrlumbu: don't think so (although @danworth may know of a way) - but you could still just check the condition holds and if not throw a RuntimeException of some sort

mrlumbu: but yeah, I guess this is just for tests for now, 👍

mrlumbu (approved):

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
kbottla: I will raise a separate PR for validation

PR #1352, opened 2019-05-29
PP-4060 test refactorings

heathd:
Some stylistic tweaks to the test

Feedback welcome. I talked this through with @nimalank7 

What I did:

- string literal rather than local vars for token names. Use of meaningful/intention revealing token names.
- date offsets also don't benefit much from local variables, especially if defined in relation to `today`
- introduce hamcrest matcher for optional which simplifies the assertions

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -342,6 +342,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>com.github.npathai</groupId>
+            <artifactId>hamcrest-optional</artifactId>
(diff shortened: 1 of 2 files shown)


alexbishop1 (commented):
This looks nice.

I’d not heard of Hamcrest Optional before. It looks like there’s a newer 2.0.0 release.

heathd:
I've updated to 2.0.0 of `hamcrest-optional`

kbottla (approved):
LGTM!

PR #1361, opened 2019-05-30
PP-5274: remove feature flag on sqs write

jankowiakmaria:
* removed feature flag when writing to the queue in the `CardCaptureService`
* test cleanup after removing config from `CardCaptureService`
* added fields to the log when receiving message from the queue - messageId and receiptHandle


sfount (approved):

jankowiakmaria:
We need to wait until e2e tests have sqs container ready (so that this change doesn't break them)

sfount:
Waiting on scripts::779

PR #1369, opened 2019-05-31
PP-5273 Write to Queue after DB transaction

kbottla:
## WHAT YOU DID
- Updating/Committing charge status (for CAPTURE_APPROVED) before sending the charge to queue so that
    - queue message receiver won't receive the charge before the charge status commited to CAPTURE_APPROVED state.

Receiving charge from queue and processing (before charge being committed with CAPTURE_APPROVED state) will lead to below exception as charge will be read/updated concurrently by `CardCaptureService` and `CardCaptureMessageProcess`

`org.eclipse.persistence.exceptions.OptimisticLockException
The object [uk.gov.pay.connector.charge.model.domain.ChargeEntity@xxxxxx] cannot be updated because it has changed or been deleted since it was last read.`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
@@ -11,8 +11,6 @@ import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.model.domain.FeeEntity;
 import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
-import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
-import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
 import uk.gov.pay.connector.fee.dao.FeeDao;
 import uk.gov.pay.connector.gateway.CaptureResponse;
 import uk.gov.pay.connector.gateway.PaymentProviders;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -85,21 +81,13 @@ public ChargeEntity prepareChargeForCapture(String chargeId) {
...
-    @Transactional
     public ChargeEntity markChargeAsEligibleForCapture(String externalId) {
-        final ChargeEntity charge = chargeService.findChargeById(externalId);
+        ChargeEntity charge = chargeService.markChargeAsEligibleForCapture(externalId);
 
-        if (charge.isDelayedCapture()) {
-            return chargeService.updateChargeStatus(charge, AWAITING_CAPTURE_REQUEST);
-
-        } else {
+        if (!charge.isDelayedCapture())
```
jankowiakmaria, line 87: Would read better without exclamation mark, but functionally it's correct.
Would be better to have brackets around the if body, but can be addressed separately.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
```diff
@@ -494,12 +494,13 @@ public void markChargeAsEligibleForCapture_shouldThrowException_WithFeatureFlagE
                 mockConfiguration);
 
         String externalId = "external-id";
-        ChargeEntity charge = createNewChargeWith("worldpay", 1L, EXPIRED, "gatewayTxId");
+        ChargeEntity charge = createNewChargeWith("worldpay", 1L, AUTHORISATION_SUCCESS, "gatewayTxId");
```
jankowiakmaria, line 497: Not sure why this change is required.

jankowiakmaria (commented):
One minor comment, one doubt comment. Other than that LGTM.

sfount (approved):

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
kbottla: This was not right before. Expected is WebApplicationException but thrown due to charge being in invalid state `EXPIRED` but not when not able to add to queue

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
kbottla: 👍 I will update it in the next

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
jankowiakmaria: 👍 Thanks for explanation.

PR #1373, opened 2019-06-03
BAU: handle AmazonServiceExveption when deleting the capture message

jankowiakmaria:
Handle AmazonServiceExveption when deferring the capture message (when the message cannot be delated it currently return `Bad request 400` - we want to get more detailed information on what's causing this behaviour)


sfount (approved):
Code LGTM 👍 should help a lot with edge cases from SQS

sfount:
These might make more sense being logged at a warning level in the future but this will likely be standardised as the pattern is used across more services.

PR #1380, opened 2019-06-04
PP-5249 Use event queue to emit PaymentCreated event

mrlumbu:
Use the event queue to emit PaymentCreated event on payment creation.
The event is emitted after the new payment is committed to the database.
This ensures that an event is emitted if and only if the payment is 
successfully created. If the event emission fails the whole request 
will fail, and the payment journey will be aborted.


sfount (approved):
LGTM 👍 
* great integration tests verifying the message should be fired
* nice that it doesn't change the interface for creating a charge

sfount:
* minor Codacy complaints in `ChargeService`

PR #1381, opened 2019-06-04
BAU: Include `CAPTURE_SUBMITTED` in capture handler

sfount:
Connector is reporting errors on processing any message in the `CAPTURE_SUBMITTED` state -- this state should be considered final according to the capture process. 

* Add `CAPTURE_SUBMITTED` state to success handler
* Log any illegal state transitions not following the happy path

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureMessageProcess.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureMessageProcess.java
@@ -90,12 +90,19 @@ public class CardCaptureMessageProcess {
     }
 
     private void handleCapturedInvalidTransition(ChargeCaptureMessage captureMessage, IllegalStateRuntimeException e) throws QueueException {
-        if (chargeService.isChargeCaptured(captureMessage.getChargeId())) {
-            LOGGER.info("Charge capture message [{}] already captured - marking as processed.", captureMessage.getChargeId());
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureMessageProcess.java
```diff
@@ -90,12 +90,19 @@ private void handleCaptureRetry(ChargeCaptureMessage captureMessage) throws Queu
...
+        if (chargeService.isChargeCaptureSuccess(captureMessage.getChargeId())) {
+            LOGGER.info(
+                    "Charge capture message [{}] already captured - marking as processed. [chargeId={}]",
+                    captureMessage.getQueueMessageId(),
+                    captureMessage.getChargeId());
             captureQueue.markMessageAsProcessed(captureMessage);
             return;
         }
 
+        LOGGER.info(
```
mrlumbu, line 102: LOGGER.warn ?

mrlumbu (approved):

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureMessageProcess.java
sfount: Decided to leave warning/ error further down the stack and just log here for message processing context

PR #1383, opened 2019-06-05
PP-5292 - Modifies populateResponseBuilderWith(...) to not generate tokens and next_url/next_url_post for search results

nimalank7:
## WHAT YOU DID
This PR modifies the populateResponseBuilderWith(...) method in ChargeService so that when we are generating search results it does not generate a token nor displays next_url or next_url_post however it continues to generate a token for finding a charge. This is done by adding a boolean buildForSearchResult parameter in populateResponseBuilderWith(...)

It adds a needsNextUrl method in ChargeService which takes a ChargeEntity as well as a boolean value buildForSearchResult. This method is then called within populateResponseBuilderWith(...)

The buildResponse method in ChargeSearchStrategy has buildForSearchResult set to true. All other calls to the populateResponseBuilderWith(...) have the buildForSearchResult set to false.

The whenTheChargeCreatedDateIsBetweenTheFromDateAndTheToDate_shouldReturnTheCharge() test in SearchChargesByDateResourceIT has the matcher changed to 2 for results[0].links.size() to reflect that next_url and next_url_post is no longer exposed in the search results.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -288,7 +288,7 @@ public class ChargeService {
 
     private boolean needsNextUrl(ChargeEntity chargeEntity, boolean buildForSearchResult) {
         ChargeStatus chargeStatus = ChargeStatus.fromString(chargeEntity.getStatus());
-        return !buildForSearchResult && !chargeStatus.toExternal().isFinished() && !chargeStatus.equals(ChargeStatus.AWAITING_CAPTURE_REQUEST);
+        return !buildForSearchResult && !chargeStatus.toExternal().isFinished() && !chargeStatus.equals(AWAITING_CAPTURE_REQUEST);
     }
 
     @Transactional
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -124,7 +124,7 @@ public class ChargeService {
         return createCharge(chargeRequest, accountId, uriInfo)
                 .map(charge -> {
                     emitCreationEvent(charge);
-                    return populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, charge).build();
+                    return populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, charge, false).build();
                 });
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -124,7 +124,7 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
         return createCharge(chargeRequest, accountId, uriInfo)
                 .map(charge -> {
                     emitCreationEvent(charge);
-                    return populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, charge).build();
+                    return populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, charge, false).build();
```
kbottla, line 127: Don't we want to generate tokens when creating charges ?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -193,7 +193,7 @@ public void abortCharge(ChargeEntity charge) {
     public Optional<ChargeResponse> findChargeForAccount(String chargeId, Long accountId, UriInfo uriInfo) {
         return chargeDao
                 .findByExternalIdAndGatewayAccount(chargeId, accountId)
-                .map(chargeEntity -> populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, chargeEntity).build());
+                .map(chargeEntity -> populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, chargeEntity, false).build());
```
kbottla, line 196: Same as above.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: We do - the false flag is for whether the populateResponseBuilder is handling a search request or finding a charge. So if it's false it means it's not a search request, if it's true then it is a search request.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: See my above comment.

kbottla (approved):

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: 👍

PR #1386, opened 2019-06-06
PP-5326 Make zap test optional on master

mrlumbu:


sfount (commented):
Should `RUN_ZAP_ON_PR` parameter be updated to reflect the run zap at all change?

mrlumbu:
_shakes fist_

sfount (approved):
PP-5326 confirms temporarily disabling zap tests on master -- follow up on scheduling tests/ updating configuration might be needed.

PR #1390, opened 2019-06-07
PP-5249 Change event schema

mrlumbu:
We have agreed that an event should only have event metadata
as top level attributes, and any payload should be under
`event_details`.
As part of this I have removed the 'payment_id' from
the event, as this is contained in the external_resource_id.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/PaymentCreatedEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/PaymentCreatedEventDetails.java
@@ -1,8 +1,8 @@
-package uk.gov.pay.connector.events.eventpayload;
+package uk.gov.pay.connector.events.eventdetails;
 
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/events/eventdetails/PaymentCreatedEventDetails.java
```diff
@@ -0,0 +1,49 @@
...
+
+public class PaymentCreatedEventDetails extends EventDetails {
+    private final Long amount;
+    private final String description;
+    private final String reference;
+    private final String returnUrl;
+    private final Long gatewayAccountId;
+    private final  String paymentProvider;
+
+    private PaymentCreatedEventDetails(ChargeEntity charge) {
```
heathd, line 13: (minor/non blocking) - wouldn't it be nicer to have a constructor that took the separate args rather than requiring building from an entity?

heathd (approved):

heathd:
looks good, one minor non blocking comment .. we could address that if we need to in the future

PR #1404, opened 2019-06-17
Run e2e inline

heathd:
Run end-to-end tests inline for connector builds

Pros:
- Test reports for e2e tests run on PRs are associated with the PR build
- test reports for connector master builds are associated with the master build

Cons:
- I can't think of any

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -83,6 +83,7 @@ pipeline {
                   '''|#!/bin/bash
                      |set -e
                      |bundle install --path gems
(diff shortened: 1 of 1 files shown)


whpearson:
Is this worth putting into the jenkins library so it is easier to reuse?

heathd:
@whpearson I'd rather prove it works reliably before abstracting it

alexbishop1 (approved):
Let’s give this a try

heathd:
thanks!

PR #1400, opened 2019-06-14
PP-5321 Old capture process clean up

jankowiakmaria:
* removed old capture process class (and tests related to it)
* removed old background task
* renamed `CardCaptureMessageProcess` to `CardCaptureProcess` (which messes up github diff)
* removed unused methods in `ChargeDao`
* removed unused config values
* updated README.md (remove env variables, update description of capture process)
* removed feature flag from `CardCaptureProcess`

PR #1405, opened 2019-06-17
PP-5321: old capture process clean up

jankowiakmaria:
* removed old capture process class (and tests related to it)
* removed old background task
* renamed `CardCaptureMessageProcess` to `CardCaptureProcess` (which messes up github diff)
* removed unused methods in `ChargeDao`
* removed unused config values
* updated README.md (remove env variables, update description of capture process)
* removed feature flag from `CardCaptureProcess`

Replacement PR for https://github.com/alphagov/pay-connector/pull/1400


sfount (approved):
Approved following https://github.com/alphagov/pay-connector/pull/1400

It would be good to follow this through deploy closely just to double check everything is as expected

PR #1411, opened 2019-06-18
PP-5321: renamed config variable

jankowiakmaria:
Renamed environment variable used in the gauge (updated the way it's been used as well: `Duration` -> `int`)
Updated `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ The following variables control the background process:
 | Variable | Default | Purpose |
 |---------|---------|---------|
 | `BACKGROUND_PROCESSING_ENABLED` | `true` | enables registering scheduled processes - at the moment it includes only queue based capture methods |
-| `CAPTURE_PROCESS_RETRY_FAILURES_EVERY` | `60 minutes` | this value is used for calculating the metric gauge of messages awaiting capture (not attempted within this interval) |
+| `CHARGES_CONSIDERED_OVERDUE_FOR_CAPTURE_AFTER` | `60 minutes` | this value is used for calculating the metric gauge of messages awaiting capture (not attempted within this interval) |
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -54,7 +54,7 @@ The following variables control the background process:
 | Variable | Default | Purpose |
 |---------|---------|---------|
 | `BACKGROUND_PROCESSING_ENABLED` | `true` | enables registering scheduled processes - at the moment it includes only queue based capture methods |
-| `CAPTURE_PROCESS_RETRY_FAILURES_EVERY` | `60 minutes` | this value is used for calculating the metric gauge of messages awaiting capture (not attempted within this interval) |
```
heathd, line 57: if we're not already committed to this naming, how about something like:

`CHARGES_CONSIDERED_OVERDUE_FOR_CAPTURE_AFTER`

?

heathd: please don't change this if it's a lot of work/has interdependencies elsewhere

src/main/resources/config/config.yaml
```diff
@@ -93,15 +93,17 @@ executorServiceConfig:
 captureProcessConfig:
   backgroundProcessingEnabled: ${BACKGROUND_PROCESSING_ENABLED:-true}
 
+  retryFailuresEvery: ${CAPTURE_PROCESS_RETRY_FAILURES_EVERY:-60 minutes}
+  chargesAwaitingCaptureNotProcessedWithinMinutes: ${CHARGES_CONSIDERED_OVERDUE_FOR_CAPTURE_AFTER:-60}
```
heathd: we should prob make this consistent

jankowiakmaria: Done

kbottla (approved):
LGTM!

jankowiakmaria:
Thanks!

PR #1413, opened 2019-06-18
BAU remove generics from events

heathd:
when you have a composite object ie. A has-a B, then if you also
use a generic to parameterise the type of B then you end up in a world of
pain.. ie A<T extends B> overspecifies the type of B that A has..

you don’t need to do that. You can just allow A to have a B and the B can be
any subclass of B. The consumers of A don't care what kind of B the B is.


mrlumbu (approved):
👍

PR #1430, opened 2019-06-25
RED-116: Use tini in docker to run java

rhowe-gds:
By having bash as PID 1, we have a problem. Signals sent to the container (e.g.
SIGTERM to shut it down) will not reach Java - bash will not pass them on. This
means that shutting the container down is done uncleanly. Once the SIGTERM is
ignored, the container orchestrator will then send a SIGKILL and forcibly exit
the process.

We could make java PID 1 instead but this creates further problems - If Java
creates any child processes and those processes (or any of their children) die,
Java will be the parent of them and will not know to waitpid() on them, so they
will remain zombies forever. This is not a problem for connector as no such child
processes should be created, but it's still not really correct.

Also, when Java receives SIGTERM, it exits with an exit code of 15, even if
everything was shut down correctly. This will be mapped to exit code 143 by
Docker. tini can map this to zero to indicate that shutdown was clean.

tini solves all of these problems - it will correctly reparent orphaned child
processes and can remap the strange exit code of Java to something more
palatable.

In order to stop bash masking signals from Java, run java with 'exec' so that
no subprocess is spawned. We no longer need the `exit 0` in `docker-startup.sh`
as tini will deal with sorting out our exit status.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,3 @@
-# adoptopenjdk/openjdk11:jre-11.0.3_7-alpine
 FROM adoptopenjdk/openjdk11@sha256:eaa182283f19d3f0ee0c6217d29e299bb4056d379244ce957e30dcdc9e278e1e
 
 RUN ["apk", "--no-cache", "upgrade"]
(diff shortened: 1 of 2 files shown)


danworth (approved):
Looks good 👍

Confirmed that the new changes see tini run as pid1 with java app as a child. Also confirmed that sending `kill -15 1` results in java existing with a 143 (SIGTERM) which tini maps to a 0 due to the `-e 143` flag.
```
gds5062:~ danworth$ docker run --rm --env-file connectorconfig.tmp govukpay/connector:local
```
another terminal
```
gds5062:pay-connector danworth$ docker exec hungry_pascal kill -15 1
```
then back to the original terminal...
```
gds5062:~ danworth$ echo $?
0
```
Also confirmed that removing the `-e 143 --` flag from the tini entrypoint results in seeing the java processes 143 exit status directly...
```
gds5062:~ danworth$ echo $?
143
```
Hopefully this will resolve this ticket where charge's were left inconsistent state when Connector exits ungracefully during its capture process.
https://payments-platform.atlassian.net/browse/PP-5363

PR #1431, opened 2019-06-25
PP-5317: emitting PaymentDetailsEnteredEvent

jankowiakmaria:
* add `PaymentDetailsEnteredEvent` (tests covering producing the JSON)
* refactor `ChargeService` with proxy methods
* update definition of `emitEvent` in `ChargeService`

Questions/Doubts:
* I've refactored how wallet payment is being updated - instead of first calling `updateChargePostAuthorisation` and then setting email and wallet type (in Transactional context) it passes optional email and wallet and the `updateChargePostAuthorisation` updates the values if they exist. It caused one of the test to fail (the setup was putting an email address in the charge and the test was authorising with null value) - Is it deliberate behaviour? Is there a scenario where we have the email/wallet type already in the charge and we want to update them to `null` values?
* the other thing is: the timestamp is supposed to be taken from db (not part of this PR) when the `ChargeEventEntity` is inserted - the current logic retrieves all the charge events and gets the latest timestamp - the potential risk is that there might be another `ChargeEventEntity` inserted in the meantime.

Comments/Thoughts more than welcome.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -317,7 +317,7 @@ public class ChargeService {
         return chargeEntity;
     }
 
-    public ChargeEntity updateChargeAndEmitEventPostAuthorisation(String chargeExternalId,
+    private ChargeEntity updateChargeAndEmitEventPostAuthorisation(String chargeExternalId,
                                                                   ChargeStatus status,
                                                                   AuthCardDetails authCardDetails,
                                                                   Optional<String> transactionId,
@@ -333,6 +333,7 @@ public class ChargeService {
         return chargeEntity;
     }
 
+    // cannot be private: Guice requires @Transactional methods to be public
     @Transactional
     public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
                                                       ChargeStatus status,
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -38,6 +38,8 @@ import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeEx
 import uk.gov.pay.connector.common.model.api.ExternalChargeState;
 import uk.gov.pay.connector.common.model.api.ExternalTransactionState;
 import uk.gov.pay.connector.common.service.PatchRequestBuilder;
+import uk.gov.pay.connector.events.PaymentDetailsEnteredEvent;
+import uk.gov.pay.connector.events.Event;
 import uk.gov.pay.connector.events.EventQueue;
 import uk.gov.pay.connector.events.PaymentCreatedEvent;
 import uk.gov.pay.connector.gateway.PaymentProviders;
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -120,11 +120,15 @@ private void logMetrics(ChargeEntity chargeEntity,
...
-                .ifPresent(charge::setGatewayTransactionId);
-        return charge;
+        try {
+            ChargeEntity charge = chargeService.lockChargeForProcessing(chargeId, OperationType.AUTHORISATION);
+            getPaymentProviderFor(charge)
+                    .generateTransactionId()
+                    .ifPresent(charge::setGatewayTransactionId);
+            return charge;
+        } catch (Exception e) {
+            throw e;
```
mrlumbu: not sure why this has been added?

jankowiakmaria: debugging leftover - thanks for catching it.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -291,19 +293,63 @@ private boolean needsNextUrl(ChargeEntity chargeEntity, boolean buildForSearchRe
...
+                                                            AuthCardDetails authCardDetails,
+                                                            WalletType walletType,
+                                                            String emailAddress) {
+        ChargeEntity chargeEntity = updateChargeAndEmitEventPostAuthorisation(chargeExternalId, status, authCardDetails, transactionId, Optional.empty(), sessionIdentifier,
+                Optional.ofNullable(walletType), Optional.ofNullable(emailAddress));
+
+        return chargeEntity;
+    }
+
+    public ChargeEntity updateChargeAndEmitEventPostAuthorisation(String chargeExternalId,
```
mrlumbu: can be private, and I think this would make it clearer what the interface is given all these proxy methods

src/main/java/uk/gov/pay/connector/events/PaymentDetailsEnteredEvent.java
```diff
@@ -0,0 +1,74 @@
...
+        return resourceExternalId.equals(that.resourceExternalId) &&
+                eventDetails.equals(that.eventDetails) &&
+                timestamp.equals(that.timestamp);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(resourceExternalId, eventDetails, timestamp);
+    }
+}
```
mrlumbu, line 77: suggestion for future: we should be able to push down more of the object construction to base class

src/main/java/uk/gov/pay/connector/events/PaymentDetailsEnteredEvent.java
```diff
@@ -0,0 +1,74 @@
...
+        this.resourceExternalId = resourceExternalId;
+        this.eventDetails = eventDetails;
+        this.timestamp = timestamp;
+    }
+
+    public static PaymentDetailsEnteredEvent from(ChargeEntity charge) {
+        ZonedDateTime lastEventDate = charge.getEvents().stream()
+                .map(ChargeEventEntity::getUpdated)
+                .max(ZonedDateTime::compareTo)
+                .orElse(null);
```
mrlumbu: think this should be an exception. We don't want to publish events with no timestamp, and there is no reason why there would be no events for a charge at this point in lifecycle

mrlumbu (changes requested):

mrlumbu (approved):
👍 Looks good. As discussed, I think this PR highlights some areas where code could be improved, but that is beyond scope of this work

PR #1432, opened 2019-06-25
PP-5271 Add logging around discrepancy resolution failure

mrlumbu:
Make it easier to figure out why some cancellations aren't working,
and generally give more visibility when resolving discrepancies.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
@@ -220,9 +220,10 @@ public class ChargeExpiryService {
         }).orElse(EXPIRE_FLOW.getFailureTerminalState());
     }
     
-    public GatewayResponse<BaseCancelResponse> forceCancelWithGateway(ChargeEntity charge) {
+    public Boolean forceCancelWithGateway(ChargeEntity charge) {
         try {
-            return doGatewayCancel(charge);
+            GatewayResponse<BaseCancelResponse> cancelResponse = doGatewayCancel(charge);
+            return handleCancelResponse(charge.getExternalId(), cancelResponse);
         } catch (GatewayException e) {
             throw new WebApplicationException(String.format(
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -220,17 +220,29 @@ private ChargeStatus determineTerminalState(GatewayResponse<BaseCancelResponse>
...
         } catch (GatewayException e) {
             throw new WebApplicationException(String.format(
                     "Unable to cancel charge %s with gateway: %s",
                     charge.getExternalId(),
                     e.getMessage()));
         }
     }
-    
+
+    private Boolean handleCancelResponse(String chargeExternalId, GatewayResponse<BaseCancelResponse> cancelResponse) {
```
jankowiakmaria, line 235: [minor] It could use (slightly) better naming - ie `hasCancellationSucceeded` or something (more along the standard naming). `handle` is a bit generic for me and in this case the method determines whether something has been successful or not.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -220,17 +220,29 @@ private ChargeStatus determineTerminalState(GatewayResponse<BaseCancelResponse>
         }).orElse(EXPIRE_FLOW.getFailureTerminalState());
     }
     
-    public GatewayResponse<BaseCancelResponse> forceCancelWithGateway(ChargeEntity charge) {
+    public Boolean forceCancelWithGateway(ChargeEntity charge) {
```
jankowiakmaria, line 223: are there unit tests covering this change?

jankowiakmaria (commented):
There is a minor comment and a question.

jankowiakmaria (approved):
Approving, although I expect follow-up PR that addresses the "naming of the method" comment.
Thanks!

PR #1435, opened 2019-06-26
PP-5367: Avoid adding custom CA certificates twice

rhowe-gds:
This needs a proper description


rhowe-gds:
Obviated by https://github.com/alphagov/pay-connector/pull/1732

PR #1436, opened 2019-06-26
PP-5349 add sqs queue healtcheck

SandorArpa:
## WHAT YOU DID
- add a healthcheck that reports whether the given sqs queues are healthy

with @nimalank7


src/main/java/uk/gov/pay/connector/healthcheck/SQSHealthCheck.java
```diff
@@ -0,0 +1,73 @@
...
+        for (SqsHealthCheckQueueHolder checkValues : checkList) {
+            Optional<String> result = checkQueue(checkValues.queueUrl);
+            if (result.isPresent()) {
+                return Result.unhealthy(format("failed to retrieve %s queue attributes: %s",checkValues.queueIdentifier, result.get()));
+            }
+        }
+        
+        return Result.healthy();
+    }
+
```
mrlumbu, line 46: A slightly more streamy solution would be 
```
        List<String> queueChecks = checkList.stream()
                .map(queue -> this.checkQueue(queue.queueUrl))
                .flatMap(Optional::stream)
                .collect(Collectors.toUnmodifiableList());
        
        if (queueChecks.isEmpty()) {
            return Result.healthy();
        } else {
            String errorSummary = String.join(",", queueChecks);
            return Result.unhealthy(format("Queue healthcheck failed: %s", errorSummary));
        }
```
which has advantage of returning result for all queues (although doesn't include the identifier as is). Arguably doesn't gain much though

src/main/java/uk/gov/pay/connector/healthcheck/SQSHealthCheck.java
```diff
@@ -0,0 +1,73 @@
...
+        GetQueueAttributesRequest queueAttributesRequest =
+                new GetQueueAttributesRequest(queueUrl)
+                        .withAttributeNames("All");
+        try {
+            sqsClient.getQueueAttributes(queueAttributesRequest);
+        } catch (AmazonSQSException | UnsupportedOperationException e) {
+            logger.error("Failed to retrieve queue attributes - {}", e.getMessage());
+            return Optional.of(e.getMessage());
+        } catch (SdkClientException e) {
+            logger.error("Failed to connect to sqs queue - {}", e.getMessage());
```
mrlumbu: you need to return an optional of something here, otherwise it falls through to returning `Optional.empty` below, and therefore returns a false positive

src/test/java/uk/gov/pay/connector/it/resources/HealthCheckResourceIT.java
```diff
@@ -14,20 +14,22 @@
...
                 .body("database.healthy", is(true))
                 .body("deadlocks.healthy", is(true))
-                .body("cardExecutorService.healthy", is(true));
+                .body("cardExecutorService.healthy", is(true))
+                .body("sqsQueue.healthy", is(false));
     }
 
     @Test
-    public void checkHealthcheck_isUnhealthy() {
+    public void checkHealthcheck_isUnhealthy_LessSqsQueue() {
```
mrlumbu: let's just keep this test and bin the others. And just call it something like `healthcheckShouldIdentifyUnhealthyDBAndQueue`

src/test/java/uk/gov/pay/connector/resources/HealthCheckResourceTest.java
```diff
@@ -50,18 +50,20 @@ public void checkHealthCheck_isUnHealthy() throws JsonProcessingException {
         registry.put("database", HealthCheck.Result.unhealthy("database in unavailable"));
```
mrlumbu: This whole file is pointless. It would be more honest if it just did
```
        registry.put("check1", HealthCheck.Result.unhealthy("unhealthy"));
        registry.put("check2", HealthCheck.Result.healthy());
```
then checked it got the right thing for check1 and check2, but even that feels a bit OTT given we have an integration test. So I suggest either change to something like ☝️ or delete this file

mrlumbu (changes requested):
One change that is needed for check to work - correctly handling `SdkClientException`, and some other suggestions around testing, and a possibly unhelpful suggestion for a more streamy approach to healthcheck processing.

mrlumbu (approved):
👍

PR #1437, opened 2019-06-26
Pp 5271 ePDQ cancel improvements

mrlumbu:
The purpose of this PR is to enable us to cancel payments with ePDQ using only the reference that pay generated for the payment (i.e. the `externalId` of a charge), rather than the reference generated by ePDQ. The actual change that allows this is contained in ea2a7f8

As part of this work I have also made some steps to 'degenerify' the communication with ePDQ. This is in 
4312398
I think this is pretty uncontroversial - I have unhooked ePDQ payload building from the hyper generic code that was running it before. It s now more flexible and more readable, and about the same number of lines. I could have done much more, but will leave it there for this PR.

b6a629a refactors the way in which we ensured deterministic ordering for ePDQ payloads. Should be fairly self explanatory. This actually revealed that despite comments all over the place, the payload for some calls was not ordered alphabetically by key, so I had to make some small changes to test assertions.


SandorArpa (approved):
LOKTM

PR #1440, opened 2019-06-27
BAU/PP-5317: update date of a charge event set by db

jankowiakmaria:
Updated the code not to set the updated for the charge event


jankowiakmaria:
It'll be superseded by a piece of work made by Dom and Steven

PR #1442, opened 2019-06-28
Pp 5356 formalise state transition events

heathd:
In this PR we have modified the `PaymentGatewayStateTransitions` state machine to annotate each arc with the event expected for that transition.

This has enabled us to enumerate the following event types:

- PaymentCreatedEvent - created (payload)
- PaymentStartedEvent - user visited payment page (no payload)
- PaymentDetailsEnteredEvent - details entered (payload)

Relating to authorisation, none of these events have a payload:

- GatewayRequires3dsAuthorisation
- AuthorisationRejected
- AuthorisationCancelledEvent
- AuthorisationSucceededEvent
- GatewayTimeoutDuringAuthorisation
- GatewayErrorDuringAuthorisation
- UnexpectedGatewayErrorDuringAuthorisation

Relating to capture or user confirmation:
- UserApprovedForCapture
- UserApprovedForCaptureAwaitingServiceApproval
- ServiceApprovedForCapture
- CaptureSubmittedEvent
- CaptureConfirmedEvent
- CaptureAbandonedAfterTooManyRetriesEvent
- CaptureErrorEvent

Expiration:
- PaymentExpiredEvent

Cancellation:
- UserCancelledEvent
- SystemCancelledEvent

We have also generated a state graph with the events annotated:

states dot

## Feedback sought

There's inconsistency between naming events as `SomethingHappened` or `SomethingHappenedEvent`. My personal bias would be towards `SomethingHappened` since the fact that it's an event should be evident from context, package name, class hierarchy etc.

Supplemental point to this is the naming of the `getEventType()` which I think should be kept strictly in sync with the class name. At present we have named events as follows:
- `PAYMENT_DETAILS_EVENT`
- `PAYMENT_CREATED`

I've renamed the `PaymentCreatedEvent` class to `PaymentCreated` to make that consistent. I would propose renaming `PaymentDetailsEvent` to `PaymentDetailsEntered` so that we'd have:

- `PAYMENT_CREATED`
- `PAYMENT_DETAILS_ENTERED`


src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
```diff
@@ -150,15 +170,63 @@ private PaymentGatewayStateTransitions() {
...
+                .map(modelledEvent ->
+                        (event instanceof UnspecifiedEvent) || modelledEvent.permits(event)
+                )
+                .orElse(false);
+    }
+
+    private static abstract class ModelledEvent {
+        public abstract boolean permits(Event event);
+
+        public static <T extends Event> ModelledEvent of(Class<T> clazz) {
```
mrlumbu, line 192: can just be 
```
public static ModelledEvent of(Class<? extends Event> clazz) {
```

mrlumbu: And perhaps there is a better name than `clazz`? Even `eventClass` would feel more informative and less ugly

src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
```diff
@@ -52,90 +72,90 @@ public static PaymentGatewayStateTransitions getInstance() {
...
+        graph.putEdgeValue(AUTHORISATION_3DS_READY, AUTHORISATION_REJECTED, ModelledEvent.of(AuthorisationRejected.class));
+        graph.putEdgeValue(AUTHORISATION_3DS_READY, AUTHORISATION_ERROR, ModelledEvent.of(GatewayErrorDuringAuthorisation.class));
+        graph.putEdgeValue(AUTHORISATION_3DS_READY, EXPIRE_CANCEL_READY, ModelledEvent.none());
+        graph.putEdgeValue(AUTHORISATION_3DS_READY, AUTHORISATION_CANCELLED, ModelledEvent.of(AuthorisationCancelled.class));
+        graph.putEdgeValue(AUTHORISATION_SUCCESS, CAPTURE_APPROVED, ModelledEvent.of(UserApprovedForCapture.class));
+        graph.putEdgeValue(AUTHORISATION_SUCCESS, CAPTURE_READY, ModelledEvent.none());
+        graph.putEdgeValue(AUTHORISATION_SUCCESS, SYSTEM_CANCEL_READY, ModelledEvent.none());
+        graph.putEdgeValue(AUTHORISATION_SUCCESS, USER_CANCEL_READY, ModelledEvent.none());
+        graph.putEdgeValue(AUTHORISATION_SUCCESS, EXPIRE_CANCEL_READY, ModelledEvent.none());
+        graph.putEdgeValue(AUTHORISATION_SUCCESS, AWAITING_CAPTURE_REQUEST, ModelledEvent.of(UserApprovedForCaptureAwaitingServiceApproval.class));
```
mrlumbu, line 128: this doesn't sound very eventy - events probably should not carry informtion about what needs to come after them. No that I have a better suggestion right now

mrlumbu (commented):
Made a couple of slightly pedantic comments. Looks great though!

mrlumbu (approved):

PR #1443, opened 2019-06-28
PP-5355 Example queue message contract test

mrlumbu:
Illustrates how to run provider side message contract tests. I have had to upgrade the pact junit library to be a bit bleeding edge, but everything seems to work so probably fine.

The basic idea is that a method annotated with `@PactVerifyProvider` is matched with a corresponding pact based on the description (which acts as a key). That method is then required to produce a message corresponding to that description, which is then verified using the pact (and the test target). It is a bit magic. To understand better how it all fits together it is worth looking at 
https://github.com/DiUS/pact-jvm/blob/master/provider/pact-jvm-provider-junit/src/main/kotlin/au/com/dius/pact/provider/junit/target/AmqpTarget.kt
https://github.com/DiUS/pact-jvm/blob/master/provider/pact-jvm-provider-junit/src/main/kotlin/au/com/dius/pact/provider/junit/PactRunner.kt

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -361,9 +361,15 @@
         </dependency>
         <dependency>
             <groupId>au.com.dius</groupId>
-            <artifactId>pact-jvm-provider-junit_2.12</artifactId>
-            <version>3.6.11</version>
+            <artifactId>pact-jvm-provider-junit</artifactId>
+            <version>4.0.0-beta.2</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>xerces</groupId>
+                    <artifactId>xercesImpl</artifactId>
+                </exclusion>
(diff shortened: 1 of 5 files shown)


jankowiakmaria (approved):

sfount (approved):
Looks great 👍 ! The producer side of this looks really simple to assert.

Two questions:
* is it possible to explain some of the underlying "glue" provided by Pact annotations? Just so that if someone comes to copy this method they will have an understanding
* is there a known timeline for when these tools will be available out of beta? That might manage the unusual dependency on a beta library

mrlumbu:
Not sure about timelines. Development is pretty active https://github.com/DiUS/pact-jvm/releases, so I would guess a month or two max for a stable version.

PR #1456, opened 2019-07-04
PP-5385 Poll and emit payment events to `eventQueue`

sfount:
* State transition event emitter 

Depends on `PaymentEvent` builder restructure -- https://github.com/alphagov/pay-connector/pull/1458
Now uses uniform factory to build payment events

Depends on (rebased, should be merged first): 
* https://github.com/alphagov/pay-connector/pull/1453
* https://github.com/alphagov/pay-connector/pull/1458

with @mrlumbu

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/PaymentStateTransitionEmitterProcess.java
+++ b/src/main/java/uk/gov/pay/connector/events/PaymentStateTransitionEmitterProcess.java
@@ -6,7 +6,6 @@ import uk.gov.pay.connector.chargeevent.dao.ChargeEventDao;
 import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
 import uk.gov.pay.connector.queue.PaymentStateTransition;
 import uk.gov.pay.connector.queue.PaymentStateTransitionQueue;
-import uk.gov.pay.connector.queue.QueueException;
 
 import javax.inject.Inject;
 import java.util.Optional;
@@ -30,14 +29,8 @@ public class PaymentStateTransitionEmitterProcess {
     }
 
     public void handleStateTransitionMessages() {
-        // poll queue for latest message
-        // try and emit event for message
-        // IF successful - exit
-        // IF failed - add message back to queue
         Optional.ofNullable(paymentStateTransitionQueue.poll())
                 .ifPresent(this::emitEvent);
-
-        LOGGER.info("Checking payment state transition queue for new transitions to process");
     }
 
     private void emitEvent(PaymentStateTransition paymentStateTransition) {
@@ -53,11 +46,11 @@ public class PaymentStateTransitionEmitterProcess {
 
     private PaymentEvent createEvent(PaymentStateTransition paymentStateTransition) throws Exception {
         return chargeEventDao.findById(ChargeEventEntity.class, paymentStateTransition.getChargeEventId())
-                .map(this::createEvent)
+                .map(chargeEvent -> createEvent(chargeEvent, paymentStateTransition.getStateTransitionEventClass()))
                 .orElseThrow(() -> new Exception());
     }
 
-    private PaymentEvent createEvent(ChargeEventEntity chargeEventEntity) {
+    private PaymentEvent createEvent(ChargeEventEntity chargeEvent, Class<? extends PaymentEvent> paymentEventClass) {
         return null;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/events/PaymentStateTransitionEmitterProcess.java
+++ b/src/main/java/uk/gov/pay/connector/events/PaymentStateTransitionEmitterProcess.java
@@ -36,23 +36,33 @@ public class PaymentStateTransitionEmitterProcess {
     }
 
     private void emitEvent(PaymentStateTransition paymentStateTransition) {
-        try {
-            PaymentEvent paymentEvent = createEvent(paymentStateTransition);
-            eventQueue.emitEvent(paymentEvent);
-        } catch (ChargeEventNotFoundRuntimeException | QueueException e) {
-            LOGGER.warn(
-                    "Failed to emit payment event for state transition [chargeEventId={}] [eventType={}]",
+        if (paymentStateTransition.shouldAttempt()) {
+            paymentStateTransition.markAttempt();
+
+            try {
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/events/PaymentStateTransitionEmitterProcess.java
```diff
@@ -0,0 +1,61 @@
...
+        try {
+            PaymentEvent paymentEvent = createEvent(paymentStateTransition);
+            eventQueue.emitEvent(paymentEvent);
+        } catch (ChargeEventNotFoundRuntimeException | QueueException e) {
+            LOGGER.warn(
+                    "Failed to emit payment event for state transition [chargeEventId={}] [eventType={}]",
+                    paymentStateTransition.getChargeEventId(),
+                    paymentStateTransition.getStateTransitionEventClass().getSimpleName()
+            );
+            paymentStateTransitionQueue.offer(paymentStateTransition);
```
sfount: If a `paymentStateTransition` can't be processed because something has externally failed this will continually try to consume this. 

We should thing about a limit to the number of times we process these `paymentStateTransition` messages.

src/main/java/uk/gov/pay/connector/events/PaymentStateTransitionEmitterProcess.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+
+    private PaymentEvent createEvent(PaymentStateTransition paymentStateTransition) throws ChargeEventNotFoundRuntimeException {
+        return chargeEventDao.findById(ChargeEventEntity.class, paymentStateTransition.getChargeEventId())
+                .map(chargeEvent -> createEvent(chargeEvent, paymentStateTransition.getStateTransitionEventClass()))
+                .orElseThrow(() -> new ChargeEventNotFoundRuntimeException("No external charge ID"));
```
sfount: This exception won't give a very useful message in stack traces -- we should use a more specific exception.

src/main/java/uk/gov/pay/connector/queue/PaymentStateTransition.java
```diff
@@ -1,16 +1,23 @@
...
-public class PaymentStateTransition implements Delayed {
+public class PaymentStateTransition<T extends PaymentEvent> implements Delayed {
     private final long chargeEventId;
+    private final Class<T> stateTransitionEventClass;
     private final Long readTime;
 
+    private static final int MAXIMUM_NUMBER_OF_ATTEMPTS = 10;
+
     private long delayDurationInMilliseconds = 100L;
+    private int attempts = 0;
```
mrlumbu: I think we either need to make this class immutable or make this field threadsafe (AtomicInteger). I think I prefer the immutable option.

src/main/java/uk/gov/pay/connector/events/exception/StateTransitionMessageProcessException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.pay.connector.events.exception;
+
+public class StateTransitionMessageProcessException extends Exception {
+    public StateTransitionMessageProcessException(Long chargeEventId) {
+        super(String.format("Failed to access charge event during state transition message processing [chargeEventId=%d]", chargeEventId));
+    }
```
mrlumbu, line 6: 👍

src/main/java/uk/gov/pay/connector/queue/PaymentStateTransition.java
sfount: Great idea! Have used the immutable option in https://github.com/alphagov/pay-connector/pull/1456/commits/913bd42e7d4e3514a59624569c4ad30580182cb5 -- this should work well across threads too

src/main/java/uk/gov/pay/connector/queue/PaymentStateTransition.java
```diff
@@ -1,19 +1,45 @@
...
+public final class PaymentStateTransition<T extends PaymentEvent> implements Delayed {
     private final long chargeEventId;
+    private final Class<T> stateTransitionEventClass;
     private final Long readTime;
+    private final long delayDurationInMilliseconds;
+    private final int attempts;
 
-    private long delayDurationInMilliseconds = 100L;
+    private static final int MAXIMUM_NUMBER_OF_ATTEMPTS = 10;
+    private static final int BASE_ATTEMPTS = 1;
```
mrlumbu, line 16: is there a reason this isn't 0?

mrlumbu: (and actually a numeric field is initialised to 0 by default, so probably no need for this)

src/main/java/uk/gov/pay/connector/queue/PaymentStateTransition.java
```diff
@@ -1,19 +1,45 @@
 package uk.gov.pay.connector.queue;
 
+import uk.gov.pay.connector.events.PaymentEvent;
+
 import java.util.concurrent.Delayed;
 import java.util.concurrent.TimeUnit;
 
-public class PaymentStateTransition implements Delayed {
+public final class PaymentStateTransition<T extends PaymentEvent> implements Delayed {
```
mrlumbu, line 8: looking at this class now it occurs to me that we have muddled up quite a few different concerns. We can try to unpick it a bit if we have the time

mrlumbu (commented):
Looks PaSTEE. Some minor comments that we can discuss

src/main/java/uk/gov/pay/connector/queue/PaymentStateTransition.java
sfount: Agreed - the constructors should definitely be restructured, slightly hodge podge as is -- will visit this in a future PR

src/main/java/uk/gov/pay/connector/queue/PaymentStateTransition.java
sfount: Agreed, separating the data of this object from the delay + retry logic would make this much clearer. Should we do that in this PR or the next?

mrlumbu (approved):

PR #1465, opened 2019-07-08
BAU fix docker cleanup for inline e2e tests

heathd:
Executing a shell script block uses the `sh` command, not `shell`


rjbaker (approved):
Good catch. Thanks @heathd 👍

PR #1466, opened 2019-07-08
PP-5386 Unify all updateChargeStatus calls

sfount:
* introduce new `transitionChargeState` call to unify before including emit event
* make sure everything that uses current special case branching logic
(update to capture submitted before capture)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -58,7 +58,6 @@ import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.core.UriInfo;
 import java.net.URI;
 import java.time.Duration;
-import java.time.ZonedDateTime;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -438,22 +442,16 @@ public ChargeEntity findChargeById(String chargeId) {
                 .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeId));
     }
 
-    public ChargeEntity updateChargeStatus(ChargeEntity chargeEntity, ChargeStatus chargeStatus) {
-        if (chargeStatus == CAPTURED) {
-            chargeEntity.setStatus(CAPTURE_SUBMITTED);
-            chargeEventDao.persistChargeEventOf(chargeEntity);
-            chargeEntity.setStatus(CAPTURED);
-            chargeEventDao.persistChargeEventOf(chargeEntity, ZonedDateTime.now());
```
mrlumbu, line 446: we have removed the line that passes through `now()` as the gateway event date. This is because ultimately that value is only used here https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java#L303, which uses the updated time if the gateway event time isn't set anyway. In future work we will look at how to set this gateway event date more elegantly

mrlumbu (approved):

PR #1474, opened 2019-07-09
PP-5409 Improve cancellation with gateways

stephencdaly:
Use same methodology as charge expiry to determine whether charges need cancelling with the gateway. This groups charge statuses into PRE_AUTHORISATION, DURING_AUTHORISATION and POST_AUTHORISATION. For charges with statuses in the DURING_AUTHORISATION category, we check with the gateway to determine whether we need to send a cancellation request.

Also add the ability to cancel charges in AUTHORISATION_READY and AUTHORISATION_3DS_READY statuses, as we should be able to cancel a charge at any point when it isn't in a terminal state.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
@@ -1,12 +1,11 @@
 package uk.gov.pay.connector.charge.service;
 
-import com.google.common.collect.ImmutableList;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
-import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
+import uk.gov.pay.connector.charge.model.domain.ExpirableChargeStatus;
 import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
 import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
 import uk.gov.pay.connector.gateway.GatewayException;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/QueryService.java
```diff
@@ -1,13 +1,21 @@
...
+import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.ChargeQueryResponse;
 
 import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
+
+import static com.amazonaws.xray.contexts.SegmentContext.logger;
```
mrlumbu: This is unused

src/main/java/uk/gov/pay/connector/charge/service/StatusFlow.java
```diff
@@ -26,34 +21,28 @@
...
+    private static final List<ChargeStatus> TERMINABLE_STATUSES = ExpirableChargeStatus.getValuesAsStream()
+            .map(ExpirableChargeStatus::getChargeStatus)
+            .collect(Collectors.toList());
+    
     public static final StatusFlow USER_CANCELLATION_FLOW = new StatusFlow("User Cancellation",
-            ImmutableList.of(
-                    ENTERING_CARD_DETAILS,
-                    AUTHORISATION_SUCCESS
-            ),
+            TERMINABLE_STATUSES,
```
mrlumbu, line 29: Good that we're making this consistent. As far as I can see there is no problem with allowing e.g. user cancel to happen from a larger set of statuses, and as we discussed in person, cancel should override any lock states etc. that are being used to ensure correct processing.

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -126,17 +134,10 @@ private void nonGatewayCancel(ChargeEntity chargeEntity, StatusFlow statusFlow)
         chargeService.transitionChargeState(chargeEntity.getExternalId(), completeStatus);
     }
 
-    private boolean gatewayIsNotAwareOfCharge(ChargeEntity chargeEntity) {
-        return nonGatewayStatuses.contains(ChargeStatus.fromString(chargeEntity.getStatus()));
-    }
-
     private void prepareForTerminate(ChargeEntity chargeEntity, StatusFlow statusFlow) {
-        ChargeStatus newStatus = statusFlow.getLockState();
+        ChargeStatus lockStatus = statusFlow.getLockState();
```
mrlumbu, line 138: smooth

mrlumbu (dismissed):
Looks good to me! Great that we're making this more consistent and comprehensible

mrlumbu (approved):

PR #1484, opened 2019-07-11
PP-5386 State transition process - Dropwizard lifecycle hook

sfount:
* feature flag for emitting state transition events 
* add state transition polling to dropwizard lifecycle 
* test state transition messages on either end of process (integration test)

with @mrlumbu


src/test/resources/config/test-it-config.yaml
```diff
@@ -8,7 +8,7 @@ server:
   registerDefaultExceptionMappers: false
 
 logging:
-  level: WARN
+  level: INFO
```
mrlumbu: maybe revert this back?

mrlumbu (approved):
Looks good 👍

PR #1485, opened 2019-07-11
PP-5386 Undefined charge status

mrlumbu:
Introduces a new undefined charge status, and on charge creation initialises a charge as undefined, then transitions it to created. This means a PaymentCreated event is emitted in the same way all other events are emitted.
The charge entity needs to be merged after transitioning to created, because for some reason that I haven't got totally to the bottom of, after the call to persist the charge entity is detached. I have chosen to do the merge as a one off thing, rather than move it into the transitionChargeStatus method, as it is seemingly only needed after entity creation.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -179,9 +179,8 @@ public class ChargeService {
                     .ifPresent(chargeEntity::setCardDetails);
 
             chargeDao.persist(chargeEntity);
-
-            chargeEventDao.persistChargeEventOf(chargeEntity);
-
+            transitionChargeState(chargeEntity, CREATED);
+            chargeDao.merge(chargeEntity);
             return chargeEntity;
         });
     }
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -179,9 +179,8 @@ private void emitEvent(Event event) {
                     .ifPresent(chargeEntity::setCardDetails);
 
             chargeDao.persist(chargeEntity);
-
-            chargeEventDao.persistChargeEventOf(chargeEntity);
-
+            transitionChargeState(chargeEntity, CREATED);
+            chargeDao.merge(chargeEntity);
```
sfount, line 183: Is it worth having a comment on this merge? It's not something we have to do anywhere else and would be good to not encourage copying this pattern.

sfount (approved):
Nice 👍. 

We'll be emitting two created and payment details entered events until https://github.com/alphagov/pay-connector/pull/1487/files is in -- this shouldn't be a problem as ledger can handle this case.

PR #1503, opened 2019-07-17
PP-5388: rename duplicated events in a given payment flow

sfount:
Some state transitions during the payment flow could result in the same event being emitted for different actions.

i.e 
`SYSTEM_CANCEL_READY` -> `SYSTEM_CANCEL_SUBMITTED` : `SystemCancelled`
`SYSTEM_CANCEL_SUBMITTED` -> `SYSTEM_CANCELLED` : `SystemCancelled` 

This event was reflecting the external state transitioned to rather than the action, update this to remove duplicate events during an individual payment flow.

EDIT: unfortunately file organisation makes it look like structural changes were made - only renaming happened in `PaymentGatewayStateTransitions`


mrlumbu (approved):
Nice, big fan of the new event names. Should probably tie this to an actual Jira ticket

sfount:
Relates to PP-5388

PR #1508, opened 2019-07-18
PP-5439 Administrative find charge by gateway transaction id

sfount:
* Stripe payments before 03 May 2019 store no meta data on Stripe regarding GOV.UK Pay charge information, the only link is the `gateway_transaction_id` on the charge table 
* expose internal resource to get a charge by this ID, allowing for intermediate payout reconciliation on older Stripe payouts


mrlumbu (changes requested):
one good way of seeing if it works would be to write a test

sfount:
@mrlumbu you may have been right https://github.com/alphagov/pay-connector/pull/1508/files/91c08a03ad4939b095812c90c4e7c8d22b456ddd..b66d42740875e37e954ded9bab67857d61aa5253

mrlumbu (approved):

PR #1515, opened 2019-07-19
PP-5391 add refund events model

SandorArpa:
## WHAT YOU DID

- introduce refund event classes
- refactor eventdetails package

We introduced parentResourceExternalId which is the charge's externalId. This is going to be the link on Ledger's side to the corresponding payment.

with @mrlumbu


mrlumbu (approved):

mrlumbu:
I'm confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1516, opened 2019-07-19
PP-5391 emit refund state transitions

mrlumbu:
## WHAT YOU DID
Emit refund state transition events by hooking into centralised `transitionRefundStatus` method. We re-use the approach taken for emitting payment state transitions, whereby the transitions is put on a queue and then read off and processed, to ensure we only emit events that completed successfully, and to be able to use DB based timestamp.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -46,7 +46,7 @@ import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 import uk.gov.pay.connector.queue.PaymentStateTransition;
-import uk.gov.pay.connector.queue.PaymentStateTransitionQueue;
+import uk.gov.pay.connector.queue.StateTransitionQueue;
 import uk.gov.pay.connector.token.dao.TokenDao;
 import uk.gov.pay.connector.token.model.domain.TokenEntity;
 import uk.gov.pay.connector.wallets.WalletType;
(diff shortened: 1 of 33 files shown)


src/main/java/uk/gov/pay/connector/refund/dao/RefundDao.java
```diff
@@ -72,6 +72,23 @@ public RefundDao(final Provider<EntityManager> entityManager) {
                         refund.get(STATUS).in(statuses));
         return entityManager.get().createQuery(criteriaQuery).getResultList();
     }
+    
+    public Optional<RefundHistory> getRefundHistoryByRefundExternalIdAndRefundStatus(String refundExternalId, RefundStatus refundStatus) {
+        String query = "SELECT rh.id, rh.external_id, rh.amount, rh.status, rh.charge_id, rh.created_date, " +
+                "rh.version, rh.reference, rh.history_start_date, rh.history_end_date, rh.user_external_id, " +
+                "rh.gateway_transaction_id, ch.external_id AS charge_external_id " +
+                "FROM refunds_history rh, charges ch " +
+                "WHERE rh.external_id = ?1 AND rh.status = ?2 AND rh.charge_id = ch.id";
```
sfount, line 81: Is there a reason this doesn't use string params set? 

`WHERE rh.external_id = :externalId` might read clearer for the query

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -430,8 +430,8 @@ public ChargeEntity transitionChargeState(
                 .ifPresent(eventType -> {
                     if (shouldEmitPaymentStateTransitionEvents) {
                         PaymentStateTransition transition = new PaymentStateTransition(chargeEventEntity.getId(), eventType);
-                        paymentStateTransitionQueue.offer(transition);
-                        logger.info("Offered payment state transition to emitter queue [from={}] [to={}] [chargeEventId={}] [chargeId={}]", fromChargeState, targetChargeState, chargeEventEntity.getId(), charge.getExternalId());
+                        stateTransitionQueue.offer(transition);
+                        logger.info("Offered payment state transition to emitter queue [from={}] [to={}] [chargeId={}]", fromChargeState, targetChargeState, chargeEventEntity.getId());
```
sfount: This ID is not the `chargeId`

```suggestion
                        logger.info("Offered payment state transition to emitter queue [from={}] [to={}] [chargeEventId={}] [chargeId={}]", fromChargeState, targetChargeState, chargeEventEntity.getId(), charge.getExternalId());
```

sfount (commented):
Looks good! `PaSTEE++`

src/main/java/uk/gov/pay/connector/refund/dao/RefundDao.java
SandorArpa: It is needed as this query is using `createNativeQuery` which does not understand string params, only positional ones. (This is a strange behaviour since although the code compiles it breaks when executed)

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: 👍

SandorArpa (approved):

PR #1526, opened 2019-07-23
PP-5447 introduce external capturable state

heathd:
to allow services to detect when a charge may be explictly captured


heathd:
re-running tests with e2e enabled

heathd:
rebased

mrlumbu (approved):

PR #1527, opened 2019-07-23
PP-5402 - Adds database migration which drops the not-null constraint on return…

nimalank7:
## What you did:
Because the payload for telephone payments do not have a return_url and since we are storing them in the charges table we have to make the return_url allow null values. This PR contains a database migration for this.


kbottla (approved):
LGTM!

PR #1528, opened 2019-07-23
BAU: changed serialization for gatewayEventDate

jankowiakmaria:
* serialisation to number causes pain when creating pact test for CaptureConfirmed event
* changing to ISO formatting will be more human readable (and shouldn't cause pain with pact test)

Unless someone has better idea how to resolve https://build.ci.pymnt.uk/job/pay-ledger/job/PR-148/13/console. (I've spent enough time - don't want to waste any more)


heathd:
👍  this is actually fixing a bug, and yay pacts for detecting it.

heathd (approved):

PR #1530, opened 2019-07-23
PP-5469: PaymentDetailsEntered emits totalAmount

jankowiakmaria:
* PaymentCreated doesn't have enough information to emit corporate_surcharge and final total_amount
* corporate_surcharge is calculate before authorisation therefore PaymentDetailsEntered should be able to emit total_amount (corporate_surcharge is already there)
* add pact for PaymentDetailsEntered event

* revert https://github.com/alphagov/pay-connector/commit/8921622cca305a1ca7d58c445c81f53c7f62b2c2 (to emit `PaymentDetailsEntered` event)

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
@@ -70,7 +70,6 @@ import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.isNull;
 import static org.mockito.Mockito.anyString;
 import static org.mockito.Mockito.doAnswer;
-import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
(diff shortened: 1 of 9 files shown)


kbottla (commented):
Changes looks good, but I wonder at what point `PaymentDetailsEntered` is going to be generated. At the moment `PaymentDetailsEntered.class` is not specified in any state transitions https://github.com/alphagov/pay-connector/blob/dcc87a4d13c5d7f130fae5f1f8e587b1c11988c8/src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java#L75

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -175,7 +174,7 @@ public void setUp() {
         when(mockedConfig.getEmitPaymentStateTransitionEvents()).thenReturn(true);
 
         service = new ChargeService(mockedTokenDao, mockedChargeDao, mockedChargeEventDao,
-                mockedCardTypeDao, mockedGatewayAccountDao, mockedConfig, mockedProviders, mockedStateTransitionQueue);
+                mockedCardTypeDao, mockedGatewayAccountDao, mockedConfig, mockedProviders, mockedStateTransitionQueue, mockedEventQueue);
```
kbottla, line 177: Might be useful to have service test to verify we are emitting payment details entered event. I see there are tests to assert the behaviour in `CardAuthoriseServiceTest` & `WalletAuthoriseServiceTest` but might be making test redundant in multiple places and may make it hard to trace the tests for `emitEvent` from ChargeService.

kbottla (commented):
LGTM! Added a minor comment about verifying emit event behaviour. Happy to discuss if it is not clear

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
jankowiakmaria: It's supposed to be a temporary change (I'll create the story that addresses "proper" implementation). Is it required to have this PR approved?

kbottla: 👍 Sounds fine

kbottla (approved):

PR #1531, opened 2019-07-23
BAU - Fix regex to convert class name to event_type

kbottla:
## WHAT YOU DID
Current regex doesn't consider numbers in class names (GatewayRequires3dsAuthorisation)

   - **Current** : Event type returned for `GatewayRequires3dsAuthorisation` is `GATEWAY_REQUIRES3DS_AUTHORISATION`
   - With the change, event type will be `GATEWAY_REQUIRES_3DS_AUTHORISATION`


jankowiakmaria (dismissed):
Test will be written in separate PR.

jankowiakmaria (approved):

PR #1537, opened 2019-07-25
Pp 5463 emit payment refund events

mrlumbu:
In order to keep track of the amount available to be refunded
in ledger, we need to emit `RefundAvailabilityUpdated` eevnts from
connector whenever refundability of a payment might change.
To do this I have changed how refund state transtions are
handled, so that rather than triggering the emsission of a single
external event it can emit a sequence of events (currently only
a sequence of either 1 or 2). The logic for all this has been moved
to the RefundEventFactory.
The refund availability of a payment only changes on refund creation
when the amount available to be refunded goes down) and refund error
(when it goes back up). The actual payload of the events is calculated
using the same tools that are currently used to populate responses,
so everything will line up.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
+++ b/src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
@@ -6,19 +6,16 @@ import uk.gov.pay.connector.chargeevent.dao.ChargeEventDao;
 import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
 import uk.gov.pay.connector.events.exception.StateTransitionMessageProcessException;
 import uk.gov.pay.connector.events.model.Event;
-import uk.gov.pay.connector.events.model.charge.PaymentEvent;
 import uk.gov.pay.connector.events.model.charge.PaymentEventFactory;
-import uk.gov.pay.connector.events.model.refund.RefundEvent;
 import uk.gov.pay.connector.events.model.refund.RefundEventFactory;
 import uk.gov.pay.connector.queue.PaymentStateTransition;
-import uk.gov.pay.connector.queue.StateTransitionQueue;
 import uk.gov.pay.connector.queue.QueueException;
 import uk.gov.pay.connector.queue.RefundStateTransition;
 import uk.gov.pay.connector.queue.StateTransition;
-import uk.gov.pay.connector.refund.dao.RefundDao;
-import uk.gov.pay.connector.refund.model.domain.RefundHistory;
+import uk.gov.pay.connector.queue.StateTransitionQueue;
 
 import javax.inject.Inject;
+import java.util.List;
 import java.util.Optional;
 
 public class StateTransitionEmitterProcess {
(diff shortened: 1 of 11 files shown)


src/test/java/uk/gov/pay/connector/events/StateTransitionEmitterProcessTest.java
```diff
@@ -92,25 +92,4 @@ public void shouldPutPaymentTransitionBackOnQueueIfEventEmitFails() throws Excep
 
         verify(mockStateTransitionQueue).offer(any(PaymentStateTransition.class));
     }
-
-    @Test
-    public void shouldNotPutPaymentTransitionBackOnQueueIfItHasExceededMaxAttempts() {
```
sfount: Was this supposed to be removed? Might need to be updated given the new `getNext` logic

mrlumbu: good spot 😬

src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
```diff
@@ -6,66 +6,63 @@
...
 public class StateTransitionEmitterProcess {
     private static final Logger LOGGER = LoggerFactory.getLogger(StateTransitionEmitterProcess.class);
 
     private final StateTransitionQueue stateTransitionQueue;
     private final EventQueue eventQueue;
-    private final ChargeEventDao chargeEventDao;
-    private final RefundDao refundDao;
+    private final RefundEventFactory refundEventFactory;
+    private final ChargeEventDao chargeEventDao
+            ;
```
jankowiakmaria: [minor] odd formatting

src/main/java/uk/gov/pay/connector/events/model/refund/RefundEventFactory.java
```diff
@@ -25,4 +69,25 @@ public static RefundEvent create(Class<? extends RefundEvent> eventClass, Refund
...
+            return Optional.ofNullable(chargeService.findChargeById(refundHistory.getChargeEntity().getExternalId()))
+                    .map(charge -> new RefundAvailabilityUpdated(
+                                charge.getExternalId(),
+                                RefundAvailabilityUpdatedEventDetails.from(
+                                        charge,
+                                        paymentProviders
+                                                .byName(charge.getPaymentGatewayName())
+                                                .getExternalChargeRefundAvailability(charge)
+                                ),
+                                refundHistory.getHistoryStartDate()
```
jankowiakmaria, line 85: What is `historyStartDate`?
Is it good enough to be a timestamp (and allows ordering events)?

src/test/java/uk/gov/pay/connector/it/events/StateTransitionsIT.java
```diff
@@ -95,24 +95,42 @@ public void shouldBeAbleToRequestARefund_partialAmount() throws Exception{
...
-        assertThat(messages.size(), is(3));
+        assertThat(messages.size(), is(4));
 
-        messages.sort(Comparator.comparing(this::getTimestampFromMessage));
-
-        JsonObject message1 = new JsonParser().parse(messages.get(0).getBody()).getAsJsonObject();
+        JsonObject message1 = messages.stream()
+                .map(m -> new JsonParser().parse(m.getBody()).getAsJsonObject())
+                .filter(m -> "REFUND_CREATED_BY_SERVICE".equals(m.get("event_type").getAsString()))
+                .findFirst().get();
```
jankowiakmaria, line 103: [minor] This could be extracted as a helper function given it repeats for each message

jankowiakmaria (dismissed):
Two minor comments and one question (but in general approved).

src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
mrlumbu: oof

jankowiakmaria (approved):

src/main/java/uk/gov/pay/connector/events/model/refund/RefundEventFactory.java
mrlumbu: It can either be this or the `historyEndDate`. These are automatically generated by the history policy see https://wiki.eclipse.org/EclipseLink/Examples/JPA/History
> When enabled EclipseLink will insert a row into the history table on every update to the primary table to track the rows state. The history table also defines a start and end timestamp, so that the state of any row at any point in time can be determined.

The start date is the date at which the state described in the row took effect, and the end date is the date at which it ceased to take effect. This means the end date can be null.

src/test/java/uk/gov/pay/connector/it/events/StateTransitionsIT.java
mrlumbu: yeah, I will look at this in the next PR

src/main/java/uk/gov/pay/connector/events/model/refund/RefundEventFactory.java
jankowiakmaria: Thanks for explaining.

PR #1542, opened 2019-07-26
PP-5492 Fix delayed capture on confirmation page

sfount:
* charge service marking charge method only seem to be used by delayed
capture, update their names to reflect this
* add explicit check for `AWAITING_CAPTURE` - this makes the method line
up with the exception thrown as `transitionChargeState` will allow
`AUTHORISATION_SUCCESS` to transition to `CAPTURE_APPROVED`
* verify the behaviour

Expected behaviour: 
* calling the delayed charge end point before the user has pressed 'confirm' during the payment flow should be rejected 
* the delayed charge end point should continue to be idempotent for successful already captured charges

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -90,14 +90,6 @@ public class ChargeService {
 
     private static final List<ChargeStatus> CURRENT_STATUSES_ALLOWING_UPDATE_TO_NEW_STATUS = newArrayList(CREATED, ENTERING_CARD_DETAILS);
 
-    private static final List<ChargeStatus> IGNORABLE_CAPTURE_STATES = ImmutableList.of(
-            CAPTURE_APPROVED,
-            CAPTURE_APPROVED_RETRY,
-            CAPTURE_READY,
-            CAPTURE_SUBMITTED,
-            CAPTURED
-    );
-
     private final ChargeDao chargeDao;
     private final ChargeEventDao chargeEventDao;
     private final CardTypeDao cardTypeDao;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -508,13 +508,19 @@ public ChargeEntity markChargeAsCaptureApproved(String externalId) {
...
 
-            try {
-                transitionChargeState(charge, CAPTURE_APPROVED);
-            } catch (InvalidStateTransitionException e) {
+            if (currentStatus == AWAITING_CAPTURE_REQUEST) {
+                try {
+                    transitionChargeState(charge, CAPTURE_APPROVED);
+                } catch (InvalidStateTransitionException e) {
+                    throw new ConflictRuntimeException(charge.getExternalId(),
+                            format("attempt to perform delayed capture on invalid charge state"));
```
alexbishop1: I don’t think there was any need to change this message: it was fine before. Right now the `format(…)` isn’t actually doing anything.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -508,13 +508,19 @@ public ChargeEntity markChargeAsCaptureApproved(String externalId) {
...
+                }
+            } else {
```
alexbishop1: So, although I’m not sure it ever worked, the original acceptance criteria for this feature (PP-4165) said that if a payment is in the `CAPTURE_APPROVED`, `CAPTURE_APPROVED_RETRY`, `CAPTURE_READY`, `CAPTURE_SUBMITTED` or `CAPTURED` states, hitting the capture endpoint should not change the state but appear to work and not return an error (making the operation idempotent).

alexbishop1 (changes requested):
Looks good: just a couple of things.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
sfount: Yeah, it looks like this line of code should still meet that criteria, bizarrely: 
https://github.com/alphagov/pay-connector/blob/b5eb31a588c038bb97361222e183330db0cc874b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java#L506

Basically short circuits if it's complete

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
sfount: Oops forgot about the format! 

For the message I just wanted to differentiate it, as there are two logical checks happening

1. if it's not `AWAITING_CAPTURE_REQUEST`, it will continue with the previous message 
`"attempt to perform delayed capture on charge not in %s state.", AWAITING_CAPTURE_REQUEST`
2. the state machine has thrown an exception saying it can't progress, at this point we can't guarantee that the relevant state is `AWAITING_CAPTURE_REQUEST`. The message here doesn't give away the full details but should describe what the issue is
`"attempt to perform delayed capture on invalid charge state"`

sfount:
@alexbishop1 thanks for the review! 

Removed the dangling `format` https://github.com/alphagov/pay-connector/pull/1542/commits/93baff853b064664cef76e2d23ff3294feb0aa24. If my reasoning around the error doesn't make sense this can be updated too https://github.com/alphagov/pay-connector/pull/1542#discussion_r308162749.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Oh, I missed that! Guess it already works then.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Fair enough.

alexbishop1: But it would be good to include the message from the `InvalidStateTransitionException` in the `ConflictRuntimeException`:

```java
throw new ConflictRuntimeException(charge.getExternalId(), "attempt to perform delayed capture on invalid charge state: " + e.getMessage()));
```

alexbishop1 (changes requested):
As evidenced by the fact I missed it, I think the existing code could be clearer. Right now we have to jump from a method in the middle, to a method at the end, to a constant at the top to follow what’s going on. I think this could be clearer with a `switch` statement:

```java
return chargeDao.findByExternalId(externalId).map(charge -> {
    switch (fromString(charge.getStatus())) {
        case AWAITING_CAPTURE_REQUEST:
            try {
                transitionChargeState(charge, CAPTURE_APPROVED);
            } catch (InvalidStateTransitionException e) {
                throw new ConflictRuntimeException(charge.getExternalId(),
                        "attempt to perform delayed capture on invalid charge state " + e.getMessage());
            }
            return charge; 
        case CAPTURE_APPROVED:
        case CAPTURE_APPROVED_RETRY:
        case CAPTURE_READY:
        case CAPTURE_SUBMITTED:
        case CAPTURED:
            return charge;
        default:
            throw new ConflictRuntimeException(charge.getExternalId(),
                    format("attempt to perform delayed capture on charge not in %s state.", AWAITING_CAPTURE_REQUEST));
    }
}).orElseThrow(() -> new ChargeNotFoundRuntimeException(externalId));
```

Then the `chargeCanBeSkipped(ChargeEntity)` method and `IGNORABLE_CAPTURE_STATES` constant can be removed.

sfount:
Understood 👍 aimed to change as little as possible get this fix in but agreed that the switch reads more clearly

sfount:
@alexbishop1 https://github.com/alphagov/pay-connector/commit/f51ca6d333c3638c0b1cc0fd9dd05d9da1e95b4a
* use proposed switch method to make code more legible 
* remove unused helper methods/ state maps

alexbishop1 (dismissed):
Much nicer!

alexbishop1 (approved):

PR #1546, opened 2019-07-29
PP-5469: add transaction id to pact contract for PaymentDetailsEntered event

jankowiakmaria:


kbottla (approved):
LGTM!  Can this PR not be tied to `PP-5469` ?

jankowiakmaria:
It can - just being lazy.

PR #1549, opened 2019-07-30
PP-5463 Emit refund availability updated on capture

mrlumbu:
Refund availability changes either on capture or capture submitted, so
emit a refund availability updated event when either of these events occur

I have consolidated the event creation into a single event factory, which
means StateTransitionEmitterProcess no longer has to know about the
details of event creation at all, which seems right.


src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
```diff
@@ -0,0 +1,165 @@
...
+    private List<Event> createRefundEvents(RefundStateTransition refundStateTransition) throws EventCreationException {
+        RefundHistory refundHistory = refundDao.getRefundHistoryByRefundExternalIdAndRefundStatus(
+                refundStateTransition.getRefundExternalId(),
+                refundStateTransition.getRefundStatus())
+                .orElseThrow(() -> new EventCreationException(refundStateTransition.getIdentifier()));
+
+        Event refundEvent = createRefundEvent(refundHistory, refundStateTransition.getStateTransitionEventClass());
+        Optional<Event> refundAvailabilityEvent = createRefundAvailabilityUpdatedEvent(
+                refundHistory.getChargeEntity().getExternalId(),
+                refundHistory.getHistoryEndDate(),
```
jankowiakmaria: You've said that `historyEndDate` can be null. Why the change from `historyStartDate`?

src/test/java/uk/gov/pay/connector/events/model/refund/EventFactoryTest.java
```diff
@@ -192,4 +209,94 @@ public void shouldCreateCorrectEventsFromRefundErrorStateTransition() throws Exc
...
+        List<Event> events = eventFactory.createEvents(paymentStateTransition);
+
+        assertThat(events.size(), is(2));
+        CaptureSubmitted event1 = (CaptureSubmitted) events.get(0);
+        Assert.assertThat(event1, instanceOf(CaptureSubmitted.class));
+
+        RefundAvailabilityUpdated event2 = (RefundAvailabilityUpdated) events.get(1);
+        Assert.assertThat(event2, instanceOf(RefundAvailabilityUpdated.class));
+        Assert.assertThat(event2.getEventDetails(), instanceOf(RefundAvailabilityUpdatedEventDetails.class));
+    }
```
jankowiakmaria, line 301: Inconsistent test naming (`should...for`, `given...should`).

jankowiakmaria (commented):
One question, one minor comment (that doesn't need to be addressed as part of this PR).

Other than that LGTM, although I'm not sure about merging factories into one bigger class (good for sharing `createRefundAvailabilityUpdatedEvent`, but slightly harder to read - would be good to improve in the next iteration).

src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
mrlumbu: yeah this is wrong. thanks

jankowiakmaria (approved):

PR #1553, opened 2019-08-01
PP-5494 Add gateway_account_id to Refund events payload

kbottla:
## WHAT 
- Adds `gateway_account_id` to `RefundCreatedByUser` & `RefundCreatedByService` events

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
+++ b/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
@@ -2,6 +2,7 @@ package uk.gov.pay.connector.refund.model.domain;
 
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.model.domain.UTCDateTimeConverter;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 
 import java.sql.Timestamp;
 import java.time.ZonedDateTime;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
```diff
@@ -18,16 +19,21 @@
 
     public RefundHistory(Long id, String externalId, Long amount, String status, Long chargeId, Timestamp createdDate, 
                          Long version, String reference, Timestamp historyStartDate, Timestamp historyEndDate, 
-                         String userExternalId, String gatewayTransactionId, String chargeExternalId) {
+                         String userExternalId, String gatewayTransactionId, String chargeExternalId, Long gatewayAccountId) {
```
jankowiakmaria, line 22: It would be nice to add builder at some point. (follow-up PR)

jankowiakmaria (approved):

src/main/java/uk/gov/pay/connector/refund/model/domain/RefundHistory.java
kbottla: Replacing with builder might not be possible as we need constructor for \@SqlResultSetMapping https://github.com/alphagov/pay-connector/blob/3865a9a5c329a33fb2874a6e10fc15e88fdd34b2/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundEntity.java#L34

PR #1556, opened 2019-08-01
PP-5407 Include event_date to check for emitted events

kbottla:
- Use `event_date` along with other parameters to check if event has been emitted before. This will also us to emit similar events with different timestamp


sfount:
Relates to https://github.com/alphagov/pay-ledger/pull/187

sfount (approved):

PR #1557, opened 2019-08-02
PP-5407 Backfill - emit all charge events

kbottla:
## WHAT YOU DID

- HistoricalEventEmitterWorker emits all charge events (except `PaymentDetailsEntered` event)
- Uses ordered list of charge events - starting with first charge event created and then the rest in order
- Uses existing `PaymentGatewayStateTransitions` to derive and emit event

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
+++ b/src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
@@ -19,7 +19,7 @@ public class EmittedEventDao extends JpaDao<EmittedEventEntity> {
         super(entityManager);
     }
 
-    public boolean hasBeenEmittedBefore(Event paymentCreatedEvent) {
+    public boolean hasBeenEmittedBefore(Event event) {
         final List<Integer> singleResult = entityManager.get()
                 .createQuery("select 1 from EmittedEventEntity e where " +
                         "e.resourceType = :resource_type AND " +
@@ -27,12 +27,12 @@ public class EmittedEventDao extends JpaDao<EmittedEventEntity> {
                         "e.resourceExternalId = :resource_external_id AND " +
                         "e.eventType = :event_type", Integer.class)
                 .setMaxResults(1)
-                .setParameter("resource_type", paymentCreatedEvent.getResourceType().getLowercase())
-                .setParameter("resource_external_id", paymentCreatedEvent.getResourceExternalId())
-                .setParameter("event_type", paymentCreatedEvent.getEventType())
-                .setParameter("event_date", paymentCreatedEvent.getTimestamp())
+                .setParameter("resource_type", event.getResourceType().getLowercase())
+                .setParameter("resource_external_id", event.getResourceExternalId())
+                .setParameter("event_type", event.getEventType())
+                .setParameter("event_date", event.getTimestamp())
                 .getResultList();
-        return singleResult.size() > 0;
+        return !singleResult.isEmpty();
     }
 
     public void recordEmission(Event event) {
@@ -42,7 +42,7 @@ public class EmittedEventDao extends JpaDao<EmittedEventEntity> {
                 event.getTimestamp(),
                 ZonedDateTime.now()
         );
-        
+
         persist(emittedEvent);
     }
 }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
```diff
@@ -51,33 +59,68 @@ public void execute(Long startId, OptionalLong maybeMaxId) {
...
+                .getEventForTransition(fromChargeState, chargeEventEntity.getStatus())
+                .ifPresent(eventType -> {
+                    Event event = getPaymentEvent(chargeEventEntity, eventType);
+                    final boolean emittedBefore = emittedEventDao.hasBeenEmittedBefore(event);
+
+                    if (emittedBefore) {
+                        logger.info("[{}/{}] - found - emitted before", currentId, maxId);
+                    } else {
+                        logger.info("[{}/{}] - found - emitting {} ", currentId, maxId, event);
+                        persistAndEmit(event);
```
sfount: Even though it will end up with Connector going to the database twice I think we should use the emitter process here instead of manually emitting -- it takes care of handling multiple events created for a given transition (refund availability etc.) and will keep up with changes to the `PaSTEE` code.

src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
```diff
@@ -19,20 +19,20 @@ protected EmittedEventDao(Provider<EntityManager> entityManager) {
...
-                .setParameter("resource_external_id", paymentCreatedEvent.getResourceExternalId())
-                .setParameter("event_type", paymentCreatedEvent.getEventType())
-                .setParameter("event_date", paymentCreatedEvent.getTimestamp())
+                .setParameter("resource_type", event.getResourceType().getLowercase())
+                .setParameter("resource_external_id", event.getResourceExternalId())
+                .setParameter("event_type", event.getEventType())
+                .setParameter("event_date", event.getTimestamp())
                 .getResultList();
-        return singleResult.size() > 0;
+        return !singleResult.isEmpty();
```
sfount, line 35: 👍

sfount (commented):
Looking good! Even though it's more expensive I think we should offer the state transition to the process that knows how to deal with it, should be good to go then

src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
```diff
@@ -99,7 +99,7 @@ public EventFactory(ChargeService chargeService, RefundDao refundDao, ChargeEven
                 .collect(Collectors.toList());
     }
 
-    private static PaymentEvent createPaymentEvent(ChargeEventEntity chargeEvent, Class<? extends PaymentEvent> eventClass) {
```
sfount, line 102: Is there a reason to update this access modifier?

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
```diff
@@ -51,33 +59,68 @@ public void execute(Long startId, OptionalLong maybeMaxId) {
...
 
-    private void persistAndEmit(PaymentCreated event) {
+    private List<ChargeEventEntity> getSortedChargeEvents(ChargeEntity charge) {
+        return charge.getEvents()
+                .stream()
+                .sorted(Comparator.comparing(ChargeEventEntity::getUpdated))
+                .collect(Collectors.toList());
+    }
+
+    private void processChargeEvents(long currentId, List<ChargeEventEntity> chargeEventEntities) {
```
sfount, line 87: Given the separation between state transition `PaSTEE` events and manual events it might be useful to name this something like `processStateTransitionChargeEvents`

src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
kbottla: Reverted to use `createEvents`

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
kbottla: Updated to use `stateTransitionQueue.offer(transition);`

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
kbottla: with the refactor added `offerPaymentStateTransitionEvents`. We can change it as we set fit when adding manual events

sfount (approved):

PR #1560, opened 2019-08-04
PP-5407 Derive intermediate ready state for transition

kbottla:
## WHAT YOU DID
- HistoricalEventEmitter uses ChargeEvents ordering to derive state transition events. But PaymentGatewayStateTransitions for ongoing operations differs as transitions are based on `READY` state.
- HistoricalEventEmitter now derives intermediate `READY` state (if a transition is not available) between two statuses and uses derived intermediate state to offer state transition event

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
@@ -5,6 +5,8 @@ import com.google.common.graph.MutableValueGraph;
 import com.google.common.graph.ValueGraphBuilder;
 import org.apache.commons.lang3.tuple.Triple;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.events.model.Event;
+import uk.gov.pay.connector.events.model.UnspecifiedEvent;
 import uk.gov.pay.connector.events.model.charge.AuthorisationCancelled;
 import uk.gov.pay.connector.events.model.charge.AuthorisationRejected;
 import uk.gov.pay.connector.events.model.charge.AuthorisationSucceeded;
@@ -21,7 +23,6 @@ import uk.gov.pay.connector.events.model.charge.CaptureAbandonedAfterTooManyRetr
 import uk.gov.pay.connector.events.model.charge.CaptureConfirmed;
 import uk.gov.pay.connector.events.model.charge.CaptureErrored;
 import uk.gov.pay.connector.events.model.charge.CaptureSubmitted;
-import uk.gov.pay.connector.events.model.Event;
 import uk.gov.pay.connector.events.model.charge.GatewayErrorDuringAuthorisation;
 import uk.gov.pay.connector.events.model.charge.GatewayRequires3dsAuthorisation;
 import uk.gov.pay.connector.events.model.charge.GatewayTimeoutDuringAuthorisation;
@@ -30,7 +31,6 @@ import uk.gov.pay.connector.events.model.charge.PaymentExpired;
 import uk.gov.pay.connector.events.model.charge.PaymentStarted;
 import uk.gov.pay.connector.events.model.charge.ServiceApprovedForCapture;
 import uk.gov.pay.connector.events.model.charge.UnexpectedGatewayErrorDuringAuthorisation;
-import uk.gov.pay.connector.events.model.UnspecifiedEvent;
 import uk.gov.pay.connector.events.model.charge.UserApprovedForCapture;
 import uk.gov.pay.connector.events.model.charge.UserApprovedForCaptureAwaitingServiceApproval;
 
@@ -190,7 +190,7 @@ public class PaymentGatewayStateTransitions {
     }
 
     public <T extends Event> Optional<Class<T>> getEventForTransition(ChargeStatus fromStatus, ChargeStatus toStatus) {
-        return graph.edgeValue(fromStatus , toStatus)
+        return graph.edgeValue(fromStatus, toStatus)
                 .map(modelledEvent -> {
                     try {
                         ModelledTypedEvent<T> modelledTypedEvent = (ModelledTypedEvent) modelledEvent;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorkerTest.java
```diff
@@ -178,7 +179,41 @@ public void executeShouldEmitManualEventsWithTerminalAuthenticationState() throw
...
+        worker.execute(1L, OptionalLong.empty());
+
+        ArgumentCaptor<StateTransition> argument = ArgumentCaptor.forClass(StateTransition.class);
+        verify(stateTransitionQueue, times(1)).offer(argument.capture());
+
+        assertThat(argument.getAllValues().get(0).getStateTransitionEventClass(), is(AuthorisationSucceeded.class));
+
+        ArgumentCaptor<Event> daoArgumentCaptor = ArgumentCaptor.forClass(Event.class);
+        verify(emittedEventDao, times(2)).recordEmission(daoArgumentCaptor.capture()); // 2 times due to paymentDetailsEnteredEvent
+        assertThat(daoArgumentCaptor.getAllValues().get(0).getEventType(), is("AUTHORISATION_SUCCEEDED"));
```
sfount, line 230: 🎉

src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
```diff
@@ -205,6 +205,24 @@ public static boolean isValidTransition(ChargeStatus state, ChargeStatus targetS
         return getInstance().isValidTransitionImpl(state, targetState, event);
     }
 
+    public Optional<ChargeStatus> getIntermediateChargeStatus(ChargeStatus fromStatus, ChargeStatus toStatus) {
```
sfount, line 208: Unsure about adding this method to the this class just for the purpose of backfilling, I wonder if it could be moved to some utility of the history event emitter?

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
```diff
@@ -125,13 +138,38 @@ private void processChargeStateTransitionEvents(long currentId, List<ChargeEvent
...
+    private Optional<Class<Event>> getEventForTransition(ChargeStatus fromChargeStatus,
+                                                         ChargeEventEntity chargeEventEntity) {
+        Optional<Class<Event>> eventForTransition = paymentGatewayStateTransitions
+                .getEventForTransition(fromChargeStatus, chargeEventEntity.getStatus());
+
+        if (eventForTransition.isEmpty()) {
+            Optional<ChargeStatus> intermediateChargeStatus = paymentGatewayStateTransitions
+                    .getIntermediateChargeStatus(fromChargeStatus, chargeEventEntity.getStatus());
+
+            if (intermediateChargeStatus.isPresent() && INTEMEDIATE_READY_STATES.contains(intermediateChargeStatus.get())) {
```
sfount: Could it make sense to put this at the top of the method as these are the only cases where an intermediate state would exist? It might simplify this method a lot, something like: 

```java
ChargeStatus currentEventStatus = chargeEventEntity.getStatus();
if (INTERMEDIATE_READY_STATES.contains(currentEventStatus)) { 

  return paymentGatewayStateTransitions.getIntermediateChargeStatus(fromChargeStatus, chargeEventStatus)
    .map(intermediate -> paymentGatewayStateTransitions.getEventForTransition(intermediate.get(), currentEventStatus);
} else {

  return paymentGatewayStateTransitions.getEventForTransition(fromChargeStatus, currentEventStatus);
}
```

kbottla: `chargeEventEntity` status will never one of the READY states.
My thought was to
 - get payment transition from previous charge event status to the current
 - If no transition is available, get available intermediate state (could also be other than ready state depending on from and to statuses)
 - if it is one of the READY states, get payment transition using ready state
 - if available, process with READY state

sfount: Ah you're right! Somehow missed that, this might be simpler with `map` `orElse` but this is readable and does what you've described.

src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
kbottla: As discussed, this is ideal place as splitting out will keep the logic at different places. It is mostly generic but the focus is more on locking states for now.

src/test/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitionsTest.java
```diff
@@ -71,4 +98,54 @@ public void getEventTransitionFor_returnsEmptyForUnmodelledEvent() {
...
+                new Object[]{SYSTEM_CANCEL_READY, AUTHORISATION_SUCCESS, SYSTEM_CANCELLED},
+                new Object[]{SYSTEM_CANCEL_READY, AUTHORISATION_SUCCESS, SYSTEM_CANCEL_ERROR},
+                new Object[]{SYSTEM_CANCEL_READY, AWAITING_CAPTURE_REQUEST, SYSTEM_CANCEL_SUBMITTED},
+                new Object[]{SYSTEM_CANCEL_READY, AWAITING_CAPTURE_REQUEST, SYSTEM_CANCELLED},
+                new Object[]{SYSTEM_CANCEL_READY, AWAITING_CAPTURE_REQUEST, SYSTEM_CANCEL_ERROR},
+        };
+    }
+
+    @Test
+    @Parameters(method = "intermediateStatesForTransitions")
```
sfount, line 144: Didn't fully appreciate what this test was doing before, nice!

sfount (approved):
Great work! Looks like a good solution to this problem. Potentially could move the more tightly coupled intermediate code out of the graph in the future but for now it should do the job well 👍

PR #1569, opened 2019-08-06
PP-5520 Delay state transition process w/ blocking queue

sfount:
State transition events are currently rate limited by a scheduled
executor service waiting delaying polling by 100ms. This ensures
Connector is not constantly polling but doesn't allow the in-memory
queue to be cleared effectively given many state transitions.

* Use the blocking queue poll timeout interface to wait for a `Delayed`
entry to become visible -- this includes the delay if there is nothing
to process but lets the emitter process all elements if there is backlog
* Just use an executor service as there is no scheduling required

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
+++ b/src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
@@ -10,10 +10,12 @@ import uk.gov.pay.connector.queue.StateTransitionQueue;
 
 import javax.inject.Inject;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 public class StateTransitionEmitterProcess {
     private static final Logger LOGGER = LoggerFactory.getLogger(StateTransitionEmitterProcess.class);
 
+    private final long STATE_TRANSITION_PROCESS_DELAY_IN_MILLISECONDS = 100;
     private final StateTransitionQueue stateTransitionQueue;
     private final EventQueue eventQueue;
     private final EventFactory eventFactory;
@@ -29,8 +31,8 @@ public class StateTransitionEmitterProcess {
         this.eventFactory = eventFactory;
     }
 
-    public void handleStateTransitionMessages() {
-        Optional.ofNullable(stateTransitionQueue.poll())
+    public void handleStateTransitionMessages() throws InterruptedException {
+        Optional.ofNullable(stateTransitionQueue.poll(STATE_TRANSITION_PROCESS_DELAY_IN_MILLISECONDS, TimeUnit.MILLISECONDS))
                 .ifPresent(this::emitEvents);
     }
 
--- a/src/main/java/uk/gov/pay/connector/queue/StateTransitionQueue.java
+++ b/src/main/java/uk/gov/pay/connector/queue/StateTransitionQueue.java
@@ -1,16 +1,20 @@
 package uk.gov.pay.connector.queue;
 
-import java.util.Queue;
 import java.util.concurrent.DelayQueue;
+import java.util.concurrent.TimeUnit;
 
 public class StateTransitionQueue {
-    private final Queue<StateTransition> queue = new DelayQueue<>();
-    
+    private final DelayQueue<StateTransition> queue = new DelayQueue<>();
+
     public boolean offer(StateTransition stateTransition) {
         return queue.offer(stateTransition);
     }
-    
-    public StateTransition poll() {
-        return queue.poll();
+
+    public StateTransition poll(long timeout, TimeUnit unit) throws InterruptedException {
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/pay/connector/queue/managed/QueueMessageReceiver.java
```diff
@@ -59,11 +59,7 @@ public void start() {
                 queueSchedulerThreadDelayInSeconds,
                 TimeUnit.SECONDS);
 
-        stateTransitionMessageExecutorService.scheduleWithFixedDelay(
-                this::stateTransitionMessageReceiver,
-                0,
-                100,
-                TimeUnit.MILLISECONDS);
+        stateTransitionMessageExecutorService.execute(this::stateTransitionMessageReceiver);
```
sfount: Looks like `execute` will only call the runnable once -- will either need to wrap in an interruptible thread or use something that will continuously call

src/main/java/uk/gov/pay/connector/queue/StateTransitionQueue.java
```diff
@@ -1,16 +1,20 @@
...
     }
-    
-    public StateTransition poll() {
-        return queue.poll();
+
+    public StateTransition poll(long timeout, TimeUnit unit) throws InterruptedException {
+        return queue.poll(timeout, unit);
+    }
+
+    public StateTransition poll() throws InterruptedException {
```
kbottla, line 18: This is now only used in test. Could be replaced by new method `poll(long timeout, TimeUnit unit)`

src/main/java/uk/gov/pay/connector/queue/managed/QueueMessageReceiver.java
```diff
@@ -9,6 +9,7 @@
 import uk.gov.pay.connector.paymentprocessor.service.CardCaptureProcess;
 
 import javax.inject.Inject;
+import java.util.concurrent.ExecutorService;
```
kbottla: Unused import

kbottla (commented):
Looks good to me. Added a couple of minor comments

src/main/java/uk/gov/pay/connector/queue/StateTransitionQueue.java
```diff
@@ -1,16 +1,20 @@
 package uk.gov.pay.connector.queue;
 
-import java.util.Queue;
 import java.util.concurrent.DelayQueue;
+import java.util.concurrent.TimeUnit;
 
 public class StateTransitionQueue {
-    private final Queue<StateTransition> queue = new DelayQueue<>();
-    
+    private final DelayQueue<StateTransition> queue = new DelayQueue<>();
```
mrlumbu: pedantry alert - would be a tiny bit nicer if you declared this as a `BlockingQueue` (i.e as the interface, not the implementation). Given this is `private final`, it doesn't make any difference functionally. It would possibly make it clearer how this all works however

src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
```diff
@@ -29,8 +31,8 @@ public StateTransitionEmitterProcess(
         this.eventFactory = eventFactory;
     }
 
-    public void handleStateTransitionMessages() {
-        Optional.ofNullable(stateTransitionQueue.poll())
+    public void handleStateTransitionMessages() throws InterruptedException {
+        Optional.ofNullable(stateTransitionQueue.poll(STATE_TRANSITION_PROCESS_DELAY_IN_MILLISECONDS, TimeUnit.MILLISECONDS))
```
mrlumbu, line 35: I wonder if PaSTEE or the StateTransitionQueue itself should decide on the poll timeout?

src/main/java/uk/gov/pay/connector/queue/StateTransitionQueue.java
sfount: This is just a convenience method for anything calling the queue that doesn't want to wait if something isn't available instantly, can remove if needed but seems a sensible default

src/main/java/uk/gov/pay/connector/queue/StateTransitionQueue.java
sfount: Good suggestion! :done:

src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
sfount: Going to leave for now as it allows other classes (tests etc.) to poll without blocking, we could move this if we end up polling in more than one place

kbottla (approved):
LGTM!

PR #1579, opened 2019-08-09
Use shared job for running end to end tests

rhowe-gds:
Make connector behave like our other apps, where end to end is done in a
separate job.


SandorArpa (approved):
LOKTM

PR #1590, opened 2019-08-13
PP-5552 Emit refund availability update on payment creation

mrlumbu:
In order for ledger to know the refund availability of a payment throughout the whole lifecycle of the payment, we need to emit this on creation as well as at other points affecting refund availability.


kbottla (approved):
LGTM!

PR #1591, opened 2019-08-13
PP-5552 Emit refund availability updated events on all terminal trans…

mrlumbu:
…itions

We need to update refund availability of a payment whenever it reaches
a terminal state. This updates event factory to enable that.


src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
```diff
@@ -200,6 +202,17 @@ private PaymentGatewayStateTransitions() {
                     }
                 });
     }
+    
+    public static List<Class> getAllEventsResultingInTerminalState() {
+        return getInstance().graph.edges()
+                .stream()
+                .filter(endpointPair -> getInstance().graph.successors(endpointPair.nodeV()).isEmpty())
```
sfount, line 208: :iq-200:

sfount (dismissed):

sfount (approved):

PR #1594, opened 2019-08-14
bau: Remove extra persistence id

oswaldquek:
There is already an `@Id` annotation on the chargeId


alexbishop1 (approved):
Yeah, that looks like it was a mistake.

PR #1600, opened 2019-08-15
PP-5544 Add 3ds_integration_version column

stephencdaly:
Add 3ds_integration_version column to the gateway_accounts table.
Current expected values are 1 or 2.

A value of 1 indicates the current 3DS integration should be used.
A value of 2, in the case of Worldpay gateway accounts for now, indicates that the 3DS Flex integration should be used.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1181,4 +1181,21 @@
                 columnName="return_url" />
     </changeSet>
 
+    <changeSet id="add integration_version_3ds column to gateway_accounts table" author="">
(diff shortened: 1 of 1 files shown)


oswaldquek:
According to https://www.braintreepayments.com/blog/safe-operations-for-high-volume-postgresql/ - "Add a column with a default", the way to do this is 1) add the column 2) in another script add the default value 3) in another script set the column to be not null

stephencdaly:
@oswaldquek I've updated to do those things in multiple stages

oswaldquek (approved):

PR #1602, opened 2019-08-15
PP-5547 Add Json Web Token support

danworth:
Create a JWTFacade class to offer simple creation of JWTs. This uses
`io.jsonwebtoken` which is listed as a reputable library on `jwt.io`, it
has a simple builder pattern to generate the tokens and an Apach 2.0
licence. It enforces key strength when signing and will not permit the
use of a key considered too weak for the desired algorithm (e.g. for
HSA256 the key must be greater than 256 bits). Since the Worldpay test
token is only 288 bits and we cannot change it, this initial
implementation uses HSA256 to sign the JWT, which is supported by
Worldpay.

This commit includes the first public method on the facade to generate a
Worldpay 3DS Flex DDC token and a unit test to ensure the token, when
decoded includes the correct header and body elements.

## How to test
I've added a unit test but you can independently check this generates valid tokens using https://www.jsonwebtoken.io

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,6 +18,7 @@
         <jooq.version>3.11.11</jooq.version>
         <postgresql.version>42.2.6</postgresql.version>
         <commons-lang3.version>3.9</commons-lang3.version>
+        <jjwt.version>0.10.7</jjwt.version>
     </properties>
     <repositories>
         <repository>
@@ -463,6 +464,23 @@
             <artifactId>testcontainers</artifactId>
             <version>1.12.0</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-api</artifactId>
+            <version>${jjwt.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-impl</artifactId>
+            <version>${jjwt.version}</version>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-jackson</artifactId>
+            <version>${jjwt.version}</version>
+            <scope>runtime</scope>
         </dependency>
     </dependencies>
     <build>
--- a/src/test/java/uk/gov/pay/connector/charge/util/JwtGeneratorTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/util/JwtGeneratorTest.java
@@ -1,0 +1,39 @@
+package uk.gov.pay.connector.charge.util;
+
+
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.Jws;
+import io.jsonwebtoken.Jwts;
+import org.junit.Test;
+
+import javax.crypto.spec.SecretKeySpec;
+
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class JwtGeneratorTest {
+
+    private static final JwtGenerator jwtGenerator = new JwtGenerator();
+
+    @Test
+    public void shouldCreateCorrectTokenForWorldpay3dsFlexDdc() {
+        String secret = "fa2daee2-1fbb-45ff-4444-52805d5cd9e0";
+        String issuer = "ME";
+        String orgId = "myOrg";
+
+        String token = jwtGenerator.createWorldpay3dsFlexDdcJwt(issuer, orgId, secret);
+
+        Jws<Claims> jws = Jwts.parser()
+                .setSigningKey(new SecretKeySpec(secret.getBytes(), "HmacSHA256"))
+                .parseClaimsJws(token);
+
+        assertThat(jws.getHeader().getAlgorithm(), is("HS256"));
+        assertThat(jws.getHeader().get("typ"), is("JWT"));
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -464,6 +464,23 @@
...
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-impl</artifactId>
+            <version>0.10.7</version>
+            <scope>runtime</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt-jackson</artifactId>
+            <version>0.10.7</version>
```
alexbishop1: Can you extract these version strings into a property?

src/main/java/uk/gov/pay/connector/charge/util/JwtFacade.java
```diff
@@ -0,0 +1,41 @@
...
+
+import io.jsonwebtoken.Jwts;
+import io.jsonwebtoken.SignatureAlgorithm;
+import uk.gov.pay.connector.util.RandomIdGenerator;
+
+import javax.crypto.spec.SecretKeySpec;
+import java.time.Instant;
+import java.util.Map;
+
+public class JwtFacade {
```
alexbishop1: I’d be really specific with this name and call it `Worldpay3dsFlexDdcJwtGenerator` or something. We’re going to want different JWTs in different places — if we have specific classes for generating each, we can inject just what we need wherever we need it.

src/main/java/uk/gov/pay/connector/charge/util/JwtFacade.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Utility method to create a JWT token for Worldpay 3DS Flex DDC based upon the required claims
```
alexbishop1: “JWT token” = “JSON Web Token token” 😀

src/test/java/uk/gov/pay/connector/charge/util/JwtFacadeTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+        String token = jwtFacade.createWorldpay3dsFlexDdcJwt(issuer, orgId, secret);
+
+        Jws<Claims> jws = Jwts.parser()
+                .setSigningKey(new SecretKeySpec(secret.getBytes(), "HmacSHA256"))
+                .parseClaimsJws(token);
+
+        assertThat(jws.getHeader().getAlgorithm(), is("HS256"));
+        assertThat(jws.getHeader().get("typ"), is("JWT"));
+        assertThat(jws.getBody().get("jti"), notNullValue());
```
alexbishop1: `is(notNullValue())` would be more consistent

alexbishop1 (changes requested):
Looks good — just a couple of things

pom.xml
danworth: yes good idea, I've got that change ready to go 👍

src/main/java/uk/gov/pay/connector/charge/util/JwtFacade.java
danworth: lol I thought I'd stopped myself from doing that. Removed 😄

src/main/java/uk/gov/pay/connector/charge/util/JwtFacade.java
danworth: Renamed it JwtGenerator and we can look at what abstraction or specificity is required as it's use evolves with the future planned uses.

src/test/java/uk/gov/pay/connector/charge/util/JwtFacadeTest.java
danworth: Done 👍

pom.xml
```diff
@@ -18,6 +18,7 @@
         <jooq.version>3.11.11</jooq.version>
         <postgresql.version>42.2.6</postgresql.version>
         <commons-lang3.version>3.9</commons-lang3.version>
+        <jsonwebtoken.version>0.10.7</jsonwebtoken.version>
```
alexbishop1: I’m being picky here but the library is called JJWT. It’s good to be unambiguous in case some horror means we end up with multiple JWT libraries!

src/test/java/uk/gov/pay/connector/charge/util/JwtGeneratorTest.java
```diff
@@ -0,0 +1,39 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class JwtGeneratorTest {
+
+    private static final JwtGenerator jwtGenerator = new JwtGenerator();
+
+    @Test
+    public void shouldCreateCorrectTokenForWorldPay3ds2() {
```
alexbishop1: Missed this last time: can you make this `shouldCreateCorrectTokenForWorldpay3dsFlexDdc` or something, as we’ll be doing other Worldpay JWTs later.

alexbishop1 (changes requested):
Thanks for making the changes. Just a couple of tiny things…

pom.xml
danworth: Yes, don't worry about being picky, I'd rather you were.

src/test/java/uk/gov/pay/connector/charge/util/JwtGeneratorTest.java
danworth: Yeah 👍

alexbishop1 (approved):

PR #1603, opened 2019-08-15
PP-5544 Include 3ds_integration_version in gateway account responses

stephencdaly:
Modify GatewayAccountEntity so that the value of the integration_version_3ds column is retrieved from and set in the database, giving it a default value of 1.

Include the value in gateway account API responses with the key "3ds_integration_version".

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
@@ -42,13 +42,12 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 @SequenceGenerator(name = "gateway_accounts_gateway_account_id_seq",
         sequenceName = "gateway_accounts_gateway_account_id_seq", allocationSize = 1)
 public class GatewayAccountEntity extends AbstractVersionedEntity {
-    
+
     @Id
     @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "gateway_accounts_gateway_account_id_seq")
     @JsonIgnore
     private Long id;
 
-    //TODO: Should we rename the columns to be more consistent?
     @Column(name = "payment_provider")
     private String gatewayName;
 
@@ -92,6 +91,9 @@ public class GatewayAccountEntity extends AbstractVersionedEntity {
 
     @Column(name = "allow_zero_amount")
     private boolean allowZeroAmount;
+    
+    @Column(name = "integration_version_3ds")
+    private int integrationVersion3ds;
 
     @Column(name = "notify_settings", columnDefinition = "json")
     @Convert(converter = JsonToMapConverter.class)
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -92,6 +91,9 @@
 
     @Column(name = "allow_zero_amount")
     private boolean allowZeroAmount;
+    
+    @Column(name = "integration_version_3ds")
+    private int threeDSIntegrationVersion = 1;
```
oswaldquek: isn't the default already 1 in the database?

stephencdaly: yeh, but if this isn't a nullable field it will be initialised with a value of 0 won't it?

oswaldquek: i think this class is only ever instantiated by the JPA when looking up a gateway account entity from the database so it should be initialised with 1 by default.

oswaldquek: best way to find is remove the assignment to 1, and the ResourceIT tests in this PR should pass!

stephencdaly: Yeh it is required, because we create one of these when we insert a new gateway account.
I need to alter the tests a bit so they will fail if it isn't assigned to 1, so I'm going to force push some more changes (that include renaming variables). 
But I've tried and the assignment is required.

oswaldquek: Which test(s) fail if there is no assignment to 1?

stephencdaly: Just `GatewayAccountFrontendResourceIT.shouldGetGatewayAccountForExistingAccount`

oswaldquek: Hmm I guess one could also set `gatewayAccountEntity.setIntegrationVersion3ds(1);` in the `GatewayAccountObjectConverter.createEntityFrom`. I think this is preferred but I'll leave it to you to decide which is better.

oswaldquek (dismissed):

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
stephencdaly: Yeh I think you're right, it makes more sense to do it there

stephencdaly: changed

oswaldquek (approved):

PR #1611, opened 2019-08-16
PP-5402-Modify domain objects and JSON serialization/deserialization classes-2

nimalank7:
## WHAT YOU DID

This PR modifies the TelephoneChargeRequest, TelephoneChargeResponse, ChargeEntity, CardDetailsEntity, PaymentOutcome and State by adding constructors, setters and changing from primitive type integer to the object type of Long.

This also modifies the TelephonePaymentJSONTest to pass in Long instead of int.

We can map amount, reference, description, status, email and cardDetails from the JSON into the ChargeEntity fields so we will need a constructor to take these. Any other remaining fields in the JSON will have to go into ExternalMetadata.

For constructing the TelephonePaymentResponse we need 2 constructors for State. This is because for success we don't want to return a code but with failure we do.

Since the Supplemental object in the JSON is optional then we need 2 constructors in PaymentOutcome to handle this (e.g. one that takes it one that doesn't). Since status in PaymentOutcome is always required in the JSON and since code won't be provided with status success we need a PaymentOutcome constructor that just takes in status.

Please review https://github.com/alphagov/pay-connector/pull/1610 before merging this PR in.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/CardDetailsEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/CardDetailsEntity.java
@@ -43,6 +43,15 @@ public class CardDetailsEntity {
 
     public CardDetailsEntity() {
     }
+    
+    // For telephone payments
+    public CardDetailsEntity(LastDigitsCardNumber lastDigitsCardNumber, FirstDigitsCardNumber firstDigitsCardNumber, String cardHolderName, String expiryDate, String cardBrand) {
+        this.lastDigitsCardNumber = lastDigitsCardNumber;
+        this.firstDigitsCardNumber = firstDigitsCardNumber;
+        this.cardHolderName = cardHolderName;
+        this.expiryDate = expiryDate;
+        this.cardBrand = cardBrand;
+    }
 
     public CardDetailsEntity(FirstDigitsCardNumber firstDigitsCardNumber, LastDigitsCardNumber lastDigitsCardNumber, String cardHolderName, String expiryDate, String cardBrand, AddressEntity billingAddress) {
         this.lastDigitsCardNumber = lastDigitsCardNumber;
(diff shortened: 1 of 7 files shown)


nimalank7:
Rebased from updated master.

src/main/java/uk/gov/pay/connector/charge/model/CardDetailsEntity.java
```diff
@@ -43,6 +43,15 @@
 
     public CardDetailsEntity() {
     }
+    
+    // For telephone payments
```
mrlumbu, line 47: I wouldn't bother with this sort of comment - it doesn't add to any understanding, and is likely to become stale - maybe some other thing will be built that also needs this constructor in the future

src/main/java/uk/gov/pay/connector/charge/model/telephone/State.java
```diff
@@ -13,6 +13,12 @@
     public State() {
         // For Jackson serialization
     }
+    
+    public State(String status, Boolean finished, String message) {
```
mrlumbu, line 17: it would be good to understand why these changes are necessary - your PR describes what you have done, but not the rationale, which makes it hard to evaluate some of your choices

mrlumbu (changes requested):

src/main/java/uk/gov/pay/connector/charge/model/CardDetailsEntity.java
nimalank7: Sure I'll remove it.

src/main/java/uk/gov/pay/connector/charge/model/telephone/State.java
nimalank7: Apologies - for constructing the TelephonePaymentResponse I'll need a constructor which have these parameters. Similarly for persisting to the database I'll need to be able to store stuff from the JSON into the ExternalMetadata. I thought it'd be easier to just group these changes together. I'll update the PR description accordingly.

nimalank7: Yes - I've updated the PR description accordingly. The constructors are required for certain fields in the JSON being optional. The alternative I suppose is to use setters instead.

mrlumbu (approved):

PR #1612, opened 2019-08-16
Pp 5547 add end point for worldpay jwt

danworth:
PP-5547 Add end-point to get worldpay 3ds flex ddc jwt
    
    - Add end-point within `ChargesFrontendResource` for getting a worldpay
    3ds flex ddc jwt.
    - Add integration tests for the new end-point in new file named
    `ChargesFrontendResourceWorldpayJwtIT` since `ChargesFrontendResourceIT`
    is very large
    - Introduced a Worldpay3dsFlexJwtService and moved a worldpay method out
    of JwtGenerator (includes tests for the new service).
    - Introduce separate exceptions for when credentials are invalid or the
    payment provider is not Worldpay since we want an alert on the former.
    Both map to a new `ConflictWebApplicationException` class so that we can use a single
    `ConflictWebApplicationExceptionMapper` to generate the 409 response (this can also be
    used elsewhere going forward).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtService.java
@@ -14,8 +14,6 @@ import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
-import static java.time.temporal.ChronoUnit.SECONDS;
-
 public class Worldpay3dsFlexJwtService {
 
     private final JwtGenerator jwtGenerator;
@@ -69,8 +67,8 @@ public class Worldpay3dsFlexJwtService {
     private Map<String, Object> generateDdcClaims(String issuer, String organisationId, ZonedDateTime chargeCreatedTime) {
         return Map.of(
                 "jti", RandomIdGenerator.newId(),
-                "iat", Instant.now().toEpochMilli(),
-                "exp", chargeCreatedTime.plus(tokenExpiryDurationSeconds, SECONDS).toInstant().toEpochMilli(),
+                "iat", Instant.now().getEpochSecond(),
+                "exp", chargeCreatedTime.plusSeconds(tokenExpiryDurationSeconds).toInstant().getEpochSecond(),
                 "iss", issuer,
                 "OrgUnitId", organisationId);
     }
--- a/src/main/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtService.java
@@ -1,0 +1,74 @@
+package uk.gov.pay.connector.charge.service;
+
+import uk.gov.pay.connector.charge.exception.Worldpay3dsFlexDdcJwtCredentialsException;
+import uk.gov.pay.connector.charge.exception.Worldpay3dsFlexDdcJwtPaymentProviderException;
+import uk.gov.pay.connector.charge.util.JwtGenerator;
+import uk.gov.pay.connector.gateway.PaymentGatewayName;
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtService.java
```diff
@@ -0,0 +1,74 @@
...
+
+        var claims = generateDdcClaims(issuer, organisationId);
+        return jwtGenerator.createJwt(claims, jwtMacId);
+    }
+
+    private Map<String, Object> generateDdcClaims(String issuer, String organisationId) {
+        return Map.of(
+                "jti", RandomIdGenerator.newId(),
+                "iat", Instant.now().toEpochMilli(),
+                "exp", Instant.now().plus(90, MINUTES).toEpochMilli(),
```
stephencdaly: Should this use the expiry threshold (defaultChargeExpiryThreshold) from the config?

And should it also use the charge created date rather than now? Probably not the biggest deal if the JWT expires later than the charge though...

stephencdaly (commented):
Looks good, just a question about the JWT expiry time

src/main/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtService.java
danworth: It should be from the payment created time so I'll update it, thanks very much, that's a good spot!

danworth: I think its also a good idea to use the `defaultChargeExpiryThreshold` so its in sync with the charge status so I'll do that too.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
```diff
@@ -49,7 +51,7 @@ public void shouldCreateCorrectTokenForWorldpay3dsFlexDdc() {
         assertThat(jws.getHeader().get("typ"), is("JWT"));
         assertThat(jws.getBody().get("jti"), is (notNullValue()));
         assertThat(jws.getBody().get("iat"), is (notNullValue()));
-        assertThat(jws.getBody().get("exp"), is (notNullValue()));
+        assertThat(jws.getBody().get("exp"), is (paymentCreationTime.plusSeconds(tokenExpiryDurationSeconds).toInstant().toEpochMilli()));
```
stephencdaly: Maybe it would be better to assert on an actual value here so we're not performing the logic under test in the test. Although maybe that doesn't add loads of value in this case.

stephencdaly (dismissed):
Looks good, left a comment about the test but up to you if you feel it's something to change or not.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
danworth: I felt that the logic was simple and used `java.time` libraries, but its a very valid point. How about this for creating the expected expiry time without performing the same calculation in the method itself?
`var expectedTokenExpiryTime = paymentCreationTime.toInstant().toEpochMilli() + (tokenExpiryDurationSeconds * 1000);`

stephencdaly: I was more thinking if we have an actual epoch millisecond value it would test that we hadn't somehow outputted something of completely the wrong format then copy-pasted that into the tests in future.

But as I said not a big deal in this case, because as you say we're using the java.time libraries.

stephencdaly (approved):
Looks good 👍

PR #1614, opened 2019-08-17
PP-5402 Creates methods in ChargeService to persist TelephoneCharge to DB-3

nimalank7:
## WHAT YOU DID

This PR creates methods in the ChargeService to allow a telephone charge to be persisted into the database.

The overloaded findCharge method takes in a telephoneChargeRequest and searches the charge database via its processor_id if it already exists. If it exists then it returns the ChargeEntity wrapped in an Optional and then uses it to call the populateTelephoneCharge method which builds the ChargeResponse class used for serialization. If it doesn't exist then the method returns an empty Optional.

The overloaded createChargeEntity method takes the telephoneChargeRequest and account_id. It then throws an error if the amount is 0 and if the gatewayAccount doesn't allow for this. Otherwise a CardDetailsEntity is created pulling out the fields from the request and this is passed into the ChargeEntity. Amount, reference, description, email_address, card_details and provider_id are pulled from this. The charge status is determined from the code (or lack thereof) using the internalChargeStatus, lack of code = AUTHORISATION_SUCCESS, P0010 = AUTHORISATION_REJECTED and P0030 = AUTHORISATION_ERROR. (Validation in a later PR will enforce the requirement that only P0010 and P0030 are allowed as error codes). Any additional fields are put into external metadata which is handled by the storeExtraFieldsInMetaDataForTelepehonePayments.

The populateResponseBuilderWith method builds the ChargeResponse using a builder. Supplemental is an optional field in the JSON and the remaining if conditions handle cases for status = failed with/without Supplemental.

I modified the AbstractChargeResponseBuilder to include the extra fields for telephone payments. So AbstractChargeResponseBuilder builds both card and telephone charges. I also modified ChargeResponse to include these extra fields.

Payment_id is stored in charge_id, card details are stored in the card_details. State is stored in ExternalTransactionState. Publicapi will then extract this out and return it as separate fields in the JSON sent back to the service.

## How to test

The ChargeService test below.


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -120,6 +129,69 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
         this.shouldEmitPaymentStateTransitionEvents = config.getEmitPaymentStateTransitionEvents();
         this.eventQueue = eventQueue;
     }
+    
+    @Transactional
+    public Optional<TelephoneChargeResponse> findTelephoneCharge(TelephoneChargeCreateRequest telephoneChargeRequest, Long accountId, UriInfo uriInfo) {
```
mrlumbu: not sure what the purpose of this method is?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -120,6 +129,69 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
...
+    
+    public Optional<TelephoneChargeResponse> createTelephoneCharge(TelephoneChargeCreateRequest telephoneChargeCreateRequest, Long accountId, UriInfo uriInfo) {
+
+        return createTelephoneChargeEntity(telephoneChargeCreateRequest, accountId, uriInfo)
+                .map(charge ->
+                        populateTelephoneCharge(charge).build());
+    }
+
+    @Transactional
+    private Optional<ChargeEntity> createTelephoneChargeEntity(TelephoneChargeCreateRequest telephoneChargeRequest, Long accountId, UriInfo uriInfo) {
```
mrlumbu: I think the name should just be createCharge, there is nothing special about a telephone charge once it is created. The method will be clear enough because of the argument types

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -212,6 +284,77 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
...
 
+    private TelephoneChargeResponse.ChargeBuilder populateTelephoneCharge(ChargeEntity chargeEntity) {
+
+        final TelephoneChargeResponse.ChargeBuilder telephoneChargeResponse = new TelephoneChargeResponse.ChargeBuilder()
+                .amount(chargeEntity.getAmount())
+                .reference(chargeEntity.getReference().toString())
+                .description(chargeEntity.getDescription())
+                .createdDate((String) chargeEntity.getExternalMetadata().get().getMetadata().get("created_date"))
+                .authorisedDate((String) chargeEntity.getExternalMetadata().get().getMetadata().get("authorised_date"))
+                .processorId((String) chargeEntity.getExternalMetadata().get().getMetadata().get("processor_id"))
```
mrlumbu: what if the Optional external metadata is empty?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -618,4 +761,31 @@ private URI nextUrl() {
     private boolean chargeIsInLockedStatus(OperationType operationType, ChargeEntity chargeEntity) {
         return operationType.getLockingStatus().equals(ChargeStatus.fromString(chargeEntity.getStatus()));
     }
+    
+    private ExternalMetadata storeExtraFieldsInMetaDataForTelephonePayments(TelephoneChargeCreateRequest telephoneChargeRequest) {
```
mrlumbu: again, no need to reference telephone in method name - will be clear enough from types.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -120,6 +129,69 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
...
+        
+        if(chargeEntity.isEmpty()) {
+            return Optional.empty();
+        }
+        
+        return chargeEntity.map(charge ->
+                populateTelephoneCharge(charge).build());
+    }
+    
+    public Optional<TelephoneChargeResponse> createTelephoneCharge(TelephoneChargeCreateRequest telephoneChargeCreateRequest, Long accountId, UriInfo uriInfo) {
```
mrlumbu: Do we definitely need to return a `TelephoneChargeResponse` as opposed to a normal `ChargeResponse`? I'm a bit concerned that you are creating loads of new objects that look a lot like existing things

mrlumbu (changes requested):
I think it would be good to have a quick chat about this, just so I understand the direction you're heading in. It maybe that it is necessary to separate the Telephone payment path as much as you have done, but I'm not totally convinced. I have also made comments about naming and unsafe use of Optionals that need addressing

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -120,6 +124,63 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
...
+        return createChargeEntity(telephoneChargeCreateRequest, accountId, uriInfo)
+                .map(charge ->
+                        populateResponseBuilderWith(aChargeResponseBuilder(), charge).build());
+    }
+
+    @Transactional
+    private Optional<ChargeEntity> createChargeEntity(TelephoneChargeCreateRequest telephoneChargeRequest, Long accountId, UriInfo uriInfo) {
+        return gatewayAccountDao.findById(accountId).map(gatewayAccount -> {
+
+            if (telephoneChargeRequest.getAmount() == 0L && !gatewayAccount.isAllowZeroAmount()) {
```
mrlumbu: does this logic not exist somewhere already?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Refactored into ChargeResponse

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Added ifPresent to handle case.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Renamed method.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Used for idempotency in ChargesApiResource.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Refactored into checkIfZeroAmountAllowed method.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -120,6 +124,61 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
...
+    
+    public Optional<ChargeResponse> create(TelephoneChargeCreateRequest telephoneChargeCreateRequest, Long accountId, UriInfo uriInfo) {
+
+        return createCharge(telephoneChargeCreateRequest, accountId, uriInfo)
+                .map(charge ->
+                        populateResponseBuilderWith(aChargeResponseBuilder(), charge).build());
+    }
+
+    @Transactional
+    private Optional<ChargeEntity> createCharge(TelephoneChargeCreateRequest telephoneChargeRequest, Long accountId, UriInfo uriInfo) {
```
mrlumbu: I've just realised - you are passing this uriInfo all the way down then not using it

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -120,6 +124,61 @@ public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao char
...
+
+            chargeDao.persist(chargeEntity);
+            return chargeEntity;
+        });
+    }
+
+    private ChargeStatus internalChargeStatus(String code) {
+        if(code == null) {
+            return AUTHORISATION_SUCCESS;
+        } else if (code.equals("P0010")) {
```
mrlumbu: as Codacy says, the literal should come first in this. Admittedly in this case you have checked null first, but still good habit to get in to

nimalank7: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Removed unused parameters.

mrlumbu (dismissed):

mrlumbu (approved):

PR #1621, opened 2019-08-19
PP-5545: Add additional3DSData element for 3DS2 Flex

oswaldquek:
Test the `additional3DSData` element exists when the
`worldpay_3ds_flex_ddc_result` property (in AuthCardDetails) is present.

WorldpayPaymentProviderTest looks a bit messy and could use some refactoring but I'll do this in another PR.


stephencdaly (approved):
Nice 👌

PR #1627, opened 2019-08-20
Pp 5545: session and shopper elements for 3ds flex auth request

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
```diff
@@ -184,4 +195,54 @@ public WorldpayOrderRequestBuilder withPaResponse3ds(String paResponse) {
...
+            month = convert(date.getMonthOfYear());
+            year = String.valueOf(date.getYear());
+        }
+
+        private String convert(int number) {
+            if (number < 10) {
+                return "0" + number;
+            } else {
+                return String.valueOf(number);
+            }
```
stephencdaly: how about `String.format ("%02d", number);`?

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -44,6 +44,17 @@
...
             </#if>
             <#if authCardDetails.worldpay3dsFlexDdcResult.isPresent()>
+            <riskData>
+                <authenticationRiskData authenticationMethod="localAccount">
+                    <authenticationTimestamp><date second="${authenticationDate.getSecond()}"
+                                                   minute="${authenticationDate.getMinute()}"
+                                                   hour="${authenticationDate.getHour()}"
+                                                   dayOfMonth="${authenticationDate.getDay()}"
+                                                   month="${authenticationDate.getMonth()}"
+                                                   year="${authenticationDate.getYear()}"/>
```
stephencdaly: Is the AuthenticationDate type a little overkill?
Is it much different to doing `String.format("%0d", dateTime.getSecondOfMinute)` here?

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -44,6 +44,17 @@
             </shopper>
             </#if>
             <#if authCardDetails.worldpay3dsFlexDdcResult.isPresent()>
+            <riskData>
+                <authenticationRiskData authenticationMethod="localAccount">
```
alexbishop1: `authenticationMethod="localAccount"` means “you have authenticated the shopper using your own systems”, which I think means “the user logged in”. However, this isn’t necessarily the case (some services will require the user to log in but others won’t), so I don’t think we can include this attribute. I’m not actually not sure we should be including the `<authenticationTimestamp>` either, since even if the user did not log in, we don’t know when they did…

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -225,6 +233,24 @@ public void shouldInclude3DS2FlexElementsWhenWorldpay3dsFlexDdcResultIsPresent()
                 is("390x400"));
         assertThat(xPath.evaluate("/paymentService/submit/order/additional3DSData/@challengePreference", document),
                 is("noPreference"));
+        assertThat(xPath.evaluate("/paymentService/submit/order/riskData/authenticationRiskData/authenticationTimestamp/date/@second", document),
+                isValidSecond());
```
stephencdaly: Could we not check against actual values for the date we're passing in rather than requiring the custom matcher?

stephencdaly (changes requested):
Looks like it does what's required but I've left a few comments

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
oswaldquek: oh yeah this is much nicer. however, we aren't going to supply the risk data element anymore so problem solved!

stephencdaly (dismissed):

stephencdaly (approved):
Looks good

PR #1632, opened 2019-08-21
PP-5520 Log npe detail

mrlumbu:
We're getting npes and struggling to diagnose. Explicitly log npe stack trace with searchable prefix to all better understanding of what's going on. (This is obviously temporary)


sfount (approved):
> (This is obviously temporary) 

PR can be reverted once backfill has been processed

PR #1634, opened 2019-08-21
PP-5520 Check null card details

mrlumbu:
From running historical event emitter we've seen PaymentDetailsEnteredEventDetails creation fail because CardDetails is null. Check this before dereferencing.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorkerTest.java
+++ b/src/test/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorkerTest.java
@@ -247,9 +247,10 @@ public class HistoricalEventEmitterWorkerTest {
         assertThat(argument.getAllValues().get(1).getStateTransitionEventClass(), is(CaptureConfirmed.class));
 
         ArgumentCaptor<Event> daoArgumentCaptor = ArgumentCaptor.forClass(Event.class);
-        verify(emittedEventDao, times(2)).recordEmission(daoArgumentCaptor.capture());
+        verify(emittedEventDao, times(3)).recordEmission(daoArgumentCaptor.capture());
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorkerTest.java
```diff
@@ -247,9 +247,10 @@ public void executeShouldOfferOutOfOrderCaptureStatesInOrder() {
         assertThat(argument.getAllValues().get(1).getStateTransitionEventClass(), is(CaptureConfirmed.class));
 
         ArgumentCaptor<Event> daoArgumentCaptor = ArgumentCaptor.forClass(Event.class);
-        verify(emittedEventDao, times(2)).recordEmission(daoArgumentCaptor.capture());
+        verify(emittedEventDao, times(3)).recordEmission(daoArgumentCaptor.capture());
         assertThat(daoArgumentCaptor.getAllValues().get(0).getEventType(), is("CAPTURE_SUBMITTED"));
         assertThat(daoArgumentCaptor.getAllValues().get(1).getEventType(), is("CAPTURE_CONFIRMED"));
+        assertThat(daoArgumentCaptor.getAllValues().get(2).getEventType(), is("PAYMENT_DETAILS_ENTERED"));
     }
```
mrlumbu, line 254: These changes are because the bug actually affected the result of this test - as you might imagine, we should actually expect 3 events to be emitted if there are 3 events associated with the charge. Previously the thrid one was not being emitted because an NPE was being thrown

sfount (approved):
Good catch with the `NPE` test! 🤦‍♂ 

This looks a lot better than it was, wondering if there's a cleaner pattern with the level of dealing with optional results but don't have a suggestion and looks like this works fine.

PR #1637, opened 2019-08-22
PP-5566 add a gauge to monitor the size of in-memory queue

jankowiakmaria:
* extend StateTransitionQueue wit a size method
* add StateTransitionQueueMetricEmitter
* add registering an emitter in StateTransitionEmitterProcess
* update test


sfount (approved):
🍠

PR #1640, opened 2019-08-23
PP-5545 Update XML 3DS Logic

rorymalcolm:
- Updates XML logic to ensure 3DS is required before the result of 3DS
Flex is displayed.

### DIFF
--- a/src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
+++ b/src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
@@ -43,11 +43,10 @@
                 </browser>
             </shopper>
             <#if authCardDetails.worldpay3dsFlexDdcResult.isPresent()>
-                <additional3DSData
+            <additional3DSData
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -42,13 +42,13 @@
                     <userAgentHeader>${authCardDetails.userAgentHeader?xml}</userAgentHeader>
                 </browser>
             </shopper>
-            </#if>
             <#if authCardDetails.worldpay3dsFlexDdcResult.isPresent()>
-            <additional3DSData
+                <additional3DSData
```
alexbishop1: Can you un-indent this? We don’t usually put indents after `#if`s because it interferes with the XML indentation.

alexbishop1 (changes requested):
This looks good — just one thing about the indentation

alexbishop1 (approved):

PR #1641, opened 2019-08-23
PP-5520 Schedule multiple state transition threads

sfount:
* thread pool size is fixed to an environment variable
* only one thread was being scheduled for execution with a `Callable`
* loop through thread pool size and schedule all threads for execution


mrlumbu (approved):

sfount:
Seems to work as expected in test, reviewing `statsd` performance in staging before pushing to production -- looks like similar performance tweaks will have to be made to ledger consuming events.

PR #1651, opened 2019-08-28
PP-5600: don't re-throw when there is a problem with amazon SQS

jankowiakmaria:
* not being able to emit event shouldn't affect the payment journey (at the moment it does)
* should be thought of how to handle the cases when there is an error (to be able to emit it in the future) - part of parity check probably

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -52,14 +52,13 @@ import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 import uk.gov.pay.connector.queue.PaymentStateTransition;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -455,7 +454,6 @@ private void emitEvent(Event event) {
             eventQueue.emitEvent(event);
         } catch (QueueException e) {
             logger.error("Error emitting {} event: {}", event.getEventType(), e.getMessage());
```
sfount, line 456: If this log line was specific (something like "In-flight payment emitting error") we can ZenDesk alert if we ever see this.

sfount: EDIT: looks like it's unique

sfount (commented):
Code change looks good! We should create the ZenDesk log alert before merging the code

+ we should link this to a JIRA ticket so that it can be referenced from any review actions

jankowiakmaria:
Splunk alert has been merged and released: https://github.com/alphagov/cyber-security-splunk-apps/pull/235

sfount (approved):
Nice, great work!

PR #1659, opened 2019-08-30
PP-5598 Emit PaymentDetailsEntered Event only once

kbottla:
## WHAT YOU DID
- If chargeEvents has both AUTHORISATION_SUCCESS and AUTHORISATION_3DS_REQUIRED events, then emit PaymenyDetailsEntered only once for AUTHORISATION_3DS_REQUIRED event

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
@@ -256,10 +256,24 @@ public class HistoricalEventEmitterWorker {
         // checking against any terminal authentication transition
         chargeEventEntities
                 .stream()
-                .filter(event -> TERMINAL_AUTHENTICATION_STATES.contains(event.getStatus()))
+                .filter(event -> isValidPaymentDetailsEnteredTransition(chargeEventEntities, event))
                 .map(PaymentDetailsEntered::from)
                 .filter(event -> !emittedEventDao.hasBeenEmittedBefore(event))
                 .forEach(this::emitAndPersistEvent);
+    }
+
+    private boolean isValidPaymentDetailsEnteredTransition(List<ChargeEventEntity> chargeEventEntities, ChargeEventEntity event) {
+        return TERMINAL_AUTHENTICATION_STATES.contains(event.getStatus())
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
```diff
@@ -257,11 +257,25 @@ private void processPaymentDetailEnteredEvent(List<ChargeEventEntity> chargeEven
         chargeEventEntities
                 .stream()
                 .filter(event -> TERMINAL_AUTHENTICATION_STATES.contains(event.getStatus()))
+                .filter(event -> isNotAuthSuccessEventOrAuthSuccessEventWithOut3DSEvent(chargeEventEntities, event))
```
sfount: I think now that there are 3 nested checks it might read a lot more clearly to call these methods by what they mean rather than what they do. 

For example: 
```java 
chargeEventEntities
    .stream()
    .filter(event -> isValidPaymentDetailsEnteredTransition(chargeEventEntities, event))
    .map(PaymentDetailsEntered::from)
    .filter(event -> !emittedEventDao.hasBeenEmittedBefore(event))
    .forEach(this::emitAndPersistEvent);
```

Then the more complex logic could be conveyed with: 
```java 
private boolean isValidPaymentDetailsEnteredTransition(List<ChargeEventEntity> chargeEvents, ChargeEventEntity currentEvent) { 
    return TERMINAL_AUTHENTICATION_STATES.contains(currentEvent.getStatus()) &&
        isNotDuplicatePaymentDetailsEvent() && 
        // ...
}
```

sfount (dismissed):
The logic here makes sense to me -- I've left a comment as a potential readability improvement as this code looks like it will be used to recover from in-flight incidents. Up to you if that will be clearer or should be done now!

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
kbottla: 👍 Updated https://github.com/alphagov/pay-connector/compare/87611c3cb407bd70dfb54cc1a8301b6658565f12..3bdf506dca51092e875f9dac4e62df202d513ab6

sfount (approved):

PR #1660, opened 2019-08-30
PP-5402 - Renames Builder method and directly tests @NotNull error message

nimalank7:
## WHAT YOU DID
This PR renames TelephoneCreateChargeRequest.ChargeBuilder to TelephoneCreateChargeRequest.ChargeBuilder.

This also ensures that only required fields are sent for the validator in the unit tests.

This also removes the previous assertion for testing null values in the unit validator tests to instead check the error message is equal to "Field [...] cannot be null."


jankowiakmaria (approved):
The only thing I don't like is the naming of `passesValidationForNull<fieldName>`. Should really be `fails`.

PR #1669, opened 2019-09-02
PP-5611 Add New Pact State For Exact Ref Match

rorymalcolm:
- Adds a new pact state for the exact match of a reference search


kbottla (approved):
LGTM!

danworth (approved):
Looks good 👍

PR #1670, opened 2019-09-02
Pp 5610 remove country length validation

danworth:
The requirements when setting country for an address of a
    `PrefilledCardHolderDetails` is different to when setting the card payment
    details via frontend. It is believed frontend can send 3 character long
    countries representing a territory and which should already have been
    validated, whilst pre-filled could originate from unvalidated user input
    and should set anything that is not a
    3166-1 alpha-2 code to null so that the default GB will be used.
    
    This introduces a new class 'PrefilledAddress'. If the length of the country
    is greater than two then it is dismissed and null used instead; this is
    the behaviour described in our api documentation. This validation can be
    improved upon to look whether the two characters represent a valid
    3166-1 alpha-2 code but to preserve existing logic (where we only check
    for two characters) we will consider that later.
    
    We considered subclassing `Address` and modifying the behavior for `country`
    but since the valdiation using
    `Address` from Frontend does not use javax.validation we decided to
    create a new class so that it is clear the validation annotations do not
    apply in all cases; also there was some debate whether an address for
    prefilled and non-prefilled are strictly an is-a relationship.


mrlumbu (approved):
Seems like a sensible solution 👍

PR #1673, opened 2019-09-02
PP-5605: refactor HistoricalEventEmitterWorker

jankowiakmaria:
extract emitting from Worker to HistoricalEventEmitter

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
@@ -1,0 +1,240 @@
+package uk.gov.pay.connector.tasks;
+
+import com.google.inject.Inject;
+import com.google.inject.persist.Transactional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
+import uk.gov.pay.connector.common.model.domain.PaymentGatewayStateTransitions;
+import uk.gov.pay.connector.events.EventQueue;
+import uk.gov.pay.connector.events.dao.EmittedEventDao;
+import uk.gov.pay.connector.events.model.Event;
+import uk.gov.pay.connector.events.model.EventFactory;
+import uk.gov.pay.connector.events.model.charge.PaymentDetailsEntered;
+import uk.gov.pay.connector.queue.PaymentStateTransition;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -0,0 +1,240 @@
...
+
+    public void processPaymentEvents(ChargeEntity charge) {
+        List<ChargeEventEntity> chargeEventEntities = getSortedChargeEvents(charge);
+
+        processChargeStateTransitionEvents(charge.getId(), chargeEventEntities);
+        processPaymentDetailEnteredEvent(chargeEventEntities);
+    }
+
+    @Transactional
+    public void processRefundEvents(ChargeEntity charge) {
```
kbottla, line 91: This is also in `HistoricalEventEmitterWorker`

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterWorker.java
```diff
@@ -7,90 +7,37 @@
...
 
 public class HistoricalEventEmitterWorker {
     private static final Logger logger = LoggerFactory.getLogger(HistoricalEventEmitterWorker.class);
     private static final int PAGE_SIZE = 100;
     private final ChargeDao chargeDao;
     private final ChargeEventDao chargeEventDao;
-    private final EmittedEventDao emittedEventDao;
-    private StateTransitionQueue stateTransitionQueue;
-    private final EventQueue eventQueue;
+    private HistoricalEventEmitter historicalEventEmitter;
```
kbottla, line 28: Looks good to have the emitter separately so it can be reused. Couple of things I have noticed
- I think we could do naming methods more clearly. We have used `process*` excessively in both classes which makes it confusing for readability and following the code. 
- When `HistoricalEventEmitter` is used to emit events for a charge, it looks into `EmittedEvents` table. So if we use the same during parity checking, events won't get emitted if `EmittedEvents` has an entry already (but ledger failed to receive or process them for some reason).

jankowiakmaria: * agree with that - but I wanted to only move things to separate class initially
* I know (but good point obviously) - it will be addressed separately

kbottla (dismissed):
Happy for comments to be addressed separately

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
jankowiakmaria: good point - must have missed it (I've done refactoring and I had to re-do it after merging your changes) - will address it now (see the new commit)

kbottla (approved):

PR #1675, opened 2019-09-03
PP-5605: database migration

jankowiakmaria:
add parity_check_status and parity_check_date to the charges table


kbottla (approved):
LGTM!

PR #1676, opened 2019-09-03
PP-5605: add parity check fields to Charge entity

jankowiakmaria:
* add parity_check_status and parity_check_date to charge entity
* updated the fixture
* added integration test for persistance verification


mrlumbu (approved):
LGTM 👍

PR #1677, opened 2019-09-03
PP-5402 Bug Fix Nested Metadata

nimalank7:
## WHAT YOU DID

External metadata should not be storing nested objects. I've refactored populateResponseBuilder and storeExtraFieldsInMetaData to store it as name-value pairs instead. All tests have been appropriately modified.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -754,22 +754,10 @@ public class ChargeService {
         telephoneJSON.put("authorised_date", telephoneChargeRequest.getAuthorisedDate());
         telephoneJSON.put("processor_id", telephoneChargeRequest.getProcessorId());
         telephoneJSON.put("auth_code", telephoneChargeRequest.getAuthCode());
-        
-        HashMap<String, Object> paymentOutcome = new HashMap<>();
-        paymentOutcome.put("status", telephoneChargeRequest.getPaymentOutcome().getStatus());
-        
-        if(telephoneChargeRequest.getPaymentOutcome().getCode() != null) {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -286,40 +286,34 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
                 .withChargeId("dummypaymentid123notpersisted");
                 
         chargeEntity.getExternalMetadata().ifPresent(externalMetadata -> {
-
-            final Map<String, Object> paymentOutcomeMap = ((Map) externalMetadata.getMetadata().get("payment_outcome"));
-
+            
             final PaymentOutcome paymentOutcome = new PaymentOutcome(
-                    paymentOutcomeMap.get("status").toString()
+                    externalMetadata.getMetadata().get("status").toString()
```
mrlumbu, line 291: is status guaranteed to be in the Map at this point?

nimalank7: Yes. Status is required and PaymentOutcome has custom validation checking if it is present.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -286,40 +286,34 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
...
             );
             
+            final Supplemental supplemental = new Supplemental();
```
mrlumbu: this doesn't appear to be used

nimalank7: Fixed.

mrlumbu (approved):

PR #1680, opened 2019-09-04
PP-5613 Enforce ExternalMetadata Validation when saving ChargeEntity

danworth:
We evaluate the `@ValidExternalMetadata` when accepting a create charge
request via an end-point however this is not evaluated when attempting
to persist a `ChargeEntity`. This means it is possible to create an
invalid instance of `ExternalMetadata` within Connector and then persist
it to the charges table. This change adds the `@Valid` annotation to the
`externalMetadata` attribute of `ChargeEntity` so that any JPA actions
to update or persist the entity will evaluate the validation, rolling
back the transaction if it fails.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -43,6 +43,7 @@ import javax.persistence.OneToOne;
 import javax.persistence.OrderBy;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
+import javax.validation.Valid;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
@@ -139,6 +140,7 @@ public class ChargeEntity extends AbstractVersionedEntity implements Nettable {
 
     @Column(name = "external_metadata", columnDefinition = "jsonb")
     @Convert(converter = ExternalMetadataConverter.class)
+    @Valid
     private ExternalMetadata externalMetadata;
 
     @Column(name = "parity_check_status")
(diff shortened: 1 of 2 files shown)


oswaldquek:
How does an external service cause a ChargeEntity to be saved without going through the create charge API? If it goes through another endpoint, we'll need an integration test for that endpoint so a proper response with clear reason is returned to that service.

oswaldquek (changes requested):
How does an external service cause a ChargeEntity to be saved without going through the create charge API? If it goes through another endpoint, we'll need an integration test for that endpoint so a proper response with clear reason is returned to that service.

danworth:
Thanks Oz, I should have added a bit more context to this, and probably best to chat. 

This came about because the in progress telephone payments that @nimalank7 is working on is re-using the `externalMetadata` field of `ChargeEntity` to store some data. It doesn't come into Connector as an `externalMetadata` field with the `@ValidExternalMetadata` annotation on it, rather elements of a `TelephoneChargeCreateRequest` are deconstructed into an `ExternalMetadata` object [here](https://github.com/alphagov/pay-connector/blob/5833cca87254064231844bb769dc12e3af13ffbb/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java#L744). The problem came about when Dom noticed during some QA of telephone payments that the metadata wasn't being displayed correctly in SelfService which was because it had objects for values, which isn't permitted by the validation. This change is really to close this loop hole and something I should have added when we originally created the `ExternalMetadata` so that it is validated in both API and persistence so they can never become out of sync. 

It's a good point about needing integration tests for this scenario. Nim is aware of this change and has updated telephone payments to conform to the validated form of `ExternalMetadata` e.g. no longer using objects or `null` for values. The response will be a 422 via the `ConstraintViolationMapper` but @nimalank7 will need to add those tests as part of his telephone payment work and probably need to add validation to any element being deconstructed into the `ExternalMetadata` to make sure it can be. 

Lets have a chat with Nim.

danworth:
Hi @nimalank7 for the benefit of a reviewer of this PR can you please confirm that you've made the necessary changes with regard to Telephone payments so that we can progress this PR, e.g. you have added all necessary validation to the elements of a create telephone payment request that will be added to external metadata so the correct error is returned.

nimalank7:
Yep @danworth  I've written validation to enforce this. See the following PR [https://github.com/alphagov/pay-connector/pull/1694]

danworth:
@oswaldquek could you please review again in light of Nim updating his work to make this a safe change.

oswaldquek (approved):

PR #1681, opened 2019-09-04
PP-5611 Remove Reference Like Searching

rorymalcolm:
- Remove use of like comparator inside of reference search, now looks
for an exact match for publicapi
- Retains case isensitivity
- Potential to update use of indexes for better adjustments in
post-fuzzy search world

with @alexbishop1


src/main/java/uk/gov/pay/connector/charge/dao/SearchParams.java
```diff
@@ -28,10 +28,11 @@
     private TransactionSearchStrategyTransactionType transactionSearchStrategyTransactionType;
     private LastDigitsCardNumber lastDigitsCardNumber;
     private FirstDigitsCardNumber firstDigitsCardNumber;
-    private CardHolderName cardHolderName;
+    private CardHolderName likeCardHolderName;
     private Long gatewayAccountId;
     private ServicePaymentReference reference;
-    private String email;
+    private ServicePaymentReference likeReference;
```
danworth, line 34: I like the addition of `like` to make it clearer 👍

src/main/java/uk/gov/pay/connector/charge/dao/SearchParams.java
```diff
@@ -189,13 +190,22 @@ public ServicePaymentReference getReference() {
...
+        return likeEmail;
     }
 
-    public SearchParams withReferenceLike(ServicePaymentReference reference) {
+    public SearchParams withReference(ServicePaymentReference reference) {
         this.reference = reference;
         return this;
     }
+    
+    public SearchParams withReferenceLike(ServicePaymentReference reference) {
```
danworth, line 201: The attribute is called `likeReference` and the getter is appropriately named `getLikeReference` but I guess you went with this because it reads nicer when its used so 👍

src/main/java/uk/gov/pay/connector/charge/dao/SearchParams.java
```diff
@@ -189,13 +190,22 @@ public ServicePaymentReference getReference() {
     }
 
     public String getEmail() {
-        return email;
+        return likeEmail;
```
danworth, line 193: This method should be renamed `getLikeEmail`?

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoIT.java
```diff
@@ -859,6 +781,57 @@ public void searchChargeByReferenceAndStatusAndToDate_ShouldReturnZeroIfToDateIs
         assertThat(charges.size(), is(0));
     }
 
+    @Test
+    public void searchChargeByPartialReference_ShouldReturnCharge() {
+        insertTestChargeWithReference(ServicePaymentReference.of("rorytest"));
+        insertTestCharge();
+        SearchParams searchParams = new SearchParams()
+                .withReferenceLike(ServicePaymentReference.of("reference"));
```
danworth, line 789: Its a little bit confusing that you're not trying to find the charge that you're setting up in each test, this reads as though you're expecting 'reference' to find 'rory test' which obviously it won't and instead relying on finding the charge inserted elsewhere with the reference of 'Test reference'. Perhaps the charge you set-up within each test could be the one that it actually finds and to test the exclusion it could perhaps not match `Test reference`? Same for the tests below.

danworth (changes requested):
Looks good. If I've understood this correctly then v1 of the charges search api will use an exact reference search, and v2 will continue to use a like reference search. There's quite a lot here but the tests look complete and its been paired on which makes me happy :)

Couple of minor things for your consideration.

src/main/java/uk/gov/pay/connector/charge/dao/SearchParams.java
rorymalcolm: Yeah we went with this because its more coherent to read when you're using the method in our opinion

src/main/java/uk/gov/pay/connector/charge/dao/SearchParams.java
rorymalcolm: Agreed!

danworth (approved):
Yep looks good, been paired on and the tests look to cover the necessary scenarios 👍

PR #1682, opened 2019-09-04
BAU: improve card capture process logging

jankowiakmaria:
* added chargeID to MDC for capture process

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureProcess.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureProcess.java
@@ -2,6 +2,7 @@ package uk.gov.pay.connector.paymentprocessor.service;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 import uk.gov.pay.connector.charge.ChargesAwaitingCaptureMetricEmitter;
 import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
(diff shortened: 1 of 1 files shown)


kbottla (approved):
LGTM!

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureProcess.java
```diff
@@ -36,19 +37,20 @@ public void handleCaptureMessages() throws QueueException {
         List<ChargeCaptureMessage> captureMessages = captureQueue.retrieveChargesForCapture();
         for (ChargeCaptureMessage message : captureMessages) {
             try {
-                LOGGER.info("Charge capture message received - [externalChargeId={}] [queueMessageId={}] [queueMessageReceiptHandle={}]",
-                        message.getChargeId(),
+                MDC.put("chargeId", message.getChargeId());
```
alexbishop1, line 40: As part of the structured logging work, we’re trying to define a consistent set of keys for things like this:

https://github.com/alphagov/pay-java-commons/blob/master/logging/src/main/java/uk/gov/pay/logging/LoggingKeys.java

For this, adding `"external_id"` and also `"resource_type"` with a value of `"payment"` and `"gateway_card_operation"` with a value of `"Capture"` would fit in with the pattern.

(The keys are very new and aren’t being used much yet, so if it requires too much contortion to use the pre-defined keys, we probably want to refine the pre-defined keys.)

alexbishop1: I forgot: more details about the approach taken with the keys can be found in commit https://github.com/alphagov/pay-java-commons/commit/37ca42954cfa19a0649a9df7e0b3b07695ae5fa9.

jankowiakmaria: I might add it separately (unless someone pick up this work first). This is just a quick side PR (while waiting for PR review).
Thanks for the heads up though.

PR #1686, opened 2019-09-04
PP-5585 Stripe capture with payment intents

mrlumbu:
We need to switch over to using Stripe payment intents to process
payments with Stripe, in order to be 3DS2 compliant. In order to
do this in a backwards compatible way we need the capturing with
stripe to work with charges created both via the charges api and
the payment intents api.
This PR uses the fact that stripe namespaces all their ids with
a prefix indicating the resource type (so a payment intent always
has a prefix 'pi_...') so that we can capture the charge
appropriately.
This PR should not affect existing payment flow at all, and will
need to be tested more thoroughly once changes to authorisation
have been implemented.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
@@ -12,12 +12,15 @@ import uk.gov.pay.connector.gateway.model.GatewayError;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
 import uk.gov.pay.connector.gateway.stripe.json.StripeCharge;
 import uk.gov.pay.connector.gateway.stripe.json.StripeErrorResponse;
+import uk.gov.pay.connector.gateway.stripe.json.StripePaymentIntent;
 import uk.gov.pay.connector.gateway.stripe.json.StripeTransferResponse;
-import uk.gov.pay.connector.gateway.stripe.request.StripeCaptureRequest;
+import uk.gov.pay.connector.gateway.stripe.request.StripeChargeCaptureRequest;
+import uk.gov.pay.connector.gateway.stripe.request.StripePaymentIntentCaptureRequest;
 import uk.gov.pay.connector.gateway.stripe.request.StripeTransferOutRequest;
 import uk.gov.pay.connector.gateway.stripe.response.StripeCaptureResponse;
 import uk.gov.pay.connector.util.JsonObjectMapper;
 
+import java.util.List;
 import java.util.Optional;
 
 import static javax.ws.rs.core.Response.Status.Family.CLIENT_ERROR;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -47,22 +49,23 @@ public CaptureResponse capture(CaptureGatewayRequest request) {
         String transactionId = request.getTransactionId();
 
         try {
-            StripeCharge stripeCaptureResponse = captureWithPlatform(request);
-            Optional<Long> maybeFee = Optional.empty();
-
-            if (stripeCaptureResponse.isPlatformCharge()) {
```
mrlumbu, line 53: all charges are now platform charges (this line was a hangover from a while back where we switched the details of how we authorise and capture), so I have removed this condition

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -91,8 +94,25 @@ public CaptureResponse capture(CaptureGatewayRequest request) {
...
+        logger.info("Captured charge id {} with platform account - stripe capture id {}",
+                request.getExternalId(),
+                stripeCharge.getId()
+        );
+
+        return stripeCharge;
+    }
+
+    private boolean usePaymentIntent(String transactionId) {
+        return transactionId.startsWith("pi_");
```
kbottla, line 119: Although remote possibility, this make us rely on Stripe not to change format/prefix for payment intents IDs unless Stripe guarantees it will never change.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -3,6 +3,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
+import uk.gov.pay.connector.events.model.charge.GatewayErrorDuringAuthorisation;
```
kbottla: Unused import

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -91,8 +96,32 @@ public CaptureResponse capture(CaptureGatewayRequest request) {
...
-    private StripeCharge captureWithPlatform(CaptureGatewayRequest request) throws GatewayException.GenericGatewayException, GatewayErrorException, GatewayException.GatewayConnectionTimeoutException {
-        String captureResponse = client.postRequestFor(StripeCaptureRequest.of(request, stripeGatewayConfig)).getEntity();
+    private StripeCharge captureWithPaymentIntentAPI(CaptureGatewayRequest request) throws GatewayException {
+        String captureResponse = client.postRequestFor(StripePaymentIntentCaptureRequest.of(request, stripeGatewayConfig)).getEntity();
+        StripePaymentIntent stripeCaptureResponse = jsonObjectMapper.getObject(captureResponse, StripePaymentIntent.class);
+        List<StripeCharge> stripeCharges = stripeCaptureResponse.getChargesCollection().getCharges();
+        
+        if (stripeCharges.size() != 1) {
+            throw new GatewayErrorException(
+                    String.format("Expected exactly one charge associated with payment intent %s, got %s", request.getTransactionId(), stripeCharges.size()));
```
kbottla, line 105: This might lead to charges capture with Payment provider but in CAPTURE_READY state with us. Might need an alert if this ever happens ?

kbottla (commented):
Looks good overall. Added a few minor comments

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
mrlumbu: To some extent existing alerting will pick this up as we already alert on capture failures. We should make sure that when we create the payment intent it is designated as single use to avoid the possibility of there being more than one charge associated with a payment intent.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
mrlumbu: agree, but this is a very established Stripe convention, and will only be required to ensure backwards compatibility during release, after which we can remove

kbottla (approved):

PR #1687, opened 2019-09-05
PP-5614 Double time allowed to retry pocessing state transition

mrlumbu:
We are seeing very occasional failures to process state transitions e.g. https://govuk.zendesk.com/agent/tickets/3793212

After investigating it seems highly likely this is due to the related transaction failing to complete in time. This PR increases time per retry. The reasoning for increasing time rather than number of attempts is that we are primarily concerned with robustness rather than speed. Increasing time will increase chances of correctness without increasing load on the system.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/StateTransition.java
+++ b/src/main/java/uk/gov/pay/connector/queue/StateTransition.java
@@ -12,7 +12,7 @@ public abstract class StateTransition implements Delayed {
 
     private static final int MAXIMUM_NUMBER_OF_ATTEMPTS = 10;
     private static final int BASE_ATTEMPTS = 1;
-    private static final long DEFAULT_DELAY_DURATION_IN_MILLISECONDS = 100L;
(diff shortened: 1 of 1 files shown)


kbottla (dismissed):
LGTM!

src/main/java/uk/gov/pay/connector/queue/StateTransition.java
```diff
@@ -10,7 +10,7 @@
     private final long delayDurationInMilliseconds;
     private final AtomicInteger attempts;
 
-    private static final int MAXIMUM_NUMBER_OF_ATTEMPTS = 10;
+    private static final int MAXIMUM_NUMBER_OF_ATTEMPTS = 50;
```
jankowiakmaria: That looks more like increase of number of attempts rather than time. (Unless it's time through attempts)

mrlumbu: yes, I hadn't pushed updated branch. Fixed now

jankowiakmaria (approved):

PR #1688, opened 2019-09-05
PP-5605: include charge's refunds when verifying ledger's parity

jankowiakmaria:
* instead of only checking the charge, there will be additional calls to check whether refunds exist as well

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
@@ -12,13 +12,15 @@ import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.events.EventQueue;
 import uk.gov.pay.connector.events.dao.EmittedEventDao;
 import uk.gov.pay.connector.paritycheck.LedgerService;
-import uk.gov.pay.connector.paritycheck.LedgerTransaction;
 import uk.gov.pay.connector.queue.StateTransitionQueue;
 import uk.gov.pay.connector.refund.dao.RefundDao;
+import uk.gov.pay.connector.refund.model.domain.RefundEntity;
 
 import javax.inject.Inject;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
```diff
@@ -75,13 +77,15 @@ public void checkParityFor(long currentId) {
...
-                    logger.info("transaction exists in ledger [id={}]", currentId);
+                List<String> ids = charge.getRefunds().stream().map(RefundEntity::getExternalId).collect(Collectors.toList());
+                ids.add(charge.getExternalId());
+
+                if (existInLedger(ids)) {
+                    logger.info("transaction (and its refunds) exists in ledger [id={}]", currentId);
                     chargeService.updateChargeParityStatus(charge.getExternalId(), ParityCheckStatus.EXISTS_IN_LEDGER);
                 } else {
-                    logger.info("transaction does not exist in ledger [id={}] -", currentId);
+                    logger.info("transaction (or its refunds) does not exist in ledger [id={}] -", currentId);
```
kbottla, line 88: Would be useful to log individual `external_id` if missing. In case of refunds it will be clear which refund is not available in connector

kbottla (approved):

src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
jankowiakmaria: Hm...with the current way if the first refund doesn't exist it will just stop instead of further calling ledger (for remaining refunds and charge). With a filter (that would return a list of missing transactions) it increases number of calls when they are not necessary.
There might be a third way (I didn't want to over-engineer the solution though).

Either way the backfill is being invoked.
Not sure how beneficial it is (comparing to the cost).

jankowiakmaria:
Merging - possibly with follow up PR if we decide on logging the missing ids.

PR #1690, opened 2019-09-05
Pp 5585 accept payment intent notifications

mrlumbu:
This PR modifies the StripeNotificationService to accept 2 new Stripe
    notifications
    `payment_intent.amount_capturable_updated`- https://stripe.com/docs/api/events/types#event_types-payment_intent.amount_capturable_updated
    `payment_intent.payment_failed` - https://stripe.com/docs/api/events/types#event_types-payment_intent.payment_failed

   If we receive a `amount_capturable_updated` notification we check whether
   the amount capturable matches the amount of the charge, and then
   set the the charge to authorised using the existing `authorise3dsResponse` method
   on the StripePaymentProvider. As with capture, I use the structure of
   Stripe ids to decide whether to process as a payment intent or the old
   way.    
   If we receive a `payment_failed` event then the charge is moved to
   `AUTHORISATION DECLINED`, similarly to how we currently process
   the `source_failed` event.    
   These changes do not look very DRY, but we only need to process both
   source notifications and payment intent notifications for the period
   in which we're releasing this change, after which the source
   notification handling code will be deleted, so for that reason have
   kept two paths quite distinct.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
@@ -1,0 +1,66 @@
+package uk.gov.pay.connector.gateway.stripe.request;
+
+import uk.gov.pay.connector.app.StripeGatewayConfig;
+import uk.gov.pay.connector.gateway.model.OrderRequestType;
+import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+
+import java.util.Map;
+
+public class StripePaymentIntentRequest extends StripeRequest {
+
+    private final String amount;
+    private final String paymentMethodId;
+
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
```diff
@@ -0,0 +1,66 @@
+package uk.gov.pay.connector.gateway.stripe.request;
+
+import uk.gov.pay.connector.app.StripeGatewayConfig;
+import uk.gov.pay.connector.gateway.model.OrderRequestType;
+import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+
+import java.util.Map;
+
+public class StripePaymentIntentRequest extends StripeRequest {
```
kbottla, line 10: This looks unused, but imagine it is required for authorisation

kbottla (approved):
LGTM!

PR #1692, opened 2019-09-06
PP-5402-Date validation for telephone payments

nimalank7:
## WHAT YOU DID
This PR adds custom validation for dates (e.g. created_date and authorised_date in the JSON) to check if they are in valid ISO-8601 format. Both createdDate and authorisedDate in TelephoneChargeCreateRequest have both been annotated with this. See the DateValidator file.

## How to test

DateValidatorTest and integration tests in ChargesApiTelephonePaymentResourceIT. I have also added a tearDown() method in ChargeServiceTest.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/telephone/TelephoneChargeCreateRequest.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/telephone/TelephoneChargeCreateRequest.java
@@ -6,7 +6,7 @@ import uk.gov.pay.connector.charge.validation.telephone.ValidCardExpiryDate;
 import uk.gov.pay.connector.charge.validation.telephone.ValidCardFirstSixDigits;
 import uk.gov.pay.connector.charge.validation.telephone.ValidCardLastFourDigits;
 import uk.gov.pay.connector.charge.validation.telephone.ValidCardType;
-import uk.gov.pay.connector.charge.validation.telephone.ValidDateTime;
+import uk.gov.pay.connector.charge.validation.telephone.ValidZonedDateTime;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/charge/validation/telephone/DateValidator.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.pay.connector.charge.validation.telephone;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeParseException;
+import java.util.regex.Pattern;
+
+public class DateValidator implements ConstraintValidator<ValidDate, String> {
```
danworth: Maybe this could be called `ZonedDateTimeValidator` (and update the annotation name itself) since it'll fail validation for just a date
```
jshell> ZonedDateTime.parse("2018-02-21")
|  Exception java.time.format.DateTimeParseException: Text '2018-02-21' could not be parsed at index 10
|        at DateTimeFormatter.parseResolved0 (DateTimeFormatter.java:2046)
|        at DateTimeFormatter.parse (DateTimeFormatter.java:1948)
|        at ZonedDateTime.parse (ZonedDateTime.java:598)
|        at ZonedDateTime.parse (ZonedDateTime.java:583)
|        at (#4:1)
```

src/main/java/uk/gov/pay/connector/charge/model/telephone/TelephoneChargeCreateRequest.java
```diff
@@ -23,8 +24,10 @@
     @NotNull(message = "Field [description] cannot be null")
     private String description;
     
+    @ValidDate(message = "Field [created_date] must be in ISO-8601 format")
```
danworth: Maybe mention in the message that it must include the time also, I think `ISO 8601` format covers the format for both date and date with a time
https://en.wikipedia.org/wiki/ISO_8601

danworth (changes requested):
Looks really good. My only suggestion is that the annotation and its validator might be called `ZonedDateTimeValidator` rather than just `DateValidator` since its doing the former (also maybe mention in the messages that you expect it to include the time also).

src/main/java/uk/gov/pay/connector/charge/validation/telephone/DateValidator.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/model/telephone/TelephoneChargeCreateRequest.java
nimalank7: Done

danworth (approved):
Looks good 👍 
Some of those commits could be squashed down but as discussed I'm happy to help run through how git rebase works later since you need to get this in.

PR #1694, opened 2019-09-06
PP-5402 Truncate fields stored in external metadata over 50 characters

nimalank7:
## WHAT YOU DID
Context for why this PR is needed is found here [PP-5613 Enforce ExternalMetadata Validation when saving ChargeEntity](https://github.com/alphagov/pay-connector/pull/1680). 

This PR modifies storeExtraFieldsInMetaData in ChargeServe to ensure that any fields stored in metadata over 50 characters are truncated to 50. Anything that is 50 characters or below are kept the same.

## How to test

See ChargeServiceTest and ChargesApiTelephonePaymentResourceIT.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -164,7 +164,7 @@ public class ChargeService {
                     telephoneChargeRequest.getProviderId(),
                     SupportedLanguage.ENGLISH
             );
-
+            
             chargeDao.persist(chargeEntity);
             return chargeEntity;
         });
@@ -744,18 +744,24 @@ public class ChargeService {
 
     private ExternalMetadata storeExtraFieldsInMetaData(TelephoneChargeCreateRequest telephoneChargeRequest) {
         HashMap<String, Object> telephoneJSON = new HashMap<>();
-        telephoneJSON.put("processor_id", telephoneChargeRequest.getProcessorId());
+        telephoneJSON.put("processor_id", telephoneChargeRequest.getProcessorId().substring(0, Math.min(telephoneChargeRequest.getProcessorId().length(), 50)));
         telephoneJSON.put("status", telephoneChargeRequest.getPaymentOutcome().getStatus());
-
         telephoneChargeRequest.getCreatedDate().ifPresent(createdDate -> telephoneJSON.put("created_date", createdDate));
         telephoneChargeRequest.getAuthorisedDate().ifPresent(authorisedDate -> telephoneJSON.put("authorised_date", authorisedDate));
-        telephoneChargeRequest.getAuthCode().ifPresent(authCode -> telephoneJSON.put("auth_code", authCode));
-        telephoneChargeRequest.getTelephoneNumber().ifPresent(telephoneNumber -> telephoneJSON.put("telephone_number", telephoneNumber));
-        telephoneChargeRequest.getPaymentOutcome().getCode().ifPresent(code -> telephoneJSON.put("code", code));
+        telephoneChargeRequest.getAuthCode().ifPresent(authCode -> telephoneJSON.put("auth_code", authCode.substring(0, Math.min(authCode.length(), 50))));
+        telephoneChargeRequest.getTelephoneNumber().ifPresent(telephoneNumber -> telephoneJSON.put("telephone_number", telephoneNumber.substring(0, Math.min(telephoneNumber.length(), 50))));
+        telephoneChargeRequest.getPaymentOutcome().getCode().ifPresent(code -> telephoneJSON.put("code", code.substring(0, Math.min(code.length(), 50))));
         telephoneChargeRequest.getPaymentOutcome().getSupplemental().ifPresent(
                 supplemental -> {
-                    supplemental.getErrorCode().ifPresent(errorCode -> telephoneJSON.put("error_code", errorCode));
-                    supplemental.getErrorMessage().ifPresent(errorMessage -> telephoneJSON.put("error_message", errorMessage));
+                    supplemental.getErrorCode().ifPresent(errorCode -> telephoneJSON.put("error_code", errorCode.substring(0, Math.min(errorCode.length(), 50))));
+                    supplemental.getErrorMessage().ifPresent(errorMessage -> {
+                        
+                        if(errorMessage.length() > 50) {
+                            logger.info("Telephone payment error message - processor_id {} is longer than 50 characters and has been truncated and stored. Actual value is {}", telephoneChargeRequest.getProcessorId(), errorMessage);
+                        }
+                        
+                        telephoneJSON.put("error_message", errorMessage.substring(0, Math.min(errorMessage.length(), 50)));
+                    });
                 }
         );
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -761,18 +761,25 @@ private boolean chargeIsInLockedStatus(OperationType operationType, ChargeEntity
 
     private ExternalMetadata storeExtraFieldsInMetaData(TelephoneChargeCreateRequest telephoneChargeRequest) {
         HashMap<String, Object> telephoneJSON = new HashMap<>();
-        telephoneJSON.put("processor_id", telephoneChargeRequest.getProcessorId());
+        telephoneJSON.put("processor_id", telephoneChargeRequest.getProcessorId().substring(0, Math.min(telephoneChargeRequest.getProcessorId().length(), 50)));
```
kbottla: Don't we have validations in place for length of the fields. For example, for Processor ID length allowed is 23 characters as per specs. https://alphagov.github.io/pay-telephone-payments/2019-07-04-spec.html#header-technical-decisions-agreed
- I don't think we need to truncate, if fields are validated for length and valid values. Or we could only focus on fields we are not aware of.
- If there is no validation and if we are truncating fields, how do we ensure that the field is returned in full (ie.., without missing info).

kbottla (commented):
Haven't gone through it in full, but added a query to know more about the reason for the changes

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Not for processor_id length in connector/publicapi. It's to prevent the validation error in metadata in case if something greater than 23 is sent. The remaining truncations are for sanity checks (e.g. created_date/authorised_date won't be 50 characters but just to close that loophole). We won't return the field in full but just truncated except for supplemental_error_message which will be logged. I had a chat with @heathd about this.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -761,18 +761,25 @@ private boolean chargeIsInLockedStatus(OperationType operationType, ChargeEntity
...
         telephoneJSON.put("status", telephoneChargeRequest.getPaymentOutcome().getStatus());
         
-        telephoneChargeRequest.getCreatedDate().ifPresent(createdDate -> telephoneJSON.put("created_date", createdDate));
-        telephoneChargeRequest.getAuthorisedDate().ifPresent(authorisedDate -> telephoneJSON.put("authorised_date", authorisedDate));
-        telephoneChargeRequest.getAuthCode().ifPresent(authCode -> telephoneJSON.put("auth_code", authCode));
-        telephoneChargeRequest.getTelephoneNumber().ifPresent(telephoneNumber -> telephoneJSON.put("telephone_number", telephoneNumber));
-        telephoneChargeRequest.getPaymentOutcome().getCode().ifPresent(code -> telephoneJSON.put("code", code));
+        telephoneChargeRequest.getCreatedDate().ifPresent(createdDate -> telephoneJSON.put("created_date", createdDate.substring(0, Math.min(createdDate.length(), 50))));
+        telephoneChargeRequest.getAuthorisedDate().ifPresent(authorisedDate -> telephoneJSON.put("authorised_date", authorisedDate.substring(0, Math.min(authorisedDate.length(), 50))));
```
kbottla: Can we not truncate `createdDate`, `authorisedDate` , `errorCode`. They are validated and is unlikely to exceed 50 chars ?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -761,18 +761,25 @@ private boolean chargeIsInLockedStatus(OperationType operationType, ChargeEntity
...
+        telephoneChargeRequest.getPaymentOutcome().getCode().ifPresent(code -> telephoneJSON.put("code", code.substring(0, Math.min(code.length(), 50))));
         telephoneChargeRequest.getPaymentOutcome().getSupplemental().ifPresent(
                 supplemental -> {
-                    supplemental.getErrorCode().ifPresent(errorCode -> telephoneJSON.put("error_code", errorCode));
-                    supplemental.getErrorMessage().ifPresent(errorMessage -> telephoneJSON.put("error_message", errorMessage));
+                    supplemental.getErrorCode().ifPresent(errorCode -> telephoneJSON.put("error_code", errorCode.substring(0, Math.min(errorCode.length(), 50))));
+                    supplemental.getErrorMessage().ifPresent(errorMessage -> {
+                        
+                        if(errorMessage.length() > 50) {
+                            logger.info(String.format("Telephone payment message for charge_processor_id of " + telephoneChargeRequest.getProcessorId()+  " is longer than 50 characters and has been truncated and stored. The full message is: " + errorMessage));
```
kbottla: Could use String.format semantics for parameters instead of concatening string `+`. Below is applicable for processor ID and need to add a new one for `errorMessage`
```suggestion
                            logger.info(String.format("Telephone payment message - processor_id is longer than 50 characters and has been truncated and stored. Actual value is [%s]", telephoneChargeRequest.getProcessorId()) );
```

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -761,18 +761,25 @@ private boolean chargeIsInLockedStatus(OperationType operationType, ChargeEntity
...
-        telephoneChargeRequest.getAuthCode().ifPresent(authCode -> telephoneJSON.put("auth_code", authCode));
-        telephoneChargeRequest.getTelephoneNumber().ifPresent(telephoneNumber -> telephoneJSON.put("telephone_number", telephoneNumber));
-        telephoneChargeRequest.getPaymentOutcome().getCode().ifPresent(code -> telephoneJSON.put("code", code));
+        telephoneChargeRequest.getCreatedDate().ifPresent(createdDate -> telephoneJSON.put("created_date", createdDate.substring(0, Math.min(createdDate.length(), 50))));
+        telephoneChargeRequest.getAuthorisedDate().ifPresent(authorisedDate -> telephoneJSON.put("authorised_date", authorisedDate.substring(0, Math.min(authorisedDate.length(), 50))));
+        telephoneChargeRequest.getAuthCode().ifPresent(authCode -> telephoneJSON.put("auth_code", authCode.substring(0, Math.min(authCode.length(), 50))));
+        telephoneChargeRequest.getTelephoneNumber().ifPresent(telephoneNumber -> telephoneJSON.put("telephone_number", telephoneNumber.substring(0, Math.min(telephoneNumber.length(), 50))));
```
kbottla: Can we also log message when truncating `authCode`, `telephoneNumber` ? May be extract into small method and use for all truncated fields

```suggestion
       checkAndGetTruncatedValue("telephone_number", telephoneNumber);
```

```
private String checkAndGetTruncatedValue(String field, String value) {
        if (value.length() > 50) {
            logger.info(String.format("Telephone payment message - [%s] is longer than 50 characters and has been truncated and stored. Actual value is [%s]", field, value));
            return value.substring(0, 50);
        }
        return value;
    }
```

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Error code isn't validated but payment_code is. Created_date and authorised_date are validated but error_Code needs to have validation. Even though payment_outcome_code is validated and will most likely be under 50 characters it could be the case in the future it changes? So this will close down the loophole.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done

kbottla (commented):
Looks good to me. Can you squash all commits as they are all related ?

kbottla (approved):

PR #1695, opened 2019-09-06
PP-5585 Refactor stripe cancel

mrlumbu:
Change stripe cancel to use Stripe*Request pattern. This will make it easier to change to handle cancelling payment intents.


kbottla (approved):
LGTM!

PR #1696, opened 2019-09-06
PP-5585 Cancel payment intents

mrlumbu:
Use transaction id to determine whether to cancel using payment intent
or charge api.


kbottla (approved):
LGTM!

PR #1699, opened 2019-09-09
PP-5619 Add used column to tokens table

danworth:
with @AlexBishop1


alexbishop1 (approved):
I paired with @danworth on this and am confident it requires no further code review. We tested the change locally and it had the desired effect (added a column called `used` of type `boolean` with all existing rows set to `null`).

PR #1701, opened 2019-09-09
PP-5619 Mark all existing tokens in the database as unused

alexbishop1:
```sql
UPDATE tokens SET used = false WHERE used IS NULL;
```


danworth (approved):
Looks good, I've checked this locally and all existing tokens are set to `false` 👍

PR #1702, opened 2019-09-09
Pp 5619 only return unused tokens

danworth:
PP-5619 Only return charges for unused tokens
    
    When searching for a charge by token id only return the charge if the
    token is unused.

    Remove the find token by charge id since it was only being used by its
    own tests.

    
    with @AlexBishop1


alexbishop1 (approved):
I paired with @danworth on this and I am confident that in line with our policies, this change does not require further review.

PR #1707, opened 2019-09-10
PP-5619 Add end-point to mark token as used

danworth:
Returns no content for successful invocation.

Update:
When truncating the test database after each test include the `tokens` table explicitly
    since it has no fk link to either `gateway_accounts` or
    `emitted_events`. This was likely causing some intermittent failures on
    the `SecurityTokenResourceIT` tests and was preventing `SecurityTokenResourceIT.shouldDeleteToken` passing for this PR.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceIT.java
@@ -20,6 +20,7 @@ import static io.restassured.http.ContentType.JSON;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.emptyOrNullString;
 import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertTrue;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CREATED;
 
 @RunWith(DropwizardJUnitRunner.class)
@@ -56,7 +57,7 @@ public class SecurityTokensResourceIT {
     }
 
     @Test
-    public void shouldSuccessfullyGetChargeForToken() {
+    public void shouldGetChargeForToken() {
         TestToken token = DatabaseFixtures
                 .withDatabaseTestHelper(databaseTestHelper)
                 .aTestToken()
@@ -85,7 +86,7 @@ public class SecurityTokensResourceIT {
     }
 
     @Test
-    public void shouldSuccessfullyGetUnusedToken() {
+    public void shouldGetUnusedToken() {
         TestToken token = DatabaseFixtures
                 .withDatabaseTestHelper(databaseTestHelper)
                 .aTestToken()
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceIT.java
```diff
@@ -157,6 +158,34 @@ public void shouldSuccessfullyDeleteToken() {
                 .body("error_identifier", is(ErrorIdentifier.GENERIC.toString()));
     }
 
+    @Test
+    public void shouldSuccessfullyMarkTokenAsUsed() {
```
oswaldquek: do we need the word 'Successfully' here?

src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceIT.java
```diff
@@ -157,6 +158,34 @@ public void shouldSuccessfullyDeleteToken() {
...
+                .withUsed(false)
+                .insert();
+
+        givenSetup()
+                .post("/v1/frontend/tokens/" + token.getSecureRedirectToken() + "/used")
+                .then()
+                .statusCode(204)
+                .body(emptyOrNullString());
+
+        assertTrue(databaseTestHelper.isChargeTokenUsed(token.getSecureRedirectToken()));
```
oswaldquek, line 176: I wonder if it will be better to use the `/v1/frontend/token/{tokenId}` endpoint to assert the endpoint actually returns `used: true`...?

oswaldquek (changes requested):
Some comments which aren't a big deal, but worth considering.

src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceIT.java
danworth: I thought that originally, but I wanted to use the database helper so that the assertion doesn't use JPA (which would be used to check the Token via the end-point). This way it makes sure that JPA entity manager is actually committing the change to the db rather than just retrieving a cached value. I considered having both but thought was unnecessary; also, we have tests for getting a used and unused token so there isn't extra benefit there. What do you think?

src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceIT.java
danworth: Not really its just following the naming convention of the other methods in the file. I'll update them all.

src/test/java/uk/gov/pay/connector/it/resources/SecurityTokensResourceIT.java
oswaldquek: Ah ok, since it's already tested elsewhere that's fine.

oswaldquek (dismissed):

oswaldquek (approved):

PR #1712, opened 2019-09-10
BAU Retry smoke test stage 3 times

mrlumbu:
Post deploy smoke tests are a bit flakey, partly due to service
not being fully stable even after AWS says it is. Retry 3 times
to mitigate impact of this.


kbottla:
We probably don't need this now as smoke tests are not flakey recently

PR #1714, opened 2019-09-10
PP-5587 Record emitted events to DB

kbottla:
## WHAT 
- Adds an entry to EmittedEvents table when an event is offered to StateTransitionQueue
- Updates `emitted_date` when event is taken from StateTransitionQueue and is emitted
- Added service `StateTransitionService` which is now resposible for offering events to StateTransitionQueue
- Added service `EventService` that emits event and records into DB as appropriate
- Updated relevant tests and added new tests

This doesn't change the way HistoricalEventEmitter emit and records events.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
+++ b/src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
@@ -17,19 +17,19 @@ public class StateTransitionEmitterProcess {
 
     private final long STATE_TRANSITION_PROCESS_DELAY_IN_MILLISECONDS = 1000;
     private final StateTransitionQueue stateTransitionQueue;
-    private final EventQueue eventQueue;
     private final EventFactory eventFactory;
+    private EventService eventService;
 
     @Inject
     public StateTransitionEmitterProcess(
             StateTransitionQueue stateTransitionQueue,
-            EventQueue eventQueue,
             EventFactory eventFactory,
-            StateTransitionQueueMetricEmitter stateTransitionQueueMetricEmitter
+            StateTransitionQueueMetricEmitter stateTransitionQueueMetricEmitter,
+            EventService eventService
     ) {
         this.stateTransitionQueue = stateTransitionQueue;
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/connector/queue/StateTransitionService.java
```diff
@@ -0,0 +1,60 @@
...
+                refundEntity.getExternalId(),
+                Event.eventTypeForClass(refundEventClass),
+                null);
+    }
+
+    @Transactional
+    public void offerPaymentStateTransition(String externalId, ChargeStatus fromChargeState, ChargeStatus targetChargeState, ChargeEventEntity chargeEventEntity) {
+        PaymentGatewayStateTransitions.getInstance()
+                .getEventForTransition(fromChargeState, targetChargeState)
+                .ifPresent(eventType -> {
```
mrlumbu: this might all be clearer if you call this `eventClass`? At the moment the line below 
```
Event.eventTypeForClass(eventType)
```
is a bit confusing

src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java
```diff
@@ -17,19 +17,19 @@
```
mrlumbu: lower down in this file I think you need to change line https://github.com/alphagov/pay-connector/blob/39ab43041ff6da73f3f2cac7d84299e558f97cf6/src/main/java/uk/gov/pay/connector/events/StateTransitionEmitterProcess.java#L85 to use the new service?

mrlumbu (dismissed):
Looks good. One minor suggestion

src/main/java/uk/gov/pay/connector/queue/StateTransitionService.java
kbottla: Updated https://github.com/alphagov/pay-connector/compare/39ab43041ff6da73f3f2cac7d84299e558f97cf6..b389d309ec64fd368398d388b62d83c88e042299

mrlumbu (approved):

PR #1716, opened 2019-09-10
PP-5605: add `do_not_reprocess_valid_records` param to parity checker

jankowiakmaria:
* it will allow ignoring already processed records (removing the external call to ledger)


kbottla (approved):
LGTM!

PR #1720, opened 2019-09-11
PP-5585 Fix capture with payment intents

mrlumbu:
Transfer needs the charge id not the payment intent id to do
the transfer phase of the capture.


kbottla (approved):
LGTM!

PR #1721, opened 2019-09-11
PP-5585 Switch to authorise using payment intents

mrlumbu:
Introduce logic to decide whether to authorise payment using payment intents or charges API, based on the `integrationVersion3DS` of the gateway account. This will allow us to switch on payment intents authorisation for selected accounts in production to test payment flow end to end, and ensure backwards compatibility. I have tried to leave alone as much of the code as possible as post switch we will delete all the charge API based code.

There are no tests for the new code path yet. I would like to get this out in the wild and test it with IRL with a single gateway account, and then come back to this and make any adjustments, and add unit tests for the new path. Feel free to tell me this is a bad idea.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
@@ -11,34 +11,40 @@ public class StripePaymentIntentRequest extends StripeRequest {
 
     private final String amount;
     private final String paymentMethodId;
+    private final String transferGroup;
+    private String frontendUrl;
+    private String chargeExternalId;
 
-    private StripePaymentIntentRequest(
-            GatewayAccountEntity gatewayAccount,
-            String idempotencyKey,
-            StripeGatewayConfig stripeGatewayConfig,
-            String amount,
-            String paymentMethodId
-    ) {
+
+    public StripePaymentIntentRequest(
+            GatewayAccountEntity gatewayAccount, String idempotencyKey, StripeGatewayConfig stripeGatewayConfig, 
+            String amount, String paymentMethodId, String transferGroup, String frontendUrl, String chargeExternalId) {
         super(gatewayAccount, idempotencyKey, stripeGatewayConfig);
         this.amount = amount;
         this.paymentMethodId = paymentMethodId;
+        this.transferGroup = transferGroup;
+        this.frontendUrl = frontendUrl;
+        this.chargeExternalId = chargeExternalId;
     }
 
     public static StripePaymentIntentRequest of(
             CardAuthorisationGatewayRequest request,
             String paymentMethodId,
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    protected Map<String, String> params() {
+        return Map.of(
+                "card[cvc]", cvc,
+                "card[exp_month]", expiryMonth,
+                "card[exp_year]", expiryYear,
+                "card[number]", cardNo,
+                "type", "card");
```
kbottla: Think we should also send `cardholder_name` & `billing address` as we do it now

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
```diff
@@ -56,7 +62,12 @@ protected OrderRequestType orderRequestType() {
                 "amount", amount,
                 "confirmation_method", "automatic",
                 "capture_method", "manual",
-                "currency", "GBP");
+                "currency", "GBP",
+                "transfer_group", transferGroup,
+                "on_behalf_of", stripeConnectAccountId,
+                "confirm", "true",
+                "return_url", String.format("%s/card_details/%s/3ds_required_in", frontendUrl, chargeExternalId)
+        );
```
kbottla, line 70: `description` similar to current method https://github.com/alphagov/pay-connector/blob/d3d3884b48094f1e1727e048bea920ca5a6b941b/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeAuthoriseRequest.java#L78

kbottla (commented):
Looks good overall. Added a few comments around additional fields required for payment intents to be consistent with current method. Would also be good to have few tests around this but happy as a separate PR.

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
mrlumbu: ooh good catch

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
mrlumbu: yeah, wasn't entirely sure why we do this

kbottla: It appears in Stripe dashboard `description` but not sure what it will be if description is not sent

kbottla (approved):
LGTM!

PR #1723, opened 2019-09-12
PP-5402-Removes dummy PaymentId

nimalank7:
## WHAT YOU DID
This PR removes the dummy payment id that is returned by Connector which was previously included for testing purposes. Now this has been replaced with the ChargeEntity.external_id which populates payment_id field in publicapi with it.

This PR also modifies the ChargeService tests to test for the junk fields in Connector JSON as well as fixing the createTelephoneChargeForOnlyRequiredFields() test to check for state.

In both ChargeService and ChargesApiTelephonePaymentResourceIT since the chargeId is random we just check for the length.


kbottla (approved):
LGTM!

PR #1726, opened 2019-09-13
PP-5605: additional param allowing to run parity checker for only given parity status

jankowiakmaria:
**Why**
During running the parity checker we've discovered a number of charges either in DATA_MISMATCH
or MISSING_IN_LEDGER statuses. In order to be able to reprocess them in a timely manner
(rather than manually retrieving charge ids that are in given status or going through all
the charges again) we're going to process only the charges in a given state.

We want to re-process them, bacause there were changes to ledger and re-processing
will allow re-calculating the transaction state.

**What**
* changes to parity worker to retrieve and pass additional parameter
	* includes some refactoring to remove duplicate code (run locally - everything works as expected)
* changes to the task to retrieve charges in appropriate state (with pagination)
	* includes some refactoring in order to share common bits of code


kbottla (approved):
LGTM!

PR #1727, opened 2019-09-13
PP-5549 Add 3DS Flex challenge data columns

stephencdaly:
Add columns to store the data we get back in the response to an authorisation request made using 3DS flex when a challenge is requested by worldpay


rorymalcolm (approved):
LGTM

PR #1728, opened 2019-09-13
PP-5549 Persist 3DS flex challenge data

stephencdaly:
Add new columns to Auth3dsDetailsEntity so they can be persisted to and read from the database.

Modify test to check that all fields in Auth3dsDetailsEntity are successfully persisted to the database.


src/test/java/uk/gov/pay/connector/model/domain/Auth3dsDetailsEntityFixture.java
```diff
@@ -0,0 +1,74 @@
...
+        this.worldpayChallengePayload = worldpayChallengePayload;
+        return this;
+    }
+
+    public Auth3dsDetailsEntityFixture withThreeDsVersion(String threeDsVersion) {
+        this.threeDsVersion = threeDsVersion;
+        return this;
+    }
+
+    public Auth3dsDetailsEntity build() {
```
danworth, line 62: perhaps this could be called `toEntity` but I don't mind

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoIT.java
```diff
@@ -912,6 +929,12 @@ public void shouldCreateANewChargeWith3dsDetails() {
```
danworth, line 929: Seems odd that we're using the Dao to persist and then retrieve it in the same test, kind of like asking it to mark its own homework, I think elsewhere we use the database test helper to do one side of the process but this isn't anything you've done, just a passing thought.

danworth (approved):
Looks good 👍

PR #1730, opened 2019-09-13
PP-5549 Handle 3DS Flex challenge response

stephencdaly:
Handle an authorisation response from worldpay which contains a challengeRequiredElement by storing the challenge details and updating the status of the charge to "AUTHORISATION 3DS REQUIRED".

Add tests for handling 3DS flex challenge response including a "contract" test that talks to a real Worldpay gateway


rorymalcolm (approved):
LGTM

PR #1739, opened 2019-09-17
PP-5553 Return challenge JWT in charge response

stephencdaly:
When the charge is in status "AUTHORISATION 3DS REQUIRED" and the 3DS data on the charge contains challenge data, calculate and return a JWT in the response to the frontend get charge endpoint.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.charge.model;
 
 import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
@@ -21,6 +22,7 @@ import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
 import static com.fasterxml.jackson.annotation.JsonInclude.Include;
 import static uk.gov.pay.commons.model.ApiResponseDateTimeFormatter.ISO_INSTANT_MILLISECOND_PRECISION;
@@ -554,8 +556,9 @@ public class ChargeResponse {
             return md;
         }
 
-        public String getWorldpayChallengeJwt() {
-            return worldpayChallengeJwt;
+        @JsonIgnore
+        public Optional<String> getWorldpayChallengeJwt() {
+            return Optional.ofNullable(worldpayChallengeJwt);
         }
 
         public void setWorldpayChallengeJwt(String worldpayChallengeJwt) {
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -559,12 +570,13 @@ public boolean equals(Object o) {
...
+                    Objects.equals(worldpayChallengeJwt, that.worldpayChallengeJwt);
         }
 
         @Override
         public int hashCode() {
-            return Objects.hash(paRequest, issuerUrl, htmlOut, md);
+            return Objects.hash(paRequest, issuerUrl, htmlOut, md, worldpayChallengeJwt);
         }
 
         @Override
```
danworth, line 585: Does `worldpayChallengeJwt` need adding to the `toString()` for Auth3dsData?

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -551,6 +554,14 @@ public String getMd() {
             return md;
         }
 
+        public String getWorldpayChallengeJwt() {
```
danworth: I think using `Optional` for the getters on the response classes helps to highlight what the response can look like but we don't really follow that pattern so happy as-is. Interested in your opinion though.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResource.java
```diff
@@ -188,6 +172,26 @@ private ChargeResponse buildChargeResponse(UriInfo uriInfo, ChargeEntity charge)
...
+        }
+        
+        if (charge.get3dsDetails() != null) {
+            var auth3dsData = new ChargeResponse.Auth3dsData();
+            auth3dsData.setPaRequest(charge.get3dsDetails().getPaRequest());
+            auth3dsData.setIssuerUrl(charge.get3dsDetails().getIssuerUrl());
+            auth3dsData.setHtmlOut(charge.get3dsDetails().getHtmlOut());
+            auth3dsData.setMd(charge.get3dsDetails().getMd());
+
+            worldpay3dsFlexJwtService.generateChallengeTokenIfAppropriate(charge).ifPresent(
```
danworth: Just wondering whether its worth checking if the payment provider is worldpay before calling `worldpay3dsFlexJwtService.generateChallengeTokenIfAppropriate` since its worldpay specific. Not necessarily from a performance perspective but perhaps from a comprehension point of view like 'why are we calling a worldpay service for all charges', just a thought 🤷‍♂

danworth (dismissed):
Looks good, couple comments but more interested in your view than things to change 👍

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
stephencdaly: Yes it does, good spot

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
stephencdaly: I thought the same and initially tried doing it but I couldn't get it to return the optional correctly in the response. I'm pretty sure I managed it before though so I'll have another look at how we do this, unless you know of the top of your head?

src/main/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResource.java
stephencdaly: yeh I think that's a good idea 👍

stephencdaly:
I've made all the suggested changes

danworth (approved):
Looks good 👍

PR #1741, opened 2019-09-18
PP-5549 Add challenge data to WorldpayOrderStatusResponse toString()

stephencdaly:
Add the data loaded from a 3DS flex challenge response to the toString() method of  WorldpayOrderStatusResponse

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -178,6 +178,15 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
         if (StringUtils.isNotBlank(issuerUrl)) {
             joiner.add("issuerURL: " + issuerUrl);
         }
+        if (StringUtils.isNotBlank(challengeAcsUrl)) {
+            joiner.add("threeDSChallengeDetails acsUrl: " + challengeAcsUrl);
+        }
+        if (StringUtils.isNotBlank(challengeTransactionId)) {
+            joiner.add("threeDSChallengeDetails transactionId3DS: " + challengeTransactionId);
+        }
+        if (StringUtils.isNotBlank(threeDsVersion)) {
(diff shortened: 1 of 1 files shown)


alexbishop1 (changes requested):
Could you make the names in the string match the XML elements/attributes i.e. `acsURL`, `transactionId3DS`, `payload`, `threeDSVersion` to make it consistent with the other ones (the idea is that it should look pretty close to what Worldpay send us, rather than our interpretation)?

Also, could you add these new bits _before_ the two error ones?

stephencdaly:
done

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -178,6 +178,18 @@ public String toString() {
...
             joiner.add("issuerURL: " + issuerUrl);
         }
+        if (StringUtils.isNotBlank(challengeAcsUrl)) {
+            joiner.add("threeDSChallengeDetails acsUrl: " + challengeAcsUrl);
+        }
+        if (StringUtils.isNotBlank(challengeTransactionId)) {
+            joiner.add("threeDSChallengeDetails transactionId3DS: " + challengeTransactionId);
+        }
+        if (StringUtils.isNotBlank(challengePayload)) {
+            joiner.add("threeDSChallengeDetails payload: " + challengePayload);
```
alexbishop1: Actually, it looks like the payload is a long base64-encoded string that is pretty opaque to us so probably not worth including when logging etc. Could you remove this one please?

alexbishop1 (changes requested):
Sorry to be a pain…

stephencdaly:
done

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
stephencdaly: done

alexbishop1 (approved):
Awesomesauce! 🎉

PR #1743, opened 2019-09-18
PP-5648 make expiry date an optional field

SandorArpa:
## WHAT YOU DID
- as pointed out by our integrating partner, card_expiry field is not present when the transaction failed. We're removing the @NotNull annotation from TelephoneChargeCreateRequest
- update relevant tests

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
@@ -571,6 +571,7 @@ public class ChargeServiceTest {
 
         TelephoneChargeCreateRequest telephoneChargeCreateRequest = telephoneRequestBuilder
                 .withPaymentOutcome(paymentOutcome)
+                .withCardExpiry(null)
                 .build();
 
         service.create(telephoneChargeCreateRequest, GATEWAY_ACCOUNT_ID);
@@ -594,7 +595,7 @@ public class ChargeServiceTest {
         assertThat(createdChargeEntity.getCardDetails().getLastDigitsCardNumber().toString(), is("1234"));
         assertThat(createdChargeEntity.getCardDetails().getFirstDigitsCardNumber().toString(), is("123456"));
         assertThat(createdChargeEntity.getCardDetails().getCardHolderName(), is("Jane Doe"));
-        assertThat(createdChargeEntity.getCardDetails().getExpiryDate(), is("01/19"));
+        assertThat(createdChargeEntity.getCardDetails().getExpiryDate(), is(nullValue()));
         assertThat(createdChargeEntity.getCardDetails().getCardBrand(), is("visa"));
         assertThat(createdChargeEntity.getProviderSessionId(), is("1PROV"));
         assertThat(createdChargeEntity.getExternalMetadata().get().getMetadata(), equalTo(metadata));
@@ -620,6 +621,7 @@ public class ChargeServiceTest {
 
         TelephoneChargeCreateRequest telephoneChargeCreateRequest = telephoneRequestBuilder
                 .withPaymentOutcome(paymentOutcome)
+                .withCardExpiry(null)
                 .build();
 
         service.create(telephoneChargeCreateRequest, GATEWAY_ACCOUNT_ID);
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/charge/validation/telephone/CardExpiryValidatorTest.java
```diff
@@ -85,7 +85,6 @@ public void passesValidationForNullCardExpiry() {
 
         Set<ConstraintViolation<TelephoneChargeCreateRequest>> constraintViolations = validator.validate(telephoneChargeCreateRequest);
 
-        assertThat(constraintViolations.size(), isNumber(1));
-        assertThat(constraintViolations.iterator().next().getMessage(), is("Field [card_expiry] cannot be null"));
+        assertThat(constraintViolations.size(), isNumber(0));
```
nimalank7: It might be better to do `assertThat(constraintViolations.isEmpty(), is(true)` as the previous test uses this. Thoughts?

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -571,6 +571,7 @@ public void shouldCreateATelephoneChargeForFailureCodeOfP0010() {
 
         TelephoneChargeCreateRequest telephoneChargeCreateRequest = telephoneRequestBuilder
                 .withPaymentOutcome(paymentOutcome)
+                .withCardExpiry(null)
```
nimalank7, line 574: It's unnecessary do to this - the TelephoneChargeCreateRequest spins up both required and optional fields. We can test for the optional fields in the integration tests.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -620,6 +621,7 @@ public void shouldCreateATelephoneChargeForFailureCodeOfP0050() {
...
```
nimalank7, line 624: See above remark ^^

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -861,6 +863,7 @@ public void shouldCreateATelephoneChargeResponseForFailure() {
...
```
nimalank7, line 866: See above remark ^^

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiTelephonePaymentResourceIT.java
```diff
@@ -128,6 +130,7 @@ public void createTelephoneChargeForFailedStatusP0010() {
                         )
                 )
         );
+        postBody.put("card_expiry", null);
```
nimalank7: I think this is unnecessary as the postBody doesn't have card_expiry so it will send null anyway.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiTelephonePaymentResourceIT.java
```diff
@@ -250,6 +255,7 @@ public void createTelephoneChargeWithTruncatedMetaData() {
         postBody.put("created_date", "2018-02-21T16:04:25Z");
         postBody.put("authorised_date", "2018-02-21T16:05:33Z");
         postBody.put("telephone_number", stringOf51Characters);
+        postBody.put("card_expiry", "02/19");
```
nimalank7: This can be removed as card_expiry isn't being truncated and stored in external meta-data.

nimalank7 (changes requested):
Some small changes needed - let me know if you agree!

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
SandorArpa: It is necessary as the test breaks for nullValue(). In setUp() telephoneRequestBuilder is created with a value and thus creation will return a value. For test readability it is nicer to show that we want card_expiry to be null than not setting it in setUp() and then adding it in tests where necessary.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
SandorArpa: Same as above, in setUp() telephoneRequestBuilder is created with a value and thus creation will return a value.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
nimalank7: Sure

nimalank7 (dismissed):
Looks good to me!

nimalank7 (approved):
Looks good to me.

PR #1744, opened 2019-09-18
PP-5553 Return 500 response if credentials for generating JWT missing

stephencdaly:
When calling the frontend endpoint to generate a Worldpay 3DS flex DDC JWT or when calling the frontend endpoint to get a charge when it is in a state that will cause the response to contain a calculated JWT, if the required credentials aren't set on the account, return a 500 rather than a 409.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
@@ -28,6 +28,7 @@ import uk.gov.pay.commons.utils.metrics.DatabaseMetricsService;
 import uk.gov.pay.commons.utils.xray.Xray;
 import uk.gov.pay.connector.cardtype.resource.CardTypesResource;
 import uk.gov.pay.connector.charge.exception.ConflictWebApplicationExceptionMapper;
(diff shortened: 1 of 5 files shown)


rhowe-gds:
We trigger alerting on 5xx response codes from microservices, which is a sign that something unexpected and unhandled is going on. I don't think we should intentionally throw 500s

stephencdaly:
> We trigger alerting on 5xx response codes from microservices, which is a sign that something unexpected and unhandled is going on. I don't think we should intentionally throw 500s

@rhowe-gds You make a good point. The reasoning for us thinking it should throw a 500 is because there isn't a problem with the request, but the state of the account, which will cause the request to fail.
This is something we'd want to know about because it will cause payments for that account to fail. If we don't return a 500 here, a 500 will be returned from frontend anyway.
Of course open to a better suggestion.

rhowe-gds:
We shouldn't throw a 500 because it's a situation we anticipate and are explicitly handling, we should return perhaps a 403 with an appropriate response. 

How frontend then responds to the user is another question - it maybe shouldn't even return an error status, but a regular HTTP 200 with information to the user perhaps?

stephencdaly:
@rhowe-gds You've won me over, I will close this pull request.

PR #1745, opened 2019-09-19
PP-5553 Test that 409 is returned when credentials are missing

stephencdaly:
Test that a 409 response is returned when calling the frontend get charge endpoint, and the charge is in a state where we calculate a challenge JWT but the required credentials are missing.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceWorldpayJwtIT.java
+++ b/src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceWorldpayJwtIT.java
@@ -161,6 +161,32 @@ public class ChargesFrontendResourceWorldpayJwtIT {
                 .body("auth_3ds_data", not(hasKey("worldpayChallengeJwt")));
     }
 
+    @Test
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/connector/charge/resource/ChargesFrontendResourceWorldpayJwtIT.java
```diff
@@ -161,6 +161,32 @@ public void shouldOmitChallengeJwtWhenChargeNotInAppropriateState() {
...
+                "a-transaction-id",
+                "a-payload",
+                "2.1.0");
+
+        connectorRestApi
+                .withAccountId(gatewayAccountId)
+                .withChargeId(chargeExternalId)
+                .getFrontendCharge()
+                .statusCode(HttpStatus.SC_CONFLICT)
+                .body("message", is(List.of("Cannot generate Worldpay 3ds Flex JWT for account 202 because the following credential is unavailable: issuer")));
```
alexbishop1, line 187: This should really say that it’s specifically the Worldpay 3DS Flex _challenge_ JWT.

alexbishop1 (approved):

rorymalcolm (approved):
looks good to me

PR #1746, opened 2019-09-19
PP-5650 Provider Id correctly stored in gateway_transaction_id

nimalank7:
At present the provider_id is incorrectly being stored in the provider_session_id column in Connector database. This PR fixes this and stores provider_id in the gateway_transaction_id field column. 

Since ChargeDAO already has a findByGatewayTransactionId I removed the findByProviderSessionId. ChargeEntity has been refactored to have it stored in gateway_transaction_id, as well as the tests. This also removes adding provider_id in the test charge which I previously added for findByProviderSessionId. Since we don't do this now I've removed that addition to the test charge.


kbottla (commented):
Looks good to me. Can you squash all commits to one ?

nimalank7:
@kbottla commits have been squashed.

kbottla (approved):

PR #1747, opened 2019-09-19
PP-5557 Send different 2nd order request for 3DS Flex

stephencdaly:
When a request is made to /v1/frontend/charges/{chargeId}/3ds when 3DS Flex is used for the payment, the request will not contain a paResponse.

In this case, the order request XML payload will now contain a `<completedAuthentication/>` element rather than a `<paRequest></paRequest>` element.


rorymalcolm (approved):
LGTM

PR #1751, opened 2019-09-20
PP-5649 Store Card Type

rorymalcolm:
- Enables the storing of 'card type' from charges, as well as adding a
column to the charges table in which to store it.

- Updates 'card brand' to be 'card brand' as it was previously 'card
type' and this causes a mismatch with elsewhere in our system

- Creates a new enum for the possible card type values


alexbishop1:
Please put the database migration in its own PR.

Database migrations aren’t run on deploy (except locally), so when the application starts, it will be running the new code but without the new database column until the migration is run.

Due to this complexity, it’s always best to have database migrations in their own PRs with no other code changes.

src/main/java/uk/gov/pay/connector/charge/model/CardDetailsEntity.java
```diff
@@ -36,6 +38,12 @@
 
     @Column(name = "card_brand")
     private String cardBrand;
+    
+    @Convert(converter = SupportedTypeConverter.class)
+    @JsonSerialize(using = ToStringSerializer.class)
+    @JsonProperty("card_type")
+    @Column(name = "card_type")
+    private SupportedType cardType;
```
stephencdaly: Rather than having a custom converter I think you can use
`@Enumerated(EnumType.STRING)`

src/main/java/uk/gov/pay/connector/charge/model/telephone/TelephoneChargeCreateRequest.java
```diff
@@ -42,13 +43,15 @@
     @Valid
     private PaymentOutcome paymentOutcome;
     
-    @ValidCardType(message = "Field [card_type] must be either master-card, visa, maestro, diners-club or american-express")
-    private String cardType;
+    @ValidCardBrand(message = "Field [card_brand] must be either master-card, visa, maestro, diners-club or american-express")
+    private String cardBrand;
```
stephencdaly: can we just change this? would public api not be expecting the field to be named `card_type`?

rorymalcolm:
> Please put the database migration in its own PR.
> 
> Database migrations aren’t run on deploy (except locally), so when the application starts, it will be running the new code but without the new database column until the migration is run.
> 
> Due to this complexity, it’s always best to have database migrations in their own PRs with no other code changes.

https://github.com/alphagov/pay-connector/pull/1752

src/test/java/uk/gov/pay/connector/it/dao/ChargeDaoCardDetailsIT.java
```diff
@@ -120,7 +121,7 @@ public void persist_shouldStoreCardDetails() {
 
         Address billingAddress = AddressFixture.anAddress().build();
         ChargeEntity chargeEntity = ChargeEntityFixture.aValidChargeEntity().build();
-        CardDetailsEntity cardDetailsEntity = new CardDetailsEntity(FirstDigitsCardNumber.of("123456"), LastDigitsCardNumber.of("1258"), "Mr. Pay Mc Payment", "03/09", "VISA", new AddressEntity(billingAddress));
+        CardDetailsEntity cardDetailsEntity = new CardDetailsEntity(FirstDigitsCardNumber.of("123456"), LastDigitsCardNumber.of("1258"), "Mr. Pay Mc Payment", "03/09", "VISA", SupportedType.DEBIT, new AddressEntity(billingAddress));
```
stephencdaly: I think we should add an assertion here that the card type is persisted

src/main/java/uk/gov/pay/connector/cardtype/model/domain/SupportedType.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Internal entity used to drive the frontend UI based on the
+ * strings stored in the table {@code card_types}. This represents
+ * which card types are supported by the service, not what card types
+ * are used for payment by the paying user
+ * <p>
+ * This should be used only for driving the UI
+ */
+public enum SupportedType {
```
stephencdaly: I'm not sure we should re-use this enum, as even though it has the same values, it's used for a different purpose, which is specifying the card types for cards allowed by the service.
If we do re-use it, the comment is no longer valid, and the name should be changed to 'CardType' as it now has a more generic use.

src/main/java/uk/gov/pay/connector/charge/model/CardDetailsEntity.java
stephencdaly: You also might not need the `@JsonSerialize` but test it out

stephencdaly (changes requested):
A few things

src/main/java/uk/gov/pay/connector/cardtype/model/domain/CardType.java
```diff
@@ -0,0 +1,14 @@
+package uk.gov.pay.connector.cardtype.model.domain;
+
+/**
+ * Internal entity used to drive the frontend UI based on the
+ * strings stored in the table {@code card_types}. This represents
+ * which card types are supported by the service, not what card types
+ * are used for payment by the paying user, which is {@code PayersCardType}
+ * <p>
+ * This should be used only for driving the UI
+ */
```
stephencdaly, line 7: This comment should be revised as it's not only used for driving the UI anymore.
Probably should keep a comment to explain the distinction between this and PayersCardType, but highlight the 2 uses it now has.

src/main/resources/migrations.xml
```diff
@@ -1255,4 +1255,12 @@
...
     </changeSet>
 
+    <changeSet id="add card type column" author="">
+        <addColumn tableName="charges">
+            <column name="card_type" type="varchar(20)">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
```
stephencdaly, line 1265: Can you rebase the PR to remove the migration from the PR.

stephencdaly (changes requested):
Looks good, just a comment about a comment.
And you need to rebase the PR to pull in the migration change.

stephencdaly (approved):

PR #1752, opened 2019-09-23
PP-5649 Add Card Type Migration

rorymalcolm:
- Adds a migration to enable the card type column on charges


stephencdaly (approved):
looks good

PR #1753, opened 2019-09-23
PP-5635: add dao method to retrieve not emitted events

jankowiakmaria:
Context:
shuts down in an ungraceful way, we need to track events that hasn't been emitted.

Changes:
* introduce dao method which retrieves all the events from emitted_events table
that don't have emitted_date set


kbottla (approved):
LGTM! We might need an index on these fields

jankowiakmaria:
Yep, was just thinking about that. I will look into it separately (it terms of PRs).

PR #1756, opened 2019-09-24
PP-5635: create EmittedEventBackfillService for backfilling missing events

jankowiakmaria:
Context:
shuts down in an ungraceful way, we need to track events and reprocess them when they haven't been emitted.

Changes:
* add processPaymentAndRefundEvents to historcal event emitter (wrapper method for otherwise
two separate methods)
* created service that
  * retrieves not emitted events from emitted events table
  * for each emitted event retrieves relevant charge and invokes charge backfill (for both payment and refund)
  * sets emittedDate to now (to mark an event as processed)

Possible improvements:
If there are multiple events for the same charge it would invoke the backfill for each of those events.
We could possibly store charge ids and verify wheteher the charge has been already processed in given run

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
+++ b/src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
@@ -1,0 +1,85 @@
+package uk.gov.pay.connector.events;
+
+import com.google.inject.persist.Transactional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.app.config.EmittedEventSweepConfig;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.service.ChargeService;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
```diff
@@ -0,0 +1,85 @@
...
+        this.emittedEventDao = emittedEventDao;
+        this.chargeService = chargeService;
+        this.refundDao = refundDao;
+        this.sweepConfig = configuration.getEmittedEventSweepConfig();
+        this.historicalEventEmitter = new HistoricalEventEmitter(emittedEventDao, refundDao, shouldForceEmission,
+                eventService, stateTransitionService);
+    }
+
+    public void backfillNotEmittedEvents() {
+        List<EmittedEventEntity> notEmittedEventsToProcess =
```
kbottla, line 50: Would be useful to have some sort of pagination as the volumes could be high if connector fails to emit to SQS for long periods

kbottla (approved):

src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
jankowiakmaria: sure - will address in separate PR.

PR #1760, opened 2019-09-24
PP-5635: resource for emitted event sweeper

jankowiakmaria:
Context:
shuts down in an ungraceful way, we need to track events and reprocess them when they haven't been emitted.

Changes:
* added resource that invokes the emittedEventsBackfillService

### DIFF
--- a/src/test/java/uk/gov/pay/connector/events/resource/EmittedEventResourceTest.java
+++ b/src/test/java/uk/gov/pay/connector/events/resource/EmittedEventResourceTest.java
@@ -1,0 +1,35 @@
+package uk.gov.pay.connector.events.resource;
+
+import io.dropwizard.testing.junit.ResourceTestRule;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.junit.MockitoJUnitRunner;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/connector/events/resource/EmittedEventResourceTest.java
```diff
@@ -0,0 +1,35 @@
...
+            .build();
+
+    @Test
+    public void shouldReturn200() {
+        Response response = resources
+                .target("/v1/tasks/emitted-events-sweep")
+                .request()
+                .post(Entity.json(""));
+
+        assertThat(response.getStatus(), is(200));
```
kbottla, line 33: [minor] for readability
```suggestion
        assertThat(response.getStatus(), is(HttpStatus.SC_OK/Status.OK));
```

kbottla (approved):
LGTM! just a minor suggestion. Might also need to hold on to merging as master is 🔴

src/test/java/uk/gov/pay/connector/events/resource/EmittedEventResourceTest.java
jankowiakmaria: Will update in follow up PR.

PR #1761, opened 2019-09-24
PP-5658 Emit card brand label with payment details entered event

sfount:
Add card brand label to details that are sent to Ledger during payment details entered. 

Join card types to card details on charge entity. This isn't needed anywhere else the charge entity is used but there is no clean abstraction for including database lookup logic anywhere in the event/ event details building process.

* verify that the `card_brand_label` attribute is set on the payment details entered event


kbottla (approved):
LGTM!

PR #1762, opened 2019-09-24
PP-5445 add integration test to check if events are created

SandorArpa:
## WHAT YOU DID
- extend ChargesApiTelephonePaymentResourceIT to check whether the relevant events were stored in the DB
- add new method to `DatabaseTestHelper`


kbottla (approved):
LGTM!

PR #1764, opened 2019-09-24
PP-5445 add event queue pact test

SandorArpa:
## WHAT YOU DID
- add an event queue message contract test for payment notification created event


kbottla (approved):
LGTM!

PR #1767, opened 2019-09-25
PP-5635: address review comments

jankowiakmaria:
* Clean up
* add pagination to findNotEmittedEventsOlderThan method

PR #1768, opened 2019-09-25
PP-5635: add indices for emitted_events date columns

jankowiakmaria:
Context:
The emitted event sweeper queries emitted_events table to retrieve
not emitted events. It uses event_date and emitted_date to determine
which records should be re-processed (https://github.com/alphagov/pay-connector/pull/1767).
In order to make sure queries are efficient indices on those columns
are required.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1271,4 +1271,10 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add partial index to emitted_events emitted_date column" runInTransaction="false" author="">
+        <sql>
+            CREATE INDEX CONCURRENTLY idx_emitted_events_null_emitted_date ON emitted_events (emitted_date) where emitted_date is null;
+        </sql>
+    </changeSet>
+
 </databaseChangeLog>


src/main/resources/migrations.xml
```diff
@@ -1271,4 +1271,17 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add indices to emitted_events date columns" author="">
+        <createIndex indexName="idx_emitted_events_event_date"
+                     tableName="emitted_events"
+                     unique="false">
+            <column name="event_date"/>
+        </createIndex>
+        <createIndex indexName="idx_emitted_events_emitted_date"
```
sfount: I'm not sure we need a full index on this field as it's mostly used for a `NOT NULL` check, I can't 100% confirm that it wouldn't speed this up but I think the index would mainly benefit fields where we expect the data to be populated.

Up to you if we think we can start without this index and add it if we see performance problems. Happy to approve if you think we should go forward with this.

jankowiakmaria:
[This time] I've run explain analyze [in test] for:
```sql
EXPLAIN ANALYZE SELECT e from emitted_events e
WHERE 
e.emitted_date is null
and e.id > 0 AND e.id <= 7283017 AND e.event_date < '2019-09-09Z'
ORDER BY e.id;

EXPLAIN ANALYZE SELECT MAX(e.id) from emitted_events e
WHERE e.event_date < '2019-09-09Z'
AND e.emitted_date is null;
```

I created test indexes/indices (for some scenarios):
```sql
create index idx_emitted_events_event_date on emitted_events (event_date);
create index idx_emitted_events_emitted_date on emitted_events (emitted_date);
create index idx_emitted_events_null_emitted_date on emitted_events (emitted_date) where emitted_date is null;
```

And run analysis:
* no new index:
```sql
Sort  (cost=102762.85..102762.85 rows=1 width=110) (actual time=469.233..469.233 rows=0 loops=1)
  Sort Key: id
  Sort Method: quicksort  Memory: 25kB
  ->  Seq Scan on emitted_events e  (cost=0.00..102762.84 rows=1 width=110) (actual time=469.228..469.228 rows=0 loops=1)
        Filter: ((emitted_date IS NULL) AND (id > 0) AND (id <= 7283017) AND (event_date < '2019-09-09 00:00:00'::timestamp without time zone))
        Rows Removed by Filter: 3232162
Planning time: 0.225 ms
Execution time: 469.249 ms

Aggregate  (cost=86602.03..86602.04 rows=1 width=8) (actual time=387.557..387.557 rows=1 loops=1)
  ->  Seq Scan on emitted_events e  (cost=0.00..86602.02 rows=1 width=8) (actual time=387.554..387.554 rows=0 loops=1)
        Filter: ((emitted_date IS NULL) AND (event_date < '2019-09-09 00:00:00'::timestamp without time zone))
        Rows Removed by Filter: 3232162
Planning time: 0.084 ms
Execution time: 387.579 ms
```
* `idx_emitted_events_event_date`
```sql
Sort  (cost=102762.85..102762.85 rows=1 width=110) (actual time=482.802..482.802 rows=0 loops=1)
  Sort Key: id
  Sort Method: quicksort  Memory: 25kB
  ->  Seq Scan on emitted_events e  (cost=0.00..102762.84 rows=1 width=110) (actual time=482.797..482.797 rows=0 loops=1)
        Filter: ((emitted_date IS NULL) AND (id > 0) AND (id <= 7283017) AND (event_date < '2019-09-09 00:00:00'::timestamp without time zone))
        Rows Removed by Filter: 3232162
Planning time: 0.144 ms
Execution time: 482.820 ms

Aggregate  (cost=86602.03..86602.04 rows=1 width=8) (actual time=354.424..354.425 rows=1 loops=1)
  ->  Seq Scan on emitted_events e  (cost=0.00..86602.02 rows=1 width=8) (actual time=354.422..354.422 rows=0 loops=1)
        Filter: ((emitted_date IS NULL) AND (event_date < '2019-09-09 00:00:00'::timestamp without time zone))
        Rows Removed by Filter: 3232162
Planning time: 0.170 ms
Execution time: 354.450 ms
```
* `idx_emitted_events_event_date`, `idx_emitted_events_emitted_date`
```sql
Sort  (cost=4.47..4.47 rows=1 width=110) (actual time=0.010..0.010 rows=0 loops=1)
  Sort Key: id
  Sort Method: quicksort  Memory: 25kB
  ->  Index Scan using idx_emitted_events_emitted_date on emitted_events e  (cost=0.43..4.46 rows=1 width=110) (actual time=0.007..0.007 rows=0 loops=1)
        Index Cond: (emitted_date IS NULL)
        Filter: ((id > 0) AND (id <= 7283017) AND (event_date < '2019-09-09 00:00:00'::timestamp without time zone))
Planning time: 0.135 ms
Execution time: 0.025 ms

Aggregate  (cost=4.45..4.46 rows=1 width=8) (actual time=0.008..0.008 rows=1 loops=1)
  ->  Index Scan using idx_emitted_events_emitted_date on emitted_events e  (cost=0.43..4.45 rows=1 width=8) (actual time=0.006..0.006 rows=0 loops=1)
        Index Cond: (emitted_date IS NULL)
        Filter: (event_date < '2019-09-09 00:00:00'::timestamp without time zone)
Planning time: 0.143 ms
Execution time: 0.028 ms
```
* `idx_emitted_events_event_date`, `idx_emitted_events_emitted_date`, `idx_emitted_events_null_emitted_date`
```sql
Sort  (cost=4.16..4.16 rows=1 width=110) (actual time=0.135..0.135 rows=0 loops=1)
  Sort Key: id
  Sort Method: quicksort  Memory: 25kB
  ->  Index Scan using idx_emitted_events_null_emitted_date on emitted_events e  (cost=0.12..4.15 rows=1 width=110) (actual time=0.033..0.033 rows=0 loops=1)
        Filter: ((id > 0) AND (id <= 7283017) AND (event_date < '2019-09-09 00:00:00'::timestamp without time zone))
Planning time: 0.243 ms
Execution time: 0.155 ms

Aggregate  (cost=4.15..4.16 rows=1 width=8) (actual time=0.002..0.002 rows=1 loops=1)
  ->  Index Scan using idx_emitted_events_null_emitted_date on emitted_events e  (cost=0.12..4.14 rows=1 width=8) (actual time=0.001..0.001 rows=0 loops=1)
        Filter: (event_date < '2019-09-09 00:00:00'::timestamp without time zone)
Planning time: 0.170 ms
Execution time: 0.022 ms
```
Given there is no easy/nice way to create partial index (and it's performance to the non-partial index on the same column is very similar) in liquibase ([docs](https://www.liquibase.org/documentation/changes/create_index.html), [hacky solution/workaround](https://forum.liquibase.org/topic/how-to-create-conditional-indexes-using-createindex)) I'm going to leave only `idx_emitted_events_emitted_date` index.

heathd:
shouldn't we `CREATE INDEX CONCURRENTLY`? eg. https://github.com/alphagov/pay-connector/blob/master/src/main/resources/migrations.xml#L890

jankowiakmaria:
I updated the index to be created concurrently and to be partial.

heathd:
thanks for making it create concurrently

Haven't used partial indexes before, but seems like it does perform better (marginally) so 👍 I guess it will also use less disk storage which is probably a bigger consideration in terms of performance (because relatively few of the rows will be null).

heathd (approved):

PR #1769, opened 2019-09-25
PP-5635: add docs about new task in api_specifiation

jankowiakmaria:


docs/api_specification.md
```diff
@@ -22,6 +22,43 @@ Content-Type: application/json
 ```
 -----------------------------------------------------------------------------------------------------------
 
+## POST /v1/tasks/emitted-events-sweep
+
+During the state transition event connector puts an event in in-memory queue (and database) which is then picked up by
```
sfount: Super minor 
```suggestion
During the state transition event connector puts an event in an in-memory queue (and database) which is then picked up by
```

docs/api_specification.md
```diff
@@ -22,6 +22,43 @@ Content-Type: application/json
...
+the background process to emit the event to SQS.
+If the process is interrupted there is a database record which indicates that the event has been
+put in in-memory queue, but not yet emitted to the SQS.
```
sfount: Very minor
```suggestion
put in an in-memory queue, but not yet emitted to the SQS.
```

sfount (dismissed):

sfount (approved):

PR #1771, opened 2019-09-26
PP-5651 Convert CardType To Lower Case On Serialise

rorymalcolm:
- Converts the cardtype to lower case from the upper case enum when it
is serialised to JSON


stephencdaly (commented):
Should there be a test somewhere that would need updating for this?

stephencdaly (approved):
🏅

PR #1772, opened 2019-09-26
PP-5651 Fix broken test

stephencdaly:
A broken test managed to get in because two interdependant PRs were
merged close together, so CI didn't rebuild. Fix this.


rorymalcolm (approved):
added herobrine

PR #1774, opened 2019-09-27
PP-5662 Send debit/credit card type to ledger

alexbishop1:
When emitting a `PaymentDetailsEntered` event to ledger, send the card type (as `"card_type"`) in the event details. The value will be either `"DEBIT"` or `"CREDIT"` or nothing at all (if we can’t be
certain).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsEnteredEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsEnteredEventDetails.java
@@ -32,31 +32,26 @@ public class PaymentDetailsEnteredEventDetails extends EventDetails {
     private final String wallet;
     private final Long totalAmount;
 
-    private PaymentDetailsEnteredEventDetails(Long corporateSurcharge, String email, String cardType, String cardBrand,
-                                              String cardBrandLabel, String gatewayTransactionId, String firstDigitsCardNumber,
-                                              String lastDigitsCardNumber, String cardholderName, String expiryDate,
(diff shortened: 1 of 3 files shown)


rorymalcolm (dismissed):
LGTM!

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsEnteredEventDetails.java
```diff
@@ -31,7 +32,7 @@
     private final String wallet;
     private final Long totalAmount;
 
-    private PaymentDetailsEnteredEventDetails(Long corporateSurcharge, String email, String cardBrand,
+    private PaymentDetailsEnteredEventDetails(Long corporateSurcharge, String email, String cardType, String cardBrand,
```
danworth: Since this is private and only called by the builder.build() it would seem preferable to just pass the builder as a single argument rather than deconstructing its contents and having to keep adding each new element to this already long constructor which is error prone. Nothing you've done but perhaps this is an opportunity to change it.

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsEnteredEventDetails.java
```diff
@@ -216,11 +220,16 @@ Builder withEmail(String email) {
             return this;
         }
 
+        Builder withCardType(String cardType) {
```
danworth, line 223: Seems a shame not to use `CardType` for greater type safety even if the field itself remains as `String`.

danworth (approved):
Looks good.

PR #1776, opened 2019-09-27
BAU: Add Sandbox Visa prepaid unknown credit or debit unknown card

alexbishop1:
Add the test card number 4000000000000010, which card ID says is a Visa card for which it is unknown whether it is prepaid and for which it is unknown whether it is a credit or debit card but which is not a corporate card, to sandbox to assist services’ testing.


rorymalcolm (approved):
LGTM

PR #1783, opened 2019-10-01
PP-5670: Specify the consumer when running contract test

oswaldquek:
Currently when a consumer build is kicked off, the provider contract test will
run in its entirety. For example, in a frontend build, the connector provider
contract tests will run as well. However, what is run is:

frontend->connector
ledger->connector
publicapi->connector
selfservice->connector

We really only want to run frontend->connector in this case.

This commit will allow a dynamic creation of the test suite to run based on a
CONSUMER system property. This will mean shorter run times when a provider
contract test is run as part of a specific consumer build.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation


stephencdaly (approved):
Looks good 🏅

PR #1789, opened 2019-10-03
PP-5672 Always authorise with payment intents

mrlumbu:
All issues have been ironed out with using payment intents, so
fully switch to using them to create and authorise payments.
All downstream operations are still backwards compatible with
both payment intents and charges. I will switch them over in
subsequent PRs.


mrlumbu:
To help reviewing:
- Reassure yourself that the tests I have changed or removed only concerned the 'old' way of authorising charges
- look at how I have changed the main method in `StripeAuthoriseHandler` from dealing with two cases to dealing with one, without changing any other logic

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeResourceAuthorizeIT.java
```diff
@@ -114,33 +101,10 @@ public void setup() {
...
-                .statusCode(BAD_REQUEST_400)
-                .body("message", contains("This transaction was declined."))
-                .body("error_identifier", is(ErrorIdentifier.GENERIC.toString()));
-
-        assertFrontendChargeStatusIs(externalChargeId, AUTHORISATION_REJECTED.toString());
-    }
-
+    
     @Test
     public void cardAuthorisationWithPaymentIntentsFailureShouldReturnBadRequest() {
```
sfount, line 109: Very minor but specific test names like this might benefit from being simplified to avoid confusion with looking for other 'Withs'.

(for example `cardAuthorisationFailureShouldReturnBadRequest`). 

This would have the benefit of showing the test is equivalent although the process has changed.

sfount (approved):
Change looks good to me 👍 happy path has been tested behind flag and is working well, all behaviour looks the same.

Test coverage looks equivalent but removes tests specific to the previous authorise flow.

Would be good to follow this release through carefully but should be good to go.

Great cleanup!

PR #1792, opened 2019-10-04
PP-5672 Make Stripe refund more robust

mrlumbu:
This PR mitigates a problem we have observed in the wild.
The problem is that the following can happen:
- refund request to Stripe succeeds
- transfer request from connected account to platform account fails
- we as GOV.UK Pay are left out of pocket
- from service's point of view refund seems to be an error, yet the
end user has recieved the refund, and the service still recieves the
funds as part of next payout (since it is still credited to their account)

To avoid this, the `StripeRefundHandler` now does the following:
- transfer funds from connect account to platform account (if this
fails, refund fails, and no harm done)
- attempt to refund payment
- if refund step fails, reverse original transfer
- iff reversal fails, log error that will be picked up by Splunk alerting

The general idea is that we try reversable things before non-reversable things,
so that in the case of failure of non-reversible operation we can revert the
whole state without causing confusion.

I have also tidied up the tests somewhat. There were tests claiming to be about
testing specific kinds of failure from Stripe, which were essentially duplicate
since all Stripe failures are being mocked. I have left tests that check the
behaviour of refund handler for main failure modes that the handler can
discriminate between.

Finally, I have removed handling of 'non-payment intents' based payments
as we now only process payments using payment intents API.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
@@ -17,10 +17,9 @@ import uk.gov.pay.connector.gateway.stripe.json.StripeTransferResponse;
 import uk.gov.pay.connector.gateway.stripe.request.StripeGetPaymentIntentRequest;
 import uk.gov.pay.connector.gateway.stripe.request.StripeRefundRequest;
 import uk.gov.pay.connector.gateway.stripe.request.StripeTransferInRequest;
+import uk.gov.pay.connector.gateway.stripe.request.StripeTransferReversalRequest;
 import uk.gov.pay.connector.gateway.stripe.response.StripeRefundResponse;
 import uk.gov.pay.connector.util.JsonObjectMapper;
-
-import java.util.Optional;
 
 import static javax.ws.rs.core.Response.Status.Family.CLIENT_ERROR;
 import static javax.ws.rs.core.Response.Status.Family.SERVER_ERROR;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
```diff
@@ -46,60 +45,63 @@ public StripeRefundHandler(
...
+        }
+        return refundResponse;
+    }
+
+    private void reverseTransfer(RefundGatewayRequest request, String transferId) throws GatewayException.GenericGatewayException, GatewayErrorException, GatewayException.GatewayConnectionTimeoutException{
+        try {
+            StripeTransferReversalRequest stripeRefundRequest = StripeTransferReversalRequest.of(transferId, request, stripeGatewayConfig);
+            client.postRequestFor(stripeRefundRequest);
+        } catch (GatewayException e) {
+            logger.error("Refund failed for refund gateway request {}. Reason: {}",
```
sfount, line 96: Should this read "Stripe transfer reversal failed"? Refund failed could be conflated with the error log from the previous step

sfount (commented):
Code looks good to me and lines up with the proposed solution in the ticket. Minor comment about logging. 

Should we have a unit or integration test for the `refund` method that just verifies that that transfer reversal is definitely invoked if the refund step throws an error? Ensuring that critical path might help if this ever needs to be changed in the future.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
mrlumbu: We could also log that. I wanted to keep the wording that the alert is looking for which I think is this (I should double check)

mrlumbu:
Re. tests - see https://github.com/alphagov/pay-connector/blob/f2577f08c4823f291bd86f4aaf6aa4ec16bd0c3b/src/test/java/uk/gov/pay/connector/gateway/stripe/StripeRefundHandlerTest.java
I think that's what you're talking about no?

sfount:
> Re. tests - see src/test/java/uk/gov/pay/connector/gateway/stripe/StripeRefundHandlerTest.java
> I think that's what you're talking about no?

The diff was large so GitHub collapsed it 🤦‍♂ . Yeah that's exactly it

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
mrlumbu: I'll do this in a separate PR

sfount (approved):

PR #1797, opened 2019-10-07
PP-561 Return Card Type In V2 Endpoint

rorymalcolm:
- Returns the `card_type` in the getCharges v2 endpoint


danworth (commented):
Is there a test to go with this, how do you know this works?

stephencdaly (dismissed):
Can you update a test for this?

src/main/java/uk/gov/pay/connector/charge/model/domain/Transaction.java
```diff
@@ -210,6 +214,10 @@ public String getCardBrand() {
     public String getCardBrandLabel() {
         return cardBrandLabel;
     }
+    
+    public String getCardType() {
```
danworth: Seems this is optional and later on you do a `!= null` so this might be better returning `Optional<String>`

src/main/java/uk/gov/pay/connector/charge/service/TransactionSearchStrategy.java
```diff
@@ -60,6 +61,9 @@ public ChargeResponse buildResponse(UriInfo uriInfo, Transaction transaction) {
         PersistedCard cardDetails = new PersistedCard();
         cardDetails.setCardBrand(transaction.getCardBrandLabel());
         cardDetails.setCardHolderName(transaction.getCardHolderName());
+        if(transaction.getCardType() != null) {
```
danworth: If you make the return type for `getCardType` `Optional<String>` you can do something like:
`transaction.getCardType().map(CardType::valueOf).ifPresent(cardDetails::setCardType);`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -362,6 +362,7 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
         if (chargeEntity.getCardDetails() != null) {
             persistedCard = chargeEntity.getCardDetails().toCard();
             persistedCard.setCardBrand(findCardBrandLabel(chargeEntity.getCardDetails().getCardBrand()).orElse(""));
+            persistedCard.setCardType(chargeEntity.getCardDetails().getCardType());
```
danworth: I don't think is necessary. `.toCard();` sets the card type so it will already be set won't it? For what its worth I don't think it even needs the `setCardBrand` besides it defaults to `""` rather than `null` - might be worth looking into and cleaning up?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiV2ResourceIT.java
```diff
@@ -203,8 +204,8 @@ public void shouldGetExpectedCharge_whenOnlySpecifiedPaymentStates() {
     @Test
     public void shouldFilterTransactionsByCardHolderName() {
         String cardHolderName = "Mr. PayMcPayment";
-        addChargeAndCardDetails(nextLong(), CREATED, ServicePaymentReference.of("ref-1"), "ref", now(), "", "http://service.url/success-page/", "aaa@bbb.test", cardHolderName, "1234");
-        addChargeAndCardDetails(nextLong(), AUTHORISATION_SUCCESS, ServicePaymentReference.of("ref-1"), "ref", now(), "", "http://service.url/success-page/", "aaa@bbb.test", cardHolderName, "1234");
+        addChargeAndCardDetails(nextLong(), CREATED, ServicePaymentReference.of("ref-1"), "ref", now(), "", "http://service.url/success-page/", "aaa@bbb.test", cardHolderName, "1234", "CREDIT");
```
danworth, line 207: These all use `CREDIT`, perhaps you could use a mix of values for a little greater coverage?

danworth (changes requested):
Few little things to consider, could you also update `TransactionDaoIT` to confirm the read of the new value?

danworth (approved):
Looks good 👍

PR #1799, opened 2019-10-07
PP-5672 Make refunds backwards compatible

mrlumbu:
Reinstate logic to do refund either using payment intent id or
just the charge id based on form of id.


stephencdaly (dismissed):
Looks good

stephencdaly (approved):

PR #1805, opened 2019-10-09
PP-5681 send card brand label with telephone payment event

SandorArpa:
## WHAT YOU DID
- send card_brand_label with the payment notification created event to ledger (see PP-5658)


kbottla (approved):
LGTM!

PR #1809, opened 2019-10-10
PP-5672 Revert Stripe refund logic

mrlumbu:
We have run into issues with the way Stripe refund currently works.
We try to do this
- transfer from connect account
- refund payment from platform account

- if refund fails reverse the transfer

However, it turns out we cannot reverse the transfer easily due to the
indebtedness of our platform account. So revert to the original order -
refund first then transfer. We will still alert on any failures and can
resolve them manually.


sfount (approved):
Change LGTM, I think this reverts correctly back to the original method. Would be good to get a second opinion from @kbottla just to double check it looks like nothing has been missed.

kbottla (commented):
LGTM!

sfount:
Looks like CI can handle this now. Thanks @kbottla

PR #1811, opened 2019-10-11
PP-5204 Limit processing fee to gross charge amount

sfount:
This will prevent us from attempting to transfer 0 or negative amounts on the Stripe platform. This sets a policy of recovering "as much fee as possible" from the payment without going over the charge limit. 

This would allow services to many any payments (down to 1p) on Stripe with the Platform covering loss of additional fee. 

Discussions around if this is a user need or can be worked around with limiting payments on Stripe to a minimum value (~20p) should be completed with this as a possible route.


sfount:
I believe this idea was shelved in favour of limiting to 20p once services have updated guidance on their end.

PR #1812, opened 2019-10-11
[PP-5704] update column not to be insertable or updatable by persistance provider

jankowiakmaria:
Currently the connector nodes generate timestamps that are later put in the db.
This can sometimes cause the problem, when the nodes have different values of time.
In order to mitigate the problem we're going to let db set the timestamp values for
ChargeEvents.
The way we currently handle timestamps is causing some problems in the local environment (if
the timezone is different than UTC).
The long term solution should be to convert all timestamps (at the moment all of them are without
a timzone info) to timestamps with timezone.
The fix put in this PR allows connector to read the timestamp data as UTC regardless the timezone
the app runs in (the explaination can be found here:
https://stackoverflow.com/questions/41618165/how-to-specify-utc-timezone-for-spring-boot-jpa-timestamp).

Changes (tl;dr):
* ChargeEventEntity.updated column changed to non-insertable and non-updatable (db sets the value)
  * replaced its converter to be able to read the value as UTC regardless the jvm timezone
* update countChargesForImmediateCapture (used for metrics only) dao method to pass the cutoff date in UTC
* update DatabaseHelper.addEvent to insert timestamp in UTC

### DIFF
--- a/src/main/java/uk/gov/pay/connector/chargeevent/model/domain/ChargeEventEntity.java
+++ b/src/main/java/uk/gov/pay/connector/chargeevent/model/domain/ChargeEventEntity.java
@@ -42,17 +42,18 @@ public class ChargeEventEntity extends AbstractVersionedEntity {
     @Convert(converter = UTCDateTimeConverter.class)
     private ZonedDateTime gatewayEventDate;
 
-    @Convert(converter = UTCDateTimeConverter.class)
+    @Column(insertable = false, updatable = false)
+    @Convert(converter = LocalDateTimeConverter.class)
     private ZonedDateTime updated;
 
     protected ChargeEventEntity() {
     }
 
-    public ChargeEventEntity(ChargeEntity chargeEntity, ChargeStatus chargeStatus, ZonedDateTime updated, Optional<ZonedDateTime> gatewayEventDate) {
+    public ChargeEventEntity(ChargeEntity chargeEntity, ChargeStatus chargeStatus, Optional<ZonedDateTime> updated, Optional<ZonedDateTime> gatewayEventDate) {
         this.chargeEntity = chargeEntity;
         this.status = chargeStatus;
         this.gatewayEventDate = gatewayEventDate.orElse(null);
-        this.updated = updated;
+        this.updated = updated.orElse(null);
     }
 
     public Long getId() {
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/chargeevent/model/domain/LocalDateTimeConverter.java
```diff
@@ -0,0 +1,29 @@
+package uk.gov.pay.connector.chargeevent.model.domain;
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Converter;
+import java.sql.Timestamp;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+@Converter
+public class LocalDateTimeConverter implements AttributeConverter<ZonedDateTime, Timestamp> {
```
kbottla, line 10: May be a test to see it does what it says ?

src/main/java/uk/gov/pay/connector/chargeevent/model/domain/ChargeEventEntity.java
```diff
@@ -42,17 +42,18 @@
     @Convert(converter = UTCDateTimeConverter.class)
     private ZonedDateTime gatewayEventDate;
 
-    @Convert(converter = UTCDateTimeConverter.class)
+    @Column(insertable = false, updatable = false)
+    @Convert(converter = LocalDateTimeConverter.class)
```
kbottla, line 46: Looks good. But as discussed, we will need to review the how we set/read ZonedTimeStamp across codebase sooner rather than later.

jankowiakmaria:
There is one more thing missing - I thought I looked at it, but I must have missed it.
Refresh the ChargeEventEntity after persisting so that StateTransitionService has updated timestamp.

src/main/java/uk/gov/pay/connector/chargeevent/model/domain/ChargeEventEntity.java
jankowiakmaria: Yes.

src/main/java/uk/gov/pay/connector/chargeevent/model/domain/LocalDateTimeConverter.java
jankowiakmaria: Adding it.

jankowiakmaria:
@kbottla Updated.

kbottla (approved):

PR #1815, opened 2019-10-15
PP-5704/BAU: remove unused dao method

jankowiakmaria:
* The `findChargesForCapture` method is only used in tests
  - removing the method and relevant tests


rorymalcolm (approved):
Good 👍

PR #1816, opened 2019-10-15
PP-5702 telephone payments with success to be in CAPTURE SUBMITTED state

SandorArpa:
## WHAT YOU DID
- change telephone payments success state from AUTHORISATION_SUCCESS to CAPTURE_SUBMITTED as this reflects the true state of the transaction
- update relevant tests


rorymalcolm (approved):
👍

PR #1818, opened 2019-10-16
PP-5714: Configure connector to use json logging

oswaldquek:
This won't break any splunk searches as the search terms used by searches relating to connector are still valid. These are `account="*prod*"`, `service="*docker*"`, and piping search results to `rex field=_raw`.


oswaldquek:
do not merge yet. need to ascertain if this will affect any splunk searches.

stephencdaly (approved):

PR #1824, opened 2019-10-17
PP-5725 Add sentry config

mrlumbu:
See https://github.com/alphagov/pay-direct-debit-connector/pull/745


danworth (approved):
Looks good 👍

PR #1829, opened 2019-10-18
PP-5666: add pre-filled card details to payment created event

jankowiakmaria:
Why?
To allow searches on pre-filled data in ledger

Why is it complicated?
* it’s impossible to distinguish the new event from event backfilling
* once user fills the card details the pre-filled data is lost

Why does this approach work?
During the backfill the states happening after AUTHORISATION_READY event
(AUTHORISATION_READY won’t appear in db) will cause PaymentDetailsEntered
to be emitted. So if there is a charge (in whatever state) that doesn’t
have any charge events in the post-AUTHORISATION_READY states
(PaymentGatewayStateTransitions.java#L112-L120) then if we have information
about card details we can freely add it to PaymentCreated event.
In the normal event flow (not a backfill) this conditions should apply too
(there shouldn’t be enough time to transition to the next state (any of the
post-auth-ready ones) and if there is then we’ve lost the pre-filled data
anyway so we shouldn’t add them to the PaymentCreated payload.

Changes:
* updated PaymentDetailsEntered event to include card-details and email
* added relevant unit tests

Todo (in separate PR) - updates to the pact test

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentCreatedEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentCreatedEventDetails.java
@@ -1,35 +1,20 @@
 package uk.gov.pay.connector.events.eventdetails.charge;
 
-import com.google.common.collect.ImmutableList;
 import uk.gov.pay.commons.model.charge.ExternalMetadata;
 import uk.gov.pay.connector.charge.model.AddressEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
+import uk.gov.pay.connector.common.model.domain.PaymentGatewayStateTransitions;
 import uk.gov.pay.connector.events.eventdetails.EventDetails;
 
-import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentCreatedEventDetails.java
```diff
@@ -44,7 +83,35 @@ public static PaymentCreatedEventDetails from(ChargeEntity charge) {
...
+        }
+
+        return builder.build();
+    }
+
+    private static boolean isInCreatedState(ChargeEntity charge) {
+        return ChargeStatus.CREATED.equals(ChargeStatus.fromString(charge.getStatus()));
+    }
+
+    private static boolean hasNotGoneThroughAuthorisation(ChargeEntity charge) {
```
kbottla, line 84: Is it possible to use PaymentGatewayStateTransitions graph to find successor nodes ? Hardcoding list may make it inconsistent for any future changes to graph

`getTransitionsForStatus(AUTHORISATION_READY);`

 ```
PaymentGatewayStateTransitions

public <T extends Event> List<ChargeStatus> getTransitionsForStatus(ChargeStatus fromStatus) {
        return graph.successors(fromStatus)
                .stream()
    }
```

kbottla (dismissed):
Looks good to me. Added a suggestion about using graph to get statuses post AUTH_READY. Happy to discuss it further

kbottla (approved):

PR #1832, opened 2019-10-21
PP-5743 Telephone payments - Additional card type

kbottla:
## WHAT
- Telephone payments should also support JCB card type


SandorArpa (approved):
LOKTM

PR #1836, opened 2019-10-23
PP-5666: update payment created message for pact tests

jankowiakmaria:
The https://github.com/alphagov/pay-connector/pull/1829 change
adds additional data to PaymentCreated message.

Change:
* update pact provider to reflect the change to the message

### DIFF
--- a/src/test/java/uk/gov/pay/connector/pact/QueueMessageContractTest.java
+++ b/src/test/java/uk/gov/pay/connector/pact/QueueMessageContractTest.java
@@ -57,6 +57,7 @@ public class QueueMessageContractTest {
         ChargeEntity charge = ChargeEntityFixture.aValidChargeEntity()
                 .withExternalMetadata(new ExternalMetadata(ImmutableMap.of("key", "value")))
                 .withCorporateSurcharge(55L)
+                .withCardDetails(anAuthCardDetails().getCardDetailsEntity())
                 .build();
 
         PaymentCreated paymentCreatedEvent = new PaymentCreated(


src/test/java/uk/gov/pay/connector/pact/QueueMessageContractTest.java
```diff
@@ -57,6 +57,7 @@ public String verifyPaymentCreatedEvent() throws Exception {
         ChargeEntity charge = ChargeEntityFixture.aValidChargeEntity()
                 .withExternalMetadata(new ExternalMetadata(ImmutableMap.of("key", "value")))
                 .withCorporateSurcharge(55L)
+                .withCardDetails(anAuthCardDetails().getCardDetailsEntity())
```
kbottla, line 60: Do we already send `email` ? and might be setting fixture with more details than required ? (i.e., I think card details entity also included first_digits_card_number, last_digits_card_number.. which are not applicable for this particular event)

jankowiakmaria: Yep, we do send email - I can be explicit though.
Yes, we do - copied this line from PaymentDetailsEntered. It should not be a problem though, because PaymentCreated event will only include https://github.com/alphagov/pay-connector/blob/89519e7d39968a2b3c411ac664b83381c6e08454/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentCreatedEventDetails.java#L30-L36 fields (it won't emit `first|last_digits_card_number`).

kbottla: `email` - is fine as long as it is not based on random string

kbottla (approved):

src/test/java/uk/gov/pay/connector/pact/QueueMessageContractTest.java
jankowiakmaria: nah, it's static:
```java
private String email = "test@email.invalid";
```

PR #1844, opened 2019-10-25
PP-5781 Transition from 3DS flow to external cancel request

sfount:
A number of requests are failing because we do not currently support
transitioning from 3DS REQUIRED to system cancelled.

This has reportedly caused issues in the past with our expiry sweeper
however these issues have been addressed.

It seems to be a valid transition tested on Worldpay accounts.
Determining how sensible this is for all Gateways might be required.


sfount:
Have confirmed this behaviour in the `test-12` environment. If the payment is cancelled by the service while the user is on the 3DS challenge they will be redirected to "Your payment has been cancelled" page when they submit this.

kbottla (approved):

sfount:
Waiting for `PP-5781` to be prioritised before releasing this

stephencdaly:
can this be merged now?

sfount:
Looking into the additional non-ready pre-authorisation states:
`AUTHORISATION_SUBMITTED` - the code treats this state as an external error (EPDQ specific?), transitioning to successfully cancelled in this case wouldn't make sense
`AUTHORISATION_3DS_REQUIRED` - added in this PR
`ENTERING_CARD_DETAILS` - supported already
`CREATED` - supported already

Any post-authorisation states will need to go through `SYSTEM_CANCEL_READY`. 

The one exception is `AUTHORISATION_READY` which will need to factor in how far the gateway is through processing the authorisation, addressing this state will follow in future work.

PR #1849, opened 2019-10-28
PP-5757 Log things not found as 'info' rather than 'error'

danworth:
If Connector cannot find the item (e.g. charge, gatewayaccount or token) that
it has been asked for then log it at level `info` rather than `error`.
Whether the item not existing is actually an `error` should be determined by the
service that called Connector since it has the context to know whether it requires
greater attention.

## WHAT YOU DID
This is to help clean up error level logging for Sentry e.g. https://pay-sentry.cloudapps.digital/sentry/connector/issues/292/?query=is%3Aunresolved


stephencdaly (approved):
👍

PR #1850, opened 2019-10-28
PP-5757: Reclassify missing charges and transactions

oswaldquek:
These shouldn't be logged at ERROR level as it doesn't stop the application
from working.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
@@ -82,7 +82,7 @@ public class WorldpayNotificationService {
         }
 
         if (isTransactionIdBlank(notification)) {
-            logger.error("{} notification {} failed verification because it has no transaction ID", gatewayName(), notification);
+            logger.warn("{} notification {} failed verification because it has no transaction ID", gatewayName(), notification);
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
```diff
@@ -69,7 +69,7 @@ private void handle(SmartpayNotification notification) {
                 notification.getOriginalReference());
 
         if (!maybeCharge.isPresent()) {
-            logger.error("{} notification {} could not be evaluated (associated charge entity not found)",
+            logger.info("{} notification {} could not be evaluated (associated charge entity not found)",
```
stephencdaly: Would it be worth changing this log to log the transaction id we couldn't find, so we can look into why this might be? It seems odd that we would get notifications for charges that don't exist, it might be interesting to see what gateway account they're from, maybe someone has set up their notifications incorrectly...

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -82,7 +82,7 @@ public boolean handleNotificationFor(String ipAddress, String payload) {
         }
 
         if (isTransactionIdBlank(notification)) {
-            logger.error("{} notification {} failed verification because it has no transaction ID", gatewayName(), notification);
+            logger.info("{} notification {} failed verification because it has no transaction ID", gatewayName(), notification);
```
stephencdaly: have we seen this error before?

oswaldquek: we haven't seen any in the last 30 days so it's safe to downgrade to WARN actually.

oswaldquek:
@stephencdaly I've updated the PR. see the new individual commits for commit messages.

stephencdaly (approved):

PR #1851, opened 2019-10-28
PP-5757 Log the correct id when we process a Smartpay notification

stephencdaly:
We were logging the "pspReference", but this is a unique reference for the operation, rather than the Smartpay's reference for the charge. This is unhelpful as we don't have an easy way to relate it back to the charge. Instead, log the "originalReference" which is Smartpay's reference for the charge.


oswaldquek:
I wonder if it's a good idea to change https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java#L36-L51 to json logging as well, but this wouldn't be required for approval of this PR.

oswaldquek (approved):

PR #1852, opened 2019-10-28
PP-5757: Increase epdq capture timeout

oswaldquek:
We have been getting a lot of timeout exceptions when capturing epdq payments -
683 events on 25/10/2019 for example. Hopefully this will go away if we increase the timeout slightly.


stephencdaly (dismissed):

danworth (approved):
I think this is a good idea and its proposed by a few people now. Lets give it a try.

PR #1853, opened 2019-10-29
PP-5757 Don't log error when notification has already been processed

stephencdaly:
We log an error if we fail to transition a charge when processing a notification. We've seen a number of errors logged when we try to transition the charge into a state it's already in. In this case we've already processed the notification so it's fine to do nothing and log nothing.

We're still logging an error if we can't transition to a different state, however we should review if we see errors like this whether they are actually errors.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/exception/InvalidStateTransitionException.java
+++ b/src/main/java/uk/gov/pay/connector/common/exception/InvalidStateTransitionException.java
@@ -5,8 +5,22 @@ import uk.gov.pay.connector.events.model.Event;
 import static java.lang.String.format;
 
 public class InvalidStateTransitionException extends IllegalStateException {
+    
+    private final String currentState;
+    
+    private final String targetState;
+
+    public String getCurrentState() {
+        return currentState;
+    }
+
+    public String getTargetState() {
+        return targetState;
+    }
 
     public InvalidStateTransitionException(String currentState, String targetState, Event event) {
         super(format("Charge state transition [%s] -> [%s] not allowed [event={}]", currentState, targetState, event));
+        this.currentState = currentState;
+        this.targetState = targetState;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
@@ -28,8 +28,12 @@ public class ChargeNotificationProcessor {
         try {
             chargeService.transitionChargeState(chargeEntity, newStatus, gatewayEventDate);
         } catch (InvalidStateTransitionException e) {
-            logger.error("{} ({}) notification '{}' could not be used to update charge: {}",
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -28,8 +28,12 @@ public void invoke(String transactionId, ChargeEntity chargeEntity, ChargeStatus
         try {
             chargeService.transitionChargeState(chargeEntity, newStatus, gatewayEventDate);
         } catch (InvalidStateTransitionException e) {
-            logger.error("{} ({}) notification '{}' could not be used to update charge: {}",
-                    gatewayAccount.getGatewayName(), gatewayAccount.getId(), transactionId, e.getMessage());
+            // don't log an error if we're trying to transition to the same state as this can happen if we've already
+            // processed the notification
+            if (!e.getCurrentState().equals(e.getTargetState())) {
```
danworth, line 33: Might it be worth logging at info level if the statuses are the same, so there's some record we attempted to process it but ignored it because there's no update to be made (or is that captured elsewhere)?

src/main/java/uk/gov/pay/connector/common/exception/InvalidStateTransitionException.java
```diff
@@ -5,8 +5,22 @@
...
+
+    public String getCurrentState() {
+        return currentState;
+    }
+
+    public String getTargetState() {
+        return targetState;
+    }
 
     public InvalidStateTransitionException(String currentState, String targetState, Event event) {
```
danworth, line 21: Is there an appropriate type for these states, are they `ChargeStatus` perhaps, if there is it might be nice to use them?

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
stephencdaly: I talked to Dom about this, and we reckoned it's probably just unnecessary noise. If you think it's worthwhile, could add it in

src/main/java/uk/gov/pay/connector/common/exception/InvalidStateTransitionException.java
stephencdaly: Seems a bit much to convert them into a state when we're using them for such a simple task? The ChargeEntity deals with them as strings as it is.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
danworth: nah if you Dom don't think its necessary then 👍 just something I thought worth considering.

src/main/java/uk/gov/pay/connector/common/exception/InvalidStateTransitionException.java
danworth: No that seems fine, if they're already Strings.

danworth (approved):
Looks good 👍

PR #1854, opened 2019-10-29
PP-5785 Worldpay Notification - Send non 2xx if charge not found

kbottla:
## WHAT
- Returns non 2xx response for WorldpayNotification and when a charge is not found for Worldpay Transaction ID.
   Currently, we always have a charge (for successful authorisation) associated with worldpay transaction ID and we ignore notifications (send 200 response to worldpay) if charge is not found for corresponding worldpay transaction ID.

   This is an issue with telephone payments, where we can get worldpay CAPTURE notification, before we receive any payload for telphone payment. In such cases, charge related to telephone payment stays in CAPTURE_SUBMITTED state forever and can't be refunded.

   Returning non 2xx response will ensure worldpay retries the notification periodically for upto 7 days and we can receive telephone payment payload & worldpay notification for CAPTURE in any order.

- Also downgraded ERROR level logging to INFO when charge is not found for notification

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
@@ -89,10 +89,13 @@ public class WorldpayNotificationService {
         Optional<ChargeEntity> optionalChargeEntity = chargeDao.findByProviderAndTransactionId(gatewayName(),
                 notification.getTransactionId());
 
-        if (!optionalChargeEntity.isPresent()) {
-            logger.error("{} notification {} could not be evaluated (associated charge entity not found)",
+        if (optionalChargeEntity.isEmpty()) {
+            logger.info("{} notification {} could not be evaluated (associated charge entity not found)",
                     gatewayName(), notification);
-            return true;
+            // Respond with an error, which will cause worldpay to try to send the notification
+            // again later — this is necessary because sometimes we might receive a notification
+            // for a telephone payment before we know about the payment itself
+            return false;
         }
 
         if (isCaptureNotification(notification)) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -90,9 +90,9 @@ public boolean handleNotificationFor(String ipAddress, String payload) {
                 notification.getTransactionId());
 
         if (!optionalChargeEntity.isPresent()) {
```
alexbishop1: This can be `if (optionalChargeEntity.isEmpty())`

SandorArpa (dismissed):
LOKTM

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -90,9 +90,9 @@ public boolean handleNotificationFor(String ipAddress, String payload) {
...
-            logger.error("{} notification {} could not be evaluated (associated charge entity not found)",
+            logger.info("{} notification {} could not be evaluated (associated charge entity not found)",
                     gatewayName(), notification);
-            return true;
+            return false;
```
alexbishop1, line 98: Could you add a little comment before this line? Something like:

```java
// Respond to Worldpay with an error, which will cause them to try to send the
// notification again later — this is necessary because sometimes we might receive
// a notification for a telephone payment before we know about the payment itself
return false;
```

I’m not a huge fan of code comments but I think it would be useful in this case because the behaviour is surprising.

alexbishop1 (changes requested):
Looks good but I think a code comment would be valuable

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
kbottla: 👍 Updated

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
kbottla: 👍 Added comment

alexbishop1 (approved):

PR #1856, opened 2019-10-29
PP-5757 Downgrade log levels for Stripe notifications

stephencdaly:
- For a notification without a transaction id, this does not suggest anissue with our service, so log as WARN rather than ERROR.
- For a notification with a transaction id we can't find, only log this as info. It happens regularly that we can receive a notification that a charge is capturable before we've received and processed the response to an authorisation request, which means we won't have yet set the payment intent id on the charge and thus won't be able to find it.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
@@ -94,14 +94,14 @@ public class StripeNotificationService {
             StripePaymentIntent paymentIntent = toPaymentIntent(notification.getObject());
 
             if (isBlank(paymentIntent.getId())) {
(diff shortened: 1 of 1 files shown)


danworth:
For this one:
`For a notification with a transaction id we can't find, only log this as info. It happens regularly that we can receive a notification that a charge is capturable before we've received and processed the response to an authorisation request, which means we won't have yet set the payment intent id on the charge and thus won't be able to find it.`
Will we try and process the notification again later on? Is there any impact to simply ignoring this race condition?

stephencdaly:
We only process the notifications if the charge is in `AUTH 3DS REQUIRED` or `AUTH 3DS READY` status as they tell us whether the charge has been authorised. In this case there shouldn't be the possibility of a race, as the user would have had to go through the 3DS process after we've received the authorisation response and updated the charge to set the payment intent id on it. Even if there was, this isn't a useful error to check on if a lot of it is just noise so having it as an info log is as useful for identifying a payment we might have an issue with.

danworth (approved):

PR #1857, opened 2019-10-29
PP-5757 Don't log generic gateway errors at error level

stephencdaly:
This log just generates a lot of noise so remove it.
The main logs come from:
- entering a non-sandbox card number for a sandbox account
- entering a sandbox card number that causes an authorisation error
- entering a non-test card number for a worldpay test account
- entering an invalid card number with a valid bin range for a worldpay account
- entering an invalid expiry date for a worldpay account.

There are a few other opaque worldpay errors, but logging them here isn't a good way of telling if there is a problem with our system. We already log all authorisation responses at info level, and have a saved search for authorisation errors due to a payment method being disabled
in worldpay, which is a better way of determining if there is something we should act upon.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
@@ -171,7 +171,6 @@ public class CardResource {
             case GATEWAY_ERROR:
                 return serviceErrorResponse(error.getMessage());
             default:
-                logger.error("Charge {}: error {}", chargeId, error.getMessage());
                 return badRequestResponse(error.getMessage());
         }
     }


src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -171,7 +171,6 @@ private Response handleError(String chargeId, GatewayError error) {
             case GATEWAY_ERROR:
                 return serviceErrorResponse(error.getMessage());
             default:
-                logger.error("Charge {}: error {}", chargeId, error.getMessage());
```
danworth, line 174: I'm not sure about not logging anything. Sounds like at present most of the time when this is reached we don't really care about it but since this is the default case we might find that some important things fall through here which we wouldn't know about (or find if something else alerted us to a problem). Could this be downgraded to `warn` or perhaps introduce more `case`s to catch the known unimportant things (which could be ignored) and leave a default case that does log?

stephencdaly: I think logging in this place is just pretty useless. We log the response in other places.
For stripe in particular we do log gateway exceptions in the stripe-specific code which seems like a better place to do it if we should treat gateway errors as actual errors, unlike for Worldpay where we shouldn't
```logger.error("GatewayException occurred for charge external id {}, error:\n {}", request.getChargeExternalId(), e);```

danworth: ok cool, if its logged elsewhere makes sense.

danworth (approved):

PR #1858, opened 2019-10-29
PP-5757 Don't log Stripe declines at error level

stephencdaly:
We currently log when a 3DS Authorisation for Stripe is declined at error level. This is not an error, but BAU, so log as info.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
@@ -137,7 +137,7 @@ public class StripePaymentProvider implements PaymentProvider {
 
             if (e.getFamily() == CLIENT_ERROR) {
                 StripeErrorResponse stripeErrorResponse = jsonObjectMapper.getObject(e.getResponseFromGateway(), StripeErrorResponse.class);
-                logger.error("3DS Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
+                logger.info("3DS Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
                         request.getChargeExternalId(), stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), e.getStatus());
 
                 return Gateway3DSAuthorisationResponse.of(BaseAuthoriseResponse.AuthoriseStatus.REJECTED);
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
@@ -63,7 +63,7 @@ public class StripeAuthoriseHandler implements AuthoriseHandler {
             String chargeExternalId = request.getChargeExternalId();
 
             if ((e.getStatus().isPresent() && e.getStatus().get() == SC_UNAUTHORIZED) || e.getFamily() == SERVER_ERROR) {
-                logger.error("Authorisation failed for charge {}. Reason: {}. Status code from Stripe: {}.",
+                logger.error("Authorisation failed for charge {} due to an internal error. Reason: {}. Status code from Stripe: {}.",
                         chargeExternalId, e.getMessage(), e.getStatus().map(String::valueOf).orElse("no status code"));
                 GatewayError gatewayError = gatewayConnectionError("There was an internal server error authorising charge_external_id: " + chargeExternalId);
                 return responseBuilder.withGatewayError(gatewayError).build();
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
```diff
@@ -63,15 +63,15 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
             String chargeExternalId = request.getChargeExternalId();
 
             if ((e.getStatus().isPresent() && e.getStatus().get() == SC_UNAUTHORIZED) || e.getFamily() == SERVER_ERROR) {
```
danworth, line 65: Logging `SC_UNAUTHORIZED` at info makes perfect sense. I wonder where `SERVER_ERROR` should be treated separately and logged at `warn` or even `error` since its a not a BAU thing.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
```diff
@@ -137,7 +137,7 @@ private Gateway3DSAuthorisationResponse authorise3DSSource(Auth3dsResponseGatewa
 
             if (e.getFamily() == CLIENT_ERROR) {
                 StripeErrorResponse stripeErrorResponse = jsonObjectMapper.getObject(e.getResponseFromGateway(), StripeErrorResponse.class);
-                logger.error("3DS Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
+                logger.info("3DS Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
```
danworth, line 140: I think the condition checking `SC_UNAUTHORIZED` is catching the BAU case of the user simply being denied. This is catching all other 4XX codes so I'm wondering if this might be better as `warn` or even kept at `error` since I don't actually see any of these for the past month?

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
stephencdaly: hmm you're right although I think SC_UNAUTHORIZED also points to a legit error as it means we're unauthorised to make the request. I will put that back to error, and slightly improve the message to make this clear.

stephencdaly: updated

src/main/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProvider.java
stephencdaly: The fact it's returning a declined response, rather than error, I think we are treating it as a BAU response so shouldn't log an error. It also matches how we handle legitimate rejections for normal authorisations.

danworth: Any idea why we treat a client error 4XX between us and Stripe the same a BAU auth rejection? It seems like they're two different things, the BAU case we don't really care about but if suddenly there is some problem between us and Stripe I'd have thought we'd want to know.

danworth: Oh looks like because 4XX can be in relation to the user being declined eg.
`Authorisation failed for charge {chargeId}. Failure code from Stripe: card_declined, failure message from Stripe: Your card has insufficient funds.. Response code from Stripe: Optional[402]`

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
```diff
@@ -63,15 +63,15 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
+                logger.error("Authorisation failed for charge {} due to an internal error. Reason: {}. Status code from Stripe: {}.",
                         chargeExternalId, e.getMessage(), e.getStatus().map(String::valueOf).orElse("no status code"));
                 GatewayError gatewayError = gatewayConnectionError("There was an internal server error authorising charge_external_id: " + chargeExternalId);
                 return responseBuilder.withGatewayError(gatewayError).build();
             }
 
             if (e.getFamily() == CLIENT_ERROR) {
                 StripeErrorResponse stripeErrorResponse = jsonObjectMapper.getObject(e.getResponseFromGateway(), StripeErrorResponse.class);
-                logger.error("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
+                logger.info("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
```
danworth, line 74: So if I've understood this correctly we're treating a 4XX response from Stripe as the authorisation being rejected, which is BAU and I agree this can just be `info`.

danworth (approved):
Looks good 👍

PR #1860, opened 2019-10-29
PP-5757 Increase ePDQ request timeout for refunds

stephencdaly:
We've seen a number of ePDQ refund requests time out. Refunds are a synchronous operation that aren't retried, so we should have the same timeout as for authorisations and cancellations.


alexbishop1 (approved):
Odd that this was different in the first place

PR #1861, opened 2019-10-29
PP-5757: Emails shouldn't be sent

oswaldquek:
If the email collection mode is OPTIONAL and there is no email associated with
the charge.

This fixes https://pay-sentry.cloudapps.digital/sentry/connector/issues/85/

Also adding tests for where email collection mode is OFF as it would be possible for a charge to have an email address whilst the email
collection mode is OFF. This can happen if the collection mode was MANDATORY, a
charge was created (but not yet captured) and the service turned the email
collection mode to OFF.


stephencdaly (approved):
Looks good

PR #1862, opened 2019-10-29
PP-5757 Improve 3DS Authorisation response

stephencdaly:
We were doing as follows:
- If the authorisation status was AUTHORISED or AUTH_3DS_READY we
returned a 200 response frontend.
- If the authorisation status was REJECTED or ERROR we returned a 400
response to frontend.
- If the authorisation status was anything else, we returned a 500 to frontend, causing it to show the technical difficulties page.

However, this did not account for the CANCELLED status which is a valid outcome for Worldpay 3DS authorisations, which we have observed come back.

Improve this by still returning a 200 for AUTHORISED or AUTH_3DS_READY statuses, and a 500 only for the EXCEPTION status, and a 400 for any other status.

Sentry issue: https://pay-sentry.cloudapps.digital/sentry/connector/issues/109

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
@@ -277,7 +277,7 @@ public class WorldpayCardResourceIT extends ChargingITestBase {
     public void shouldReturnStatus500_WhenAuthorisationCallThrowsException() {
         String chargeId = createNewCharge(AUTHORISATION_3DS_REQUIRED);
 
-        String expectedErrorMessage = "Failed";
+        String expectedErrorMessage = "There was an error when attempting to authorise the transaction.";
         worldpayMockClient.mockServerFault();
 
         givenSetup()
@@ -296,7 +296,7 @@ public class WorldpayCardResourceIT extends ChargingITestBase {
     public void shouldReturnStatus500_AWorldpayPaResParseError() {
         String chargeId = createNewCharge(AUTHORISATION_3DS_REQUIRED);
 
-        String expectedErrorMessage = "Failed";
+        String expectedErrorMessage = "There was an error when attempting to authorise the transaction.";
         worldpayMockClient.mockAuthorisationPaResParseError();
 
         givenSetup()
--- a/src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/response/Gateway3DSAuthorisationResponse.java
@@ -33,11 +33,6 @@ public class Gateway3DSAuthorisationResponse {
         return new Gateway3DSAuthorisationResponse(authorisationStatus, null, "");
     }
 
-    public boolean isDeclined() {
-        return authorisationStatus == BaseAuthoriseResponse.AuthoriseStatus.REJECTED ||
-                authorisationStatus == BaseAuthoriseResponse.AuthoriseStatus.ERROR;
-    }
-
     public boolean isSuccessful() {
         return authorisationStatus == BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED
                 || authorisationStatus == BaseAuthoriseResponse.AuthoriseStatus.AUTH_3DS_READY;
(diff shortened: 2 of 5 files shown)


src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
```diff
@@ -274,36 +274,17 @@ public void shouldReturnStatus400_WhenAuthorisationFails() {
...
-                .statusCode(INTERNAL_SERVER_ERROR.getStatusCode())
-                .contentType(JSON)
-                .body("message", contains(expectedErrorMessage))
-                .body("error_identifier", is(ErrorIdentifier.GENERIC.toString()));
-
-        assertFrontendChargeStatusIs(chargeId, AUTHORISATION_ERROR.getValue());
-    }
-
-    @Test
-    public void shouldReturnStatus500_AWorldpayPaResParseError() {
```
stephencdaly, line 296: removed this test as it's essentially the same as above - not worthy of it's own resource test

oswaldquek:
Is there a link to the connector issue in Sentry?

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
```diff
@@ -274,36 +274,17 @@ public void shouldReturnStatus400_WhenAuthorisationFails() {
...
-
-        String expectedErrorMessage = "Failed";
-        worldpayMockClient.mockAuthorisationPaResParseError();
-
-        givenSetup()
-                .body(buildJsonWithPaResponse())
-                .post(authorise3dsChargeUrlFor(chargeId))
-                .then()
-                .statusCode(INTERNAL_SERVER_ERROR.getStatusCode())
+                .statusCode(BAD_REQUEST.getStatusCode())
```
oswaldquek: It's not clear to me why this scenario returns a 400. The test is mocking out a server fault (line 281) but I don't see how this maps to the PR description where we're accounting "for the CANCELLED status which is a valid outcome for Worldpay 3DS".

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
oswaldquek: It doesn't look like it's the same as the above test. The above test is mocking out a server fault while this one is mocking out a worldpay auth PaRes response...?

oswaldquek (commented):
some questions...

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
stephencdaly: Hmm I guess maybe not, I'd say the difference between them should be tested on a service level, but yanking out the resource test without making sure there's a service level test is a bad idea. I will put this back.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
stephencdaly: I think it should return a 400, because we handle the error internally and transition the charge to an appropriate state. There's no need to return a 500 to frontend when we've suitably handled the exception, as frontend is able to cope with an authorisation error state by showing the user an appropriate screen. It turns out that they are shown the same screen regardless so behaviour there isn't changed.

oswaldquek: ok thanks for clarifying

oswaldquek: hmm, actually, if it's a server fault it seems like returning 400 is not the right thing to do... it's technically a 500.

stephencdaly:
I've modified so we still return a 500 for the "EXCEPTION" authorisation result, but 400 for any other result that isn't success.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
stephencdaly: Updated so an internal error we catch does still return a 500

oswaldquek (approved):
Looks good

PR #1864, opened 2019-10-30
PP-5757 Allow CAPTURE SUBMITTED to CAPTURE ERROR transition

stephencdaly:
We can receive Smartpay notifications informing us that a capture didn't succeed. Currently, as this state transition is disallowed we do nothing to the charge when we receive this notification. The charge therefore remains stuck in the "CAPTURE SUBMITTED" state.

This also means that a ticket is not created in Zendesk to inform us of the capture error, which should be the case.


oswaldquek (approved):

danworth (approved):
This look sensible if we're being told the charge has an error post submission we should update it to reflect that. Would be good to keep an eye on this and understand why they're failing with Smartpay after authorisation has succeeded.

kbottla:
@stephencdaly  @danworth  This might potentially have an impact on services communicating with Users about the Payment.
When the charge is in `CAPTURE SUBMITTED` we return `Success` status to service which may be communicated to users. If a `CAPTURE SUBMITTED` is allowed to `CAPTURE ERROR` then service may be in the dark about failed payments ?

PR #1865, opened 2019-10-30
PP-5757 Only log error on PSP status code of 500+

mrlumbu:
The GatewayClient was logging every non-success status code as an error. Some PSPs use these codes to indicate BAU info - e.g. Stripe declines a card with a 402 response. So only log an error at this point if we get a 500 or 'more'.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
@@ -24,6 +24,7 @@ import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
 import static javax.ws.rs.core.Response.Status.OK;
 
 public class GatewayClient {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
```diff
@@ -70,10 +71,12 @@ public GatewayClient(Client client, MetricRegistry metricRegistry) {
...
             } else {
-                logger.error("Gateway returned unexpected status code: {}, for gateway url={} with type {} with order request type {}",
-                        statusCode, url, account.getType(), request.getOrderRequestType().toString());
-                incrementFailureCounter(metricRegistry, metricsPrefix);
-                throw new GatewayErrorException("Unexpected HTTP status code " + statusCode + " from gateway", gatewayResponse.getEntity(), statusCode);
+                if (statusCode >= INTERNAL_SERVER_ERROR.getStatusCode()) {
+                    logger.error("Gateway returned unexpected status code: {}, for gateway url={} with type {} with order request type {}",
+                            statusCode, url, account.getType(), request.getOrderRequestType());
+                    incrementFailureCounter(metricRegistry, metricsPrefix);
+                }
```
oswaldquek, line 81: do we want to log things at WARN/INFO is the statusCode is less than 500?

mrlumbu: yeah that's probably a good idea for now

oswaldquek (approved):

PR #1867, opened 2019-10-30
PP-5757: Log gateway account username for bad smartpay notifications

oswaldquek:
We are seeing quite a lot of smartpay notifications for which there is no
charge entity. After some investigation, we can confidently rule out that we're
not storing the smartpay pspReference from the authorisation response. A theory
could be services using their Smartpay account for payments outside of Pay so
we get notifications about payments we don’t know about. It'd be interesting to
see if this is the case so log the username for their gateway account.


oswaldquek:
Do not merge yet. Pending further investigation.

PR #1871, opened 2019-10-31
PP-5779 (PATCH) Backfill force emit live and card brand label

sfount:
Depends on https://github.com/alphagov/pay-connector/pull/1870.

Forces emission of `PaymentCreated` and `PaymentDetailsEntered` events. This will apply new schema to all Ledger events allowing transactions to be built with up to date info. 

Key changes to `HistoricalEventEmitter.java `.

Parameter for backfill task is `force_patch`, expecting `true` value with everything else defaulting to false.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
@@ -100,6 +100,13 @@ public class HistoricalEventEmitter {
         processPaymentDetailEnteredEvent(chargeEventEntities);
     }
 
+    public void processPatchPaymentCreatedAndPaymentDetailsEnteredEvents(ChargeEntity charge) { 
+        List<ChargeEventEntity> chargeEventEntities = getSortedChargeEvents(charge);
+
+        processSingleChargeStateTransitionEvent(charge.getId(), ChargeStatus.UNDEFINED, chargeEventEntities.get(0));
+        processPaymentDetailEnteredEvent(chargeEventEntities);
+    }
+
     @Transactional
     public void processRefundEvents(ChargeEntity charge) {
         List<RefundHistory> refundHistories = refundDao.searchAllHistoryByChargeId(charge.getId());
--- a/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterTask.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterTask.java
@@ -15,6 +15,7 @@ import uk.gov.pay.connector.queue.StateTransitionService;
 import uk.gov.pay.connector.refund.dao.RefundDao;
 
 import java.io.PrintWriter;
+import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.SynchronousQueue;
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -100,6 +100,13 @@ public void processPaymentEvents(ChargeEntity charge) {
         processPaymentDetailEnteredEvent(chargeEventEntities);
     }
 
+    public void processPatchPaymentCreatedAndPaymentDetailsEnteredEvents(ChargeEntity charge) { 
+        List<ChargeEventEntity> chargeEventEntities = getSortedChargeEvents(charge);
+
+        processSingleChargeStateTransitionEvent(charge.getId(), ChargeStatus.UNDEFINED, chargeEventEntities.get(0));
```
kbottla, line 106: We may need to filter for `CREATED` event and emit these two events as this process is not applicable for telephone payments

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterTask.java
```diff
@@ -46,12 +57,18 @@ public HistoricalEventEmitterTask(HistoricalEventEmitterWorker worker, Environme
     public void execute(ImmutableMultimap<String, String> parameters, PrintWriter output) throws Exception {
         Long startId = getParam(parameters, "start_id").orElse(0);
         final OptionalLong maybeMaxId = getParam(parameters, "max_id");
+        Long patchBackfillParam = getParam(parameters, "force_patch").orElse(0);
```
kbottla: Although it is a PATCH, this could be true/false which avoids deriving `forcePatchBackfill`

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterTask.java
```diff
@@ -46,12 +57,18 @@ public HistoricalEventEmitterTask(HistoricalEventEmitterWorker worker, Environme
...
+        boolean forcePatchBackfill = patchBackfillParam != 0;
 
         logger.info("Execute called start_id={} max_id={} - processing", startId, maybeMaxId);
         
         try {
             logger.info("Request accepted");
-            executor.execute(() -> worker.execute(startId, maybeMaxId));
+            if (forcePatchBackfill) {
```
kbottla: could just be patchBackfill ? As we create are object with `shouldForceEmission=true`, we force emit even if it is not patch

kbottla (commented):
Looks good. Added a few minor comments

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
sfount: 👍 we'll have to look into telephone payments events to make sure they have the same schema. As discussed we can look into applying these manually as there are so few live payments

sfount:
@kbottla thanks for the comments, have updated in https://github.com/alphagov/pay-connector/pull/1871/commits/3616035a48167f792d838db992b28a69f2b9abff

kbottla (approved):

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
sfount: Quickly clarifying this comment as I had misunderstood: as this uses the first (0th) recorded charge event this will respect if it was a standard payment or a telephone payment

PR #1878, opened 2019-11-04
PP-5779 Emit only if charge events exists

kbottla:
## WHAT

- For some charges, events don't exist (initial stages of developing telephone payments). Don't emit events if no charge events are found.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
@@ -103,8 +103,10 @@ public class HistoricalEventEmitter {
     public void processPatchPaymentCreatedAndPaymentDetailsEnteredEvents(ChargeEntity charge) { 
         List<ChargeEventEntity> chargeEventEntities = getSortedChargeEvents(charge);
 
-        processSingleChargeStateTransitionEvent(charge.getId(), ChargeStatus.UNDEFINED, chargeEventEntities.get(0));
-        processPaymentDetailEnteredEvent(chargeEventEntities);
+        if(!chargeEventEntities.isEmpty()) {
+            processSingleChargeStateTransitionEvent(charge.getId(), ChargeStatus.UNDEFINED, chargeEventEntities.get(0));
+            processPaymentDetailEnteredEvent(chargeEventEntities);
+        }
     }
 
     @Transactional


src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -103,8 +103,10 @@ public void processPaymentEvents(ChargeEntity charge) {
     public void processPatchPaymentCreatedAndPaymentDetailsEnteredEvents(ChargeEntity charge) { 
         List<ChargeEventEntity> chargeEventEntities = getSortedChargeEvents(charge);
 
-        processSingleChargeStateTransitionEvent(charge.getId(), ChargeStatus.UNDEFINED, chargeEventEntities.get(0));
-        processPaymentDetailEnteredEvent(chargeEventEntities);
+        if(!chargeEventEntities.isEmpty()) {
+            processSingleChargeStateTransitionEvent(charge.getId(), ChargeStatus.UNDEFINED, chargeEventEntities.get(0));
+            processPaymentDetailEnteredEvent(chargeEventEntities);
+        }
```
jankowiakmaria, line 109: Is it possible to have a [telephone] payment without any charge events?
Is it correct?

kbottla: Not now, but initially telephone charges were created without charge events. And if no events exists, we can't emit events

jankowiakmaria: Are we going to fix them or leave? (If not fixing that would mean that we cannot backfill them)

jankowiakmaria (approved):

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
kbottla: They are test payments by us. Don't think backfilling is required for these.

jankowiakmaria: Yeah, but it leaves data in a dirty state (the same as refunds not being in refund_history table).
If it's test data we should either remove it or fix (I'm not saying it needs to be done as part of this story - maybe as a police-cat thingy).

PR #1886, opened 2019-11-05
PP-5812 Add payment provider to StateTransitionService log line

danworth:
To enable our payment journey SLI we need to know the payment
provider when emitting events so that we can exclude `sandbox`. This is
being added as a structured argument.

Also take the opportunity to add the `to_state` and `from_state` as
structured arguments to the logger to make the implementation of the sli
query in Splunk simpler.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/StateTransitionService.java
+++ b/src/main/java/uk/gov/pay/connector/queue/StateTransitionService.java
@@ -17,6 +17,8 @@ import javax.inject.Inject;
 import java.time.ZoneId;
 
 import static java.time.ZonedDateTime.now;
+import static net.logstash.logback.argument.StructuredArguments.kv;
+import static uk.gov.pay.logging.LoggingKeys.PROVIDER;
 
 public class StateTransitionService {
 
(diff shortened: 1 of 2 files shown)


stephencdaly (changes requested):
Doesn't look quite right, I'm seeing the following
```
{"@timestamp":"2019-11-05T16:30:55.646Z","@version":"1","message":"Offered payment state transition to emitter queue [from=provider=sandbox] [to=from_state=CREATED] [chargeEventId=to_state=ENTERING CARD DETAILS] [chargeId={}]","logger_name":"uk.gov.pay.connector.queue.StateTransitionService$$EnhancerByGuice$$62bcb127","thread_name":"dw-26 - PUT /v1/frontend/charges/1ou22ol90r4pbj66l054nvec2t/status","level":"INFO","level_value":20000,"x_request_id":"","provider":"sandbox","from_state":"CREATED","to_state":"ENTERING_CARD_DETAILS","container":"connector"}
```
I think the string format placeholders might be wrong so they aren't being substituted, and then the `{}` is substituted by the keyValue.


Also, is it definitely this line we want to add the provider to, or the `"Changing charge status for externalId"` line, or both? That's just a question rather than an issue with what you've done.

danworth:
Thanks Stephen, I've corrected the format string, tested locally and now get what we expect.
```
{"@timestamp":"2019-11-06T08:36:19.185Z","@version":"1","message":"Offered payment state transition to emitter queue [from=SYSTEM CANCEL READY] [to=SYSTEM CANCELLED] [chargeEventId=16] [chargeId=charge1147303691]","logger_name":"uk.gov.pay.connector.queue.StateTransitionService$$EnhancerByGuice$$62bcb127","thread_name":"dw-63 - POST /v1/api/accounts/1595730147/charges/charge1147303691/cancel","level":"INFO","level_value":20000,"chargeId":"charge1147303691","provider":"worldpay","from_state":"SYSTEM_CANCEL_READY","to_state":"SYSTEM_CANCELLED","container":"connector"}
```

Wrt to your question, I don't think we need it in both places for our SLI and the line we're using comes after the `Changing charge status` and after the charge status has actually been updated, so I think its preferable.

stephencdaly (approved):
Looks good 👍

PR #1887, opened 2019-11-05
PP-5709: add shopperIPAddress to authorisation request (Worldpay)

jankowiakmaria:
Adding shopperIPAddress attribute to authorisation request can be
used by Worldpay for fraud checks.
It's an IP address which has been used by payer.
We want to add it in all cases (not only for 3ds payments).

Changes:
* update to AuthCardDetails to include ip address
* update to WorldpayAuthoriseOrderTemplate.xml to include additional
attribute and change to the condition for displaying session element
when it's 3ds payment or we have information about IP address
* update to logic in WorldpayPaymentProvider to include property determining
when session element should be added to the request
* update to tests

### DIFF
--- a/src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
+++ b/src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
@@ -31,7 +31,9 @@
                     </cardAddress>
                     </#if>
                 </VISA-SSL>
-                <#if requires3ds>
+                <#if authCardDetails.ipAddress.isPresent()>
+                <session id="${sessionId?xml}" shopperIPAddress="${authCardDetails.ipAddress.get()}"/>
+                <#elseif requires3ds>
                 <session id="${sessionId?xml}"/>
                 </#if>
             </paymentDetails>
--- a/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
@@ -71,7 +71,7 @@ public class WorldpayPaymentProviderTest extends WorldpayBasePaymentProviderTest
 
     @Mock
     private GatewayClientFactory gatewayClientFactory;
-    
+
     private ChargeEntityFixture chargeEntityFixture;
 
     @Before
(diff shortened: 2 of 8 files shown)


src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -31,8 +31,12 @@
...
                     </#if>
                 </VISA-SSL>
-                <#if requires3ds>
-                <session id="${sessionId?xml}"/>
+                <#if includesSession>
+                <session id="${sessionId?xml}"
+                <#if authCardDetails.ipAddress.isPresent()>
+                    shopperIPAddress="${authCardDetails.ipAddress.get()}"
+                </#if>
+                />
```
alexbishop1: It might be possible to do this without needing `includesSession` (at the cost of a little bit of verbosity) by doing something like this:

```xml
<#if authCardDetails.ipAddress.isPresent()> 
<session id="${sessionId?xml}" shopperIPAddress="${authCardDetails.ipAddress.get()}"/>
<#elseif requires3ds>
<session id="${sessionId?xml}"/>
</#if>
```

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -121,29 +121,32 @@ public void testRefundRequestContainsReference() throws Exception {
...
+        ChargeEntity chargeEntity = chargeEntityFixture
+                .withExternalId("uniqueSessionId")
+                .withTransactionId("transaction-id")
+                .build();
 
+        GatewayClient.Response gatewayResponse = mock(GatewayClient.Response.class);
+        when(gatewayResponse.getEntity()).thenReturn(TestTemplateResourceLoader.load(WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE));
         when(mockGatewayClient.postRequestFor(any(URI.class), any(GatewayAccountEntity.class), any(GatewayOrder.class), anyMap()))
-                .thenThrow(new GatewayException.GatewayErrorException("Unexpected HTTP status code 400 from gateway"));
+                .thenReturn(gatewayResponse);
```
alexbishop1, line 141: I don’t understand why this has changed.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -121,29 +121,32 @@ public void testRefundRequestContainsReference() throws Exception {
...
 
-        try {
-            var worldpayPaymentProvider = new WorldpayPaymentProvider(configuration, gatewayClientFactory, environment);
-            worldpayPaymentProvider.authorise(getCardAuthorisationRequest(chargeEntity));
-        } catch (GatewayException.GatewayErrorException e) {
```
alexbishop1, line 141: Why is the try/catch being removed?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -226,41 +229,12 @@ public void shouldIncludeCompletedAuthenticationInSecondOrderWhenPaRequestNotInF
         assertXMLEqual(TestTemplateResourceLoader.load(WORLDPAY_VALID_3DS_FLEX_RESPONSE_AUTH_WORLDPAY_REQUEST),
                 gatewayOrderArgumentCaptor.getValue().getPayload());
     }
-    
-    @Test
-    public void shouldNotIncludeElementsWhen3DSecureNotEnabled() throws Exception {
```
alexbishop1, line 231: I think we still want this test (in a modified form) because 3DS authorisations include an extra `browser` element with `acceptHeader` and `userAgentHeader` child elements.

alexbishop1 (changes requested):
Looks good but I have a suggestion that I think might make it a bit simpler and there’s also some changes to the tests I don’t quite follow.

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
jankowiakmaria: I considered this option.
TBH I don't mind - I can change it to ^^ if it's preferred.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
jankowiakmaria: This is a duplicate of `shouldNotInclude3dsElementsWhen3dsToggleDisabled` test.
The difference is - this one checks whether specific elements DON'T exist (`session`, `acceptHeader` and `userAgentHeader`), and the other one check the exact value of the payload (which is https://github.com/alphagov/pay-connector/blob/master/src/test/resources/templates/worldpay/valid-authorise-worldpay-request-excluding-3ds.xml file).
When you compare it with https://github.com/alphagov/pay-connector/blob/master/src/test/resources/templates/worldpay/valid-authorise-worldpay-request-including-3ds.xml (which is used in `shouldInclude3dsElementsWhen3dsToggleEnabled` test), exactly the above properties are missing from `...excluding...` file.

That's the reason I removed it - it's a duplicate.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
jankowiakmaria: So that it doesn't throw - It didn't seem like the intention of the test is to validate anything other than constructed order request.
Not mocking a gateway with a valid response but to throw an exception seems like a quicker (?) way. The intention is not to test anything along these lines, so I thought I'll clean it up - given I'm changing it anyway.
(that's the reason I removed try/catch as well).

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
alexbishop1: Yes, I think I would prefer that because otherwise `includesSession` is set quite far away from where it’s used, which means one has to look in two places to figure out what’s going on (not to mention that it’s an extra variable and one that unfortunately has the ultra-generic word “session” in the name).

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
alexbishop1: OK, thanks for explaining. I looked at the full test and what you say makes sense. I thought perhaps the exception was thrown in order to check `WorldpayPaymentProvider` handles it but it just gets propagated up anyway.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
alexbishop1: Yes, you’re right. Makes sense.

jankowiakmaria:
I'm wondering whether we should sanitise the value of the ip address - in case the connector is provided with some sort of gibberish shouldn't we just not use it (worldpay may fail otherwise (?)), but not fail the request?
Thoughts?

alexbishop1:
Yeah, interesting question.

We can probably assume that any time we’re getting the IP address directly from the connection it’s valid. But most of the time it will be coming from the `X-Forwarded-For` header. While we can probably assume an IP address added by our loadbalancers is legit, we have no control over anything else in the header (it’s essentially untrusted user input).

Validating the value could be tricky — IP addresses can be surprisingly complex, especially when IPv6 is involved. And we have no guarantee that whatever validation rules we use will match what Worldpay supports. As far as I’m aware,`X-Forwarded-For` is non-standard, so there’s no format specified.

We could try to see how Worldpay handles invalid input. Do an authorisation against a Worldpay test account with gibberish in the `shopperIPAddress` and see what happens (you can use `uk.gov.pay.connector.it.contract.WorldpayPaymentProviderTest` for this).

alexbishop1 (changes requested):
This looks good but could you please squash the commits? It will confuse future code archaeologists if we add `includesSession` in one commit and then remove it in the next one!

jankowiakmaria:
Ready for merge (the potential validation/sanitisation would be done separately).

alexbishop1 (approved):

PR #1895, opened 2019-11-07
TMP: PP-5794: validating behaviour in test-12

sfount:
Testing Stripe MOTO payment flow integration.

This will only be used to test a spike assertion, it should not be merged into master.


sfount:
This spike was completed

PR #1900, opened 2019-11-11
PP-5792 Worldpay 3ds flex JWT token credentials use new table

sfount:
* Add `jwtMacKey` to flex credentials object as a write only object,
this means it will only be serialised and shouldn't be deserialised

Token generation:
* use worldpay 3ds flex credentials POJO in favour of bundles gateway
account credentials column
* pass credentials to service from dependant resources
* update tests to mock 3ds flex credentials entity and database fixtures
* removed integration tests for partially complete flex credentials in
the database, this is now enforced at creation (non-null columns)

The details required by this change will block 3ds flex smoke tests until the account details are updated in each environment.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
@@ -203,7 +203,7 @@ public class GatewayAccountEntity extends AbstractVersionedEntity {
 
     @JsonInclude(NON_NULL)
     @JsonProperty("worldpay_3ds_flex")
-    public Worldpay3dsFlexCredentials getWorldpay3dsFlexCredentialsEntity() {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -363,6 +363,10 @@ public void setIntegrationVersion3ds(int integrationVersion3ds) {
         this.integrationVersion3ds = integrationVersion3ds;
     }
 
+    public void setWorldpay3dsFlexCredentialsEntity(Worldpay3dsFlexCredentialsEntity worldpay3dsFlexCredentialsEntity) {
```
rorymalcolm, line 366: good - correct - naming ;)

rorymalcolm (approved):
👍 LGTM! Nice work!

sfount:
Updated credentials in tests so post merge jobs should pass for build

PR #1903, opened 2019-11-12
PP-5409 Improve charge cancellation check

nimalank7:
## WHAT YOU DID

- Adds an isCancelOrExpiryInProgress() method to StatusFlow to indicate whether a cancel or expiry is in progress
- Modifies validateChargeStatus so that it checks whether a cancel or expiry is in progress (i.e. statusFlow.isStarted) and also checks whether the charge is in a terminable state. In both cases if the condition fails an exception is thrown
- Modifies validateChargeStatus method signature as 2 parameters are unused and moves method call into doCancel()

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
@@ -6,7 +6,6 @@ import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
-import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
 import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
 import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
 import uk.gov.pay.connector.gateway.GatewayException;
@@ -17,14 +16,11 @@ import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 
 import javax.inject.Inject;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CREATED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
 import static uk.gov.pay.connector.charge.service.StatusFlow.SYSTEM_CANCELLATION_FLOW;
@@ -68,6 +64,8 @@ public class ChargeCancelService {
     }
 
     private void doCancel(ChargeEntity chargeEntity, StatusFlow statusFlow) {
+        validateChargeStatus(statusFlow, chargeEntity);
+        
         if (gatewayIsNotAwareOfCharge(chargeEntity)) {
             nonGatewayCancel(chargeEntity, statusFlow);
         } else {
@@ -78,14 +76,16 @@ public class ChargeCancelService {
     private void cancelChargeWithGatewayCleanup(ChargeEntity chargeEntity, StatusFlow statusFlow) {
         prepareForTerminate(chargeEntity, statusFlow);
 
-        ChargeStatus chargeStatus = null;
-        String stringifiedResponse = null;
+        ChargeStatus chargeStatus;
+        String stringifiedResponse;
 
         try {
             final GatewayResponse<BaseCancelResponse> gatewayResponse = doGatewayCancel(chargeEntity);
 
-            if (!gatewayResponse.getBaseResponse().isPresent()) gatewayResponse.throwGatewayError();
-
+            if (gatewayResponse.getBaseResponse().isEmpty()) {
+                gatewayResponse.throwGatewayError();
+            }
+            
             chargeStatus = determineTerminalState(gatewayResponse.getBaseResponse().get(), statusFlow);
             stringifiedResponse = gatewayResponse.getBaseResponse().get().toString();
         } catch (GatewayException e) {
@@ -131,11 +131,9 @@ public class ChargeCancelService {
     }
 
     private void prepareForTerminate(ChargeEntity chargeEntity, StatusFlow statusFlow) {
-        ChargeStatus newStatus = statusFlow.getLockState();
+        ChargeStatus lockState = statusFlow.getLockState();
         ChargeStatus currentStatus = ChargeStatus.fromString(chargeEntity.getStatus());
-
-        validateChargeStatus(statusFlow, chargeEntity, newStatus, currentStatus);
-
+        
         // Used by Sumo Logic saved search
         logger.info("Card cancel request sent - charge_external_id={}, charge_status={}, account_id={}, transaction_id={}, amount={}, operation_type={}, provider={}, provider_type={}, locking_status={}",
                 chargeEntity.getExternalId(),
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -146,27 +144,27 @@ private void prepareForTerminate(ChargeEntity chargeEntity, StatusFlow statusFlo
...
-                throw new ConflictRuntimeException(chargeEntity.getExternalId());
-            }
+    private void validateChargeStatus(StatusFlow statusFlow, ChargeEntity chargeEntity) {
+        ChargeStatus chargeStatus = ChargeStatus.fromString(chargeEntity.getStatus());
 
+        if (statusFlow.isInProgress(chargeStatus)) {
+            throw new OperationAlreadyInProgressRuntimeException(statusFlow.getName(), chargeEntity.getExternalId());
+        }
+        
+        if (!chargeIsInTerminableStatus(statusFlow, chargeStatus)) {
```
danworth, line 159: Nit picking, so entirely up to you. As Robert C. Martin says in Clean Code:

> G29. Avoid Negative Conditionals
"Negatives are just a bit harder to understand than positives. So, when possible, conditionals should be expressed as positives"

Seems that `chargeIsInTerminableStatus` could be modified and this becomes:
`if (chargeIsNotInTerminableStatus) {...`

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -78,14 +76,16 @@ private void doCancel(ChargeEntity chargeEntity, StatusFlow statusFlow) {
...
-        String stringifiedResponse = null;
+        ChargeStatus chargeStatus;
+        String stringifiedResponse;
 
         try {
             final GatewayResponse<BaseCancelResponse> gatewayResponse = doGatewayCancel(chargeEntity);
 
-            if (!gatewayResponse.getBaseResponse().isPresent()) gatewayResponse.throwGatewayError();
-
+            if (gatewayResponse.getBaseResponse().isEmpty()) {
```
danworth, line 85: 👍

src/main/java/uk/gov/pay/connector/charge/service/StatusFlow.java
```diff
@@ -7,29 +7,16 @@
...
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.EXPIRE_CANCEL_SUBMITTED;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.SYSTEM_CANCELLED;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.SYSTEM_CANCEL_ERROR;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.SYSTEM_CANCEL_READY;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.SYSTEM_CANCEL_SUBMITTED;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.USER_CANCELLED;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.USER_CANCEL_ERROR;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.USER_CANCEL_READY;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.USER_CANCEL_SUBMITTED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.*;
```
danworth: Codacy doesn't like this and our style guide says to avoid wildcard imports. If Intelij did this automatically you can change the settings:
https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelFrontendResourceIT.java
```diff
@@ -121,7 +121,7 @@ public void respondWith204WithLockingState_whenCancelFailsAfterAuth() {
     }
 
     @Test
-    public void respondWith204With3DSRequeirdState_whenCancellationBeforeAuth() {
+    public void respondWith204With3DSRequiredState_whenCancellationBeforeAuth() {
```
danworth, line 124: ha 👍

src/main/java/uk/gov/pay/connector/charge/service/StatusFlow.java
```diff
@@ -99,4 +87,8 @@ public ChargeStatus getSubmittedState() {
     public ChargeStatus getFailureTerminalState() {
         return failureTerminalState;
     }
+    
+    public boolean isInProgress(ChargeStatus chargeStatus) {
+        return List.of(lockState, submittedState).contains(chargeStatus);
```
danworth: Nit picking, only mentioning it because we were speaking about different collection types, using a `Set` rather than `List` might be more appropriate since it holds unique values and being used to performa a contains check https://www.baeldung.com/java-collections-complexity

Since there's only two values it is absolutely negligible performance benefit O(1) v O(n), as I said mentioning it since it relates to previous discussions.

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -146,27 +144,27 @@ private void prepareForTerminate(ChargeEntity chargeEntity, StatusFlow statusFlo
...
                 chargeEntity.getGatewayAccount().getType(),
-                newStatus);
+                lockState);
 
-        chargeService.transitionChargeState(chargeEntity.getExternalId(), newStatus);
+        chargeService.transitionChargeState(chargeEntity.getExternalId(), lockState);
     }
 
-    private void validateChargeStatus(StatusFlow statusFlow, ChargeEntity chargeEntity, ChargeStatus newStatus, ChargeStatus oldStatus) {
-        if (!chargeIsInTerminatableStatus(statusFlow, oldStatus)) {
```
danworth, line 155: I guess there's a subtle change in logic here, thought I'd mention it just to make sure its intended (and that I've understood it). Previously, we would throw `OperationAlreadyInProgressRuntimeException` only if the charge was in a non terminable status AND the new status = old status. With this change we'll throw a `OperationAlreadyInProgressRuntimeException` if the new status = old status regardless of whether the charge is in a non terminable status.

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelFrontendResourceIT.java
```diff
@@ -131,14 +131,14 @@ public void respondWith204With3DSRequeirdState_whenCancellationBeforeAuth() {
...
-    public void respondWith409_whenCancellationDuringAuth3DSReady() {
+    public void respondWith400_whenCancellationDuringAuth3DSReady() {
         String chargeId = addCharge(AUTHORISATION_3DS_READY, "ref", ZonedDateTime.now().minusHours(1), "transaction-id");
         worldpayMockClient.mockCancelSuccess();
 
         connectorRestApiClient
                 .withChargeId(chargeId)
                 .postFrontendChargeCancellation()
-                .statusCode(409);
+                .statusCode(400);
```
danworth, line 141: Since you're changing the response code I assume this doesn't break any logic in any client services that might be processing Connector's response based on the http code, e.g. Frontend.

danworth (changes requested):
Nice change Nim 👍 If you could please remove the wildcard import then its good to go (few other thoughts but that's up to you).

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
stephencdaly: Yeh, it didn't seem like there was any good reason to only check for non-gateway terminable states. I think it will never throw the `OperationAlreadyInProgressRuntimeException` for these, but it could throw an `IllegalStateRuntimeException` so we should be doing the check whether it is in a terminable state, and it seems cleaner to do all the state validation in one place.

src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
stephencdaly: I generally agree, which is why we changed a conditional higher up in this file to get rid of a negative. I wouldn't say in this case say that I prefer the second option because I think this actually makes the method more difficult to understand due to the negative in the title. We're just moving where the negative is.

src/main/java/uk/gov/pay/connector/charge/service/StatusFlow.java
stephencdaly: 👍

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelFrontendResourceIT.java
stephencdaly: It could return 400s in other cases already so I'd hope it's handled correctly by frontend. Can double check.

src/main/java/uk/gov/pay/connector/charge/service/StatusFlow.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelFrontendResourceIT.java
danworth: Yeah maybe worth double checking, quick look appears to show there is some logic in Frontend using various 4XX but doesn't look like its related to cancelling charges so probably ok..
```
GDS5062 > grep -rA3 "case 4\d\d"  ./app/
./app//controllers/three_d_secure_controller.js:    case 400:
./app//controllers/three_d_secure_controller.js-      redirect(res).toConfirm(charge.id)
./app//controllers/three_d_secure_controller.js-      break
./app//controllers/three_d_secure_controller.js-    case 202:
--
./app//controllers/three_d_secure_controller.js:    case 409:
./app//controllers/three_d_secure_controller.js-      redirect(res).toAuthWaiting(charge.id)
./app//controllers/three_d_secure_controller.js-      break
./app//controllers/three_d_secure_controller.js-    case 500:
--
./app//controllers/charge_controller.js:    case 409:
./app//controllers/charge_controller.js-      logging.failedChargePost(409)
./app//controllers/charge_controller.js-      redirect(res).toAuthWaiting(req.chargeId)
./app//controllers/charge_controller.js-      break
--
./app//controllers/web-payments/handle-auth-response-controller.js:    case 409:
./app//controllers/web-payments/handle-auth-response-controller.js-      logging.failedChargePost(409)
./app//controllers/web-payments/handle-auth-response-controller.js-      redirect(res).toAuthWaiting(req.chargeId)
./app//controllers/web-payments/handle-auth-response-controller.js-      break
--
./app//controllers/web-payments/handle-auth-response-controller.js:    case 400:
./app//controllers/web-payments/handle-auth-response-controller.js-    case 500:
./app//controllers/web-payments/handle-auth-response-controller.js-      logging.failedChargePost(409)
./app//controllers/web-payments/handle-auth-response-controller.js-      logging.systemError('Wallet authorisation response', req.headers && req.headers[CORRELATION_HEADER], charge.id)
```

danworth (approved):
Thanks Nim, looks good 👍

src/test/java/uk/gov/pay/connector/it/resources/ChargeCancelFrontendResourceIT.java
stephencdaly: Just had a look, we don't handle any specific status codes for cancellation, we treat all failure responses the same

stephencdaly: Same in public api

PR #1907, opened 2019-11-13
PP-5788: Log Stripe account updated events

oswaldquek:
At the moment we're only concerned with `requirements` and `payouts_enabled`
information.

The idea is that a Splunk alert will be set up to look for a `Received an account.updated event for stripe account` string and create a ticket in Zendesk

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
@@ -1,0 +1,63 @@
+package uk.gov.pay.connector.gateway.stripe;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.gateway.stripe.response.StripeNotification;
+
+import javax.inject.Inject;
+
+import static net.logstash.logback.argument.StructuredArguments.kv;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.STRIPE;
+
+public class StripeAccountUpdatedHandler {
+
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    public StripeAccountUpdatedHandler(ObjectMapper objectMapper) {
+        this.objectMapper = objectMapper;
+    }
+
+    public void process(StripeNotification notification) {
+        try {
+            var dataObject = objectMapper.readValue(notification.getObject(), DataObject.class);
+            LOGGER.info(String.format("Received an account.updated event for stripe account %s", dataObject.getId()),
+                    kv("payouts_enabled", dataObject.isPayoutsEnabled()),
+                    kv("requirements", dataObject.getRequirements()));
```
alexbishop1, line 33: Since this relates to a specific gateway account, can we also add a key-value pair for that (using `LoggingKeys.GATEWAY_ACCOUNT_ID`)?

alexbishop1 (changes requested):
Looks good! One small request…

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
oswaldquek: I thought about that but it would mean retrieving a gateway account by the stripe account id (e.g. acct_xxxxxxxxx) and it would involve new methods and tests in the gateway account service and DAO and there's no functionality to search by stripe account id at the moment.

oswaldquek: Perhaps it would be better if that were done in another PR...

alexbishop1: Oh, I see. Still, even if it’s the Stripe account ID, it would be nice to extract it to a separate field (`"stripe_account_id"`, say).

oswaldquek: added `kv("stripe_account_id", dataObject.getId())`

alexbishop1 (approved):

PR #1908, opened 2019-11-13
PP-5409 Updates ChargeExpiryServiceTest

nimalank7:
## WHAT YOU DID

This PR modifies the ChargeExpiryServiceTest by updating a test and adding one to check that it does expire with payment providers when gateway state is cancelled as we don't test this currently.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
@@ -1,18 +1,21 @@
 package uk.gov.pay.connector.charge.service;
 
 import com.google.common.collect.ImmutableList;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 import uk.gov.pay.connector.app.ChargeSweepConfig;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
-import uk.gov.pay.connector.gateway.ChargeQueryResponse;
 import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.PaymentProvider;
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -176,24 +182,60 @@ public void shouldExpireChargesWithStatus_awaitingCaptureRequest_byCallingProvid
...
+        ArgumentCaptor<CancelGatewayRequest> cancelCaptor = ArgumentCaptor.forClass(CancelGatewayRequest.class);
+
+        ChargeEntity expiredCharge = mockExpiredChargeEntity();
+        when(mockChargeService.transitionChargeState(any(String.class), any())).thenReturn(expiredCharge);
+
+        chargeExpiryService.expire(singletonList(chargeEntity));
+
+        verify(mockPaymentProvider).cancel(cancelCaptor.capture());
+        assertThat(cancelCaptor.getValue().getTransactionId(), is(chargeEntity.getGatewayTransactionId()));
+        verify(mockChargeService).transitionChargeState(chargeEntity.getExternalId(), EXPIRED);
```
stephencdaly: I think we can actually improve this slightly by removing the `verify` and improving the stub for this method so that it expects exact arguments. For newer versions of Mockito, we don't need a stub and a verify if the method is only called once as Mockito will complain if the stub is unused.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -176,24 +182,60 @@ public void shouldExpireChargesWithStatus_awaitingCaptureRequest_byCallingProvid
...
+        when(mockQueryService.isTerminableWithGateway(chargeEntity)).thenReturn(true);
+
+        when(mockWorldpayCancelResponse.cancelStatus()).thenReturn(CancelStatus.CANCELLED);
+
+        when(mockChargeDao.findByExternalId(chargeEntity.getExternalId())).thenReturn(Optional.of(chargeEntity));
+        when(mockPaymentProvider.cancel(any())).thenReturn(gatewayResponse);
+        ArgumentCaptor<CancelGatewayRequest> cancelCaptor = ArgumentCaptor.forClass(CancelGatewayRequest.class);
+
+        ChargeEntity expiredCharge = mockExpiredChargeEntity();
+        when(mockChargeService.transitionChargeState(any(String.class), any())).thenReturn(expiredCharge);
```
stephencdaly: This is how I originally did this, but I think we can make it better
```suggestion
        when(mockChargeService.transitionChargeState(chargeEntity.getExternalId(), EXPIRED)).thenReturn(expiredCharge);
```

stephencdaly (changes requested):
Just an improvement that I noticed over the original test which we were re-adding

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
nimalank7: Done

stephencdaly (approved):
👍

PR #1913, opened 2019-11-14
PP-5842: include shopperIPAddress property when gateway account configured

jankowiakmaria:
It's been decided to configure whether shopperIPAddress is sent in Worldpay
authorisation at the gateway account level.
The changes to db have been introduced https://github.com/alphagov/pay-connector/pull/1912.

This PR makes use of the newly introduced db column.
* add new column to GatewayAccountEntity
* update the order template to include shopperIPAddress when ip address is present
and gateway account enables it
* update tests

I haven't extended resources - they won't include information about the new setting at the moment.
It will be required for adding UI for this piece of configuration.


alexbishop1 (changes requested):
Rather than setting a property in the XML template that says whether to include the IP address, I think it might be more straightforward set a property containing the IP address itself but only if there is one and the gateway account is set to send it.

So the code in `WorldpayPaymentProvider` would look something like this:

```java
var builder = aWorldpayAuthoriseOrderRequestBuilder().
        .withSessionId(request.getChargeExternalId())
        .with3dsRequired(is3dsRequired)
        .withDate(DateTime.now(DateTimeZone.UTC)); // etc. etc.

if (request.getGatewayAccount().isSendPayerIpAddressToGateway()) {
    request.getAuthCardDetails().getIpAddress.ifPresent(builder::withPayerIpAddress);
}

return builder.build();
```

Then in the XML template we could directly do:

```xml
<#if payerIpAddress>
    <session id="${sessionId?xml}" shopperIPAddress="${payerIpAddress?xml}"/>
<#else>
    <session id="${sessionId?xml}"/>
</#if>
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -181,12 +187,20 @@ public ExternalChargeRefundAvailability getExternalChargeRefundAvailability(Char
...
                 .withSessionId(request.getChargeExternalId())
                 .with3dsRequired(is3dsRequired)
-                .withDate(DateTime.now(DateTimeZone.UTC))
+                .withDate(DateTime.now(DateTimeZone.UTC));
+
+        if (request.getGatewayAccount().isSendPayerIpAddressToGateway()) {
+            request.getAuthCardDetails().getIpAddress().ifPresent(builder::withPayerIpAddress);
+        }
+
+        return builder
```
alexbishop1, line 203: I think it would be a bit clearer if all the happens-every-time builder operations were first, followed by the happens-some-of-the-time builder operations. Like this:

```java
var builder = aWorldpayAuthoriseOrderRequestBuilder()
        .withSessionId(request.getChargeExternalId())
        .withDate(DateTime.now(DateTimeZone.UTC))
        .withTransactionId(request.getTransactionId().orElse(""))
        // … and all the others

if (request.getGatewayAccount().isSendPayerIpAddressToGateway()) {
    request.getAuthCardDetails().getIpAddress().ifPresent(builder::withPayerIpAddress);
}

return builder.build()
```

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -32,8 +32,12 @@
                     </#if>
                 </VISA-SSL>
                 <#if requires3ds>
+                <#if payerIpAddress??>
```
alexbishop1, line 35: I am so _au fait_ with FreeMarker that I had to check what this syntax means!

https://freemarker.apache.org/docs/dgui_template_exp.html#dgui_template_exp_missing_test

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -159,20 +162,41 @@ public void shouldInclude3dsElementsWhen3dsToggleEnabled() throws Exception {
...
-                    gatewayOrderArgumentCaptor.getValue().getPayload());
-        }
+        ArgumentCaptor<GatewayOrder> gatewayOrderArgumentCaptor = ArgumentCaptor.forClass(GatewayOrder.class);
+        verify(mockGatewayClient).postRequestFor(eq(WORLDPAY_URL), eq(gatewayAccountEntity), gatewayOrderArgumentCaptor.capture(), anyMap());
+        assertXMLEqual(TestTemplateResourceLoader.load(WORLDPAY_VALID_AUTHORISE_WORLDPAY_REQUEST_INCLUDING_3DS),
+                gatewayOrderArgumentCaptor.getValue().getPayload());
+    }
+
+    @Test
+    public void shouldInclude3dsElementsWithoutShopperIpAddressWhen3dsToggleEnabled() throws Exception {
```
alexbishop1: This is a good test but the name is a little confusing because it focuses on the 3DS toggle rather than the send IP address toggle.

Since `shouldInclude3dsElementsWithoutShopperIpAddressWhen3dsAndSendIpAddressTogglsDisabled`, we could do this:

- rename `shouldInclude3dsElementsWhen3dsToggleEnabled` to `shouldInclude3dsElementsWithIpAddress`

- rename `shouldInclude3dsElementsWithoutShopperIpAddressWhen3dsToggleEnabled` to `shouldInclude3dsElementsWithoutIpAddress`

… and just have it be implicit that the behaviour in each test is that way because the equivalent gateway account toggles are set or not.

src/test/java/uk/gov/pay/connector/util/TestTemplateResourceLoader.java
```diff
@@ -19,6 +19,7 @@
     public static final String WORLDPAY_VALID_AUTHORISE_WORLDPAY_REQUEST_EXCLUDING_3DS = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-request-excluding-3ds.xml";
     public static final String WORLDPAY_VALID_AUTHORISE_WORLDPAY_REQUEST_FULL_ADDRESS = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-request-full-address.xml";
     public static final String WORLDPAY_VALID_AUTHORISE_WORLDPAY_REQUEST_INCLUDING_3DS = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-request-including-3ds.xml";
```
alexbishop1: Maybe rename this to `valid-authorise-worldpay-request-including-3ds-with-ip-address.xml` for clarity.

alexbishop1 (changes requested):
This looks good. Just a few trivial requests.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
jankowiakmaria: The top bit has `WorldpayOrderRequestBuilder` and the bottom bit has `OrderRequestBuilder` type.
Given `withPayerIpAddress` returns `WorldpayOrderRequestBuilder` type I can put the `if` at the top or in the middle.

alexbishop1: Oh, I’ve been bitten by that… _interesting_ aspect of our code before but I’d forgotten about it! OK then, how you have it is fine.

alexbishop1 (approved):

PR #1922, opened 2019-11-19
PP-5855 Add do_not_retry_emit_until column to emitted events

kbottla:
## WHAT
- Adds new column to emitted events table to be used by parity checker to indicate it is processing events
- Emitted events sweeper to consider emitting events if the column is null or the value of this column is in the past

with @alexbishop1


alexbishop1 (approved):
I paired with @kbottla on this am confident it does not require further review in line with our security procedures.

PR #1924, opened 2019-11-20
PP-5855 JPA for column `do_not_retry_emit_until`

kbottla:
## WHAT
- Adds new DB column to JPA entity and sets it null for now. Further changes for parity checker will set the column value as needed

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/EmittedEventEntity.java
+++ b/src/main/java/uk/gov/pay/connector/events/EmittedEventEntity.java
@@ -39,15 +39,22 @@ public class EmittedEventEntity {
     @Column(name = "emitted_date")
     private ZonedDateTime emittedDate;
 
+    @Convert(converter = UTCDateTimeConverter.class)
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/events/model/Event.java
```diff
@@ -21,6 +21,7 @@
     private String resourceExternalId;
     private EventDetails eventDetails;
     private ZonedDateTime timestamp;
+    private ZonedDateTime doNotRetryEmitUntil;
```
sfount: Is this field intentionally added the event model as well?

kbottla: yes, this PR updates only entity & model for new field. Subsequent PRs will use the same

kbottla: Removed field from event

sfount (approved):

PR #1925, opened 2019-11-20
PP-5855 EmittedEventSweeper - Ignore events being processed

kbottla:
## WHAT
- Emitted event sweeper currently picks up all events from emitted events and tries to emit them even though there may be another background process (parity checker, historical events emitter) that is doing the same job.
- We want to change it so that events emitted by the parity checker are not immediately picked up by the emitted events sweeper.
  This is done by using `emitted_events -> do_not_retry_emit_until`. If field value is null or is in the past emitted events sweeper can process the events. If `do_not_retry_emit_until` value is in the future, events are ignored by sweeper

with @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
@@ -51,7 +51,6 @@ public class ConnectorModule extends AbstractModule {
         bind(HashUtil.class);
         bind(RequestValidator.class);
         bind(GatewayAccountRequestValidator.class).in(Singleton.class);
-        bind(StateTransitionQueue.class).in(Singleton.class);
 
         install(jpaModule(configuration));
         install(new FactoryModuleBuilder().build(GatewayAccountServicesFactory.class));
@@ -133,6 +132,16 @@ public class ConnectorModule extends AbstractModule {
         return new NotifyClientFactory(connectorConfiguration);
     }
 
+    @Provides
+    @Singleton
+    public StateTransitionQueue stateTransitionQueue() {
+        return getStateTransitionQueue();
+    }
+
+    protected StateTransitionQueue getStateTransitionQueue() {
+        return new StateTransitionQueue();
+    }
+    
     @Provides
     public AmazonSQS sqsClient(ConnectorConfiguration connectorConfiguration) {
 
--- a/src/main/java/uk/gov/pay/connector/events/EventService.java
+++ b/src/main/java/uk/gov/pay/connector/events/EventService.java
@@ -27,7 +27,7 @@ public class EventService {
             eventQueue.emitEvent(event);
             emittedEventDao.recordEmission(event);
         } catch (QueueException e) {
-            emittedEventDao.recordEmission(event.getResourceType(), event.getResourceExternalId(), event.getEventType(), event.getTimestamp());
+            emittedEventDao.recordEmission(event.getResourceType(), event.getResourceExternalId(), event.getEventType(), event.getTimestamp(), null);
             logger.error("Failed to emit event {} due to {} [externalId={}]", event.getEventType(), e.getMessage(), event.getResourceExternalId());
         }
     }
@@ -38,6 +38,6 @@ public class EventService {
     }
 
     public void recordOfferedEvent(ResourceType resourceType, String externalId, String eventType, ZonedDateTime eventDate) {
-        emittedEventDao.recordEmission(resourceType, externalId, eventType, eventDate);
+        emittedEventDao.recordEmission(resourceType, externalId, eventType, eventDate, null);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
+++ b/src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
@@ -53,14 +53,15 @@ public class EmittedEventDao extends JpaDao<EmittedEventEntity> {
         persist(emittedEvent);
     }
 
-    public void recordEmission(ResourceType resourceType, String externalId, String eventType, ZonedDateTime eventDate) {
+    public void recordEmission(ResourceType resourceType, String externalId, String eventType, ZonedDateTime eventDate,
+                               ZonedDateTime doNotRetryEmitUntil) {
         final EmittedEventEntity emittedEvent = new EmittedEventEntity(
                 resourceType.getLowercase(),
                 externalId,
                 eventType,
                 eventDate,
                 null,
-                null
+                doNotRetryEmitUntil
         );
         persist(emittedEvent);
     }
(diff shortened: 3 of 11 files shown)


src/main/java/uk/gov/pay/connector/events/EventService.java
```diff
@@ -27,7 +27,7 @@ public void emitAndRecordEvent(Event event) {
             eventQueue.emitEvent(event);
             emittedEventDao.recordEmission(event);
         } catch (QueueException e) {
-            emittedEventDao.recordEmission(event.getResourceType(), event.getResourceExternalId(), event.getEventType(), event.getTimestamp());
+            emittedEventDao.recordEmission(event.getResourceType(), event.getResourceExternalId(), event.getEventType(), event.getTimestamp(), null);
```
jankowiakmaria, line 30: [minor] It seems to me that it's better (in terms of readability) to have overloaded method int the `EmittedEventDao` without the last parameter than passing `null` (which then invokes original method with `null` obviously).

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -133,6 +132,16 @@ protected NotifyClientFactory getNotifyClientFactory(ConnectorConfiguration conn
         return new NotifyClientFactory(connectorConfiguration);
     }
 
+    @Provides
+    @Singleton
+    public StateTransitionQueue stateTransitionQueue() {
+        return getStateTransitionQueue();
+    }
```
jankowiakmaria, line 139: [Question] Now I may sound ignorant, but I'm trying to figure out what's the difference between the 
```bind(StateTransitionQueue.class).in(Singleton.class);```
and 
```@Provides @Singleton``` on the method.

I looked at:
* https://stackoverflow.com/questions/14781471/guice-differences-between-singleton-class-and-singleton and
* https://github.com/google/guice/wiki/Scopes
and I don't see the difference.

Could you explain, or point me to the resource, please?

Or is it just for the testing purpose? To override the `getStateTransitionQueue` method?

File: src/test/java/uk/gov/pay/connector/events/dao/EmittedEventDaoIT.java [hunk 3]
```diff
@@ -160,6 +162,35 @@ public void findNotEmittedEventsOlderThan_shouldReturnEventsWithEmptyEmittedDate
...
+        assertThat(maxId.isPresent(), is(true));
+
+        List<EmittedEventEntity> notEmittedEvents = emittedEventDao.findNotEmittedEventsOlderThan(
+                ZonedDateTime.parse("2019-01-01T14:00:01Z"), 2, 0L, Long.MAX_VALUE, ZonedDateTime.now(UTC));
+
+        assertThat(notEmittedEvents.size(), is(2));
+        assertThat(notEmittedEvents.get(0).getEmittedDate(), nullValue());
+        assertThat(notEmittedEvents.get(0).getEventType(), is(paymentCreatedEvent.getEventType()));
+        assertThat(notEmittedEvents.get(0).getResourceExternalId(), is(paymentCreatedEvent.getResourceExternalId()));
+        assertThat(notEmittedEvents.get(0).getResourceType(), is(paymentCreatedEvent.getResourceType().getLowercase()));
```
jankowiakmaria, line 186: [minor] This could be tested for appropriate `doNotRetryEmitUntil` value (given the same event is used to emit two events).

src/test/java/uk/gov/pay/connector/it/events/EmittedEventResourceIT.java
```diff
@@ -0,0 +1,163 @@
...
+
+        connectorRestApiClient
+                .postEmittedEventsSweepTask()
+                .statusCode(OK.getStatusCode());
+
+        assertThat(stateTransitionQueue.size(), is(1));
+
+        List<Map<String, Object>> emittedEvents = databaseTestHelper.readEmittedEvents();
+
+        // emitted events sweeper adds duplicate event to table (and doesn't remove existing event)
```
jankowiakmaria, line 72: [just saying] but the first one should have the emitted date set

src/test/java/uk/gov/pay/connector/it/events/EmittedEventResourceIT.java
```diff
@@ -0,0 +1,163 @@
...
+                .withStatus(CREATED)
+                .build());
+        return chargeId;
+    }
+
+    private void assertEmittedEvent(Map<String, Object> emittedEvent, Instant expectedDoNotRetryEmitUntil) {
+        assertThat(emittedEvent.get("event_type"), is("PAYMENT_CREATED"));
+        Optional.ofNullable(expectedDoNotRetryEmitUntil).ifPresentOrElse(
+                value -> assertThat(emittedEvent.get("do_not_retry_emit_until"), is(Timestamp.from(expectedDoNotRetryEmitUntil))),
+                () -> assertThat(emittedEvent.get("do_not_retry_emit_until"), is(nullValue()))
```
jankowiakmaria, line 140: [minor] a bit too much logic in the assertion itself

src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
```diff
@@ -84,27 +85,32 @@ public void markEventAsEmitted(Event event) {
         query.executeUpdate();
     }
 
-    public Optional<Long> findNotEmittedEventMaxIdOlderThan(ZonedDateTime cutOffDate) {
+    public Optional<Long> findNotEmittedEventMaxIdOlderThan(ZonedDateTime cutOffDate, ZonedDateTime now) {
         String query = "SELECT MAX(e.id) from EmittedEventEntity e " +
                 "WHERE e.eventDate < :cutOffDate " +
-                "AND e.emittedDate is null";
+                "AND e.emittedDate is null " +
+                "AND (e.doNotRetryEmitUntil is null or e.doNotRetryEmitUntil < :currentDate)";
```
jankowiakmaria, line 92: [Question] Have you considered using database now?

jankowiakmaria (approved):
Few minor comments and questions.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
kbottla: There is no difference in this case. It is for testing only as we have no other way of asserting that events are added to StateTransitionQueue

src/main/java/uk/gov/pay/connector/events/EventService.java
kbottla: Will look into this in the next PR

src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
kbottla: now() doesn't work with JPA. We will have to use native sql which complicates it.

jankowiakmaria:
Another question - do you plan to add an index to the new field? (or is it fine without?)

src/test/java/uk/gov/pay/connector/it/events/EmittedEventResourceIT.java
kbottla: We don't have background process here consuming events and if emitted_date is set, sweeper won't emit another event.

kbottla (commented):
Thanks @jankowiakmaria. Yes we need an index for new field. Will be in a separate PR

src/main/java/uk/gov/pay/connector/events/dao/EmittedEventDao.java
jankowiakmaria: I see

src/test/java/uk/gov/pay/connector/it/events/EmittedEventResourceIT.java
jankowiakmaria: The task that runs the backfill sets emitted date, so on top of generating the new one - the old one is marked as "done".
https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java#L93

kbottla: Weird, it is not being set in the test. Will check and update accordingly.

PR #1928, opened 2019-11-22
PP-5855 ParityChecker - mark events as being processed

kbottla:
## WHAT

- Adds configuration (default value) & task parameter to set  `do_not_retry_emit_until` duration.
  Emitted event sweeper ignores events, if `do_not_retry_emit_until` value is in the future as per https://github.com/alphagov/pay-connector/commit/64d7e343e1a327b3aa000bff4baf2b2aba65e850
- Sets `do_not_retry_emit_until` date for events being processed by parity checker worker.
- Rest of the tasks (EmittedEventsSweeper, HistoricalEventEmitter) do not need to set `do_not_retry_emit_until`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
@@ -20,11 +20,13 @@ import uk.gov.pay.connector.refund.dao.RefundDao;
 import uk.gov.pay.connector.refund.model.domain.RefundHistory;
 import uk.gov.pay.connector.refund.service.RefundStateEventMap;
 
+import java.time.ZonedDateTime;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import static java.time.ZonedDateTime.now;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ABORTED;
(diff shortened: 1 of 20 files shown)


src/main/resources/config/config.yaml
```diff
@@ -227,6 +227,9 @@ chargesSweepConfig:
 emittedEventSweepConfig:
   notEmittedEventMaxAgeInSeconds: ${NOT_EMITTED_EVENT_MAX_AGE_IN_SECONDS:-1800}
 
+parityCheckerConfig:
+  defaultDoNotRetryEmittingEventUntilDurationInSeconds: ${DEFAULT_DO_NOT_RETRY_EMITTING_EVENT_UNTIL_DURATION_IN_SECONDS:-7200}
```
jankowiakmaria, line 231: [minor] It would be good to add the parameter description to `README.md`.

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -68,24 +70,32 @@
...
                                   RefundDao refundDao,
                                   EventService eventService,
                                   StateTransitionService stateTransitionService) {
-        this(emittedEventDao, refundDao, false, eventService, stateTransitionService);
+        this(emittedEventDao, refundDao, false, eventService, stateTransitionService, null);
     }
 
     public HistoricalEventEmitter(EmittedEventDao emittedEventDao,
                                   RefundDao refundDao, boolean shouldForceEmission, EventService eventService,
                                   StateTransitionService stateTransitionService) {
```
jankowiakmaria, line 85: So this constructor is used by `HistoricalEventEmitterTask`. Ideally it should follow the same logic as Parity Checker - otherwise a clash of the backfill and emitted events sweeper can happen (causing the same problems).
What's the plan? Update it in the next PR?

jankowiakmaria (approved):
Two things (that can be done outside of this PR).
Looks good.

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
kbottla: Good point. will create a self contained story/PR for `HistoricalEventEmitterTask`

PR #1930, opened 2019-11-22
PP-5855 HistoricalEventEmitter - mark events as being processed

kbottla:
## WHAT
- Sets `do_not_retry_emit_until` date for events being processed by historical event emitter worker so EmittedEventsSweeper donot process the same

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterByDateRangeTask.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterByDateRangeTask.java
@@ -7,11 +7,14 @@ import io.dropwizard.servlets.tasks.Task;
 import io.dropwizard.setup.Environment;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.app.config.EventEmitterConfig;
 
 import java.io.PrintWriter;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
 import java.util.Optional;
+import java.util.OptionalLong;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.SynchronousQueue;
 
@@ -19,6 +22,7 @@ public class HistoricalEventEmitterByDateRangeTask extends Task {
     private static final String TASK_NAME = "historical-event-emitter-by-date";
     private final Logger logger = LoggerFactory.getLogger(getClass());
     private HistoricalEventEmitterWorker worker;
+    private EventEmitterConfig eventEmitterConfig;
     private ExecutorService executor;
 
     public HistoricalEventEmitterByDateRangeTask() {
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterByDateRangeTask.java
```diff
@@ -77,4 +86,21 @@ private Optional getParam(ImmutableMultimap<String, String> parameters, String p
...
     }
+
+    private Long getDoNotRetryEmitUntilDuration(ImmutableMultimap<String, String> parameters) {
+        OptionalLong doNotRetryEmitUntil = getParam(parameters,
+                "do_not_retry_emit_until");
+        return doNotRetryEmitUntil.orElse(
+                eventEmitterConfig.getDefaultDoNotRetryEmittingEventUntilDurationInSeconds());
+    }
+
+    private OptionalLong getParam(ImmutableMultimap<String, String> parameters, String paramName) {
```
jankowiakmaria, line 97: [minor] it's more like `getLongParam`

jankowiakmaria: The nicest thing would be to extract https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/tasks/ParityCheckTask.java#L71-L99 to separate class and reuse it everywhere.

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterByDateRangeTask.java
```diff
@@ -43,28 +49,31 @@ public HistoricalEventEmitterByDateRangeTask(HistoricalEventEmitterWorker worker
...
+                executor.execute(() -> worker.executeForDateRange(startDate.get(), endDate.get(),
+                        doNotRetryEmitUntilDuration));
                 output.println("Accepted");
-            } catch (java.util.concurrent.RejectedExecutionException e) {
-                logger.info("Rejected request, worker already running");
-                output.println("Rejected request, worker already running");
             }
+        } catch (java.util.concurrent.RejectedExecutionException | NumberFormatException e) {
+            logger.info("Rejected request, worker already running");
+            output.println("Rejected request, worker already running");
```
jankowiakmaria, line 72: [Question] Why was the `try/catch` moved?

jankowiakmaria (approved):
A few comments, bud I don't think they affect mergability.

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitterByDateRangeTask.java
kbottla: It is to handle invalid `do_not_retry_emit_until` and return `Rejected` response instead of returning 500

jankowiakmaria: Yeah...I guess it's just the error message is not clear in case of the parameter being invalid.
Probably could be addressed as part of refactoring parameter extraction.

kbottla: There is lots around the tasks to improve for validation / refactoring. Would prefer a separate story to improve it all

PR #1936, opened 2019-11-27
PP-5839 Stripe - set request_three_d_secure to any

kbottla:
## WHAT

- For Stripe payment intents - `request_three_d_secure` attribute is not set currently, which will default it to `automatic`. In this case, 3DS authentication is only enforced if card issuer requests it.
- With `request_three_d_secure` set to `any`, Stripe requires users to perform 3DS authentication if supported by card


sfount (changes requested):
Code looks good to me! 

As discussed with tech/ product, it looks like it would be a good idea to understand our completion rates and understand what we're doing by forcing 3ds on 100% of payments. 

Setting up a splunk dashboard/ db query to compare where we are now vs. what impact this has will be useful. 

(made the PR review so that others don't pick up the PR)

kbottla:
Not to enforce 3DS, until we have reliable Stripe 3DS smoke test

PR #1948, opened 2019-12-03
PP-5424 Remove /v2 search endpoint

kbottla:
## WHAT
- Removes `/v2/api/accounts/{accountId}/charges` and relevant as the endpoint is no longer used by selfservice


alexbishop1 (approved):
🗑

PR #1949, opened 2019-12-03
PP-5424 Remove search refunds endpoint

kbottla:
## WHAT
- Removes search refunds endpoint (`/v1/api/accounts/{accountId}/refunds`) and related service classes as publicapi uses ledger to search for refunds


sfount (approved):
`-472` 🎉

sfount:
(as discussed README will need updating for this path but this will come in a future PR)

PR #1969, opened 2019-12-09
PP-5893: Add gateway tx id to mdc

oswaldquek:


src/main/java/uk/gov/pay/connector/filters/LoggingMDCRequestFilter.java
```diff
@@ -48,6 +48,9 @@ public void filter(ContainerRequestContext requestContext) {
 
         getPathParameterFromRequest("chargeTokenId", requestContext)
                 .ifPresent(token -> MDC.put(SECURE_TOKEN, token));
+
+        getPathParameterFromRequest("gatewayTransactionId", requestContext)
+                .ifPresent(gatewayTxId -> MDC.put("gateway_transaction_id", gatewayTxId));
```
stephencdaly: could we use `PROVIDER_ID` in pay-java-commons for this?

danworth: Looks good, I guess you're assuming this key is unlikely to be used anywhere else and hence no need to be added to pay-java-common/loggingKeys

danworth (dismissed):
Looks good

src/main/java/uk/gov/pay/connector/filters/LoggingMDCRequestFilter.java
oswaldquek: I'm not really sure as we're trying to avoid "generic" keys. see https://github.com/alphagov/pay-java-commons/pull/177.

oswaldquek: yes

stephencdaly: I think specifically this key is worth having a constant for, I can see it being used in different places in connector (such as for processing notifications) and also potentially in ledger

stephencdaly: So maybe add a new non-generic one?

stephencdaly: or change "provider_id" to be "provider_payment_id" to be less generic. we can also use that in dd_connector too then.

oswaldquek: fixed

stephencdaly (approved):

PR #1977, opened 2019-12-12
[PP-5936] update api_specification.md

jankowiakmaria:
Why?
The new property `block_prepaid_cards` has been added to the gateway_accounts
table. It will be used for determining whether payments are blocked for pre-paid
cards. The new property has been added and exposed through connector api therefore
updating the docs is necessary.

Changes:
* update affected endpoints with the new property


sfount (approved):
👍

PR #1980, opened 2019-12-13
PP-JFDI Add columns to track MOTO payments

sfount:
* charges track "is moto" through `moto` column
* gateway accounts can be configured to allow or block moto payment
requests through `allow_moto`


sfount:
This work was superseded by `PP-5952`

PR #1982, opened 2019-12-13
PP-JFDI Pass MOTO flag through to Stripe

sfount:
Depends on https://github.com/alphagov/pay-connector/pull/1980

Original code was split into 4 pull requests but ended up building incorrectly 🤦‍♂ . 

Top level: 
* column migrations
* entities and resource updated
* patching resources as required 
* logic for passing flag through to Stripe if enabled on the charge and the gateway account


sfount:
This work was superseded by `PP-5952`

PR #1985, opened 2019-12-17
BAU Add `to_state` and `from_state` when logging state change

danworth:
This will make extracting and searching for payment state changes
easier.


jankowiakmaria (approved):

kbottla:
Confirmed with @danworth, we don't this this change anymore

PR #1988, opened 2019-12-19
[PP-5924]: add new log for missing ddc when 3ds Flex

jankowiakmaria:
Why?
In order to investigate the issue with smoke test flakiness we need more information.
Few errors indicate missing `additional3DSData` element which is populated with ddc information.
The log will appear only for gateway accounts with 3ds integration enabled and version 2 when there
is missing device collection data.


SandorArpa (approved):
LOKTM

PR #1991, opened 2019-12-20
PP-5986 Poll for charge status before proceeding with Stripe 3DS confirmation

kbottla:
## WHAT

Introduces following changes to avoid `OptimisticLockException` due to Frontend (which tries to update charge to AUTHORISATION_3DS_READY) and Stripe notification (which updates charge to AUTHORASATION_SUCCESS/REJECTED) trying to update charge at the same time

- Wait on frontend locking state (`AUTHORISATION_3DS_READY`) to proceed with Stripe 3DS confirmation
- Retry every 200 milliseconds and until specified delay (as per config variable `NOTIFICATION_3DS_WAIT_DELAY`) for charge to be in AUTHORISATION_3DS_READY state before processing 3DS authorisation
- Relevant unit tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
@@ -22,6 +22,7 @@ import javax.ws.rs.WebApplicationException;
 import java.util.List;
 import java.util.Optional;
 
+import static net.logstash.logback.argument.StructuredArguments.kv;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -192,6 +195,25 @@ private void executePost3DSAuthorisation(ChargeEntity charge, String notificatio
...
+    private void delayFor3dsReady(ChargeEntity charge) {
+        int totalTimeDelayedInMillis = 0;
+        int delayInMillis = 200;
+        while (totalTimeDelayedInMillis < stripeGatewayConfig.getNotification3dsWaitDelay()) {
+            ChargeEntity chargeEntity = chargeService.findChargeById(charge.getExternalId());
+            if (ChargeStatus.fromString(chargeEntity.getStatus()) == AUTHORISATION_3DS_READY) {
+                break;
+            }
+            logger.info("Frontend has not posted 3ds ready locking state for charge [{}]," +
+                    " waiting for 200ms", charge.getExternalId());
```
sfount: Probably worth parameterising this with `delayInMillis` so it doesn't get out of date!

sfount: EDIT: this will likely log ~5 times per payment depending on how long this will be waiting. Maybe it's a good idea to only log when this method has completed, once per payment. 

Logging `totalTimeDelayedInMillis` at the end will show how long it took and will be inline making it easier for our monitoring

kbottla: Updated! Now logs only once for each notification

sfount (approved):
Looks good to me! 👍

PR #1994, opened 2019-12-23
PP-5964 Add userEmail to RefundEntity

kbottla:
## WHAT YOU DID

- Adds new DB column `user_email` to RefundEntity & RefundsHistory
- Updates relevant tests


sfount (approved):
Code looks good to me 👍nice one! I wonder if it would be useful to have a direct assertion that refund table and entity map correctly. There is one for refund history so that certainly implies it is propagating correctly

kbottla:
Thanks @sfount, going to add assertion on refundEntity persist in the next PR

PR #1997, opened 2019-12-23
PP-5966 Add user_email to RefundCreatedByUser event

kbottla:
## WHAT
- **Propagate user_email to Ledger** : Adds `user_email` field to RefundCreateByUser event payload


sfount (approved):
🎉

PR #2002, opened 2019-12-30
[PP-5823] change when email notification is sent

jankowiakmaria:
Send card payment confirmation emails when user successfully completes
the payment journey (non-delayed captures).
It used to be send after the successful gateway request to capture the payment.

Why?
See RFC-049;

tl;dr;
we frequently get timeouts from ePDQ when sending the capture request.
That causes us not to send confirmation emails to the payers.

Changing the time of sending the email will mitigate the issue.

During the kick-off we've decided that for the delayed-capture payments the
confirmation emails will be sent in the old way though.

Changes:
* update the logic for sending the confirmation email in `markChargeAsEligibleForCapture`
and `processGatewayCaptureResponse` methods
* update the tests (to verify sending email invocation)
* replace deprecated `verifyZeroInteractions` method with `verifyNoInteractions`
* remove redundant tests


kbottla (approved):
LGTM!

PR #2007, opened 2020-01-07
Run builds on travis

rjbaker:
## WHAT YOU DID
Adds a simple .travis.yml build configuration for running CI java build tasks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,5 @@ before_install:
 cache:
   directories:
     - $HOME/.m2
+
+script: mvn clean verify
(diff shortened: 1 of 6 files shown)


danworth (approved):
Looks good, I'm wondering whether reducing the log level and removing the request/response loggers might make it harder to debug failed builds but since its necessary to run on Travis I guess 🤷‍♂ 

👍

rjbaker:
> Looks good, I'm wondering whether reducing the log level and removing the request/response loggers might make it harder to debug failed builds but since its necessary to run on Travis I guess 🤷‍♂
> 
> 👍

I'd argue it makes the test results harder to read as it produces thousands of useless lines. If you need to debug further, run the failed tests in isolation locally with a different log level.

PR #2011, opened 2020-01-09
PP-6001 upgrade dropwizard to 2.0.0

SandorArpa:
## WHAT YOU DID
- upgrade dropwizard to 2.0.0
- fix maven imports that's been omitted from the upgrade
- fix/refactor code that needs to be changed due to the upgrade
- fix tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,15 +4,20 @@
     <groupId>uk.gov.pay</groupId>
     <version>0.1-SNAPSHOT</version>
     <artifactId>pay-connector</artifactId>
+    <parent>
+        <groupId>io.dropwizard</groupId>
+        <artifactId>dropwizard-dependencies</artifactId>
+        <version>2.0.0</version>
+    </parent>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dropwizard.version>1.3.17</dropwizard.version>
+        <dropwizard.version>2.0.0</dropwizard.version>
         <liquibase.version>3.7.0</liquibase.version>
         <wiremock.version>2.25.1</wiremock.version>
         <eclipselink.version>2.7.5</eclipselink.version>
         <guice.version>4.2.2</guice.version>
         <jackson.version>2.10.2</jackson.version>
-        <pay-java-commons.version>1.0.20191224120315</pay-java-commons.version>
+        <pay-java-commons.version>1.0.20200107160358</pay-java-commons.version>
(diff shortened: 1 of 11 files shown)


pom.xml
```diff
@@ -475,6 +474,12 @@
             <version>1.12.4</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jdbi</groupId>
```
kbottla, line 478: We may use latest version of jdbi instead of sticking to old version

```
<dependency>
            <groupId>io.dropwizard</groupId>
            <artifactId>dropwizard-jdbi3</artifactId>
            <version>${dropwizard.version}</version>
            <scope>test</scope>
        </dependency>
```

pom.xml
```diff
@@ -305,6 +305,11 @@
             <artifactId>jetty-util</artifactId>
             <version>9.4.25.v20191220</version>
         </dependency>
+        <dependency>
```
kbottla: Is it still needed after upgrade to latest java-commons version ?

pom.xml
```diff
@@ -6,13 +6,13 @@
     <artifactId>pay-connector</artifactId>
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dropwizard.version>1.3.17</dropwizard.version>
+        <dropwizard.version>2.0.0</dropwizard.version>
```
kbottla, line 14: Might be worth adding 
```
<parent>
  <groupId>io.dropwizard</groupId>
  <artifactId>dropwizard-dependencies</artifactId>
  <version>2.0.0</version>
</parent>
```
so there won't be different library versions https://www.dropwizard.io/en/release-2.0.x/manual/upgrade-notes/upgrade-notes-2_0_x.html

kbottla (commented):
Looks good, a few comments on libraries

pom.xml
SandorArpa: this is not dropwizard jdbi. this one is used for in TestContext for DBTestHelper (org.skife.jdbi.v2.DBI).

pom.xml
SandorArpa: 👍

pom.xml
SandorArpa: No, it seems

kbottla (approved):

pom.xml
kbottla: This can be replaced by Jdbi on Jdbiv3, but can be done in separate PR

PR #2016, opened 2020-01-14
PP-6014 Add source column

kbottla:
## WHAT YOU DID

- Adds new column to charges table which is to be used to track source of payment


SandorArpa (approved):

PR #2018, opened 2020-01-14
PP-6018 Add `source` column to ChargeEntity

kbottla:
## WHAT YOU DID
- Adds new db column `source` and to ChargeEntity class
- Integration test

with @sandorarpa


SandorArpa (approved):

PR #2019, opened 2020-01-14
PP-6018 source field for create charge endpoint

kbottla:
## WHAT YOU DID

- Added new optional field `source` for create charge endpoint - allowed values (CARD_API, CARD_PAYMENT_LINK, empty)
- Validation for field
- Relevant unit & integration tests

with @sandorarpa


SandorArpa (approved):

PR #2022, opened 2020-01-15
PP-6021 Specify source for external telephone payments

kbottla:
## WHAT YOU DID
- Sets the `source` field to `CARD_EXTERNAL_TELEPHONE` for payments reported via endpoint v1/api/accounts/{accountId}/telephone-charges
- Added relevant tests


SandorArpa (approved):
LOKTM

PR #2038, opened 2020-01-24
Revert "PP-6073 Return allow_moto field in gateway account response"

oswaldquek:
Reverts alphagov/pay-connector#2035

The build for https://github.com/alphagov/pay-connector/pull/2034 failed and it needs to be its own green build so the database migration can run. If we deploy that PR and this PR at the same time things might break.


rorymalcolm (approved):
👍

PR #2039, opened 2020-01-24
Revert "Revert "PP-6073 Return allow_moto field in gateway account response""

oswaldquek:
Reverts alphagov/pay-connector#2038


rorymalcolm (approved):
hey dawg we heard you liked reverting so we reverted your revert

PR #2040, opened 2020-01-24
PP-6073: Integration test for patching allow_moto field

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
@@ -13,6 +13,8 @@ import uk.gov.pay.connector.it.dao.DatabaseFixtures;
 import uk.gov.pay.connector.junit.DropwizardConfig;
 import uk.gov.pay.connector.junit.DropwizardJUnitRunner;
 
+import java.util.Map;
+
 import static io.restassured.http.ContentType.JSON;
 import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -449,6 +451,38 @@ public void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+        givenSetup()
+                .body(payload)
+                .patch("/v1/api/accounts/" + gatewayAccountId)
+                .then()
+                .statusCode(OK.getStatusCode());
+        givenSetup()
+                .get("/v1/api/accounts/" + gatewayAccountId)
+                .then()
+                .statusCode(OK.getStatusCode())
+                .body("allow_moto", is(false));
```
stephencdaly, line 484: reckon this should be 2 tests as it's testing 2 different things?

stephencdaly: My thinking on this was the service test and an integration test that tests patching some path was enough, which is why I didn't add an IT. Happy if you think otherwise

oswaldquek: Ok, I'll close this PR then. I generally prefer IT tests over service tests if they're virtually the same as IT tests make it clearer how the API is used.

PR #2044, opened 2020-01-27
PP-6075 Better rationalise ChargeEntity constructors

stephencdaly:
It was unclear what the difference between the various ChargeEntity constructors were. Make this clearer by:

- Moving `ChargeEntityFixture` to the same package as `ChargeEntity` in test code and making the constructor used only by `ChargeEntityFixture` package-private.

- Adding static factory methods for constructing `ChargeEntity` to make it immediately clear that one is for creating a web payment and the other is for telephone payments

- Reordering the parameters of the constructor used by telephone payments to make it easier to see what the difference is between this constructor and the constructor used for non-telephone payments. Also remove `status`, `source` and `language` parameters as these are always given static values.


oswaldquek (dismissed):
This represents an improvement but I think we could use builders, for e.g. TelephoneChargeEntityBuilder and WebChargeEntityBuilder to make it better!

stephencdaly:
Yeh if you think builders would make it better I could do that. Would you have 2 separate builders or just allow the service to decide what properties it wants to set on the ChargeEntity?

oswaldquek:
@stephencdaly two separate builders would be nice!

stephencdaly:
I'm slightly worried 2 builders will create a lot of boilerplate and make the code confusing again. Are you really wedded to this idea? If not I think I'd prefer to keep the constructors even though I would generally prefer having a builder.

oswaldquek:
I'm not really wedded to the idea. I think what you've done in this PR is already an improvement so you could just go ahead and merge it!
I assume web and telephone charges need separate fields populated which I think having builders for each would make what fields are required much clearer via the presence/absence of setter methods. So I don't think it would make the code confusing... just my opinion though!

stephencdaly:
I've gone ahead and added the builders, it is probably nicer

oswaldquek (approved):

PR #2049, opened 2020-01-28
PP-6067 Remove Chained JPA Calls From doRefund

rorymalcolm:
- Removes chained JPA calls from doRefund() in favour of direct access
via the relevant DAO classes.

- Updates tests with new mocks to validate this behaviour.


SandorArpa (approved):
LOKTM

PR #2058, opened 2020-01-31
PP-6067 Pass GatewayAccountEntity to RefundGatewayRequest.value()

kbottla:
## WHAT YOU DID
- Passes GatewayAccountEntity to build refund gateway request instead of using changeEntity.getGatewayAccountEntity()
- Few other places still need JPA way as GatewayAccount can only be derived from charge entity. This can be changed once Connector starts using chargeExternalId everywhere for refunds and also uses ledger to get payment information if charge doesn't exists in connector
  https://github.com/alphagov/pay-connector/blob/d239965797d5f1966e61d3484828f53b53134844/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java#L53
https://github.com/alphagov/pay-connector/blob/3865a9a5c329a33fb2874a6e10fc15e88fdd34b2/src/main/java/uk/gov/pay/connector/events/model/refund/RefundCreatedByService.java#L20


SandorArpa (approved):
LOKTM

PR #2064, opened 2020-02-03
PP-5976 Send moto flag to Stripe

stephencdaly:
If a charge has the `moto` flag set to true, include the parameter `payment_method_options[card[moto]]` in the request payload with a value of true.

If the charge is not a moto charge, omit this parameter.

I've tested that this creates the charge successfully in Stripe testing using Pay local against the test production Stripe account. When the payment intent is created in Stripe with the moto flag, the response contains:

```
"payment_method_details": {
          "card": {
            ...
            "moto": true,
            ...
          },
          "type": "card"
        },
```

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequestTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequestTest.java
@@ -14,11 +14,13 @@ import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayReques
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 
 import java.net.URI;
+import java.net.URLEncoder;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.not;
-import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.when;
 
 @RunWith(MockitoJUnitRunner.class)
@@ -70,7 +72,7 @@ public class StripePaymentIntentRequestTest {
         assertThat(payload, containsString("on_behalf_of=" + stripeConnectAccountId));
         assertThat(payload, containsString("confirm=true"));
         assertThat(payload, containsString("description=" + description));
-        assertThat(payload, containsString("return_url=" + frontendUrl + "%2Fcard_details%2F" + chargeExternalId + "%2F3ds_required_in"));
+        assertThat(payload, containsString("return_url=" + URLEncoder.encode(frontendUrl + "/card_details/" + chargeExternalId + "/3ds_required_in", UTF_8)));
     }
 
     @Test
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequestTest.java
```diff
@@ -75,13 +75,43 @@ public void shouldHaveCorrectParametersWithAddress() {
...
+
+
+    @Test
+    public void shouldIncludeMotoFlagWhenChargeIsMoto() {
+        when(charge.isMoto()).thenReturn(true);
+
+        StripePaymentIntentRequest stripePaymentIntentRequest = createStripePaymentIntentRequest();
+
+        String payload = stripePaymentIntentRequest.getGatewayOrder().getPayload();
+        assertThat(payload, containsString("payment_method_options%5Bcard%5Bmoto%5D%5D=true"));
```
oswaldquek: could we not put the string through a some http encoder rather than hard code them (e.g. `%5B`) here?

src/test/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequestTest.java
```diff
@@ -75,13 +75,43 @@ public void shouldHaveCorrectParametersWithAddress() {
...
+    }
+    
+    @Test
+    public void shouldNotIncludeMotoFlagWhenChargeIsNotMoto() {
+        when(charge.isMoto()).thenReturn(false);
+
+        StripePaymentIntentRequest stripePaymentIntentRequest = createStripePaymentIntentRequest();
+
+        String payload = stripePaymentIntentRequest.getGatewayOrder().getPayload();
+        assertThat(payload, not(containsString("payment_method_options%5Bcard%5Bmoto%5D%5D")));
```
oswaldquek: same comment as above

src/test/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequestTest.java
stephencdaly: This is test code, so doesn't seem too unreasonable to compare to a hardcoded string? This is following what is done for the existing tests

oswaldquek: I think test code should be as readable as production code. I doubt most developers would be able to know `%5B` is the `[` character off the top of their head?
Yeah I think the existing tests should be changed too, I wouldn't mind making a PR for that.

stephencdaly: okay, I'll look into it

oswaldquek: `URLEncoder.encode` should do the trick

stephencdaly: done

oswaldquek: Nice, thanks for doing this!

oswaldquek (approved):

PR #2070, opened 2020-02-04
PP-6095 add findRefundsRefundsByChargeExternalId to RefundDao

SandorArpa:
## WHAT YOU DID
- add new method to RefundDao (so getRefunds() can be decoupled from ChargeEntity)
- update DatabaseFixtures accordingly
- update existing tests to use chargeExternalId
- add JPA test for RefundDao


sfount (approved):
LGTM 👍

PR #2073, opened 2020-02-05
PP-6097 Add DAO method `findChargeToExpunge`

kbottla:
## WHAT YOU DID
- Added new DAO method `findChargeToExpunge` to find a charge to expunge based on the 'ageOfcharge` and which has not been parity checked within last week

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
@@ -188,4 +188,26 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
                 .setMaxResults(size)
                 .getResultList();
     }
+
+    public Optional<ChargeEntity> findChargeToExpunge(int minimumAgeOfChargeInDays) {
+        String query = "SELECT c FROM ChargeEntity c " +
+                "WHERE (c.parityCheckDate is null or c.parityCheckDate < :parityCheckedBeforeDate)" +
+                " AND c.createdDate < :createdBeforeDate " +
+                " ORDER BY c.createdDate asc";
+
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
```diff
@@ -188,4 +188,26 @@ public Long findMaxId() {
                 .setMaxResults(size)
                 .getResultList();
     }
+
+    public Optional<ChargeEntity> findChargeToExpunge(int ageOfChargeInDays) {
```
sfount: I think we could be more specific with the naming of `ageOfChargeInDays`, the interface sounds a bit like it could find a charge of exactly this number of days. 

Not sure exactly what it should be, maybe something like `minimumChargeAgeInDays`, the world `threshold` might fit in somewhere

src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
```diff
@@ -188,4 +188,26 @@ public Long findMaxId() {
...
+        String query = "SELECT c FROM ChargeEntity c " +
+                "WHERE c.parityCheckDate < :parityCheckedBeforeDate" +
```
sfount: Should this query factor in if `parityCheckDate` is `NULL`? 

My understanding is that if it's `NULL` it has never been checked and should qualify for expunging

sfount (commented):
Code and query look good to me! Small comments about the condition/ naming, nothing should be blocking though

src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
kbottla: Good spot! updated

sfount (approved):

PR #2079, opened 2020-02-05
PP-6085: Use env-map buildpack to set env vars

richardTowers:
WHAT YOU DID
------------

Should mostly be similar to the other applications we've been doing. A
few things to note:

* because this is a java app, env-map.yml needs to live in src/main/resources so it gets built and deployed with the jar file
* the jq is quite complicated because:
  * there's more than one user-provided service, so we need to disambiguate by name, not index
  * card-connector-db might be "postgres" (staging) or "user-provided" (dev)
  * ssl_option is only set in the user-provided service, and it needs a default

How to test
-----------

`cf push --vars-file staging.yml --var space=staging`

Code review checklist
---------------------

- Logging
- Documentation
    - Introduced any new environment variables / removed existing environment variable
    - Added new API / updated existing API definition

### DIFF
--- a/src/main/resources/env-map.yml
+++ b/src/main/resources/env-map.yml
@@ -1,0 +1,11 @@
+env_vars:
+  DB_HOST:                         '.[][] | select(.name == "card-connector-db") | .credentials.host                           '
+  DB_NAME:                         '.[][] | select(.name == "card-connector-db") | .credentials.name                           '
+  DB_PASSWORD:                     '.[][] | select(.name == "card-connector-db") | .credentials.password                       '
+  DB_USER:                         '.[][] | select(.name == "card-connector-db") | .credentials.username                       '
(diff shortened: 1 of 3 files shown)


src/main/resources/env-map.yml
```diff
@@ -0,0 +1,11 @@
+env_vars:
+  DB_HOST:                         '.[][] | select(.name == "card-connector-db") | .credentials.host                           '
```
danworth, line 2: 👍 great point about selecting on the service name to allow for dev set-up , I'll need to update Adminusers.

danworth:
@richardTowers I've rebased with master since pact tests were failing on CI due to I think missing changes from master (was quite a way behind).

danworth (approved):
Looks good 👍

PR #2087, opened 2020-02-06
PP-6077 Refactor get payment with metadata pact

nimalank7:
Description:
- Modifies the state for the pact test so that it holds metadata

### DIFF
--- a/src/test/java/uk/gov/pay/connector/pact/ContractTest.java
+++ b/src/test/java/uk/gov/pay/connector/pact/ContractTest.java
@@ -254,7 +254,7 @@ public class ContractTest {
     }
 
     @State("a charge exists")
-    public void aChargeExists(Map<String, String> params) {
+    public void aChargeExists(Map<String, String> params) throws Exception {
         String gatewayAccountId = params.get("gateway_account_id");
         Long chargeId = ThreadLocalRandom.current().nextLong(100, 100000);
         String chargeExternalId = params.get("charge_id");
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/connector/pact/ContractTest.java
```diff
@@ -270,6 +270,7 @@ public void aChargeExists(Map<String, String> params) {
                 .withTransactionId(params.get("gateway_transaction_id"))
                 .withDescription("Test description")
                 .withReference(ServicePaymentReference.of("aReference"))
+                .withExternalMetadata(new ExternalMetadata(objectMapper.readValue("{\"ledger_code\":123, \"some_key\":\"key\"}", Map.class)))
```
oswaldquek, line 273: is there a reason for putting this json string in here and not in the params here https://github.com/alphagov/pay-publicapi/blob/master/src/test/resources/pacts/publicapi-connector-get-payment.json#L14?

nimalank7: It's for the reason we discussed - because the pact is running against master it won't pass as the param isn't being sent over. So for the time being hard-code then when publicapi is merged I can then submit a follow up PR to use params

oswaldquek: ah ok!

oswaldquek (approved):

PR #2092, opened 2020-02-07
PP-6098 Extend LedgerTransaction

kbottla:
## WHAT YOU DID
- Extended LedgerTransaction with all attributes required to deserialize ledger response during parity check
- LedgerTransaction can potentially be moved to pay-java-commons to use in ledger as well as connector to serialise/deserialise Ledger transaction


sfount (approved):
Code looks good to me 👍 .

As discussed this morning this could look at sharing serialisation code with public API -- in the future we shouldn't need to do as in depth a parity check before removing charges. 

We can trust the infrastructure will propagate events as long as they are validly emitted from Connector, while we validate historically backfilled payments this will be useful.

PR #2093, opened 2020-02-10
PP-6098 Parity checker: match common fields

kbottla:
## WHAT 
- Updated ParityCheckerService to match common fields for payments. More matching to come in next set of PRs
- Transaction & charge matcher matches all fields and logs (mismatches) before returning true/false.
- Updated POJO/Fixtures for common fields accessors/mutators
- Updated relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paritycheck/LedgerTransaction.java
+++ b/src/main/java/uk/gov/pay/connector/paritycheck/LedgerTransaction.java
@@ -45,16 +45,32 @@ public class LedgerTransaction {
         return amount;
     }
 
+    public void setAmount(Long amount) {
+        this.amount = amount;
+    }
+
     public String getDescription() {
         return description;
+    }
+
+    public void setDescription(String description) {
+        this.description = description;
     }
 
     public String getReference() {
         return reference;
     }
 
+    public void setReference(String reference) {
+        this.reference = reference;
+    }
+
     public String getEmail() {
         return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
     }
 
     public boolean getDelayedCapture() {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/paritycheck/LedgerTransaction.java
```diff
@@ -45,18 +45,34 @@ public Long getAmount() {
...
 
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
     public String getReference() {
         return reference;
     }
 
+    public void setReference(String reference) {
```
alexbishop1, line 64: We actually have a wrapper type for this (`ServicePaymentReference`)

src/main/java/uk/gov/pay/connector/tasks/ParityCheckService.java
```diff
@@ -95,4 +100,54 @@ public boolean parityCheckChargeForExpunger(ChargeEntity chargeEntity) {
...
     }
+
+    private ParityCheckStatus checkParity(ChargeEntity chargeEntity, LedgerTransaction transaction) {
+        String externalId = chargeEntity.getExternalId();
+        ParityCheckStatus parityCheckStatus;
+
+        MDC.put(PAYMENT_EXTERNAL_ID, externalId);
+
+        if (transaction == null) {
+            logger.info("Transaction missing in Ledger for Charge");
```
alexbishop1, line 111: Even though we have the payment ID as metadata, it might be good to put it in the message as well, in case we ever find ourselves looking at this in a context where we only see the message (since it’s so easy to do)

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/tasks/ParityCheckService.java
kbottla: 👍 will update in next PR

src/main/java/uk/gov/pay/connector/paritycheck/LedgerTransaction.java
kbottla: A bit hesitated to use `ServicePaymentReference`. Will update in next PR

PR #2095, opened 2020-02-10
PP-6072 Add DAO method to search for gateway accounts

stephencdaly:
Add method to search gateway accounts by multiple parameters. Currently this is just by gateway account ids and whether MOTO is enabled or not. The is being added so we can filter gateway accounts in toolbox, and is intended to be extended in future with additional search parameters.

Add GatewayAccountSearchParams which is the query parameters bean which will be populated by the resource. This follows the pattern used for searching in ledger.

This new DAO method will be used by the GET `/v1/api/accounts` search resource, and will replace the existing `list` method in the DAO that takes a list of gateway account ids as a parameter.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/api/CommaDelimitedSetParameter.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/api/CommaDelimitedSetParameter.java
@@ -1,0 +1,30 @@
+package uk.gov.pay.connector.common.model.api;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+public class CommaDelimitedSetParameter {
+    private List<String> elements;
+    private String queryString;
+
+    public CommaDelimitedSetParameter(String queryString) {
+        this.queryString = queryString;
+        elements = isBlank(queryString)
+                ? new ArrayList<>()
+                : List.of(queryString.split(","));
+    }
+
+    public boolean isNotEmpty() {
+        return !elements.isEmpty();
+    }
+
+    public String getRawString() {
+        return queryString;
+    }
+
+    public List<String> getParameters() {
+        return elements;
+    }
+}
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/common/model/api/CommaDelimitedSetParameter.java
```diff
@@ -0,0 +1,30 @@
...
+    private String queryString;
+
+    public CommaDelimitedSetParameter(String queryString) {
+        this.queryString = queryString;
+        elements = isBlank(queryString)
+                ? new ArrayList<>()
+                : List.of(queryString.split(","));
+    }
+
+    public boolean isNotEmpty() {
```
rorymalcolm, line 19: I think, `isEmpty()` would be better here, but if that's convention from ledger, that's fine

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParamsTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertThat(queryMap, IsMapWithSize.anEmptyMap());
+    }
+
+    @Test
+    public void shouldNotIncludeAccountIdsInQueryMapForEmptyString() {
+        var params = new GatewayAccountSearchParams();
+        params.setAccountIds(new CommaDelimitedSetParameter(""));
+
+        Map<String, Object> queryMap = params.getQueryMap();
+        assertThat(queryMap, IsMapWithSize.anEmptyMap());
```
rorymalcolm: static import?

rorymalcolm (dismissed):

src/main/java/uk/gov/pay/connector/common/model/api/CommaDelimitedSetParameter.java
stephencdaly: Because I copied this class from ledger. It's useful

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParamsTest.java
stephencdaly: staticly imported now

src/main/java/uk/gov/pay/connector/common/model/api/CommaDelimitedSetParameter.java
rorymalcolm: nah I meant the equality of it

stephencdaly: what's wrong with it?

stephencdaly: would you prefer it it be `isEmpty`? I'm not fussed either way, just copied it as was

rorymalcolm (approved):

PR #2099, opened 2020-02-11
PP-6072 Allow searching for gateway accounts by additional parameters

stephencdaly:
Amend query parameters and query construction for the DAO to allow for
filtering gateway accounts by additional parameters:
`apple_pay_enabled`
`google_pay_enabled`
`requires_3ds`
`type` (test/live)
`payment_provider`

This is for use in toolbox for finding which gateway accounts have certain features enabled etc, which is implemented here https://github.com/alphagov/pay-toolbox/pull/295

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -196,15 +196,20 @@ Content-Type: application/json
 
 Retrieves a collection of accounts, optionally filtering by the provided query parameters
 
-| Field          | Always present | Description                                                                                                                                |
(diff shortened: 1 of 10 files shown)


oswaldquek:
I've nothing against this but isn't PP-6072 about searching for moto enabled gateway accounts?

stephencdaly:
You caught me! that bit is already done and this was only a tiny bit of extra effort to sneak in. I think people will find it useful

nimalank7 (approved):
Looks good to me!

PR #2100, opened 2020-02-11
PP-6065 remove foreign key constrain on charge_id on refunds table

SandorArpa:
## WHAT YOU DID
This is the second part of the bigger story of de-coupling RefundsEntity from ChargeEntity.

- drop foreign key constraint
- drop not null constraints for charge_id
- drop indexes for charge_id
- add index for charge_external_id on refunds_history

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1430,4 +1430,36 @@
         </createIndex>
     </changeSet>
 
+    <changeSet id="drop foreign key constraint on charge_id on refunds table" author="">
+        <dropForeignKeyConstraint
+                baseTableName="refunds"
+                constraintName="fk__refunds_charges"/>
+    </changeSet>
+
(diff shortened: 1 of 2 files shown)


src/main/resources/migrations.xml
```diff
@@ -1430,4 +1430,38 @@
...
+    
+    <changeSet id="drop index on charge_id on refunds table" author="">
+        <dropIndex tableName="refunds" indexName="idx_refunds_charge_id"/>
+    </changeSet>
+
+    <changeSet id="drop index on charge_id on refunds_history table" author="">
+        <dropIndex tableName="refunds_history" indexName="idx_refunds_history_charge_id"/>
+    </changeSet>
+
+    <changeSet id="create charge_external_id index on refunds_history table" author="">
```
kbottla: Do we need to indicate `CONCURRENTLY` to avoid locking table and also `runInTransaction=false`

kbottla (approved):

PR #2101, opened 2020-02-11
PP-6098 Parity checker: match gateway_account and feature specific fields

kbottla:
## WHAT

- Updated ParityCheckerService to match gateway account details on Ledger transaction and feature specific fields
- Updated relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/ParityCheckService.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/ParityCheckService.java
@@ -13,6 +13,7 @@ import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.model.domain.ParityCheckStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.paritycheck.Address;
 import uk.gov.pay.connector.paritycheck.CardDetails;
 import uk.gov.pay.connector.paritycheck.LedgerService;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/tasks/ParityCheckService.java
```diff
@@ -144,6 +148,9 @@ private boolean matchCommonFields(ChargeEntity chargeEntity, LedgerTransaction t
         fieldsMatch = fieldsMatch && isEquals(chargeEntity.getEmail(), transaction.getEmail(), "email");
         fieldsMatch = fieldsMatch && isEquals(chargeEntity.getReturnUrl(), transaction.getReturnUrl(), "return_url");
         fieldsMatch = fieldsMatch && isEquals(chargeEntity.getGatewayTransactionId(), transaction.getGatewayTransactionId(), "gateway_transaction_id");
+        fieldsMatch = fieldsMatch && isEquals(
+                ISO_INSTANT_MILLISECOND_PRECISION.format(chargeEntity.getCreatedDate()),
```
sfount, line 152: I'm not sure we'll be able to match created date to millisecond precision, I seem to remember there was some reason these could vary by (mili? micro?) seconds due to how the events are stored or emitted.

sfount (commented):
Code looks good to me, well thought through test cases 👍 question around the feasibility of matching exact created dates, potentially rounding up to seconds will get around this

src/main/java/uk/gov/pay/connector/tasks/ParityCheckService.java
kbottla: We used charge CREATED event to set Ledger transaction created_date. So current matching will be updated in next PR to consider ChargeEvent date. It should then be precise.

sfount (approved):

PR #2105, opened 2020-02-12
PP-6098 Parity checker: match settlement summary and refundability

kbottla:
## WHAT 
- Updated ParityChecker to match settlement, refund status fields
- `createdDate` is taken from CREATED event which is what used for PaymentCreated event
- `Expunge charge` : Updated dao method to also remove charge_events. More to be deleted in PP-6130
- Updated relevant tests
- Fixed integration test for changes - for refund_summary.status used hard coded value for
  now but this can be extended if we want to try tests with different refund-ability status


rorymalcolm (approved):

rorymalcolm:
this was the 5000th commit to connector

PR #2106, opened 2020-02-12
PP-6090 Refund events use Charge POJO

sfount:
Remove reliance on ChargeEntity for refund event calculation.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -273,6 +273,11 @@ public class ChargeService {
                 .orElseGet(Optional::empty);
     }
 
+    public Optional<Charge> findCharge(String chargeExternalId) {
+        return chargeDao.findByExternalId(chargeExternalId)
+                .map(Charge::from);
+    }
+
     public Optional<Charge> findCharge(String chargeExternalId, Long gatewayAccountId) {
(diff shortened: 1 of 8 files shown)


kbottla (dismissed):
Looks good. Might need to rebase with master before merging

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -273,6 +273,17 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
...
     }
 
+    public Optional<Charge> findCharge(String chargeExternalId) {
+        Optional<ChargeEntity> maybeChargeEntity = chargeDao.findByExternalId(chargeExternalId);
+
+        if(maybeChargeEntity.isPresent()) {
+            return maybeChargeEntity.map(Charge::from);
+        }
+        else{
+            return ledgerService.getTransaction(chargeExternalId).map(Charge::from);
```
kbottla, line 283: is this reverting https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java#L283

sfount: This path is used when we're calculating events (non user inputted values), the route for checking with gateway account is still used for the refund process (user requested external ID)

kbottla (approved):

PR #2107, opened 2020-02-12
BAU — Make some specific string types extend WrappedStringValue

alexbishop1:
Make some specific string types extend `WrappedStringValue` to leverage its `equals(…)`, `hashCode()` and `toString()` methods.

Delete `CardHolderName` because it is unused.


SandorArpa (approved):
LOKTM

PR #2110, opened 2020-02-13
PP-6127 Add connector secret service

danworth:
Env vars which are specific to card-connector and differ by environment
are to be provided via the card-connector-secret-service. The mapping
between these vars and their keys within the vars returned by
card-connector-secret-service are defined within env-map.yml.

## WHAT
I've tested manually pushing this into staging and card-connector starts. It also appears to have the correct vars set.


rjbaker (approved):
👍

richardTowers (approved):

PR #2112, opened 2020-02-13
PP-6065 drop charge_id column in refunds and refunds_history tables

SandorArpa:
## WHAT YOU DID
- drop no longer needed column


sfount (approved):
Code changes look to me 👍 our policy is generally to separate migrations and code (DAO) changes but it looks like this is small and self contained enough to not cause issues down the line 🤞

PR #2113, opened 2020-02-13
PP-6065 switch ChargeRefundsResource to use ChargeService

SandorArpa:
## WHAT YOU DID
- use ChargeService instead of ChargeDao

This is due to the new purge of charges change

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/resource/ChargeRefundsResource.java
+++ b/src/main/java/uk/gov/pay/connector/refund/resource/ChargeRefundsResource.java
@@ -1,9 +1,12 @@
 package uk.gov.pay.connector.refund.resource;
 
-import uk.gov.pay.connector.charge.dao.ChargeDao;
-import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.Charge;
+import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.gateway.model.GatewayError;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/refund/resource/ChargeRefundsResource.java
```diff
@@ -74,7 +77,7 @@ private void validateRefundRequest(long amount) {
     @Path("/v1/api/accounts/{accountId}/charges/{chargeId}/refunds/{refundId}")
     @Produces(APPLICATION_JSON)
     public Response getRefund(@PathParam("accountId") Long accountId, @PathParam("chargeId") String chargeId, @PathParam("refundId") String refundId, @Context UriInfo uriInfo) {
-        return chargeDao.findByExternalIdAndGatewayAccount(chargeId, accountId)
+        return chargeService.findCharge(chargeId, accountId)
                 .map(chargeEntity -> getRefundResponse(chargeEntity, refundId, accountId, uriInfo))
```
kbottla:
```suggestion
                .map(charge -> getRefundResponse(charge, refundId, accountId, uriInfo))
```

kbottla (dismissed):
Looks good. Just a minor suggestion

kbottla (approved):

sfount:
I think our APIs default strategy is to go to Ledger for refunds, I don't think these resources need to be updated to factor in expunged charges. We'll have to review the behaviour of get one refund but this should end up going to Ledger anyway.

PR #2116, opened 2020-02-13
PP-6130 Expunger Expunges Further Dependent Tables

rorymalcolm:
- The expunging job, when run, will now delete auxiliary information
from tables that are not charges but contain fields which reference it,
either implicitly through FK relationships or tenuously.

- A test has been written to ensure this functionality works as we
expect it to.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
@@ -212,9 +212,25 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
                 .getResultList().stream().findFirst();
     }
 
-    public void expungeCharge(Long id) {
+    public void expungeCharge(Long id, String externalId) {
+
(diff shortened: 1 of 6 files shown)


sfount (approved):
LGTM 👍 really good stuff wrangling the tests

src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
```diff
@@ -212,12 +212,28 @@ public Long findMaxId() {
...
                 .setParameter(1, id)
                 .executeUpdate();
 
+        entityManager.get()
+                .createNativeQuery("delete from tokens where charge_id = ?1")
+                .setParameter(1, id)
+                .executeUpdate();
+
+        entityManager.get()
+                .createNativeQuery("delete from emitted_events where resource_external_id = ?1")
```
kbottla, line 228: This is likely to be very slow. Could also pass `resource_type=PAYMENT` so index will be used.

PR #2117, opened 2020-02-13
PP-6130 Modify Delete Emitted Events Query

rorymalcolm:
- Krishna rightly raised that the lack of indexed use on the
`external_id` field would make the current delete query slow, this adds
an extra discriminatory parameter to use an index and speed up query
execution.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
@@ -225,8 +225,9 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
                 .executeUpdate();
 
         entityManager.get()
-                .createNativeQuery("delete from emitted_events where resource_external_id = ?1")
-                .setParameter(1, externalId)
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
```diff
@@ -225,7 +225,7 @@ public void expungeCharge(Long id, String externalId) {
                 .executeUpdate();
 
         entityManager.get()
-                .createNativeQuery("delete from emitted_events where resource_external_id = ?1")
+                .createNativeQuery("delete from emitted_events where resource_external_id = ?1 AND resource_type=PAYMENT")
```
kbottla: Should be other way round `where resource_type=PAYMENT" and resource_external_id = ?1` as column order on index is resource_type, resource_external_id

rorymalcolm: cheers, fixed it

kbottla (approved):

PR #2122, opened 2020-02-14
PP-6090 Refund availability for events support historic

sfount:
Charges not fetched from the database will have an external status
already provided; this should be used in favour of re-calculating from
the charge entity itself.


src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
```diff
@@ -158,14 +159,22 @@ public static Event createRefundEvent(RefundHistory refundHistory, Class<? exten
                     Optional.ofNullable(charge)
                     .map(c -> {
                         List<RefundEntity> refundEntityList = refundDao.findRefundsByChargeExternalId(c.getExternalId());
+                        ExternalChargeRefundAvailability refundAvailability;
+
+                        if(charge.isHistoric()) {
```
kbottla, line 164: Can we have a test to cover this change ?

kbottla (dismissed):

kbottla (approved):

PR #2123, opened 2020-02-14
PP-6098 Paritychecker - Fix matcher for non existing cardType

kbottla:
- Fix matching cardType when value is empty in connector
- rename ExpungeResource query parameter

### DIFF
--- a/src/main/java/uk/gov/pay/connector/expunge/resource/ExpungeResource.java
+++ b/src/main/java/uk/gov/pay/connector/expunge/resource/ExpungeResource.java
@@ -26,7 +26,7 @@ public class ExpungeResource {
     @POST
     @Path("/v1/tasks/expunge")
     @Produces(APPLICATION_JSON)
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/expunge/resource/ExpungeResource.java
```diff
@@ -26,7 +26,7 @@ public ExpungeResource(ChargeExpungeService chargeExpungeService) {
     @POST
     @Path("/v1/tasks/expunge")
     @Produces(APPLICATION_JSON)
-    public Response expungeCharges(@QueryParam("number_of_charges_to_expire") Integer noOfChargesToExpunge) {
+    public Response expungeCharges(@QueryParam("number_of_charges_to_expunge") Integer noOfChargesToExpunge) {
```
sfount, line 29: 👏

sfount (approved):

PR #2126, opened 2020-02-18
PP-6143 Resource identifier for expunging job

sfount:
* use existing correlation id if it exists, if not generate one for the
running expunger job
* include payment external id uniformly across logging contexts

### DIFF
--- a/src/main/java/uk/gov/pay/connector/expunge/resource/ExpungeResource.java
+++ b/src/main/java/uk/gov/pay/connector/expunge/resource/ExpungeResource.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.connector.expunge.resource;
 
+import org.slf4j.MDC;
 import uk.gov.pay.connector.expunge.service.ChargeExpungeService;
 
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/expunge/resource/ExpungeResource.java
```diff
@@ -27,6 +31,8 @@ public ExpungeResource(ChargeExpungeService chargeExpungeService) {
     @Path("/v1/tasks/expunge")
     @Produces(APPLICATION_JSON)
     public Response expungeCharges(@QueryParam("number_of_charges_to_expunge") Integer noOfChargesToExpunge) {
+        String correlationId = MDC.get(HEADER_REQUEST_ID) == null ? "ExpungeResource-" + UUID.randomUUID().toString() : MDC.get(HEADER_REQUEST_ID);
```
kbottla, line 34: (non-blocking) We may add header directly as it is unlikely to have correlation id (unless there is a plan to have the task invoked from toolbox ?)

kbottla (approved):
Looks good. Just a minor comment

PR #2127, opened 2020-02-18
PP-6186: Cancel charge with gateway if it's in a terminable state

oswaldquek:
The scenario here is the attempt to cancel an ePDQ charge by a user or via the
API. With ePDQ integration, some services configure their payments to be
"one-step", that is, a charge is authorised and captured in one step. (GOV.UK
Pay assumes "two-step", that is, separate API calls have to be made for
authorisation and capturing a charge.)

The result of this is that when a cancel is attempted, the charge may already
have been captured, which is a non-terminable state. This commit adds a check
to see if the charge is in a terminable state before canceling.

At present, the only implementation of the PaymentProvider that implements the
`queryPaymentStatus` method is the EpdqPaymentProvider. We have therefore
amended the `QueryService.isTerminableWithGateway` method to return `true` if
the implementation of PaymentProvider throws an UnsupportedOperationException.
This is the equivalent of saying we assume the charge can be canceled if the
PSP is not epdq which is a reasonable assumption because the only complaints we get
are services which use ePDQ as their PSP.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
@@ -70,11 +70,15 @@ public class ChargeCancelService {
                 .of(chargeStatus).getAuthorisationStage();
 
         boolean cancellableWithGateway = authorisationStage == ExpirableChargeStatus.AuthorisationStage.POST_AUTHORISATION
-                || (authorisationStage == ExpirableChargeStatus.AuthorisationStage.DURING_AUTHORISATION
-                && queryService.isTerminableWithGateway(chargeEntity));
+                || authorisationStage == ExpirableChargeStatus.AuthorisationStage.DURING_AUTHORISATION;
 
         if (cancellableWithGateway) {
-            cancelChargeWithGatewayCleanup(chargeEntity, statusFlow);
+            boolean isTerminable = queryService.isTerminableWithGateway(chargeEntity);
+
+            if (isTerminable) {
+                cancelChargeWithGatewayCleanup(chargeEntity, statusFlow);
+            }
+
         } else {
             nonGatewayCancel(chargeEntity, statusFlow);
         }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -70,11 +70,15 @@ private void doCancel(ChargeEntity chargeEntity, StatusFlow statusFlow) {
                 .of(chargeStatus).getAuthorisationStage();
 
         boolean cancellableWithGateway = authorisationStage == ExpirableChargeStatus.AuthorisationStage.POST_AUTHORISATION
-                || (authorisationStage == ExpirableChargeStatus.AuthorisationStage.DURING_AUTHORISATION
-                && queryService.isTerminableWithGateway(chargeEntity));
+                || authorisationStage == ExpirableChargeStatus.AuthorisationStage.DURING_AUTHORISATION;
 
         if (cancellableWithGateway) {
-            cancelChargeWithGatewayCleanup(chargeEntity, statusFlow);
+            boolean isTerminable = queryService.isTerminableWithGateway(chargeEntity);
```
stephencdaly, line 76: I think what was happening before was:

- If the authorisation stage is POST_AUTHORISATION we will always try to cancel with the gateway in order to ensure funds are released. This includes for the gateways we don't query with first.
- For DURING_AUTHORISATION we query with the gateways (that we've implement this for) whether the payment is cancellable. And if it is try to cancel it. I think this is mainly for the sake of ePDQ where we might not think the charge is authorised yet when it actually is. Possibly for Smartpay and Stripe we can be more sure that the charge hasn't been authorised, so there's no need to try to cancel with them and they'll expire the payment in their system normally.

So I think for our change, we need to do our checking with the gateway only if that's implemented. And if it isn't still always try to cancel for POST_AUTHORISATION and don't for DURING_AUTHORISATION states.

oswaldquek:
this will be replaced by https://github.com/alphagov/pay-connector/compare/PP-6186-update-charge-with-gateway-status-if-not-cancellable-with-gateway?expand=1, closing

PR #2129, opened 2020-02-19
PP-6184 Handle notifications for expunged charges

kbottla:
## WHAT 
- Refunds are allowed for charges expunged from connector and so  we should be able to process relevant refunds notifications for expunge charges. Refund notifications for all payment providers need to get charge and also gateway account information to be able to verify notification signature, send refund emails to users.

- Added a new ChargeService method `findByProviderAndTransactionIdFromDbOrLedger` which looks  up Ledger if charge has been expunged
- Updated notifications to use new ChargeService method
- Epdq notifications still need connector internal charge status, to map target status but
  only for inflight payments. So it should be safe to continue using it
  https://github.com/alphagov/pay-connector/blob/5ac4d251a1b7009cfc4f90260565951e8e4556f3/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java#L145

More to come ...
- To remove reliance on internal `ChargeStatus` for all SmartPay notifications

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -296,15 +296,13 @@ public class ChargeService {
     }
 
     public Optional<Charge> findByProviderAndTransactionIdFromDbOrLedger(String paymentGatewayName, String gatewayTransactionId) {
-        Optional<ChargeEntity> maybeChargeEntity = chargeDao.findByProviderAndTransactionId(paymentGatewayName, gatewayTransactionId);
+        return Optional.ofNullable(chargeDao.findByProviderAndTransactionId(paymentGatewayName, gatewayTransactionId)
+                .map(Charge::from)
+                .orElseGet(() -> findChargeFromLedger(paymentGatewayName, gatewayTransactionId).orElse(null)));
+    }
 
-        if(maybeChargeEntity.isPresent()) {
-            return maybeChargeEntity.map(Charge::from);
-        }
-        else{
-            return ledgerService.getTransactionForProviderAndGatewayTransactionId(paymentGatewayName,gatewayTransactionId)
-                    .map(Charge::from);
-        }
+    private Optional<Charge> findChargeFromLedger(String paymentGatewayName, String gatewayTransactionId) {
+        return ledgerService.getTransactionForProviderAndGatewayTransactionId(paymentGatewayName,gatewayTransactionId).map(Charge::from);
     }
 
     @Transactional
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -302,7 +302,7 @@ public class ChargeService {
             return maybeChargeEntity.map(Charge::from);
         }
         else{
-            return ledgerService.findByProviderAndGatewayTransactionId(paymentGatewayName,gatewayTransactionId)
+            return ledgerService.getTransactionForProviderAndGatewayTransactionId(paymentGatewayName,gatewayTransactionId)
                     .map(Charge::from);
         }
     }
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -295,6 +295,18 @@ public class ChargeService {
         }
     }
 
+    public Optional<Charge> findByProviderAndTransactionIdFromDbOrLedger(String paymentGatewayName, String gatewayTransactionId) {
+        Optional<ChargeEntity> maybeChargeEntity = chargeDao.findByProviderAndTransactionId(paymentGatewayName, gatewayTransactionId);
+
+        if(maybeChargeEntity.isPresent()) {
+            return maybeChargeEntity.map(Charge::from);
+        }
+        else{
+            return ledgerService.findByProviderAndGatewayTransactionId(paymentGatewayName,gatewayTransactionId)
+                    .map(Charge::from);
+        }
+    }
+
     @Transactional
     public Optional<ChargeEntity> updateCharge(String chargeId, PatchRequestBuilder.PatchRequest chargePatchRequest) {
         return chargeDao.findByExternalId(chargeId)
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
@@ -19,7 +19,9 @@ import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import static net.logstash.logback.argument.StructuredArguments.kv;
 import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
@@ -30,11 +32,12 @@ import static uk.gov.pay.connector.gateway.epdq.EpdqNotification.SHASIGN_KEY;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_SHA_OUT_PASSPHRASE;
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUNDED;
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUND_ERROR;
+import static uk.gov.pay.logging.LoggingKeys.PAYMENT_EXTERNAL_ID;
 
 public class EpdqNotificationService {
 
     private static final String PAYMENT_GATEWAY_NAME = PaymentGatewayName.EPDQ.getName();
-    private final Logger logger = LoggerFactory.getLogger(getClass());
+    private static final Logger logger = LoggerFactory.getLogger(EpdqNotificationService.class);
     private final ChargeService chargeService;
     private final SignatureGenerator signatureGenerator;
     private final ChargeNotificationProcessor chargeNotificationProcessor;
@@ -93,7 +96,7 @@ public class EpdqNotificationService {
 
         logger.info("Evaluating {} notification {}", PAYMENT_GATEWAY_NAME, notification);
 
-        final Optional<ChargeStatus> newChargeStatus = newChargeStateForChargeNotification(notification.getStatus(), ChargeStatus.fromString(charge.getStatus()));
+        final Optional<ChargeStatus> newChargeStatus = newChargeStateForChargeNotification(notification.getStatus(), charge);
 
         if (newChargeStatus.isPresent()) {
             chargeNotificationProcessor.invoke(notification.getTransactionId(), charge, newChargeStatus.get(), null);
@@ -139,7 +142,7 @@ public class EpdqNotificationService {
         return gatewayAccountEntity.getCredentials().get(CREDENTIALS_SHA_OUT_PASSPHRASE);
     }
 
-    private static Optional<ChargeStatus> newChargeStateForChargeNotification(String notificationStatus, ChargeStatus chargeStatus) {
+    private static Optional<ChargeStatus> newChargeStateForChargeNotification(String notificationStatus, Charge charge) {
         final EpdqNotification.StatusCode statusCode = EpdqNotification.StatusCode.byCode(notificationStatus);
 
         switch (statusCode) {
@@ -149,22 +152,34 @@ public class EpdqNotificationService {
                 return Optional.of(AUTHORISATION_SUCCESS);
 
             case EPDQ_AUTHORISED_CANCELLED:
-                switch (chargeStatus) {
-                    case USER_CANCEL_SUBMITTED:
-                        return Optional.of(USER_CANCELLATION_FLOW.getSuccessTerminalState());
-
-                    case EXPIRE_CANCEL_SUBMITTED:
-                        return Optional.of(EXPIRE_FLOW.getSuccessTerminalState());
-
-                    case SYSTEM_CANCEL_SUBMITTED:
-                    case CREATED:
-                    case ENTERING_CARD_DETAILS:
-                        return Optional.of(SYSTEM_CANCELLATION_FLOW.getSuccessTerminalState());
-                    default:
-                        return Optional.empty();
-                }
+                return newChargeStateForAuthorisationCancelledNotification(charge);
             case EPDQ_PAYMENT_REQUESTED:
                 return Optional.of(CAPTURED);
+            default:
+                return Optional.empty();
+        }
+    }
+
+    private static Optional<ChargeStatus> newChargeStateForAuthorisationCancelledNotification(Charge charge) {
+        if (charge.isHistoric() || isEmpty(charge.getStatus())) {
+            logger.info("Could not derive charge status for authorisation cancelled notification as charge [{}] has been expunged or status is empty {}",
+                    charge.getExternalId(),
+                    kv(PAYMENT_EXTERNAL_ID, charge.getExternalId()));
+            return Optional.empty();
+        }
+
+        ChargeStatus chargeStatus = ChargeStatus.fromString(charge.getStatus());
+        switch (chargeStatus) {
+            case USER_CANCEL_SUBMITTED:
+                return Optional.of(USER_CANCELLATION_FLOW.getSuccessTerminalState());
+
+            case EXPIRE_CANCEL_SUBMITTED:
+                return Optional.of(EXPIRE_FLOW.getSuccessTerminalState());
+
+            case SYSTEM_CANCEL_SUBMITTED:
+            case CREATED:
+            case ENTERING_CARD_DETAILS:
+                return Optional.of(SYSTEM_CANCELLATION_FLOW.getSuccessTerminalState());
             default:
                 return Optional.empty();
         }
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
@@ -5,12 +5,14 @@ import com.google.inject.persist.Transactional;
 import org.apache.http.NameValuePair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import uk.gov.pay.connector.charge.dao.ChargeDao;
-import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.processor.ChargeNotificationProcessor;
 import uk.gov.pay.connector.gateway.processor.RefundNotificationProcessor;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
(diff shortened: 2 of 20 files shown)


src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
```diff
@@ -98,10 +106,10 @@ private void handle(SmartpayNotification notification) {
         }
     }
 
-    private InterpretedStatus interpretStatus(SmartpayNotification notification, ChargeEntity charge) {
+    private InterpretedStatus interpretStatus(SmartpayNotification notification, String status) {
         return SmartpayStatusMapper.from(
                 notification.getStatus(),
-                ChargeStatus.fromString(charge.getStatus())
+                ChargeStatus.fromString(status)
```
sfount: As discussed this will throw an invalid charge status exception if the Connector status is unknown, just a reminder to myself that this should be removed immediately following (charge status isn't actually used further down the chain)

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -295,6 +295,18 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
...
+        Optional<ChargeEntity> maybeChargeEntity = chargeDao.findByProviderAndTransactionId(paymentGatewayName, gatewayTransactionId);
+
+        if(maybeChargeEntity.isPresent()) {
+            return maybeChargeEntity.map(Charge::from);
+        }
+        else{
+            return ledgerService.getTransactionForProviderAndGatewayTransactionId(paymentGatewayName,gatewayTransactionId)
+                    .map(Charge::from);
+        }
+    }
```
alexbishop1, line 306: Might be slightly nicer (but still not ideal) to do something like this?

```java
    public Optional<Charge> findByProviderAndTransactionIdFromDbOrLedger(String paymentGatewayName, String gatewayTransactionId) {
        return Optional.ofNullable(chargeDao.findByProviderAndTransactionId(paymentGatewayName, gatewayTransactionId)
                .map(Charge::from)
                .orElseGet(() -> findChargeFromLedger(paymentGatewayName, gatewayTransactionId).orElse(null)));
    }
    
    private Optional<Charge> findChargeFromLedger(String paymentGatewayName, String gatewayTransactionId) {
        return ledgerService.getTransactionForProviderAndGatewayTransactionId(paymentGatewayName,gatewayTransactionId).map(Charge::from);
    }
```

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -65,22 +68,26 @@ public void handleNotificationFor(String payload) {
...
             logger.error("{} notification {} could not be verified (associated charge entity not found)",
                     PAYMENT_GATEWAY_NAME, notification);
             return;
         }
 
-        ChargeEntity charge = maybeCharge.get();
-        if (!isValidNotificationSignature(notification, charge)) {
+        Charge charge = maybeCharge.get();
+        GatewayAccountEntity gatewayAccountEntity
+                = gatewayAccountService.getGatewayAccount(charge.getGatewayAccountId()).get();
```
alexbishop1, line 91: Rather than using `get()`, an `orElseThrow(…)` with an appropriate “Oh no! We couldn’t find a gateway account that should definitely be there!” message might be useful to us if the gateway account ever isn’t there.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -21,7 +22,8 @@
         this.chargeService = chargeService;
     }
 
-    public void invoke(String transactionId, ChargeEntity chargeEntity, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
+    public void invoke(String transactionId, Charge charge, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
+        ChargeEntity chargeEntity = chargeService.findChargeByExternalId(charge.getExternalId());
```
alexbishop1, line 26: This will throw if the charge isn’t in connector but I guess we’re assuming that’s OK because we’ll only get notifications for charges that haven’t been expunged.

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
```diff
@@ -74,8 +79,11 @@ private void handle(SmartpayNotification notification) {
             return;
         }
 
-        ChargeEntity charge = maybeCharge.get();
-        InterpretedStatus interpretedStatus = interpretStatus(notification, charge);
+        Charge charge = maybeCharge.get();
```
alexbishop1, line 83: Another place I think `orElseThrow(…)` would be good

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
```diff
@@ -74,8 +79,11 @@ private void handle(SmartpayNotification notification) {
...
+        GatewayAccountEntity gatewayAccountEntity =
+                gatewayAccountService.getGatewayAccount(charge.getGatewayAccountId()).get();
```
alexbishop1, line 86: And another!

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
```diff
@@ -74,8 +79,11 @@ private void handle(SmartpayNotification notification) {
...
+        
+        InterpretedStatus interpretedStatus = interpretStatus(notification, charge.getStatus());
```
alexbishop1: I think this will throw an exception if the charge has been expunged from connector.

As I understand it, if the change has been retrieved from Ledger, `charge.getStatus()` will return `null`. The `interpretStatus(…)` method makes a call to `ChargeStatus.fromString(status)`, which will throw an `IllegalArgumentException` is passed with `null`.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -1329,4 +1329,50 @@ public void shouldFindCharge_fromLedgerIfNotExists() {
...
+                chargeEntity.getExternalId())).thenReturn(Optional.of(transaction));
+
+        Optional<Charge> charge = service.findByProviderAndTransactionIdFromDbOrLedger("sandbox",
+                chargeEntity.getExternalId());
+
+        assertThat(charge.isPresent(), is(true));
+        final Charge result = charge.get();
+        assertThat(result.getAmount(), is(chargeEntity.getAmount()));
+        assertThat(result.getExternalId(), is(chargeEntity.getExternalId()));
+    }
```
alexbishop1, line 1377: Maybe a test for if it’s not in either?

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqNotificationServiceTest.java
```diff
@@ -43,19 +44,23 @@
...
-        when(mockGatewayAccountEntity.getCredentials()).thenReturn(ImmutableMap.of(CREDENTIALS_SHA_OUT_PASSPHRASE, shaPhraseOut));
+        gatewayAccountEntity = ChargeEntityFixture.defaultGatewayAccountEntity();
+        gatewayAccountEntity.setCredentials(ImmutableMap.of(CREDENTIALS_SHA_OUT_PASSPHRASE, shaPhraseOut));
+        charge = Charge.from(ChargeEntityFixture.aValidChargeEntity()
+                .withGatewayAccountEntity(gatewayAccountEntity)
+                .build());
 
-        when(mockChargeDao.findByProviderAndTransactionId(EPDQ.getName(), payId)).thenReturn(Optional.of(mockCharge));
+        when(mockGatewayAccountService.getGatewayAccount(charge.getGatewayAccountId())).thenReturn(Optional.of(gatewayAccountEntity));
+        when(mockChargeService.findByProviderAndTransactionIdFromDbOrLedger(EPDQ.getName(), payId)).thenReturn(Optional.of(charge));
```
alexbishop1, line 60: I think we’re going to need to be more rigorous here because `findByProviderAndTransactionIdFromDbOrLedger(…)` actually returns different sorts of charge objects depending on whether the charge came from connector or ledger, with consequential differences like whether `getStatus()` returns `null` or not.

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
kbottla: This will be handled in followup PR which considers if the charge is historic or not before using ChargeStatus.from(status)

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
```diff
@@ -55,59 +57,71 @@ public void shouldUpdateChargeToAuthorisationRejected_IfEpdqStatusIs2() {
...
     @Test
-    public void shouldUpdateChargeToSystemCancel_IfEpdqStatusIs6WithRelevantChargeStatus() {
-
-        List<ChargeStatus> ChargeStatuses = ImmutableList.of(SYSTEM_CANCEL_SUBMITTED, CREATED, ENTERING_CARD_DETAILS);
+    @Parameters({"SYSTEM CANCEL SUBMITTED", "CREATED", "ENTERING CARD DETAILS"})
+    public void shouldUpdateChargeToSystemCancel_IfEpdqStatusIs6WithRelevantChargeStatus(String status) {
+        charge = Charge.from(ChargeEntityFixture.aValidChargeEntity()
+                .withStatus(ChargeStatus.fromString(status))
+                .build());
+        when(mockChargeService.findByProviderAndTransactionIdFromDbOrLedger(any(), any())).thenReturn(Optional.of(charge));
```
alexbishop1: Not a fan of the `any()`s here: can we provide more specific values?

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -479,8 +479,6 @@ public void shouldNotReturn3dsFlexCredentials_whenGatewayIsNotAWorldpayAccount()
         givenSetup()
                 .get("/v1/frontend/accounts/" + gatewayAccountId)
                 .then()
-                .log()
-                .all()
```
alexbishop1, line 483: Good catch!

alexbishop1 (changes requested):
I’m concerned that some of the existing notification processing code seems to assume that every charge has a non-null status, which doesn’t seem to be the case for charges retrieved from ledger.

There’s also the issue of what happens if we get a notification for a charge that’s been expunged. We’ve said we think that won’t happen but I think it could if there’s some sort of problem (e.g. we timed out when authorising a payment so we have the expendable `AUTHORISATION UNEXPECTED ERROR` state but actually authorisation succeeded and one-step capture is on, so we get a notification telling us the payment has been captured). Even if not expunged, we won’t process that notification because it doesn’t fit with our idea of valid state transitions but the failure mode for an expunged charge will be different and I think it would be good to make sure there’s a code path for that that results in an appropriate error being logged.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
kbottla: For now, we are assuming we get charge notifications for inflight/recent payments. We will need to factor in expunged charge, if we get closer to expunging all terminal payments (which should also factor in the process to send events to ledger)

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
kbottla: This will be handled in a followup PR which updates `interpretStatus(SmartpayNotification notification, String status)` -> `interpretStatus(SmartpayNotification notification)` as we are not using ChargeStatus to get interpreted status.
Haven't updated here to keep PR small

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
kbottla: This should be fine as `chargeService.findByProviderAndTransactionIdFromDbOrLedger` returns empty optional if charge doesn't exist in both connector and ledger. We check for isEmpty before getting it https://github.com/alphagov/pay-connector/blob/d8103b2d110a14d5631df7b3a0caef32e798a8f8/src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java#L76

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
kbottla: 👍 Added

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
kbottla: Updated!

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqNotificationServiceTest.java
kbottla: Added tests in `EpdqNotificationServiceStatusMapperTest` for EpdqStatus `6` which Authorisation cancelled and the only place charge status is used.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
kbottla: Thought about it and wondered if we will ever have any charge with gateway_accound_id but no gateway account entity!! As discussed, will handle the exception if we ever come across this scenario

kbottla (commented):
Thanks for the review @alexbishop1 . I have addressed most of your comments in this PR. 

As mentioned in the PR desc, SmartpayNotificationService will be updated to remove reliance on ChargeStatus. And will also add some checks around gatewayAccountEntity. Both will be in followup PR(s).

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: 👍 Updated

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
kbottla: Actually @oswaldquek has removed this here https://github.com/alphagov/pay-connector/pull/2130

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -142,27 +152,39 @@ private String getShaOutPassphrase(ChargeEntity charge) {
...
             case EPDQ_PAYMENT_REQUESTED:
                 return Optional.of(CAPTURED);
             default:
                 return Optional.empty();
         }
     }
 
+    private static Optional<ChargeStatus> newChargeStateForAuthorisationCancelledNotification(Charge charge) {
+        if (charge.isHistoric() || isEmpty(charge.getStatus())) {
+            logger.info("Could not derive charge status for authorisation cancelled notification as charge [{}] has been expunged or status is empty {}",
```
alexbishop1, line 165: I think this should be logged at error level since it should Never Happen™

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -142,27 +152,39 @@ private String getShaOutPassphrase(ChargeEntity charge) {
...
+                    charge.getExternalId(),
+                    kv(PAYMENT_EXTERNAL_ID, charge.getExternalId()));
+            return Optional.empty();
```
alexbishop1, line 168: I pondered if throwing an exception would be better than returning an empty optional but that might cause ePDQ to retry, which probably isn’t so useful.

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
alexbishop1: Ah, missed that!

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
```diff
@@ -55,59 +58,98 @@ public void shouldUpdateChargeToAuthorisationRejected_IfEpdqStatusIs2() {
...
-
-        for (ChargeStatus chargeStatus : ChargeStatuses) {
-            when(mockCharge.getStatus()).thenReturn(chargeStatus.getValue());
-            final String payload = notificationPayloadForTransaction(payId, EPDQ_AUTHORISED_CANCELLED);
-            notificationService.handleNotificationFor(payload);
-        }
-        verify(mockChargeNotificationProcessor, times(3)).invoke(payId, mockCharge, SYSTEM_CANCELLATION_FLOW.getSuccessTerminalState(), null);
+        final String payload = notificationPayloadForTransaction(payId, EPDQ_AUTHORISED_CANCELLED);
+        notificationService.handleNotificationFor(payload);
+        verify(mockChargeNotificationProcessor, times(1)).invoke(payId, charge, SYSTEM_CANCELLATION_FLOW.getSuccessTerminalState(), null);
```
alexbishop1, line 92: Don’t need `times(1)`, it’s the default.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
```diff
@@ -55,59 +58,98 @@ public void shouldUpdateChargeToAuthorisationRejected_IfEpdqStatusIs2() {
...
-        when(mockCharge.getStatus()).thenReturn(CAPTURED.getValue());
+        charge = Charge.from(ChargeEntityFixture.aValidChargeEntity()
+                .withStatus(CAPTURED)
+                .build());
+        when(mockChargeService.findByProviderAndTransactionIdFromDbOrLedger(any(), any())).thenReturn(Optional.of(charge));
+
+        final String payload = notificationPayloadForTransaction(payId, EPDQ_AUTHORISED_CANCELLED);
+
+        notificationService.handleNotificationFor(payload);
+        verify(mockChargeNotificationProcessor, never()).invoke(any(), any(), any(), any());
```
alexbishop1, line 131: Probably just `verifyNoInteractions(mockChargeNotificationProcessor)` would work here.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
```diff
@@ -169,4 +211,10 @@ public void shouldNotProcessRefund_IfEpdqStatusIsUnknown() {
         notificationService.handleNotificationFor(payload);
         verify(mockRefundNotificationProcessor, never()).invoke(any(), any(), any(), any(), any(), any());
     }
+    
+    private Charge getCharge(boolean isHistoric){
```
alexbishop1, line 215: I think if this were two methods, `getInflightCharge()` and `getHistoricCharge()`, with the former always having a status, it would be a little clearer.

alexbishop1 (approved):
Thanks for making the changes. As discussed, approving this conditional on it only going out alongside any other changes that are necessary to ensure nothing will blow up if there’s a `null` status for a (historic, only-in-ledger) charge.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationServiceStatusMapperTest.java
kbottla: 👍 Followed existing tests without thinking much. Cleaned up in https://github.com/alphagov/pay-connector/pull/2136

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
kbottla: Updated in https://github.com/alphagov/pay-connector/pull/2136

PR #2136, opened 2020-02-20
PP-6184 Handle notifications for historic charges

kbottla:
## WHAT YOU DID
Followup for https://github.com/alphagov/pay-connector/pull/2129

- If for any reason we receive charge notifications for historic charges (expunged from connector), they are ignored
  and logged as error (this is required until notifications can be handled by connector for expunged charges)

- Similarly, if gateway account is not found for a charge (from Connector / Ledger),
  error is logger (as gateway account should always be there!)
   - For Worldpay notifications, follows the pattern when Charge is not found (which returns false) for Worldpay to retry notifications.

- Added new tests for above scenarios

- Cleaned up notifications tests to use `verifyNoInteractions` instead of verify(xxx, never())
- Also removed `gatewayName` from WorldpayNotificationService as linter keeps complaining about invoking conditionally during logging

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
@@ -32,6 +32,7 @@ import static uk.gov.pay.connector.gateway.epdq.EpdqNotification.SHASIGN_KEY;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_SHA_OUT_PASSPHRASE;
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUNDED;
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUND_ERROR;
+import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
 import static uk.gov.pay.logging.LoggingKeys.PAYMENT_EXTERNAL_ID;
 
 public class EpdqNotificationService {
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -99,6 +113,14 @@ public void handleNotificationFor(String payload) {
         final Optional<ChargeStatus> newChargeStatus = newChargeStateForChargeNotification(notification.getStatus(), charge);
 
         if (newChargeStatus.isPresent()) {
+            if(charge.isHistoric()){
+                logger.error("{} notification {} could not be processed as charge [{}] has been expunged from connector {} {}",
```
sfount, line 117: Should this be classified as an error? I wonder if we should just `warn` here as nothing has gone wrong we just don't yet support processing this. 

Same goes for other gateways.

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
```diff
@@ -81,13 +82,32 @@ private void handle(SmartpayNotification notification) {
...
         Charge charge = maybeCharge.get();
-        
-        GatewayAccountEntity gatewayAccountEntity =
-                gatewayAccountService.getGatewayAccount(charge.getGatewayAccountId()).get();
+
+        Optional<GatewayAccountEntity> mayBeGatewayAccountEntity =
+                gatewayAccountService.getGatewayAccount(charge.getGatewayAccountId());
+
+        if (mayBeGatewayAccountEntity.isEmpty()) {
+            logger.error("{} notification {} could not be processes (associated gateway account [{}] not found for charge [{}] {}, {})",
```
sfount, line 90:
```suggestion
            logger.error("{} notification {} could not be processed (associated gateway account [{}] not found for charge [{}] {}, {})",
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -20,11 +20,16 @@
...
 
+import static net.logstash.logback.argument.StructuredArguments.kv;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
+import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
+import static uk.gov.pay.logging.LoggingKeys.PAYMENT_EXTERNAL_ID;
 
 public class WorldpayNotificationService {
 
+    private static final String PAYMENT_GATEWAY_NAME = PaymentGatewayName.WORLDPAY.getName();
```
sfount, line 31: Nice 👍

sfount (approved):
Code LGTM, really good cleanup with constants and tests 👍 very minor comments

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
kbottla: We shouldn't ideally get notifications (we care about) once charge is expunged. As this is very rare, logging as error will have sentry alert us if this the case and also if we reduce connector to concept of inflight payments only. 

This could get a closer look once we start thinking about connector processing charge notifications for expunged charges

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayNotificationService.java
kbottla: good catch! Fixing in another open PR https://github.com/alphagov/pay-connector/pull/2137

PR #2154, opened 2020-02-25
PP-6186 allow the status of a charge to be forcibly updated

oswaldquek:
- Allow a charge to be forcibly transitioned from any state to certain
target states (for now just `CAPTURED`).
- This is to allow for the gateway considering the charge to be in a
different state to what we think it's in.
- This adds new events to be emitted to ledger per status to be
transitioned to. These events need to be added into the enum in the
ledger codebase to link them to the appropriate TransactionStatus.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -25,11 +25,11 @@ import uk.gov.pay.connector.charge.model.PrefilledCardHolderDetails;
 import uk.gov.pay.connector.charge.model.ServicePaymentReference;
 import uk.gov.pay.connector.charge.model.builder.AbstractChargeResponseBuilder;
 import uk.gov.pay.connector.charge.model.domain.Auth3dsDetailsEntity;
+import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.model.domain.ParityCheckStatus;
 import uk.gov.pay.connector.charge.model.domain.PersistedCard;
-import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.telephone.PaymentOutcome;
 import uk.gov.pay.connector.charge.model.telephone.Supplemental;
 import uk.gov.pay.connector.charge.model.telephone.TelephoneChargeCreateRequest;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -644,6 +648,24 @@ public ChargeEntity transitionChargeState(String chargeId, ChargeStatus targetCh
...
+        ChargeStatus fromChargeState = ChargeStatus.fromString(charge.getStatus());
+
+        return PaymentGatewayStateTransitions.getEventForForceUpdate(targetChargeState).map(eventClass -> {
+            charge.setStatusIgnoringValidTransitions(targetChargeState);
+            ChargeEventEntity chargeEventEntity = chargeEventDao.persistChargeEventOf(charge);
+
+            if (shouldEmitPaymentStateTransitionEvents) {
+                stateTransitionService.offerPaymentStateTransition(
+                        charge.getExternalId(), fromChargeState, targetChargeState, chargeEventEntity,
+                        StatusCorrectedToCapturedToMatchGatewayStatus.class);
```
stephencdaly: The last argument here should be `eventClass`

stephencdaly (changes requested):
Spotted a small mistake.

Also do you think it's worth adding a test in StateTransitionServiceTest for the new public method we've added?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: Sorry I don't follow... `StatusCorrectedToCapturedToMatchGatewayStatus` is an event class?

oswaldquek:
I don't think we need to add a test. There is a new method but it's just a refactor - pulling out existing code into another which is already tested in https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/queue/StateTransitionServiceTest.java#L54

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
stephencdaly: PaymentGatewayStateTransitions.getEventForForceUpdate returns an Optional<Class<T extends Event> which we then map to the variable `eventClass` in the lambda. We should use that variable. The variable will happen to have a value of `StatusCorrectedToCapturedToMatchGatewayStatus.class` always, as currently we're only allowing transitions to `CAPTURED` but the point of the method is we could allow transitions to other states if `getEventForForceUpdate` returns an event class which would be a different class for different target states.

stephencdaly:
> I don't think we need to add a test. There is a new method but it's just a refactor - pulling out existing code into another which is already tested in https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/queue/StateTransitionServiceTest.java#L54

Yeh that's fair enough

stephencdaly (approved):

PR #2156, opened 2020-02-25
BAU Log payment external id when transition fails for notification

stephencdaly:
It's difficult linking the failures to payments as currently we're only logging the gateway id. Log the external id too.

Add structured arguments for the other fields we log too.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
@@ -12,6 +12,12 @@ import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import javax.inject.Inject;
 import java.time.ZonedDateTime;
 
+import static net.logstash.logback.argument.StructuredArguments.kv;
+import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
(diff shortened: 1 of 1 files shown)


sfount (dismissed):

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -33,8 +39,12 @@ public void invoke(String transactionId, Charge charge, ChargeStatus newStatus,
             // don't log an error if we're trying to transition to the same state as this can happen if we've already
             // processed the notification
             if (!e.getCurrentState().equals(e.getTargetState())) {
-                logger.error("{} ({}) notification '{}' could not be used to update charge: {}",
-                        gatewayAccount.getGatewayName(), gatewayAccount.getId(), transactionId, e.getMessage());
+                logger.error(String.format("%s (%s) notification '%s' could not be used to update charge: %s",
+                        gatewayAccount.getGatewayName(), gatewayAccount.getId(), transactionId, e.getMessage()),
```
sfount: Not introduced in this PR but `transactionId` was hard to reason about making sure the it was the payment provider's ID, a more specific name used throughout might make it easier to understand; something like `gatewayTransactionId`

stephencdaly: good point, i've updated here

sfount (approved):

PR #2165, opened 2020-03-02
PP-6189 Use gateway status if terminal instead of expiring

stephencdaly:
If a charge is picked up by the expiry sweep has a status that indicates
the gateway knows about it, then for ePDQ and Worldpay payments:

- we get the status of the charge with the gateway and map it to a
ChargeStatus.
- if the mapped status is terminal (CAPTURED, AUTHORISATION REJECTED)
then we don't expire and instead attempt to update the status of the
charge to match the gateway status.
- if the mapped status is not terminal, we attempt to cancel the charge
with the gateway and then expire it.

For other gateways, we do not check with the gateway and instead always
try to cancel with the gateway if the charge has been authorised
successfully.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
@@ -16,6 +16,7 @@ import uk.gov.pay.connector.charge.model.domain.ExpirableChargeStatus;
 import uk.gov.pay.connector.charge.model.domain.ExpirableChargeStatus.AuthorisationStage;
 import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
 import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
+import uk.gov.pay.connector.common.exception.InvalidForceStateTransitionException;
 import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
 import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.PaymentProviders;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -161,38 +171,93 @@ private int expireChargesWithoutGateway(List<ChargeEntity> nonAuthSuccessCharges
...
+                                    kv(GATEWAY_ACCOUNT_ID, chargeEntity.getGatewayAccount().getId()),
+                                    kv(PROVIDER, chargeEntity.getGatewayAccount().getGatewayName()));
+
+                            try {
+                                chargeService.forceTransitionChargeState(chargeEntity, status);
+                                expireCancelFailed.getAndIncrement();
+                            } catch (InvalidForceStateTransitionException e) {
+                                logger.error(format("Cannot expire charge as it is in a terminal state of [%s] with the " +
+                                                "gateway provider. The charge's state could not be transitioned to [%s].",
+                                        status.getValue(), status.getValue()),
```
oswaldquek: I don't think there's a need to log the same information `status.getValue()` twice here. A better log line might be `Cannot transition state from chargeEntity.getStatus() to status.getValue()`?

oswaldquek (changes requested):
Looks good! One minor comment about logging.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
stephencdaly: I pretty much lifted this logging from here https://github.com/alphagov/pay-connector/blob/b73bc6fbf61181cfcd11c2378a38c9484ebf183d/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java#L103 
Should I change it there too?

oswaldquek: I can do that as part of my up and coming PR for PP-6186!

stephencdaly: How about:

> Cannot expire charge as it is in a terminal state of [AUTHORISATION REJECTED] with the gateway provider and it is not possible to transition the charge into this state.

oswaldquek: 👍 I think it'll also be good to log the charge's current state.

stephencdaly: updated

oswaldquek: Sorry I meant 
```
Cannot expire charge as it is in a terminal state of [AUTHORISATION REJECTED] with the gateway provider and it is not possible to transition the charge into this state. Current state: [status.getValue()]
```

oswaldquek (dismissed):

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
stephencdaly: Updated

oswaldquek (approved):

PR #2172, opened 2020-03-03
PP-6217 Fix stringification for Worldpay payment query response

alexbishop1:
When stringifying a Worldpay payment query response, describe it as a “Worldpay query response” rather than a “Worldpay authorisation response” because it was making our logs confusing.


danworth (approved):
Thanks 👍

PR #2175, opened 2020-03-04
PP-6186 Cancel charge if has unknown status with gateway

stephencdaly:
Querying the gateway and not getting a response back is BAU, as this
occurs e.g. when a charge is in AUTHORISATION_3DS_READY state with
Worldpay.

Therefore when this happens, cancel the charge locally without
cancelling with the gateway.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
@@ -73,7 +73,7 @@ public class ChargeCancelService {
         validateChargeStatus(statusFlow, chargeEntity);
         ChargeStatus currentChargeStatus = ChargeStatus.fromString(chargeEntity.getStatus());
         var authorisationStage = ExpirableChargeStatus.of(currentChargeStatus).getAuthorisationStage();
-        
+
         if ((authorisationStage == DURING_AUTHORISATION || authorisationStage == POST_AUTHORISATION)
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java
```diff
@@ -86,40 +86,39 @@ private void doCancel(ChargeEntity chargeEntity, StatusFlow statusFlow) {
...
+                                CancelConflictException.ConflictResult.CHARGE_FORCIBLY_TRANSITIONED);
+                    } else {
+                        cancelChargeWithGatewayCleanup(chargeEntity, statusFlow);
+                    }
+                },
+                () -> {
+                    logger.info(format("Gateway status does not map to any charge status in %s, cancelling without " +
+                                    "cancelling on the gateway", ChargeStatus.class.getCanonicalName()),
+                            kv(PAYMENT_EXTERNAL_ID, chargeEntity.getExternalId()));
+                    nonGatewayCancel(chargeEntity, statusFlow);
```
stephencdaly, line 120: This is the bit that actually changed. The rest is because I modified to use `ifPresentOrElse` and it changed the indentation.

stephencdaly: If you hide whitespace changes it's clearer

Screenshot 2020-03-04 at 10 46 13

oswaldquek (approved):

PR #2176, opened 2020-03-04
PP-6199 - Force state transitions for capture notifications for charges in errored states

nimalank7:
Description:
- Reuses ChargeService.forceTransitionChargeState and placed in catch clause so that it only attempts to force if it has already tried to update using the normal flow
- I'm unsure of the logging messages here. It will initially log at ERROR level that it couldn't be transitioned the normal way then attempt to force capture. If this fails then it will log at ERROR level.

Any thoughts on the exact wording for the error message and the placement of the force transition code?

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
@@ -6,6 +6,7 @@ import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.common.exception.InvalidForceStateTransitionException;
 import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -46,7 +47,18 @@ public void invoke(String gatewayTransactionId, Charge charge, ChargeStatus newS
                         kv(GATEWAY_ACCOUNT_ID, gatewayAccount.getId()),
                         kv(PROVIDER, gatewayAccount.getGatewayName()));
             }
-            return;
+            
+            try {
+                chargeService.forceTransitionChargeState(chargeEntity, newStatus);
+            } catch (InvalidForceStateTransitionException ie) {
+                logger.error(String.format("%s (%s) notification '%s' could not be used to force capture charge: %s",
```
stephencdaly: This error message is stating that it's a capture notification, but it won't necessarily be.

Instead we should log the state we're trying to transition from and to.

stephencdaly: While you're here, I think we could improve the above code so that we check whether the `newStatus` is the same as the current status of the `chargeEntity` before trying to do any transition and skip the notification if it is. We could then remove the check for logging the error: `if (!e.getCurrentState().equals(e.getTargetState())) {`

Otherwise we're going to try to force transition when the statuses are the same which might be a bit confusing.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -46,7 +47,18 @@ public void invoke(String gatewayTransactionId, Charge charge, ChargeStatus newS
...
+                        gatewayAccount.getGatewayName(), gatewayAccount.getId(), gatewayTransactionId, ie.getMessage()),
+                        kv(PAYMENT_EXTERNAL_ID, chargeEntity.getExternalId()),
+                        kv(PROVIDER_PAYMENT_ID, gatewayTransactionId),
+                        kv(GATEWAY_ACCOUNT_ID, gatewayAccount.getId()),
+                        kv(PROVIDER, gatewayAccount.getGatewayName()));
+                return;
+            }
```
oswaldquek: it would be good to clear this code up by moving it to a private method. a try...catch block within ia try...catch block is generally quite nasty to read.

oswaldquek (changes requested):
Looks good! Just a minor comment about code readability.
I also assume there'll be another PR for sending an event to ledger too :)

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceIT.java
```diff
@@ -69,6 +69,21 @@ public void shouldHandleACaptureNotification(ChargeStatus chargeStatus) {
...
+        String response = notifyConnector(transactionId, "1", "9", getCredentials().get(CREDENTIALS_SHA_OUT_PASSPHRASE))
+                .statusCode(200)
+                .extract().body()
+                .asString();
+
+        assertThat(response, is(RESPONSE_EXPECTED_BY_EPDQ));
+
+        assertFrontendChargeStatusIs(chargeId, CAPTURED.getValue());
+    }
+
```
oswaldquek, line 101: could we also have another test where the notification is not a capture notification please (e.g. USER CANCELLED)? in this case the charge's original status should remain the same.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqNotificationResourceIT.java
nimalank7: Done

oswaldquek (approved):
👍

PR #2178, opened 2020-03-05
PP-6189 Allow for normal state transition to terminal state on expiry

stephencdaly:
If when we query the payment gateway before expiry, and it tells us that
the charge maps to a terminal state that we can normally transition to
without "forcing" it, then transition to that state.

This covers that case where 3DS authorisation has been rejected for an
ePDQ payment, but the user was never returned to our pages from the ePQQ
3DS page so we never knew about it until this point.

Also, when we update the status of the charge rather than expiring it,
treat this as an expiry success in the response returned from the
expiry sweep.

Noticed this was something can happen from this error logged to sentry https://pay-sentry.cloudapps.digital/sentry/connector/issues/3873/events/115290/

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
@@ -17,6 +17,7 @@ import uk.gov.pay.connector.charge.model.domain.ExpirableChargeStatus.Authorisat
 import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
 import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
 import uk.gov.pay.connector.common.exception.InvalidForceStateTransitionException;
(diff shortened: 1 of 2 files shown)


oswaldquek:
The reason I'm not sure about this PR is because the AUTHORISATION REJECTED (from the sentry link) is terminal and does not need expiring...

stephencdaly:
The problem is that it's not currently being expired or being updated to AUTHORISATION REJECTED. This PR will mean that it's updated to AUTHORISATION REJECTED and not expired.

stephencdaly:
I decided to do this rather than adding another event to "force" this state because it is already a valid transition, we're just finding out about it later. Maybe we should able to "force" transition to this state too, but in this case it seems wrong to suggest we're "forcing" it into this state.

oswaldquek (approved):
Sorry my bad - i think transitioning normally before forcefully transitioning makes sense.

PR #2182, opened 2020-03-06
PP-6238 Enable prometheus metrics

danworth:
If `PROMETHEUS_ENABLED` is `true` add a `DropwizardExports` to process
the default `MetricsRegistry` into the format required by Prometheus and
add it to the default Prometheus `CollectorRegistry`. To provide an
end-point which Prometheus can scrape, add a `MetricsServlet` exposed on
a configurable path and add serve it via the admin port.

The default settings are to not enable Prometheus.

This is for a spike and will not be moved to prod. Please do not merge and no need to review. Creating a draft PR to add to the jira ticket.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -502,6 +502,22 @@
             <artifactId>dropwizard-db</artifactId>
             <version>${dropwizard.version}</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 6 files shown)


SandorArpa:
Old PR

PR #2183, opened 2020-03-06
zd3934900: Don't throw exception if the response is successful

oswaldquek:
A capture of a charge failed because a 202 was returned. This led to the error

```
Error capturing charge from SQS message [queueMessageId=a4f345d9-c063-4a87-ad35-c31503a031f7] [errorMessage=HTTP 202 Accepted]
```

See the ticket for more details. I'm not actually sure how `errorMessage=HTTP
202 Accepted` came about; following the code path, a 202 should throw a
GatewayErrorException which gets rethrown as a runtime exception in
StripeCaptureHandler (`new RuntimeException("Unrecognised response status
during capture.")`). But it's still worth checking the response is successful,
rather than just a 200 OK.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
@@ -24,6 +24,8 @@ import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
 import static java.util.Collections.emptyList;
+import static javax.ws.rs.core.Response.Status.Family.SUCCESSFUL;
+import static javax.ws.rs.core.Response.Status.Family.familyOf;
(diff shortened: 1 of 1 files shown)


sfount (approved):
Looks reasonable to me -- note that this also will return a successful response given `201 Created`, `204 No content`, `205 Reset content` and `206 Partial content`. 

The only requirement on this ticket is `202 Accepted` for the idempotent request response, to avoid any future side effects just checking for this should also resolve the issue.

oswaldquek:
I think this is ok as the GatewayClient is meant to be a generic thing. We should allow all responses in the "successful 2xx" category.

PR #2190, opened 2020-03-09
PP-6194 Add DAO method for getting charges with provider and status

stephencdaly:
Want to be able to get all epdq charges in statuses 'AUTHORISATION ERROR',
'AUTHORISATION TIMEOUT' and 'AUTHORISATION UNEXPECTED ERROR' in order to
clean them up with the gateway.

Add this DAO method so we can do that.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
@@ -240,4 +240,12 @@ public class ChargeDao extends JpaDao<ChargeEntity> {
                 .setParameter(1, id)
                 .executeUpdate();
     }
+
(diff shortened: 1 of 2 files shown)


oswaldquek (approved):
With this PR I assume we're going to go with finding all epdq charges in the states of interest without regard to the date range.

stephencdaly:
Yes, I ran the query on production and it was really quick to run

oswaldquek:
It's not the being quick to query the charges, it's more like whether we have any API limits when querying epdq.

stephencdaly:
Ahh hmm good point. Do you think we want a limit on the number of charges we process at a time? Or should we try it as is and monitor if we get any issues?
It's not something that we've considered for expiry, where we already query ePDQ.

PR #2194, opened 2020-03-10
PP-6201 - Force capture notifications for expunged charges

nimalank7:
Description:
- Corresponds to PP-6201-Add CAPTURE_NOTIFICATION_FOR_EXPUNGED_CHARGE event to Ledger
- Since the charge doesn't exist in the database (as it is expunged) when we process capture notification we send it directly to the Event SQS queue without persisting

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -679,7 +679,7 @@ public class ChargeService {
     public Optional<ChargeEntity> findByProviderAndTransactionId(String paymentGatewayName, String transactionId) {
         return chargeDao.findByProviderAndTransactionId(paymentGatewayName, transactionId);
     }
-
+    
     @Transactional
     public ChargeEntity markChargeAsEligibleForCapture(String externalId) {
         return chargeDao.findByExternalId(externalId).map(charge -> {
--- a/src/main/java/uk/gov/pay/connector/events/EventService.java
+++ b/src/main/java/uk/gov/pay/connector/events/EventService.java
@@ -21,7 +21,15 @@ public class EventService {
         this.eventQueue = eventQueue;
         this.emittedEventDao = emittedEventDao;
     }
-
+    
+    public void emitEvent(Event event) {
+        try {
+            eventQueue.emitEvent(event);
+        } catch (QueueException e) {
+            logger.error("Failed to emit event {} due to {} [externalId={}]", event.getEventType(), e.getMessage(), event.getResourceExternalId());
+        }
+    }
+    
     public void emitAndRecordEvent(Event event, ZonedDateTime doNotRetryEmitUntilDate) {
         try {
             eventQueue.emitEvent(event);
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
@@ -13,12 +13,14 @@ import uk.gov.pay.connector.gateway.processor.ChargeNotificationProcessor;
 import uk.gov.pay.connector.gateway.processor.RefundNotificationProcessor;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
+import uk.gov.pay.connector.queue.QueueException;
 import uk.gov.pay.connector.refund.model.domain.RefundStatus;
 
 import java.util.List;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import static java.lang.String.format;
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isEmpty;
@@ -114,6 +116,11 @@ public class EpdqNotificationService {
 
         if (newChargeStatus.isPresent()) {
             if(charge.isHistoric()){
+                if (CAPTURED.getValue().equals(newChargeStatus.get().getValue())) {
+                    chargeNotificationProcessor.processCaptureNotificationForExpungedCharge(gatewayAccountEntity, notification.getTransactionId(), charge, newChargeStatus.get(), null);
+                    return;
+                }
+                
                 logger.error("{} notification {} could not be processed as charge [{}] has been expunged from connector {} {}",
                         PAYMENT_GATEWAY_NAME, notification,
                         charge.getExternalId(),
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
@@ -8,11 +8,16 @@ import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.common.exception.InvalidForceStateTransitionException;
 import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
+import uk.gov.pay.connector.events.EventService;
+import uk.gov.pay.connector.events.model.Event;
+import uk.gov.pay.connector.events.model.charge.CaptureConfirmedByGatewayNotification;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 
 import javax.inject.Inject;
 import java.time.ZonedDateTime;
+import java.util.List;
 
+import static java.lang.String.format;
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
 import static uk.gov.pay.logging.LoggingKeys.PAYMENT_EXTERNAL_ID;
@@ -22,11 +27,13 @@ import static uk.gov.pay.logging.LoggingKeys.PROVIDER_PAYMENT_ID;
 public class ChargeNotificationProcessor {
 
     private ChargeService chargeService;
+    private EventService eventService;
     private Logger logger = LoggerFactory.getLogger(getClass());
 
     @Inject
-    ChargeNotificationProcessor(ChargeService chargeService) {
+    ChargeNotificationProcessor(ChargeService chargeService, EventService eventService) {
         this.chargeService = chargeService;
+        this.eventService = eventService;
     }
 
     public void invoke(String gatewayTransactionId, Charge charge, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
@@ -85,5 +92,22 @@ public class ChargeNotificationProcessor {
                     kv(PROVIDER, gatewayAccount.getGatewayName()));
             return false;
         }
+    }
+    
+    public void processCaptureNotificationForExpungedCharge(GatewayAccountEntity gatewayAccount, String gatewayTransactionId, Charge charge, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
+        logger.info(String.format("Received capture notification for charge that was already expunged from Connector"),
+                kv(PAYMENT_EXTERNAL_ID, charge.getExternalId()),
+                kv(PROVIDER_PAYMENT_ID, gatewayTransactionId),
+                kv(GATEWAY_ACCOUNT_ID, gatewayAccount.getId()),
+                kv(PROVIDER, gatewayAccount.getGatewayName()));
+        
+        Event event = new CaptureConfirmedByGatewayNotification(charge.getExternalId(), gatewayEventDate);
+
+        logger.info(format("Force state transition from [%s] to [%s]", charge.getStatus(),  newStatus.getValue()),
+                List.of(kv(PAYMENT_EXTERNAL_ID, charge.getExternalId()),
+                        kv(GATEWAY_ACCOUNT_ID, gatewayAccount.getId()),
+                        kv(PROVIDER, gatewayAccount.getGatewayName())));
+        
+         eventService.emitEvent(event);
     }
 }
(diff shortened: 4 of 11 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -114,6 +115,11 @@ public void handleNotificationFor(String payload) {
 
         if (newChargeStatus.isPresent()) {
             if(charge.isHistoric()){
+                if ("CAPTURED".equals(newChargeStatus.get().getValue())) {
```
stephencdaly: Use the ChargeStatus enum value rather than a hardcoded string.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -86,4 +86,15 @@ private boolean forceTransitionChargeState(GatewayAccountEntity gatewayAccount,
             return false;
         }
     }
+    
+    public void processForcedCaptureForExpungedCharge(GatewayAccountEntity gatewayAccount, String gatewayTransactionId, Charge charge, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
+        logger.info(String.format("Forcibly changing charge status for externalId [%s] [%s]->[%s]",
+                charge.getExternalId(), charge.getStatus(), newStatus.getValue()),
+                kv(PAYMENT_EXTERNAL_ID, charge.getExternalId()),
```
stephencdaly, line 99: can use `ChargeEntity.getStructuredLoggingArgs` to save lines of code to log all these keys.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -86,4 +86,15 @@ private boolean forceTransitionChargeState(GatewayAccountEntity gatewayAccount,
             return false;
         }
     }
+    
+    public void processForcedCaptureForExpungedCharge(GatewayAccountEntity gatewayAccount, String gatewayTransactionId, Charge charge, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
+        logger.info(String.format("Forcibly changing charge status for externalId [%s] [%s]->[%s]",
```
stephencdaly: Could the log line better describe what's going on? We don't actually know whether we're forcing a status change, all we know is we've got a capture notification for a charge that's been expunged.
Something like:
"Received capture notification for charge that was already expunged from connector"

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -86,4 +86,15 @@ private boolean forceTransitionChargeState(GatewayAccountEntity gatewayAccount,
             return false;
         }
     }
+    
+    public void processForcedCaptureForExpungedCharge(GatewayAccountEntity gatewayAccount, String gatewayTransactionId, Charge charge, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
```
stephencdaly: could the method be renamed to `processCaptureNotificationForExpungedCharge`

src/main/java/uk/gov/pay/connector/events/model/charge/ExpungedChargeStatusCorrectedToCapturedToMatchGatewayStatus.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.pay.connector.events.model.charge;
+
+import java.time.ZonedDateTime;
+
+public class ExpungedChargeStatusCorrectedToCapturedToMatchGatewayStatus extends PaymentEventWithoutDetails {
```
stephencdaly: Hmm do we actually know that we've corrected a status though? For all we know the charge is already in a captured state in ledger.
I think it should be something along the lines of "CaptureNotificationReceivedForExpungedCharge" unless someone else had something to say about this.

src/main/java/uk/gov/pay/connector/events/EventService.java
```diff
@@ -21,6 +21,10 @@ public EventService(EventQueue eventQueue, EmittedEventDao emittedEventDao) {
         this.eventQueue = eventQueue;
         this.emittedEventDao = emittedEventDao;
     }
+    
+    public void emitEvent(Event event) throws QueueException {
```
stephencdaly: why do we need this new method?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -679,6 +681,21 @@ public ChargeEntity transitionChargeState(String chargeId, ChargeStatus targetCh
     public Optional<ChargeEntity> findByProviderAndTransactionId(String paymentGatewayName, String transactionId) {
         return chargeDao.findByProviderAndTransactionId(paymentGatewayName, transactionId);
     }
+    
+    public void forceCaptureForExpungedCharge(Charge charge, GatewayAccountEntity gatewayAccountEntity, ChargeStatus targetChargeState, ZonedDateTime gatewayEventDate) {
```
stephencdaly: I don't think we need a method in the ChargeService for this, could we just directly emit the event from the NotificationProcessor?

I also don't think we're handling the QueueException well here, I think we should just use the `emitAndRecordEvent` method. Is there a reason not to?

stephencdaly (changes requested):

src/main/java/uk/gov/pay/connector/events/model/charge/ExpungedChargeStatusCorrectedToCapturedToMatchGatewayStatus.java
stephencdaly: or CAPTURE_CONFIRMED_BY_GATEWAY_NOTIFICATION as Steven suggested on the other PR

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: emitAndRecordEvent persists the ChargeEventEntity into the database which I don't think we want to do as we only want to send a capture notification to Ledger.

src/main/java/uk/gov/pay/connector/events/EventService.java
nimalank7: The other methods persist the recorded event to the database which I don't think we want to do as the charge is historic.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/events/model/charge/ExpungedChargeStatusCorrectedToCapturedToMatchGatewayStatus.java
oswaldquek: or CaptureNOtificationForExpungedCharge as per the story in Jira.

src/main/java/uk/gov/pay/connector/events/EventService.java
oswaldquek: If the event was successfully put on the Queue we don't want to record it; however if it was not successful we want to to record it so it will be picked up to be processed again. So I think the code here should look like 
```
    public void emitEvent(Event event, ZonedDateTime doNotRetryEmitUntilDate) {
        try {
            eventQueue.emitEvent(event);
        } catch (QueueException e) {
            emittedEventDao.recordEmission(event.getResourceType(), event.getResourceExternalId(),
                    event.getEventType(), event.getTimestamp(), doNotRetryEmitUntilDate);
            logger.error("Failed to emit event {} due to {} [externalId={}]",
                    event.getEventType(), e.getMessage(), event.getResourceExternalId());
        }
    }
```

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
oswaldquek: yeah I think we can just call `EventService.emitEvent` after this line https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java#L121.
adding this  `forceCaptureForExpungedCharge` method seems excessive.

src/main/java/uk/gov/pay/connector/events/EventService.java
oswaldquek: so after speaking to Steven F, the original 
```
public void emitEvent(Event event) throws QueueException {
  eventQueue.emitEvent(event);
}
```
looks fine. In theory if we return a 500 in response to `/v1/api/notifications/epdq` epdq should retry the notification. I will check with them.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
nimalank7: Sure but I'd have to recreate the ChargeEntity from the Charge object to do that which would generate more lines of code to do that.

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/events/model/charge/ExpungedChargeStatusCorrectedToCapturedToMatchGatewayStatus.java
nimalank7: I've changed it to CAPTURE_CONFIRMED_BY_GATEWAY_NOTIFICATION

src/main/java/uk/gov/pay/connector/events/EventService.java
nimalank7: I've kept it to merely emitEvent as above

oswaldquek (approved):

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
stephencdaly: oh yeh my bad

src/main/java/uk/gov/pay/connector/events/EventService.java
```diff
@@ -21,7 +21,15 @@ public EventService(EventQueue eventQueue, EmittedEventDao emittedEventDao) {
         this.eventQueue = eventQueue;
         this.emittedEventDao = emittedEventDao;
     }
-
+    
+    public void emitEvent(Event event) {
```
stephencdaly, line 25: Could it be worth adding a comment here about what situation this method should be called in as opposed to `emitAndRecord` event - i.e. when the charge has already been expunged? It might save someone doing something wrong in the future.

stephencdaly (approved):
Just a comment about adding a comment

src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
```diff
@@ -86,4 +93,21 @@ private boolean forceTransitionChargeState(GatewayAccountEntity gatewayAccount,
...
+    public void processCaptureNotificationForExpungedCharge(GatewayAccountEntity gatewayAccount, String gatewayTransactionId, Charge charge, ChargeStatus newStatus, ZonedDateTime gatewayEventDate) {
+        logger.info(String.format("Received capture notification for charge that was already expunged from Connector"),
+                kv(PAYMENT_EXTERNAL_ID, charge.getExternalId()),
+                kv(PROVIDER_PAYMENT_ID, gatewayTransactionId),
+                kv(GATEWAY_ACCOUNT_ID, gatewayAccount.getId()),
+                kv(PROVIDER, gatewayAccount.getGatewayName()));
+        
+        Event event = new CaptureConfirmedByGatewayNotification(charge.getExternalId(), gatewayEventDate);
+
+        logger.info(format("Force state transition from [%s] to [%s]", charge.getStatus(),  newStatus.getValue()),
```
stephencdaly, line 106: probably don't need this additional line here, we're already logging a line above that gives us the information we need

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -114,6 +116,11 @@ public void handleNotificationFor(String payload) {
 
         if (newChargeStatus.isPresent()) {
             if(charge.isHistoric()){
+                if (CAPTURED.getValue().equals(newChargeStatus.get().getValue())) {
```
stephencdaly, line 119: don't need the `.getValue` in both cases, can just be 
```if (CAPTURED.equals(newChargeStatus.get())) {```

File: src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java [hunk 11]
```diff
@@ -114,6 +116,11 @@ public void handleNotificationFor(String payload) {
 
         if (newChargeStatus.isPresent()) {
```
stephencdaly, line 117: I think we generally prefer to use 
```.ifPresentOrElse(status -> {```

to avoid having to call `.get` on the Optional, as it's the more modern java-y way to do it.

src/main/java/uk/gov/pay/connector/events/EventService.java
oswaldquek: I'll do this. I also want to write a comment on why we catch the QueueException and don't do anything further. I did think about this but didn't want to delay the PR any further as it's been open for a while.

PR #2196, opened 2020-03-10
PP-6194 Add service for ePDQ gateway cleanup

stephencdaly:
Add service for the recurring job to cleanup ePDQ charges that are in "AUTHORISATION ERROR", "AUTHORISATION UNEXPECTED ERROR" or "AUTHORISATION TIMEOUT" statuses.

This job is required as when we enter into one of these statuses we are unsure whether the charge was actually authorised with the gateway. If it was authorised, we need to cancel it on the gateway to ensure that funds are released in the paying users' bank account.

The job will move the charge into one of three statuses when it has successfully handled it:
"AUTHORISATION ERROR CANCELLED" - the charge was authorised on the gateway but has now been cancelled.
"AUTHORISATION ERROR REJECTED" - the authorisation was rejected on the gateway and no action needed to be taken to clean up.
"AUTHORISATION ERROR CHARGE MISSING" - the charge was not found on the gateway, most likely because the error was before the gateway processed the authorisation.

Map ePDQ status code "0" to "AUTHORISATION ERROR" status when querying the charge status, as this represents an authorisation response that we would have handled by moving the charge into this state.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/EpdqAuthorisationErrorGatewayCleanupService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/EpdqAuthorisationErrorGatewayCleanupService.java
@@ -33,13 +33,13 @@ public class EpdqAuthorisationErrorGatewayCleanupService {
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
-    public static final String CLEANUP_SUCCESS = "cleanup-success";
-    public static final String CLEANUP_FAILED = "cleanup-failed";
+    static final String CLEANUP_SUCCESS = "cleanup-success";
+    static final String CLEANUP_FAILED = "cleanup-failed";
 
-    private ChargeDao chargeDao;
-    private ChargeService chargeService;
-    private QueryService queryService;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/charge/service/EpdqAuthorisationErrorGatewayCleanupService.java
```diff
@@ -0,0 +1,151 @@
...
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    public static final String CLEANUP_SUCCESS = "cleanup-success";
+    public static final String CLEANUP_FAILED = "cleanup-failed";
+
+    private ChargeDao chargeDao;
+    private ChargeService chargeService;
+    private QueryService queryService;
+    private PaymentProviders providers;
```
oswaldquek: be nice to make these `final` too

oswaldquek (dismissed):
Minor comment but it doesn't stop this PR from being approved.

oswaldquek (approved):

PR #2203, opened 2020-03-12
PP-6224: Don't expunge epdq charges in select states

oswaldquek:
Don't expunge if in AUTHORISATION ERROR, AUTHORISATION TIMEOUT and
AUTHORISATION UNEXPECTED ERROR states.
The reason we're doing this is because we don't trust that epdq charges in
these states match the states at the epdq end. On the epdq side, these charges
can be:
  - missing
  - captured
  - authorised
  - rejected

By making these not expungable, the EpdqAuthorisationErrorGatewayCleanupService
will be able to pick these charges up and clean them up.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeStatus.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeStatus.java
@@ -78,7 +78,7 @@ public enum ChargeStatus implements Status {
         return value;
     }
 
-    public boolean  isExpungeable() {
+    public boolean isExpungeable() {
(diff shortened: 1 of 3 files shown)


oswaldquek:
DO NOT MERGE YET

sfount (approved):
The reasoning for this is well explained in the commit message, hopefully we can optimise these two jobs to remove this custom exception in the future but we'll monitor it initially to see how this works

PR #2208, opened 2020-03-13
DRAFT Validate External Metadata on ChargeEntity

danworth:
Validate the `ExternalMetadata` when creating a new `ChargeEntity`
to prevent invalid external meta data being created from within
Connector (e.g. not via an api request). This can pose a problem for
`Selfservice` when it attempts to display a charge with invalid meta
data.

Previous attempt used the `@Valid` annotation but this prevented existing
charges with invalid metadata being updated (the validation was called upon
any JPA update and persist operation). Furthermore the error message reported was
unclear that the exception was being thrown by JPA and not the API
layer.

This approach permits updating existing charges which have metadata
considered to be invalid by the latest standards of `ValidExternalMetadata`
which is important since this standard can evolve over time. It
prevents new `ChargeEntities` from being created if the external
metadata does not conform to the latest `ValidExternalMetadata`
standard. The exception message should provide a clear indication what
is wrong.

## WHAT YOU DID
I was curious as to a solution and got carried away. Raising as a draft in case it's useful for you. Can likely be tidied up, please be gentle I've been writing `.yml` for the last 6 months...

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -44,13 +44,18 @@ import javax.persistence.OneToOne;
 import javax.persistence.OrderBy;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
-import javax.validation.Valid;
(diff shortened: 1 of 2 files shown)


stephencdaly:
This is very old, closing

PR #2209, opened 2020-03-13
PP-6259 Rename 3DS-related model objects

alexbishop1:
Despite their similar names, `Auth3dsDetailsEntity` and `Auth3dsDetails` do not represent the same thing. The former is data needed when 3D Secure is required (e.g. a `paRequest`), while the latter is the result of 3D Secure (e.g. a `paResponse`).

Rename `Auth3dsDetailsEntity` to `Auth3dsRequiredEntity` and `Auth3dsDetails` to `Auth3dsResult` (including renaming the existing `Auth3dsResult` — an inner enum of `Auth3dsDetails` — to `Auth3dsResultOutcome`).

with @SandorArpa


SandorArpa (approved):

PR #2210, opened 2020-03-13
PP-6259 Rename GatewayParamsFor3ds to Gateway3dsRequiredParams

alexbishop1:
Rename `GatewayParamsFor3ds` to `Gateway3dsRequiredParams` to make it clear that the parameters relate to what the gateway tells us when 3D Secure is required and not anything to do with the 3D Secure result.

with @SandorArpa


SandorArpa (approved):

SandorArpa (approved):

PR #2211, opened 2020-03-13
PP-6259 Create specific types for session IDs

alexbishop1:
We have a couple of different things called session identifiers or IDs:

* A thing that a payment provider sends to us and requires us to store and send back to them. This includes Worldpay’s `machine` cookie, which is used as part of 3D Secure.
* A unique identifier Worldpay requires us to generate and send to them when authorising a payment.

To avoid confusion, introduce specific types for these two things, rather than having them both be strings with confusingly similar names.

with @SandorArpa


SandorArpa (approved):

PR #2215, opened 2020-03-16
PP-6254: Add Carbon Relay metrics config

rjbaker:
Set the `METRICS_HOST` and `METRICS_PORT` from the app-catalog to send
metrics to the carbon-relay app.


test.sh
```diff
@@ -0,0 +1,17 @@
+
+
+
+latest_release_number=dfdfd
```
danworth: is this supposed to be in this commit?

danworth (approved):
Looks good 👍

PR #2217, opened 2020-03-16
PP-6259 Stop using Optional params for some authorisation methods

alexbishop1:
Having `Optional`s as method parameters is generally frowned upon, so replace some of them in authorisation-related methods with just regular might-be-null objects.

with @SandorArpa


SandorArpa (approved):

PR #2219, opened 2020-03-16
PP-6259 Worldpay: store 3DS required data if asked to do 3DS again

alexbishop1:
If attempting to authorise a 3D Secure result with Worldpay leads to Worldpay sending back a response that contains details required to complete 3D Secure again (e.g. a `paRequest`), store these on the charge.

Note that this won’t actually change any behaviour because a transition from `AUTHORISATION 3DS READY` to `AUTHORISATION 3DS REQUIRED` is still considered invalid and we’re also not yet indicating to frontend that the operation was successful.

with @SandorArpa


SandorArpa (approved):

PR #2221, opened 2020-03-17
PP-6194 Allow force transition to auth error and rejected states

stephencdaly:
The charge could be in a terminal state corresponding to AUTHORISATION
ERROR or AUTHORISATION REJECTED with the gateway. Currently when we come
across charges in these states as part of the expiry job we don't expire
them and leave them to be picked up continuously by the job.

Instead we should transition them to the terminal state they are in on
the gateway.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
@@ -34,6 +34,8 @@ import uk.gov.pay.connector.events.model.charge.PaymentExpired;
 import uk.gov.pay.connector.events.model.charge.PaymentNotificationCreated;
 import uk.gov.pay.connector.events.model.charge.PaymentStarted;
 import uk.gov.pay.connector.events.model.charge.ServiceApprovedForCapture;
+import uk.gov.pay.connector.events.model.charge.StatusCorrectedToAuthorisationErrorToMatchGatewayStatus;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/common/model/domain/PaymentGatewayStateTransitions.java
```diff
@@ -141,6 +146,7 @@ private PaymentGatewayStateTransitions() {
         graph.putEdgeValue(AUTHORISATION_3DS_REQUIRED, AUTHORISATION_SUCCESS, ModelledEvent.of(AuthorisationSucceeded.class));
         graph.putEdgeValue(AUTHORISATION_3DS_REQUIRED, AUTHORISATION_REJECTED, ModelledEvent.of(AuthorisationRejected.class));
         graph.putEdgeValue(AUTHORISATION_3DS_REQUIRED, AUTHORISATION_CANCELLED, ModelledEvent.of(AuthorisationCancelled.class));
+        graph.putEdgeValue(AUTHORISATION_3DS_REQUIRED, AUTHORISATION_ERROR, ModelledEvent.of(GatewayErrorDuringAuthorisation.class));
```
oswaldquek: Is this part of this PR?

stephencdaly: oh good spot, this was what I was originally going to do but decided against it, will revert that change

oswaldquek (approved):

PR #2224, opened 2020-03-17
PP-6259 Return charge status to frontend when 3DS auth declined

alexbishop1:
When a call to `/v1/frontend/charges/{chargeId}/3ds` results in the payment’s authorisation being rejected, return a 400 with a body containing the response, rather than just a generic error message.

Before:

```json
{
  "error_identifier": "GENERIC",
  "message": "This transaction was declined."
}
```

After:

```json
{
  "status": "AUTHORISATION REJECTED"
}
```

This makes 400 responses consistent with 200 responses. Only frontend uses this endpoint.

with @SandorArpa


SandorArpa (approved):

PR #2226, opened 2020-03-17
PP-6194 Fix charge status transition not persisted for expiry job

stephencdaly:
When updating the charge status to match the gateway status, the change was not being persisted because the ChargeEntity being modified was not part of the current transaction.

Ensure we are loading the ChargeEntity into the transaction and add a test for this.

Improve the tests for the expiry service to ensure we are persisting the status change.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
@@ -210,7 +210,7 @@ public class ChargeExpiryService {
                             // first try to transition to the terminal state gracefully if allowed, otherwise force the
                             // transition
                             try {
-                                chargeService.transitionChargeState(chargeEntity, status);
+                                chargeService.transitionChargeState(chargeEntity.getExternalId(), status);
                                 expireCancelled.getAndIncrement();
                             } catch (InvalidStateTransitionException e) {
                                 if (forceTransitionChargeState(chargeEntity, status)) {
@@ -247,7 +247,7 @@ public class ChargeExpiryService {
     
     private boolean forceTransitionChargeState(ChargeEntity chargeEntity, ChargeStatus status) {
         try {
-            chargeService.forceTransitionChargeState(chargeEntity, status);
+            chargeService.forceTransitionChargeState(chargeEntity.getExternalId(), status);
             return true;
         } catch (InvalidForceStateTransitionException e) {
             logger.error(format("Cannot expire charge as it is in a terminal state of [%s] with " +
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -641,14 +641,21 @@ public class ChargeService {
     }
 
     @Transactional
-    public ChargeEntity transitionChargeState(String chargeId, ChargeStatus targetChargeState) {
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -247,7 +247,7 @@ private int expireChargesWithoutGateway(List<ChargeEntity> nonAuthSuccessCharges
     
     private boolean forceTransitionChargeState(ChargeEntity chargeEntity, ChargeStatus status) {
         try {
-            chargeService.forceTransitionChargeState(chargeEntity, status);
+            chargeService.forceTransitionChargeState(chargeEntity.getExternalId(), status);
```
oswaldquek, line 250: is there a reason why we can't use the `public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus targetChargeState)` method here?

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -210,7 +210,7 @@ private int expireChargesWithoutGateway(List<ChargeEntity> nonAuthSuccessCharges
                             // first try to transition to the terminal state gracefully if allowed, otherwise force the
                             // transition
                             try {
-                                chargeService.transitionChargeState(chargeEntity, status);
+                                chargeService.transitionChargeState(chargeEntity.getExternalId(), status);
```
oswaldquek, line 213: I think there needs to be a test (or modify an existing one) for this line as `ChargeExpiryResourceIT.shouldExpireChargesAndHaveTheCorrectEvents` and `ChargeExpiryResourceEpdqIT.shouldUpdateChargeStatusToMatchTerminalStateOnGateway ` pass without this change.

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
stephencdaly: Good point, added a new test

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
stephencdaly: Because the chargeEntity won't be associated with the transaction that's opened by that method, so need to load it into the entity manager (in this case, by querying it again).

oswaldquek: ah, so even if `public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus targetChargeState)` is annotated with `@Transactional` it won't work?

oswaldquek (approved):

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
stephencdaly: yeh because we retrieved the ChargeEntity as part of a query outside the transaction so it's not registered with the entity manager

oswaldquek: I wonder what the point of having the `@Transactional` on that method is then! Be good to delete it.

PR #2227, opened 2020-03-18
PP-6259 Allow users to retry 3DS again if the gateway asks

alexbishop1:
When we authorise the 3D Secure result, if the gateway response says to do 3D Secure again, transition to payment status to `AUTHORISATION 3DS REQUIRED` and send this in the response to frontend, which will make frontend show the 3DS page again. However, only do this if this means the user will be able to attempt 3DS at most three times in total (measured by counting how many times an `AUTHORISATION 3DS REQUIRED` event appears in the charge events for the payment); otherwise set to payment status to `AUTHORISATION REJECTED`.

While this change is not specific to one payment gateway, in reality we only allow Worldpay to ask us to do 3DS again (due to the way gateway responses to 3DS result authorisations are handled). If there is a desire for another gateway to exhibit the same behaviour, we will have to update how we determine the new status from the response and also replace the data required to do 3DS (which is stored on the charge entity) as it may be different for the second and subsequent attempts.

An important part of this change is that it allows a state transition from `AUTHORISATION 3DS READY` to `AUTHORISATION 3DS REQUIRED`, though as described above, this can only happen in limited circumstances.

The limit of three attempts at 3DS is in place to stop the user repeatedly attempting 3DS again and again. It’s controlled by the `MAXIMUM_NO_USER_3DS_ATTEMPTS` environment variable and defaults to 3 if not explicitly set.

with @SandorArpa

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import io.dropwizard.Configuration;
 import io.dropwizard.client.JerseyClientConfiguration;
 import io.dropwizard.db.DataSourceFactory;
+import uk.gov.pay.connector.app.config.Authorisation3dsConfig;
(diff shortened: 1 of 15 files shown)


SandorArpa (approved):

PR #2232, opened 2020-03-20
PP-6302 Add language to PAYMENT_NOTIFICATION_CREATED event details

kbottla:
## WHAT YOU DID
- Payment notification end point doesn't receive `language` in the request body, but connector
DB column defaults to 'en' and parity check fails as language is not included in details sent to Ledger.
Added language to payment notification event details.


alexbishop1 (changes requested):
Don’t we also need to update `equals(…)` and `hashCode()`?

kbottla:
Thanks @alexbishop1, overlooked those details. Updated!

alexbishop1 (approved):

PR #2235, opened 2020-03-24
PP-6259 Store machine cookie when Worldpay asks us to do 3DS again

alexbishop1:
Store the `machine` cookie sent by Worldpay with the response an attempt to authorise a 3D Secure result, if there is one. This means that if Worldpay asks us to do 3DS again, our attempt to authorise this subsequent 3DS result will use the new cookie and not the old one, which won’t work.

with @SandorArpa


SandorArpa (approved):

PR #2237, opened 2020-03-25
PP-6265 Handle Payout Notifications

rorymalcolm:
- Connector now handles `payout.created` notifications from Stripe,
parsing them and outputting logs that show information about the event.

- It also includes a test that ensures that the payout event is
correctly deserialised.

- This will be adapted later to trigger more meaningful actions when
such an event is received.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
@@ -22,6 +22,7 @@ import javax.ws.rs.WebApplicationException;
 import java.util.List;
 import java.util.Optional;
 
+import static java.lang.String.format;
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static org.apache.commons.lang3.StringUtils.isBlank;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -94,13 +100,19 @@ public void handleNotificationFor(String payload, String signatureHeader) {
...
+            processPayoutNotification(notification);
         }
     }
 
     private boolean isAnAccountUpdatedNotification(StripeNotification notification) {
         return StripeNotificationType.byType(notification.getType()) == ACCOUNT_UPDATED;
     }
 
+    private void processPayoutNotification(StripeNotification notification) {
+        logger.info("{} payout created notification with id [{}] and body [{}] was received", PAYMENT_GATEWAY_NAME, notification.getId(), notification.getObject());
```
kbottla, line 113: As discussed, logging `notification.object` will be temporary as all info need not be logged all the time

kbottla, line 113: We could further serialise the notification object to specific payout type and ensure it is serialisable. But can be in separate PR.

kbottla (approved):

PR #2245, opened 2020-03-27
BAU Stripe refund handler sets transfer id for gateway transaction

sfount:
We currently set the ID of the refund against the platform for the
refund `gateway_transaction_id`. The actual refund relative to the
Connect account is the transfer from the Connect account to the
platform.

Store this ID so we can use to relate refund transactions and GOV.UK Pay
refunds during reconcilation.

See `PP-6266`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
@@ -54,10 +54,10 @@ public class StripeRefundHandler {
             } else {
                 stripeChargeId = request.getTransactionId();
             }
-            StripeRefund refundResponse = refundCharge(request, stripeChargeId);
-            transferFromConnectAccount(request, stripeChargeId);
+            refundCharge(request, stripeChargeId);
+            StripeTransferResponse transferResponse = transferFromConnectAccount(request, stripeChargeId);
 
-            return fromBaseRefundResponse(StripeRefundResponse.of(refundResponse.getId()), GatewayRefundResponse.RefundState.COMPLETE);
+            return fromBaseRefundResponse(StripeRefundResponse.of(transferResponse.getId()), GatewayRefundResponse.RefundState.COMPLETE);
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeRefundHandler.java
```diff
@@ -54,10 +54,10 @@ public GatewayRefundResponse refund(RefundGatewayRequest request) {
             } else {
                 stripeChargeId = request.getTransactionId();
             }
-            StripeRefund refundResponse = refundCharge(request, stripeChargeId);
-            transferFromConnectAccount(request, stripeChargeId);
+            refundCharge(request, stripeChargeId);
+            StripeTransferResponse transferResponse = transferFromConnectAccount(request, stripeChargeId);
 
-            return fromBaseRefundResponse(StripeRefundResponse.of(refundResponse.getId()), GatewayRefundResponse.RefundState.COMPLETE);
+            return fromBaseRefundResponse(StripeRefundResponse.of(transferResponse.getId()), GatewayRefundResponse.RefundState.COMPLETE);
```
kbottla, line 60: _Just a thought -_ Don't we lose the ability to identify refunds in Stripe without refund ID. We may go to payment view and match by date though but will lose search by refund ID functionality

Do we need a note somewhere (manual) that the past refunds will have a different ID ? 

Also not quite clear of the role of transfer ID in payout reconciliation. But up for a chat later

sfount: These are good questions; the transfer ID referred to here is the transfer between the connect account and the platform to facilitate the refund, it's not related to the payment. The ID should be just as searchable on Stripe's end and will still have the same group ID all the other transactions. 

A note in the manual if it does change would definitely be useful, refund gateway transaction IDs were only actually added when we did the work on Stripe fees so the value in this field has changed a few times since the integration.

sfount:
Superseded by https://github.com/alphagov/pay-connector/pull/2329

PR #2268, opened 2020-04-14
PP-6267 Add Payout reconcile process config

kbottla:
## WHAT
- Added config required for processing payout messages received from payout-reconcile SQS queue


SandorArpa (approved):
LOKTM

PR #2275, opened 2020-04-15
PP-6267 Added PayoutReconciliation background job

kbottla:
## WHAT 
- Added background job `PayoutReconcileMessageReceiver` that is scheduled at fixed internal, and which process payout messages from SQS queue.
- Sets `payoutReconcileQueueEnabled` to false to disable background processing until payout reconcile SQS queue is available and also payout reconciliation algorithm added
- Added `PayoutReconcileProcess` to receive & remove messages from queue. Payout reconciliation algorithm defined in PP-6266 to be added in future

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/managed/PayoutReconcileMessageReceiver.java
+++ b/src/main/java/uk/gov/pay/connector/queue/managed/PayoutReconcileMessageReceiver.java
@@ -1,0 +1,67 @@
+package uk.gov.pay.connector.queue.managed;
+
+import io.dropwizard.lifecycle.Managed;
+import io.dropwizard.setup.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.payout.PayoutReconcileProcess;
+
+import javax.inject.Inject;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+public class PayoutReconcileMessageReceiver implements Managed {
+
+    private static final String PAYOUT_RECONCILE_MESSAGE_RECEIVER_THREAD_NAME = "sqs-message-payoutReconcileMessageReceiver";
+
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/queue/managed/PayoutReconcileMessageReceiver.java
```diff
@@ -0,0 +1,67 @@
...
+                .threads(queueScheduleNumberOfThreads)
+                .build();
+
+        queueSchedulerThreadDelayInSeconds = connectorConfiguration.getPayoutReconcileProcessConfig().getQueueSchedulerThreadDelayInSeconds();
+        payoutReconcileQueueEnabled = connectorConfiguration.getPayoutReconcileProcessConfig().getPayoutReconcileQueueEnabled();
+    }
+
+    @Override
+    public void start() {
+        if (payoutReconcileQueueEnabled) {
```
sfount, line 45: Looks good for during development, we could make a note in upcoming stories to remove this as part of the AC.

src/main/java/uk/gov/pay/connector/queue/managed/PayoutReconcileMessageReceiver.java
```diff
@@ -0,0 +1,67 @@
+package uk.gov.pay.connector.queue.managed;
```
sfount, line 1: Now that this is in the `queue.managed` package, it looks like it would be good to rename the `QueueMessageReceiver` as differentiating between that and the `PayoutReconcileMessageReceiver` would be difficult.

https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/queue/managed/QueueMessageReceiver.java

sfount (approved):
Code LGTM 👍

src/main/java/uk/gov/pay/connector/queue/managed/PayoutReconcileMessageReceiver.java
kbottla: Agree! QueueMessageReceiver will be refactored as part of this story which includes separating multiple schedulers (which is currently the case) into their own receiver

PR #2278, opened 2020-04-16
PP-6267 Abstract queue functions and common fields

kbottla:
## WHAT
- Move SQS queue specific functionality to abstract class which makes it easier to replace with different queue technology if needed and reuse functionality if new queues are added
- Also extracts common fields to parent class

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/AbstractQueue.java
+++ b/src/main/java/uk/gov/pay/connector/queue/AbstractQueue.java
@@ -1,0 +1,41 @@
+package uk.gov.pay.connector.queue;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import uk.gov.pay.connector.queue.sqs.SqsQueueService;
+
+import java.util.List;
+
+public abstract class AbstractQueue {
+
+    private static final String MESSAGE_ATTRIBUTES_TO_RECEIVE = "All";
+    protected ObjectMapper objectMapper;
+    private String queueUrl;
+    private int failedMessageRetryDelayInSeconds;
+    private SqsQueueService sqsQueueService;
+
+    public AbstractQueue(SqsQueueService sqsQueueService, ObjectMapper objectMapper,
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/queue/AbstractQueue.java
```diff
@@ -0,0 +1,41 @@
+package uk.gov.pay.connector.queue;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import uk.gov.pay.connector.queue.sqs.SqsQueueService;
+
+import java.util.List;
+
+public abstract class AbstractQueue {
```
sfount, line 8: Does this need to be called `AbstractQueue` if it has the `abstract` property? `CaptureQueue extends Queue` should be understandable

src/main/java/uk/gov/pay/connector/queue/CaptureQueue.java
```diff
@@ -15,42 +15,32 @@
...
+                        .getFailedCaptureRetryDelayInSeconds());
     }
 
     public void sendForCapture(ChargeEntity charge) throws QueueException {
         String message = new GsonBuilder()
                 .create()
                 .toJson(ImmutableMap.of("chargeId", charge.getExternalId()));
 
-        QueueMessage queueMessage = sqsQueueService.sendMessage(captureQueueUrl, message);
+        QueueMessage queueMessage = sendMessageToQueue(message);
```
sfount, line 37: nice 👍

sfount (approved):
Really nice clean up 👍

src/main/java/uk/gov/pay/connector/queue/AbstractQueue.java
kbottla: Think we have mixed conventions in this area and either way seems fine. Java core API has lots of classes starting with Abstract* https://docs.oracle.com/javase/6/docs/api/ which this is based on.

PR #2280, opened 2020-04-16
PP-6267 Move queue level classes to domain specific packages

kbottla:
WHAT
- Moved `queue` level classes to domain specific packages


alexbishop1 (approved):
https://www.youtube.com/watch?v=zkHOVJINRD8

PR #2285, opened 2020-04-23
PP-6362: Add color depth

oswaldquek:
Add color depth to AuthCardDetails and EpdqPayloadDefinitionForNewOrder.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
@@ -23,6 +23,7 @@ import uk.gov.pay.connector.gateway.epdq.model.response.EpdqAuthorisationRespons
 import uk.gov.pay.connector.gateway.epdq.model.response.EpdqCancelResponse;
 import uk.gov.pay.connector.gateway.epdq.model.response.EpdqQueryResponse;
 import uk.gov.pay.connector.gateway.model.Auth3dsResult.Auth3dsResultOutcome;
+import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 import uk.gov.pay.connector.gateway.model.request.Auth3dsResponseGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
@@ -300,6 +301,11 @@ public class EpdqPaymentProvider implements PaymentProvider {
                         anEpdq3DsAuthoriseOrderRequestBuilder(frontendUrl) :
                         anEpdqAuthoriseOrderRequestBuilder();
 
+        AuthCardDetails authCardDetails = request.getAuthCardDetails();
+        if (request.getIntegrationVersion3ds() != 2 && !request.isRequires3ds()) {
+            authCardDetails.setJsScreenColorDepth(null);
+        }
+
         return epdqOrderRequestBuilder
                 .withOrderId(request.getChargeExternalId())
                 .withPassword(request.getGatewayAccount().getCredentials().get(CREDENTIALS_PASSWORD))
@@ -308,7 +314,7 @@ public class EpdqPaymentProvider implements PaymentProvider {
                 .withMerchantCode(request.getGatewayAccount().getCredentials().get(CREDENTIALS_MERCHANT_ID))
                 .withDescription(request.getDescription())
                 .withAmount(request.getAmount())
-                .withAuthorisationDetails(request.getAuthCardDetails())
+                .withAuthorisationDetails(authCardDetails)
                 .build();
     }
 
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
@@ -122,8 +122,11 @@ public class EpdqPaymentProvider implements PaymentProvider {
     @Override
     public GatewayResponse<BaseAuthoriseResponse> authorise(CardAuthorisationGatewayRequest request) throws GatewayException {
         URI url = URI.create(String.format("%s/%s", gatewayUrlMap.get(request.getGatewayAccount().getType()), ROUTE_FOR_NEW_ORDER));
-        GatewayClient.Response response = authoriseClient.postRequestFor(url, request.getGatewayAccount(), 
-                buildAuthoriseOrder(request, frontendUrl), getGatewayAccountCredentialsAsAuthHeader(request.getGatewayAccount()));
+        GatewayClient.Response response = authoriseClient.postRequestFor(
+                url, 
+                request.getGatewayAccount(), 
+                buildAuthoriseOrder(request, frontendUrl), 
+                getGatewayAccountCredentialsAsAuthHeader(request.getGatewayAccount()));
         return getEpdqGatewayResponse(response, EpdqAuthorisationResponse.class);
     }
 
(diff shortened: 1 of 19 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -297,16 +301,20 @@ private GatewayOrder buildAuthoriseOrder(CardAuthorisationGatewayRequest request
                         anEpdq3DsAuthoriseOrderRequestBuilder(frontendUrl) :
                         anEpdqAuthoriseOrderRequestBuilder();
 
+        AuthCardDetails authCardDetails = request.getAuthCardDetails();
+        if (request.getIntegrationVersion3ds() != 2 && !request.isRequires3ds()) {
+            authCardDetails.setJsScreenColorDepth(null);
```
alexbishop1, line 306: I’m not sure mutating the request is the right thing to do here. I think we want to go down the path of only adding a value to the request to ePDQ if all the necessary conditions are met.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNewOrder.java
```diff
@@ -37,6 +38,10 @@
                 .add(EXPIRY_DATE_KEY, templateData.getAuthCardDetails().getEndDate())
                 .add(OPERATION_KEY, templateData.getOperationType())
                 .add(ORDER_ID_KEY, templateData.getOrderId());
+        
+        if (templateData.getAuthCardDetails().getJsScreenColorDepth().isPresent()) {
+            parameterBuilder.add(BROWSER_COLOR_DEPTH, templateData.getAuthCardDetails().getJsScreenColorDepth().get());
+        }
```
alexbishop1, line 44: This can be written as:

```java
templateData.getAuthCardDetails().getJsScreenColorDepth()
    .ifPresent(bcp -> parameterBuilder.add(BROWSER_COLOR_DEPTH, bcp));
```

However, I don’t think this is actually the right class. `EpdqPayloadDefinitionForNewOrder` is only used for non-3DS payments. For 3DS ones, we have `EpdqPayloadDefinitionForNew3dsOrder`, which extends this one. The decision of which one to use is made at the beginning of `buildAuthoriseOrder(…)` in `EpdqPaymentProvider`.

Since the data ePDQ require for a 3DS1 payment is a superset of the data required for a non-3DS payment, and the data ePDQ require for a 3DS2 payment is a superset of the data required for a 3DS1 payment, we could add another variant, `EpdqPayloadDefinitionForNew3ds2Order`, extending the existing `EpdqPayloadDefinitionForNew3dsOrder`.

This would mean that `buildAuthoriseOrder(…)` in `EpdqPaymentProvider` could decide, essentially, which parameters we’re going to send (based on the payment being non-3DS, 3DS1 or 3DS2) and `EpdqPayloadDefinitionForNew3ds2Order` could encapsulate the logic to always supply valid data (e.g checking the provided colour depth is valid according to ePDQ and putting in the default in if not).

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForNewOrderTest.java
```diff
@@ -58,44 +58,54 @@
...
+        List<NameValuePair> result = epdqPayloadDefinitionForNewOrder.extract(epdqTemplateData);
+        assertThat(result, hasItem(new BasicNameValuePair(BROWSER_COLOR_DEPTH, "1")));
+    }
 
-        List<NameValuePair> result = epdqPayloadDefinitionForNewOrder.extract(mockTemplateData);
+    @Test
+    public void shouldNotExtract_js_screen_color_depth_ifNotPresent() {
+        List<NameValuePair> result = epdqPayloadDefinitionForNewOrder.extract(epdqTemplateData);
+        assertEquals(0, result.stream().filter(nameValuePair -> nameValuePair.getName().equals(BROWSER_COLOR_DEPTH)).count());
+    }
```
alexbishop1, line 103: Not sure these two tests are really testing the right thing. Ultimately, whether we send the `browserColorDepth` parameter is controlled by whether 3DS2 is enabled on the gateway account, not the content of the request from frontend.

We either need some logic earlier on to decide whether to populate the `browserColorDepth` in `epdqTemplateData` based on whether 3DS2 is enabled, or we need to always populate the `browserColorDepth` in `epdqTemplateData` and decide when we build the request to ePDQ whether to send it or not (which I think is more consistent with what we do elsewhere and therefore my preference).

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProviderTest.java
```diff
@@ -33,6 +35,15 @@ public void shouldGenerateNoTransactionId() {
...
     }
 
+    @Test
+    public void doNotInclude_js_screen_color_depth_inAuthoriseOrderIf3ds2NotEnabled() throws Exception {
+        mockPaymentProviderResponse(200, successAuthResponse());
+        CardAuthorisationGatewayRequest request = buildTestAuthorisationRequest();
+        request.getAuthCardDetails().setJsScreenColorDepth("1");
+        provider.authorise(request);
+        verifyPaymentProviderRequest(successAuthRequest());
+    }
```
alexbishop1, line 45: How is this actually testing the `browserColorDepth` parameter isn’t sent?

alexbishop1 (changes requested):
I would have preferred it if the refactoring and the new functionality were in two separate pull requests: I found it quite taxing to follow.

It doesn’t look like there’s anything in here to ensure that the `browserColorDepth` we send to ePDQ is one of the ones they support (1, 2, 4, 8, 15, 16, 24 or 32) nor to ensure that we always send a value for every 3DS2 payment (defaulting to 24 if frontend either didn’t send a value at all or sent an invalid one).

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNewOrder.java
oswaldquek: I'm not sure how I followed through to the wrong class. This is probably as a result of how confusing the code is. I might have a think about refactoring the `buildAuthoriseOrder(…)` first. 👍 for introducing a `EpdqPayloadDefinitionForNew3ds2Order`, I will definitely do that.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForNewOrderTest.java
oswaldquek: I was trying to munge 3DS1 and 3DS2 code together here, which was why I made a test to ensure the `browserColorDepth` isn't sent if it isn't provided. However with the new `EpdqPayloadDefinitionForNew3ds2Order` class this will go away.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProviderTest.java
oswaldquek: Ah, because the `successAuthRequest` contains a sha sign calculated from all the params in the `AuthCardDetails`. The test will fail if the screen color depth is included in the caluclation. This was definitely not clear in the test and is worth a comment.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
oswaldquek: I'm generally for immutability but Java doesn't do this well and our code doesn't do this either with the getter and setter methods. I did this mutation because the screen color depth could (but very unlikely of course) be included in a non-3DS2 auth request, and the result would be the inclusion of it in the sha sign calculation.
I don't like this coupling though, and will think about how to make it better.

oswaldquek:
Thanks for the review. Yes, I made the carnal mistake of refactoring and coding new functionality in the same PR, something I find annoying in other PRs!
I also made some code mistakes here, partly because the epdq code is really confusing, so I think the best thing is to close this PR and make a new one.

oswaldquek:
I also found it difficult to figure out where to start with this story. Originally I did an integration test but realised it would touch too many pieces of code and decided `EpdqPayloadDefinitionForNewOrder` (or `EpdqPayloadDefinitionForNew3ds2Order` in the up-and-coming PR) is the best place to start.

PR #2288, opened 2020-04-24
PP-6362: Add new epdq payload definition for 3ds2

oswaldquek:
Also adds the `browserColorDepth` to the payload definition.
Note this PR doesn't handle calculation of the sha sign.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
@@ -1,0 +1,28 @@
+package uk.gov.pay.connector.gateway.epdq.payload;
+
+import org.apache.http.NameValuePair;
+import uk.gov.pay.connector.gateway.epdq.EpdqOrderRequestBuilder;
+
+import java.util.List;
+import java.util.Set;
+
+public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionForNew3dsOrder {
+
+    public final static String BROWSER_COLOR_DEPTH = "browserColorDepth";
+    
+    private final static String DEFAULT_BROWSER_COLOR_DEPTH = "24";
+    private final static Set<String> VALID_SCREEN_COLOR_DEPTHS = Set.of("1", "2", "4", "8", "15", "16", "24", "32");
+
+    @Override
+    public List<NameValuePair> extract(EpdqOrderRequestBuilder.EpdqTemplateData templateData) {
+        List<NameValuePair> nameValuePairs = super.extract(templateData);
+        return newParameterBuilder(nameValuePairs)
+                .add(BROWSER_COLOR_DEPTH, getBrowserColorDepth(templateData))
+                .build();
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -0,0 +1,26 @@
...
+import org.apache.http.NameValuePair;
+import uk.gov.pay.connector.gateway.epdq.EpdqOrderRequestBuilder;
+
+import java.util.List;
+
+public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionForNew3dsOrder {
+
+    public final static String BROWSER_COLOR_DEPTH = "browserColorDepth";
+    
+    private final static List<String> VALID_SCREEN_COLOR_DEPTHS = List.of("1", "2", "4", "8", "15", "16", "24", "32");
```
alexbishop1: A `Set` may be more appropriate

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -0,0 +1,26 @@
...
+    public List<NameValuePair> extract(EpdqOrderRequestBuilder.EpdqTemplateData templateData) {
+        List<NameValuePair> nameValuePairs = super.extract(templateData);
+        return newParameterBuilder(nameValuePairs)
+                .add(BROWSER_COLOR_DEPTH, getBrowserColorDepth(templateData))
+                .build();
+    }
+
+    private String getBrowserColorDepth(EpdqOrderRequestBuilder.EpdqTemplateData templateData) {
+        String browserColorDepth = templateData.getAuthCardDetails().getJsScreenColorDepth().orElse("24");
+        return VALID_SCREEN_COLOR_DEPTHS.contains(browserColorDepth) ? browserColorDepth : "24";
```
alexbishop1:
``` java
return templateData.getAuthCardDetails().getJsScreenColorDepth()
    .filter(VALID_SCREEN_COLOR_DEPTHS::contains).orElse("24");
```

Probably best to put `"24"` in a constant, just so we can give it a meaningful name like `DEFAULT_BROWSER_COLOR_DEPTH`.

alexbishop1 (changes requested):
Looks good — just a couple of small suggestions.

I think the `SHASIGN` stuff will just work without you having to do anything else.

Of course, there’s still no code to switch to using the new `EpdqPayloadDefinitionForNew3ds2Order` class yet but it’s probably a good idea to fill out that class before adjusting the code path to use it when 3DS2 is enabled.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
oswaldquek: done

alexbishop1 (approved):

PR #2289, opened 2020-04-27
PP-6362: Delete EpdqOrderRequestBuilder

oswaldquek:
EpdqOrderRequestBuilder was a class that had lots of indirection which it made
it difficult to figure out how a GatewayOrder was being built from the
operations (e.g. capture, authorise, cancel etc). This commit simplifies this
indirection by

- instantiating concrete subclasses of `EpdqPayloadDefinition`.  These are
   EpdqPayloadDefinitionFor[Cancel|Capture|Maintenance|New3ds2|New3ds|NewOrder|Query|Refund]Order.java.
- moving the building of the GatewayOrder into
  `EpdqPayloadDefinition.createGatewayOrder`
- removing inappropriate uses of inheritance and interfaces where possible


src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionMaintenanceOrderTest.java
```diff
@@ -24,29 +27,51 @@
...
+                    new BasicNameValuePair("PAYID", PAY_ID),
+                    new BasicNameValuePair("PSPID", PSP_ID),
+                    new BasicNameValuePair("PSWD", PASSWORD),
+                    new BasicNameValuePair("USERID", USER_ID))
+                    .build()));
+        });
     }
 
+    @Test
+    public void testOnlyTransactionIdIsSentIfBothTransactionIdAndOrderIdAvailable() {
```
oswaldquek: this test was directly copied from https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java#L59. i think the test name is inaccurate as the `extractPairs` variable in the test will have both the transaction id and order id populated.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionMaintenanceOrderTest.java
```diff
@@ -24,29 +27,51 @@
...
+        epdqTemplateData.setOrderId("Order-Id");
+        List<NameValuePair> extractPairs = captureOrder.extract(epdqTemplateData);
+        assertThat(extractPairs, hasItem(new BasicNameValuePair("PAYID", PAY_ID)));
+    }
 
+    @Test
+    public void testThatOrderIdIsSentIfTransactionIsNotAvailable() {
```
oswaldquek: copied from https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java#L70

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqCaptureHandler.java
```diff
@@ -47,13 +47,13 @@ public CaptureResponse capture(CaptureGatewayRequest request) {
...
-                .withTransactionId(request.getTransactionId())
-                .build();
+        EpdqTemplateData templateData = new EpdqTemplateData();
+        templateData.setUserId(request.getGatewayAccount().getCredentials().get(CREDENTIALS_USERNAME));
+        templateData.setPassword(request.getGatewayAccount().getCredentials().get(CREDENTIALS_PASSWORD));
+        templateData.setShaInPassphrase(request.getGatewayAccount().getCredentials().get(CREDENTIALS_SHA_IN_PASSPHRASE));
+        templateData.setMerchantCode(request.getGatewayAccount().getCredentials().get(CREDENTIALS_MERCHANT_ID));
+        templateData.setTransactionId(request.getTransactionId());
+        
+        return new EpdqPayloadDefinitionForCaptureOrder().createGatewayOrder(templateData);
```
alexbishop1, line 57: Perhaps not for now but it doesn’t look like `EpdqPayloadDefinitionForCaptureOrder` and friends actually hold any state. Could it be a singleton?

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -273,61 +274,61 @@ private static boolean responseDoesNotMatchWithUserResult(BaseAuthoriseResponse.
...
+        templateData.setPassword(request.getGatewayAccount().getCredentials().get(CREDENTIALS_PASSWORD));
+        templateData.setShaInPassphrase(request.getGatewayAccount().getCredentials().get(CREDENTIALS_SHA_IN_PASSPHRASE));
+        templateData.setUserId(request.getGatewayAccount().getCredentials().get(CREDENTIALS_USERNAME));
+        templateData.setMerchantCode(request.getGatewayAccount().getCredentials().get(CREDENTIALS_MERCHANT_ID));
+
+        var epdqPayloadDefinitionForQueryOrder = new EpdqPayloadDefinitionForQueryOrder();
+        return epdqPayloadDefinitionForQueryOrder.createGatewayOrder(templateData);
     }
 
     private GatewayOrder buildQueryOrderRequestFor(ChargeEntity charge) {
```
alexbishop1, line 277: Not for now, but this method appears to be identical to the one above except for how it obtains the charge ID.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinition.java
```diff
@@ -1,50 +1,49 @@
...
-import java.util.Comparator;
 import java.util.List;
 
-import static org.apache.commons.lang3.StringUtils.isEmpty;
-
 public abstract class EpdqPayloadDefinition {
 
-    static ParameterBuilder newParameterBuilder() {
-        return new ParameterBuilder();
+    private static final SignatureGenerator signatureGenerator = new EpdqSha512SignatureGenerator();
```
alexbishop1: This constant name should be in `SCREAMING_SNAKE_CASE`.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinition.java
```diff
@@ -1,50 +1,49 @@
...
+
+    /**
+     * ePDQ have never confirmed that they use Windows-1252 to decode
+     * application/x-www-form-urlencoded payloads sent by us to them and use
+     * Windows-1252 to encode application/x-www-form-urlencoded notification
+     * payloads sent from them to us but experimentation — and specifically the
+     * fact that ’ (that’s U+2019 right single quotation mark in Unicode
+     * parlance) seems to encode to %92 — makes us believe that they do
+     */
+    private static final Charset EPDQ_APPLICATION_X_WWW_FORM_URLENCODED_CHARSET = Charset.forName("windows-1252");
```
alexbishop1: This constant is now here and in `EpdqPaymentProvider` — I think it should only be in the latter.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinition.java
```diff
@@ -1,50 +1,49 @@
...
+
+    protected abstract List<NameValuePair> extract(EpdqTemplateData templateData);
+
+    public GatewayOrder createGatewayOrder(EpdqTemplateData templateData) {
+        templateData.setOperationType(getOperationType());
```
alexbishop1, line 34: Not for now, but I think every concrete subclass of `EpdqPayloadDefinition` probably knows enough to add its own operation type as a constant rather than it being added to the data passed to it here.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinition.java
```diff
@@ -1,50 +1,49 @@
...
+        List<NameValuePair> templateParams = extract(templateData);
+        List<NameValuePair> requestParams = new ArrayList<>(templateParams);
+        requestParams.add(new BasicNameValuePair("SHASIGN", signatureGenerator.sign(templateParams, templateData.getShaInPassphrase())));
```
alexbishop1: I found the `templateParams` and `requestParams` variable names a little non-obvious here. Maybe we could avoid needing both by restructuring the code a little:

```java
var params = new ArrayList<NameValuePair> = extract(templateData);
String signature = signatureGenerator.sign(params, templateData.getShaInPassphrase());
params.add(new BasicNameValuePair("SHASIGN", signature));
```

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/ParameterBuilder.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.http.NameValuePair;
+import org.apache.http.message.BasicNameValuePair;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+import static org.apache.commons.lang3.StringUtils.isEmpty;
+
+public class ParameterBuilder {
```
alexbishop1: Could we call this `EpdqParameterBuilder`? The current name is a bit too generic.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForCancelOrderTest.java
```diff
@@ -0,0 +1,27 @@
...
+import uk.gov.pay.connector.gateway.model.OrderRequestType;
+import uk.gov.pay.connector.util.TestTemplateResourceLoader;
+
+import static org.junit.Assert.assertEquals;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.EPDQ_CANCEL_REQUEST;
+
+public class EpdqPayloadDefinitionForCancelOrderTest {
+
+    @Test
+    public void assertGatewayOrder() {
```
alexbishop1: I think this test name (and the ones with the same name in other test classes) could be more descriptive.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java
```diff
@@ -1,85 +0,0 @@
...
-import org.junit.Test;
-import uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForMaintenanceOrder;
-
-import java.util.List;
-
-import static org.hamcrest.Matchers.hasItem;
-import static org.hamcrest.collection.IsIterableContainingInOrder.contains;
-import static org.junit.Assert.assertThat;
-
-public class EpdqPayloadDefinitionForMaintenanceOrderTest {
```
alexbishop1, line 15: If these test class were still here, would these tests still pass? It may well be that these tests are unnecessary but I’d feel more confident if we left them in place and saw them still pass after this refactoring. We can look at removing them later.

src/test/java/uk/gov/pay/connector/gateway/templates/FormUrlEncodedStringBuilderTest.java
```diff
@@ -1,74 +0,0 @@
...
-
-import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
-
-import static org.hamcrest.core.Is.is;
-import static org.junit.Assert.assertThat;
-import static org.mockito.Mockito.when;
-
-@RunWith(MockitoJUnitRunner.class)
-public class FormUrlEncodedStringBuilderTest {
```
alexbishop1, line 21: This test is doing some interesting checks relating to encoding that I don’t think we want to lose, to the point where I’m pondering if merging `FormUrlEncodedStringBuilder` into `EpdqPaymentDefintion` is necessarily the right decision. Perhaps we could modify some of the new `assertGatewayOrder()` tests so they have data that covers the cases here.

alexbishop1 (changes requested):
This is a really good improvement.

I would have preferred it to be delivered a bit more incrementally but I recognise that would be hard. Perhaps some of the changes like coalescing `EpdqSignedPayloadDefinition` and `FormUrlEncodedStringBuilder` into `EpdqPayloadDefinition` could have been done separately.

I like how a lot of the tests still assert the results by comparing to data loaded from (unchanged!) files. That gives me confidence that the behaviour is the same.

I’ve left some comments. Since this PR is already quite big, I’ve prefixed anything that can probably wait until a subsequent PR with “Perhaps not for now”. Anything that should definitely wait until a subsequent PR has been prefixed with “Not for now”.

One thing I’m less sure about is how the subclasses of `EpdqPayloadDefinition` now have `extract(…)` methods and an inherited `createGatewayOrder(…)` method that seem to operate at different levels. This is particularly noticeable in the tests. I think we should try to smooth that out in the future.

Not for now (obviously) but I think we can go even further than this and remove more intermediary classes like `EpdqTemplateData`. But what we have now, particularly the break-up of `EpdqOrderRequestBuilder`, is a great step forward in improving comprehensibility.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinition.java
oswaldquek: Oh I forgot to delete it from `EpdqPaymentProvider`. Why should it in `EpdqPaymentProvider`? It's only needed when building the payload which happens in `EpdqPayloadDefinition`.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java
oswaldquek: The tests here haven't been deleted. `testBaseValuePairs` and `testBaseValuePairsWithAmount` have been merged into `EpdqPayloadDefinitionMaintenanceOrderTest.shouldExtractParametersFromTemplate`. 
`testOnlyTransactionIdIsSentIfBothTransactionIdAndOrderIdAvailable` and `testThatOrderIdIsSentIfTransactionIsNotAvailable` have also been moved to `EpdqPayloadDefinitionMaintenanceOrderTest`.

src/test/java/uk/gov/pay/connector/gateway/templates/FormUrlEncodedStringBuilderTest.java
oswaldquek: The reason why I didn't think these were necessary is because in the `assertGatewayOrder()` tests the following is tested:
```
assertEquals(TestTemplateResourceLoader.load(EPDQ_AUTHORISATION_3DS_REQUEST), gatewayOrder.getPayload());
```
This loads a file (https://github.com/alphagov/pay-connector/blob/master/src/test/resources/templates/epdq/authorisation-3ds-request.txt in this case) which essentially covers our bases here.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForCancelOrderTest.java
oswaldquek: changed to `assert_payload_and_order_request_type_are_as_expected`

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java
oswaldquek: I probably should have kept the name of the test to `EpdqPayloadDefinitionForMaintenanceOrderTest`. I don't know why I changed it. I'll change it back.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinition.java
oswaldquek: 👍 I'll do in the next PR.

oswaldquek:
Thanks for the review @alexbishop1. I'll make the "Not for now" and "Perhaps not for now" changes in the next PR.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinition.java
alexbishop1: Ah, yes, in that case it should only be in `EpdqPayloadDefinition`.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java
alexbishop1: I see. I thought we’d lost some of the tests.

And how we ended up with  both `EpdqPayloadDefinitionMaintenanceOrderTest` and `EpdqPayloadDefinitionForMaintenanceOrderTest` makes the mind boggle somewhat!

src/test/java/uk/gov/pay/connector/gateway/templates/FormUrlEncodedStringBuilderTest.java
alexbishop1: It covers things like encoding spaces but not some of the obscure cases like encoding characters that encode differently in Windows-1252 to UTF-8 (like é).

oswaldquek: Yeah but `FormUrlEncodedStringBuilder` was really just encapsulating `org.apache.http.client.utils.URLEncodedUtils.format` so it wasn't doing anything special. Do we really need to test library code?
I'll add it back in if you insist though.

alexbishop1: Nah, you’re good. I’ve just been burned by the earlier times when we got the character encoding wrong!

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPayloadDefinitionForMaintenanceOrderTest.java
oswaldquek: there's only `EpdqPayloadDefinitionForMaintenanceOrderTest` now :)

alexbishop1 (approved):
Good stuff!

PR #2290, opened 2020-04-27
PP-6362: Set browser language to be sent to epdq for 3ds2 payment

oswaldquek:
This PR does not include the scenario where the request sent from frontend includes no js_navigator_language.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
@@ -4,6 +4,8 @@ import org.apache.http.NameValuePair;
 import uk.gov.pay.connector.gateway.epdq.EpdqTemplateData;
 
 import java.util.List;
+import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqParameterBuilder.newParameterBuilder;
@@ -11,8 +13,9 @@ import static uk.gov.pay.connector.gateway.epdq.payload.EpdqParameterBuilder.new
 public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionForNew3dsOrder {
 
     public final static String BROWSER_COLOR_DEPTH = "browserColorDepth";
-    
-    private final static String DEFAULT_BROWSER_COLOR_DEPTH = "24";
+    public final static String BROWSER_LANGUAGE = "browserLanguage";
+    public final static String DEFAULT_BROWSER_COLOR_DEPTH = "24";
+
     private final static Set<String> VALID_SCREEN_COLOR_DEPTHS = Set.of("1", "2", "4", "8", "15", "16", "24", "32");
 
     public EpdqPayloadDefinitionForNew3ds2Order(String frontendUrl) {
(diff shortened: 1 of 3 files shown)


oswaldquek:
Probably best to wait for https://github.com/alphagov/pay-connector/pull/2289 to be merged.

src/main/java/uk/gov/pay/connector/gateway/model/AuthCardDetails.java
```diff
@@ -166,4 +167,13 @@ public PayersCardPrepaidStatus getPayersCardPrepaidStatus() {
     public Optional<String> getIpAddress() {
         return Optional.ofNullable(ipAddress);
     }
+
+    @JsonProperty("js_navigator_language")
+    public void setJsNavigatorLanguage(String jsNavigatorLanguage) {
+        this.jsNavigatorLanguage = jsNavigatorLanguage;
+    }
+
+    public String getJsNavigatorLanguage() {
```
alexbishop1, line 176: Since this could be null, we should return an `Optional<String>` here

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqParameterBuilder.java
```diff
@@ -28,6 +28,12 @@ public EpdqParameterBuilder add(String name, String value) {
         return this;
     }
 
+    public EpdqParameterBuilder replace(String name, String value) {
+        parameters.removeIf(x -> x.getName().equals(name));
```
alexbishop1: Maybe `existingParameter` rather than just `x`.

If we’re having to do this, I wonder if making this a map rather than a list would be more appropriate.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -22,9 +26,18 @@ public EpdqPayloadDefinitionForNew3ds2Order(String frontendUrl) {
     @Override
     public List<NameValuePair> extract(EpdqTemplateData templateData) {
         List<NameValuePair> nameValuePairs = super.extract(templateData);
-        return newParameterBuilder(nameValuePairs)
-                .add(BROWSER_COLOR_DEPTH, getBrowserColorDepth(templateData))
-                .build();
+        EpdqParameterBuilder parameterBuilder = newParameterBuilder(nameValuePairs)
+                .add(BROWSER_COLOR_DEPTH, getBrowserColorDepth(templateData));
+
+        if (isNotBlank(templateData.getAuthCardDetails().getJsNavigatorLanguage())) {
```
alexbishop1: I think we should have a null check here but I don’t think we need a blank string check. `Locale.forLanguageTag("").toLanguageTag()` or `Locale.forLanguageTag(" ").toLanguageTag()` or similar will return `"und"` (undefined), which is a legitimate language tag. I don’t think we need to worry too much about the edge case of browsers sending something that isn’t a valid language tag (however, we do need to deal with JavaScript not being enabled and do the null check).

Plus, if we do a null check rather than a blank string check, we can have cool code like this:

```java
String language = Optional.ofNullable(templateData.getAuthCardDetails().getJsNavigatorLanguage())
    .map(Locale::forLanguageTag).map(Locale::toLanguageTag).orElse("en-GB");
```

alexbishop1 (changes requested):
Couple of minor qualms but this looks fine. Happy for it to go in now without the code to default to the right language since none of this will actually be used yet.

src/main/java/uk/gov/pay/connector/gateway/model/AuthCardDetails.java
oswaldquek: I know but I want to use `Optional<String>` when a test calls for it. Maybe I should include the `public void should_include_default_browserLanguage` test in this PR too then.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqParameterBuilder.java
oswaldquek: I think I will revert this code as it strictly test code used in `EpdqPayloadDefinitionForNew3ds2OrderTest.should_include_browserColorDepth` and is not production code.

src/main/java/uk/gov/pay/connector/gateway/model/AuthCardDetails.java
oswaldquek: Actually I will hold off on this for now. The reason I want to have the `should_include_default_browserLanguage` test in another PR is because the `browserLanguage` default depends on whether the card payment is in english or welsh which adds more extra code and is deserving of its own PR.

oswaldquek: So the `Optional<String>` will go into the next PR.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
oswaldquek: i've amended to 
```
Optional.ofNullable(templateData.getAuthCardDetails().getJsNavigatorLanguage())
                .map(Locale::forLanguageTag).map(Locale::toLanguageTag).ifPresent(lang -> parameterBuilder.add(BROWSER_LANGUAGE, lang));
``` 
for now. 
Default values will be dealt with in the next PR.

alexbishop1 (changes requested):
Looks good.

Could you squash the two commits together just so we don’t have code that’s added in one commit and then removed in the next?

oswaldquek:
@alexbishop1 done

alexbishop1 (approved):

PR #2291, opened 2020-04-28
PP-6362: Set browser language default

oswaldquek:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
@@ -1,11 +1,11 @@
 package uk.gov.pay.connector.gateway.epdq.payload;
 
 import org.apache.http.NameValuePair;
+import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.connector.gateway.epdq.EpdqTemplateData;
 
 import java.util.List;
 import java.util.Locale;
-import java.util.Optional;
 import java.util.Set;
 
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqParameterBuilder.newParameterBuilder;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -17,25 +17,44 @@
...
+                .map(Locale::forLanguageTag)
+                .map(Locale::toLanguageTag)
+                .orElse(getDefaultBrowserLanguage());
+    }
+
+    private String getDefaultBrowserLanguage() {
+        String defaultBrowserLanguage = null;
+        switch (paymentLanguage) {
+            case ENGLISH: 
+                defaultBrowserLanguage = Locale.forLanguageTag("en-GB").toLanguageTag();
```
alexbishop1: Since `""en-GB"` is a constant, we already know what the result of `Locale.forLanguageTag("en-GB").toLanguageTag()` will be (it will be `"en-GB"`) so we can just use `"en-GB"` here.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -17,25 +17,44 @@
...
+                break;
+            case WELSH: 
+                defaultBrowserLanguage = Locale.forLanguageTag("cy-GB").toLanguageTag();
```
alexbishop1: Just `"cy"`. There aren’t really variants of Welsh (in the same way there are, for example, British and American variants of English), so `"cy-GB"` isn’t really a thing.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -17,25 +17,44 @@
...
+                break;
+        }
+        return defaultBrowserLanguage;
```
alexbishop1: If we ever add another `SupportedLanguage` and don’t update this code, it return null.

Fortunately, `toString()`ing `SupportedLanguage` gives a language tag and — except in the case of English — it’s already what we want. So we could do something like this:

```java
private String getDefaultBrowserLanguage() {
    if (paymentLanguage == SupportedLanguage.ENGLISH) {
        return "en-GB";
    }
    return paymentLanguage.toString();
}
```

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -82,28 +88,27 @@ public void setup() {
...
+                .withBrowserLanguage(Locale.forLanguageTag("de").toLanguageTag())
                 .build()));
     }
 
-
     @Test
-    public void should_include_browserColorDepth() {
-        authCardDetails.setJsScreenColorDepth("1");
+    @Parameters({"ENGLISH, en-GB", "WELSH, cy-GB"})
+    public void should_include_default_browserLanguage(SupportedLanguage language, String expectedBrowserLanguage) {
```
alexbishop1: Maybe something like `should_include_payment_browserLanguage_if_none_provided`?

alexbishop1 (changes requested):
Looks good — few small suggestions

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
oswaldquek: Great spot, thanks.

alexbishop1 (approved):

PR #2292, opened 2020-04-29
PP-6381 Allow Worldpay users to try 3DS twice

alexbishop1:
Allow paying users paying for a service that uses a Worldpay gateway account to try 3DS for a second time if Worldpay asks us to do so when we try to authorise the first attempt (previously we treated this as an `AUTHORISATION REJECTED`).

We know from experience that this will cause a lot of “verification of PaRes failed” errors (error code 7) when we try to authorise the second attempt, which results in the user being shown an error page after trying 3DS twice (with their payment becoming `AUTHORISATION ERROR`).

The idea is to keep this on for just long enough to collect some more logging information (we enhanced the logging in PP-6380) so we can send it to Worldpay and they can hopefully figure out why the errors are occurring.

with @nimalank7


nimalank7 (approved):
LGTM!

PR #2294, opened 2020-04-29
bau: Make epdq order classes singletons

oswaldquek:
Where subclasses of `EpdqPayloadDefinition` have no state, these can be made singletons.


alexbishop1 (commented):
Is this enough? Don’t we need to also inject them into `EpdqPaymentProvider` and store them as fields there (rather than `new`ing them each time)?

oswaldquek:
Ugh 🤦 you're right

oswaldquek:
@alexbishop1 updated

alexbishop1 (approved):

PR #2295, opened 2020-04-29
Revert "PP-6381 Allow Worldpay users to try 3DS twice"

alexbishop1:
Reverts alphagov/pay-connector#2292 (since we’ll have enough logging data)


nimalank7 (approved):
LGTM!

PR #2303, opened 2020-05-01
PP-6362 - Add timezone offset to ePDQ 3DS2 parameters

nimalank7:
Description:
- This adds jsTimezoneOffsetMins to ePDQ 3DS2 parameters

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
@@ -4,6 +4,9 @@ import org.apache.http.NameValuePair;
 import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.connector.gateway.epdq.EpdqTemplateData;
 
+import java.time.Clock;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
@@ -17,18 +20,22 @@ public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionF
     public final static String BROWSER_LANGUAGE = "browserLanguage";
     public final static String BROWSER_SCREEN_HEIGHT = "browserScreenHeight";
     public final static String BROWSER_SCREEN_WIDTH = "browserScreenWidth";
+    public final static String BROWSER_TIMEZONE_OFFSET_MINS = "browserTimezoneOffsetMins";
     public final static String DEFAULT_BROWSER_COLOR_DEPTH = "24";
     public final static String DEFAULT_BROWSER_SCREEN_HEIGHT = "480";
     public final static String DEFAULT_BROWSER_SCREEN_WIDTH = "320";
     
     private final static Pattern NUMBER_FROM_0_TO_999999 = Pattern.compile("0|[1-9][0-9]{0,5}");
+    private final static Pattern NUMBER_FROM_MINUS_999_TO_999 = Pattern.compile("-[1-9][0-9]{0,2}|0|[1-9][0-9]{0,2}");
     private final static Set<String> VALID_SCREEN_COLOR_DEPTHS = Set.of("1", "2", "4", "8", "15", "16", "24", "32");
     
     private final SupportedLanguage paymentLanguage;
+    private final Clock clock;
 
-    public EpdqPayloadDefinitionForNew3ds2Order(String frontendUrl, SupportedLanguage paymentLanguage) {
+    public EpdqPayloadDefinitionForNew3ds2Order(String frontendUrl, SupportedLanguage paymentLanguage, Clock clock) {
         super(frontendUrl);
         this.paymentLanguage = paymentLanguage;
+        this.clock = clock;
     }
 
     @Override
@@ -38,9 +45,19 @@ public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionF
                 .add(BROWSER_COLOR_DEPTH, getBrowserColorDepth(templateData))
                 .add(BROWSER_LANGUAGE, getBrowserLanguage(templateData))
                 .add(BROWSER_SCREEN_HEIGHT, getBrowserScreenHeight(templateData))
-                .add(BROWSER_SCREEN_WIDTH, getBrowserScreenWidth(templateData));
+                .add(BROWSER_SCREEN_WIDTH, getBrowserScreenWidth(templateData))
+                .add(BROWSER_TIMEZONE_OFFSET_MINS, getBrowserTimezoneOffsetMins(templateData));
 
         return parameterBuilder.build();
+    }
+
+    private String getBrowserTimezoneOffsetMins(EpdqTemplateData templateData) {
+        return templateData.getAuthCardDetails().getJsTimezoneOffsetMins()
+                .filter(timezoneOffsetMins -> NUMBER_FROM_MINUS_999_TO_999.matcher(timezoneOffsetMins).matches())
+                .map(Integer::parseInt)
+                .filter(timezoneOffsetMins -> timezoneOffsetMins >= -840 && timezoneOffsetMins <= 720)
+                .map(timezoneOffsetMins -> Integer.toString(timezoneOffsetMins))
+                .orElseGet(() -> getDefaultBrowserOffsetInMinutes());
     }
 
     private String getBrowserScreenWidth(EpdqTemplateData templateData) {
@@ -80,5 +97,12 @@ public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionF
     private String getBrowserColorDepth(EpdqTemplateData templateData) {
         return templateData.getAuthCardDetails().getJsScreenColorDepth()
                 .filter(VALID_SCREEN_COLOR_DEPTHS::contains).orElse(DEFAULT_BROWSER_COLOR_DEPTH);
+    }
+
+    private String getDefaultBrowserOffsetInMinutes() {
+        ZoneOffset currentUkOffset = ZoneId.of("Europe/London").getRules().getOffset(clock.instant());
+        int currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive = currentUkOffset.getTotalSeconds() / 60;
+        int currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative = -currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive;
+        return String.valueOf(currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative);
     }
 }
--- a/src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
@@ -12,6 +12,9 @@ import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.connector.gateway.epdq.EpdqTemplateData;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 
+import java.time.Clock;
+import java.time.Instant;
+import java.time.ZoneOffset;
 import java.util.List;
 import java.util.Locale;
 
@@ -23,6 +26,7 @@ import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionFor
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNew3ds2Order.BROWSER_LANGUAGE;
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNew3ds2Order.BROWSER_SCREEN_HEIGHT;
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNew3ds2Order.BROWSER_SCREEN_WIDTH;
+import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNew3ds2Order.BROWSER_TIMEZONE_OFFSET_MINS;
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNew3ds2Order.DEFAULT_BROWSER_COLOR_DEPTH;
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNew3ds2Order.DEFAULT_BROWSER_SCREEN_HEIGHT;
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNew3ds2Order.DEFAULT_BROWSER_SCREEN_WIDTH;
@@ -64,9 +68,11 @@ public class EpdqPayloadDefinitionForNew3ds2OrderTest {
     private static final String FRONTEND_URL = "http://www.frontend.example.com";
     private static final String ACCEPT_HEADER = "Test Accept Header";
     private static final String USER_AGENT_HEADER = "Test User Agent Header";
-
+    private static final Clock BRITISH_SUMMER_TIME_OFFSET_CLOCK = Clock.fixed(Instant.parse("2020-05-06T10:10:10.100Z"), ZoneOffset.UTC);
+    private static final Clock GREENWICH_MERIDIAN_TIME_OFFSET_CLOCK = Clock.fixed(Instant.parse("2020-01-01T10:10:10.100Z"), ZoneOffset.UTC);
+    
     private EpdqPayloadDefinitionForNew3ds2Order epdqPayloadDefinitionFor3ds2NewOrder = 
-            new EpdqPayloadDefinitionForNew3ds2Order(FRONTEND_URL, SupportedLanguage.ENGLISH);
+            new EpdqPayloadDefinitionForNew3ds2Order(FRONTEND_URL, SupportedLanguage.ENGLISH, BRITISH_SUMMER_TIME_OFFSET_CLOCK);
 
     private EpdqTemplateData epdqTemplateData = new EpdqTemplateData();
     private AuthCardDetails authCardDetails = new AuthCardDetails();
@@ -136,7 +142,7 @@ public class EpdqPayloadDefinitionForNew3ds2OrderTest {
     @Test
     @Parameters({"ENGLISH, en-GB", "WELSH, cy"})
     public void should_include_payment_browserLanguage_if_none_provided(SupportedLanguage language, String expectedBrowserLanguage) {
-        var epdqPayloadDefinitionFor3ds2NewOrder = new EpdqPayloadDefinitionForNew3ds2Order(FRONTEND_URL, language);
+        var epdqPayloadDefinitionFor3ds2NewOrder = new EpdqPayloadDefinitionForNew3ds2Order(FRONTEND_URL, language, BRITISH_SUMMER_TIME_OFFSET_CLOCK);
         List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
         assertThat(result, is(aParameterBuilder().withBrowserLanguage(expectedBrowserLanguage).build()));
     }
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -17,11 +20,14 @@
     public final static String BROWSER_LANGUAGE = "browserLanguage";
     public final static String BROWSER_SCREEN_HEIGHT = "browserScreenHeight";
     public final static String BROWSER_SCREEN_WIDTH = "browserScreenWidth";
+    public final static String BROWSER_TIMEZONE_OFFSET_MINS = "browserTimezoneOffsetMins";
     public final static String DEFAULT_BROWSER_COLOR_DEPTH = "24";
     public final static String DEFAULT_BROWSER_SCREEN_HEIGHT = "480";
     public final static String DEFAULT_BROWSER_SCREEN_WIDTH = "320";
+    public final static String DEFAULT_BROWSER_TIMEZONE_OFFSET_MINS = EpdqPayloadDefinitionForNew3ds2Order.getCurrentOffsetInMinutes();
```
alexbishop1: We shouldn’t store this in a constant because it will only be evaluated when the class loads for the first time after the application starts and the current UK time zone offset could change before the next restart. We should instead calculate this every time.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -17,11 +20,14 @@
...
     
     private final static Pattern NUMBER_FROM_0_TO_999999 = Pattern.compile("0|[1-9][0-9]{0,5}");
+    private final static Pattern NUMBER_FROM_MINUS_840_TO_720 = Pattern.compile("-?[0-9]{1,3}");
```
alexbishop1: The name isn’t quite right here because this allows anything from -999 to 999. Also, it will allow strange things like -0 or 012 (which would get parsed as octal). I think `-[1-9][0-9]{0,2}|0|[1-9][0-9]{0,2}` would work better.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -81,4 +97,11 @@ private String getBrowserColorDepth(EpdqTemplateData templateData) {
         return templateData.getAuthCardDetails().getJsScreenColorDepth()
                 .filter(VALID_SCREEN_COLOR_DEPTHS::contains).orElse(DEFAULT_BROWSER_COLOR_DEPTH);
     }
+
+    private static String getCurrentOffsetInMinutes() {
+        Instant now = Instant.now();
+        ZoneOffset currentUKOffset = ZoneId.of("Europe/London").getRules().getOffset(now);
```
alexbishop1: Usually if there’s an initialism (like UK) in a camel case identifier, it’s treated as if it were a word: `currentUkOffset`.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -81,4 +97,11 @@ private String getBrowserColorDepth(EpdqTemplateData templateData) {
...
+        Integer currentUKOffsetInMins = currentUKOffset.getTotalSeconds()/-60;
```
alexbishop1: The negative sign is a bit of a surprising head-scratcher here because JavaScript has this weird behaviour where ahead of UTC is a negative offset (I don’t know why: I found https://stackoverflow.com/questions/21102435/why-does-javascript-date-gettimezoneoffset-consider-0500-as-a-positive-off but I’m not sure it really clears it up). I’d be really tempted to make it a bit more verbose and name the variables to try to give the hapless reader a clue as to what’s going on:

```java
ZoneOffset currentUkOffset = ZoneId.of("Europe/London").getRules().getOffset(Instant.now());
int currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive = currentUkOffset.getTotalSeconds() / 60;
int currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative = -currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive;
return String.valueOf(currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative);
```

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -81,4 +97,11 @@ private String getBrowserColorDepth(EpdqTemplateData templateData) {
...
+        return currentUKOffsetInMins.toString();
```
alexbishop1: There’s no need to box the `int` into an `Integer` just to be able to `toString()` it. This would be equivalent:

```java
int currentUKOffsetInMins = currentUKOffset.getTotalSeconds()/-60;
return String.valueOf(currentUKOffsetInMins);
```

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -169,12 +171,30 @@ public void should_include_browserColorDepth() {
...
+        assertThat(result, is(aParameterBuilder().withBrowserTimezoneOffsetMins(timeZoneOffsetMins).build()));
+    }
+
+    @Test
+    @Parameters({"null", "-900", "800", "invalid", "0x6", "123L", "1.2"})
+    public void should_include_default_browserTimezoneOffSetMins_when_timezone_offset_provided_is_invalid
+            (@Nullable String timeZoneOffsetMins) {
+        authCardDetails.setJsTimezoneOffsetMins(timeZoneOffsetMins);
+        List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
+        assertThat(result, is(aParameterBuilder().build()));
```
alexbishop1: Maybe I’m missing something here but I don’t see how this is checking the default offset is added.

alexbishop1 (changes requested):
Looking good overall. A few things…

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
nimalank7: I agree with re-naming it. Is it necessarily bad to not allow -0 or 012 - we could just parse them as 0 and 12?

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
nimalank7: Ah sorry yes! I've fixed that now.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Fixed when we build the parameter class the default offset is added.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
nimalank7: Yep

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
alexbishop1: I thought `Integer.parseInt(…)` did things like treated leading zeros as octal but I’ve [just checked and apparently it doesn’t do that](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Integer.html#parseInt(java.lang.String)). So maybe it wouldn’t be too bad. I do like the appeal of having a regex that tightly constrains what we accept though. Leaves fewer possible surprises in the future.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -17,11 +20,13 @@
     public final static String BROWSER_LANGUAGE = "browserLanguage";
     public final static String BROWSER_SCREEN_HEIGHT = "browserScreenHeight";
     public final static String BROWSER_SCREEN_WIDTH = "browserScreenWidth";
+    public final static String BROWSER_TIMEZONE_OFFSET_MINS = "browserTimezoneOffsetMins";
     public final static String DEFAULT_BROWSER_COLOR_DEPTH = "24";
     public final static String DEFAULT_BROWSER_SCREEN_HEIGHT = "480";
     public final static String DEFAULT_BROWSER_SCREEN_WIDTH = "320";
     
     private final static Pattern NUMBER_FROM_0_TO_999999 = Pattern.compile("0|[1-9][0-9]{0,5}");
+    private final static Pattern NUMBER_FROM_MINUS_840_TO_720 = Pattern.compile("-[1-9][0-9]{0,2}|0|[1-9][0-9]{0,2}");
```
alexbishop1: This should really be `NUMBER_FROM_MINUS_999_TO_999` to avoid confusion.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -38,11 +43,21 @@ public EpdqPayloadDefinitionForNew3ds2Order(String frontendUrl, SupportedLanguag
...
         return parameterBuilder.build();
     }
 
+    private String getBrowserTimezoneOffsetMins(EpdqTemplateData templateData) {
+        return templateData.getAuthCardDetails().getJsTimezoneOffsetMins()
+                .filter(timezoneOffsetMins -> NUMBER_FROM_MINUS_840_TO_720.matcher(timezoneOffsetMins).matches())
+                .map(Integer::parseInt)
+                .filter(timezoneOffsetMins -> timezoneOffsetMins >= -840 && timezoneOffsetMins <= 720)
+                .map(timezoneOffsetMins -> Integer.toString(timezoneOffsetMins))
+                .orElse(getCurrentBrowserOffsetInMinutes());
```
alexbishop1: This should be `.orElseGet(this::getDefaultBrowserLanguage)` or the default will be calculated every time even if it’s not needed.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
nimalank7: Yep - the regex has been changed accordingly

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
alexbishop1: Oh, I see now.

Ideally, we shouldn’t be relying on the current time in tests because it makes each run slightly different. In reality, we do that a lot. However, in this case we have the added twist that when this test runs in winter we’ll be testing the default winter offset (but not the summer one) and when it runs in summer we’ll be testing the default summer offset (but not the winter one).

There is a solution to this: it’s to use a [`Clock`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Clock.html). We can make `EpdqPayloadDefinitionForNew3ds2Order` take a `Clock` as a constructor parameter and call `clock.instant()` whenever it wants the current time.

In the real code to instantiate the `EpdqPayloadDefinitionForNew3ds2Order` object (which I believe hasn’t been written yet), we can pass in `Clock.systemUTC()`, which provides the real time. In the tests, we can pass in `Clock.fixed(…)` set to whatever non-changing time we like. This would allow us to have two variants of this test: one where it’s winter time and where it’s summer.

I don’t think we use `Clock`s for testing in this way yet but I think the unique daylight saving time differences here make it worth introducing.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
alexbishop1: This should be `NUMBER_FROM_MINUS_999_TO_999`.

alexbishop1 (changes requested):
A couple of small things and an idea that should make this easier to test.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -81,4 +99,11 @@ private String getBrowserColorDepth(EpdqTemplateData templateData) {
         return templateData.getAuthCardDetails().getJsScreenColorDepth()
                 .filter(VALID_SCREEN_COLOR_DEPTHS::contains).orElse(DEFAULT_BROWSER_COLOR_DEPTH);
     }
+
+    private String getCurrentBrowserOffsetInMinutes() {
```
alexbishop1: This should have “default” somewhere in the name e.g. `getDefaultBrowserTimezoneOffsetInMinutes()`

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -64,9 +68,10 @@
     private static final String FRONTEND_URL = "http://www.frontend.example.com";
     private static final String ACCEPT_HEADER = "Test Accept Header";
     private static final String USER_AGENT_HEADER = "Test User Agent Header";
-
+    private static final Clock britishSummerTimeClock = Clock.fixed(ZonedDateTime.parse("2020-05-06T10:10:10.100Z").toInstant(), ZoneId.of("Europe/London"));
```
alexbishop1: Since this is a constant, we want `BRITISH_SUMMER_TIME_CLOCK` or similar.

We don’t actually need to use a `ZonedDateTime` here: `Instant` also has a `parse(…)` method so `Instant.parse("2020-05-06T10:10:10.100Z")` will have the same effect as `ZonedDateTime.parse("2020-05-06T10:10:10.100Z").toInstant()`.

Also, the time zone of the `Clock` probably shouldn’t be Europe/London. A bit confusing, I know, but connector always runs in UTC time, so the time zone of the real clock will be UTC. It doesn’t really matter though because `clock.instant()` returns an `Instant`, which is independent of time zones. Then we’ll use this instant to figure out what the current offset is for Europe/London. Confusing? Yes. Long and short of it is we should probably use `ZoneOffset.UTC` here.

Come to think of it, maybe the name should be `CLOCK_WITH_TIME_WHERE_UK_IS_IN_BST` or something.

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -169,12 +174,41 @@ public void should_include_browserColorDepth() {
...
+        authCardDetails.setJsTimezoneOffsetMins(timeZoneOffsetMins);
+        List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
+        assertThat(result, is(aParameterBuilder().build()));
+    }
+
+    @Test
+    @Parameters({"null", "-900", "800", "invalid", "0x6", "123L", "1.2"})
+    public void should_include_default_browserTimezoneOffsetMins_for_Gmt_when_timezone_offset_provided_is_invalid
+            (@Nullable String timeZoneOffsetMins) {
+        Clock gmtClock = Clock.fixed(ZonedDateTime.parse("2020-01-01T10:10:10.100Z").toInstant(), ZoneId.of("Europe/London"));
```
alexbishop1: This clock should be a constant like the other one (or the other one should be a local variable; I don’t mind as long as it’s consistent).

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -169,12 +174,41 @@ public void should_include_browserColorDepth() {
...
+        List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
+        assertThat(result, is(aParameterBuilder().withBrowserTimezoneOffsetMins("0").build()));
+    }
     
     static class ParameterBuilder {
         private String browserLanguage = "en-GB";
         private String browserScreenHeight = DEFAULT_BROWSER_SCREEN_HEIGHT;
         private String browserScreenWidth = DEFAULT_BROWSER_SCREEN_WIDTH;
         private String browserColorDepth = DEFAULT_BROWSER_COLOR_DEPTH;
+        private String browserTimezoneOffsetMins = "-60";
```
alexbishop1, line 212: It’s a bit confusing to have this as a default: `should_include_default_browserTimezoneOffsetMins_for_summer_time_when_timezone_offset_provided_is_invalid` should probably behave more similarly to `should_include_default_browserTimezoneOffsetMins_for_Gmt_when_timezone_offset_provided_is_invalid` where the default time zone offset to expect is explicitly specified in the test.

alexbishop1 (changes requested):
Thanks for introducing the `Clock`. Just a couple of things.

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -169,12 +175,40 @@ public void should_include_browserColorDepth() {
...
+    @Parameters({"-840", "720", "500"})
+    public void should_include_browserTimezoneOffSetMins(String timeZoneOffsetMins) {
+        authCardDetails.setJsTimezoneOffsetMins(timeZoneOffsetMins);
+        List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
+        assertThat(result, is(aParameterBuilder().withBrowserTimezoneOffsetMins(timeZoneOffsetMins).build()));
+    }
+
+    @Test
+    @Parameters({"null", "-900", "800", "invalid", "0x6", "123L", "1.2"})
+    public void should_include_default_browserTimezoneOffsetMins_for_summer_time_when_timezone_offset_provided_is_invalid
```
alexbishop1, line 189: I know `var epdqPayloadDefinitionFor3ds2NewOrder = new EpdqPayloadDefinitionForNew3ds2Order(FRONTEND_URL, language, BRITISH_SUMMER_TIME_OFFSET_CLOCK)` constructs the `EpdqPayloadDefinitionForNew3ds2Order` object with the appropriate `Clock` as a field but it would be good to have it explicitly here (like it is the other test) just to make it really clear what’s going on.

alexbishop1 (changes requested):
Looks good: just one small thing.

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

alexbishop1 (approved):

PR #2306, opened 2020-05-01
PP-6427 Add `account` property to Stripe Notification

kbottla:
## WHAT YOU DID
- Added `account` property which is included in notifications sent by Stripe.
  This info will only be available in notifications (ex: payouts) sent for a specific connect account.

with @sandorarpa


SandorArpa (dismissed):

sfount:
I think this PR is made against the wrong ticket, that might be worth clarifying

SandorArpa (approved):

PR #2317, opened 2020-05-05
PP-6362 Refactor ePDQ 3DS tests

nimalank7:
Description:
- Our current tests for 3DS do not have enough test coverage as they are service-level integration tests with no corresponding unit tests for EpdqPaymentProvider.
These service-level integration tests only test whether the GatewayOrder payload is correct regardless of what type of GatewayOrder it is. For example the
shouldRequire3dsAuthoriseRequest() test spins up a test gatewayaccount entity without 3DS enabled.
- This PR ensures that these service-level integration tests are spun up with test gateway accounts with 3DS and integration_version 1 enabled.
- A future PR will rename these tests as IT tests and create a separate class for unit testing EpdqPaymentProvider

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
@@ -40,6 +40,7 @@ import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import java.net.URI;
+import java.util.Map;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.Is.is;
@@ -113,7 +114,7 @@ public abstract class BaseEpdqPaymentProviderTest {
         when(gatewayConfig.getUrls()).thenReturn(ImmutableMap.of(TEST.toString(), "http://epdq.url"));
 
         when(configuration.getLinks()).thenReturn(linksConfig);
-        when(linksConfig.getFrontendUrl()).thenReturn("http://frontendUrl");
+        when(linksConfig.getFrontendUrl()).thenReturn("http://www.frontend.example.com");
 
         provider = new EpdqPaymentProvider(
                 configuration, 
@@ -139,6 +140,10 @@ public abstract class BaseEpdqPaymentProviderTest {
     void verifyPaymentProviderRequest(String requestPayload) {
         verify(mockClientInvocationBuilder).post(Entity.entity(requestPayload,
                 MediaType.APPLICATION_FORM_URLENCODED));
+    }
+    
+    String successAuth3dsRequest() {
+        return TestTemplateResourceLoader.load(TestTemplateResourceLoader.EPDQ_AUTHORISATION_3DS_REQUEST);
     }
 
     String successAuthRequest() {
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
```diff
@@ -140,6 +140,10 @@ void verifyPaymentProviderRequest(String requestPayload) {
         verify(mockClientInvocationBuilder).post(Entity.entity(requestPayload,
                 MediaType.APPLICATION_FORM_URLENCODED));
     }
+    
+    String successAuth3DSRequest() {
```
alexbishop1: Convention is usually that “3DS” becomes “3ds” when it’s in a camel case method or variable name.

Also, I think this method should probably be `private` rather than default access (I know it just follows similar existing methods; we can change those too).

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
```diff
@@ -229,13 +233,33 @@ private GatewayAccountEntity buildTestGatewayAccountEntity() {
         return gatewayAccount;
     }
 
+    private GatewayAccountEntity buildTestGatewayAccountWith3DSEntity() {
+        GatewayAccountEntity gatewayAccount = new GatewayAccountEntity();
+        gatewayAccount.setId(1L);
+        gatewayAccount.setGatewayName("epdq");
+        gatewayAccount.setRequires3ds(true);
+        gatewayAccount.setIntegrationVersion3ds(1);
+        gatewayAccount.setCredentials(ImmutableMap.of(
```
alexbishop1: We can use the built-in `Map.of(…)` here.

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
```diff
@@ -245,11 +269,19 @@ Auth3dsResponseGatewayRequest buildTestAuthorisation3dsVerifyRequest(String auth
...
     }
 
     RefundGatewayRequest buildTestRefundRequest() {
-        return buildTestRefundRequest(buildTestGatewayAccountEntity());
+        return buildTestRefundRequest(buildTestGatewayAccountWith3DSEntity());
+    }
+
+    private CardAuthorisationGatewayRequest buildTestAuthorisation3DSRequest(GatewayAccountEntity accountEntity) {
+        ChargeEntity chargeEntity = aValidChargeEntity()
+                .withExternalId("mq4ht90j2oir6am585afk58kml")
```
alexbishop1, line 282: Do we actually care about setting a specific external ID here?

alexbishop1 (changes requested):
Looks fine: just a couple of things.

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
nimalank7: I don't think this can be the case as these methods are inherited in the EpdqPaymentProvider3ds test and so can't be made private.

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderTest.java
nimalank7: No but the successAuth3dsRequest is pulling in EPDQ_AUTHORISATION_3DS_REQUEST template. This template is used in the EpdqPayloadDefinitionForNewOrderTest where it is needed. I decided to use this template for the EpdqPaymentProviderTest to reduce duplication as it seems unnecessary to generate a separate fixture for this.

alexbishop1 (approved):

PR #2319, opened 2020-05-06
PP-6362 - Send extra 3DS2 details to ePDQ

nimalank7:
Description:
- If the gateway account has 3DS2 enabled then it will send the extra parameter details across


src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -312,10 +314,18 @@ private GatewayOrder buildAuthoriseOrder(CardAuthorisationGatewayRequest request
...
         templateData.setAuthCardDetails(request.getAuthCardDetails());
+
+        EpdqPayloadDefinition epdqPayloadDefinition;
         
-        var epdqPayloadDefinition = request.getGatewayAccount().isRequires3ds() ?
-                        new EpdqPayloadDefinitionForNew3dsOrder(frontendUrl) :
-                        epdqPayloadDefinitionForNewOrder;
+        if (request.getGatewayAccount().isRequires3ds()) {
+            if (request.getGatewayAccount().getIntegrationVersion3ds() == 2) {
+                epdqPayloadDefinition = new EpdqPayloadDefinitionForNew3ds2Order(frontendUrl, request.getCharge().getLanguage(), Clock.systemUTC());
```
alexbishop1: Per its Javadoc, `Clock.systemUTC()` returns an object that’s immutable and thread-safe, which are my two favourite words in the Java language. So we could have a singleton `Clock` and pass that in to each new `EpdqPayloadDefinitionForNew3ds2Order` we instantiate.

To do this, add a method to the Guice `ConnectorModule`:

```java
@Provides
@Singleton
public Clock systemUtcClock() {
    return Clock.systemUTC();
}
```

Then just add a `Clock` parameter to the `EpdqPaymentProvider` constructor annotated with the `@Inject` annotation. Guice will find the `Clock` from the `@Provides` method in `ConnectorModule` and inject it. The body of the constructor can then assign this to a field.

The beauty of doing this is that it means that anywhere we have class with an `@Inject` constructor, we can add a `Clock` parameter and Guice will inject the singleton UTC `Clock`.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -312,10 +314,18 @@ private GatewayOrder buildAuthoriseOrder(CardAuthorisationGatewayRequest request
...
+            } else {
+                epdqPayloadDefinition = new EpdqPayloadDefinitionForNew3dsOrder(frontendUrl);
+            }
```
alexbishop1, line 328: Maybe a `switch` expression would work here?

```java
switch(request.getGatewayAccount().getIntegrationVersion3ds()) {
    case 2:
        epdqPayloadDefinition = new EpdqPayloadDefinitionForNew3ds2Order(frontendUrl, request.getCharge().getLanguage(), Clock.systemUTC());
        break;
    case 1:
    default:
        epdqPayloadDefinition = new EpdqPayloadDefinitionForNew3dsOrder(frontendUrl);
}
```

Not sure if it’s actually any better. I’ll leave it up to you.

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
```diff
@@ -250,10 +254,30 @@ private GatewayAccountEntity buildTestGatewayAccountWith3DSEntity() {
...
+        gatewayAccount.setIntegrationVersion3ds(2);
+        gatewayAccount.setCredentials(Map.of(
+                CREDENTIALS_MERCHANT_ID, "merchant-id",
+                CREDENTIALS_USERNAME, "username",
+                CREDENTIALS_PASSWORD, "password",
+                CREDENTIALS_SHA_IN_PASSPHRASE, "sha-passphrase"
+        ));
+        gatewayAccount.setType(TEST);
+        return gatewayAccount;
+    }
```
alexbishop1, line 260: Possibly this could share some code with `buildTestGatewayAccountWith3DSEntity()`?

(Also, as I’ve said before, it should really be `3ds` in a method name.)

src/test/resources/templates/epdq/authorisation-3ds2-request.txt
```diff
@@ -0,0 +1 @@
+ACCEPTURL=http%3A%2F%2Fwww.frontend.example.com%2Fcard_details%2Fmq4ht90j2oir6am585afk58kml%2F3ds_required_in%2Fepdq&AMOUNT=500&CARDNO=5555444433331111&CN=Mr.+Payment&CURRENCY=GBP&CVC=737&DECLINEURL=http%3A%2F%2Fwww.frontend.example.com%2Fcard_details%2Fmq4ht90j2oir6am585afk58kml%2F3ds_required_in%2Fepdq%3Fstatus%3Ddeclined&ED=08%2F18&EXCEPTIONURL=http%3A%2F%2Fwww.frontend.example.com%2Fcard_details%2Fmq4ht90j2oir6am585afk58kml%2F3ds_required_in%2Fepdq%3Fstatus%3Derror&FLAG3D=Y&HTTP_ACCEPT=text%2Fhtml&HTTP_USER_AGENT=Mozilla%2F5.0&LANGUAGE=en_GB&OPERATION=RES&ORDERID=mq4ht90j2oir6am585afk58kml&OWNERADDRESS=41%2C+Scala+Street&OWNERCTY=GB&OWNERTOWN=London&OWNERZIP=EC2A+1AE&PSPID=merchant-id&PSWD=password&USERID=username&WIN3DS=MAINW&browserColorDepth=24&browserLanguage=en-GB&browserScreenHeight=480&browserScreenWidth=320&browserTimezoneOffsetMins=-60&SHASIGN=06c7f863a28ddac84d525e30447dc650ad161d0924812bdeaed0e08f568ec021178636a3f7fab064c905b0a4d198430751271d75e758f5ef20ec195396efffda
```
alexbishop1: All the 3DS2 parameters are the defaults, aren’t they? Since we’ve decided that our testing is really an integration test, I think we might want to check actually setting the parameters on the `AuthCardDetails` to non-default values so that we check they actually eventually become the correct parameters in the payload.

However, testing the defaults does seem useful, so maybe we could have two tests: one where the parameters are set on `AuthCardDetails` and one where they’re not.

alexbishop1 (changes requested):
Looks promising but I have a few suggestions and I think we can make the test a bit more rigorous, as described below.

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
nimalank7: Not entirely sure what this adds so I'll leave it as it is.

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
nimalank7: Refactored

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
nimalank7: Done

src/test/resources/templates/epdq/authorisation-3ds2-request.txt
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
alexbishop1: Went to all the effort of setting up a Guice-provided `Clock` but we’re not actually using it!

We want to add the `Clock` to the constructor parameters of this class:

```java
@Inject
public EpdqPaymentProvider(ConnectorConfiguration configuration,
                           GatewayClientFactory gatewayClientFactory,
                           Environment environment,
                           EpdqPayloadDefinitionForCancelOrder epdqPayloadDefinitionForCancelOrder,
                           EpdqPayloadDefinitionForCaptureOrder epdqPayloadDefinitionForCaptureOrder,
                           EpdqPayloadDefinitionForNewOrder epdqPayloadDefinitionForNewOrder,
                           EpdqPayloadDefinitionForQueryOrder epdqPayloadDefinitionForQueryOrder,
                           Clock clock) {
```

… then pass it in to the `EpdqPayloadDefinitionForNew3ds2Order` constructor when we instantiate it:

```java
epdqPayloadDefinition = new EpdqPayloadDefinitionForNew3ds2Order(frontendUrl, request.getCharge().getLanguage(), clock);
```

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -32,6 +33,7 @@
     private final SupportedLanguage paymentLanguage;
     private final Clock clock;
 
+    @Inject
```
alexbishop1: Ah, I see what you were trying to do!

This object is just instantiated by our code using `new` when needed, so `@Inject` won’t work here (we need to inject it in to the `EpdqPaymentProvider` — which is Guice-managed — then pass it down).

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
```diff
@@ -270,20 +283,45 @@ Auth3dsResponseGatewayRequest buildTestAuthorisation3dsVerifyRequest(String auth
...
+    private CardAuthorisationGatewayRequest buildTestAuthorisation3ds2RequestWithProvidedParameters(GatewayAccountEntity accountEntity) {
+        ChargeEntity chargeEntity = aValidChargeEntity()
+                .withExternalId("mq4ht90j2oir6am585afk58kml")
+                .withGatewayAccountEntity(accountEntity)
+                .build();
+
+        AuthCardDetails authCardDetails = buildTestAuthCardDetails();
+        authCardDetails.setJsScreenColorDepth("15");
+        authCardDetails.setJsScreenHeight("1500");
+        authCardDetails.setJsScreenWidth("1500");
```
alexbishop1: Minor, but could you make the height and width different? It provides a little bit of protection against the values being accidentally transposed.

alexbishop1 (changes requested):
Looks good. Just a couple of small things.

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java

alexbishop1 (approved):

PR #2321, opened 2020-05-07
PP-6363 - Send HTTP Accept Header for authorisation for ePDQ 3DS2

nimalank7:
Description:
- Sends the HTTP Accept Header for authorisation and provides a default value if none provided

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
@@ -9,9 +9,11 @@ import java.time.ZoneId;
 import java.time.ZoneOffset;
 import java.util.List;
 import java.util.Locale;
+import java.util.Optional;
 import java.util.Set;
 import java.util.regex.Pattern;
 
+import static java.util.function.Predicate.not;
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqParameterBuilder.newParameterBuilder;
 
 public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionForNew3dsOrder {
@@ -21,6 +23,7 @@ public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionF
     public final static String BROWSER_SCREEN_HEIGHT = "browserScreenHeight";
     public final static String BROWSER_SCREEN_WIDTH = "browserScreenWidth";
     public final static String BROWSER_TIMEZONE_OFFSET_MINS = "browserTimezoneOffsetMins";
+    public final static String BROWSER_ACCEPT_HEADER = "browserAcceptHeader";
     public final static String DEFAULT_BROWSER_COLOR_DEPTH = "24";
     public final static String DEFAULT_BROWSER_SCREEN_HEIGHT = "480";
     public final static String DEFAULT_BROWSER_SCREEN_WIDTH = "320";
@@ -46,7 +49,8 @@ public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionF
                 .add(BROWSER_LANGUAGE, getBrowserLanguage(templateData))
                 .add(BROWSER_SCREEN_HEIGHT, getBrowserScreenHeight(templateData))
                 .add(BROWSER_SCREEN_WIDTH, getBrowserScreenWidth(templateData))
-                .add(BROWSER_TIMEZONE_OFFSET_MINS, getBrowserTimezoneOffsetMins(templateData));
+                .add(BROWSER_TIMEZONE_OFFSET_MINS, getBrowserTimezoneOffsetMins(templateData))
+                .add(BROWSER_ACCEPT_HEADER, getBrowserAcceptHeader(templateData));
 
         return parameterBuilder.build();
     }
@@ -104,5 +108,9 @@ public class EpdqPayloadDefinitionForNew3ds2Order extends EpdqPayloadDefinitionF
         int currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive = currentUkOffset.getTotalSeconds() / 60;
         int currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative = -currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive;
         return String.valueOf(currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative);
+    }
+
+    String getBrowserAcceptHeader(EpdqTemplateData templateData) {
+        return super.getBrowserAcceptHeader(templateData);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3dsOrder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3dsOrder.java
@@ -3,12 +3,13 @@ package uk.gov.pay.connector.gateway.epdq.payload;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.http.NameValuePair;
 import uk.gov.pay.connector.common.model.domain.Address;
+import uk.gov.pay.connector.gateway.epdq.EpdqTemplateData;
 import uk.gov.pay.connector.gateway.model.OrderRequestType;
 
 import java.util.List;
+import java.util.Optional;
 
-import uk.gov.pay.connector.gateway.epdq.EpdqTemplateData;
-
+import static java.util.function.Predicate.not;
 import static uk.gov.pay.connector.gateway.epdq.payload.EpdqParameterBuilder.newParameterBuilder;
 
 public class EpdqPayloadDefinitionForNew3dsOrder extends EpdqPayloadDefinitionForNewOrder {
@@ -18,11 +19,13 @@ public class EpdqPayloadDefinitionForNew3dsOrder extends EpdqPayloadDefinitionFo
     public static final String DECLINEURL_KEY = "DECLINEURL";
     public static final String EXCEPTIONURL_KEY = "EXCEPTIONURL";
     public static final String FLAG3D_KEY = "FLAG3D";
-    public static final String HTTPACCEPT_URL = "HTTP_ACCEPT";
+    public static final String HTTPACCEPT_KEY = "HTTP_ACCEPT";
     public static final String HTTPUSER_AGENT_URL = "HTTP_USER_AGENT";
     public static final String LANGUAGE_URL = "LANGUAGE";
     public static final String PARAMPLUS_URL = "PARAMPLUS";
     public static final String WIN3DS_URL = "WIN3DS";
+    
+    public static final String DEFAULT_BROWSER_ACCEPT_HEADER = "*/*";
     
     private final String frontendUrl;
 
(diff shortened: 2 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -105,4 +109,8 @@ private String getDefaultBrowserOffsetInMinutes() {
         int currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative = -currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive;
         return String.valueOf(currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative);
     }
+
+    private String getBrowserAcceptHeader(EpdqTemplateData templateData) {
+        return Optional.ofNullable(templateData.getAuthCardDetails().getAcceptHeader()).orElse(DEFAULT_BROWSER_ACCEPT_HEADER);
```
alexbishop1: We should also be using the `DEFAULT_BROWSER_ACCEPT_HEADER` if we the submitted one is the empty string.

alexbishop1 (changes requested):
Looks good apart from one small thing…

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
nimalank7: Good catch!

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java
```diff
@@ -105,4 +109,10 @@ private String getDefaultBrowserOffsetInMinutes() {
         int currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative = -currentUkOffsetMinsInJavaFormatWithAheadOfUtcPositive;
         return String.valueOf(currentUkOffsetMinsInJavaScriptFormatWithAheadOfUtcNegative);
     }
+
+    private String getBrowserAcceptHeader(EpdqTemplateData templateData) {
+        return Optional.ofNullable(templateData.getAuthCardDetails().getAcceptHeader())
+                .filter(header -> !header.equals(""))
```
alexbishop1: Another way to do this is to statically import `java.util.function.Predicate.not` and then do `filter(not(String::isEmpty))`.

src/test/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider3dsIT.java
```diff
@@ -33,6 +34,7 @@ public void shouldRequire3dsAuthoriseRequest() throws Exception {
     }
 
     @Test
+    @Ignore
```
alexbishop1, line 37: As agreed, these tests are being temporarily ignored until all the new parameters have been implemented. There’s no risk to do this because no-one is using 3DS2 with ePDQ yet.

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -203,13 +203,29 @@ public void should_include_browserTimezoneOffSetMins(String timeZoneOffsetMins)
...
+
+    @Test
+    public void should_include_accepted_browserAcceptHeader() {
+        authCardDetails.setAcceptHeader("text/html");
+        List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
+        assertThat(result, hasItem(new BasicNameValuePair(BROWSER_ACCEPT_HEADER, "text/html")));
+    }
+
+    @Test
+    @Parameters({"null", ""})
```
alexbishop1, line 217: I didn’t know  the string `"null"` would become an actual `null`!

alexbishop1 (approved):

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -253,7 +274,6 @@ public ParameterBuilder withBrowserTimezoneOffsetMins(String browserTimezoneOffs
                     .add(EXPIRY_DATE_KEY, END_DATE)
                     .add(EXCEPTIONURL_KEY, expectedFrontend3dsIncomingUrl + "?status=error")
                     .add(FLAG3D_KEY, "Y")
-                    .add(HTTPACCEPT_URL, ACCEPT_HEADER)
```
alexbishop1, line 256: Don’t remove this. We still want this to be sent.

alexbishop1 (changes requested):
Whoops! Scrolled too fast and missed an unwanted change!

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Yep hence the `@Nullable` annotation!

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: What's the difference between ACCEPT_HEADER and the BROWSER_ACCEPT_HEADER? Do I change it to `.add(HTTPACCEPT_URL, BROWSER_ACCEPT_HEADER)?`

alexbishop1: Basically, we have to send the `Accept` header twice, in two different parameters: `HTTP_ACCEPT` (which is required for all 3DS payments) and `browserAcceptHeader` (which is required for 3DS2 payments).

We should probably make it so that we send the same value for both (right now `HTTP_ACCEPT` doesn’t get the same defaulting to `*/*` behaviour that `browserAcceptHeader` does.

In the test, we should be checking for the presence of both of parameters.

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2Order.java

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3dsOrder.java
```diff
@@ -47,11 +46,16 @@ public EpdqPayloadDefinitionForNew3dsOrder(String frontendUrl) {
...
                 .add(HTTPUSER_AGENT_URL, templateData.getAuthCardDetails().getUserAgentHeader())
                 .add(LANGUAGE_URL, "en_GB")
                 .add(OPERATION_KEY, templateData.getOperationType())
                 .add(ORDER_ID_KEY, templateData.getOrderId());
+        
+        if(StringUtils.isBlank(templateData.getAuthCardDetails().getAcceptHeader())) {
+            epdqParameterBuilder.add(HTTPACCEPT_URL, "*/*");
+        } else {
+            epdqParameterBuilder.add(HTTPACCEPT_URL, templateData.getAuthCardDetails().getAcceptHeader());
+        }
```
alexbishop1: This should really use the same code as `getBrowserAcceptHeader()` in `EpdqPayloadDefinitionForNew3dsOrder`. You can probably extract the common stuff into a single method.

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -66,7 +68,7 @@
     private static final String PASSWORD = "password";
     private static final String USER_ID = "User";
     private static final String FRONTEND_URL = "http://www.frontend.example.com";
-    private static final String ACCEPT_HEADER = "Test Accept Header";
+    private static final String ACCEPT_HEADER = "*/*";
```
alexbishop1: Maybe we want this to not be the default one we use.

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -203,13 +205,29 @@ public void should_include_browserTimezoneOffSetMins(String timeZoneOffsetMins)
...
+        List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
+        assertThat(result, is(aParameterBuilder().withBrowserAcceptHeader("text/html").build()));
+    }
+
+    @Test
+    @Parameters({"null", ""})
+    public void browserAcceptHeader_should_include_default_if_browser_accept_header_not_provided(@Nullable String browserAcceptHeader) {
+        authCardDetails.setAcceptHeader(browserAcceptHeader);
+        List<NameValuePair> result = epdqPayloadDefinitionFor3ds2NewOrder.extract(epdqTemplateData);
+        assertThat(result, is(aParameterBuilder().build()));
```
alexbishop1: Could you explicitly do `assertThat(result, is(aParameterBuilder().withBrowserAcceptHeader(EpdqPayloadDefinitionForNewOrderTest.DEFAULT_BROWSER_ACCEPT_HEADER).build()));` so it’s clear we expect the default to be used?

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
```diff
@@ -253,7 +276,7 @@ public ParameterBuilder withBrowserTimezoneOffsetMins(String browserTimezoneOffs
                     .add(EXPIRY_DATE_KEY, END_DATE)
                     .add(EXCEPTIONURL_KEY, expectedFrontend3dsIncomingUrl + "?status=error")
                     .add(FLAG3D_KEY, "Y")
-                    .add(HTTPACCEPT_URL, ACCEPT_HEADER)
+                    .add(HTTPACCEPT_URL, browserAcceptHeader)
```
alexbishop1: While you’re here, could you rename `HTTPACCEPT_URL` to `HTTP_ACCEPT_KEY`? The current name looks like a copypasta that doesn’t make much sense.

alexbishop1 (changes requested):
Getting there! Just a couple of things…

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3ds2OrderTest.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForNew3dsOrder.java
nimalank7: Done I put it in the super class and invoked the super class method

alexbishop1 (approved):

PR #2323, opened 2020-05-07
PP-6477 Fix Refund availability for historic payments

kbottla:
## WHAT 
- We currently calculate refund available in 2 ways
  1. For in-flight (charge still in connector) payments, refund availability is calculated based on     internal charge status. Internal statuses considered are also slightly to EPDQ compared to the      rest of the payment providers. i.e., for EPDQ, CAPTURED_SUBMITTED is refundable and pending for others
  2. Historic payments (payments expunged and only exists in ledger)
     So far, for historic payments `refundAvailability` is set based on ledger refund availability which is always `available` (or the latest `refundAvailability` that Ledger holds),  but this should be calculated based on charge status and current list of refunds.

Changes addresses the issue with historic payments and aims to match the functionality of calculating refundAvailability of in-flight payments

As ledger doesn't store equivalent of connector internal status, we are now relying on
 1. `captured_date` from Ledger : (for all payment providers) to calculate amount refundable. This matches current functionality of `CAPTURED` state
 2. `capture_submitted_date` : For EPDQ, refunds are allowed immediately when charge is 
 submitted for capture. So refundable amount is calculated for EPDQ if capture submitted date is available.
 3. For any in-progress external states, refundAvailability is `pending`. We should not have any historic payments that are expunged from connector but still in progress

All other states or if connector couldn't map external state from Ledger for any reason, refundAvailability will be `unavailable`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/util/DefaultExternalRefundAvailabilityCalculator.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/util/DefaultExternalRefundAvailabilityCalculator.java
@@ -2,14 +2,15 @@ package uk.gov.pay.connector.gateway.util;
 
 import com.google.common.collect.ImmutableList;
 import uk.gov.pay.connector.charge.model.domain.Charge;
-import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.util.RefundCalculator;
 import uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability;
+import uk.gov.pay.connector.common.model.api.ExternalChargeState;
 import uk.gov.pay.connector.refund.model.domain.RefundEntity;
 
 import java.util.List;
 
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_READY;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_READY;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/util/DefaultExternalRefundAvailabilityCalculator.java
```diff
@@ -46,9 +53,20 @@
     );
     private static final List<ChargeStatus> STATUSES_THAT_MAP_TO_EXTERNAL_AVAILABLE_OR_EXTERNAL_FULL = ImmutableList.of(CAPTURED);
 
+    private static final List<ExternalChargeState> EXTERNAL_STATUSES_THAT_MAP_TO_REFUND_EXTERNAL_PENDING = ImmutableList.of(
```
sfount, line 56: I'm hesitant to introduce a third level of mapping after we've already translated from Connector to Ledger and back again, I think this will make it particularly difficult to reason about the criteria for this process. 

I propose that we simplify this based on the following assumptions: 

1. something can't be expunged if it's not terminal, we have very strict checks in place to make sure this is the case
2. the information Connector has about refund database entities is up to date at the time of refund
3. if we use the current refundability status from Ledger there are only three options:
  a. the refundability is set to "full" or "unavailable", we can't reach this code
  b. the refundability is is set to "available" and the ledger record is consistent, we can trust it
  c. the refundability is set to "available" but the ledger record is not yet consistent (two refunds processed directly in a row)

In order to guard against `3.c`, moving the abstraction down to this class and always calculating refund-ability is great idea, I think we can just simplify it to say that if everything has happened to get the refund to this point (see assumptions) then the only thing left is to make sure there's an up to date available balance. 

The proposed algorithm: 
``` 
if (charge.isHistoric()) {
  // we would actually call the same method, calculateRefundAvailability
  if (amountAvailableToBeRefunded > 0) {
      return EXTERNAL_AVAILABLE;
  } else {
      return EXTERNAL_FULL;
  }
} else {
  // do what we've always done
}

sfount (commented):
The investigation and write up look great, some questions about the approach that we can pair on as discussed.

src/main/java/uk/gov/pay/connector/gateway/util/DefaultExternalRefundAvailabilityCalculator.java
kbottla: As discussed, using refundability from ledger can simplify things, but could add 
- unexpected behaviour if Ledger is not upto date i.e. if the events are not processed
 by Ledger yet, it would not have latest refundability which could let connector take a different step than expected.
- and this closely ties refundability to every piece of code that emits a terminal state and for every manual change (updating to terminal state) we should also calculate refundability. This could add some overhead if we are updating ledger manually and if not automated.

On the other hand external status never changes. One other thing to consider is that the logic can be simplified later to only use external statuses and remove reliance on Charge internal statuses so we don't end up with 2 different paths forever.

PR #2328, opened 2020-05-11
PP-6363 - Send browserJavaEnabled parameter to ePDQ 3DS2

nimalank7:
Description:
- Sends the browserJavaEnabled parameter to ePDQ 3DS


alexbishop1 (changes requested):
Seems to be some confusion here. The IP address (`REMOTE_ADDR`) should only be sent if the `send_payer_ip_address_to_gateway` flag is `true`. However, `browserJavaEnabled` should always be sent with a value of `false`.

alexbishop1:
Turns out the confusion was entirely caused by my making a copy-paste mistake with the acceptance criteria in the story. Sorry!

alexbishop1 (approved):

PR #2334, opened 2020-05-12
PP-6470 Retrieve balance transfers for payout from Stripe

stephencdaly:
When we process a PayoutReconcileMessage, make a request to Stripe to
retrieve all BalanceTransactions for the payout.

Add skeleton structure for processing of Payments and Refunds included
in the transfer. We will reconcile these and emit events to ledger in
PP-6472.

Add mocks to the existing happy path test to verify the payout is
processed as expected.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
@@ -7,6 +7,7 @@ import java.util.Map;
 
 public class StripeCredentials {
 
+    public static final String STRIPE_ACCOUNT_ID = "stripe_account_id";
     private String accountId;
 
     public StripeCredentials(@JsonProperty("stripe_account_id") String accountId) {
@@ -14,6 +15,6 @@ public class StripeCredentials {
     }
 
     public Map<String, String> toMap() {
-        return ImmutableMap.of("stripe_account_id", accountId);
+        return ImmutableMap.of(STRIPE_ACCOUNT_ID, accountId);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
@@ -7,14 +7,14 @@ import java.util.Map;
 
 public class StripeCredentials {
 
-    public static final String STRIPE_ACCOUNT_ID = "stripe_account_id";
+    public static final String STRIPE_ACCOUNT_ID_KEY = "stripe_account_id";
     private String accountId;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
```diff
@@ -7,13 +7,14 @@
 
 public class StripeCredentials {
 
+    public static final String STRIPE_ACCOUNT_ID = "stripe_account_id";
     private String accountId;
 
     public StripeCredentials(@JsonProperty("stripe_account_id") String accountId) {
```
sfount: Is it reasonable to set this `@JsonProperty` to `STRIPE_ACCOUNT_ID` as well, it's unlikely but the json property defines what's written into `jsonb` in the database and if these two get out of line looking up accounts by this won't work. 

Also I think `STRIPE_ACCOUNT_ID_KEY` would be more specific

src/main/java/uk/gov/pay/connector/payout/StripeClientWrapper.java
```diff
@@ -0,0 +1,24 @@
+package uk.gov.pay.connector.payout;
+
+import com.stripe.exception.StripeException;
+import com.stripe.model.BalanceTransaction;
+import com.stripe.net.RequestOptions;
+
+import java.util.List;
+import java.util.Map;
+
+class StripeClientWrapper {
```
sfount, line 10: Really clean abstraction 👍

sfount (dismissed):
Looks like a great start 👍 some minor comments

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
stephencdaly: Yeh good point, made this change

sfount (approved):

src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
```diff
@@ -42,4 +98,11 @@ public void processPayouts() throws QueueException {
             }
         }
     }
+
+    private String getStripeApiKey(String stripeAccountId) {
```
kbottla, line 102: We might have to keep this on par with https://github.com/alphagov/pay-connector/blob/1bd16bb4f6f5a6be81b5e23ab5b5b77857c1cb35/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java#L29 or use one place for this. Version is important in case Stripe SDK uses latest version by default which could be incompatibile with our expectation

stephencdaly: Good point, I'll look into this

PR #2340, opened 2020-05-13
PP-6475 Fix: use `ResourceType` enum when removing emitted events

sfount:
Previous code removed emitted events during the expunge process with the
condition `resource_type='PAYMENT'`. Records are written to the database
with the values `payment` or `refund` and so this criteria was never met
and emitted events were not removed as part of the expunge process.

This was covered by a test and story review but these also used `"PAYMENT"`
in line with the code change (resulting in a false positive as no
results didn't mean anything had been deleted).

Use the `ResourceType` enum that is used for writing to the database to
ensure consistency.


kbottla (approved):

stephencdaly:
:oof:

PR #2344, opened 2020-05-14
PP-6469 Add PayoutCreated event

kbottla:
## WHAT YOU DID
- Added new payout created event with details to emit to ledger for PAYOUT_CREATED notifications from Stripe and relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripePayout.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripePayout.java
@@ -4,6 +4,10 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.databind.PropertyNamingStrategy;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
 
+import java.time.ZonedDateTime;
+
+import static uk.gov.pay.connector.util.DateTimeUtils.toUTCZonedDateTime;
+
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
 public class StripePayout {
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripePayout.java
```diff
@@ -12,6 +16,21 @@
...
+    String type;
+    String statementDescriptor;
+
+    public StripePayout() {
+    }
+
+    public StripePayout(String id, Long amount, Long arrivalDate, Long created, String status, String type, String statementDescriptor) {
+        this.id = id;
+        this.amount = amount;
+        this.arrivalDate = arrivalDate;
```
sfount, line 28: From the Stripe docs:
> Date the payout is expected to arrive in the bank. This factors in delays like weekends or bank holidays.

This will go into the payout details JSON column which seems like a good fit.

We have a top level column in Ledger called `paid_out_date` which sounds more like it would be set when the `payout.paid` event is received, I've noted that for the ticket dealing with those webhooks.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripePayout.java
```diff
@@ -12,6 +16,21 @@
...
+        this.created = created;
+        this.status = status;
```
sfount, line 30: It might be useful to call this `gateway_status` or something to differentiate it from our `state` which is controlled by the Ledger consumer. 

This `status` won't actually drive our understanding of the payout state (which will be done through salient events). 

The renaming would clarify that when looking in the details.

sfount (commented):
LGTM :+1: clear and well covered with tests - question about the name of `status` which will go into the payout details object

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripePayout.java
kbottla: Good point! Will rename to gateway_status. Also renaming type, arrival_date to destination_type, estimated_bank_arrival_date to make these clearer when only looking into ledger

kbottla: Updated https://github.com/alphagov/pay-connector/compare/663615b3afe66396dc575f1ea0a2718e9f28de36..872922df4afa24aed594e185e71fe02624b2eacc

sfount (approved):
Great clarifications 👍

PR #2346, opened 2020-05-14
PP-6475 Add swap table to take over for emitted events

sfount:
An issue with expunging emitted events from Connector was recently
addressed in https://github.com/alphagov/pay-connector/pull/2340. This
issue means that the `emitted_events` table has a large number of rows
that need to be safely removed.

The first step in the process outlined in this ticket is to create an
identical table that live Connector nodes can use to track emitted
events for in-flight payments.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1470,4 +1470,8 @@
         <dropColumn tableName="refunds_history" columnName="charge_id"/>
     </changeSet>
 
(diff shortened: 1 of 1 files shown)


sfount:
Verified in psql 
Screenshot 2020-05-14 at 18 44 26
Screenshot 2020-05-14 at 18 44 40

kbottla (approved):

PR #2350, opened 2020-05-15
PP-6475 Record when exceptions are thrown during backfill process

sfount:
We are silently consuming any exceptions thrown during the backfill
process when sweeping non-emitted events. Generally exceptions are
handled at the actual point of event emission (an unrelated process)
but there are a number of cases where this would happen before that 
(charge doesn't exist given inconsistent data etc.)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
+++ b/src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
@@ -98,6 +98,16 @@ public class EmittedEventsBackfillService {
             charge.ifPresent(c -> MDC.put("chargeId", c.getExternalId()));
             charge.ifPresent(historicalEventEmitter::processPaymentAndRefundEvents);
             event.setEmittedDate(now(ZoneId.of("UTC")));
+        } catch (Exception e) {
+            logger.error(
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
```diff
@@ -98,6 +98,15 @@ public void backfillEvent(EmittedEventEntity event) {
             charge.ifPresent(c -> MDC.put("chargeId", c.getExternalId()));
             charge.ifPresent(historicalEventEmitter::processPaymentAndRefundEvents);
             event.setEmittedDate(now(ZoneId.of("UTC")));
+        } catch (Exception e) {
+            logger.error(
```
kbottla, line 102: Looks fine but catching exception here means task will end up with HTTP status 200 instead of 500 / something else. May be throw exception or send relevant HTTP status if there is any exception ?

sfount: Good call 👍 I don't think anything consumes the HTTP response but making sure something gets back would be a good idea, done in https://github.com/alphagov/pay-connector/commit/1a9ad2cf4be9507cbb41ccc3c82906552696bf4a

kbottla (approved):

PR #2356, opened 2020-05-18
PP-6472 Use serializers from pay-java-commons

stephencdaly:
Use MicrosecondPrecisionDateTimeSerializer and MicrosecondPrecisionDateTimeDeserializer from pay-java-commons


sfount (approved):
Great cleanup 👍

PR #2358, opened 2020-05-18
PP-6364 Make ePDQ 3DS1 “contract” integration tests pass

alexbishop1:
Make the “contract” (with the payment provider) integration tests for ePDQ that actually talk to ePDQ send the correct card number to simulate a requires 3DS response when necessary.


nimalank7 (approved):
LGTM!

PR #2359, opened 2020-05-18
PP-6469 Add gateway_account_id to PayoutCreated event

kbottla:


src/main/java/uk/gov/pay/connector/payout/PayoutEmitterService.java
```diff
@@ -42,14 +51,18 @@ private void sendToEventQueue(Event event) {
...
     }
 
-    private Event getPayoutEvent(Class eventClass, String connectAccount, StripePayout payout) {
+    private Event getPayoutEvent(Class<? extends PayoutEvent> eventClass, String connectAccount, StripePayout payout) {
         if (eventClass == PayoutCreated.class) {
-            Event event = PayoutCreated.from(payout);
-            return event;
+            GatewayAccountEntity gatewayAccount = gatewayAccountDao.findByCredentialsKeyValue(
+                    STRIPE_ACCOUNT_ID_KEY, connectAccount)
+                    .orElseThrow(() -> new GatewayAccountNotFoundException(
```
stephencdaly: Is this going to cause us to respond with a 400 to the notification? Is this what we want to do?

kbottla: yeah we respond with a 500 status. We should not have any prod accounts (receiving payouts) with connect account not in gateway accounts. So this is very unlikely to happen and may be fine for now. We probably need to have another look at payout exception handling for when to alert or let Stripe try notification again.

sfount: I think it's okay to respond with a non `200 OK` to this, it will raise visibility for it happening. For us to be getting notifications about unknown accounts in production something will have gone very wrong.

stephencdaly: Should we log an error so it gets into Sentry too? I'm not sure it's going to be that visible otherwise?

kbottla: Agree! Updated to log an error.

stephencdaly (approved):
Looks good

PR #2360, opened 2020-05-18
PP-6472 Emit events for transactions included in payout

stephencdaly:
When we process PayoutReconcileMessages from the SQS queue, emit events to ledger for each payment and refund included in a payout.

Events we emit are PAYMENT_INCLUDED_IN_PAYOUT and REFUND_INCLUDED_IN_PAYOUT. These both contain a `gateway_payout_id` field.

Only emit events if the feature flag is switched on using the existing EMIT_PAYOUT_EVENTS environment variable.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
+++ b/src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
@@ -101,7 +101,7 @@ public class PayoutReconcileProcess {
                     payoutReconcileQueue.markMessageAsProcessed(payoutReconcileMessage.getQueueMessage());
                 }
             } catch (Exception e) {
-                LOGGER.warn("Error processing payout from SQS message [queueMessageId={}] [errorMessage={}]",
+                LOGGER.error("Error processing payout from SQS message [queueMessageId={}] [errorMessage={}]",
                         payoutReconcileMessage.getQueueMessageId(),
                         e.getMessage()
                 );
@@ -113,7 +113,7 @@ public class PayoutReconcileProcess {
         return gatewayAccountDao.findByCredentialsKeyValue(StripeCredentials.STRIPE_ACCOUNT_ID_KEY, stripeAccountId)
                 .map(gatewayAccountEntity ->
                         gatewayAccountEntity.isLive() ? stripeGatewayConfig.getAuthTokens().getLive() : stripeGatewayConfig.getAuthTokens().getTest())
-                .orElseThrow(() -> new GatewayAccountNotFoundException(format("Gateway account with Stripe connect account ID %s not found.", stripeAccountId)));
+                .orElseThrow(() -> new RuntimeException(format("Gateway account with Stripe connect account ID [%s] not found.", stripeAccountId)));
     }
 
     private void emitPaymentEvent(PayoutReconcileMessage payoutReconcileMessage, BalanceTransaction balanceTransaction) {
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
```diff
@@ -84,7 +89,7 @@ public void processPayouts() throws QueueException {
                         });
 
                 if (payments.intValue() == 0 && refunds.intValue() == 0) {
-                    LOGGER.error("No payments or refunds retrieved for payout [{}] for connect account [{}]. Requires investigation.", 
+                    LOGGER.error("No payments or refunds retrieved for payout [{}] for connect account [{}]. Requires investigation.",
```
sfount, line 92: It's not actually part of this PR but should the top level process `catch` also log at an `error` level? Is failing to process something we want to actively get in Sentry or just be able to monitor over time

sfount: This line in question
```
LOGGER.warn("Error processing payout from SQS message [queueMessageId={}] [errorMessage={}]",
```

sfount (dismissed):
Really clear code and well covered by tests 👍 looks good to me

src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
```diff
@@ -110,4 +115,45 @@ private String getStripeApiKey(String stripeAccountId) {
...
+                        paymentExternalId, payoutReconcileMessage.getGatewayPayoutId(), e.getMessage()), e);
+            }
+        }
+
+        LOGGER.info("Emitted event for payment [{}] included in payout [{}]", paymentExternalId, payoutReconcileMessage.getGatewayPayoutId());
+    }
+
+    private void emitRefundEvent(PayoutReconcileMessage payoutReconcileMessage, BalanceTransaction balanceTransaction) {
+        var sourceTransfer = (Transfer) balanceTransaction.getSourceObject();
+        var metadata = StripeTransferMetadata.from(sourceTransfer.getMetadata());
```
sfount, line 145: Looks like this deserialise will happily add `nulls` if the metadata `Map<String, String>` isn't formatted correctly (if it didn't have the govuk transaction ID in there).

It might be worth throwing an exception in `from` or at this method level if the transaction ID can't be found (is `null`), that way we don't accidentally associate invalid data.

src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
stephencdaly: I could have sworn I'd made this change in an earlier PR. I will do it for real and not just in my head. Good spot

src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
stephencdaly: I've pushed 2 new commits to address this and do a bit of cleanup. Sorry if this PR is getting a bit big now

sfount: Looks great 👍

src/test/java/uk/gov/pay/connector/payout/PayoutReconcileProcessTest.java
```diff
@@ -108,15 +147,78 @@ public void shouldNotMarkMessageAsSuccessfullyProcessedIfNoPaymentsOrRefundsFoun
...
+        when(refundBalanceTransaction.getSourceObject()).thenReturn(refundTransferSource);
+        when(refundTransferSource.getMetadata()).thenReturn(Map.of());
+
+        when(stripeClientWrapper.getBalanceTransactionsForPayout(eq(payoutId), eq(stripeAccountId), eq(stripeApiKey)))
+                .thenReturn(List.of(refundBalanceTransaction));
+
+        payoutReconcileProcess.processPayouts();
+
+        verify(logAppender).doAppend(loggingEventArgumentCaptor.capture());
+        assertThat(loggingEventArgumentCaptor.getValue().getFormattedMessage(), containsString("Transaction external ID missing in metadata"));
```
sfount, line 193: Well covered 👍

sfount (approved):

PR #2362, opened 2020-05-19
PP-6486 Add payout notification types

kbottla:
## WHAT YOU DID
- Added additional payout notification types we are expected to receive from Stripe
- Also introduced `eventClass` to notificationType, so notification specific event can be derived and sent to event queue.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
@@ -12,7 +12,7 @@ import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
-import uk.gov.pay.connector.events.model.payout.PayoutCreated;
+import uk.gov.pay.connector.events.model.payout.PayoutEvent;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.model.Auth3dsResult;
 import uk.gov.pay.connector.gateway.stripe.json.StripePaymentIntent;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationType.java
```diff
@@ -30,6 +42,10 @@ public String getType() {
         return type;
     }
 
+    public Class<? extends PayoutEvent> getEventClass() {
```
stephencdaly: Should this return an optional as there won't always be an event class?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -119,19 +119,23 @@ public void handleNotificationFor(String payload, String signatureHeader) {
...
+            StripeNotificationType stripeNotificationType = byType(notification.getType());
 
-            sendToPayoutReconcileQueue(notification.getAccount(), payout);
+            if (PAYOUT_CREATED.equals(stripeNotificationType)) {
+                Payout payout = new Payout(stripePayout.getId(), notification.getAccount(), stripePayout.getCreated());
+                sendToPayoutReconcileQueue(notification.getAccount(), payout);
+            }
 
-            payoutEmitterService.emitPayoutEvent(PayoutCreated.class, notification.getAccount(), stripePayout);
+            payoutEmitterService.emitPayoutEvent(stripeNotificationType.getEventClass(),
```
stephencdaly: even though through the current code path we only get PAYOUT_CREATED events here, should we protect against an EventClass not being returned for the notification type as not all notification types will have one?

kbottla: Getting null here is fine as PayoutEmitterService ignores unknown eventClasses including null
https://github.com/alphagov/pay-connector/blob/3493bf18be98e6673b07f8c7afd9cfee00c1dfa3/src/main/java/uk/gov/pay/connector/payout/PayoutEmitterService.java#L59

Now that eventClass returned as optional, this is handled

stephencdaly: ah 👍

stephencdaly (dismissed):

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationType.java
kbottla: PayoutEmitterService (which is what used this return) ignores null. Changed to Optional return to be defensive for future changes

stephencdaly (approved):

PR #2377, opened 2020-05-26
BAU — Simply method to check if card brand requires 3DS

alexbishop1:
- Remove code that filters all returned `CardBrandEntity` objects so that we’re only left with ones where the brand matches the argument: the DAO already does this for us.
- Don’t unnecessarily collect the stream into a list only to immediately create another stream from that.


rorymalcolm (approved):
a great contribution on the road to 0 lines of code

PR #2386, opened 2020-06-01
PP-6564 Emit additional payout events for payouts with terminal status

kbottla:
## WHAT YOU DID
- StripeNotificationService emits all (except PAYOUT_CREATED) payout events to payment event queue as soon as webhook notification is received from Stripe. So if for any reason event cannot be emitted StripeNotificationService, a webhook need to be triggered manually from Stripe dashboard.
- This change adds a mechanism to also emit additional events (when emitting PAYOUT_CREATED event) if payout status is terminal.
  So instead of using Stripe dashboard to resend notifications, we can emit an event to
  PayoutReconcileQueue which will complete the payout lifecycle (with emitting additional payout events) in ledger


SandorArpa (approved):
LOKTM

PR #2387, opened 2020-06-01
PP-6565 Use structured arguments for payout logging

stephencdaly:
Extract the connect account id, gateway payout id, ledger event type
and payment and refund external ids into structured arguments.

This makes constructing searches in Splunk to link together events
easier.


SandorArpa (approved):
LOKTM

PR #2404, opened 2020-06-22
PP-6680 Include tag of Docker base image in FROM line in Dockerfile

alexbishop1:
This should help Dependabot to automatically find updates.


SandorArpa (approved):
LOKTM

PR #2416, opened 2020-07-01
PP-6712 EpdqSha512SignatureGenerator: be smarter with streams

alexbishop1:
Be smarter with streams in `EpdqSha512SignatureGenerator` to avoid both an intermediate list and an intermediate `StringJoiner`.


kbottla (approved):
LGTM!

PR #2417, opened 2020-07-01
PP-6712 Make subclasses of EpdqPayloadDefinition not be singletons

alexbishop1:
Make subclasses of `EpdqPayloadDefinition` not be singletons for they will soon hold state (except for
`EpdqPayloadDefinitionForNew3dsOrder` and `EpdqPayloadDefinitionForNew3ds2Order`, which already aren’t singletons and hold state).

Fun fact: despite being annotated with `@Singleton`, `EpdqPayloadDefinitionForRefundOrder` wasn’t a singleton.


SandorArpa (approved):
LOKTM

PR #2421, opened 2020-07-02
PP-6712 Remove SHA-IN passphrase from EpdqTemplateData

alexbishop1:
Move the SHA-IN passphrase from being a field of `EpdqTemplateData` to being a field of `EpdqPayloadDefinition`.

This means that only the `extract()` method on `EpdqPayloadDefinition` now accesses the `EpdqTemplateData` inside the `EpdqPayloadDefinition`.


stephencdaly (approved):
Looks like a sensible refactor

PR #2423, opened 2020-07-02
PP-6712 Don’t use EpdqTemplateData for maintenance requests

alexbishop1:
When building a payload to cancel, capture or refund an ePDQ payment, don’t use `EpdqTemplateData`. Instead, set the data directly on the `EpdqPayloadDefinition` subclass object.

Push some stuff down from `EpdqPayloadDefinitionForMaintenanceOrder` to the appropriate subclass to better reflect reality. For example, a refund is the only maintenance order that can have an amount, so only it needs an amount field. Similarly, a cancel is the only maintenance order that either has a `PAYID` or an `ORDERID`, so only it needs the logic for dealing with this. After pushing  these things down, EpdqPayloadDefinitionForMaintenanceOrder isn’t really pulling its weight, so delete it and have its subclasses directly extend `EpdqPayloadDefinition` instead.


src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForCancelOrder.java
```diff
@@ -1,8 +1,52 @@
...
+import org.apache.http.NameValuePair;
 import uk.gov.pay.connector.gateway.model.OrderRequestType;
 
-public class EpdqPayloadDefinitionForCancelOrder extends EpdqPayloadDefinitionForMaintenanceOrder {
+import java.util.List;
+import java.util.Optional;
+
+import static uk.gov.pay.connector.gateway.epdq.payload.EpdqParameterBuilder.newParameterBuilder;
+
+public class EpdqPayloadDefinitionForCancelOrder extends EpdqPayloadDefinition {
```
danworth, line 11: I'm not familiar with these `EpdqPayloadDefinition` classes but curious as to your thoughts on wether its practical for them to be made immutable with a builder pattern rather than using the various `set` methods?

danworth (approved):
Looks good

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForCancelOrder.java
alexbishop1: They essentially are builders (and I might rename them as such). The `createGatewayOrder()` method is analogous to a builder `build()` method. There’s some weirdness like there being a protected `extract()` method that’s called by the `createGatewayOrder()` method, which I may or may not get time to address before the week is out.

PR #2428, opened 2020-07-03
PP-6712 Don’t use EpdqTemplateData for authorisation requests

alexbishop1:
When building a payload to authorise an ePDQ payment, don’t use `EpdqTemplateData`. Instead, set the data directly on the
`EpdqPayloadDefinition` subclass object.

Delete `EpdqTemplateData` as it’s no longer used anywhere.


src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqPaymentProvider.java
```diff
@@ -292,15 +292,6 @@ private GatewayOrder buildQueryOrderRequestFor(ChargeEntity charge) {
     }
 
     private GatewayOrder buildAuthoriseOrder(CardAuthorisationGatewayRequest request, String frontendUrl) {
-        EpdqTemplateData templateData = new EpdqTemplateData();
-        templateData.setOrderId(request.getChargeExternalId());
-        templateData.setPassword(request.getGatewayAccount().getCredentials().get(CREDENTIALS_PASSWORD));
-        templateData.setUserId(request.getGatewayAccount().getCredentials().get(CREDENTIALS_USERNAME));
-        templateData.setMerchantCode(request.getGatewayAccount().getCredentials().get(CREDENTIALS_MERCHANT_ID));
-        templateData.setDescription(request.getDescription());
```
kbottla, line 300: Don't we need `description`  ?

alexbishop1: It never gets sent to ePDQ so we don’t need it for constructing payloads.

kbottla (approved):

PR #2429, opened 2020-07-03
PP-6712 Remove getOperationType() from EpdqPayloadDefinition

alexbishop1:
Remove `getOperationType()` from `EpdqPayloadDefinition` and its subclasses because it’s only used internally by the same class and doesn’t need to exposed.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForCaptureOrder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForCaptureOrder.java
@@ -33,17 +33,12 @@ public class EpdqPayloadDefinitionForCaptureOrder extends EpdqPayloadDefinition 
     @Override
     public List<NameValuePair> extract() {
         return newParameterBuilder()
-                .add("OPERATION", getOperationType())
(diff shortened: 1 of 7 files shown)


kbottla (approved):

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForCaptureOrder.java
```diff
@@ -33,19 +33,14 @@ public void setPassword(String password) {
     @Override
     public List<NameValuePair> extract() {
         return newParameterBuilder()
-                .add("OPERATION", getOperationType())
+                .add("OPERATION", "SAS")
```
sfount, line 36: Any reason these don't use the `OPERATION_KEY` type constants used for other payloads?

sfount (approved):

src/main/java/uk/gov/pay/connector/gateway/epdq/payload/EpdqPayloadDefinitionForCaptureOrder.java
alexbishop1: Probably no good reason. It was like that when I got here.

PR #2432, opened 2020-07-06
BAU — Include gateway name in failure to query payment status error

alexbishop1:
Make the “Unable to retrieve status for charge x: yyy” message be “Unable to retrieve status for charge x for gateway y: zzz” because the problem may well be at the gateway’s end and it’s useful to know.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/QueryService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/QueryService.java
@@ -8,10 +8,15 @@ import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.ChargeQueryResponse;
+import uk.gov.pay.logging.LoggingKeys;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/QueryService.java
```diff
@@ -34,7 +34,8 @@ public boolean canQueryChargeGatewayStatus(PaymentGatewayName paymentGatewayName
         try {
             return getChargeGatewayStatus(charge).getMappedStatus();
         } catch (WebApplicationException | UnsupportedOperationException | GatewayException | IllegalArgumentException e) {
-            logger.info("Unable to retrieve status for charge {}: {}", charge.getExternalId(), e.getMessage());
+            logger.info("Unable to retrieve status for charge {} for gateway {}: {}", charge.getExternalId(), charge.getPaymentGatewayName().toString(),
+                    e.getMessage());
```
stephencdaly: could we make this log line have structured key/values while you're here?

alexbishop1: Yeah, OK

stephencdaly (approved):
nice

PR #2436, opened 2020-07-08
PP-6688 Split VatNumberCompanyNumber task

kbottla:
## WHAT YOU DID
 - Splits `VAT_NUMBER_COMPANY_NUMBER` to individual tasks `VAT_NUMBER` and `COMPANY_NUMBER` as we will be removing `check your answers` page in selfservice. By removing check your answers page we are intending to submit `vat_number` as soon as form is submitted and `company_number` separately.

   This change will enable us to direct service users to specific task due (vat_number or company_number) during their onboarding flow.
- `VAT_NUMBER_COMPANY_NUMBER` task still remains and to be removed once selfservice stops using this task


stephencdaly (approved):
Looks good

PR #2443, opened 2020-07-16
PP-6756 Add gateway_transaction_id to refunds payload

kbottla:
## WHAT
  - We store gateway reference for refunds in `reference` column. This contradicts with payments as   `gateway_transaction_id` (charges table) is used to store payment reference provided by PSPs.  Additionally `reference` on payment is the service provided value which further adds confusion to using the name reference on refunds.

**Changes**
   - Starts setting gateway_transaction_id on refunds table. `reference` is still set on refunds table but soon to be dropped.
   - Adding gateway_transaction_id (reference in refunds table) to refund events sent to Ledger. This will make the gateway_transaction_id naming consistent with payment event. `reference` will be removed from refund events payload once we fully switch over to using gateway_transaction_id column.

with @sandorarpa


SandorArpa (approved):

PR #2448, opened 2020-07-17
PP-6614 - Add empty string for last four digits to Sandbox for Wallet Payments

nimalank7:
Description:
- The lastFourDigitsCardNumber is not used for authorisation purposes by Apple and Google Pay it is purely cosmetic. However in our tests
Sandbox determines the authorisation status based on the lastFourDigitsCardNumber field which is not ideal. To ensure our pact test passes (see PP-6614 in frontend) we have added an empty string to
Sandbox allowed card numbers
- We add a check for whether lastFourDigitsCardNumber is of length 4 to handle the case of an empty string as otherwise an Exception will be thrown. This is to fix a bug as lastFourDigitsCardNumber isn't used for authorisation so upon successful authorisation the contents of lastFourDigitsCardNumber shouldn't matter.
- Apple/Google Pay don't use the last_four_digits as part of authorisation but we do for producing canned responses for authorisation. To better test Apple/Google Pay the test numbers have been moved to a separate class to ensure separation of concerns.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -745,8 +745,11 @@ public class ChargeService {
         if (hasFullCardNumber(authCardDetails)) { // Apple Pay etc. don’t give us a full card number, just the last four digits here
             detailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of(StringUtils.left(authCardDetails.getCardNo(), 6)));
         }
-        detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
-
+        
+        if (hasLastFourCharactersCardNumber(authCardDetails)) {
+            detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
+        }
+        
         if (authCardDetails.getAddress().isPresent())
             detailsEntity.setBillingAddress(new AddressEntity(authCardDetails.getAddress().get()));
 
@@ -757,6 +760,10 @@ public class ChargeService {
 
     private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
         return authCardDetails.getCardNo().length() > 6;
+    }
+    
+    private boolean hasLastFourCharactersCardNumber(AuthCardDetails authCardDetails) {
+        return authCardDetails.getCardNo().length() >= 4;
     }
 
     private TokenEntity createNewChargeEntityToken(ChargeEntity chargeEntity) {
--- a/src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
@@ -1,34 +1,24 @@
 package uk.gov.pay.connector.gateway.sandbox;
 
-import com.google.common.collect.ImmutableSet;
-
 import java.util.Map;
 import java.util.Set;
-import java.util.function.Function;
-import java.util.stream.Stream;
 
-import static java.util.stream.Collectors.toMap;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ERROR;
 
 public class SandboxCardNumbers {
-
+    
     private static final String GOOD_CARD_PREPAID_NON_CORPORATE = "4000160000000004";
-    private static final String GOOD_WALLET_LAST_DIGITS_CARD_NUMBER = "4242";
     private static final String GOOD_CORPORATE_PREPAID_UNKNOWN_CREDIT_CARD = "5101180000000007";
     private static final String GOOD_CORPORATE_PREPAID_UNKNOWN_DEBIT_CARD = "5200828282828210";
     private static final String GOOD_NON_CORPORATE_NON_PREPAID = "4000020000000000";
     private static final String GOOD_MASTERCARD_CREDIT_CARD = "5101110000000004";
     private static final String GOOD_VISA_PREPAID_UNKNOWN_CREDIT_OR_DEBIT_UNKNOWN_CARD = "4000000000000010";
-    private static final String DECLINED_WALLET_LAST_DIGITS_CARD_NUMBER = "0002";
     private static final String DECLINED_CARD_NUMBER = "4000000000000002";
-    private static final String CVC_ERROR_WALLET_LAST_DIGITS_CARD_NUMBER = "0127";
     private static final String CVC_ERROR_CARD_NUMBER = "4000000000000127";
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -758,6 +761,10 @@ private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails
     private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
         return authCardDetails.getCardNo().length() > 6;
     }
+    
+    private boolean hasLastFourCharactersCardNumber(AuthCardDetails authCardDetails) {
+        return authCardDetails.getCardNo().length() >=4;
```
alexbishop1: The lack of a space between `>=` and `4` annoys me to an unusual degree!

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
```diff
@@ -11,8 +11,13 @@
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ERROR;
 
 public class SandboxCardNumbers {
+    /*
+    The last four digits field for Apple and Google Pay aren't used for authorisation.
+    However for testing purposes Sandbox determines the authorisation status based on the last four digits field.
+     */
 
     private static final String GOOD_CARD_PREPAID_NON_CORPORATE = "4000160000000004";
+    private static final String GOOD_WALLET_EMPTY_LAST_DIGITS_CARD_NUMBER = "";
```
alexbishop1: I’m not sure this is the best place for this because it’s not restricted to just Apple Pay and Google Pay.

Fortunately, we have `SandboxWalletAuthorisationHandler`:

```java
public class SandboxWalletAuthorisationHandler implements WalletAuthorisationHandler, SandboxGatewayResponseGenerator {
    @Override
    public GatewayResponse<BaseAuthoriseResponse> authorise(WalletAuthorisationGatewayRequest request) {
        String lastDigitsCardNumber = request.getWalletAuthorisationData().getLastDigitsCardNumber();
        return getSandboxGatewayResponse(lastDigitsCardNumber);
    }

}
```

We could put the check in there:

```java
public class SandboxWalletAuthorisationHandler implements WalletAuthorisationHandler, SandboxGatewayResponseGenerator {

    @Override
    public GatewayResponse<BaseAuthoriseResponse> authorise(WalletAuthorisationGatewayRequest request) {
        String lastDigitsCardNumber = request.getWalletAuthorisationData().getLastDigitsCardNumber();
        if (lastDigitsCardNumber.isEmpty()) {
            getSandboxGatewayResponse(true);
        }
        return getSandboxGatewayResponse(lastDigitsCardNumber);
    }

}
```

This would entail changing the visibility of the `getSandboxGatewayResponse(boolean)` method but I think that’s OK.

alexbishop1 (changes requested):
I think we might be able to approach this in another way that’s a bit more limited in terms of its effect on what sandbox considers to be valid card numbers.

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
nimalank7: Sure. I don't think we need to change the visibility as it's a static method and the interface is implemented so we can just call `SandboxGatewayResponseGenerator.getSandboxGatewayResponse`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Ha fixed

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxCardNumbers.java
nimalank7: Come to think of it why don't we move all the Apple/Google Pay numbers into that class?

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxWalletCardNumbers.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+
+    private static final String GOOD_WALLET_LAST_DIGITS_CARD_NUMBER = "4242";
+    private static final String GOOD_WALLET_EMPTY_STRING_CARD_NUMBER = "";
+    private static final String DECLINED_WALLET_LAST_DIGITS_CARD_NUMBER = "0002";
+    private static final String CVC_ERROR_WALLET_LAST_DIGITS_CARD_NUMBER = "0127";
+    private static final String EXPIRED_WALLET_LAST_DIGITS_CARD_NUMBER = "0069";
+    private static final String PROCESSING_ERROR_WALLET_LAST_DIGITS_CARD_NUMBER = "0119";
+
+    private static final Set<String> GOOD_CARDS = ImmutableSet.of(
```
alexbishop1: We can use `Set.of(…)` here

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxWalletCardNumbers.java
```diff
@@ -0,0 +1,57 @@
...
+    private static final Set<String> REJECTED_CARDS = ImmutableSet.of(
+            DECLINED_WALLET_LAST_DIGITS_CARD_NUMBER,
+            EXPIRED_WALLET_LAST_DIGITS_CARD_NUMBER,
+            CVC_ERROR_WALLET_LAST_DIGITS_CARD_NUMBER);
+
+    private static Map<String, CardError> ERROR_CARDS = Stream.of(
+            PROCESSING_ERROR_WALLET_LAST_DIGITS_CARD_NUMBER)
+            .collect(toMap(
+                    Function.identity(),
+                    cardNumber -> new CardError(AUTHORISATION_ERROR, "This transaction could be not be processed.")));
```
alexbishop1: This could just be `Map.of(PROCESSING_ERROR_WALLET_LAST_DIGITS_CARD_NUMBER, new CardError(AUTHORISATION_ERROR, "This transaction could be not be processed."))`

alexbishop1 (changes requested):
This looks like it will do the trick. Just a couple of small requests.

Also, could you update the commit message to better reflect the approach we’re taking now?

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxWalletCardNumbers.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/sandbox/SandboxWalletCardNumbers.java
nimalank7: Done

alexbishop1 (dismissed):
This is nicer than what we had before.

alexbishop1 (approved):

PR #2449, opened 2020-07-19
PP-6756 Remove reference on Refunds

kbottla:
## WHAT YOU DID
- Removed `reference` field on Refunds/RefundHistory tables and its usages. Uses gatewayTransactionId instead which is now being set on refunds.


SandorArpa (approved):
LOKTM

PR #2450, opened 2020-07-20
PP-6756 Remove reference column on refund tables

kbottla:
## WHAT
- Removes `reference` column on refund & refund_history tables as we started using `gateway_transaction_id` through out code base and reference is no longer needed

Depends on https://github.com/alphagov/pay-connector/pull/2449


SandorArpa (approved):
LOKTM

PR #2451, opened 2020-07-20
WIP Stripe auth handles non card errors

sfount:
POST to Stripe authorisation will reject with `4xx` for both card
valid card declines and corrupt requests. All of the codes listed in
https://stripe.com/docs/api/errors have never fired other than
`card_error` and `validation_error` when there is a problem with the
Platform accounts.

By default we should therefore have immediate alerts around anything
that isn't being treated as a `card_error`, this case will usually
require us to intervene.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
@@ -71,8 +71,20 @@ public class StripeAuthoriseHandler implements AuthoriseHandler {
 
             if (e.getFamily() == CLIENT_ERROR) {
                 StripeErrorResponse stripeErrorResponse = jsonObjectMapper.getObject(e.getResponseFromGateway(), StripeErrorResponse.class);
-                logger.info("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeAuthoriseHandler.java
```diff
@@ -71,8 +71,20 @@ public GatewayResponse authorise(CardAuthorisationGatewayRequest request) {
...
-                        chargeExternalId, stripeErrorResponse.getError().getCode(), stripeErrorResponse.getError().getMessage(), e.getStatus());
+
+                switch(stripeErrorResponse.getError().getCode()) {
+                    /**
+                     * Stripe clients throw well formed errors for different scenarios, as we are directly POSTing, handle
+                     * the card_error token as business as usual and alert about everything else.
+                     * https://stripe.com/docs/api/errors#errors-validation_error
+                     */
+                    case "card_error":
+                        logger.info("Authorisation failed for charge {}. Failure code from Stripe: {}, failure message from Stripe: {}. Response code from Stripe: {}",
```
alexbishop1, line 82: I think “Authorisation rejected” would be clearer: “failed” is a bit ambiguous as to whether it’s an expected or unexpected failure.

sfount: Oh just noticed this is commented on the existing log message (which is just moved), agreed with the comment but we would probably have to verify upstream to make sure no one is checking for this

sfount:
Superseded by https://github.com/alphagov/pay-connector/pull/2464

PR #2452, opened 2020-07-20
PP-6756 Drop gateway_processing_details column on refunds

kbottla:
## WHAT YOU DID
 - Drops `gateway_processing_details` column from `refunds` table as it is never used and not required at this time.

with @sandorarpa


SandorArpa (approved):
LOKTM

PR #2464, opened 2020-07-28
PP-6852 Categorise Stripe authorisation client error as REJECTED/ERROR

kbottla:
## WHAT YOU DID
- Currently we treat all 4xx error types as REJECTED, but it is only `card_error` type that is expected for payments rejected by card issuer.
  Rest of the 4xx error types (like api_error, rate_limit_error) are unexpected and should be treated as ERROR.

**Changes** 
- Only `card_error` type is now treated as AUTHORISATION_REJECTED for 4xx status and rest of error types will be marked as AUTHORISATION_ERROR

- Also improved toString() representation for `StripeAuthorisationFailedResponse`. It is current logged with default toString() method (which appears like `StripeAuthorisationFailedResponse@36e182ac`) in the upstream which is not helpful.


stephencdaly (approved):
Looks good

PR #2487, opened 2020-08-13
PP-6883 Add Notify Base URL

rorymalcolm:
- To follow the convention set by adminnusers, we want to configure the
notify base url parameter in the manifest.yml instead of the env-map,
this means that we only have to configure a shared variable like this in
one place per environment.


danworth (approved):
Makes sense 👍

PR #2490, opened 2020-08-16
PP-6916 IGNORE: Raising to check concourse

danworth:
Will use to measure build time and worker utilisation for java app.

## WHAT YOU DID
No op change to test run time and worker utilisation


SandorArpa:
I think we can safely close this PR.

PR #2494, opened 2020-08-17
PP-6256 - Add address_state_province column to charges database

nimalank7:
Description:
- As part of PP-6256 to determine the US state or Canadian province from the postal code we need to add a new column to the database

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1490,4 +1490,11 @@
         <dropColumn tableName="refunds" columnName="gateway_processing_details"/>
     </changeSet>
 
+    <changeSet id="add address_state_province column to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="address_state_province" type="varchar(2)">
+            </column>
+        </addColumn>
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1490,4 +1490,17 @@
...
+            <column name="address_state_province" type="varchar(2)">
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="update address_state_province column to null" author="">
+        <update tableName="charges">
+            <column name="address_state_province" value="null"/>
+        </update>
+    </changeSet>
```
alexbishop1: Shouldn’t need this: all the existing rows will get a `NULL` value for that column.

alexbishop1 (changes requested):

src/main/resources/migrations.xml
nimalank7: Done

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #2495, opened 2020-08-17
PP-6256 - Add addressStateProvince field to ChargeEntity

nimalank7:
Description:
- Since the database migration in PR 2494 has been run we can now add the field to the JPA entity

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/AddressEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/AddressEntity.java
@@ -22,6 +22,8 @@ public class AddressEntity {
     private String county;
     @Column(name = "address_country")
     private String country;
+    @Column(name = "address_state_province")
+    private String stateOrProvince;
 
     public AddressEntity() {
         //for jpa
@@ -70,6 +72,14 @@ public class AddressEntity {
 
     public String getCountry() {
         return country;
+    }
+
+    public String getStateOrProvince() {
+        return stateOrProvince;
+    }
+
+    public void setStateOrProvince(String stateOrProvince) {
+        this.stateOrProvince = stateOrProvince;
     }
 
     public String getCounty() {


oswaldquek (dismissed):

alexbishop1 (changes requested):
This belongs inside `AddressEntity` with the other cardholder address properties.

src/main/java/uk/gov/pay/connector/charge/model/AddressEntity.java
```diff
@@ -22,6 +22,8 @@
     private String county;
     @Column(name = "address_country")
     private String country;
+    @Column(name = "address_state_province")
+    private String addressStateProvince;
```
alexbishop1: The field name can just be `stateProvince` (or maybe `stateOrProvince`): we don’t need the `address` prefix because it’s obvious from the context of the class.

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -161,7 +161,7 @@
 
     @Column(name = "moto")
     private boolean moto;
-
+
```
alexbishop1: I think all the changes to this file can be reverted.

alexbishop1 (changes requested):
Almost there!

src/main/java/uk/gov/pay/connector/charge/model/AddressEntity.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
nimalank7: Done

alexbishop1 (approved):

PR #2501, opened 2020-08-19
PP-6256 - Add US states and Canadian provinces and territories mapping

nimalank7:
Description:
- This gives the mapping from the various ZIP and postal codes to the relevant US and Canadian provinces.
- Since a later PR will turn on the actual mapping in ChargeService it merely runs through the code and calculates the provinces but doesn't do anything with them
- The Utils have been made static so that we can simply call the method
- I have excluded the Santa postcode. I didn't do any further granular analysis of Canadian post codes as there are far too many exceptions in say the 'A' category which made it unfeasible. So even if 'A6...' is an unassigned postcode the code will assign it to the relevant territory/province.

## How to test
- Run the Utils test


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -750,8 +752,11 @@ private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails
             detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
         }
         
-        if (authCardDetails.getAddress().isPresent())
-            detailsEntity.setBillingAddress(new AddressEntity(authCardDetails.getAddress().get()));
+        if (authCardDetails.getAddress().isPresent()) {
+            AddressEntity addressEntity = new AddressEntity(authCardDetails.getAddress().get());
+            processRegionForCanadaOrUs(addressEntity);
+            detailsEntity.setBillingAddress(addressEntity);
+        }
```
alexbishop1: We can use `ifPresent(…)` rather than `isPresent()` here:

```java
if (authCardDetails.getAddress().ifPresent(address -> {
        AddressEntity addressEntity = new AddressEntity(address);
        processRegionForCanadaOrUs(addressEntity);
        detailsEntity.setBillingAddress(addressEntity);
});

File: src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java [hunk 2]
```diff
@@ -765,7 +770,26 @@ private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
     private boolean hasLastFourCharactersCardNumber(AuthCardDetails authCardDetails) {
         return authCardDetails.getCardNo().length() >= 4;
     }
-
+    
+    private void processRegionForCanadaOrUs(AddressEntity addressEntity) {
+        String country = addressEntity.getCountry().toLowerCase();
```
alexbishop1: We don’t actually need to do this for the reasons described below.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -765,7 +770,26 @@ private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
...
+        String normalisedPostalCode = addressEntity.getPostcode().replaceAll("\\s", "");
```
alexbishop1: It would be more efficient to only do this once we know we’re dealing with a US or Canadian address.

Also, we should upper-case the postcode here, using `toUpperCase(Locale.ENGLISH)`.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -765,7 +770,26 @@ private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
...
+        switch (country) {
+            case "america":
+            case "united states":
+            case "united states of america":
+            case "us":
+            case "usa":
```
alexbishop1: So it’s not at all obvious at this point in the code but `addressEntity.getCountry()` will always give you an ISO 3166-1 alpha-2 country code, so we only need to check for “US” and “CA”.

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritory.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.connector.util;
+
+public class CanadaProvinceTerritory {
```
alexbishop1: This would work great as an enum.

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
```diff
@@ -0,0 +1,71 @@
+package uk.gov.pay.connector.util;
+
+import java.util.AbstractMap;
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+public class CanadaProvinceTerritoryUtils {
```
alexbishop1: `Utils` is so generic. I’d call this `CanadianPostcodeToProvinceOrTerritoryMapper` or something.

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private final static Pattern postalAddress = Pattern.compile("[A-Z][0-9][A-Z][0-9][A-Z][0-9]");
```
alexbishop1: Use `SCREAMING_SNAKE_CASE` for constant names.

alexbishop1: I suppose `[A-WY-Z][0-9][A-Z][0-9][A-Z][0-9]` would exclude an initial X and make the two regexen mutually exclusive.

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
```diff
@@ -0,0 +1,71 @@
...
+    private final static Pattern xPostalAddress = Pattern.compile("X[0-9][A-Z][0-9][A-Z][0-9]");
+    
+    public static final Map<String, CanadaProvinceTerritory> NON_X_POSTAL_CODE_TERRITORY_PROVINCE_MAP = Map.ofEntries(
+            new AbstractMap.SimpleEntry<>("A", provinceTerritoryOf("NL", "Newfoundland and Labrador")),
```
alexbishop1: We can use `Map.entry(…)` in place of `new AbstractMap.SimpleEntry<>(…)` to make it a lot shorter (and we can even static import `Map.entry` to make it even shorter).

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
```diff
@@ -0,0 +1,71 @@
...
+            new AbstractMap.SimpleEntry<>("B", provinceTerritoryOf("NS", "Nova Scotia")),
+            new AbstractMap.SimpleEntry<>("C", provinceTerritoryOf("PE", "Prince Edward Island")),
+            new AbstractMap.SimpleEntry<>("E", provinceTerritoryOf("NB", "New Brunswick")),
+            new AbstractMap.SimpleEntry<>("G", provinceTerritoryOf("QC", "Quebec")),
+            new AbstractMap.SimpleEntry<>("H", provinceTerritoryOf("QC", "Quebec")),
```
alexbishop1: If we use an enum for provinces and territories as suggested above, we can just reference that rather than repeatedly stating that Quebec’s abbreviation is QC.

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
```diff
@@ -0,0 +1,71 @@
...
+    private static Optional<CanadaProvinceTerritory> getXPostalStateTerritory(String normalisedPostalCode) {
+        String firstThreeLetters = normalisedPostalCode.substring(0, 3);
+        return Optional.ofNullable(X_POSTAL_CODE_TERRITORY_MAP.get(firstThreeLetters));
+        
+    }
+    
+    private static Optional<CanadaProvinceTerritory> getPostalStateTerritory(String normalisedPostalCode) {
+        String firstLetter = normalisedPostalCode.substring(0, 1);
+        
+        if ("H0H0H0".equals(normalisedPostalCode)) {
```
alexbishop1: Should probably extract “H0H0H0” to a constant with a meaningful name like `SANTA_CLAUS_POSTAL_CODE` or `I_BET_WHEN_CANADA_POST_WERE_BEING_CUTE_GIVING_SANTA_HIS_OWN_POSTAL_CODE_THEY_DID_NOT_CONSIDER_COMPUTER_PROGRAMMERS_WOULD_HAVE_TO_SPECIAL_CASE_IT`.

src/main/java/uk/gov/pay/connector/util/UsState.java
```diff
@@ -0,0 +1,60 @@
+package uk.gov.pay.connector.util;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class UsState {
```
alexbishop1: I think it would be cleanest to separate out the domain object that represents a US state from the ZIP codes that match it (much like you’ve done for Canada).

src/main/java/uk/gov/pay/connector/util/UsState.java
```diff
@@ -0,0 +1,60 @@
...
+    public String getFullStateName() {
+        return fullStateName;
+    }
+
+    public Set<String> getZipCodesNotInUse() {
+        return zipCodesNotInUse;
+    }
+    
+    private boolean zipCodeInRange(String zipCode, int zipCodeInteger) {
+        return zipCodeInteger <= getHighestZipCode() && zipCodeInteger >= getLowestZipCode();
```
alexbishop1: This all feels quite complicated. Honestly, I think having a thousand-line file that maps each three-digit code to a state would be the simplest thing to do. (If we grab the page source of https://en.wikipedia.org/wiki/List_of_ZIP_Code_prefixes it should be fairly easy to use some text editor magic to generate most of the file for us.)

alexbishop1 (changes requested):
This looks like it’s along the right lines. I’ve left some suggestions.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaProvinceTerritoryUtils.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsState.java
nimalank7: Done - I replaced it with an enum following what I did with Canada

src/main/java/uk/gov/pay/connector/util/UsState.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -765,7 +771,22 @@ private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
     private boolean hasLastFourCharactersCardNumber(AuthCardDetails authCardDetails) {
         return authCardDetails.getCardNo().length() >= 4;
     }
-
+    
+    private void processRegionForCanadaOrUs(AddressEntity addressEntity) {
+        String country = addressEntity.getCountry();
```
alexbishop1: I’d just inline this into the `switch` statement.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -765,7 +771,22 @@ private boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
...
+        String normalisedPostalCode = addressEntity.getPostcode().replaceAll("\\s", "").toUpperCase(Locale.ENGLISH);
```
alexbishop1: It would be more efficient to only do this when we know the country is the US or Canada (we can extract it into a private method to avoid duplicating the code).

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
+package uk.gov.pay.connector.util;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+import static java.util.Map.entry;
+
+public class CanadaPostcodeToProvinceOrTerritoryMapper {
```
alexbishop1: It turns out in Canada they’re call “postal codes” rather than “postcodes” so to be pedantic, this should really be `CanadaPostalCodeToProvinceOrTerritoryMapper`.

Sorry.

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
...
+    
+    private final static Pattern POSTAL_ADDRESS = Pattern.compile("[A-Z][0-9][A-Z][0-9][A-Z][0-9]");
```
alexbishop1: This is really a `WELL_FORMED_POSTAL_CODE`.

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
...
+    private final static Pattern X_POSTAL_ADDRESS = Pattern.compile("X[0-9][A-Z][0-9][A-Z][0-9]");
+    private final static String SANTA_CLAUS_POSTAL_CODE = "H0H0H0";
```
alexbishop1: 🎅🏻

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
...
+            entry("K", CanadaProvinceTerritory.ONTARIO),
+            entry("L", CanadaProvinceTerritory.ONTARIO),
+            entry("M", CanadaProvinceTerritory.ONTARIO),
+            entry("N", CanadaProvinceTerritory.ONTARIO),
+            entry("P", CanadaProvinceTerritory.ONTARIO),
+            entry("R", CanadaProvinceTerritory.MANITOBA),
+            entry("S", CanadaProvinceTerritory.SASKATCHEWAN),
+            entry("T", CanadaProvinceTerritory.ALBERTA),
+            entry("V", CanadaProvinceTerritory.BRITISH_COLOMBIA),
+            entry("Y", CanadaProvinceTerritory.YUKON)
```
alexbishop1: I’d static import `CanadaProvinceTerritory.NEWFOUNDLAND_AND_LABRADOR` and friends to make this a bit cleaner.

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
...
+        String firstLetter = normalisedPostalCode.substring(0, 1);
+        
+        if (SANTA_CLAUS_POSTAL_CODE.equals(normalisedPostalCode)) {
+            return Optional.empty();
+        }
+        
+        return Optional.ofNullable(NON_X_POSTAL_CODE_TERRITORY_PROVINCE_MAP.get(firstLetter));
+    }
+
+    public enum CanadaProvinceTerritory {
```
alexbishop1: I think this can live in its own file.

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
...
+        NEW_BRUNSWICK("NB", "New Brunswick"),
+        QUEBEC("QC", "Quebec"),
+        ONTARIO("ON", "Ontario"),
+        MANITOBA("MB", "Manitoba"),
+        SASKATCHEWAN("SK", "Saskatchewan"),
+        ALBERTA("AB", "Alberta"),
+        BRITISH_COLOMBIA("BC", "British Colombia"),
+        YUKON("YT", "Yukon"),
+        NUNAVUT("NU", "Nunavut"),
+        NORTHWEST_TERRITORIES("NT", "Northwest Territories");
```
alexbishop1: Can we put these in some sort of logical order?

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
...
+
+        private final String abbreviation;
+        private final String fullName;
+
+        CanadaProvinceTerritory(String abbreviation, String fullStateName) {
```
alexbishop1: It’s a province or territory, not a state!

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,101 @@
...
+            this.abbreviation = abbreviation;
+            this.fullName = fullStateName;
```
alexbishop1: Although it’s strictly unnecessary, if you do this:

```java
this.abbreviation = Objects.requireNonNull(abbreviation);
this.fullName = Objects.requireNonNull(fullStateName);
```

… it makes it really obvious to the reader that neither value can be null.

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
```diff
@@ -0,0 +1,946 @@
+package uk.gov.pay.connector.util;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class UsStateMapping {
```
alexbishop1: Maybe `UsZipCodeToStateMapper` for consistency?

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
```diff
@@ -0,0 +1,946 @@
...
+    
+    /**
+    We have to use this initializiation method because it causes a compiler error otherwise. See https://youtrack.jetbrains.com/issue/IDEA-245963
```
alexbishop1: We can be a bit more explicit here and say “Map.ofEntries(…) with hundreds of arguments is excessively slow to compile” or similar.

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
```diff
@@ -0,0 +1,946 @@
...
+     */
+
+    public final static Map<String, String> ZIP_CODE_TO_US_STATE_ABBREVIATIONS = new HashMap<>() {{
```
alexbishop1: We can wrap this whole thing in `Map.copyOf(…)` to make it unmodifiable.

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
```diff
@@ -0,0 +1,946 @@
...
+        put("005", "NY");
```
alexbishop1: Might be more flexible if we mapped to a `UsState` rather than an abbreviation string.

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
```diff
@@ -0,0 +1,168 @@
+package uk.gov.pay.connector.util;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+import static java.util.Map.entry;
+
+public class UsStateUtils {
```
alexbishop1: I think this can probably be merged into `UsStateMapping`.

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
```diff
@@ -0,0 +1,168 @@
...
+
+    private final static Pattern stateWithZipCode = Pattern.compile("[A-Z]{2}[0-9]{5}");
```
alexbishop1: 🐍

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
```diff
@@ -0,0 +1,168 @@
...
+    private final static Pattern stateWithZipCodeFourPlus = Pattern.compile("[A-Z]{2}[0-9]{5}-[0-9]{4}");
+    private final static Pattern zipCode = Pattern.compile("[0-9]{5}");
+    private final static Pattern zipCodeFourPlus = Pattern.compile("[0-9]{5}-[0-9]{4}");
```
alexbishop1: It’s called ZIP+4 so `ZIP_PLUS_FOUR` would be a more appropriate name.

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
```diff
@@ -0,0 +1,168 @@
...
+    
+    private final static Map<String, UsState> US_STATE_ABBREVIATIONS = Map.ofEntries(
```
alexbishop1: I think if we make `ZIP_CODE_TO_US_STATE_ABBREVIATIONS` map to a `UsState`, we won’t need this.

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+    
+    private static Optional<UsState> getStateFromZipCodeAndState(String zipCodeWithState) {
+        String initialTwoChars = zipCodeWithState.substring(0, 2);
+        return Optional.ofNullable(US_STATE_ABBREVIATIONS.get(initialTwoChars));
+    }
+    
+    private static Optional<UsState> getStateFromZipCode(String zipCode) {
+        Optional<String> stateAbbreviation = Optional.ofNullable(UsStateMapping.ZIP_CODE_TO_US_STATE_ABBREVIATIONS.get(zipCode));
+        return stateAbbreviation.map(state -> US_STATE_ABBREVIATIONS.get(state));
```
alexbishop1:
```java
return stateAbbreviation.map(US_STATE_ABBREVIATIONS::get);
```

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
```diff
@@ -0,0 +1,168 @@
...
+        NEW_JERSEY("NJ", "New Jersey"),
+        ARMED_FORCES_EUROPE("AE", "Armed Forces Europe"), // Double check this
+        PENNSYLVANIA("PA", "Pennsylvania"),
+        DELAWARE("DE", "Delaware"),
+        WASHINGTON_DC("DC", "Washington DC"), // Double check this
+        VIRGINIA("VA", "Virginia"),
+        MARYLAND("MD", "Maryland"),
+        WEST_VIRGINIA("WV", "West Virginia"),
+        NORTH_CAROLINA("NC", "North Carolina"),
+        SOUTH_CAROLINA("NC", "South Carolina"),
```
alexbishop1: Don’t conflate the Carolinas!

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
```diff
@@ -0,0 +1,168 @@
...
+        ARIZONA("AZ", "Arizona"),
+        NEW_MEXICO("NM", "New Mexico"),
+        NEVADA("NV", "Nevada"),
+        CALIFORNIA("CA", "California"),
+        ARMED_FORCES("AP", "Armed Forces"),
+        HAWAII("HI", "Hawaii"),
+        GUAM("GU", "Guam"),
+        OREGON("OR", "Oregon"),
+        WASHINGTON("WA", "Washington"),
+        ALASKA("AK", "Alaska");
```
alexbishop1: Order seems a bit arbitrary here. Can we order them alphabetically or by order of admission to the union or something?

alexbishop1 (changes requested):
This is getting there. I made a few suggestions.

Can we also think about an appropriate package to put all this stuff in rather than just `uk.gov.pay.connector.util`? Maybe `uk.gov.pay.connector.address` or something?

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
nimalank7: We do need this because we want to be able to go from a state abbreviation in the ZIP code to the US state.

alexbishop1: Ah, I see. But we should be able to avoid explicitly listing the associations between the abbreviations and the states…

```java
private static final Map<String, UsState> US_STATE_ABBREVIATIONS = Arrays.stream(UsState.values()).collect(toUnmodifiableMap(UsState::getAbbreviation, identity()));
```

(Need to static import `java.util.stream.Collectors.toUnmodifiableMap` and ` java.util.function.Function.identity` for that to work.)

nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/CanadaPostcodeToProvinceOrTerritoryMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsStateMapping.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
nimalank7: Perhaps but I think that would end up cluttering the class and make it a little unreadable. It's easier to just have it in its own file.

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/util/UsStateUtils.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,83 @@
...
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.ALBERTA;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.BRITISH_COLOMBIA;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.YUKON;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.NUNAVUT;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.NORTHWEST_TERRITORIES;
+
+public class CanadaPostalcodeToProvinceOrTerritoryMapper {
+    
+    private final static Pattern WELL_FORMED_POSTAL_ADDRESS = Pattern.compile("[A-Z][0-9][A-Z][0-9][A-Z][0-9]");
+    private final static Pattern WELL_FORMED_X_POSTAL_ADDRESS = Pattern.compile("X[0-9][A-Z][0-9][A-Z][0-9]");
```
alexbishop1: Being pedantic here but these should really be `WELL_FORMED_POSTAL_CODE` and `WELL_FORMED_X_POSTAL_CODE`.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,83 @@
...
+            entry("L", ONTARIO),
+            entry("M", ONTARIO),
+            entry("N", ONTARIO),
+            entry("P", ONTARIO),
+            entry("R", MANITOBA),
+            entry("S", SASKATCHEWAN),
+            entry("T", ALBERTA),
+            entry("V", BRITISH_COLOMBIA),
+            entry("Y", YUKON)
+    );
```
alexbishop1, line 46: Checked that these are correct.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,83 @@
...
+
+    public static final Map<String, CanadaProvinceOrTerritory> X_POSTAL_CODE_TERRITORY_MAP = Map.ofEntries(
+            entry("X0A", NUNAVUT),
+            entry("X0B", NUNAVUT),
+            entry("X0C", NUNAVUT),
+            entry("X0E", NORTHWEST_TERRITORIES),
+            entry("X0G", NORTHWEST_TERRITORIES)
+    );
```
alexbishop1, line 54: Checked that these are correct.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,83 @@
...
+            return getProvinceOrTerritoryFromXPostalCode(normalisedPostalCode);
+        } else if (WELL_FORMED_POSTAL_ADDRESS.matcher(normalisedPostalCode).matches()) {
+            return getProvinceOrTerritoryFromPostalCode(normalisedPostalCode);
+        }
+
+        return Optional.empty();
+    }
+    
+    private static Optional<CanadaProvinceOrTerritory> getProvinceOrTerritoryFromXPostalCode(String normalisedPostalCode) {
+        String firstThreeLetters = normalisedPostalCode.substring(0, 3);
```
alexbishop1: We could avoid having to do this by using capturing groups in the regexen.

If we changed `WELL_FORMED_X_POSTAL_ADDRESS` to be `(X[0-9][A-Z])[0-9][A-Z][0-9]` (note the parentheses around the first three characters, we could do something like this:

```java
var xPostalCodeMatcher = WELL_FORMED_X_POSTAL_ADDRESS.matcher(normalisedPostalCode);
if (xPostalCodeMatcher.matches()) {
    var firstThreeCharacters = xPostalCodeMatcher.group(1);
    return Optional.ofNullable(X_POSTAL_CODE_TERRITORY_MAP.get(firstThreeCharacters));
}
```

And similar for the non-X postal codes.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaProvinceOrTerritory.java
```diff
@@ -0,0 +1,37 @@
...
+    ONTARIO("ON", "Ontario"),
+    PRINCE_EDWARD_ISLAND("PE", "Prince Edward Island"),
+    QUEBEC("QC", "Quebec"),
+    SASKATCHEWAN("SK", "Saskatchewan"),
+    YUKON("YT", "Yukon");
+
+    private final String abbreviation;
+    private final String fullName;
+
+    CanadaProvinceOrTerritory(String abbreviation, String fullStateName) {
```
alexbishop1: It’s Canada. No states here!

src/main/java/uk/gov/pay/connector/northamericaregion/NorthAmericaRegion.java
```diff
@@ -0,0 +1,6 @@
+package uk.gov.pay.connector.northamericaregion;
+
+public interface NorthAmericaRegion {
```
alexbishop1, line 3: Other than the geography lesson, is this interface actually buying us anything?

src/main/java/uk/gov/pay/connector/northamericaregion/UsState.java
```diff
@@ -0,0 +1,80 @@
...
+    ARMED_FORCES_AMERICAS("AA", "Armed Forces Americas"),
+    CALIFORNIA("CA", "California"),
+    COLORADO("CO", "Colorado"),
+    CONNECTICUT("CT", "Connecticut"),
+    DELAWARE("DE", "Delaware"),
+    FLORIDA("FL", "Florida"),
+    GEORGIA("GA", "Georgia"),
+    GUAM("GU", "Guam"),
+    HAWAII("HI", "Hawaii"),
+    IDAHO("ID", "Idaho"),
```
alexbishop1, line 21: Note to self: watch _My Own Private Idaho_ some time.

src/main/java/uk/gov/pay/connector/northamericaregion/UsState.java
```diff
@@ -0,0 +1,80 @@
...
+    PUERTO_RICO("PR", "Puerto Rico"),
+    RHODE_ISLAND("RI", "Rhode Island"),
+    SOUTH_CAROLINA("SC", "South Carolina"),
+    SOUTH_DAKOTA("SD", "South Dakota"),
+    TENNESSEE("TN", "Tennessee"),
+    TEXAS("TX", "Texas"),
+    UTAH("UT", "Utah"),
+    VERMONT("VT", "Vermont"),
+    VIRGINIA("VA", "Virginia"),
+    WASHINGTON_DC("DC", "District of Colombia"),
```
alexbishop1: “District of Columbia”.

src/main/java/uk/gov/pay/connector/northamericaregion/UsState.java
```diff
@@ -0,0 +1,80 @@
...
+    WASHINGTON("WA", "Washington"),
+    WEST_VIRGINIA("WV", "West Virginia"),
+    WISCONSIN("WI", "Wisconsin"),
+    WYOMING("WY", "Wyoming");
```
alexbishop1, line 62: Apart from the exceptions noted, these look correct.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMap.java
```diff
@@ -0,0 +1,1003 @@
...
+    
+    /**
+    We have to use this initializiation method because using 'Map.ofEntries(...)' with hundreds of arguments is excessively slow to compile. See https://youtrack.jetbrains.com/issue/IDEA-245963
+     */
+
+    public final static Map<String, UsState> ZIP_CODE_TO_US_STATE_ABBREVIATIONS = Map.copyOf(new HashMap<>() {{
+        put("005", NEW_YORK);
+        put("006", PUERTO_RICO);
+        put("007", PUERTO_RICO);
+        put("008", VIRGINIA);
```
alexbishop1: I think this should be the US Virgin Islands.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMap.java
```diff
@@ -0,0 +1,1003 @@
...
+        put("990", WASHINGTON);
+        put("991", WASHINGTON);
+        put("992", WASHINGTON);
+        put("993", WASHINGTON);
+        put("994", WASHINGTON);
+        put("995", ALASKA);
+        put("996", ALASKA);
+        put("997", ALASKA);
+        put("998", ALASKA);
+        put("999", ALASKA);
```
alexbishop1, line 1002: I’ve checked these and — with the exception noted above — I think they’re correct.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
```diff
@@ -0,0 +1,42 @@
...
+import java.util.Optional;
+import java.util.regex.Pattern;
+
+import static java.util.function.UnaryOperator.identity;
+import static java.util.stream.Collectors.toUnmodifiableMap;
+
+public class UsZipCodeToStateMapper {
+
+    private final static Pattern WELL_FORMED_STATE_WITH_ZIP_CODE = Pattern.compile("[A-Z]{2}[0-9]{5}");
+    private final static Pattern WELL_FORMED_STATE_WITH_ZIP_FOUR_PLUS = Pattern.compile("[A-Z]{2}[0-9]{5}-[0-9]{4}");
```
alexbishop1: `WELL_FORMED_STATE_WITH_ZIP_PLUS_FOUR`

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
```diff
@@ -0,0 +1,42 @@
...
+    
+    private final static Map<String, UsState> US_STATE_ABBREVIATIONS_MAPPING = Arrays.stream(UsState.values()).collect(
+            toUnmodifiableMap(UsState::getAbbreviation, identity())); 
+    
+    public static Optional<UsState> getState(String normalisedZipCode) {
+        
+        if (WELL_FORMED_STATE_WITH_ZIP_CODE.matcher(normalisedZipCode).matches() || WELL_FORMED_STATE_WITH_ZIP_FOUR_PLUS.matcher(normalisedZipCode).matches()) {
+            return getStateFromZipCodeAndState(normalisedZipCode);
+        } else if (WELL_FORMED_ZIP_CODE.matcher(normalisedZipCode).matches() || WELL_FORMED_ZIP_FOUR_PLUS.matcher(normalisedZipCode).matches()) {
+            return getStateFromZipCode(normalisedZipCode.substring(0, 3));
```
alexbishop1: We could use the same trick to use the regex to extract the first few characters.

src/test/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapperTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class CanadaPostalcodeToProvinceOrTerritoryMapperTest {
+
+    @Test
+    public void shouldReturnTheCorrectStateForNonXPostalCode() {
+        Optional<CanadaProvinceOrTerritory> canadaProvinceTerritory = CanadaPostalcodeToProvinceOrTerritoryMapper.getProvinceOrTerritory("A1A1A1");
+
+        assertThat(canadaProvinceTerritory.isPresent(), is (true));
+        assertThat(canadaProvinceTerritory.get().getAbbreviation(), is ("NL"));
+        assertThat(canadaProvinceTerritory.get().getFullName(), is ("Newfoundland and Labrador"));
```
alexbishop1: We can just assert that it’s equal to the enum value, rather than checking the abbreviation and full name separately.

src/test/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapperTest.java
```diff
@@ -0,0 +1,46 @@
...
+    public void shouldReturnTheCorrectStateForNorthwestTerritoriesPostalCode() {
+        Optional<CanadaProvinceOrTerritory> canadaProvinceTerritory = CanadaPostalcodeToProvinceOrTerritoryMapper.getProvinceOrTerritory("X0E1Z0");
+
+        assertThat(canadaProvinceTerritory.isPresent(), is (true));
+        assertThat(canadaProvinceTerritory.get().getAbbreviation(), is ("NT"));
+        assertThat(canadaProvinceTerritory.get().getFullName(), is ("Northwest Territories"));
+    }
+    
+    @Test
+    public void shouldNotReturnAStateForSantaPostCode() {
```
alexbishop1: This should really be `PostalCode` for consistency.

src/test/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapperTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public void shouldReturnTheCorrectStateForLowestValidZipCode() {
+        Optional<UsState> usState = UsZipCodeToStateMapper.getState("05020");
+        
+        assertThat(usState.isPresent(), is (true));
+        assertThat(usState.get().getAbbreviation(), is ("VT"));
+        assertThat(usState.get().getFullName(), is ("Vermont"));
+    }
+
+    @Test
+    public void shouldReturnTheCorrectStateForHighestValidZipCode() {
```
alexbishop1: Are these highest and lowest tests the most appropriate now we’re not using ranges to match?

src/test/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapperTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public void shouldReturnCorrectStateForZipCodeInExclusionListOfAnotherState() {
+        Optional<UsState> usState = UsZipCodeToStateMapper.getState("05510");
+        assertThat(usState.isPresent(), is (true));
+        assertThat(usState.get().getAbbreviation(), is ("MA"));
+        assertThat(usState.get().getFullName(), is ("Massachusetts"));
+    }
+    
+    @Test
+    public void shouldNotReturnStateForZipCodeNotInUse() {
+        Optional<UsState> usState = UsZipCodeToStateMapper.getState("00000");
```
alexbishop1: Wouldn’t 000 or something be better?

alexbishop1 (changes requested):
My main suggestion is to look at using capturing groups in the regexen to extract the prefixes for look-ups. Other than that, I think this is almost there.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
nimalank7: Done. I didn't do it for the US as I'm using `||` with 2 matchers.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaProvinceOrTerritory.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/northamericaregion/NorthAmericaRegion.java
nimalank7: In the next PR it saves a few lines of code. 
`switch(addressEntity.getCountry()) {
case "US"
case "CA"
 Optional<NorthAmericanRegion>(...).map(region -> region.getAbbreviation)
}`

src/main/java/uk/gov/pay/connector/northamericaregion/UsState.java
nimalank7: Ha

src/main/java/uk/gov/pay/connector/northamericaregion/UsState.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMap.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
nimalank7: I don't think we can. In the case for WELL_FORMED_ZIP_CODE and WELL_FORMED_ZIP_FOUR_PLUS - we'd have to create 2 String variables which hold the first 3 characters from either matcher. Otherwise we'd have to have 4 separate if blocks.

src/test/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapperTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapperTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapperTest.java
nimalank7: No apologies I forgot to remove these.

src/test/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapperTest.java
nimalank7: No because the test is checking whether it is in the correct zip code format but isn't actually assigned to anything. So it has to be 5 characters.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,74 @@
...
+            entry("X0A", NUNAVUT),
+            entry("X0B", NUNAVUT),
+            entry("X0C", NUNAVUT),
+            entry("X0E", NORTHWEST_TERRITORIES),
+            entry("X0G", NORTHWEST_TERRITORIES)
+    );
+    
+    public static Optional<CanadaProvinceOrTerritory> getProvinceOrTerritory(String normalisedPostalCode) {
+        var xPostalCodeMatcher = WELL_FORMED_X_POSTAL_CODE.matcher(normalisedPostalCode);
+        var postalCodeMatcher = WELL_FORMED_POSTAL_CODE.matcher(normalisedPostalCode);
```
alexbishop1, line 58: Sucks a bit that we create matchers we might not use but I can’t thing of easy way to avoid this without making the code complicated.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
alexbishop1: Still looks like `FOUR_PLUS` to me…

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
alexbishop1: We could do it if we used combined the regular ZIP and ZIP+4 regexen…

```java
private final static Pattern WELL_FORMED_STATE_WITH_ZIP_AND_OPTIONAL_PLUS_FOUR = Pattern.compile("([A-Z]{2})[0-9]{5}(?:-[0-9]{4})?");
private final static Pattern WELL_FORMED_ZIP_WITH_OPTIONAL_PLUS_FOUR = Pattern.compile("([0-9]{3})[0-9]{2}(?:-[0-9]{4})?");
```

src/test/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapperTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class CanadaPostalcodeToProvinceOrTerritoryMapperTest {
+
+    @Test
+    public void shouldReturnTheCorrectStateForNonXPostalCode() {
+        Optional<CanadaProvinceOrTerritory> canadaProvinceTerritory = CanadaPostalcodeToProvinceOrTerritoryMapper.getProvinceOrTerritory("A1A1A1");
+
+        assertThat(canadaProvinceTerritory.isPresent(), is (true));
+        assertThat(canadaProvinceTerritory.get().getAbbreviation(), is (CanadaProvinceOrTerritory.NEWFOUNDLAND_AND_LABRADOR.getAbbreviation()));
+        assertThat(canadaProvinceTerritory.get().getFullName(), is (CanadaProvinceOrTerritory.NEWFOUNDLAND_AND_LABRADOR.getFullName()));
```
alexbishop1: This could just be:

```java
assertThat(canadaProvinceTerritory.get(), is(CanadaProvinceOrTerritory.NEWFOUNDLAND_AND_LABRADOR));
```

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
```diff
@@ -0,0 +1,74 @@
...
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.MANITOBA;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.ALBERTA;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.BRITISH_COLUMBIA;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.YUKON;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.NUNAVUT;
+import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.NORTHWEST_TERRITORIES;
+
+public class CanadaPostalcodeToProvinceOrTerritoryMapper {
+    
+    private final static Pattern WELL_FORMED_POSTAL_CODE = Pattern.compile("([A-Z])[0-9][A-Z][0-9][A-Z][0-9]");
```
alexbishop1: Not going to insist on this but generally the order is `private static final` (applies to a few places).

alexbishop1 (changes requested):
One request relating to a naming thing. Another suggestion of how we could use possibly use capturing groups for US ZIP codes, which I’ll leave to your discretion. Other than that, looks good!

src/test/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapperTest.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
nimalank7: Yeah - I had to quickly look up what non-capturing groups in regexes are! But that change does make it easier.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public static Optional<UsState> getState(String normalisedZipCode) {
+        var wellFormedZipCodeAndOptionalPlusFourMatcher = WELL_FORMED_STATE_WITH_ZIP_CODE_AND_OPTIONAL_PLUS_FOUR.matcher(normalisedZipCode);
+        var wellFormedStateWithZipAndOptionalPlusFourMatcher = WELL_FORMED_ZIP_CODE_AND_OPTIONAL_PLUS_FOUR.matcher(normalisedZipCode);
+
+        if (wellFormedStateWithZipAndOptionalPlusFourMatcher.matches()) {
+            String stateAbbreviation = wellFormedStateWithZipAndOptionalPlusFourMatcher.group(1);
+            return Optional.ofNullable(UsZipCodeToStateMap.ZIP_CODE_TO_US_STATE_ABBREVIATIONS.get(stateAbbreviation));
+        } else if (wellFormedZipCodeAndOptionalPlusFourMatcher.matches()) {
+            String FirstThreeCharacters = wellFormedZipCodeAndOptionalPlusFourMatcher.group(1);
```
alexbishop1: A capital `F`?

alexbishop1 (changes requested):
Looks great apart from a rogue capital letter at the beginning of a variable name.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
nimalank7: Done.

alexbishop1 (approved):
🎉

PR #2502, opened 2020-08-19
PP-6256 - Send US states and Canadian provinces or territories to gateway, ledger and database

nimalank7:
Description:
- Since the mappings are now in place we can store the US states and Canadian provinces in the charge database, a PaymentDetailsEnteredEvent which contains the address_state_province and sent to Ledger as well as being sent to the payment gateways
- The StripePaymentProvider contract test previously had an incorrect configuration. This is because the previous configuration uses the DropwizardAppRule to spin up the Guice container which contains the StripePaymentProvider class as well as recreating it manually which causes an exception as the metric already exists. The new configuration merely retrieves the existing class.
- Since some of the code is reused across the payment provider builders it has been extracted into NorthAmericanRegionMapper

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -56,6 +56,8 @@ import uk.gov.pay.connector.gateway.model.PayersCardType;
 import uk.gov.pay.connector.gateway.model.ProviderSessionIdentifier;
 import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.northamericaregion.NorthAmericaRegion;
+import uk.gov.pay.connector.northamericaregion.NorthAmericanRegionMapper;
 import uk.gov.pay.connector.paritycheck.LedgerService;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 import uk.gov.pay.connector.queue.statetransition.StateTransitionService;
@@ -120,13 +122,14 @@ public class ChargeService {
     private final Boolean shouldEmitPaymentStateTransitionEvents;
     private final RefundDao refundDao;
     private EventService eventService;
+    private final NorthAmericanRegionMapper northAmericanRegionMapper;
 
     @Inject
     public ChargeService(TokenDao tokenDao, ChargeDao chargeDao, ChargeEventDao chargeEventDao,
                          CardTypeDao cardTypeDao, GatewayAccountDao gatewayAccountDao,
                          ConnectorConfiguration config, PaymentProviders providers,
                          StateTransitionService stateTransitionService, LedgerService ledgerService, EventService eventService,
-                         RefundDao refundDao) {
+                         RefundDao refundDao, NorthAmericanRegionMapper northAmericanRegionMapper) {
         this.tokenDao = tokenDao;
         this.chargeDao = chargeDao;
         this.chargeEventDao = chargeEventDao;
@@ -140,6 +143,7 @@ public class ChargeService {
         this.ledgerService = ledgerService;
         this.eventService = eventService;
         this.refundDao = refundDao;
+        this.northAmericanRegionMapper = northAmericanRegionMapper;
     }
 
     @Transactional
@@ -752,10 +756,14 @@ public class ChargeService {
         if (hasLastFourCharactersCardNumber(authCardDetails)) {
             detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
         }
-        
+
         authCardDetails.getAddress().ifPresent(address -> {
-            AddressEntity addressEntity = new AddressEntity(authCardDetails.getAddress().get());
-            processRegionForCanadaOrUs(addressEntity);
+            var addressEntity = new AddressEntity(address);
+            northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
+                    .map(NorthAmericaRegion::getAbbreviation)
+                    .ifPresent(stateOrProvinceAbbreviation -> {
+                        addressEntity.setStateOrProvince(stateOrProvinceAbbreviation);
+                    });
             detailsEntity.setBillingAddress(addressEntity);
         });
 
@@ -770,23 +778,6 @@ public class ChargeService {
     
     private boolean hasLastFourCharactersCardNumber(AuthCardDetails authCardDetails) {
         return authCardDetails.getCardNo().length() >= 4;
-    }
-    
-    private void processRegionForCanadaOrUs(AddressEntity addressEntity) {
-        switch (addressEntity.getCountry()) {
-            case "US":
-                Optional.of(UsZipCodeToStateMapper.getState(getNormalisedPostalCode(addressEntity)));
-                break;
-            case "CA":
-                Optional.of(CanadaPostalcodeToProvinceOrTerritoryMapper.getProvinceOrTerritory(getNormalisedPostalCode(addressEntity)));
-                break;
-            default:
-                break;
-        }
-    }
-
-    private String getNormalisedPostalCode(AddressEntity addressEntity) {
-        return addressEntity.getPostcode().replaceAll("\\s", "").toUpperCase(Locale.ENGLISH);
     }
 
     private TokenEntity createNewChargeEntityToken(ChargeEntity chargeEntity) {
--- a/src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilder.java
@@ -1,17 +1,24 @@
 package uk.gov.pay.connector.gateway.smartpay;
 
+import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 import uk.gov.pay.connector.gateway.model.OrderRequestType;
 import uk.gov.pay.connector.gateway.templates.PayloadBuilder;
 import uk.gov.pay.connector.gateway.templates.TemplateBuilder;
 import uk.gov.pay.connector.gateway.OrderRequestBuilder;
+import uk.gov.pay.connector.northamericaregion.CanadaPostalcodeToProvinceOrTerritoryMapper;
+import uk.gov.pay.connector.northamericaregion.NorthAmericaRegion;
+import uk.gov.pay.connector.northamericaregion.NorthAmericanRegionMapper;
+import uk.gov.pay.connector.northamericaregion.UsZipCodeToStateMapper;
 
 import javax.ws.rs.core.MediaType;
+import java.util.Locale;
 
 public class SmartpayOrderRequestBuilder extends OrderRequestBuilder {
     static public class SmartpayTemplateData extends TemplateData {
         private String reference;
         private String paResponse;
         private String md;
+        private String stateOrProvince;
 
         public String getReference() {
             return reference;
@@ -36,6 +43,14 @@ public class SmartpayOrderRequestBuilder extends OrderRequestBuilder {
         public void setMd(final String md) {
             this.md = md;
         }
+
+        public String getStateOrProvince() {
+            return stateOrProvince;
+        }
+
+        public void setStateOrProvince(String stateOrProvince) {
+            this.stateOrProvince = stateOrProvince;
+        }
     }
 
     public static final TemplateBuilder AUTHORISE_ORDER_TEMPLATE_BUILDER = new TemplateBuilder("/smartpay/SmartpayAuthoriseOrderTemplate.xml");
@@ -46,6 +61,7 @@ public class SmartpayOrderRequestBuilder extends OrderRequestBuilder {
     public static final TemplateBuilder REFUND_ORDER_TEMPLATE_BUILDER = new TemplateBuilder("/smartpay/SmartpayRefundOrderTemplate.xml");
 
     private SmartpayTemplateData smartpayTemplateData;
+    private final NorthAmericanRegionMapper northAmericanRegionMapper;
 
     public static SmartpayOrderRequestBuilder aSmartpayAuthoriseOrderRequestBuilder() {
         return new SmartpayOrderRequestBuilder(new SmartpayTemplateData(), AUTHORISE_ORDER_TEMPLATE_BUILDER, OrderRequestType.AUTHORISE);
@@ -73,6 +89,7 @@ public class SmartpayOrderRequestBuilder extends OrderRequestBuilder {
 
     public SmartpayOrderRequestBuilder(SmartpayTemplateData smartpayTemplateData, PayloadBuilder payloadBuilder, OrderRequestType orderRequestType) {
         super(smartpayTemplateData, payloadBuilder, orderRequestType);
+        this.northAmericanRegionMapper = new NorthAmericanRegionMapper();
         this.smartpayTemplateData = smartpayTemplateData;
     }
 
@@ -89,6 +106,16 @@ public class SmartpayOrderRequestBuilder extends OrderRequestBuilder {
     public SmartpayOrderRequestBuilder withMd(String md) {
         smartpayTemplateData.setMd(md);
         return this;
+    }
+
+    @Override
+    public OrderRequestBuilder withAuthorisationDetails(AuthCardDetails authCardDetails) {
+        OrderRequestBuilder orderRequestBuilder = super.withAuthorisationDetails(authCardDetails);
+        authCardDetails.getAddress()
+                .flatMap(northAmericanRegionMapper::getNorthAmericanRegionForCountry)
+                .map(NorthAmericaRegion::getAbbreviation)
+                .ifPresent(smartpayTemplateData::setStateOrProvince);
+        return orderRequestBuilder;
     }
 
     @Override
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
@@ -2,16 +2,24 @@ package uk.gov.pay.connector.gateway.stripe.request;
 
 import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
+import uk.gov.pay.connector.charge.model.AddressEntity;
+import uk.gov.pay.connector.common.model.domain.Address;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 import uk.gov.pay.connector.gateway.model.OrderRequestType;
 import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.northamericaregion.CanadaPostalcodeToProvinceOrTerritoryMapper;
+import uk.gov.pay.connector.northamericaregion.NorthAmericaRegion;
+import uk.gov.pay.connector.northamericaregion.NorthAmericanRegionMapper;
+import uk.gov.pay.connector.northamericaregion.UsZipCodeToStateMapper;
 
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 public class StripePaymentMethodRequest extends StripeRequest {
     private final AuthCardDetails authCardDetails;
+    private final NorthAmericanRegionMapper northAmericanRegionMapper;
     
     public StripePaymentMethodRequest(
             GatewayAccountEntity gatewayAccount,
@@ -21,6 +29,7 @@ public class StripePaymentMethodRequest extends StripeRequest {
     {
         super(gatewayAccount, idempotencyKey, stripeGatewayConfig);
         this.authCardDetails = authCardDetails;
+        this.northAmericanRegionMapper = new NorthAmericanRegionMapper();
     }
     
     public static StripePaymentMethodRequest of(CardAuthorisationGatewayRequest request, StripeGatewayConfig config) {
@@ -47,6 +56,9 @@ public class StripePaymentMethodRequest extends StripeRequest {
             if (StringUtils.isNotBlank(address.getLine2())) {
                 localParams.put("billing_details[address[line2]]", address.getLine2());
             }
+            northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
+                    .map(NorthAmericaRegion::getFullName)
+                    .ifPresent(stateOrProvince -> localParams.put("billing_details[address[state]]", stateOrProvince));
             localParams.put("billing_details[address[city]]", address.getCity());
             localParams.put("billing_details[address[country]]", address.getCountry());
             localParams.put("billing_details[address[postal_code]]", address.getPostcode());
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
@@ -4,13 +4,19 @@ import org.joda.time.DateTime;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.gateway.OrderRequestBuilder;
+import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 import uk.gov.pay.connector.gateway.model.OrderRequestType;
 import uk.gov.pay.connector.gateway.templates.PayloadBuilder;
 import uk.gov.pay.connector.gateway.templates.TemplateBuilder;
+import uk.gov.pay.connector.northamericaregion.CanadaPostalcodeToProvinceOrTerritoryMapper;
+import uk.gov.pay.connector.northamericaregion.NorthAmericaRegion;
+import uk.gov.pay.connector.northamericaregion.NorthAmericanRegionMapper;
+import uk.gov.pay.connector.northamericaregion.UsZipCodeToStateMapper;
 import uk.gov.pay.connector.wallets.WalletType;
 import uk.gov.pay.connector.wallets.model.WalletAuthorisationData;
 
 import javax.ws.rs.core.MediaType;
+import java.util.Locale;
 import java.util.Optional;
 
 public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
@@ -27,6 +33,7 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
         private boolean requires3ds;
         private String paResponse3ds;
         private String payerIpAddress;
+        private String state;
 
         public String getReference() {
             return reference;
@@ -101,6 +108,14 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
         public void setPayerIpAddress(String payerIpAddress) {
             this.payerIpAddress = payerIpAddress;
         }
+
+        public String getState() {
+            return state;
+        }
+
+        public void setState(String state) {
+            this.state = state;
+        }
     }
 
     public static final TemplateBuilder AUTHORISE_ORDER_TEMPLATE_BUILDER = new TemplateBuilder("/worldpay/WorldpayAuthoriseOrderTemplate.xml");
@@ -113,6 +128,7 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
     public static final TemplateBuilder INQUIRY_TEMPLATE_BUILDER = new TemplateBuilder("/worldpay/WorldpayInquiryOrderTemplate.xml");
 
     private final WorldpayTemplateData worldpayTemplateData;
+    private final NorthAmericanRegionMapper northAmericanRegionMapper;
 
     public static WorldpayOrderRequestBuilder aWorldpayAuthoriseOrderRequestBuilder() {
         return new WorldpayOrderRequestBuilder(new WorldpayTemplateData(), AUTHORISE_ORDER_TEMPLATE_BUILDER, OrderRequestType.AUTHORISE);
@@ -144,6 +160,7 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
 
     private WorldpayOrderRequestBuilder(WorldpayTemplateData worldpayTemplateData, PayloadBuilder payloadBuilder, OrderRequestType orderRequestType) {
         super(worldpayTemplateData, payloadBuilder, orderRequestType);
+        this.northAmericanRegionMapper = new NorthAmericanRegionMapper();
         this.worldpayTemplateData = worldpayTemplateData;
     }
 
@@ -191,6 +208,16 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
     public WorldpayOrderRequestBuilder withPayerIpAddress(String payerIpAddress) {
         worldpayTemplateData.setPayerIpAddress(payerIpAddress);
         return this;
+    }
+
+    @Override
+    public OrderRequestBuilder withAuthorisationDetails(AuthCardDetails authCardDetails) {
+        OrderRequestBuilder orderRequestBuilder = super.withAuthorisationDetails(authCardDetails);
+        authCardDetails.getAddress()
+                .flatMap(northAmericanRegionMapper::getNorthAmericanRegionForCountry)
+                .map(NorthAmericaRegion::getAbbreviation)
+                .ifPresent(worldpayTemplateData::setState);
+        return orderRequestBuilder;
     }
 
     public WorldpayOrderRequestBuilder withPaResponse3ds(String paResponse) {
--- a/src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
+++ b/src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
@@ -21,11 +21,11 @@ import static uk.gov.pay.connector.northamericaregion.CanadaProvinceOrTerritory.
 
 public class CanadaPostalcodeToProvinceOrTerritoryMapper {
 
-    private static final Pattern WELL_FORMED_POSTAL_CODE = Pattern.compile("([A-Z])[0-9][A-Z][0-9][A-Z][0-9]");
-    private static final Pattern WELL_FORMED_X_POSTAL_CODE = Pattern.compile("(X[0-9][A-Z])[0-9][A-Z][0-9]");
-    private static final String SANTA_CLAUS_POSTAL_CODE = "H0H0H0";
+    private final Pattern WELL_FORMED_POSTAL_CODE = Pattern.compile("([A-Z])[0-9][A-Z][0-9][A-Z][0-9]");
+    private final Pattern WELL_FORMED_X_POSTAL_CODE = Pattern.compile("(X[0-9][A-Z])[0-9][A-Z][0-9]");
+    private final String SANTA_CLAUS_POSTAL_CODE = "H0H0H0";
 
-    public static final Map<String, CanadaProvinceOrTerritory> NON_X_POSTAL_CODE_TERRITORY_PROVINCE_MAP = Map.ofEntries(
+    public final Map<String, CanadaProvinceOrTerritory> NON_X_POSTAL_CODE_TERRITORY_PROVINCE_MAP = Map.ofEntries(
             entry("A", NEWFOUNDLAND_AND_LABRADOR),
             entry("B", NOVA_SCOTIA),
             entry("C", PRINCE_EDWARD_ISLAND),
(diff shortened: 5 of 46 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -754,7 +754,7 @@ private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails
         }
         
         authCardDetails.getAddress().ifPresent(address -> {
-            AddressEntity addressEntity = new AddressEntity(authCardDetails.getAddress().get());
+            AddressEntity addressEntity = new AddressEntity(address);
```
alexbishop1: Good spot!

We could actually do this:

```java
authCardDetails.getAddress().map(AddressEntity::new).ifPresent(addressEntity -> {
    // …
});
```

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -775,10 +775,12 @@ private boolean hasLastFourCharactersCardNumber(AuthCardDetails authCardDetails)
     private void processRegionForCanadaOrUs(AddressEntity addressEntity) {
```
alexbishop1: The word “process” is a bit vague. Maybe `addUsStateOrCanadaProvinceTerritory` or something?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -775,10 +775,12 @@ private boolean hasLastFourCharactersCardNumber(AuthCardDetails authCardDetails)
...
         switch (addressEntity.getCountry()) {
             case "US":
-                Optional.of(UsZipCodeToStateMapper.getState(getNormalisedPostalCode(addressEntity)));
+                UsZipCodeToStateMapper.getState(getNormalisedPostalCode(addressEntity))
```
alexbishop1: I think it would be preferable if `UsZipCodeToStateMapper` (and `CanadaPostalcodeToProvinceOrTerritoryMapper`) were dependency-injected instantiated objects that we called non-static methods on.

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilder.java
```diff
@@ -91,8 +104,34 @@ public SmartpayOrderRequestBuilder withMd(String md) {
...
+                    break;
+                case "CA":
+                    CanadaPostalcodeToProvinceOrTerritoryMapper.getProvinceOrTerritory(normalisedPostalCode)
+                            .ifPresent(provinceOrTerritory -> smartpayTemplateData.setStateOrProvince(provinceOrTerritory.getFullName()));
+                    break;
+                default:
+                    break;
+            }
+        });
+    }
```
alexbishop1: This is very similar to the code in `ChargeService` (and I expect other classes I haven’t gotten to yet). I think we should extract this to a separate class with a method that takes an `Address` (from where it can get both the country and the ready-for-normalisation postcode) and returns a `Optional<NorthAmericanRegion>` (I take back everything I said about that being an unnecessary interface!). The caller can then decide if it wants the abbreviation or the full name and set it on whatever the appropriate object is.

src/main/resources/templates/smartpay/Smartpay3dsRequiredOrderTemplate.xml
```diff
@@ -32,7 +32,7 @@
                         <ns2:houseNumberOrName>${authCardDetails.address.get().line1?xml}</ns2:houseNumberOrName>
                         <ns2:street><#if authCardDetails.address.get().line2?has_content>${authCardDetails.address.get().line2?xml}<#else>N/A</#if></ns2:street>
                         <ns2:postalCode>${authCardDetails.address.get().postcode?xml}</ns2:postalCode>
-                        <ns2:stateOrProvince><#if authCardDetails.address.get().county??>${authCardDetails.address.get().county?xml}</#if></ns2:stateOrProvince>
+                        <ns2:stateOrProvince><#if stateOrProvince?has_content>${stateOrProvince}</#if></ns2:stateOrProvince>
```
alexbishop1: This should be `${stateOrProvince?xml}` to make the necessary escaping is applied.

src/main/resources/templates/smartpay/SmartpayAuthoriseOrderTemplate.xml
```diff
@@ -21,7 +21,7 @@
                         <ns2:houseNumberOrName>${authCardDetails.address.get().line1?xml}</ns2:houseNumberOrName>
                         <ns2:street><#if authCardDetails.address.get().line2?has_content>${authCardDetails.address.get().line2?xml}<#else>N/A</#if></ns2:street>
                         <ns2:postalCode>${authCardDetails.address.get().postcode?xml}</ns2:postalCode>
-                        <ns2:stateOrProvince><#if authCardDetails.address.get().county??>${authCardDetails.address.get().county?xml}</#if></ns2:stateOrProvince>
+                        <ns2:stateOrProvince><#if stateOrProvince?has_content>${stateOrProvince}</#if></ns2:stateOrProvince>
```
alexbishop1: `${stateOrProvince?xml}`

src/test/java/uk/gov/pay/connector/charge/model/domain/ChargeEntityFixture.java
```diff
@@ -88,6 +88,24 @@ public static CardDetailsEntity defaultCardDetails() {
...
+        cardDetailsEntity.setCardHolderName("Test");
+        cardDetailsEntity.setCardBrand("visa");
+        cardDetailsEntity.setExpiryDate("11/99");
+        cardDetailsEntity.setFirstDigitsCardNumber(FirstDigitsCardNumber.of("123456"));
+        cardDetailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of("1234"));
+        cardDetailsEntity.setCardType(CardType.DEBIT);
+        cardDetailsEntity.setCardTypeDetails(CardBrandLabelEntity);
+        cardDetailsEntity.setBillingAddress(addressEntity);
+
+        return cardDetailsEntity;
```
alexbishop1: Any reason this can’t be just…?

```java
var cardDetailsEntity = defaultCardDetails().setBillingAddress(addressEntity);
return cardDetailsEntity;
```

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -427,6 +429,52 @@ public void updateChargeAndEmitEventPostAuthorisation_shouldEmitEvent() {
...
+        when(chargeEvent.getStatus()).thenReturn(ENTERING_CARD_DETAILS);
+        when(chargeEvent.getUpdated()).thenReturn(now());
+        when(mockedChargeEventDao.persistChargeEventOf(chargeSpy, null)).thenReturn(chargeEvent);
+        when(mockedChargeDao.findByExternalId(chargeSpy.getExternalId())).thenReturn(Optional.of(chargeSpy));
+        when(chargeSpy.getEvents()).thenReturn(List.of(chargeEvent));
+
+        service.updateChargeAndEmitEventPostAuthorisation(chargeSpy.getExternalId(), ENTERING_CARD_DETAILS,
+                authCardDetails, null, null, null, null, null);
+
+        assertThat(chargeSpy.getCardDetails().getBillingAddress().get().getStateOrProvince(), is("DC"));
```
alexbishop1, line 487: If we injected the mapper as suggested in an earlier comment, we could stub out the ZIP code to state mapping, which would mean this test wouldn’t rely on the external knowledge that 20500 is Washington DC.

src/test/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilderTest.java
```diff
@@ -48,7 +50,7 @@ public void tearDown() {
 
     @Test
     public void shouldGenerateValidAuthoriseOrderRequestForAddressWithAllFields() throws Exception {
-        Address address = new Address("41", "Scala Street", "EC2A 1AE", "London", "London", "GB");
+        Address address = new Address("41", "Scala Street", "EC2A 1AE", "London", null, "GB");
```
alexbishop1, line 53: I’d be really tempted to explore a follow-up PR that removes the county stuff from `Address` altogether. It’s just confusing.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProviderTest.java
```diff
@@ -114,6 +114,20 @@ public void shouldAuthoriseImmediately_whenPaymentIntentReturnsAsRequiresCapture
         assertThat(response.getBaseResponse().get().getTransactionId(), is("pi_1FHESeEZsufgnuO08A2FUSPy"));
     }
 
+    @Test
+    public void shouldAuthorise_ForAddressInUs() throws Exception {
```
alexbishop1, line 118: Not sure this test is buying us a huge deal as we don’t assert anything about the request and it doesn’t change anything about the response.

src/test/java/uk/gov/pay/connector/it/contract/SmartpayPaymentProviderTest.java
```diff
@@ -127,6 +128,27 @@ public void shouldSendSuccessfullyAnOrderForMerchantWithNoAddressInRequest() thr
         assertTrue(response.isSuccessful());
     }
 
+    @Test
+    public void shouldSendSuccessfullyAnOrderForMerchantWithUsAddressInRequest() throws Exception {
```
alexbishop1, line 132: Maybe one for Canada too?

src/test/java/uk/gov/pay/connector/it/resources/CardResourceAuthoriseIT.java
```diff
@@ -90,6 +90,30 @@ public void shouldStoreCardDetailsForAuthorisedCharge() {
         assertThat(databaseTestHelper.getChargeCardBrand(chargeId), is(cardBrand));
     }
 
+    @Test
+    public void shouldStoreAddressStateProvinceForAuthorisedChargeFromUs() {
```
alexbishop1: `FromUs` makes it sound like it’s… from us. Maybe `FromUsa` or `FromUnitedStates`?

src/test/java/uk/gov/pay/connector/it/resources/CardResourceAuthoriseIT.java
```diff
@@ -90,6 +90,30 @@ public void shouldStoreCardDetailsForAuthorisedCharge() {
...
+                "Mr. Name",
+                "4444333322221111",
+                "123",
+                "10/99",
+                "visa",
+                "CREDIT",
+                "Line1",
+                "Line2",
+                "Washington D.C.",
+                "DC",
```
alexbishop1: Is this right? This test is testing a call to connector, right? And connector never receives the state in the call, it infers it internally.

src/test/resources/templates/smartpay/special-char-valid-authorise-smartpay-3ds-request.xml
```diff
@@ -31,7 +31,7 @@
                         <ns2:houseNumberOrName>41</ns2:houseNumberOrName>
                         <ns2:street>Scala &amp; Haskell Rocks</ns2:street>
                         <ns2:postalCode>EC2A 1AE</ns2:postalCode>
-                        <ns2:stateOrProvince>London --&gt;</ns2:stateOrProvince>
```
alexbishop1, line 34: Maybe the `--&gt;` should be moved to another element. Looks like this test is trying to check entities are handled correctly.

src/test/resources/templates/worldpay/valid-authorise-worldpay-request-full-address.xml
```diff
@@ -20,7 +20,6 @@
                             <address2>This road</address2>
                             <postalCode>SW8URR</postalCode>
                             <city>London</city>
-                            <state>London county</state>
```
alexbishop1, line 23: https://commons.wikimedia.org/wiki/File:County_Hall_(8710618372).jpg

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -427,6 +429,52 @@ public void updateChargeAndEmitEventPostAuthorisation_shouldEmitEvent() {
...
+        when(mockedChargeEventDao.persistChargeEventOf(chargeSpy, null)).thenReturn(chargeEvent);
+        when(mockedChargeDao.findByExternalId(chargeSpy.getExternalId())).thenReturn(Optional.of(chargeSpy));
+        when(chargeSpy.getEvents()).thenReturn(List.of(chargeEvent));
+
+        service.updateChargeAndEmitEventPostAuthorisation(chargeSpy.getExternalId(), ENTERING_CARD_DETAILS,
+                authCardDetails, null, null, null, null, null);
+
+        assertThat(chargeSpy.getCardDetails().getBillingAddress().get().getStateOrProvince(), is(nullValue()));
+    }
+
```
alexbishop1, line 512: We might want a variant of `updateChargeAndEmitEventPostAuthorisation_shouldEmitEvent` for when the country is the US or Canada to check the state/province/territory gets sent to ledger.

alexbishop1 (changes requested):
This looks like it’s along the right lines. Comments mostly relate to reducing duplication and tests.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done.

src/main/resources/templates/smartpay/Smartpay3dsRequiredOrderTemplate.xml
nimalank7: Done.

src/main/resources/templates/smartpay/SmartpayAuthoriseOrderTemplate.xml
nimalank7: Done.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripePaymentProviderTest.java
nimalank7: Yeah that's true - and also it's covered in the contract tests.

src/test/java/uk/gov/pay/connector/it/resources/CardResourceAuthoriseIT.java
nimalank7: Oh yes your right I've set it to null.

src/main/resources/templates/smartpay/Smartpay3dsRequiredOrderTemplate.xml
nimalank7: Done

src/test/java/uk/gov/pay/connector/charge/model/domain/ChargeEntityFixture.java
nimalank7: Ah I just realised this method is never called so I've removed it.

src/test/resources/templates/smartpay/special-char-valid-authorise-smartpay-3ds-request.xml
nimalank7: Done

src/test/java/uk/gov/pay/connector/it/contract/SmartpayPaymentProviderTest.java
nimalank7: Done

src/test/java/uk/gov/pay/connector/it/resources/CardResourceAuthoriseIT.java
nimalank7: Done

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilder.java
nimalank7: Done.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
nimalank7: I've stubbed it out now.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -753,9 +756,10 @@ private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails
             detailsEntity.setLastDigitsCardNumber(LastDigitsCardNumber.of(StringUtils.right(authCardDetails.getCardNo(), 4)));
         }
         
-        authCardDetails.getAddress().ifPresent(address -> {
-            AddressEntity addressEntity = new AddressEntity(authCardDetails.getAddress().get());
-            processRegionForCanadaOrUs(addressEntity);
+        authCardDetails.getAddress().map(AddressEntity::new).ifPresent(addressEntity -> {
+            northAmericanRegionMapper
+                    .getNorthAmericanRegionForCountry(addressEntity)
+                    .ifPresent(northAmericanRegion -> addressEntity.setStateOrProvince(northAmericanRegion.getAbbreviation()));
```
alexbishop1: I spent longer than I care to admit fiddling with this to see if there’s a way we can remove the need for the nested `ifPresent(…)` methods. It’s hard because we have to deal with both the address and a state or province maybe not being present. In the end I came up with this:

```java
authCardDetails.getAddress().ifPresent(address -> {
    northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
            .map(NorthAmericaRegion::getAbbreviation)
            .ifPresent(stateOrProvinceAbbreviation -> {
                var addressEntity = new AddressEntity(address);
                addressEntity.setStateOrProvince(stateOrProvinceAbbreviation);
                detailsEntity.setBillingAddress(addressEntity);
            });
});
```

It’s not beautiful by any means but it does mean we can call `NorthAmericanRegionMapper.getNorthAmericanRegionForCountry(Address)` rather than `NorthAmericanRegionMapper.getNorthAmericanRegionForCountry(AddressEntity)`, which then means we don’t need the latter, which removes some duplication.

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilder.java
```diff
@@ -91,6 +107,16 @@ public SmartpayOrderRequestBuilder withMd(String md) {
         return this;
     }
 
+    @Override
+    public OrderRequestBuilder withAuthorisationDetails(AuthCardDetails authCardDetails) {
+        OrderRequestBuilder orderRequestBuilder = super.withAuthorisationDetails(authCardDetails);
+        authCardDetails.getAddress().ifPresent(address ->
+                northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
+                        .ifPresent(stateOrProvince -> smartpayTemplateData.setStateOrProvince(stateOrProvince.getAbbreviation()))
+        );
```
alexbishop1:
```java
authCardDetails.getAddress()
        .flatMap(northAmericanRegionMapper::getNorthAmericanRegionForCountry)
        .map(NorthAmericaRegion::getAbbreviation)
        .ifPresent(smartpayTemplateData::setStateOrProvince);
```

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
```diff
@@ -47,6 +55,8 @@ public static StripePaymentMethodRequest of(CardAuthorisationGatewayRequest requ
             if (StringUtils.isNotBlank(address.getLine2())) {
                 localParams.put("billing_details[address[line2]]", address.getLine2());
             }
+            northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
+                    .ifPresent(stateOrProvince -> localParams.put("billing_details[address[state]]", stateOrProvince.getFullName()));
```
alexbishop1: Could also do this:

```java
northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
        .map(NorthAmericaRegion::getFullName)
        .ifPresent(stateOrProvince -> localParams.put("billing_details[address[state]]", stateOrProvince));
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
```diff
@@ -193,6 +209,16 @@ public WorldpayOrderRequestBuilder withPayerIpAddress(String payerIpAddress) {
         return this;
     }
 
+    @Override
+    public OrderRequestBuilder withAuthorisationDetails(AuthCardDetails authCardDetails) {
+        OrderRequestBuilder orderRequestBuilder = super.withAuthorisationDetails(authCardDetails);
+        authCardDetails.getAddress().ifPresent(address ->
+                northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
+                        .ifPresent(stateOrProvince -> worldpayTemplateData.setState(stateOrProvince.getAbbreviation()))
```
alexbishop1:
```java
authCardDetails.getAddress()
        .flatMap(northAmericanRegionMapper::getNorthAmericanRegionForCountry)
        .map(NorthAmericaRegion::getAbbreviation)
        .ifPresent(worldpayTemplateData::setState);
```

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
```diff
@@ -45,15 +45,15 @@
...
+    public final Map<String, NorthAmericaRegion> X_POSTAL_CODE_TERRITORY_MAP = Map.ofEntries(
             entry("X0A", NUNAVUT),
             entry("X0B", NUNAVUT),
             entry("X0C", NUNAVUT),
             entry("X0E", NORTHWEST_TERRITORIES),
             entry("X0G", NORTHWEST_TERRITORIES)
     );
-    
-    public static Optional<CanadaProvinceOrTerritory> getProvinceOrTerritory(String normalisedPostalCode) {
+
```
alexbishop1, line 55: It should be possible to make this `CanadaProvinceOrTerritory` rather than `NorthAmericaRegion`. Might have to change `getNorthAmericanRegionForCountry(…)` in `NorthAmericanRegionMapper` to return `Optional<? extends NorthAmericaRegion>` to get it all to compile.

src/main/java/uk/gov/pay/connector/northamericaregion/NorthAmericanRegionMapper.java
```diff
@@ -0,0 +1,45 @@
...
+        this.canadaPostalcodeToProvinceOrTerritoryMapper = new CanadaPostalcodeToProvinceOrTerritoryMapper();
+    }
+
+    @Inject
+    public NorthAmericanRegionMapper(UsZipCodeToStateMapper usZipCodeToStateMapper, CanadaPostalcodeToProvinceOrTerritoryMapper canadaPostalcodeToProvinceOrTerritoryMapper) {
+        this.usZipCodeToStateMapper = usZipCodeToStateMapper;
+        this.canadaPostalcodeToProvinceOrTerritoryMapper = canadaPostalcodeToProvinceOrTerritoryMapper;
+    }
+    
+    public Optional<NorthAmericaRegion> getNorthAmericanRegionForCountry(AddressEntity addressEntity) {
```
alexbishop1: As suggested in earlier comments, I think we can just have the `Address` variant of this method.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
```diff
@@ -10,12 +10,12 @@
...
-    private static final Pattern WELL_FORMED_STATE_WITH_ZIP_CODE_AND_OPTIONAL_PLUS_FOUR = Pattern.compile("([A-Z]{2})[0-9]{5}(?:-[0-9]{4})?");
-    private static final Map<String, UsState> US_STATE_ABBREVIATIONS_MAPPING = Arrays.stream(UsState.values()).collect(
-            toUnmodifiableMap(UsState::getAbbreviation, identity())); 
+    private final Pattern WELL_FORMED_ZIP_CODE_AND_OPTIONAL_PLUS_FOUR = Pattern.compile("([0-9]{3})[0-9]{2}(?:-[0-9]{4})?");
+    private final Pattern WELL_FORMED_STATE_WITH_ZIP_CODE_AND_OPTIONAL_PLUS_FOUR = Pattern.compile("([A-Z]{2})[0-9]{5}(?:-[0-9]{4})?");
+    private final Map<String, NorthAmericaRegion> US_STATE_ABBREVIATIONS_MAPPING = Arrays.stream(UsState.values()).collect(
+            toUnmodifiableMap(NorthAmericaRegion::getAbbreviation, identity())); 
 
-    public static Optional<UsState> getState(String normalisedZipCode) {
+    public Optional<NorthAmericaRegion> getState(String normalisedZipCode) {
```
alexbishop1: I think this can actually stay as `UsState`.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -427,6 +435,77 @@ public void updateChargeAndEmitEventPostAuthorisation_shouldEmitEvent() {
...
+        when(chargeSpy.getEvents()).thenReturn(List.of(chargeEvent));
+
+        service.updateChargeAndEmitEventPostAuthorisation(chargeSpy.getExternalId(), ENTERING_CARD_DETAILS,
+                authCardDetails, null, null, null, null, null);
+
+        verify(mockEventService).emitAndRecordEvent(PaymentDetailsEntered.from(chargeSpy));
+    }
+
+    @Test
+    public void shouldUpdateStateOrProvinceForChargeWhenCountryIsUs() {
```
alexbishop1, line 467: Does this test differ significantly from the previous one?

src/main/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilder.java
```diff
@@ -73,6 +88,7 @@ public static SmartpayOrderRequestBuilder aSmartpayRefundOrderRequestBuilder() {
 
     public SmartpayOrderRequestBuilder(SmartpayTemplateData smartpayTemplateData, PayloadBuilder payloadBuilder, OrderRequestType orderRequestType) {
         super(smartpayTemplateData, payloadBuilder, orderRequestType);
+        this.northAmericanRegionMapper = new NorthAmericanRegionMapper();
```
alexbishop1, line 92: Not sure about using `new` here. If it were injected, we could stub it in the tests.

src/test/java/uk/gov/pay/connector/gateway/smartpay/SmartpayOrderRequestBuilderTest.java
```diff
@@ -117,9 +119,28 @@ public void shouldGenerateValidAuthoriseOrderRequestForAddressWithOptionalFields
         assertEquals(OrderRequestType.AUTHORISE, actualRequest.getOrderRequestType());
     }
 
+    @Test
+    public void shouldGenerateValidAuthoriseOrderRequestForUsAddress() throws Exception {
+
+        Address usAddress = new Address("10 WCB", null, "20500", "Washington D.C.", null, "US");
```
alexbishop1, line 125: If `NorthAmericanRegionMapper` were injected into `SmartpayOrderRequestBuilder`, we could just stub it with `given(mockNorthAmericanRegionMapper.get(usAddress)).willReturn(UsState.WASHINGTON_DC)`, which would mean this test wouldn’t rely on the mapping between the 20500 ZIP code and Washington DC.

src/test/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequestTest.java
```diff
@@ -95,6 +95,46 @@ public void shouldHaveCorrectParametersWithAddress() {
...
+    
+    @Test
+    public void shouldHaveCanadianProvinceOrTerritoryInBillingDetailsWhenProvided() {
+        Address address = new Address();
+        address.setLine1(line1);
+        address.setLine2(line2);
+        address.setCity(city);
+        address.setCountry("CA");
+        address.setPostcode("X0A0A0");
+        authCardDetails.setAddress(address);
```
alexbishop1, line 107: Again, if `NorthAmericanRegionMapper` were injected, we could just stub it rather than having to provide a real Canadian address.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
```diff
@@ -87,6 +89,51 @@ public void shouldGenerateValidAuthoriseOrderRequestForAddressWithMinimumFieldsW
         assertEquals(OrderRequestType.AUTHORISE, actualRequest.getOrderRequestType());
     }
 
+    @Test
+    public void shouldGenerateValidAuthoriseOrderRequestForAddressWithMState() throws Exception {
```
alexbishop1: `M`

src/test/java/uk/gov/pay/connector/northamericaregion/NorthAmericanRegionMapperTest.java
```diff
@@ -0,0 +1,62 @@
...
+        addressEntity.setCountry("CA");
+        addressEntity.setCity("Arctic Region");
+        addressEntity.setPostcode("X0A0A0");
+
+        Optional<? extends NorthAmericaRegion> northAmericaRegion = mapper.getNorthAmericanRegionForCountry(addressEntity);
+
+        assertThat(northAmericaRegion.isPresent(), is (true));
+        assertThat(northAmericaRegion.get(), is (CanadaProvinceOrTerritory.NUNAVUT));
+    }
+}
```
alexbishop1, line 63: You’ll be stunned to read that once again, I’m recommending injecting `UsZipCodeToStateMapper` and `CanadaPostalcodeToProvinceOrTerritoryMapper`.

That way, we could really hone this test down to just checking the logic it’s responsible for (checking if the country is “US” or “CA” and getting the calling the appropriate mapper). So we could end up with tests like this:

```java
private static final String POSTCODE = "I am a postcode!";

@Before
public void setUp() {
    given(mockUsZipCodeToStateMapper.getState(POSTCODE)).willReturn(UsState.WASHINGTON_DC);
    given(mockCanadaPostalcodeToProvinceOrTerritoryMapper.getProvinceOrTerritory(POSTCODE)).willReturn(CanadaProvinceOrTerritory.NUNAVUT);
}

@Test
public void shouldReturnTheCorrectProvinceOrTerritoryForValidAddressInCanada() {
    var Address new Address();
    address.setCountry("CA");
    address.setPostcode(POSTCODE);

    var provinceOrTerritory = mapper.getNorthAmericanRegionForCountry(address);

    assertThat(provinceOrTerritory.isPresent(), is (true));
    assertThat(provinceOrTerritory.get(), is (CanadaProvinceOrTerritory.NUNAVUT));
}

// And a similar test for the US

// And a similar test for a country that’s neither the US or Canada
```

alexbishop1 (changes requested):
I think this is almost there. A few minor suggestions about code style and a couple of slightly larger ones about injecting more so that we can stub things out in tests and not keep repeatedly checking that the 20500 ZIP code is Washington DC.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
nimalank7: Yes. The other test is checking whether an event has been emitted - this test is for whether the ChargeEntity has been updated with the correct state. I did think about combining the tests together but that would be mixing state and behaviour verification together.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: This doesn't work I don't think. In the event that the stateOrProvinceAbbreviation isn't present then the consumer won't be called and so the addressEntity will never be passed into detailsEntity which is incorrect. Regardless of whether stateOrProvinceAbbreviation is present we want the AddressEntity to be set in the CardDetailsEntity. I've done a proposed refactoring.

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
nimalank7: Sure.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Whoops, you’re right. Something like this then:

```java
authCardDetails.getAddress().ifPresent(address -> {
    var addressEntity = new AddressEntity(address);
    northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
            .map(NorthAmericaRegion::getAbbreviation)
            .ifPresent(addressEntity::setStateOrProvince);
    detailsEntity.setBillingAddress(addressEntity);
});
```

Alternatively, we could make it so that `Address` as well as the `AddressEntity` stores the state or province (it’s a bit weird that it’s only on one right now) and do something like this:

```java
authCardDetails.getAddress().ifPresent(address -> {
    northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
            .map(NorthAmericaRegion::getAbbreviation)
            .ifPresent(address::setStateOrProvince);
    });

    var addressEntity = new AddressEntity(address);
    detailsEntity.setBillingAddress(addressEntity);
```

But then that means we’d be mutating something in the `AuthCardDetails`, which I’m not thrilled about.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/northamericaregion/NorthAmericanRegionMapper.java
nimalank7: Done.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
nimalank7: Oops!

src/test/java/uk/gov/pay/connector/northamericaregion/NorthAmericanRegionMapperTest.java
nimalank7: Sure unfortunately I can't do DI because the Smartpay/Worldpay methods constructors are called via static methods and I can't inject using `@Inject` as the classes aren't managed by the Guice container. Similar remarks apply to Stripe.

src/main/java/uk/gov/pay/connector/northamericaregion/CanadaPostalcodeToProvinceOrTerritoryMapper.java
nimalank7: I agree. Originally I couldn't get Mockito to compile but now I've figured it out using the `Answer` interface.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
nimalank7: Yep that's the one I've done.

src/main/java/uk/gov/pay/connector/northamericaregion/UsZipCodeToStateMapper.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -752,10 +756,14 @@ private CardDetailsEntity buildCardDetailsEntity(AuthCardDetails authCardDetails
...
+
         authCardDetails.getAddress().ifPresent(address -> {
-            AddressEntity addressEntity = new AddressEntity(authCardDetails.getAddress().get());
-            processRegionForCanadaOrUs(addressEntity);
+            var addressEntity = new AddressEntity(address);
+            northAmericanRegionMapper.getNorthAmericanRegionForCountry(address)
+                    .map(NorthAmericaRegion::getAbbreviation)
+                    .ifPresent(stateOrProvinceAbbreviation -> {
+                        addressEntity.setStateOrProvince(stateOrProvinceAbbreviation);
+                    });
```
alexbishop1, line 766: This could be just `ifPresent(addressEntity::setStateOrProvince)`.

src/test/java/uk/gov/pay/connector/northamericaregion/NorthAmericanRegionMapperTest.java
alexbishop1: Even if we can’t easily do DI in the gateway-specific order request builders, we could still have this class have `UsZipCodeToStateMapper` and `CanadaPostalcodeToProvinceOrTerritoryMapper` injected (the order request builders would just have to do something like `mapper = new NorthAmericanRegionMapper(new UsZipCodeToStateMapper(), new CanadaPostalcodeToProvinceOrTerritoryMapper())`).

alexbishop1 (approved):
I think this is good to go in. Thanks for your hard work responding to all the comments!

PR #2510, opened 2020-08-25
PP-6960 Support setting MOTO mask properties on gateway accounts

alexbishop1:
Support sending PATCH requests to `/v1/api/accounts/{accountId}` with JSON bodies like this:

```json
{
  "op": "replace",
  "path": "moto_mask_card_number_input",
  "value": true
}
```

… or this:

```json
{
  "op": "replace",
  "path": "moto_mask_card_security_code_input",
  "value": false
}
```

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
@@ -26,6 +26,7 @@ import java.util.Optional;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
+import static java.util.Map.entry;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.CREDENTIALS_GATEWAY_MERCHANT_ID;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_ALLOW_APPLE_PAY;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_ALLOW_GOOGLE_PAY;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
```diff
@@ -88,50 +91,83 @@ public GatewayAccountResponse createGatewayAccount(GatewayAccountRequest gateway
...
+            ),
+            entry(
+                FIELD_MOTO_MASK_CARD_NUMBER_INPUT,
+                (gatewayAccountRequest, gatewayAccountEntity) -> gatewayAccountEntity.setMotoMaskCardNumberInput(gatewayAccountRequest.valueAsBoolean())
+            ),
+            entry(
+                FIELD_MOTO_MASK_CARD_SECURITY_CODE_INPUT,
+                (gatewayAccountRequest, gatewayAccountEntity) -> gatewayAccountEntity.setMotoMaskCardSecurityCodeInput(gatewayAccountRequest.valueAsBoolean())
+            )
+    );
```
oswaldquek, line 170: nice refactor

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
```diff
@@ -274,7 +287,63 @@ public void shouldUpdateAllowMotoFalse() {
...
+        Optional<GatewayAccount> optionalGatewayAccount = gatewayAccountService.doPatch(GATEWAY_ACCOUNT_ID, request);
+        assertThat(optionalGatewayAccount.isPresent(), is(true));
+        verify(mockGatewayAccountEntity).setMotoMaskCardNumberInput(false);
+        verify(mockGatewayAccountDao).merge(mockGatewayAccountEntity);
+    }
+
+    @Test
+    public void shouldUpdateMotoMaskCardSecurityCodeInputTrue() {
+        JsonPatchRequest request = JsonPatchRequest.from(new ObjectMapper().valueToTree(Map.of(
+                "op", "replace",
```
oswaldquek, line 322: Doesn't stop the approval of this PR but I was wondering if we should test the operations of "add" and "remove" too. I realise we haven't tested these operations for the other things though.

oswaldquek (approved):
Optional comment to be addressed

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
alexbishop1: We don’t actually support them (we cheat and just do everything with `"replace"`) so not much point. (Ideally, we’d try to make our JSON Patch implementation more conformant at some stage!)

PR #2521, opened 2020-09-03
PP-7077 Get refunds from Ledger

kbottla:
## WHAT YOU DID
- Added service method to get refund transactions for payment from Ledger.
  This is going to be used to calculate refundability when refunds have been expunged from connector

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paritycheck/LedgerService.java
+++ b/src/main/java/uk/gov/pay/connector/paritycheck/LedgerService.java
@@ -1,8 +1,11 @@
 package uk.gov.pay.connector.paritycheck;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 
 import javax.inject.Inject;
+import javax.ws.rs.ProcessingException;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/paritycheck/LedgerService.java
```diff
@@ -51,12 +51,23 @@ public LedgerService(Client client, ConnectorConfiguration configuration) {
         return getTransactionFromLedger(uri);
     }
 
+    public Optional<RefundTransactionsForPayment> getRefundsForPayment(Long gatewayAccountId, String paymentExternalId) {
```
sfount: As discussed should we change the interface to this method to not give the impression that it would validly return an empty result? Potentially the code would be read clearer if it directly returned a `RefundTransactionsForPayment` and through an error for any other case.

sfount (commented):
Question about the use case/ interface around this method, if this is what we want to go ahead with happy to approve as is

src/main/java/uk/gov/pay/connector/paritycheck/LedgerService.java
kbottla: yes, it would be useful to handle as an exception. empty() optional could be misinterpreted as no refunds for payment. Updated to throw relevant exceptions https://github.com/alphagov/pay-connector/compare/4400b5d5ed0f15c3b562e3430569332d4ea9dd44..ad2a5197a55a13c5893ba738a055a3d03dbd0d90

Only edge case is, Ledger returns 404 if payment is not found but likely to happen in extreme cases (payment events are still in queue due to issues with processing payment or any lag in processing).

SandorArpa (approved):
LOKTM

PR #2522, opened 2020-09-03
PP-7077 Reorganise Ledger classes

kbottla:
## WHAT YOU DID
- Reorganised Ledger related service and models into ledger specific package

Follows: https://github.com/alphagov/pay-connector/pull/2521


SandorArpa (approved):
LOKTM

PR #2524, opened 2020-09-07
PP-7020: Fix env-map db user parameter name

rjbaker:
## WHAT YOU DID
Corrects the name of the database username property provided by the app secret service and mapped to a local environment variable when running on GOV.UK PaaS.


danworth (approved):

rorymalcolm (approved):
💯

PR #2526, opened 2020-09-08
PP-6703 Set cookie header directly for gateway request

stephencdaly:
Worldpay 3DS authorisation requests are currently failing for test accounts because we are setting the machine cookie header twice on requests. We've possibly always been doing this but Worldpay recently started returning an error in this case.

It looks like  we are adding the $Version=1 attribute to the cookie which is part of obsolete  specification rfc2965 for HTTP State Management Mechanism and has been removed in the latest rfc6265. On top of it RequestAddCookies adds the same cookie without $Version making it 2 different cookies.

This solution attempts to set the cookie on the request without a version so that it is only included once.


kbottla (approved):
LGTM!

PR #2540, opened 2020-09-15
PP-7189 Make telephone payment requests idempotent

kbottla:
## WHAT YOU DID
- Avoid creating additional charge (for telephone payment) when another already exists in charges table for a given `provider_id`
  Using `orElse` invokes the supplier function even when the optional value exists which is causing additional records to be created in database.
  With `orElseGet`, supplying function will only be invoked if value is not present
- Added an assertion to ensure only one record is created for multiple requests with same provider_id

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
@@ -88,8 +88,7 @@ public class ChargesApiResource {
     ) {
         return chargeService.findCharge(telephoneChargeCreateRequest)
                 .map(response -> Response.status(200).entity(response).build())
-                .orElse(Response.status(201).entity(chargeService.create(telephoneChargeCreateRequest, accountId).get())
-                .build());
+                .orElseGet(() -> Response.status(201).entity(chargeService.create(telephoneChargeCreateRequest, accountId).get()).build());
     }
 
     @POST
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -88,8 +88,7 @@ public Response createNewTelephoneCharge(
     ) {
         return chargeService.findCharge(telephoneChargeCreateRequest)
                 .map(response -> Response.status(200).entity(response).build())
-                .orElse(Response.status(201).entity(chargeService.create(telephoneChargeCreateRequest, accountId).get())
-                .build());
+                .orElseGet(() -> Response.status(201).entity(chargeService.create(telephoneChargeCreateRequest, accountId).get()).build());
```
danworth, line 91: Looks good Krishna 👍  My only comment it about this `.get()` without checking a value is present. It seems `.create` would return an empty `Optional` if the gateway account cannot be found from the account id for example. Do you think it's necessary to check that the Optional from `.create` has a value?

kbottla: Good point. Although we shouldn’t have the case that request got through publicapi and connector doesn’t have corresponding gateway account, would be good consider the possibility. Will look into it separately and the response we send in these cases

danworth: yeah sounds good 👍

danworth (approved):
Looks good, nice catch 👏

PR #2543, opened 2020-09-21
PP-6994 Remove ChargeService dependency from ChargeRefundService

stephencdaly:
Will require a dependency the other way round so that the ChargeService can retrieve the refunds for a charge in order to build the refund summary for a charge when we also look in ledger for refunds.

Instead, lookup the charge in the resource and pass it into the RefundService method that requires it.


kbottla (dismissed):
LGTM!

sfount (approved):

PR #2544, opened 2020-09-21
PP-6988 Refactor ParityCheckerService

kbottla:
## WHAT YOU DID
- Created separate ChargeParityChecker which is responsible for checking parity between charge and ledger transaction.
  This allows us to create separate RefundParityChecker and keep ParityCheckerService simple


stephencdaly (approved):
LGTM

PR #2545, opened 2020-09-21
PP-6988 Add RefundParityChecker

kbottla:
## WHAT YOU DID
- Added RefundParityChecker which will be used by ExpungeService to parity check on refunds (between connector and ledger)  before removing from connector database.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/service/RefundParityChecker.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/service/RefundParityChecker.java
@@ -1,0 +1,88 @@
+package uk.gov.pay.connector.tasks.service;
+
+import com.google.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+import uk.gov.pay.connector.charge.model.domain.ParityCheckStatus;
+import uk.gov.pay.connector.client.ledger.model.LedgerTransaction;
+import uk.gov.pay.connector.refund.dao.RefundDao;
+import uk.gov.pay.connector.refund.model.domain.RefundEntity;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/tasks/service/RefundParityChecker.java
```diff
@@ -0,0 +1,88 @@
...
+                refundDao.getRefundHistoryByRefundExternalIdAndRefundStatus(refundExternalId, CREATED);
+
+        return refundHistoryForCreatedEvent.map(RefundHistory::getHistoryStartDate);
+    }
+
+    private boolean isEquals(Object value1, Object value2, String fieldName) {
+        if (Objects.equals(value1, value2)) {
+            return true;
+        } else {
+            logger.info("Field value does not match between ledger and connector [field_name={}] {}", fieldName,
```
stephencdaly: Will the {} at the end of this log line cause the StructuredArgument to be substituted into the string rather than appearing as a JSON field?

stephencdaly (commented):
Looks good, just one question about the logging

src/main/java/uk/gov/pay/connector/tasks/service/RefundParityChecker.java
kbottla: good spot! linter complained about unused argument but I guess we need to ignore this. There may be few more instances I substituted 🤦 in the past. Will look into them separately.

stephencdaly: Using String.format is a slightly more annoying way to do it but avoids the lint warning if that is a concern

stephencdaly (approved):

PR #2549, opened 2020-09-22
PP-6994 Use RefundService when calculating refundability

stephencdaly:
Following the same `ChargeEntity` -> `Charge` POJO pattern, add a POJO to represent refunds. These will be buildable from Connector database entities and historic transactions if the database entities have been expunged.

When calculating refundability, instead of directly getting refunds from the database using the RefundDao, use a new method in RefundService, which returns a list of these new Refund objects.

This method will later be updated to retrieve refunds from ledger and combine this set of refunds with the set retrieved from the connector database so that we can handle calculating refundability when refunds are expunged.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -61,8 +61,8 @@ import uk.gov.pay.connector.northamericaregion.NorthAmericaRegion;
 import uk.gov.pay.connector.northamericaregion.NorthAmericanRegionMapper;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 import uk.gov.pay.connector.queue.statetransition.StateTransitionService;
-import uk.gov.pay.connector.refund.dao.RefundDao;
-import uk.gov.pay.connector.refund.model.domain.RefundEntity;
+import uk.gov.pay.connector.refund.model.domain.Refund;
+import uk.gov.pay.connector.refund.service.RefundService;
 import uk.gov.pay.connector.token.dao.TokenDao;
 import uk.gov.pay.connector.token.model.domain.TokenEntity;
 import uk.gov.pay.connector.wallets.WalletType;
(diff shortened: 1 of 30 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -796,10 +801,10 @@ private TokenEntity createNewChargeEntityToken(ChargeEntity chargeEntity) {
     private ChargeResponse.RefundSummary buildRefundSummary(ChargeEntity chargeEntity) {
         ChargeResponse.RefundSummary refund = new ChargeResponse.RefundSummary();
         Charge charge = Charge.from(chargeEntity);
-        List<RefundEntity> refundEntityList = refundDao.findRefundsByChargeExternalId(chargeEntity.getExternalId());
-        refund.setStatus(providers.byName(chargeEntity.getPaymentGatewayName()).getExternalChargeRefundAvailability(charge, refundEntityList).getStatus());
-        refund.setAmountSubmitted(RefundCalculator.getRefundedAmount(refundEntityList));
-        refund.setAmountAvailable(RefundCalculator.getTotalAmountAvailableToBeRefunded(charge, refundEntityList));
+        List<Refund> refundList = refundService.findRefunds(charge);
```
kbottla, line 804: Can we only pass chargeExternalId ?

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -228,22 +230,33 @@ private void checkIfChargeIsRefundableOrTerminate(Charge reloadedCharge, Externa
...
                 .byName(PaymentGatewayName.valueFrom(gatewayAccountEntity.getGatewayName()))
-                .getExternalChargeRefundAvailability(charge, refundEntityList);
+                .getExternalChargeRefundAvailability(charge, refundList);
         checkIfChargeIsRefundableOrTerminate(charge, refundAvailability, gatewayAccountEntity);
 
-        List<RefundEntity> refundEntities = refundDao.findRefundsByChargeExternalId(charge.getExternalId());
+//      @TODO(sfount) is there a reason this currently needs to request new values from the database? This all happens in an @Transactional method
+//                    can we just pass the refund list that's been passed into this method (will the newly created refund be picked up before the
+//                    transaction has commited?         
+        List<Refund> postRefundList = findRefunds(charge);
```
kbottla, line 244: Based on code comments, we may need to query DB to handle refunds submitted in succession (due to double click error or partial refunds) and use refundList passed to this method (to avoid querying ledger again). We can get closest snapshot of refunds before calculating refund availability.

src/main/java/uk/gov/pay/connector/tasks/service/ChargeParityChecker.java
```diff
@@ -243,10 +243,10 @@ private boolean matchCaptureFields(ChargeEntity chargeEntity, LedgerTransaction
     }
 
     private boolean matchRefundSummary(ChargeEntity chargeEntity, LedgerTransaction transaction) {
-        List<RefundEntity> refundsList = refundDao.findRefundsByChargeExternalId(chargeEntity.getExternalId());
+        List<Refund> refundList = refundService.findRefunds(Charge.from(chargeEntity));
```
kbottla, line 246: We can try avoiding querying ledger in future PR as we won't be expunging refunds until charge is expunged

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
sfount: This feature work has been broken up into a number of pieces but the design is for the service to figure out if it needs to go Ledger or not and one of the criteria would be if the charge was "in-flight" or historic, as it will use multiple qualities of the charge the whole charge was passed through. That could be changed if needed but the full use isn't shown in this first PR.

src/main/java/uk/gov/pay/connector/tasks/service/ChargeParityChecker.java
sfount: This will also be delegated to the service so that upstream code doesn't have to worry about how this works underneath, it will be able to figure this out based on the `historic` flag on charges (can potentially be renamed?) ref: https://github.com/alphagov/pay-connector/pull/2549#discussion_r493339417

kbottla (approved):

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
stephencdaly: Okay that makes sense, I'll address that in the next PR when we are querying ledger so that here we only re-retrieve from the database and merge into the our existing list

sfount: Great thanks @kbottla, I realised that we're dealing with JPA transactions + Postgres transactions so figuring out exactly if reading from the database a second time would return new or different data inside the transaction wasn't clear. This sounds reasonable to me, I think this comment can be removed

PR #2550, opened 2020-09-22
PP-6988 Add parity check fields on refunds

kbottla:
## WHAT YOU DID
- Added new fields on refunds and refunds_history tables to be used to record parity check status


SandorArpa (approved):
LOKTM

PR #2551, opened 2020-09-23
PP-6988 Refund DAO methods for expunger

kbottla:
## WHAT YOU DID
- Added refund DAO methods to
   **find refund to expunge** : Returns only the refunds for which charge has been expunged and eligible for expunging ( created before configured days and not parity checked recently)
   **expunge refunds** : expunged refund (removes records from emitted_events, refunds_history and refunds table)


stephencdaly (approved):
Looks good 👍

PR #2553, opened 2020-09-23
PP-6994 Use external refund status to calculate refundability

stephencdaly:
Refunds that we get from ledger have a status corresponding to an external status in connector.

Use the external status for charges from the connector database when determining which refunds should be included in the refundability calculation so that it will work for refunds retrieved from ledger.


sfount (approved):
I've checked the refund states against these and can confirm the logic is sound (https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/refund/model/domain/RefundStatus.java#L17-L20). This also lines up with Ledger mappings which is assured by transaction expunge checks as well as various pact/ integration tests. Passing all existing tests should reenforce this.

PR #2554, opened 2020-09-23
PP-6988 Parity check refund and emit events

kbottla:
## WHAT YOU DID
- Added functionality to parity check refund for expunging and backfill if parity check fails


stephencdaly (approved):
Looks good!

PR #2557, opened 2020-09-24
PP-6994 Use expunged refunds from ledger when calculating refundability

stephencdaly:
If the charge is historic, make a request to ledger to get any refunds that exist there. If the refunds also exist in the connector database, prefer the record from the database as the refund may be in-flight.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundsResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundsResourceIT.java
@@ -8,11 +8,16 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 import uk.gov.pay.commons.model.ErrorIdentifier;
 import uk.gov.pay.connector.app.ConnectorApp;
+import uk.gov.pay.connector.charge.model.ChargeResponse;
+import uk.gov.pay.connector.client.ledger.model.LedgerTransaction;
+import uk.gov.pay.connector.common.model.api.ExternalRefundStatus;
 import uk.gov.pay.connector.it.base.ChargingITestBase;
 import uk.gov.pay.connector.it.dao.DatabaseFixtures;
 import uk.gov.pay.connector.junit.DropwizardConfig;
 import uk.gov.pay.connector.junit.DropwizardJUnitRunner;
+import uk.gov.pay.connector.refund.model.domain.RefundStatus;
 
+import java.time.ZonedDateTime;
 import java.util.HashMap;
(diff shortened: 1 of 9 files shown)


src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundsResourceIT.java
```diff
@@ -286,22 +293,87 @@ public void shouldFailRequestingARefund_whenAPartialRefundMakesTotalRefundedAmou
         assertRefundsHistoryInOrderInDBForSuccessfulOrPartialRefund(defaultTestCharge);
     }
 
+    @Test
+    public void shouldFailRequestingARefundForHistoricCharge_whenPartialRefundAmountGreaterThanRemainingAmount_whenExistingPartialRefundsHaveBeenExpunged() throws Exception {
```
sfount, line 297: 😮

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -616,6 +640,55 @@ public void shouldFindRefundsGivenValidCharge() {
...
+        
+        var refundTransactionsForPayment = aValidRefundTransactionsForPayment()
+                .withParentTransactionId(charge.getExternalId())
+                .withTransactions(List.of(ledgerRefund1, ledgerRefund2))
+                .build();
+        when(mockLedgerService.getRefundsForPayment(charge.getGatewayAccountId(), charge.getExternalId()))
+                .thenReturn(refundTransactionsForPayment);
+
+        List<Refund> refunds = refundService.findRefunds(charge);
+        assertThat(refunds, hasSize(3));
```
sfount, line 689: Would it be useful to add an assertion here to assert that the in-flight refund is always selected? It looks like the `refundInDatabaseAndLedgerId` have different mocked statuses so potentially an assertion on the status of that entry returned.

sfount (dismissed):
Looks great to me 👍  I think it would be good to also add ledger/ connector db  path scenarios to the integration tests of critical code paths that use this service but that would be ideal as a separate PR

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
stephencdaly: ah yeh good spot, I intended in those assertions but then must have got distracted

stephencdaly: added in those assertions

sfount (approved):

PR #2558, opened 2020-09-24
PP-6988 Expunge refunds

kbottla:
## WHAT YOU DID
- Starts expunging refunds (if `EXPUNGE_REFUNDS_ENABLED` is enabled) when expunge task is invoked.


SandorArpa (approved):
LOKTM

PR #2562, opened 2020-09-25
PP-7224 Fix sweeping emitted refunds events for historic charges

kbottla:
## WHAT YOU DID
- Emitted events sweeper for refunds currently relies on related charge being in connector database.
  But we should be able to emit refund events for expunged charges. This fixes backfilling refund events for historic charges
- Also took opportunity to split full backfill (charges + refunds) for a given event. For a charge event, only payment events are emitted and for refund event, only refund events are emitted. This can be improved further to  emit specific event only.
- In case of any exception when processing event, event is marked to not retry instead of throwing exception. The emitted events sweeper continues emitting pending events.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
+++ b/src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
@@ -6,12 +6,10 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.app.config.EmittedEventSweepConfig;
-import uk.gov.pay.connector.charge.dao.ChargeDao;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/events/EmittedEventsBackfillService.java
```diff
@@ -81,9 +86,9 @@ public void backfillEvent(EmittedEventEntity event) {
                     event.getEventType(),
                     event.getEventDate()
             );
-            throw e;
+            event.setDoNotRetryEmitUntil(ZonedDateTime.now(UTC).minusSeconds(doNotRetryEmittingEventUntilDurationInSeconds));
```
SandorArpa, line 89: Is it intentional that the do not retry emit time is in the past?

SandorArpa (changes requested):

SandorArpa (dismissed):
LGTM

SandorArpa (approved):
LOKTM

PR #2569, opened 2020-09-29
PP-7229 Remove Joda-Time

alexbishop1:
Stop using Joda-Time’s `DateTime` to represent the requested date to capture a payment when a capture request is created to send to Worldpay.

While the direct `java.time` equivalent of Joda-Time’s `DateTime` is `ZonedDateTime`, use `LocalDate` instead because only the date component is actually used. Retain the behaviour of determining the current date based what the day is right now in the UTC time zone.

Stop setting a capture date altogether for authorisation requests to Worldpay because it’s never used and doesn’t end up in the final payload sent to Worldpay.

In addition to the usual automated test suite, also ran the contract test that connects to Worldpay and inspected the generated capture request XML to make sure it looks good.

Also remove `jackson-datatype-joda` dependency from the POM because it’s not used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -266,11 +266,6 @@
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.datatype</groupId>
-            <artifactId>jackson-datatype-joda</artifactId>
-            <version>${jackson.version}</version>
-        </dependency>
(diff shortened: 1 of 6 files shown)


rorymalcolm (approved):
this LGTM - nice touch on removing the withDate calls

PR #2573, opened 2020-10-01
PP-7242 Use new CardExpiryDate type instead of strings

alexbishop1:
Use the new `CardExpiryDate` instead of strings in most places in connector to represent card expiry dates.

This does not change how card expiry dates are represented in the API (in either requests or responses), nor does it change how card expiry dates are formatted when sent to payment gateways, nor does it change how card expiry dates are stored in the database.

Due to the limitations of our current validation mechanism, there is a change what happens when an invalid expiry date is supplied to `/v1/frontend/charges/{chargeId}/cards`. Previously the validator would catch it and return a 400 response with a friendly error message. Now it will fail when Jersey tries to deserialise it into a `CardExpiryDate` object and a 400 response will be returned with an exception message. This should be OK because only frontend calls this endpoint and only for regular card payments (not Apple Pay or Google Pay) and it has its own validation (tightened in PP-7080) so should never send invalid expiry dates. The format required by `CardExpiryDate` is actually stricter than the previous validation that was in place for this endpoint but since frontend should only send valid expiry dates this make no difference.

`/v1/api/accounts/{accountId}/telephone-charges` continues to deserialise the expiry date to a string. It is only converted to a `CardExpiryDate` object later, after the endpoint has performed its own validation, which allows card expiry dates in the exact same format as required by `CardExpiryDate` (and now it actually uses `CardExpiryDate`’s code to check the format). Therefore the behaviour of this endpoint when supplied with an invalid expiry date should not have changed.

The mechanism by which 2-digit years are made into 4-digit years by prefixing them with “20” when sending them in authorisation requests to Worldpay and Smartpay remains but this logic is now encapsulated in `CardExpiryDate` rather than in the Freemarker XML templates.

Prior to this change, an expiry date associated with a payment was converted to a string when retrieved from the database. Now it is immediately converted to a `CardExpiryDate` object, which will fail if the expiry date in the database is not in the required format. A check of all the expiry dates stored in the production database a couple of days ago confirmed that all are in the correct format and we do not expect this to have changed since then due to the existing validation we have in place in connector and frontend.

Expiry dates in payments retrieved from ledger remain as strings, as is the convention in this area of code (in addition, we haven’t verified that every expiry date in the ledger database matches the required format). Expiry dates that are sent to ledger in events are converted to strings prior to the event object being serialised to JSON; again, this matches the existing conventions in this area of code.

In addition to the automated tests, I tested this by creating a payment, filling in the card details and checking the database and the API response to make sure the expiry date matched the existing format. I also threw a payload at `/v1/frontend/charges/{chargeId}/cards` to verify that it deserialised valid expiry dates and gave an error for invalid ones.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -4,6 +4,7 @@ import com.google.inject.persist.Transactional;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.commons.model.CardExpiryDate;
 import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.commons.model.charge.ExternalMetadata;
 import uk.gov.pay.connector.app.CaptureProcessConfig;
@@ -172,7 +173,7 @@ public class ChargeService {
                     LastDigitsCardNumber.ofNullable(telephoneChargeRequest.getLastFourDigits().orElse(null)),
                     FirstDigitsCardNumber.ofNullable(telephoneChargeRequest.getFirstSixDigits().orElse(null)),
                     telephoneChargeRequest.getNameOnCard().orElse(null),
-                    telephoneChargeRequest.getCardExpiry().orElse(null),
+                    telephoneChargeRequest.getCardExpiry().map(CardExpiryDate::valueOf).orElse(null),
                     telephoneChargeRequest.getCardType().orElse(null),
                     null
             );
--- a/src/main/java/uk/gov/pay/connector/charge/validation/telephone/CardExpiryValidator.java
+++ b/src/main/java/uk/gov/pay/connector/charge/validation/telephone/CardExpiryValidator.java
@@ -1,20 +1,15 @@
 package uk.gov.pay.connector.charge.validation.telephone;
 
+import uk.gov.pay.commons.model.CardExpiryDate;
+
 import javax.validation.ConstraintValidator;
 import javax.validation.ConstraintValidatorContext;
-import java.util.regex.Pattern;
 
 public class CardExpiryValidator implements ConstraintValidator<ValidCardExpiryDate, String> {
 
-    private Pattern pattern = Pattern.compile("(0[1-9]|1[0-2])\\/[0-9]{2}");
-
     @Override
     public boolean isValid(String value, ConstraintValidatorContext context) {
-
-        if(value == null) {
-            return true;
-        }
-        
-        return pattern.matcher(value).matches();
+        return value == null || CardExpiryDate.CARD_EXPIRY_DATE_PATTERN.matcher(value).matches();
     }
+
 }
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentNotificationCreatedEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentNotificationCreatedEventDetails.java
@@ -1,12 +1,18 @@
 package uk.gov.pay.connector.events.eventdetails.charge;
 
+import uk.gov.pay.commons.model.CardExpiryDate;
 import uk.gov.pay.commons.model.Source;
 import uk.gov.pay.commons.model.charge.ExternalMetadata;
+import uk.gov.pay.connector.cardtype.model.domain.CardBrandLabelEntity;
+import uk.gov.pay.connector.charge.model.CardDetailsEntity;
+import uk.gov.pay.connector.charge.model.FirstDigitsCardNumber;
+import uk.gov.pay.connector.charge.model.LastDigitsCardNumber;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.events.eventdetails.EventDetails;
 
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 
 public class PaymentNotificationCreatedEventDetails extends EventDetails {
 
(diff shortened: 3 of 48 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -172,7 +173,7 @@ public ChargeService(TokenDao tokenDao,
                     LastDigitsCardNumber.ofNullable(telephoneChargeRequest.getLastFourDigits().orElse(null)),
                     FirstDigitsCardNumber.ofNullable(telephoneChargeRequest.getFirstSixDigits().orElse(null)),
                     telephoneChargeRequest.getNameOnCard().orElse(null),
-                    telephoneChargeRequest.getCardExpiry().orElse(null),
+                    telephoneChargeRequest.getCardExpiry().map(CardExpiryDate::valueOf).orElse(null),
```
danworth, line 176: I think you discuss this in Chapter 4 of your commit message but thought I'd ask just to confirm. Since you've updated the `CardExpiryValidator` to perform the same validation as that on `CardExpiryDate` which is used on the `TelephoneChargeCreateRequest.cardExpiry` can this not be changed from a `String` to a `CardExpiryDate` type also, so it is deserialised straight into `CardExpiryDate`? Or does this effect the error responses/handling to the client?
https://github.com/alphagov/pay-connector/blob/f97b14bdd7a40d0ccb74facc5540b174161da172/src/main/java/uk/gov/pay/connector/charge/model/telephone/TelephoneChargeCreateRequest.java#L53

src/main/java/uk/gov/pay/connector/charge/validation/telephone/CardExpiryValidator.java
```diff
@@ -1,20 +1,15 @@
...
-
     @Override
     public boolean isValid(String value, ConstraintValidatorContext context) {
-
-        if(value == null) {
-            return true;
-        }
-        
-        return pattern.matcher(value).matches();
+        return value == null || CardExpiryDate.CARD_EXPIRY_DATE_PATTERN.matcher(value).matches();
```
danworth, line 12: 👍

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentNotificationCreatedEventDetails.java
```diff
@@ -55,44 +61,24 @@ private PaymentNotificationCreatedEventDetails(Long gatewayAccountId, Long amoun
...
-                    expiryDate,
-                    cardBrand,
-                    cardBrandLabel,
-                    charge.getGatewayAccount().isLive(),
-                    charge.getGatewayAccount().getGatewayName(),
-                    charge.getExternalMetadata().map(ExternalMetadata::getMetadata).orElse(null),
-                    charge.getSource(),
-                    charge.getLanguage().toString());
+        Optional<CardDetailsEntity> cardDetails = Optional.ofNullable(charge.getCardDetails());
+        return new PaymentNotificationCreatedEventDetails(charge.getGatewayAccount().getId(),
```
danworth, line 65: Could do with a builder but that's for another day :)

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
```diff
@@ -44,8 +39,8 @@ public static StripePaymentMethodRequest of(CardAuthorisationGatewayRequest requ
     @Override
     protected Map<String, String> params() {
         Map<String, String> localParams = new HashMap<>();
-        localParams.put("card[exp_month]", authCardDetails.expiryMonth());
-        localParams.put("card[exp_year]", authCardDetails.expiryYear());
+        localParams.put("card[exp_month]", Integer.valueOf(authCardDetails.getEndDate().getTwoDigitMonth()).toString());
```
danworth, line 42: Wonder why one is required as an `Integer` and the other a `String` 😕

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -10,7 +10,7 @@
                 <VISA-SSL>
                     <cardNumber>${authCardDetails.cardNo}</cardNumber>
                     <expiryDate>
-                        <date month="${authCardDetails.endDate?split('/')?first}" year="20${authCardDetails.endDate?split('/')?last}"/>
+                        <date month="${authCardDetails.endDate.twoDigitMonth?xml}" year="${authCardDetails.endDate.fourDigitYear?xml}"/>
```
danworth, line 13: Having the separate `twoDigitMonth` and `twoDigitYear` is much nicer 👍

danworth (approved):
Having reviewed the changes they seems to match the plot of your excellently written novella.

These changes seem very well considered and tested; by both automatic testing and supplemented by some more manual QA.  Thought has been given to how this could impact the api response and it appears it should not do so in any detrimental or unacceptable manner. Changing from a `String` to a custom type should make things better going forward.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Yes, the reasoning was that it would change the failure mode for an invalid expiry date from being a validation error to a Jersey deserialisation error, which would change the response from connector, which I thought would change the response from public API, which we don’t want to change because that would change what services see.

However, I’ve just checked and it seems public API validates the expiry date before it sends the request to connector (using the same format as `CardExpiryDate`). Therefore, we should be safe to change this. I’ll have a look at doing that in a follow-up PR.

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentMethodRequest.java
alexbishop1: They’re actually both `Strings` (`toString()` is called on the `Integer`). The intermediate `Integer` object is to strip the leading zero from months that have one, which matches the existing behaviour. The year, on the other hand, is always sent as 2 digits. This is a bit inconsistent but I didn’t want to change what we send to gateways.

PR #2578, opened 2020-10-02
PP-7242 Use CardExpiryDate in TelephoneCreateChargeRequest

alexbishop1:
In `TelephoneChargeCreateRequest`, represent the card expiry date as a `CardExpiryDate` rather than a string.

This changes what happens if an invalid card expiry date is provided to `/v1/api/accounts/{accountId}/telephone-charges` (but what is considered to be invalid has not changed).

Previously, an invalid card expiry date would cause the request to fail validation and a 400 response with a friendly error message would be returned.

Now Jersey immediately tries to deserialise the card expiry date into a `CardExpiryDate` object. `CardExpiryDate` performs its own validation checks when it’s instantiated, so if an invalid card expiry date is provided (and what’s considered to be invalid hasn’t changed), it will fail to deserialise and a 400 response with an exception message will be returned.

However, public API already validates the card expiry date (using the same format as `CardExpiryDate`) so a request with an invalid card expiry date should never get this far. If it does (say, because the validation in public API changes), the request will still fail and return a 400 response, just with a different error message.

Since as discussed above `CardExpiryDate` will not allow invalid instances of itself to be created, some external validation code can be removed.

A request that does not contain a card expiry date (either no `card_expiry` property or a null `card_expiry` property) is still valid.


src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceTelephonePaymentsIT.java
```diff
@@ -362,16 +362,6 @@ public void shouldReturn422ForInvalidCardType() {
                 .body("message[0]", is("Field [card_type] must be either master-card, visa, maestro, diners-club, american-express or jcb"));
     }
 
-    @Test
-    public void shouldReturn422ForInvalidCardExpiryDate() {
```
oswaldquek, line 366: shouldn't this test be changed to return a 400 response instead of it being deleted altogether?

alexbishop1: That’s a very good point. I’ll update the test and add it back in.

oswaldquek (approved):

PR #2584, opened 2020-10-06
PP-6994 Do not re-request ledger refunds during refund

stephencdaly:
After making a network request to get the external refund-ability, we were re-querying the refunds in the database to ensure there had been no changes in the meantime.

When the change was made to also get refunds that only exist in ledger for the refund-ability calculation, in order to allow us to expunge refunds from connector, this was changed to also re-query ledger.

We shouldn't need to re-query ledger as the expunged refunds will be in a terminal state - with the exception of being able to be updated by an admin.

We should however only really care if the amount refundable has decreased rather than increased. And in the case of an admin updating the status of a historic refund from "error" -> "success", we would expect the external refund-ability check with the payment provider to be unsuccessful.

It could also be possible, depending on how long we wait before expunging refunds in the future, that a refund is expunged before we query the database a second time. If this is the case, we are fine to continue using the record we obtained the first time without re-querying ledger as it is not possible that the amount available to refund would have decreased, even if the status of that refund had changed in the meantime.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
+++ b/src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
@@ -26,7 +26,9 @@ import uk.gov.pay.connector.usernotification.service.UserNotificationService;
 import javax.inject.Inject;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -248,13 +250,17 @@ private long validateRefundAndGetAvailableAmount(Charge charge,
...
+        List<RefundEntity> databaseRefunds = findNotExpungedRefunds(charge.getExternalId());
 
-//      @TODO(sfount) is there a reason this currently needs to request new values from the database? This all happens in an @Transactional method
-//                    can we just pass the refund list that's been passed into this method (will the newly created refund be picked up before the
-//                    transaction has commited?         
-        List<Refund> postRefundList = findRefunds(charge);
+        // Add in any new or updated refunds. If a refund has been expunged since we last queried and so no longer
+        // exists in the database, continue using the existing record from the original refund list.
+        Map<String, Refund> updatedRefunds = refundList.stream().collect(Collectors.toMap(Refund::getExternalId, refund -> refund));
+        databaseRefunds.forEach(refundEntity -> updatedRefunds.put(refundEntity.getExternalId(), Refund.from(refundEntity)));
```
sfount: Nice solution! I wonder if encapsulating this in a private service method would help the readability? (I've fully understood everything here so this is totally up to you, this is great code as is). 

It might have an interface like
```java
// List<Refund>
var updatedRefunds = updateInFlightRefunds(refundList)
```

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -547,6 +547,66 @@ public void shouldFailWhenAmountAvailableForRefundWithCorporateSurchargeMismatch
...
+        when(mockRefundDao.findRefundsByChargeExternalId(chargeEntity.getExternalId()))
+                .thenReturn(List.of(refundExpungedSinceWeFirstChecked))
+                .thenReturn(List.of(newlyCreatedRefund));
+
+        var refundTransactionsForPayment = aValidRefundTransactionsForPayment()
+                .withTransactions(List.of(refundObtainedFromLedger))
+                .withParentTransactionId(externalChargeId).build();
+        when(mockLedgerService.getRefundsForPayment(accountId, externalChargeId)).thenReturn(refundTransactionsForPayment);
+
+        expectedException.expect(RefundException.class);
```
sfount, line 600: Nice :+1:

sfount (dismissed):
LGTM 👍

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
stephencdaly: good idea, extracted

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -263,6 +264,16 @@ private long validateRefundAndGetAvailableAmount(Charge charge,
         return availableToBeRefunded;
     }
 
+    private List<Refund> checkForNewRefunds(Charge charge, List<Refund> refundList) {
+        List<RefundEntity> databaseRefunds = findNotExpungedRefunds(charge.getExternalId());
+
+        // Add any new or updated refunds to our existing list. If a refund has been expunged since we last queried and
```
sfount, line 270: Nice one, splitting the explanation into two parts makes this really easy to understand

sfount (approved):

PR #2585, opened 2020-10-06
PP-6988 Use native query to update refund parity check status

kbottla:
## WHAT YOU DID
- Uses native sql query to update refund parity check status as using JPA to update refund also adds additional refunds_history record. As we generate events to Ledger based on refunds_history and history timestamp, additional records in refund_history causes duplicate events to be emitted to ledger but with different timestamps.
- Also added check to fail parity check on refund when gateway account is not available

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/dao/RefundDaoJpaIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/dao/RefundDaoJpaIT.java
@@ -23,6 +23,7 @@ import static org.apache.commons.lang3.RandomUtils.nextLong;
 import static org.hamcrest.CoreMatchers.hasItems;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasEntry;
+import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.Matchers.oneOf;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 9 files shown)


SandorArpa (dismissed):
OKTM

src/test/java/uk/gov/pay/connector/it/dao/RefundDaoJpaIT.java
```diff
@@ -445,4 +446,26 @@ public void getRefundHistoryByRefundExternalId_shouldReturnResultsCorrectly() {
...
+        assertThat(refundHistoryList.size(), is(1));
+
+        ZonedDateTime parityCheckDate = ZonedDateTime.now(UTC);
+        refundDao.updateParityCheckStatus(refundEntity.getExternalId(), parityCheckDate, MISSING_IN_LEDGER);
+
+        mayBeRefundEntityFromDB = refundDao.findByExternalId(refundEntity.getExternalId());
+        refundHistoryList = refundDao.getRefundHistoryByRefundExternalId(refundEntity.getExternalId());
+        assertThat(mayBeRefundEntityFromDB.get().getParityCheckDate().toLocalDateTime(), is(parityCheckDate.toLocalDateTime()));
+        assertThat(mayBeRefundEntityFromDB.get().getParityCheckStatus(), is(MISSING_IN_LEDGER));
+        assertThat(refundHistoryList.size(), is(1));
```
sfount, line 469: Great assertion 👍  that should guard against future JPA shenanigans

sfount (approved):
LGTM

PR #2586, opened 2020-10-06
PP-6988 Fix parity checker for card expiry date

kbottla:
## WHAT YOU DID
- Use Optional.ofNullable to map and parity check expiry date as date may not be available for wallet payments


stephencdaly (approved):

alexbishop1:
Whoops.

PR #2589, opened 2020-10-07
PP-6988 Refactor expunge service

kbottla:
## WHAT YOU DID

- We are using stream to loop through to find and expunge n (no of charges/refunds) records from connector.
  This is causing *ExpungeService to query for expungeable charges/refunds n times even when there are
  no records to expunge or the records available to expunge are less than n.

- Replaced IntStream with loop to stop trying to find & expunge when there are no records returned.


SandorArpa (approved):
LOKTM

PR #2590, opened 2020-10-08
PP-7253 Add consumer pact tests with ledger as provider

stephencdaly:
Add configuration for running pact consumer tests.

Replace unit tests that use templates with pact tests with ledger as the
provider.


kbottla (approved):
LGTM!

PR #2592, opened 2020-10-08
PP-7253 Fix Jenkins file

stephencdaly:


kbottla (approved):

sfount (approved):
These are then used in `pact:publish`, looks good 👍

PR #2598, opened 2020-10-08
PP-6989 Add gateway error states to events affecting refundabiltiy

kbottla:
## WHAT YOU DID
- AUTHORISATION_ERROR, AUTHORISATION_TIMEOUT, AUTHORISATION_UNEXPECTED_ERROR states are terminal states which affects refundability. But these states are now appearing as non-terminal (as per payment gateway state transition graph) due to new states introduced for cleanup up job. As a result, refundability events are not generated for ERROR states and refund_summary status in ledger ended up as 'pending' instead of 'unavailable'.

  This has majorly impacted expunge process, as refund_summary.status doesn't match with ledger and connector, but little or no impact for services. Services using API are getting incorrect refund_summary.status in search results.

- Added error states to events affecting refundability. For existing charges failing parity check, new refundability events will be emitted by connector which also fixes ledger.


SandorArpa (approved):
LOKTM

PR #2600, opened 2020-10-09
PP-6989 Refund expunger moves "charge exists" check from SQL

sfount:
Move the check to see if a charge exists for a given refund that will be
expunged to refund expunge service. This moves it out of the DAO layer
which allows us to remove the join between refunds and charges.

The hypothesis is that this left outer join is complex as the buffer
gets recalculated every time a refund is removed from the bottom of the
table.

If this change has no impact on RDS CPU utilisation, it should be
reverted.


kbottla (approved):
LGTM!

PR #2606, opened 2020-10-12
PP-6989 Parity check email only for successful charges

kbottla:
## WHAT YOU DID
- Parity check email only when the charges state is successful. Frontend currently sends a separate PATCH request to update email, and email is sent to ledger after an authorisation attempt (included in PAYMENT_DETAILS_ENTERED event). If charge is cancelled or expired before authorisation, email is never sent to ledger which results into failed parity check.

  Ignore email in this case (loses data which could potential help services to locate failed user attempts), or 
  to send separate event to ledger when email is patched (but we can't distinguish service pre-filled email to patch event for backfilling).


kbottla:
To be addressed by PP-7276

PR #2610, opened 2020-10-13
PP-7276 Emit email collected event

kbottla:
## WHAT YOU DID
- Emits separate event to ledger when email is patched by frontend. This is so email can be sent to ledger
  even when user cancels payment during authoration operation.
  When a charge is cancelled by user, authorisation operation in progress fails on invalid state transition and no further events related to authorisation are emitted to ledger.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -4,7 +4,6 @@ import com.google.inject.persist.Transactional;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import uk.gov.pay.commons.model.CardExpiryDate;
 import uk.gov.pay.commons.model.SupportedLanguage;
 import uk.gov.pay.commons.model.charge.ExternalMetadata;
 import uk.gov.pay.connector.app.CaptureProcessConfig;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -321,6 +321,7 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
                 .map(chargeEntity -> {
                     if (chargePatchRequest.getPath().equals(ChargesApiResource.EMAIL_KEY)) {
                         chargeEntity.setEmail(sanitize(chargePatchRequest.getValue()));
+                        eventService.emitAndRecordEvent(UserEmailCollected.from(chargeEntity));
```
sfount, line 324: As this method is `@Transactional`, will the charge still be updated if the emit and record event method throws anything other than `QueueException` (for some reason I can't think of off the top of my head).

Is it worth even further guarding against this failing because the event failed or do we think that will make the exception handling hard to follow?

stephencdaly (approved):
Looks good to me

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: For any exceptions transaction will be rolled back. 

Exceptions thrown from downstream code are `ChargeEventNotFoundRuntimeException` or any `runtime exception` (if QueueException can't handle the expected in event service).

`ChargeEventNotFoundRuntimeException` shouldn't happen as it is raised only when charge doesn't CREATED or ENTERING_CARD_DETAILS event.

We can handle all exceptions in EventService instead of only QueueException to be extra safe which can also cover exception handling for `PaymentDetailsEntered` event. (will raise separate PR if we would like go with this)

PR #2611, opened 2020-10-13
PP-6989 Allow parity checker to pass if created_date within 5s

stephencdaly:
Due to a historical issue with the clock of an instance of connector becoming out of sync before we were using the database to set the timestamp, the CREATED and ENTERING CARD DETAILS event dates are out of order for some charges causing ledger to store incorrect date.

Allow for the difference in the created_date to be up to 5s to allow these charges to be expunged. The precision of this time for this limited number of charges does not seem important enough to correct this data.


SandorArpa (approved):
LGTM

PR #2612, opened 2020-10-14
PP-7277 include gateway transaction id in cancelled by user event

SandorArpa:
## WHAT YOU DID
- due to the nature of how gateway transaction id is emitted to Ledger, when user cancels a payment the aforementioned id is stored in Connector but not sent to Ledger. Thus causing a disparity between the two systems. This change ensures that the id is in Ledger and we can expunge those charges from Connector.
- also includes the producer part of a queue pact test


kbottla (approved):
LGTM!

PR #2613, opened 2020-10-14
PP-7276 Add UserEmailCollected pact state

kbottla:
## WHAT YOU DID
- Add new pact state for `user email collected` event
- removed email from `payment details entered` event as email is included in separate event


src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
```diff
@@ -134,8 +135,9 @@ public static PaymentEvent createPaymentEvent(ChargeEventEntity chargeEvent, Cla
                 return PaymentNotificationCreated.from(chargeEvent);
             } else if (eventClass == CancelledByUser.class) {
                 return CancelledByUser.from(chargeEvent);
-            }
-            else {
+            } else if (eventClass == UserEmailCollected.class) {
+                return UserEmailCollected.from(chargeEvent.getChargeEntity());
```
sfount, line 139: This commit also adds an entry to the event factory which is required for any event with charge body details (in this case the charge email). 

This code will be used any time the event has to be reconstructed (it initially failed to emit or a payment is being backfilled). 

Having this included in the PR/ commit message would be useful (if it's not inconvenient) for reading back through the code. We can have a think about how we might include this in documentation/ team manuals to make sure the knowledge is shared (I'm just reminding myself of it after months!).

sfount (approved):
LGTM 👍

PR #2617, opened 2020-10-15
PP-6501 Log more about 3DS data sent to gateways when authorising

alexbishop1:
When we send an authorisation request to a gateway, we currently log something like:

“Authorisation with billing address for abcdefghijklmnopqrstuvwxyz (worldpay xxx) for…”

The word “with” can be “without” if appropriate.

Enhance this to also include more information about what additional data used for 3D Secure we’re sending to the payment gateway, if any.

In all the below examples, each “with” will be “without” as appropriate.

Worldpay:

“Authorisation with billing address and with 3DS data and with device data collection result for abcdefghijklmnopqrstuvwxyz (worldpay xxx) for…”

Stripe:

“Authorisation with billing address for abcdefghijklmnopqrstuvwxyz (stripe xxx) for …”

Smartpay:

“Authorisation with billing address and with 3DS data for abcdefghijklmnopqrstuvwxyz (smartpay xxx) for…”

ePDQ:

“Authorisation with billing address and with 3DS data and with 3DS2 data for abcdefghijklmnopqrstuvwxyz (epdq xxx) for…”

Sandbox:

“Authorisation for abcdefghijklmnopqrstuvwxyz (sandbox xxx) for…”

If a gateway account’s 3DS configuaration is changed close to an authorisation request being sent to the gateway, it’s possible we might misreport whether more 3D Secure data was being sent or not. However, this will be exceptionally rare so isn’t worth the effort to fix.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,7 @@
         <commons-lang3.version>3.11</commons-lang3.version>
         <jjwt.version>0.11.2</jjwt.version>
         <pact.version>3.6.15</pact.version>
+        <mockito.version>3.5.13</mockito.version>
         <PACT_BROKER_URL/>
         <PACT_BROKER_USERNAME/>
         <PACT_BROKER_PASSWORD/>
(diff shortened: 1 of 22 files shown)


rorymalcolm (approved):
lgtm, concerns relieved!

PR #2620, opened 2020-10-19
PP-6989 Include card details on payment created event for expired/user cancelled charges

kbottla:
## WHAT

- When a service prefills billing address, card holder details are included in payment created event details.
  During the backfilling, we only include card holder details on payment created events if charge has not gone through authorisation to avoid including details when not necessary.
  Payment state transition graph is used to get possible states post authorisation which also included USER_CANCELLED and EXPIRED states. But these states are also possible before any authorisation activity and to be excluded as states post authorisation


SandorArpa (approved):
LGTM

PR #2622, opened 2020-10-20
PP-6991: Manually set the JVM TTL for DNS Name Lookups

oswaldquek:
According to the Worldpay Payment Gateway best practice guide
(https://developer.worldpay.com/docs/wpg/reference/wpgbestpractice), domains
should only be cached "within the published Time To Live, in order to ensure
that you respond to DNS changes in a timely manner". From further
correspondence with Worldpay, the time-to-live of a DNS
record is an attribute of the individual result, and is sent as part of the DNS
lookup protocol. It is not published separately, because it can change from
time to time and from record to record, but it is typically of the order of a
few minutes.

Our current TTL (networkaddress.cache.ttl) is set based on adoptopenjdk docker
image we use, currently set to 15 seconds. This disadvantage of relying on the
adoptopenjdk docker image is that the TTL might fail silently if we switch to
another JDK distro. It therefore seems pertinent to set it manually here.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
@@ -86,6 +86,8 @@ public class ConnectorApp extends Application<ConnectorConfiguration> {
 
     @Override
     public void initialize(Bootstrap<ConnectorConfiguration> bootstrap) {
+        java.security.Security.setProperty("networkaddress.cache.ttl" , "15");
+        
         bootstrap.setConfigurationSourceProvider(
                 new SubstitutingSourceProvider(bootstrap.getConfigurationSourceProvider(),
                         new EnvironmentVariableSubstitutor(NON_STRICT_VARIABLE_SUBSTITUTOR)


oswaldquek:
I've tested locally by 
* Adding `logger.error("The networkaddress.cache.ttl from Security.getProperty is " + Security.getProperty("networkaddress.cache.ttl"));` in `ConnectorApp.java`.
* Spinning up a local version of docker connector and viewing the logs to verify that `Security.getProperty("networkaddress.cache.ttl")` has value `15`. One can do this for other values too to verify.

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -103,6 +103,8 @@ public DataSourceFactory getDataSourceFactory(ConnectorConfiguration configurati
         bootstrap.addCommand(new RenderStateTransitionGraphCommand());
         bootstrap.getObjectMapper().getSubtypeResolver().registerSubtypes(LogstashConsoleAppenderFactory.class);
         bootstrap.getObjectMapper().getSubtypeResolver().registerSubtypes(GovUkPayDropwizardRequestJsonLogLayoutFactory.class);
+
+        java.security.Security.setProperty("networkaddress.cache.ttl" , "15");
```
alexbishop1: Maybe put it right at the beginning of the method in case any of the other initialisation stuff tries to do anything with DNS first?

alexbishop1 (changes requested):
This looks like a positive.

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
oswaldquek: fixed

adityapahuja (approved):

alexbishop1 (approved):

PR #2623, opened 2020-10-20
PP-6991: Log IP address from worldpay

oswaldquek:
We are not sure if the load balancers/Nginxs are properly propagating the IP
address from the external systems through to connector, so logging the IP
addresses to be sure.

The logging of the worldpay IP only meant to be temporary.


rorymalcolm (commented):
Could we add a "//TODO: revoke this" ?

rorymalcolm (approved):
LGTM

PR #2628, opened 2020-10-21
Revert "PP-6991: Manually set the JVM TTL for DNS Name Lookups"

oswaldquek:
I believe this commit was causing a smoke test failure.

Reverts alphagov/pay-connector#2622


alexbishop1 (approved):
I would say I’m surprised this breaks things but nothing surprises me any more.

PR #2629, opened 2020-10-21
PP-6991: Make reverse dns lookup more stringent

oswaldquek:
Incoming traffic from worldpay will always be from the domain `*.worldpay.com`.
It was possible for us to receive traffic from `somebadattackerurlworldpay.com`
before.

There will be another pay-infra PR.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayNotificationResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayNotificationResourceIT.java
@@ -1,22 +1,29 @@
 package uk.gov.pay.connector.it.resources.worldpay;
 
+import io.dropwizard.setup.Environment;
 import io.restassured.response.ValidatableResponse;
+import org.junit.BeforeClass;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import uk.gov.pay.connector.app.ConnectorApp;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.app.ConnectorModule;
 import uk.gov.pay.connector.it.base.ChargingITestBase;
 import uk.gov.pay.connector.it.dao.DatabaseFixtures;
+import uk.gov.pay.connector.junit.ConfigOverride;
 import uk.gov.pay.connector.junit.DropwizardConfig;
 import uk.gov.pay.connector.junit.DropwizardJUnitRunner;
+import uk.gov.pay.connector.util.DnsPointerResourceRecord;
 import uk.gov.pay.connector.util.RandomIdGenerator;
+import uk.gov.pay.connector.util.ReverseDnsLookup;
 import uk.gov.pay.connector.util.TestTemplateResourceLoader;
 
 import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayNotificationResourceIT.java
```diff
@@ -201,7 +203,7 @@ private ValidatableResponse notifyConnector(String transactionId, String status,
     }
 
     private ValidatableResponse notifyConnector(String payload) {
-        String validIp = new DnsUtils().dnsLookup("build.ci.pymnt.uk").get();
```
heathd: ugh, I really dislike that this is looking up a real IP address and so the network is required to run this test.

Why are we not injecting an DNS lookup service and stubbing out the lookup behaviour?

oswaldquek: We could just set `validIp` to an existing worldpay one, e.g. 195.35.90.1?

alexbishop1: I think ideally (and this might be a bit of a pain to set up), we’d make it so that when running integration tests, `DnsUtils` is replaced by a stub that doesn’t actually hit the network and just has hard-coded hostname-to-IP-address mappings in it.

`DnsUtils` is already injected so it might be possible to have it provided by `ConnectorModule`, which could read a configuration value to determine whether to provide the real one or the stub one (with `test-it-config.yaml` configured to make it use the stub).

oswaldquek: Sure, but even if we replace `DnsUtils` with a stub we still require the network due to the tests in https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/util/DnsUtilsITest.java, unless we want to get rid of `DnsUtilsITest` so we don't rely on the network at all for the build?

oswaldquek: I've updated the test to stub out the reverse lookup behavior

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayNotificationResourceIT.java
```diff
@@ -153,24 +175,13 @@ public void shouldReturnNon2xxStatusIfChargeIsNotFoundForTransaction() throws Ex
...
-                .header("X-Forwarded-For", "8.8.8.8, 123.1.23.32")
+                .header("X-Forwarded-For", UNEXPECTED_IP_ADDRESS)
                 .contentType(TEXT_XML)
                 .post(NOTIFICATION_PATH)
                 .then()
                 .statusCode(403);
     }
-
-    @Test
-    public void shouldReturnForbiddenIfXForwardedForHeaderIsMalformed() {
```
oswaldquek, line 164: This can be removed as its equivalently tested in `IpDomainMatcherTest.reverseDnsShouldFailGracefullyIfForwardedHeaderIsNotValid`

alexbishop1 (approved):

PR #2630, opened 2020-10-22
PP-6991: Delete non-production code

oswaldquek:
`DnsUtils.dnsLookup` is not actually production code.

This is the first of a few PRs that aims to make the pay-connector build not depend on having a network available.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/util/DnsUtilsITest.java
+++ b/src/test/java/uk/gov/pay/connector/util/DnsUtilsITest.java
@@ -1,10 +1,7 @@
 package uk.gov.pay.connector.util;
 
-import org.apache.commons.validator.routines.InetAddressValidator;
 import org.junit.Before;
 import org.junit.Test;
-
-import java.util.Optional;
 
 import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/connector/util/DnsUtilsITest.java
```diff
@@ -21,9 +19,7 @@ public void setup() {
     @Test
     public void shouldReturnAnIpAddressForAValidHostname() {
         InetAddressValidator validator = InetAddressValidator.getInstance();
-        Optional<String> maybeIp = dnsUtils.dnsLookup("build.ci.pymnt.uk");
-        assertThat(maybeIp.isPresent(), is(true));
-        assertThat(validator.isValidInet4Address(maybeIp.get()), is (true));
+        assertThat(validator.isValidInet4Address("216.58.210.238"), is (true));
```
alexbishop1: This isn’t really testing anything other than `InetAddressValidator` now, which is third-party code, so this test can probably go.

alexbishop1 (changes requested):
I think we can delete even more!

alexbishop1 (approved):

PR #2631, opened 2020-10-22
PP-6991: Split out DnsUtils

oswaldquek:
Split out DnsUtils into its two functions: an IP to domain matcher and reverse
DNS lookup.

Doing this will allow the reverse DNS lookup to be stubbed out in tests, thus
removing the need for an actual network to be present. We don't need tests for
the reverse DNS lookup as code to do this is from a cookbook and not part of
our business logic.

This is the second of a series of PRs that aims to make the pay-connector build not depend on having a network available.


oswaldquek:
@alexbishop1 We could still keep the reverse dns lookup tests but put an `@Ignore` on it?

src/main/java/uk/gov/pay/connector/util/IpDomainMatcher.java
```diff
@@ -4,30 +4,32 @@
...
 import java.util.Optional;
 
 import static java.lang.String.format;
-import static java.lang.String.join;
 
-public class DnsUtils {
-    private static final Logger logger = LoggerFactory.getLogger(DnsUtils.class);
+public class IpDomainMatcher {
+    
+    private static final Logger logger = LoggerFactory.getLogger(IpDomainMatcher.class);
```
alexbishop1: This should really be `LOGGER` because it’s a constant.

src/main/java/uk/gov/pay/connector/util/IpDomainMatcher.java
```diff
@@ -4,30 +4,32 @@
...
+    
+    private ReverseDnsLookup reverseDnsLookup;
```
alexbishop1: This can be `final`.

src/main/java/uk/gov/pay/connector/util/IpDomainMatcher.java
```diff
@@ -4,30 +4,32 @@
...
         String extractedIp = forwardedAddress.split(",")[0];
         logger.debug("Extracted ip {} from X-Forwarded-For '{}'", extractedIp, forwardedAddress);
         return extractedIp;
     }
+    
     public boolean ipMatchesDomain(String forwardedAddress, String domain) {
         try {
             String ipAddress = extractForwardedIp(forwardedAddress);
-            Optional<String> host = reverseDnsLookup(ipAddress);
+            Optional<String> host = reverseDnsLookup.lookup(ipAddress);
```
alexbishop1: I know this was like this when you got here but we can make it a bit more functional. Something like this:

```java
return reverseDnsLookup(ipAddress).map(host -> {
    if (!host.endsWith(domain + ".")) {
        logger.error("Reverse DNS lookup on ip '{}' - resolved domain '{}' does not match '{}'", ipAddress, host, domain);
        return false;
    }
    return true;
}).orElseThrow(() -> new Exception(format("Host not found for ip address '%s'", ipAddress)));
```

src/main/java/uk/gov/pay/connector/util/ReverseDnsLookup.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Optional;
+
+import static java.lang.String.join;
+
+public class ReverseDnsLookup {
+
+    public Optional<String> lookup(String hostIp) {
+        List<String> components = Arrays.asList(hostIp.split("\\."));
+        Collections.reverse(components);
+        String reverseIp = join(".", components.toArray(new String[0])) + ".in-addr.arpa";
```
alexbishop1: I think these lines could go in `IpDomainMatcher` (changing what this method expects) so that this class literally just has the lookup and no other logic in it.

alexbishop1: The last line could also be written as `String reverseIp = components.stream().collect(Collectors.joining(".", "", ".in-addr.arpa"));` — not sure if that’s really any nicer though.

src/main/java/uk/gov/pay/connector/util/ReverseDnsLookup.java
```diff
@@ -0,0 +1,31 @@
...
+        try {
+            Hashtable env = new Hashtable();
+            env.put("java.naming.factory.initial", "com.sun.jndi.dns.DnsContextFactory");
```
alexbishop1: Please put generic type parameters on the `Hashtable`!

In fact, because the Javadoc for `InitialDirContext` says it won’t modify the `Hashtable`, we can probably make it a constant and just keep reusing the same one:

```java
private static final Hashtable<String, String> INITIAL_DIR_CONTEXT_ENV = new Hashtable<>(Map.of("java.naming.factory.initial", "com.sun.jndi.dns.DnsContextFactory"));
```

src/test/java/uk/gov/pay/connector/util/IpDomainMatcherTest.java
```diff
@@ -0,0 +1,42 @@
...
+        ipDomainMatcher = new IpDomainMatcher(reverseDnsLookup);
+    }
+
+    @Test
+    public void reverseDnsShouldCorrectlyMatchValidForwardedHeaderToDomain() {
+        when(reverseDnsLookup.lookup("195.35.90.1")).thenReturn(Optional.of("worldpay.com."));
+        when(reverseDnsLookup.lookup("8.8.8.8")).thenReturn(Optional.of("dns.google."));
+        assertThat(ipDomainMatcher.ipMatchesDomain("195.35.90.1, 8.8.8.8", "worldpay.com"), is(true));
+        assertThat(ipDomainMatcher.ipMatchesDomain("8.8.8.8, 8.8.8.8", "worldpay.com"), is(false));
+    }
```
alexbishop1, line 41: Maybe also a test with just a single IP address too (similar to `reverseDnsShouldReturnHostIfIpIsValid` in `DnsUtilsITest`)?

alexbishop1 (changes requested):
Looks good. Just a few small suggestions.

alexbishop1:
> @alexbishop1 We could still keep the reverse dns lookup tests but put an `@Ignore` on it?

We could keep them unignored. They’d hit the network still but at least it would only be in a set of well-contained tests that don’t test anything else.

src/main/java/uk/gov/pay/connector/util/ReverseDnsLookup.java
oswaldquek: I think these lines should remain in this class as extracting the "QUESTION SECTION" is part of doing a reverse dns lookup. It's nicer for the method signature to take the IP address. See this `dig` command to see what I mean about the "QUESTION SECTION":
```
>dig -x 195.35.90.1 

; <<>> DiG 9.10.6 <<>> -x 195.35.90.1
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 24835
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;1.90.35.195.in-addr.arpa.	IN	PTR

;; ANSWER SECTION:
1.90.35.195.in-addr.arpa. 500	IN	PTR	hello.worldpay.com.
```

alexbishop1: I see your point. It’s just that we end up with a few lines of code that would ordinarily be easy to test but which are hard to test because they’re mixed up in this class.

We could have a separate (injected) class that’s just responsible for going from an IP address to the `.in-addr.arpa` name. That would be then be easy to test but would still leave the method signature in `ReverseDnsLookup` as it is.

I’ll leave it up to your judgement.

oswaldquek: I created a `DnsPointerResourceRecord` class (and a corresponding test) for this.

src/main/java/uk/gov/pay/connector/util/DnsPointerResourceRecord.java
```diff
@@ -0,0 +1,48 @@
...
+    public boolean equals(Object o) {
+        if (this == o) return true;
+
+        if (o == null || getClass() != o.getClass()) return false;
+
+        DnsPointerResourceRecord that = (DnsPointerResourceRecord) o;
+
+        return new EqualsBuilder()
+                .append(reverseIp, that.reverseIp)
+                .isEquals();
```
alexbishop1, line 39: Just `return this.reverseIp.equals(that.reverseIp);` would do the trick.

alexbishop1 (approved):
Nice!

PR #2640, opened 2020-10-26
PP-7255 Add DAO method to find refunds by parity check status

kbottla:
## WHAT YOU DID
- Added DAO method to find refunds by parity check status. This is to be used to invoke parity check task on refunds for a given parity check status


SandorArpa (approved):
LOKTM

PR #2644, opened 2020-10-27
PP-7255 Parity check task for refunds

kbottla:
## WHAT YOU DID
- Currently we can parity check charges for a given ID range or parity check status. There is no way to parity check refund(s) using the task. For any fixes to refund parity checker we will have to wait until expunger picks the refund and parity checks the same.
Added functionality to parity check refunds by ID range and parity check status.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
+++ b/src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
@@ -13,13 +13,20 @@ import uk.gov.pay.connector.events.EventService;
 import uk.gov.pay.connector.events.dao.EmittedEventDao;
 import uk.gov.pay.connector.queue.statetransition.StateTransitionService;
 import uk.gov.pay.connector.refund.dao.RefundDao;
+import uk.gov.pay.connector.refund.model.domain.RefundEntity;
+import uk.gov.pay.connector.refund.service.RefundService;
 import uk.gov.pay.connector.tasks.service.ParityCheckService;
 
 import javax.inject.Inject;
 import java.util.List;
 import java.util.Optional;
 
+import static java.util.Optional.ofNullable;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
```diff
@@ -75,6 +84,27 @@ public void execute(Long startId, Optional<Long> maybeMaxId, boolean doNotReproc
...
+            } else {
+                maxId = ofNullable(maxId).orElseGet(refundDao::findMaxId);
+                processRefundsByIdRange(startId, maxId, doNotReprocessValidRecords);
+            }
+        } catch (Exception e) {
+            logger.error("Error parity checking refunds on job [start={}] [max={}] [error={}]",
+                    startId, maxId, e.getMessage(), e);
+        }
+        logger.info("Terminating");
+        MDC.remove(HEADER_REQUEST_ID);
```
stephencdaly, line 105: do we need to remove from the MDC for `execute` as well, if we're removing here?

src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
```diff
@@ -115,27 +145,81 @@ public void checkParityForIdRange(long startId, long maxId, boolean doNotReproce
     @Transactional
     public void checkParityFor(ChargeEntity charge, boolean doNotReprocessValidRecords) {
         try {
-            MDC.put("chargeId", charge.getExternalId());
+            MDC.put(PAYMENT_EXTERNAL_ID, charge.getExternalId());
```
stephencdaly, line 148: thanks for fixing that!

stephencdaly (approved):
Looks good. Question about MDC doesn't stop this PR being merged.

src/main/java/uk/gov/pay/connector/tasks/ParityCheckWorker.java
kbottla: yeah missed that. Will remove in the next PR

PR #2651, opened 2020-10-29
PP-6990 Reduce time before charge is expunged to 2 days

kbottla:
## WHAT YOU DID
- Reduce the number of days connector retains charges to 2 days. This is the minimum we need so daily stats can be reported to performance platform. We can reduce to inflight payments/refunds only when all the reporting is moved to ledger.


SandorArpa (approved):
LOKTM

PR #2654, opened 2020-11-02
PP-6948 - Add structured logging for 3DS

nimalank7:
Description:
- This adds structured logging for the 3DS response we receive from the payment gateway
- This is necessary so we can log out whether the 3DS request is 3DS1 or 3DS Flex for Worldpay


alexbishop1 (dismissed):
Good improvement

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #2670, opened 2020-11-09
PP-7382: Log paRequest

oswaldquek:
Include the paRequest in the toString so it will be logged in
https://github.com/alphagov/pay-connector/blob/eacd57a5a0c99c31f67c507fb5bb4c2d9c26dd66/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java#L98.
We want to verify the paRequest is not empty as that would cause an error when
trying to do a 3ds verification at the frontend.


alexbishop1:
PA requests tend to be really, really long. We do store them in the database though so can we check there instead?

PR #2673, opened 2020-11-10
PP-7382: Store 3ds details for wallet payments

oswaldquek:
Until Worldpay returned a 3ds required response for a Google Payment, we
weren't updating the charge with the auth3ds details, which meant the issuerURL
and paResponse from Worldpay weren't being stored in the database for the
charge.

Against best practices, I have chosen to include an `Optional<Auth3dsRequiredEntity>` parameter to the `ChargeService.updateChargePostWalletAuthorisation` method. This is because having an overloaded method (i.e. a `updateChargePostWalletAuthorisation` with no `Auth3dsRequiredEntity` parameter would cause the code in `WalletAuthoriseService.processGatewayAuthorisationResponse` to be unnecessarily lengthy to read. So from this
```
    private void processGatewayAuthorisationResponse(
            String chargeExternalId,
            ChargeStatus oldChargeStatus,
            WalletAuthorisationData walletAuthorisationData,
            String responseFromGateway,
            String transactionId,
            ProviderSessionIdentifier sessionIdentifier,
            ChargeStatus status,
            Optional<Auth3dsRequiredEntity> auth3dsRequiredDetails) {

        logger.info("Processing gateway auth response for {}", walletAuthorisationData.getWalletType().toString());
        AuthCardDetails authCardDetailsToBePersisted = authCardDetailsFor(walletAuthorisationData);
        ChargeEntity updatedCharge = chargeService.updateChargePostWalletAuthorisation(
                chargeExternalId,
                status,
                transactionId,
                sessionIdentifier,
                authCardDetailsToBePersisted,
                walletAuthorisationData.getWalletType(),
                walletAuthorisationData.getPaymentInfo().getEmail(),
                auth3dsRequiredDetails);
```
to this
```
    private void processGatewayAuthorisationResponse(
            String chargeExternalId,
            ChargeStatus oldChargeStatus,
            WalletAuthorisationData walletAuthorisationData,
            String responseFromGateway,
            String transactionId,
            ProviderSessionIdentifier sessionIdentifier,
            ChargeStatus status,
            Optional<Auth3dsRequiredEntity> auth3dsRequiredDetails) {

        logger.info("Processing gateway auth response for {}", walletAuthorisationData.getWalletType().toString());
        AuthCardDetails authCardDetailsToBePersisted = authCardDetailsFor(walletAuthorisationData);

        if (auth3dsRequiredDetails.isPresent()) {
            ChargeEntity updatedCharge = chargeService.updateChargePostWalletAuthorisation(
[...]

PR #2676, opened 2020-11-12
bau: Simplify and rename CardAuthoriseBaseService

oswaldquek:
It was clear from https://github.com/alphagov/pay-connector/pull/2673 that
there is some refactoring that could be done to remove code duplication between
WalletAuthoriseService and CardAuthoriseService. As there are still slight
authorisation implementation differences between the two it is envisaged there
will be a common AuthoriseService. However looking at the current code there is
a plethora of authorise services:

CardAuthoriseService
CardAuthoriseBaseService
Card3dsResponseAuthService
WalletAuthoriseService

The term "BaseService" is quite annoying in the sense that it doesn't really mean anything,
so is a candidate for simplification and renaming. This would be a good first
step before unifying the WalletAuthoriseService and CardAuthoriseService.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationExecutor.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationExecutor.java
@@ -1,13 +1,9 @@
 package uk.gov.pay.connector.paymentprocessor.service;
 
-
-import com.codahale.metrics.MetricRegistry;
-import io.dropwizard.setup.Environment;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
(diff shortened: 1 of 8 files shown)


alexbishop1 (changes requested):
I like the principle of renaming this but I’m not sure about `AuthorisationExecutor` because generally we only use `Executor` for things that run on a separate thread. Maybe just `AuthorisationService`?

oswaldquek:
@alexbishop1 I was trying to avoid `AuthorisationService` as that's what I planned to name the up-and-coming class that contains common methods between WalletAuthoriseService and CardAuthoriseService. How about `ExecuteAuthorise`?

alexbishop1:
Bah! Might it be the case that `AuthorisationService` includes the functionality of `AuthorisationExecutor` in the end? The division of responsibilities seems a bit arbitrary at the moment. That said, I’d prefer us to have more small classes than fewer large ones. Maybe just `ChargeAuthoriser` or something?

oswaldquek:
You know, let me try and put everything in an AuthorisationService (in another PR). I think it can be done.

PR #2677, opened 2020-11-12
bau: Rename CardAuthoriseBaseService

oswaldquek:
The term "BaseService" is quite annoying in the sense that it doesn't really
mean anything. CardAuthoriseBaseService has been renamed to 
"AuthorisationService" and the intention is it will handle both Card and 
Wallet authorisations in up-and-coming PRs.


alexbishop1 (approved):
All your base are belong to us!

PR #2683, opened 2020-11-16
PP-7336: Update Stripe notification service to perform IP firewalling

adityapahuja:
## WHAT YOU DID
This pull requests contains 3 commits.

1. The first commit adds IP Address Matcher. It is responsible for matching IP address with one of allowed IP addresses.
2. The second commit updates Stripe Notification Service to perform IP firewalling so that only requests coming from Stripe can be processed. It adds unit tests and an integration test to ensure that Stripe Notification Service can process requests coming from Stripe and reject other requests which are not from Stripe.
3. The third commit adds "STRIPE_ALLOWED_IP_ADDRESSES" environment variable to README for Connector application. Connector application uses this environment variable containing a list of allowed Stripe IP addresses so that it can perform IP firewalling on incoming notifications from Stripe. 

I have tested the changes locally and found them working fine. Stripe Notification Service can accept notifications coming from IP address present in the allowed list and reject notifications coming from IP address not present in the allowed list.

### Logging

### Documentation

Yes

No

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
@@ -25,6 +25,7 @@ import uk.gov.pay.connector.paymentprocessor.service.CardExecutorService;
 import uk.gov.pay.connector.queue.statetransition.StateTransitionQueue;
 import uk.gov.pay.connector.usernotification.govuknotify.NotifyClientFactory;
 import uk.gov.pay.connector.util.HashUtil;
+import uk.gov.pay.connector.util.IpAddressMatcher;
 import uk.gov.pay.connector.util.JsonObjectMapper;
 import uk.gov.pay.connector.util.ReverseDnsLookup;
 import uk.gov.pay.connector.util.XrayUtils;
@@ -53,6 +54,7 @@ public class ConnectorModule extends AbstractModule {
         bind(HashUtil.class);
         bind(RequestValidator.class);
         bind(GatewayAccountRequestValidator.class).in(Singleton.class);
+        bind(IpAddressMatcher.class).in(Singleton.class);
 
         install(jpaModule(configuration));
         install(new FactoryModuleBuilder().build(GatewayAccountServicesFactory.class));
--- a/src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
@@ -38,6 +38,10 @@ public class StripeGatewayConfig extends Configuration {
     @Min(1)
     private int notification3dsWaitDelay;
 
+    @Valid
+    @NotNull
+    private List<String> allowedIpAddresses;
+
     public String getUrl() {
         return url;
     }
@@ -64,5 +68,9 @@ public class StripeGatewayConfig extends Configuration {
 
     public int getNotification3dsWaitDelay() {
         return notification3dsWaitDelay;
+    }
+
+    public List<String> getAllowedIpAddresses() {
+        return allowedIpAddresses;
     }
 }
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -88,6 +88,7 @@ epdq:
       readTimeout: 20000ms
 
 stripe:
+  allowedIpAddresses: ${STRIPE_ALLOWED_IP_ADDRESSES}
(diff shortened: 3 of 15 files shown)


src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -53,6 +54,7 @@ protected void configure() {
         bind(HashUtil.class);
         bind(RequestValidator.class);
         bind(GatewayAccountRequestValidator.class).in(Singleton.class);
+        bind(IpAddressMatcher.class).in(Singleton.class);
```
oswaldquek, line 57: I'd preferred it if `@Singleton` was placed in the `IpAddressMatcher` class as it makes it clearer.

src/main/resources/config/config.yaml
```diff
@@ -88,6 +88,7 @@ epdq:
       readTimeout: 20000ms
 
 stripe:
+  allowedIpAddresses: $STRIPE_ALLOWED_IP_ADDRESSES
```
oswaldquek: missing curly braces, should be `${STRIPE_ALLOWED_IP_ADDRESSES}`

src/test/resources/config/test-it-config.yaml
```diff
@@ -53,6 +53,7 @@ epdq:
       readTimeout: 1000ms
 
 stripe:
+  allowedIpAddresses: ${STRIPE_ALLOWED_IP_ADDRESSES:-[]}
```
oswaldquek: If the default list if empty as it here, how do we make integration tests in a build pass? Just curious.

oswaldquek (changes requested):
Looks good! Some minor changes requested.

src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
```diff
@@ -38,6 +38,10 @@
     @Min(1)
     private int notification3dsWaitDelay;
 
+    @Valid
+    @NotNull
+    private List<String> allowedIpAddresses;
```
oswaldquek, line 43: I think we also need more confidence that way the IP addresses are configured in terraform (i.e. https://github.com/alphagov/pay-infra/pull/2712) actually resolve to a java `List<String>` here. I'm not sure whether the following config works
```
stripe:
  allowedIpAddresses: ["1.2.3.4", "3.5.7.8"]
```

adityapahuja: This is the right format. I already tested this change locally before raising this pull request. It is working fine. 

Would you like me to drop configOverrides which sets stripe.allowedIpAddresses to the list of allowed IP addresses in StripeNotificationResourceIT class and set this list explicitly in the config file under test resources?

src/test/resources/config/test-it-config.yaml
adityapahuja: I used configOverrides which sets stripe.allowedIpAddresses to the list of allowed IP addresses in StripeNotificationResourceIT class.

adityapahuja: Without the default value containing an empty list, the integration test will throw non-null error. It will need the variable to be defined before running integration tests.

src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
adityapahuja: I will move `stripe.allowedIpAddresses` from configOverrides to the config file to increase our confidence.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
adityapahuja: Done.

src/main/resources/config/config.yaml
adityapahuja: Done.

src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
adityapahuja: Done.

oswaldquek (approved):

src/test/resources/config/test-it-config.yaml
oswaldquek: Ah, gotcha.

oswaldquek:
https://github.com/alphagov/pay-infra/pull/2712 should be applied to the environments before deploying this change.

PR #2684, opened 2020-11-17
PP-6948 - Log Worldpay 3DS Flex for 2nd authorisation request

nimalank7:
Description:
- This PR fills a gap in our logging for Worldpay 3DS/Flex payments by logging out whether the payload for the 2nd authorisation request is 3DS or Flex

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.paymentprocessor.service;
 
+import net.logstash.logback.argument.StructuredArgument;
+import org.apache.commons.lang3.ArrayUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -8,6 +10,7 @@ import uk.gov.pay.connector.app.config.Authorisation3dsConfig;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.model.Auth3dsResult;
 import uk.gov.pay.connector.gateway.model.Gateway3dsRequiredParams;
@@ -18,13 +21,28 @@ import javax.inject.Inject;
 import java.util.Locale;
 import java.util.Optional;
 
+import static net.logstash.logback.argument.StructuredArguments.kv;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
 import static uk.gov.pay.connector.paymentprocessor.model.OperationType.AUTHORISATION_3DS;
+import static uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService.TypeOf3dsRequest.NONE_OF_THE_ABOVE;
+import static uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService.TypeOf3dsRequest.WORLDPAY_3DS_CLASSIC;
+import static uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService.TypeOf3dsRequest.WORLDPAY_3DS_FLEX;
 
 public class Card3dsResponseAuthService {
 
+    enum TypeOf3dsRequest {
+        WORLDPAY_3DS_CLASSIC(" using 3DS"), WORLDPAY_3DS_FLEX(" using 3DS Flex"), NONE_OF_THE_ABOVE("");
+
+        public final String logMessage;
+
+        TypeOf3dsRequest(String logMessage) {
+            this.logMessage = logMessage;
+        }
+    }
+
     private static final Logger LOGGER = LoggerFactory.getLogger(Card3dsResponseAuthService.class);
+    public static final String INTEGRATION_3DS_VERSION = "integration_3ds_version";
 
     private final ChargeService chargeService;
     private final AuthorisationService authorisationService;
@@ -72,11 +90,11 @@ public class Card3dsResponseAuthService {
             }
         }
 
-        processGateway3DSecureResponse(charge.getExternalId(), ChargeStatus.fromString(charge.getStatus()), gateway3DSAuthorisationResponse);
+        processGateway3DSecureResponse(charge.getExternalId(), ChargeStatus.fromString(charge.getStatus()), gateway3DSAuthorisationResponse, auth3dsResult);
         return gateway3DSAuthorisationResponse;
     }
 
-    private void processGateway3DSecureResponse(String chargeExternalId, ChargeStatus oldChargeStatus, Gateway3DSAuthorisationResponse operationResponse) {
+    private void processGateway3DSecureResponse(String chargeExternalId, ChargeStatus oldChargeStatus, Gateway3DSAuthorisationResponse operationResponse, Auth3dsResult auth3dsResult) {
         Optional<String> transactionId = operationResponse.getTransactionId();
 
         ChargeStatus newStatus = operationResponse.getMappedChargeStatus();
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
```diff
@@ -123,4 +125,12 @@ private void processGateway3DSecureResponse(String chargeExternalId, ChargeStatu
 
         authorisationService.emitAuthorisationMetric(updatedCharge, "authorise-3ds");
     }
+
+    private String logWorldpay3dsOrFlex(ChargeEntity chargeEntity) {
+        if (PaymentGatewayName.WORLDPAY.equals(chargeEntity.getPaymentGatewayName())) {
+            return chargeEntity.getGatewayAccount().getWorldpay3dsFlexCredentials() != null ? " using Flex" : " using 3DS1";
```
alexbishop1: I’m not quite sure about doing it this way. We want to make it possible to disable 3DS Flex without removing the credentials.

Ideally, we’d inspect the payload that we’re about to send to Worldpay to see if it has a device data collection result.

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
nimalank7: Done used the paResponse instead

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
```diff
@@ -123,4 +125,12 @@ private void processGateway3DSecureResponse(String chargeExternalId, ChargeStatu
 
         authorisationService.emitAuthorisationMetric(updatedCharge, "authorise-3ds");
     }
+
+    private String logWorldpay3dsOrFlex(ChargeEntity chargeEntity, Auth3dsResult auth3dsResult) {
+        if (PaymentGatewayName.WORLDPAY.equals(chargeEntity.getPaymentGatewayName())) {
+            return auth3dsResult.getPaResponse() != null ? " using 3DS1" : " using Flex";
```
alexbishop1: Slightly picky but let’s call it “3DS Flex” to match the name of the Worldpay product.

alexbishop1 (changes requested):
Looks good. Can we also add structured logging? We could call it `3ds_integration` or something with a value of `3DS1` or `3DS Flex`. That way we could reuse the same key with different values for other gateways if necessary (but we should only add it for Worldpay requests for now, since they’re the only ones that have different variants).

src/main/java/uk/gov/pay/connector/gateway/util/Authorisation3dsResponseSummaryStructuredLogging.java
```diff
@@ -0,0 +1,28 @@
...
+import net.logstash.logback.argument.StructuredArgument;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.gateway.PaymentGatewayName;
+import uk.gov.pay.connector.gateway.model.Auth3dsResult;
+
+import java.util.ArrayList;
+
+import static net.logstash.logback.argument.StructuredArguments.kv;
+
+public class Authorisation3dsResponseSummaryStructuredLogging {
```
alexbishop1: I can see you’re following the pattern we use for the first authorisation request but I think this might be overkill to just add one parameter that we can easily calculate at the point we do the logging.

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/gateway/util/Authorisation3dsResponseSummaryStructuredLogging.java
nimalank7: Modified it to remove the classes and place it into the method.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
```diff
@@ -117,10 +124,29 @@ private void processGateway3DSecureResponse(String chargeExternalId, ChargeStatu
...
                 updatedCharge.getStatus());
 
-        var structuredLoggingArguments = updatedCharge.getStructuredLoggingArgs();
-
         LOGGER.info(logMessage, structuredLoggingArguments);
 
         authorisationService.emitAuthorisationMetric(updatedCharge, "authorise-3ds");
     }
+
+    private String logWorldpay3dsOrFlex(Auth3dsResult auth3dsResult, ChargeEntity charge, Object[] structuredLoggingArguments) {
```
alexbishop1: I think this method is trying to do too much. How about you instead you define an private enum in this class called `TypeOf3dsRequest` or something with values like `WORLDPAY_3DS_CLASSIC`, `WORLDPAY_3DS_FLEX` and `NONE_OF_THE_ABOVE`. Then have this method return that. Then you can call this method once near the start and store its result in a local variable. Then use if statements or switch statements to add the appropriate structured logging arguments or bits to the log line.

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
nimalank7: Yep changes made.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
```diff
@@ -18,13 +21,28 @@
...
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
 import static uk.gov.pay.connector.paymentprocessor.model.OperationType.AUTHORISATION_3DS;
+import static uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService.TypeOf3dsRequest.NONE_OF_THE_ABOVE;
+import static uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService.TypeOf3dsRequest.WORLDPAY_3DS_CLASSIC;
+import static uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService.TypeOf3dsRequest.WORLDPAY_3DS_FLEX;
 
 public class Card3dsResponseAuthService {
 
+    enum TypeOf3dsRequest {
```
alexbishop1, line 34: I think we can make this private.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
```diff
@@ -117,10 +139,38 @@ private void processGateway3DSecureResponse(String chargeExternalId, ChargeStatu
...
                 updatedCharge.getStatus());
 
-        var structuredLoggingArguments = updatedCharge.getStructuredLoggingArgs();
-
         LOGGER.info(logMessage, structuredLoggingArguments);
 
         authorisationService.emitAuthorisationMetric(updatedCharge, "authorise-3ds");
     }
+
+    private TypeOf3dsRequest logWorldpay3dsOrFlex(Auth3dsResult auth3dsResult, ChargeEntity charge) {
```
alexbishop1: The name of this method doesn’t really fit any more.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
```diff
@@ -117,10 +139,38 @@ private void processGateway3DSecureResponse(String chargeExternalId, ChargeStatu
...
+                structuredLoggingArguments = ArrayUtils.addAll(
+                        updatedCharge.getStructuredLoggingArgs(),
+                        new StructuredArgument[] {kv(INTEGRATION_3DS_VERSION, "3DS Flex")}
+                );
+                break;
+            default:
+                structuredLoggingArguments = updatedCharge.getStructuredLoggingArgs();
+        }
+        return structuredLoggingArguments;
+    }
```
alexbishop1: I still think this is a bit overwrought.

How about immediately after the `var worldPay3dsOrFlexLogMessage = logWorldpay3dsOrFlex(auth3dsResult, updatedCharge);` line we just do this:

```java
var structuredLoggingArguments = updatedCharge.getStructuredLoggingArgs();
if (worldPay3dsOrFlexLogMessage == WORLDPAY_3DS_CLASSIC) {
    structuredLoggingArguments = ArrayUtils.addAll(structuredLoggingArguments, new StructuredArgument[] {kv(INTEGRATION_3DS_VERSION, "3DS")});
} else if (worldPay3dsOrFlexLogMessage == WORLDPAY_3DS_FLEX) {
    structuredLoggingArguments = ArrayUtils.addAll(structuredLoggingArguments, new StructuredArgument[] {kv(INTEGRATION_3DS_VERSION, "3DS Flex")});
}
```

alexbishop1 (changes requested):
Getting there…

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
nimalank7: Changed.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/Card3dsResponseAuthService.java
nimalank7: We won't be able to do the static import if it is made private.

alexbishop1 (approved):

PR #2690, opened 2020-11-20
PP-7363 Add logging for refund state changes

kbottla:
## WHAT YOU DID
- Added logging for refund state changes so we can use it to understand refund states. Refund status can be changed to refunded immediately or when we receive notification. Currently we only have detailed (with gateway account info) logs for notifications but not when refund is succeeded immediately.


SandorArpa (approved):
LGTM

PR #2691, opened 2020-11-20
PP-7336: Update Stripe to use a list of allowed CIDRs for firewall

adityapahuja:
## WHAT YOU DID

This change updates Stripe Gateway Config to pick up a list of allowed CIDRs instead of IP addresses because using CIDR helps to reduce a large number of IP addresses needed to be specified in the config file. It also updates IP Address Matcher to validate a forwarded IP address before processing it.

### Logging

### Documentation

I renamed an environment variable from STRIPE_ALLOWED_IP_ADDRESSES to STRIPE_ALLOWED_CIDRS. There is a pull request for this and it needs to be merged before this pull request is merged.

No.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
@@ -10,9 +10,11 @@ import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
 import com.google.inject.Singleton;
 import com.google.inject.assistedinject.FactoryModuleBuilder;
+import com.google.inject.name.Named;
 import com.google.inject.persist.jpa.JpaPersistModule;
 import io.dropwizard.db.DataSourceFactory;
 import io.dropwizard.setup.Environment;
+import org.apache.commons.validator.routines.InetAddressValidator;
 import uk.gov.pay.connector.charge.service.Worldpay3dsFlexJwtService;
 import uk.gov.pay.connector.charge.util.JwtGenerator;
 import uk.gov.pay.connector.common.validator.RequestValidator;
@@ -24,6 +26,7 @@ import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountServicesFactory
 import uk.gov.pay.connector.paymentprocessor.service.CardExecutorService;
 import uk.gov.pay.connector.queue.statetransition.StateTransitionQueue;
 import uk.gov.pay.connector.usernotification.govuknotify.NotifyClientFactory;
+import uk.gov.pay.connector.util.CidrUtils;
 import uk.gov.pay.connector.util.HashUtil;
 import uk.gov.pay.connector.util.IpAddressMatcher;
 import uk.gov.pay.connector.util.JsonObjectMapper;
@@ -34,6 +37,7 @@ import uk.gov.pay.connector.wallets.applepay.ApplePayDecrypter;
 import javax.ws.rs.client.Client;
 import java.time.Clock;
 import java.util.Properties;
+import java.util.Set;
 
 public class ConnectorModule extends AbstractModule {
     final ConnectorConfiguration configuration;
@@ -54,7 +58,7 @@ public class ConnectorModule extends AbstractModule {
         bind(HashUtil.class);
         bind(RequestValidator.class);
         bind(GatewayAccountRequestValidator.class).in(Singleton.class);
-        bind(IpAddressMatcher.class).in(Singleton.class);
+        bind(InetAddressValidator.class).in(Singleton.class);
 
         install(jpaModule(configuration));
         install(new FactoryModuleBuilder().build(GatewayAccountServicesFactory.class));
@@ -117,6 +121,19 @@ public class ConnectorModule extends AbstractModule {
     @Provides
     public StripeGatewayConfig stripeGatewayConfig(ConnectorConfiguration connectorConfiguration) {
         return connectorConfiguration.getStripeConfig();
+    }
+
+    @Provides
+    @Singleton
+    @Named("AllowedStripeIpAddresses")
+    public Set<String> allowedStripeIpAddresses(StripeGatewayConfig config) {
+        return CidrUtils.getIpAddresses(config.getAllowedCidrs());
+    }
+
+    @Provides
+    @Singleton
+    public IpAddressMatcher ipAddressMatcher(InetAddressValidator inetAddressValidator) {
+        return new IpAddressMatcher(inetAddressValidator);
     }
 
     @Provides
--- a/src/main/java/uk/gov/pay/connector/util/IpAddressMatcher.java
+++ b/src/main/java/uk/gov/pay/connector/util/IpAddressMatcher.java
@@ -1,14 +1,25 @@
 package uk.gov.pay.connector.util;
 
+import org.apache.commons.validator.routines.InetAddressValidator;
+
 import javax.inject.Singleton;
-import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 @Singleton
 public class IpAddressMatcher {
-    public boolean isMatch(String forwardedIpAddresses, List<String> allowedIpAddresses) {
+    private final InetAddressValidator ipAddressValidator;
+
+    public IpAddressMatcher(InetAddressValidator ipAddressValidator) {
+        this.ipAddressValidator = ipAddressValidator;
+    }
+
+    public boolean isMatch(String forwardedIpAddresses, Set<String> allowedIpAddresses) {
         if (Objects.nonNull(forwardedIpAddresses) && Objects.nonNull(allowedIpAddresses)) {
-            return allowedIpAddresses.contains(getFirstIpAddress(forwardedIpAddresses));
+            String ipAddress = getFirstIpAddress(forwardedIpAddresses);
+            if (ipAddressValidator.isValid(ipAddress)) {
+                return allowedIpAddresses.contains(ipAddress);
+            }
         }
         return false;
     }
--- a/src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
@@ -10,6 +10,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
 import net.logstash.logback.marker.ObjectAppendingMarker;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.validator.routines.InetAddressValidator;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
@@ -28,6 +29,7 @@ import uk.gov.pay.connector.payout.PayoutEmitterService;
 import uk.gov.pay.connector.queue.QueueException;
 import uk.gov.pay.connector.queue.payout.Payout;
 import uk.gov.pay.connector.queue.payout.PayoutReconcileQueue;
+import uk.gov.pay.connector.util.CidrUtils;
 import uk.gov.pay.connector.util.IpAddressMatcher;
 import uk.gov.pay.connector.util.TestTemplateResourceLoader;
 
@@ -36,6 +38,7 @@ import java.time.Instant;
 import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.is;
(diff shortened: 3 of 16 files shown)


src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -119,6 +125,25 @@ public StripeGatewayConfig stripeGatewayConfig(ConnectorConfiguration connectorC
...
+    @Singleton
+    @Named("AllowedStripeIpAddresses")
+    public Set<String> allowedStripeIpAddresses(StripeGatewayConfig config) {
+        ImmutableSet.Builder<String> allowedStripeIpAddresses = ImmutableSet.builder();
+        for (String cidr : config.getAllowedCidrs()) {
+            SubnetUtils subnet = new SubnetUtils(cidr);
+            subnet.setInclusiveHostCount(true);
+            allowedStripeIpAddresses.addAll(List.of(subnet.getInfo().getAllAddresses()));
+        }
+        return allowedStripeIpAddresses.build();
```
alexbishop1: Firstly, do we actually need to include the network and broadcast addresses, as `setInclusiveHostCount(true)` does? I’m no expert on these on things but I would have thought we won’t receive any requests with these as the source IP because they don’t belong to actual hosts.

Secondly, we can do this without using Guava:

```java
var allowedStripeIpAddresses = new HashSet<String>();
// … for loop
return Set.copyOf(allowedStripeIpAddresses);
```

Alternatively, we could do it like this:

```java
Set<SubnetUtils> subnets = config.getAllowedCidrs().stream().map(SubnetUtils::new).collect(toUnmodifiableSet());
subnets.forEach(subnet -> subnet.setInclusiveHostCount(true));
return subnets.stream().map(SubnetUtils::getInfo).map(SubnetUtils.SubnetInfo::getAllAddresses).flatMap(Arrays::stream).collect(toUnmodifiableSet());
```

But I’m not sure if that’s any clearer. I’ll leave that decision up to you.

Of course, if it turns out we don’t need to do `setInclusiveHostCount(true)`, then it becomes much simpler:

```java
return config.getAllowedCidrs()
        .stream()
        .map(SubnetUtils::new)
        .map(SubnetUtils::getInfo)
        .map(SubnetUtils.SubnetInfo::getAllAddresses)
        .flatMap(Arrays::stream)
        .collect(toUnmodifiableSet());
```

src/main/java/uk/gov/pay/connector/util/IpAddressMatcher.java
```diff
@@ -1,14 +1,25 @@
...
+    public IpAddressMatcher(InetAddressValidator ipAddressValidator) {
+        this.ipAddressValidator = ipAddressValidator;
+    }
+
+    public boolean isMatch(String forwardedIpAddresses, Set<String> allowedIpAddresses) {
         if (Objects.nonNull(forwardedIpAddresses) && Objects.nonNull(allowedIpAddresses)) {
-            return allowedIpAddresses.contains(getFirstIpAddress(forwardedIpAddresses));
+            String ipAddress = getFirstIpAddress(forwardedIpAddresses);
+            if (ipAddressValidator.isValid(ipAddress)) {
+                return allowedIpAddresses.contains(getFirstIpAddress(forwardedIpAddresses));
```
alexbishop1: We already have `ipAddress` so we can just do this:

```java
return allowedIpAddresses.contains(ipAddress);
```

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -73,7 +77,14 @@
...
     private static final String FORWARDED_IP_ADDRESSES = "1.2.3.4, 102.108.0.6";
-    private static final List<String> ALLOWED_IP_ADDRESSES = List.of("1.2.3.4", "9.9.9.9");
+    private static final Set<String> ALLOWED_IP_ADDRESSES;
+
+    static {
+        ImmutableSet.Builder<String> allowedIpAddresses = ImmutableSet.builder();
+        addIpAddresses(allowedIpAddresses, "1.2.3.0/24");
+        addIpAddresses(allowedIpAddresses, "9.9.9.9/32");
+        ALLOWED_IP_ADDRESSES = allowedIpAddresses.build();
+    }
```
alexbishop1: We can avoid Guava using the suggestion I made in an earlier comment.

Having a `static` block is kind of _eww_ so maybe `ALLOWED_IP_ADDRESSES` could just be a regular field that’s initialised in the `@Before` method?

src/test/java/uk/gov/pay/connector/util/IpAddressMatcherTest.java
```diff
@@ -1,15 +1,29 @@
...
-    private IpAddressMatcher ipAddressMatcher = new IpAddressMatcher();
+    private static final Set<String> ALLOWED_IP_ADDRESSES;
+
+    static {
+        ImmutableSet.Builder<String> allowedIpAddresses = ImmutableSet.builder();
+        addIpAddresses(allowedIpAddresses, "9.9.9.9/32");
+        addIpAddresses(allowedIpAddresses, "1.2.3.0/24");
+        addIpAddresses(allowedIpAddresses, "3.6.9.12/32");
+        ALLOWED_IP_ADDRESSES = allowedIpAddresses.build();
+    }
```
alexbishop1: Same comment as the last file applies here.

alexbishop1 (changes requested):
Looks straightforward enough. I’m not sure about if we need to include the network and broadcast addresses though. Other than that, just a few comments on avoiding Guava (it’s lovely but the standard Java SE stuff is enough for us), the possibility for a few more tests and some stylistic things.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
adityapahuja: Unfortunately, yes. We need to include `setInclusiveHostCount(true)`. For example, if the CIDR is `1.1.1.1/32` and `setInclusiveHostCount(false)` (default) is used, the SubnetUtils.getAllAddresses() will return an empty list of IP addresses.

I will replace Guava with Java.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
adityapahuja: If we make `ALLOWED_IP_ADDRESSES` as a regular field then the field has to be initialised (e.g. converting from CIDRs into a set of IP addresses) before each test, it will take longer to execute all tests cases. That's a reason why I moved it to static initialisation block.

adityapahuja: I will replace Gauva with Java.

src/main/java/uk/gov/pay/connector/util/IpAddressMatcher.java
adityapahuja: Thanks for spotting it. I have fixed it.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
adityapahuja: Done.

src/test/java/uk/gov/pay/connector/util/IpAddressMatcherTest.java
adityapahuja: Done.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
adityapahuja: Done.

alexbishop1: I doubt reinitialising the set of allowed IP addresses before every test will make a huge difference to the speed. You could always keep it as a static field but initialise it in a `@BeforeClass` method.

adityapahuja: I have moved it to `@BeforeAll` because it is for JUnit5.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
```diff
@@ -119,6 +125,24 @@ public StripeGatewayConfig stripeGatewayConfig(ConnectorConfiguration connectorC
...
+    @Provides
+    @Singleton
+    @Named("AllowedStripeIpAddresses")
+    public Set<String> allowedStripeIpAddresses(StripeGatewayConfig config) {
+        Set<String> allowedStripeIpAddresses = new HashSet<>();
+        config.getAllowedCidrs().stream().map(SubnetUtils::new).forEach(subnet -> {
+            subnet.setInclusiveHostCount(true);
+            allowedStripeIpAddresses.addAll(List.of(subnet.getInfo().getAllAddresses()));
+        });
+        return Set.copyOf(allowedStripeIpAddresses);
```
alexbishop1: OK, I’ve been tinkering with this a bit more:

```java
return config.getAllowedCidrs()
        .stream()
        .map(SubnetUtils::new)
        .peek(subnet -> subnet.setInclusiveHostCount(true))
        .map(SubnetUtils::getInfo)
        .map(SubnetInfo::getAllAddresses)
        .flatMap(Arrays::stream)
        .collect(toUnmodifiableSet());
```

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -73,7 +78,7 @@
 @ExtendWith(MockitoExtension.class)
 class StripeNotificationServiceTest {
     private static final String FORWARDED_IP_ADDRESSES = "1.2.3.4, 102.108.0.6";
-    private static final List<String> ALLOWED_IP_ADDRESSES = List.of("1.2.3.4", "9.9.9.9");
+    private static Set<String> allowedIpAddresses = new HashSet<>();
```
alexbishop1: Since it’s static, can this be `ALLOWED_IP_ADDRESSES`?

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -105,6 +110,13 @@
     private final String webhookTestSigningSecret = "whtest";
     private final String webhookLiveSigningSecret = "whlive";
 
+    @BeforeAll
+    static void setUpBeforeAll() {
+        addIpAddresses(allowedIpAddresses, "1.2.3.0/24");
+        addIpAddresses(allowedIpAddresses, "9.9.9.9/32");
+        allowedIpAddresses = Set.copyOf(allowedIpAddresses);
```
alexbishop1: We could avoid the need for the `addIpAddresses(…)` method by doing something similar to what I suggested for `ConnectorModule`:

```java
allowedIpAddresses = Set.of("1.2.3.0/24", "9.9.9.9/32")
        .stream()
        .map(SubnetUtils::new)
        .peek(subnet -> subnet.setInclusiveHostCount(true))
        .map(SubnetUtils::getInfo)
        .map(SubnetInfo::getAllAddresses)
        .flatMap(Arrays::stream)
        .collect(toUnmodifiableSet());
```

Actually, we might not even need the `@BeforeAll` method!

```java
private static final Set<String> ALLOWED_IP_ADDRESSES = Set.of("1.2.3.0/24", "9.9.9.9/32")
        .stream()
        .map(SubnetUtils::new)
        .peek(subnet -> subnet.setInclusiveHostCount(true))
        .map(SubnetUtils::getInfo)
        .map(SubnetInfo::getAllAddresses)
        .flatMap(Arrays::stream)
        .collect(toUnmodifiableSet());
```

Having code that’s near identical in `ConnectorModule` and this test sucks a little bit, so maybe we could have a public static method somewhere that takes a `Collection<String>` of CIDRs and returns a set of IP addresses. We could also then easily unit-test this to make sure it does reasonable conversions between CIDRs and IP addresses.

alexbishop1 (changes requested):
I think we can make this a little nicer and remove some duplication.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
adityapahuja: It cannot use `ALLOWED_IP_ADDRESSES` because it is not using `final` and is no longer constant.

alexbishop1: Oh, sorry, I missed that.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
adityapahuja: Thanks. It is little bit difficult to know what goes inside the stream. I will move it to a separate class and unit test it to ensure that it works.

adityapahuja: This has been done.

src/main/java/uk/gov/pay/connector/app/ConnectorModule.java
adityapahuja: Done.

alexbishop1 (approved):
This ended up quite nice in the end!

PR #2697, opened 2020-11-25
PP-6500 Wait for processing of payouts or charge captures before exit

danworth:
Wait to finish processing payouts and charge captures before exiting.
The `ExecutorService.shutdown` method only stops scheduling new tasks
and it doesn't wait for existing tasks to complete. Use
`awaitTermination` with a 15 second wait for any inflight payout or
charge processing to finish.

## WHAT YOU DID
The solution is based up on the example from the `ExecutorService` documentation https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ExecutorService.html#awaitTermination

Details of the investigation can be found in the Jira ticket here: https://payments-platform.atlassian.net/browse/PP-6500

#### Isn't there a bit of duplication?
Yes. This could be factored into some static method on a utility class which accepts an `ExecutorService` but the log entires are different and I'm not sure it makes things any simpler. Happy to be convinced otherwise.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/managed/CaptureMessageReceiver.java
+++ b/src/main/java/uk/gov/pay/connector/queue/managed/CaptureMessageReceiver.java
@@ -51,7 +51,26 @@ public class CaptureMessageReceiver implements Managed {
 
     @Override
     public void stop() {
+        LOGGER.info("Shutting down card capture service");
         chargeCaptureMessageExecutorService.shutdown();
+        try {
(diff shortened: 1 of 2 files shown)


alexbishop1 (commented):
Fits the pattern described at https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/ExecutorService.html

alexbishop1 (approved):

PR #2702, opened 2020-11-26
PP-7446 Add external_id to gateway account

kbottla:
## WHAT YOU DID
- Sets external_id when creating gateway account and includes the same in responses (get one/search gateway accounts)


stephencdaly (approved):
Looks good to me

PR #2703, opened 2020-11-26
PP-7280 - Use connector timestamp for USER_EMAIL_COLLECTED event

kbottla:
## WHY
- We currently emit `USER_EMAIL_COLLECTED` event to ledger with the timestamp of ENTERING_CARD_DETAILS/CREATED event.
  This was done as there is no charge event associated when frontend patches charge for email. We didn't also want to rely on connector timestamp as it may not be in sync with database or even other instances and can lead to out of order ledger events.
  `email` field has been removed from PAYMENT_DETAILS_ENTERED event due to new event USER_EMAIL_COLLECTED.

  This led to following issues:
  - Frontend can patch email multiple times and connector emits USER_EMAIL_COLLECTED events to ledger with same timestamp. But ledger processes only one event and ignores the rest as they are treated as duplicate events due to same event_name and timestamp. This is resulting in incorrect email address in ledger for transactions.
  - Wallet payments: It seems that for wallet payments, frontend doesn't patch charge, so email is not included in any event emitted to ledger. However, when connector tries to expunge charge, parity check fails on email field and a new event (`USER_EMAIL_COLLECTED`) is emitted. Until then ledger transactions will not have associated email address.

## WHAT
 - Use connector instance timestamp when `USER_EMAIL_COLLECTED` event is emitted. This ensures different events are emitted every time email is patched and ledger processes them all.
 - For backfilling we can continue to use ENTERING_CARD_DETAILS/CREATED charge event date. Emits event `BACKFILLER_RECREATED_USER_EMAIL_COLLECTED` to distinguish USER_EMAIL_COLLECTED events that happens when frontend patches charge with `email`
 - Added `email` back to PAYMENT_DETAILS_ENTERED event. This ensures `email` is emitted to ledger for wallet payments. There will be some duplication for card payments, as email will be included in both `USER_EMAIL_COLLECTED` and `PAYMENT_DETAILS_ENTERED` events. If we would like to avoid this, we can add email to PAYMENT_DETAILS_ENTERED event only for wallet payments.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -50,8 +50,8 @@ import uk.gov.pay.connector.common.model.domain.PaymentGatewayStateTransitions;
 import uk.gov.pay.connector.common.model.domain.PrefilledAddress;
 import uk.gov.pay.connector.common.service.PatchRequestBuilder;
 import uk.gov.pay.connector.events.EventService;
-import uk.gov.pay.connector.events.model.charge.UserEmailCollected;
 import uk.gov.pay.connector.events.model.charge.PaymentDetailsEntered;
+import uk.gov.pay.connector.events.model.charge.UserEmailCollected;
 import uk.gov.pay.connector.gateway.PaymentProviders;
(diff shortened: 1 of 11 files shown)


sfount (approved):
Looks great!

PR #2716, opened 2020-12-01
PP-7417: Service level test for valid 3ds flex credentials

oswaldquek:
Service level test for the following two scenarios in PP-7417:
Scenario 1: working live account credentials
Scenario 3: working test account credentials

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationService.java
@@ -1,0 +1,50 @@
+package uk.gov.pay.connector.gateway.worldpay;
+
+import io.dropwizard.setup.Environment;
+import org.apache.http.HttpStatus;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.charge.service.Worldpay3dsFlexJwtService;
+import uk.gov.pay.connector.gateway.ClientFactory;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccount;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.model.Worldpay3dsFlexCredentials;
+
+import javax.inject.Inject;
+import javax.ws.rs.client.Client;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.Response;
+import java.time.ZonedDateTime;
+import java.util.Map;
+
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
+
+public class Worldpay3dsFlexCredentialsValidationService {
+
+    private final Client client;
+    private final Worldpay3dsFlexJwtService worldpay3dsFlexJwtService;
+    private final Map<String, String> threeDsFlexDdcUrls;
+
+    @Inject
+    public Worldpay3dsFlexCredentialsValidationService(ClientFactory clientFactory,
+                                                       Environment environment,
+                                                       Worldpay3dsFlexJwtService worldpay3dsFlexJwtService,
+                                                       ConnectorConfiguration connectorConfiguration) {
+        this.worldpay3dsFlexJwtService = worldpay3dsFlexJwtService;
+        client = clientFactory.createWithDropwizardClient(WORLDPAY, environment.metrics());
+        threeDsFlexDdcUrls = connectorConfiguration.getWorldpayConfig().getThreeDsFlexDdcUrls();
+    }
+
+    public boolean validateCredentials(GatewayAccountEntity gatewayAccountEntity, Worldpay3dsFlexCredentials flexCredentials) {
+        String ddcToken = worldpay3dsFlexJwtService.generateDdcToken(GatewayAccount.valueOf(gatewayAccountEntity), 
+                flexCredentials, ZonedDateTime.now());
+
+        var formData = new MultivaluedHashMap<String, String>();
+        formData.add("JWT", ddcToken);
+
+        String url = threeDsFlexDdcUrls.get(gatewayAccountEntity.getType());
+
+        Response response = client.target(url).request().post(Entity.form(formData));
+        return response.getStatus() == HttpStatus.SC_OK;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccount.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccount.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableMap;
 
 import java.util.Map;
+import java.util.Objects;
 
 public class GatewayAccount {
     public static final String CREDENTIALS_MERCHANT_ID = "merchant_id";
@@ -47,5 +48,20 @@ public class GatewayAccount {
 
     public static GatewayAccount valueOf(GatewayAccountEntity entity) {
         return new GatewayAccount(entity.getId(), entity.getGatewayName(), entity.getCredentials());
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        GatewayAccount that = (GatewayAccount) o;
+        return Objects.equals(id, that.id) &&
+                Objects.equals(gatewayName, that.gatewayName) &&
+                Objects.equals(credentials, that.credentials);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(id, gatewayName, credentials);
     }
 }
--- a/src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
@@ -1,0 +1,126 @@
+package uk.gov.pay.connector.gateway.worldpay;
+
+import com.codahale.metrics.MetricRegistry;
+import io.dropwizard.setup.Environment;
+import org.apache.http.HttpStatus;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.app.WorldpayConfig;
+import uk.gov.pay.connector.charge.service.Worldpay3dsFlexJwtService;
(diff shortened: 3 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationService.java
```diff
@@ -0,0 +1,55 @@
...
+        boolean isValid = gatewayAccountService
+                .getGatewayAccount(gatewayAccountId)
+                .map(gatewayAccountEntity -> {
+                    String ddcToken = worldpay3dsFlexJwtService.generateDdcToken(
+                            GatewayAccount.valueOf(gatewayAccountEntity), flexCredentials, ZonedDateTime.now());
+                    var formData = new MultivaluedHashMap<String, String>();
+                    formData.add("JWT", ddcToken);
+                    String url = threeDsFlexDdcUrls.get(gatewayAccountEntity.getType());
+                    Response response = client.target(url).request().post(Entity.form(formData));
+                    return response.getStatus() == HttpStatus.SC_OK;
```
alexbishop1: Feel free to put some blank lines in this to break it up into logical parts.

src/main/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationService.java
```diff
@@ -0,0 +1,55 @@
...
+                }).get();
```
alexbishop1: This `get()` is going to crash and burn if the gateway account isn’t there. Can we do something more graceful, even if it’s throwing an exception?

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private void mockJerseyClient() {
+        MetricRegistry metricRegistry = mock(MetricRegistry.class);
+        when(environment.metrics()).thenReturn(metricRegistry);
+        when(clientFactory.createWithDropwizardClient(any(PaymentGatewayName.class), any(MetricRegistry.class)))
+                .thenReturn(client);
+        WebTarget webTarget = mock(WebTarget.class);
+        when(client.target(any(String.class))).thenReturn(webTarget);
+        when(webTarget.request()).thenReturn(invocationBuilder);
+        when(invocationBuilder.post(any(Entity.class))).thenReturn(response);
```
alexbishop1: Is there any way we can make this stubbing more specific? Right now, once the service has the gateway account, it could be making basically any request.

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private void mockWorldpay3dsFlexJwtService() {
+        when(worldpay3dsFlexJwtService.generateDdcToken(any(GatewayAccount.class), any(Worldpay3dsFlexCredentials.class), any(ZonedDateTime.class)))
```
alexbishop1: We should be able to stub exact values of the gateway account and its credentials.

alexbishop1 (changes requested):
Looks like a good start but I’m concerned the tests are a little too loose in terms of what they will allow.

src/main/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationService.java
oswaldquek: On balance I thought leaving it as `get()` would be ok for now as I have the following tests to implement:
```
    @Test
    @Disabled
    void gateway_account_does_not_exist() {
        //TODO implement
    }

    @Test
    @Disabled
    void gateway_account_is_not_a_worldpay_account() {
        //TODO implement
    }
```
and I want to make PRs as small as possible. But I can just throw a RuntimeException for now if you prefer.

alexbishop1: Since it’s currently unused code, I’m happy with whatever approach you want to take.

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
oswaldquek: I do have a `verify(client).target(threeDsFlexDdcUrls.get(type));` in the test which resolves the `client.target` part. I can add a verification for the `invocationBuilder.post`. I'd rather do this than make the stubbing very specific as I like how clean the tests look currently as I don't want to pollute them with `when...thenReturn` lines.

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
oswaldquek: I think this will be difficult as the Worldpay3dsFlexJwtService is called with a `ZonedDateTime.now()`: 
```worldpay3dsFlexJwtService.generateDdcToken(GatewayAccount.valueOf(gatewayAccountEntity), flexCredentials, ZonedDateTime.now())```
And unfortunately everything needs to be mocked exactly or mocked with `any`.
I can rectify this by using a `verify` in the test.

File: src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java (hunk 3 above)
Comment (alexbishop1) [reply src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java]: Yeah, it’s the actual POST body I’m most concerned about.

File: src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java (hunk 4 above)
Comment (alexbishop1) [reply src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java]: You have to use matchers for every argument but they don’t all need to be `any(…)`. You should be able to do something like this:

```java
when(worldpay3dsFlexJwtService.generateDdcToken(eq(exactGatewayAccount), eq(exactWorldpay3dsFlexCredentials), any(ZonedDateTime.class)))
```

oswaldquek: Ugh, it doesn't work :(
However this does ```verify(worldpay3dsFlexJwtService).generateDdcToken(eq(GatewayAccount.valueOf(gatewayAccount)), eq(flexCredentials), any(ZonedDateTime.class));```

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
oswaldquek: Unfortunately the Entity class doesn't have a nice equals method so I've gone with
```
        verify(invocationBuilder).post(entityArgumentCaptor.capture());
        assertEquals(((Form) entityArgumentCaptor.getValue().getEntity()).asMap(),
                new MultivaluedHashMap<String, String>() {{
                    add("JWT", DDC_TOKEN);
                }});
        assertEquals(entityArgumentCaptor.getValue().getMediaType(), MediaType.APPLICATION_FORM_URLENCODED_TYPE);

```

src/main/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationService.java
oswaldquek: I've made it throw a RuntimeException for now.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccount.java
```diff
@@ -48,4 +50,28 @@ public String getGatewayName() {
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+
+        if (o == null || getClass() != o.getClass()) return false;
+
+        GatewayAccount that = (GatewayAccount) o;
+
+        return new EqualsBuilder()
```
alexbishop1: `Objects.equals(…)` does basically the same thing and is part of the standard library.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccount.java
```diff
@@ -48,4 +50,28 @@ public String getGatewayName() {
...
+                .append(id, that.id)
+                .append(gatewayName, that.gatewayName)
+                .append(credentials, that.credentials)
+                .isEquals();
+    }
+
+    @Override
+    public int hashCode() {
+        return new HashCodeBuilder(17, 37)
```
alexbishop1: `Objects.hash(…)` does basically the same thing and is part of the standard library.

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
```diff
@@ -0,0 +1,157 @@
...
+        var gatewayAccount = aGatewayAccountEntity().withType(GatewayAccountEntity.Type.fromString(type)).build();
+        
+        when(gatewayAccountService.getGatewayAccount(gatewayAccount.getId())).thenReturn(Optional.of(gatewayAccount));
+        when(response.getStatus()).thenReturn(HttpStatus.SC_OK);
+        
+        var flexCredentials = new Worldpay3dsFlexCredentials("53f0917f101a4428b69d5fb0",
+                "57992a087a0c4849895ab8a2", "4cabd5d2-0133-4e82-b0e5-2024dbeddaa9"); 
+        assertTrue(service.validateCredentials(gatewayAccount.getId(), flexCredentials));
+
+        verify(client).target(threeDsFlexDdcUrls.get(type));
```
alexbishop1: I’m amazed we have to do this. Surely ` when(client.target(threeDsFlexDdcUrls.get(type))).thenReturn(webTarget)` works?

alexbishop1 (changes requested):
Looks good but can we have another go at trying use stubbing instead of verification? Happy to lend a hand when I have a free moment.

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
oswaldquek: I mentioned before about not liking polluting code with `when...thenReturn` lines as I personally find it distracting when reading test code, but putting in `when(client.target(threeDsFlexDdcUrls.get(type))).thenReturn(webTarget)` doesn't look so bad actually, so I will do that.

src/test/java/uk/gov/pay/connector/gateway/worldpay/Worldpay3dsFlexCredentialsValidationServiceTest.java
oswaldquek: I'm not sure why this didn't work for me yesterday but it does now 🤷‍♂️

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #2719, opened 2020-12-02
PP-7515 Update refund-ability on forced state transitions

sfount:
Charge state can be force updated to match the gateway when we receive a
notification, the notification is treated as the truth.

This impacts the refundability of the charge which should be reflected
in the event factory.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
@@ -21,6 +21,9 @@ import uk.gov.pay.connector.events.model.charge.PaymentEvent;
 import uk.gov.pay.connector.events.model.charge.PaymentNotificationCreated;
 import uk.gov.pay.connector.events.model.charge.RefundAvailabilityUpdated;
 import uk.gov.pay.connector.events.model.charge.UnexpectedGatewayErrorDuringAuthorisation;
+import uk.gov.pay.connector.events.model.charge.StatusCorrectedToCapturedToMatchGatewayStatus;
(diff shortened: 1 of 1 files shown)


danworth (approved):
Looks like it'll do it. Are there tests to update?

sfount:
@danworth There's an existing test that verifies that anything in this list gets shipped off with a refund-ability updated event, for this exact scenario I'm trusting that I think. We could consider what our end-to-end paths look like given some of these less called journey branches

src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
```diff
@@ -57,7 +60,10 @@
             CaptureSubmitted.class,
             GatewayErrorDuringAuthorisation.class,
             GatewayTimeoutDuringAuthorisation.class,
-            UnexpectedGatewayErrorDuringAuthorisation.class
+            UnexpectedGatewayErrorDuringAuthorisation.class,
+            StatusCorrectedToAuthorisationErrorToMatchGatewayStatus.class,
+            StatusCorrectedToAuthorisationRejectedToMatchGatewayStatus.class,
+            StatusCorrectedToCapturedToMatchGatewayStatus.class
```
kbottla, line 66: These are fine but there is one more `CaptureConfirmedByGatewayNotification` that need to be in the list ? 

Also for `StatusCorrectedToCapturedToMatchGatewayStatus` we need to add event details to include captured_date and anything else possible from https://github.com/alphagov/pay-connector/blob/57c4efd5e71578f0150ab5cda10f11eee39811f1/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/CaptureConfirmedEventDetails.java#L10

PR #2721, opened 2020-12-02
PP-7417: Handle exceptions and invalid response status codes

oswaldquek:
This handles scenarios 5 and 6 of PP-7417.


adityapahuja (approved):
Look good to me.

PR #2723, opened 2020-12-03
PP-7417: Integration tests for /v1/api/accounts/{accountId}/worldpay/…

oswaldquek:
…check-3ds-flex-config

Tests for valid and invalid account credentials

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResource.java
@@ -1,6 +1,8 @@
 package uk.gov.pay.connector.gatewayaccount.resource;
 
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
+import uk.gov.pay.connector.gateway.worldpay.Worldpay3dsFlexCredentialsValidationService;
+import uk.gov.pay.connector.gatewayaccount.model.Worldpay3dsFlexCredentials;
 import uk.gov.pay.connector.gatewayaccount.model.Worldpay3dsFlexCredentialsRequest;
 import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
 import uk.gov.pay.connector.gatewayaccount.service.Worldpay3dsFlexCredentialsService;
@@ -22,12 +24,15 @@ public class GatewayAccount3dsFlexCredentialsResource {
 
     private final GatewayAccountService gatewayAccountService;
     private final Worldpay3dsFlexCredentialsService worldpay3dsFlexCredentialsService;
+    private final Worldpay3dsFlexCredentialsValidationService worldpay3dsFlexCredentialsValidationService;
 
     @Inject
     public GatewayAccount3dsFlexCredentialsResource(GatewayAccountService gatewayAccountService,
-                                                    Worldpay3dsFlexCredentialsService worldpay3dsFlexCredentialsService) {
+                                                    Worldpay3dsFlexCredentialsService worldpay3dsFlexCredentialsService, 
+                                                    Worldpay3dsFlexCredentialsValidationService worldpay3dsFlexCredentialsValidationService) {
         this.gatewayAccountService = gatewayAccountService;
         this.worldpay3dsFlexCredentialsService = worldpay3dsFlexCredentialsService;
+        this.worldpay3dsFlexCredentialsValidationService = worldpay3dsFlexCredentialsValidationService;
     }
 
     @POST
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResource.java
```diff
@@ -47,4 +52,30 @@ public Response createOrUpdateWorldpay3dsCredentials(@PathParam("accountId") Lon
...
+        private final String result;
+
+        private ValidationResult(boolean isValid) {
+            this.result = isValid ? "valid" : "invalid";
+        }
+
+        public String getResult() {
+            return result;
+        }
+    }
```
nimalank7, line 80: A POJO in a controller? Could we not move it into a separate class?

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResourceIT.java
```diff
@@ -50,10 +61,54 @@ public void setUp() {
...
+    }
+    
+    @Test
+    public void validate_valid_3ds_flex_credentials() throws Exception {
+        stubFor(post("/shopper/3ds/ddc.html").willReturn(ok()));
+
+        var validIssuer = "54i0917n10va4428b69l5id0";
+        var validOrgUnitId = "57992i087n0v4849895alid2";
+        var validJwtMacKey = "4inva5l2-0133-4i82-d0e5-2024dbeddaa9";
+        var payload = new ObjectMapper().writeValueAsString(Map.of(
```
nimalank7: Could we move new ObjectMapper() into the `@Before` section as we're reusing it for all the tests as far as I can see.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResource.java
oswaldquek: Given this class is really simple and only used in this resource class, I'd rather keep this as an inner class.

nimalank7 (changes requested):
Looks good. Just left a few comments. Can we also convert this to a JUnit 5 test?

oswaldquek:
@nimalank7 I don't think it'd be good idea to convert this to a JUnit5 test as that might be quite a significant piece of work. I'm not sure how well DropwizardJUnitRunner would play with JUnit5 so if we did want to convert to JUnit5 it'd best be left as it's own piece of work/PR.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResourceIT.java
oswaldquek: Yeah, good shout. Done.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResource.java
oswaldquek: The problem with making this its own class is that it would indicate it could be used by other classes and people might start doing stuff to it (e.g. adding fields) to reuse it which would make its original usage confusing. There's a point where generalisation of code is unproductive.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResourceIT.java
```diff
@@ -50,13 +62,57 @@ public void setUp() {
...
+                .body(payload)
+                .post(format(VALIDATE_3DS_FLEX_CREDENTIALS_URL, testAccount.getAccountId()))
+                .then()
+                .statusCode(200)
+                .body("result", is("invalid"));
+    }
+
+    @Test
+    public void should_return_503_if_error_communicating_with_3ds_flex_ddc_endpoint() {
+        //TODO implement
```
nimalank7, line 110: Forgot about this! This can be removed as presumably in a follow up PR you'll implement this?

nimalank7:
@oswaldquek sure that's fine. One final comment which I just spotted regarding the //TODO

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResource.java
nimalank7: Sure.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccount3dsFlexCredentialsResourceIT.java
oswaldquek: Yeah, but in the interests of not waiting for another build, can we leave this?

oswaldquek: Yeah, it will be in the next PR. In the interests of not wanting to wait for another build, can we just leave this?

nimalank7: Sure

nimalank7 (approved):

PR #2738, opened 2020-12-09
PP-5640: Add exemption request to authorisation order

oswaldquek:
An exemption request should be added to the authorisation order if:
- Worldpay gateway account has exemption_engine set to true in the
  worldpay_3ds_flex_credentials table
- the gateway account has 3-D Secure enabled

This PR actually covers all the scenarios in PP-5640.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
@@ -31,7 +31,7 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
         private String paResponse3ds;
         private String payerIpAddress;
         private String state;
-        private boolean exemptionEngine;
+        private boolean exemptionEngineEnabled;
 
         public String getReference() {
             return reference;
@@ -115,12 +115,12 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
             this.state = state;
         }
 
-        public boolean isExemptionEngine() {
-            return exemptionEngine;
+        public boolean isExemptionEngineEnabled() {
+            return exemptionEngineEnabled;
         }
 
-        public void setExemptionEngine(boolean exemptionEngine) {
-            this.exemptionEngine = exemptionEngine;
+        public void setExemptionEngineEnabled(boolean exemptionEngineEnabled) {
+            this.exemptionEngineEnabled = exemptionEngineEnabled;
         }
     }
 
@@ -212,7 +212,7 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
     }
     
     public WorldpayOrderRequestBuilder withExemptionEngine(boolean exemptionEngine) {
-        worldpayTemplateData.setExemptionEngine(exemptionEngine);
+        worldpayTemplateData.setExemptionEngineEnabled(exemptionEngine);
         return this;
     }
 
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
@@ -31,6 +31,7 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
         private String paResponse3ds;
         private String payerIpAddress;
         private String state;
+        private boolean exemptionEngine;
 
         public String getReference() {
             return reference;
@@ -112,6 +113,14 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
 
         public void setState(String state) {
             this.state = state;
+        }
+
+        public boolean isExemptionEngine() {
+            return exemptionEngine;
+        }
+
+        public void setExemptionEngine(boolean exemptionEngine) {
+            this.exemptionEngine = exemptionEngine;
         }
     }
 
@@ -199,6 +208,11 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
     public WorldpayOrderRequestBuilder with3dsRequired(boolean requires3ds) {
         logger.info("3DS requirement is: " + requires3ds + " for " + worldpayTemplateData.sessionId);
         worldpayTemplateData.setRequires3ds(requires3ds);
+        return this;
+    }
+    
+    public WorldpayOrderRequestBuilder withExemptionEngine(boolean exemptionEngine) {
+        worldpayTemplateData.setExemptionEngine(exemptionEngine);
         return this;
     }
 
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -222,16 +222,14 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
         boolean is3dsRequired = request.getAuthCardDetails().getWorldpay3dsFlexDdcResult().isPresent() ||
                 request.getGatewayAccount().isRequires3ds();
 
-        boolean exemptionEngine = request.getGatewayAccount().isRequires3ds() ? 
-                Optional.ofNullable(request.getGatewayAccount().getWorldpay3dsFlexCredentials())
-                        .map(Worldpay3dsFlexCredentials::isExemptionEngine)
-                        .orElse(false) 
-                : false;
+        boolean exemptionEngineEnabled = Optional.ofNullable(request.getGatewayAccount().getWorldpay3dsFlexCredentials())
+                        .map(Worldpay3dsFlexCredentials::isExemptionEngineEnabled)
+                        .orElse(false);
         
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
```diff
@@ -31,6 +31,7 @@
         private String paResponse3ds;
         private String payerIpAddress;
         private String state;
+        private boolean exemptionEngine;
```
alexbishop1: `exemptionEngineEnabled` maybe? Just to make the method names read more naturally.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -221,9 +222,16 @@ private GatewayOrder buildAuthoriseOrder(CardAuthorisationGatewayRequest request
         boolean is3dsRequired = request.getAuthCardDetails().getWorldpay3dsFlexDdcResult().isPresent() ||
                 request.getGatewayAccount().isRequires3ds();
 
+        boolean exemptionEngine = request.getGatewayAccount().isRequires3ds() ? 
+                Optional.ofNullable(request.getGatewayAccount().getWorldpay3dsFlexCredentials())
+                        .map(Worldpay3dsFlexCredentials::isExemptionEngine)
+                        .orElse(false) 
+                : false;
```
alexbishop1: This can be rewritten as:

```java
boolean exemptionEngine = request.getGatewayAccount().isRequires3ds() &&
    Optional.ofNullable(request.getGatewayAccount().getWorldpay3dsFlexCredentials())
            .map(Worldpay3dsFlexCredentials::isExemptionEngine)
            .orElse(false);
```

However, I wonder if we actually need this flag to consider whether or not 3DS is enabled. Presumably it’s just going to be used inside the template and we can nest it inside a check that 3DS is enabled.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentials.java
```diff
@@ -14,13 +14,25 @@
 
     @JsonIgnore
     private String jwtMacKey;
+    private boolean exemptionEngine;
```
alexbishop1: Do we need to update the `equals(…)` and `hashCode()` methods here?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentials.java
```diff
@@ -14,13 +14,25 @@
...
+    }
 
     public Worldpay3dsFlexCredentials(String issuer, String organisationalUnitId, String jwtMacKey) {
         this.issuer = issuer;
         this.organisationalUnitId = organisationalUnitId;
         this.jwtMacKey = jwtMacKey;
     }
 
+    public boolean isExemptionEngine() {
+        return exemptionEngine;
```
alexbishop1: Looks like this could return null in some cases. Do we want that or should we default it to false?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -333,15 +334,119 @@ void should_include_completed_authentication_in_second_order_when_paRequest_not_
...
+                eq(WORLDPAY_URL),
+                eq(gatewayAccountEntity),
+                gatewayOrderArgumentCaptor.capture(),
+                anyMap());
+
+        Document document = XPathUtils.getDocumentXmlString(gatewayOrderArgumentCaptor.getValue().getPayload());
+        assertThat(getNodeListFromExpression(document, "/paymentService/submit/order/exemption").getLength(),
+                is(0));
+    }
+
```
alexbishop1, line 425: Is it feasible to remove some of the duplication in these tests?

alexbishop1 (changes requested):
Looks like it should do the job. Just a few small things.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: It looks like the following works in the template:
```            
            <#if requires3ds && exemptionEngineEnabled>
                <exemption type="OP" placement="AUTHORISATION"/>
            </#if>
```

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentials.java
oswaldquek: I'm not sure... I think it's the Worldpay3dsFlexCredentials that could be null, not the exemption engine flag which is by default false in the database?

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
oswaldquek: 👍

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: Done

oswaldquek:
@alexbishop1 Thanks for the review. I've addressed the comments in a few commit to make viewing the changes easier. Happy to merge the commits if everything is ok.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentials.java
alexbishop1: The existing constructor will leave it as null if it’s invoked (maybe that constructor isn’t required any more).

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentialsEntity.java
```diff
@@ -49,6 +52,10 @@ public Worldpay3dsFlexCredentialsEntity(Long gatewayAccountId, String issuer, St
         this.jwtMacKey = jwtMacKey;
     }
 
+    public boolean isExemptionEngine() {
+        return exemptionEngine;
+    }
```
alexbishop1: I think this could do with the `Enabled` treatment too.

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -53,6 +53,9 @@
             />
             </#if>
             </#if>
+            <#if requires3ds && exemptionEngineEnabled>
+                <exemption type="OP" placement="AUTHORISATION"/>
+            </#if>
```
alexbishop1: If this was moved a line up so it’s above the `</#if>` that closes the `<#if requires3ds>`, it could just be `<#if exemptionEngineEnabled>`.

alexbishop1 (changes requested):
Looks good, just a couple of tiny things.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentials.java
oswaldquek: Won't it be `false` by default?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentialsEntity.java
oswaldquek: Done

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
oswaldquek: Done

src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentials.java
alexbishop1: Yeah, you’re right, it will. But is the constructor that doesn’t set it even needed now?

alexbishop1 (dismissed):
This looks good. Can you check if the three-arg variant of the `Worldpay3dsFlexCredentials` constructor can be removed? If so, please do that, otherwise I think this is good to go.

oswaldquek:
@alexbishop1 Strictly speaking the three-arg constructor can be removed. I didn't because it's used in a few places (mainly tests) so I thought it'd be convenient to leave it.
I've discovered that https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/model/Worldpay3dsFlexCredentialsEntity.java#L44-L50 can be removed though!

alexbishop1 (approved):
Looks good!

PR #2743, opened 2020-12-11
PP-5640: Contract test for exemption engine flag

oswaldquek:


alexbishop1 (approved):
I’m curious to know what the response says about if the exemption was honoured or not.

oswaldquek:
It's `lastEvent: REFUSED, ISO8583ReturnCode code: 65, ISO8583ReturnCode description: Authentication requested`. Makes sense as the GDS account I'm using probably doesn't have the exemption engine set up.

alexbishop1:
Did you happen to see if it had `<exemptionResponse result="REJECTED"/>` or `<exemptionResponse result="OUT_OF_SCOPE">` in the response? We’d treat that as a soft decline and retry without the exemption. But if we don’t see either of those we’d treat it as a rejection authorisation and not try again, so we’d have to be really careful to make sure we don’t send exemptions for accounts that don’t have the Exemption Engine enabled.

oswaldquek:
Yes it does have `<exemptionResponse result="REJECTED" reason="NOT_SUBSCRIBED"/>`.
(For some reason github doesn't like when I post the entire xml response)

alexbishop1:
Ah, that’s good. We plan to treat anything with `<exemptionResponse result="REJECTED">` as a soft decline and retry without the exemption.

PR #2746, opened 2020-12-11
PP-5641: Add exemption to authorisation request summary

oswaldquek:


src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthorisationRequestSummaryTest.java
```diff
@@ -25,6 +27,35 @@
...
+        given(mockChargeEntity.getGatewayAccount()).willReturn(mockGatewayAccountEntity);
+        given(mockGatewayAccountEntity.isRequires3ds()).willReturn(true);
+        given(mockGatewayAccountEntity.getWorldpay3dsFlexCredentials()).willReturn(mockWorldpay3dsFlexCredentials);
+        given(mockWorldpay3dsFlexCredentials.isExemptionEngineEnabled()).willReturn(false);
+        var worldpayAuthorisationRequestSummary = new WorldpayAuthorisationRequestSummary(mockChargeEntity, mockAuthCardDetails);
+        assertThat(worldpayAuthorisationRequestSummary.exemptionRequest(), is(NOT_PRESENT));
+    }
+
+    @Test
+    void requires3ds_false_means_exemption_request_not_present() {
```
alexbishop1, line 53: To be even more rigorous, this should enable the Exemption Engine.

alexbishop1 (changes requested):
Looks good, just one thing about a test.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthorisationRequestSummaryTest.java
oswaldquek: Sorry but what do you mean by that?

alexbishop1: Sorry, I was thinking it should enable the Exemption Engine but have 3DS disabled so we’re checking that either of those things being false means we’ll log there’s no exemption. But Mockito won’t like that because of the unnecessary stubbing.

alexbishop1 (approved):

PR #2750, opened 2020-12-14
PP-5641: The worldpay 3ds flex creds of a gateway account is optional

oswaldquek:
It is difficult to realise the worldpay 3ds flex creds may not be present in
code. This caused an end-to-end test error for
https://github.com/alphagov/pay-connector/pull/2746.

The effect of this is that the `worldpay_3ds_flex` in the JSON of a gatweay
account is serialised as `"worldpay_3ds_flex": null` rather than it not being
there at all. This could be rectified by placing
`@JsonProperty("worldpay_3ds_flex")` in the GatewayAccountEntity on the
`worldpay3dsFlexCredentialsEntity` field rather than the method
`getWorldpay3dsFlexCredentials`. However this results in a StackOverflow error
in some tests.


rorymalcolm (approved):
LGTM

PR #2755, opened 2020-12-14
PP-5641: Move authorisation code into WorldpayAuthoriseHandler

oswaldquek:
Given we are going to introduce more logic to retry authorisations in response
to a soft decline, it seems good to extract authorisation code into its own
class so that retry logic can go there in following commits. This follows the
pattern of StripeAuthoriseHandler, WorldpayCaptureHandler,
WorldpayRefundHandler etc.

All the test code are pretty much copy and paste, there should be no changes.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandler.java
@@ -1,0 +1,116 @@
+package uk.gov.pay.connector.gateway.worldpay;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.gateway.AuthoriseHandler;
+import uk.gov.pay.connector.gateway.GatewayClient;
+import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.GatewayOrder;
+import uk.gov.pay.connector.gateway.model.GatewayError;
+import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
+import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.model.Worldpay3dsFlexCredentials;
+
+import java.net.URI;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandler.java
```diff
@@ -0,0 +1,116 @@
...
+import static org.apache.http.HttpStatus.SC_UNAUTHORIZED;
+import static uk.gov.pay.connector.gateway.model.GatewayError.gatewayConnectionError;
+import static uk.gov.pay.connector.gateway.model.response.GatewayResponse.GatewayResponseBuilder.responseBuilder;
+import static uk.gov.pay.connector.gateway.util.AuthUtil.getGatewayAccountCredentialsAsAuthHeader;
+import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderRequestBuilder.aWorldpayAuthoriseOrderRequestBuilder;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_MERCHANT_ID;
+
+public class WorldpayAuthoriseHandler implements AuthoriseHandler, WorldpayGatewayResponseGenerator {
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
```
adityapahuja, line 30:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(WorldpayAuthoriseHandler.class);
```

oswaldquek: We seem to be using `LoggerFactory.getLogger(getClass())` throughout our code base. I'm not sure what the pros and cons are between using that and `private static final Logger LOGGER = LoggerFactory.getLogger(Clazz.class)`?

adityapahuja: Including `static` can create one Logger per class. Without `static`, there will be one logger per instance of this class.  There is no need to have more than one instance for this class. It saves CPU time and smaller memory footprint. SLF4J has stopped recommending that loggers members be declared as instance variables instead of static.

Source: http://slf4j.org/faq.html#declared_static

adityapahuja (approved):
All other changes look fine. There is a minor suggestion but it should not block.

PR #2756, opened 2020-12-15
PP-5641: Build auth order in response to soft decline

oswaldquek:
In the event where Worldpay responds to an exemption request with a soft
decline, the auth request should be sent again with no `<exemption>` element.
This commit doesn't contain code to retry the auth request, but to ensure an
auth request without an `<exemption>` element can be created even for a gateway
account with exemption engine enabled.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java
@@ -27,10 +27,10 @@ import static java.util.Collections.emptyList;
 import static javax.ws.rs.core.Response.Status.Family.SUCCESSFUL;
 import static javax.ws.rs.core.Response.Status.Family.familyOf;
 import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
-import static javax.ws.rs.core.Response.Status.OK;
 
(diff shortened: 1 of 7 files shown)


heathd:
Big fan on functional programming where it adds clarity or simplicity, but I'm not convinced that we're achieving that here. 

From what I can understand `buildAuthoriseOrder` operations which do slightly different but related things

You could achieve that by using the "template method" implementation of `buildAuthoriseOrder` in a base class and two different subclasses of `WorldpayAuthoriseOrderBuilder` with concrete implementations of `buildAuthoriseOrder` specialising the different behaviours. Or even just controlling the behaviour of the order build when an instance of it is created. 

I may not be seeing the bigger picture as I've not looked at the whole story. There may be a reason why the extra flexibility is valuable that I'm not aware of.

oswaldquek:
@heathd Cheers for that. Longer term I thought method signature could become
```
GatewayOrder buildOrder(WorldpayOrderRequestBuilder builder, GatewayRequest request, Strategy strategy)
```
that could be used by the other types or orders (cancel, refund, auth3ds) but this wouldn't work as the GatewayRequests are different for each order type.
So I think you're right.

adityapahuja (approved):

PR #2758, opened 2020-12-15
PP-7422: Add mocked Worldpay DDC server for PACT

adityapahuja:
This change adds mocked Worldpay Device Data Collection (DDC) server to Self Service Contract test. This will allow us to perform PACT test on Worldpay 3DS Flex credentials verification between Self Service and Connector applications.

I ran Self Service Contract Test locally and it passed all tests without errors.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/pact/SelfServiceContractTest.java
+++ b/src/test/java/uk/gov/pay/connector/pact/SelfServiceContractTest.java
@@ -4,7 +4,14 @@ import au.com.dius.pact.provider.junit.PactRunner;
 import au.com.dius.pact.provider.junit.Provider;
 import au.com.dius.pact.provider.junit.loader.PactBroker;
 import au.com.dius.pact.provider.junit.loader.PactBrokerAuth;
+import com.github.tomakehurst.wiremock.junit.WireMockClassRule;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
 import org.junit.runner.RunWith;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.post;
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
 
 @RunWith(PactRunner.class)
 @Provider("connector")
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/pact/SelfServiceContractTest.java
```diff
@@ -4,12 +4,26 @@
...
+import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
+
 @RunWith(PactRunner.class)
 @Provider("connector")
 @PactBroker(scheme = "https", host = "${PACT_BROKER_HOST:pact-broker-test.cloudapps.digital}", tags = {"${PACT_CONSUMER_TAG}", "test"},
         authentication = @PactBrokerAuth(username = "${PACT_BROKER_USERNAME}", password = "${PACT_BROKER_PASSWORD}"),
         consumers = {"selfservice"})
 public class SelfServiceContractTest extends ContractTest {
+    @ClassRule
+    public static WireMockClassRule worldpayDeviceDataCollectionRule = new WireMockClassRule(10108);
```
oswaldquek: It would be better to use a random port for this, e.g. `int worldpay3dsFlexDdcPort = PortFactory.findFreePort()`. 
Then you can initialise [DropwizardAppWithPostgresRule](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/pact/ContractTest.java#L50) like:
```
public static DropwizardAppWithPostgresRule app = new DropwizardAppWithPostgresRule(
ConfigOverride.config("captureProcessConfig.backgroundProcessingEnabled", "false"),
ConfigOverride.config("worldpay. threeDsFlexDdcUrls.test", format("http://localhost:%s/shopper/3ds/ddc.html", worldpay3dsFlexDdcPort)),
ConfigOverride.config("worldpay. threeDsFlexDdcUrls.live", format("http://localhost:%s/shopper/3ds/ddc.html", worldpay3dsFlexDdcPort))
);
```

oswaldquek: Because we also use random ports [here](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/pact/ContractTest.java#L53) there's always a chance the ports could clash.

oswaldquek (commented):
Nice. Great to know this works. Just need to use a random port and then we're done.

src/test/java/uk/gov/pay/connector/pact/SelfServiceContractTest.java
adityapahuja: I agree. I tried this approach but did not take it further. I have put the original changes back.

oswaldquek (approved):

PR #2763, opened 2020-12-16
Pp 5641 scenario 2

oswaldquek:
These commits fulfil scenario 2 of PP-5641.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -91,11 +91,8 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
     }
     
     public boolean isSoftDecline() {
-        return Optional.ofNullable(lastEvent)
-                .map(e -> e.equals("REFUSED"))
-                .map(e -> e && exemptionResponseResult != null)
-                .map(e -> e && SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS.contains(exemptionResponseResult))
-                .orElse(false);
+        return Optional.ofNullable(lastEvent).map("REFUSED"::equals).orElse(false)
+                && Optional.ofNullable(exemptionResponseResult).map(SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS::contains).orElse(false);
     }
 
     public void setLastEvent(String lastEvent) {
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -101,8 +101,8 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
         return issuerUrl;
     }
 
-    public String getLastEvent() {
-        return lastEvent;
+    public Optional<String> getLastEvent() {
+        return Optional.ofNullable(lastEvent).map(r -> isBlank(r) ? null : lastEvent);
     }
 
     public String getChallengeAcsUrl() {
@@ -201,8 +201,8 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
         if (isNotBlank(getTransactionId())) {
             joiner.add("orderCode: " + getTransactionId());
         }
-        if (isNotBlank(getLastEvent())) {
-            joiner.add("lastEvent: " + getLastEvent());
+        if (isNotBlank(lastEvent)) {
+            joiner.add("lastEvent: " + lastEvent);
         }
         if (isNotBlank(getRefusedReturnCode())) {
             joiner.add("ISO8583ReturnCode code: " + getRefusedReturnCode());
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -8,15 +8,16 @@ import uk.gov.pay.connector.gateway.model.response.BaseInquiryResponse;
 
 import javax.xml.bind.annotation.XmlRootElement;
 import java.util.Optional;
+import java.util.Set;
 import java.util.StringJoiner;
 
-import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.apache.commons.lang3.StringUtils.trim;
 
 @XmlRootElement(name = "paymentService")
 public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseCancelResponse, BaseInquiryResponse {
 
+    private static final Set<String> SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS = Set.of("REJECTED", "OUT_OF_SCOPE");
     private static final String WORLDPAY_AUTHORISED_EVENT = "AUTHORISED";
     private static final String WORLDPAY_REFUSED_EVENT = "REFUSED";
     private static final String WORLDPAY_CANCELLED_EVENT = "CANCELLED";
@@ -88,9 +89,13 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
     public void setChallengeAcsUrl(String challengeAcsUrl) {
         this.challengeAcsUrl = challengeAcsUrl != null ? challengeAcsUrl.trim() : null;
     }
-
-    public Optional<String> getExemptionResponseResult() {
-        return Optional.ofNullable(exemptionResponseResult).map(r -> isBlank(r) ? null : exemptionResponseResult);
+    
+    public boolean isSoftDecline() {
+        return Optional.ofNullable(lastEvent)
+                .map(e -> e.equals("REFUSED"))
+                .map(e -> e && exemptionResponseResult != null)
+                .map(e -> e && SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS.contains(exemptionResponseResult))
+                .orElse(false);
     }
 
     public String getPaRequest() {
@@ -102,7 +107,7 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
     }
 
     public Optional<String> getLastEvent() {
-        return Optional.ofNullable(lastEvent).map(r -> isBlank(r) ? null : lastEvent);
+        return Optional.ofNullable(lastEvent).map(r -> r.isBlank() ? null : lastEvent);
     }
 
     public String getChallengeAcsUrl() {
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -98,6 +98,14 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
                 .orElse(false);
     }
 
+    public void setLastEvent(String lastEvent) {
+        this.lastEvent = lastEvent;
+    }
+
+    public void setExemptionResponseResult(String exemptionResponseResult) {
+        this.exemptionResponseResult = exemptionResponseResult;
+    }
+
     public String getPaRequest() {
         return paRequest;
     }
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -1,6 +1,5 @@
 package uk.gov.pay.connector.gateway.worldpay;
 
-import org.apache.commons.lang3.StringUtils;
 import org.eclipse.persistence.oxm.annotations.XmlPath;
 import uk.gov.pay.connector.gateway.model.Gateway3dsRequiredParams;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
@@ -11,6 +10,8 @@ import javax.xml.bind.annotation.XmlRootElement;
 import java.util.Optional;
 import java.util.StringJoiner;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.apache.commons.lang3.StringUtils.trim;
 
 @XmlRootElement(name = "paymentService")
@@ -86,6 +87,10 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
     @XmlPath("/reply/orderStatus/challengeRequired/threeDSChallengeDetails/acsURL/text()")
     public void setChallengeAcsUrl(String challengeAcsUrl) {
         this.challengeAcsUrl = challengeAcsUrl != null ? challengeAcsUrl.trim() : null;
+    }
+
+    public Optional<String> getExemptionResponseResult() {
+        return Optional.ofNullable(exemptionResponseResult).map(r -> isBlank(r) ? null : exemptionResponseResult);
     }
 
     public String getPaRequest() {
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -136,7 +152,43 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
+                        request.getCharge().getGatewayAccount().getAnalyticsId(),
+                        request.getCharge().getGatewayAccount().getId(),
+                        response, 
+                        ChargeStatus.fromString(request.getCharge().getStatus()), request.getCharge().getStatus());
+
+                var structuredLoggingArguments = ArrayUtils.addAll(
+                        request.getCharge().getStructuredLoggingArgs(),
+                        authorisationRequestSummaryStructuredLogging.createArgs(authorisationRequestSummary));
+
+                LOGGER.info(logMessage, structuredLoggingArguments);
```
oswaldquek: This has been copied from CardAuthoriseService. I'd like to extract this out into a LoggerForSplunk class of some sort but that would done in another PR.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -88,6 +89,10 @@ public void setChallengeAcsUrl(String challengeAcsUrl) {
         this.challengeAcsUrl = challengeAcsUrl != null ? challengeAcsUrl.trim() : null;
     }
 
+    public Optional<String> getExemptionResponseResult() {
+        return Optional.ofNullable(exemptionResponseResult).map(r -> isBlank(r) ? null : exemptionResponseResult);
```
alexbishop1: Java 11+ has an `isBlank()` method on `String`, so we can use that if we know the string isn’t null (which of course it won’t be here).

Also, we can use `filter(…)` and `Predicate.not(…)` here:

```java
return Optional.ofNullable(exemptionResponseResult).filter(not((String::isBlank)));
```

However, do we really think we’re going to get an `exemptionResponse` element with an blank `result` attribute? The docs at https://developer.worldpay.com/docs/wpg/scaexemptionservices/exemptionengine#exemption-response say there’s only three possible values, so I think we can assume it won’t happen.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -53,7 +58,9 @@
 public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGatewayResponseGenerator {
 
     static final String WORLDPAY_MACHINE_COOKIE_NAME = "machine";
+    
     private static final Logger LOGGER = LoggerFactory.getLogger(WorldpayPaymentProvider.class);
+    private static final List<String> SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS = List.of("REJECTED", "OUT_OF_SCOPE");
```
alexbishop1: A `Set` is probably more appropriate here (though it feels like we should be able to do all this with enums).

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -136,7 +152,43 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
 
     @Override
     public GatewayResponse<WorldpayOrderStatusResponse> authorise(CardAuthorisationGatewayRequest request) {
-        return worldpayAuthoriseHandler.authorise(request);
+        
+        GatewayResponse<WorldpayOrderStatusResponse> response = worldpayAuthoriseHandler.authorise(request);
+        
+        if (response.getBaseResponse().isPresent()) {
```
alexbishop1: Is there someone we can yank these 30 or so lines to a separate method or, even better, a separate class?

Ideally, we probably want it to look something like this:

```java
GatewayResponse<WorldpayOrderStatusResponse> response = worldpayAuthoriseHandler.authoriseWithPossibleExemption(request);
if (worldpayExemptionResponseFigurerOuterer.isSoftDecline()) {
    // Some logging stuff here?
    response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
}
return response;
```

alexbishop1: Should be able to use `ifPresent(…)` here.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -136,7 +152,43 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
+        
+            var worldpayOrderResponse = response.getBaseResponse().get();
+        
+            if (worldpayOrderResponse.getLastEvent().isPresent() && 
+                    worldpayOrderResponse.getLastEvent().get().equals("REFUSED") && 
+                    worldpayOrderResponse.getExemptionResponseResult().isPresent() &&
+                    SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS.contains(worldpayOrderResponse.getExemptionResponseResult().get())) {
```
alexbishop1: A little gnarly. How about something like this?

```java
boolean lastEventIsRefused = worldpayOrderResponse.getLastEvent().map("REFUSED"::equals).orElse(false);
boolean exemptionResultIsRejectedOrOutOfScope = worldpayOrderResponse.getExemptionResponseResult()
            .map(SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS::contains)
            .orElse(false);

if (lastEventIsRefused && exemptionResultIsRejectedOrOutOfScope) {
    // etc. etc.
}
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -136,7 +152,43 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
+
+                var authorisationRequestSummary = generateAuthorisationRequestSummary(request.getCharge(), request.getAuthCardDetails());
+
+                Optional<String> transactionId = authorisationService.extractTransactionId(request.getCharge().getExternalId(), response);
+                
+                // Used by Splunk saved search
```
alexbishop1: This probably won’t be used in a saved search, at least not initially

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
alexbishop1: Yeah, this is proliferating. As well `CardAuthoriseService`, we also have this in the Apple Pay and Google Pay authorisation code and probably somewhere else. We probably want a separate class we can just throw responses at to get logged whenever we deem it necessary.

alexbishop1 (changes requested):
Logic looks fine but I think we could try to encapsulate some of it a bit more.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: Logging aside (which I'm going to get rid in another PR), this is what it looks like now:
```
        GatewayResponse<WorldpayOrderStatusResponse> response = worldpayAuthoriseHandler.authorise(request);
        if (response.getBaseResponse().isPresent() && response.getBaseResponse().get().isSoftDecline()) {
            return worldpayAuthoriseHandler.authorise(request, true);
        }
        return response;
```

oswaldquek: I've now got it as 
```
        if (response.getBaseResponse().isPresent() && response.getBaseResponse().get().isSoftDecline()) {
            return worldpayAuthoriseHandler.authorise(request, true);
        }
```
which is easy to read. I'm not really sure about using `.ifPresent` for the sake of using it; sometimes I find it makes code more clunky to read.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I've put this all in `WorldpayOrderStatusResponse.isSoftDecline` now.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I'll leave this to another PR as I think there's enough going on this PR at the moment!

oswaldquek:
I still want to add some testing around soft declines as per the following spec from Worldpay:
Screenshot 2020-12-18 at 10 03 34

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -87,6 +89,22 @@ public void set3dsIssuerUrl(String issuerUrl) {
     public void setChallengeAcsUrl(String challengeAcsUrl) {
         this.challengeAcsUrl = challengeAcsUrl != null ? challengeAcsUrl.trim() : null;
     }
+    
+    public boolean isSoftDecline() {
```
alexbishop1, line 93: I’d not thought about putting the logic to determine if it’s a soft decline in `WorldpayOrderStatusResponse`. Good idea. (While I still think this class and its friends conflate being a representation of the actual response and our interpretation of it, this is what we have for now, so putting this new logic in here makes sense).

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -87,6 +89,22 @@ public void set3dsIssuerUrl(String issuerUrl) {
...
+        return Optional.ofNullable(lastEvent)
+                .map(e -> e.equals("REFUSED"))
+                .map(e -> e && exemptionResponseResult != null)
+                .map(e -> e && SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS.contains(exemptionResponseResult))
+                .orElse(false);
```
alexbishop1: It’s a little confusing that `e` is repeated but means a different thing each time.

Could do this:

```java
return Optional.ofNullable(lastEvent)
        .filter("REFUSED"::equals)
        .flatMap(lastEvent -> Optional.ofNullable(exemptionResponseResult).map(SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS::contains))
        .orElse(false);
```

Or maybe just this:

```java
return Optional.ofNullable(lastEvent).map("REFUSED"::equals).orElse(false)
        && Optional.ofNullable(exemptionResponseResult).map(SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS::contains).orElse(false);
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -96,8 +114,8 @@ public String getIssuerUrl() {
         return issuerUrl;
     }
 
-    public String getLastEvent() {
-        return lastEvent;
+    public Optional<String> getLastEvent() {
+        return Optional.ofNullable(lastEvent).map(r -> r.isBlank() ? null : lastEvent);
```
alexbishop1, line 115: I think I said in the last review that I don’t think we need to worry about `lastEvent` being a blank string.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -136,7 +151,34 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
 
     @Override
     public GatewayResponse<WorldpayOrderStatusResponse> authorise(CardAuthorisationGatewayRequest request) {
-        return worldpayAuthoriseHandler.authorise(request);
+
+        GatewayResponse<WorldpayOrderStatusResponse> response = worldpayAuthoriseHandler.authorise(request);
+
+        if (response.getBaseResponse().isPresent() && response.getBaseResponse().get().isSoftDecline()) {
```
alexbishop1: If you really want to avoid `ifPresent(…)` (which I can see would be a bit of a pain because you want to return from the method but you’ll be stuck in the lambda), you can do this:

```java
if (response.getBaseResponse().map(WorldpayOrderStatusResponse::isSoftDecline).orElse(false)) {
  // …
}
```

Or possibly that but with a boolean local variable:

```java
boolean isSoftDecline = response.getBaseResponse().map(WorldpayOrderStatusResponse::isSoftDecline).orElse(false);

if (isSoftDecline) {
    // …
}
```

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponseTest.java
```diff
@@ -0,0 +1,30 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class WorldpayOrderStatusResponseTest {
+
+    @ParameterizedTest
+    @ValueSource(strings = { "OUT_OF_SCOPE", "REJECTED", })
```
alexbishop1: Stray comma here (not that it likely makes a difference).

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -76,13 +100,61 @@
...
+        when(worldpayAuthoriseHandler.authorise(cardAuthRequest)).thenReturn(firstResponse);
+        when(worldpayAuthoriseHandler.authorise(cardAuthRequest, true)).thenReturn(secondResponse);
+
+        GatewayResponse<WorldpayOrderStatusResponse> response = worldpayPaymentProvider.authorise(cardAuthRequest);
+        
+        assertTrue(response.getBaseResponse().isPresent());
+        assertEquals(secondResponse.getBaseResponse().get(), response.getBaseResponse().get());
+        
+        verify(worldpayAuthoriseHandler).authorise(cardAuthRequest);
+        verify(worldpayAuthoriseHandler).authorise(cardAuthRequest, true);
```
alexbishop1: Shouldn’t need to verify if we’re stubbing.

alexbishop1 (changes requested):
Looks good. Just one thing about a test and a couple of other suggestions.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
oswaldquek: We do actually. Some contract tests fail (shouldBeAbleToSendAuthorisationRequestForMerchantUsing3dsWithoutAddress, submit3DS2FlexAuthRequest_requiresChallenge, shouldBeAbleToSendAuthorisationRequestForMerchantUsing3dsWithoutAddress) because it appears there isn't a `lastEvent`. However as there's a new `WorldpayOrderStatusResponse.isSoftDecline` method I should be able to revert this.

alexbishop1: To be clear, I was querying that we try to handle a blank `lastEvent` (e.g. `""` or `" "`) as opposed to a null `lastEvent` because there’s no such element in the XML response.

oswaldquek: Oh sorry, I misread. However I think wanting to return an Optional.empty if something is blank is a reasonable line of thinking and nothing too controversial.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I've gone with the former here.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
alexbishop1: It’s not controversial if a blank `lastEvent` is actually a possibility but it doesn’t sound like it can be so I think it just adds a little extra complexity. Furthermore, if we have code to handle a blank `lastEvent` then it could make our future selves think that’s a possibility, which clouds our understanding of what Worldpay may respond with.

oswaldquek: > it could make our future selves think that’s a possibility, which clouds our understanding of what Worldpay may respond with.

Very valid point. But it's possible I might have seen that `isNotBlank` is used copiously in the `toString` method and took that as a hint. Anyway, if there's nothing else, can you approve the PR?

alexbishop1: Yeah, I’m not keen on how we use `isNotBlank(…)` in the `toString()`. I’m not going to hold approval for this PR on this.

alexbishop1 (approved):

PR #2765, opened 2020-12-21
PP-5641: Authorising with exemption flag results in decline

oswaldquek:
It appears no code changes needed to be made for this scenario.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
@@ -52,6 +52,7 @@ import static org.mockito.ArgumentMatchers.anyList;
 import static org.mockito.ArgumentMatchers.anyMap;
 import static org.mockito.Mockito.eq;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -68,6 +69,8 @@ import static uk.gov.pay.connector.model.domain.AuthCardDetailsFixture.anAuthCar
 import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_3DS_RESPONSE;
 import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE;
 import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_AUTHORISED_INQUIRY_RESPONSE;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_EXEMPTION_REQUEST_DECLINE_RESPONSE;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_EXEMPTION_REQUEST_HONOURED_RESPONSE;
 import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_EXEMPTION_REQUEST_SOFT_DECLINE_RESPONSE;
 import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_VALID_3DS_FLEX_RESPONSE_AUTH_WORLDPAY_REQUEST;
 import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_VALID_3DS_RESPONSE_AUTH_WORLDPAY_REQUEST;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -112,6 +115,42 @@ void setup() {
         root.setLevel(Level.INFO);
         root.addAppender(mockAppender);
     }
+    
+    @Test
+    void should_not_retry_without_exemption_flag_when_authorising_with_exemption_flag_results_in_declined() throws Exception {
```
alexbishop1: I think this could be named a bit better: we’re specifically handling the case where the exemption is honoured but the payment is still refused

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -121,6 +123,28 @@ void setup() {
         root.addAppender(mockAppender);
     }
     
+    @Test
+    void authorise_with_exemption_flag_results_in_decline() throws Exception {
```
alexbishop1: Again, this is specifically that the exemption was honoured but it was still refused (I don’t think we should use “decline” here: it’s not a term either we or Worldpay use).

alexbishop1 (changes requested):
Looks good but I think some of the test names could be a little more specific.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -112,6 +115,42 @@ void setup() {
...
+        when(worldpayAuthoriseHandler.authorise(cardAuthRequest))
+                .thenReturn(getGatewayResponse(WORLDPAY_EXEMPTION_REQUEST_DECLINE_RESPONSE));
+
+        worldpayPaymentProvider.authorise(cardAuthRequest);
+
+        verify(worldpayAuthoriseHandler, never()).authorise(cardAuthRequest, true);
+    }
+    
+    @Test
+    void should_not_retry_without_exemption_flag_when_authorising_with_exemption_flag_results_in_authorised() throws Exception {
```
alexbishop1, line 138: I don’t this name should mention retrying, since we never going to retry if it was authorised

oswaldquek: Hmm why not? The principle thing I'm testing here is 
```
verify(worldpayAuthoriseHandler, never()).authorise(cardAuthRequest, true);
```
Unless you think we don't need this test at all? (I think it's useful as a guarantee we don't retry.)

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
oswaldquek: I used "decline" as its how it's described in the acceptance criteria. It's also definitely used as a concept in connector.
Does `authorise_with_exemption_flag_results_in_worldpay_honouring_the_request_but_refusing_to_authorise_the_order` work?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: Let me know if this works: `should_not_retry_authorising_without_exemption_flag_when_authorising_with_exemption_flag_results_in_worldpay_honouring_the_request_but_refusing_to_authorise_the_order`

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
alexbishop1: Oh, I didn’t realise that. I think this test is probably useful because it’s checking what happens when we get an honoured exemption and it’s authorised.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
alexbishop1: Yeah, that works. Maybe we could make it a bit shorter. Something like `should_not_retry_without_exemption_when_authorising_with_exemption_results_exemption_honoured_but_authorisation_refused`?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
alexbishop1: We do use “decline” but within this little bit of connector we’ve just got `REFUSED` (Worldpay’s term) and `[AUTHORISATION] REJECTED` (our term). So I think something like `authorise_with_exemption_when_exemption_honoured_but_authorisation_refused_results_in_rejected` or something.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: Sounds great, done!

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
oswaldquek: Done.

alexbishop1 (approved):

PR #2774, opened 2020-12-29
PP-7596 Rename `toggle_3ds` to `requires3ds`

kbottla:
## WHAT YOU DID
- We currently return `requires_3ds` (from database) as two different fields in API.
    Most of the endpoints returns field as `toggle_3ds` and `/v1/frontend/accounts/{accountId}` returns field as `requires3ds`.

     Endpoints that return `toggle_3ds`

     `/v1/api/accounts/{accountId}`
     `/v1/api/accounts`
     `/v1/api/accounts/external-id/{externalId}`
     `/v1/frontend/accounts`

- Renamed `toggle_3ds` to `requires3ds` for consistency and the field is currently used only used by toolbox (to be updated to use `requires3ds`).

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -231,7 +231,7 @@ Content-Type: application/json
     "block_prepaid_cards": false,
     "allow_zero_amount": false,
     "email_collection_mode": "MANDATORY",
-    "toggle_3ds": false,
+    "requires3ds": false,
     "allow_moto": false,
     "moto_mask_card_number_input": false,
     "moto_mask_card_security_code_input": false,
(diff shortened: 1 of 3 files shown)


sfount (dismissed):
To clarify my understanding I think it's only the `/v1/api/accounts/external_id` route that returns `toggle_3ds`? Unless it's both that and the `/v1/frontend/accounts` path. 

Change looks good to me either way 👍

kbottla:
@sfount good spot! That was supposed to be other way around as all except /frontend/../accountId returns requires3ds. Updated PR desc and commit msg

sfount (approved):

PR #2776, opened 2020-12-30
PP-7590 ChargeEntity: use Instant not ZonedDateTime for createdDate

alexbishop1:
Update the `ChargeEntity` class so that the `createdDate` field is an `Instant` rather than a `ZonedDateTime`.

`Instant` is a better fit than `ZonedDateTime` because it represents an exact moment in time but without the complexity of time zones. When instantiating a `ZonedDateTime` to assign to a `createdDate` field, we usually used the UTC time zone (e.g. `ZonedDateTime.now(ZoneOffset.UTC))`) but sometimes relied on the default time zone (UTC for the deployed application servers, possibly not elsewhere). Since an `Instant` does not have a time zone, we can avoid this brain tax while losing nothing of value, since we never relied on being able to have different `createdDate` fields using different time zones (we basically always wanted UTC, even if we did not explicitly ask for it).

Use `InstantToUtcTimestampWithoutTimeZoneConverter` instead of `UTCDateTimeConverter` to persist the `createdDate` to the `created_date` `TIMESTAMP WITHOUT TIME ZONE` column in the `charges` table of the database (see commit `f3d40f91e6b309ea6f14e10e6229f8086c050dcd` in pay-java-commons for more information). This will actually store the `created_date` as a date and time in the UTC time zone, rather than the default time zone. Since the default time zone of the deployed application servers is UTC, this won’t make a difference. This may cause a behaviour change when running with a default time zone that is not UTC (including changing how existing dates and times in the database are interpreted) but since this only happens when running locally for development and testing purposes, this should not be a real problem. Indeed, from now on, UTC will always be used, regardless of the default time zone, making the behaviour less environment-specific and more predictable.

In order to minimise the number of changes, we convert the `Instant` to a `ZonedDateTime` (using UTC), or convert a `ZonedDateTime` to an `Instant`, in some parts of the code. This is intended to be temporary, with `Instant` eventually being used everywhere to represent charge creation dates.

There are no changes to how the `createdDate` becomes the `created_date` in JSON responses (we convert to a `ZonedDateTime` in UTC and use the existing serialisation code).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
+++ b/src/main/java/uk/gov/pay/connector/charge/dao/ChargeDao.java
@@ -16,6 +16,7 @@ import javax.persistence.criteria.CriteriaQuery;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 import java.time.Duration;
+import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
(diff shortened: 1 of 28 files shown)


kbottla (approved):
LGTM!

PR #2777, opened 2020-12-30
PP-5642: Set exemption_3ds for a charge

oswaldquek:
Set exemption_3ds for a charge to EXEMPTION_NOT_REQUESTED if:

- The auth request is for Worldpay
- The charge's status resulting from the auth request is updated to either
  AUTHORISATION SUCCESS, AUTHORISATION REJECTED, AUTHORISATION CANCELLED or
AUTHORISATION ERROR

Also added an integration test which is useful to verify all the persistence code has been wired correctly.

ChargeService looks quite bloated but any refactoring will be saved for another PR.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -453,7 +453,8 @@ public class ChargeService {
                 .withLink("self", GET, selfUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeId))
                 .withLink("refunds", GET, refundsUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeEntity.getExternalId()))
                 .withWalletType(chargeEntity.getWalletType())
-                .withMoto(chargeEntity.isMoto());
+                .withMoto(chargeEntity.isMoto())
+                .withExemption3ds(chargeEntity.getExemption3ds());
 
         chargeEntity.getFeeAmount().ifPresent(builderOfResponse::withFee);
         chargeEntity.getExternalMetadata().ifPresent(builderOfResponse::withExternalMetadata);
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -60,6 +60,7 @@ import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.northamericaregion.NorthAmericaRegion;
 import uk.gov.pay.connector.northamericaregion.NorthAmericanRegionMapper;
+import uk.gov.pay.connector.paymentprocessor.model.Exemption3ds;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 import uk.gov.pay.connector.queue.statetransition.StateTransitionService;
 import uk.gov.pay.connector.refund.model.domain.Refund;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -535,13 +541,15 @@ public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
...
-                                                      String emailAddress) {
+                                                      String emailAddress, 
+                                                      Exemption3ds exemption3ds) {
         return chargeDao.findByExternalId(chargeExternalId).map(charge -> {
             setTransactionId(charge, transactionId);
             Optional.ofNullable(sessionIdentifier).map(ProviderSessionIdentifier::toString).ifPresent(charge::setProviderSessionId);
             Optional.ofNullable(auth3dsRequiredDetails).ifPresent(charge::set3dsRequiredDetails);
             Optional.ofNullable(walletType).ifPresent(charge::setWalletType);
             Optional.ofNullable(emailAddress).ifPresent(charge::setEmail);
+            Optional.ofNullable(exemption3ds).ifPresent(e -> charge.setExemption3ds(exemption3ds));
```
alexbishop1, line 552: `Optional.ofNullable(exemption3ds).ifPresent(charge::setExemption3ds);`

src/main/java/uk/gov/pay/connector/charge/service/UpdateChargePostAuthorisation.java
```diff
@@ -0,0 +1,135 @@
...
+
+public class UpdateChargePostAuthorisation {
+
+    private String chargeExternalId;
+    private ChargeStatus status;
+    private String transactionId;
+    private Auth3dsRequiredEntity auth3dsRequiredDetails;
+    private ProviderSessionIdentifier sessionIdentifier;
+    private AuthCardDetails authCardDetails;
+    private Exemption3ds exemption3ds;
```
alexbishop1, line 19: It’s a shame these can’t be `final`. Perhaps we could consider a (private) constructor?

src/main/java/uk/gov/pay/connector/paymentprocessor/model/Exemption3ds.java
```diff
@@ -0,0 +1,32 @@
...
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ERROR;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
+
+public enum Exemption3ds {
+    
+    EXEMPTION_NOT_REQUESTED;
+    
+    private static final Set<ChargeStatus> EXEMPTION_NOT_REQUESTED_STATUSES = 
+            Set.of(AUTHORISATION_SUCCESS, AUTHORISATION_REJECTED, AUTHORISATION_CANCELLED, AUTHORISATION_ERROR);
```
alexbishop1, line 19: Why do we need this? If it’s to make sure we only store the exemption in certain circumstances (and not, say, when there’s a soft decline), I think it would probably be best to check the status at the point we’re considering storing the exemption response, rather than changing what the exemption response.

It also seems unlikely we’re going to get the behaviour we want when we get an `AUTHORISATION_ERROR` because I suspect the response wouldn’t tell us if we’d requested an exemption.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -115,6 +120,25 @@ public AuthorisationResponse doAuthorise(String chargeId, AuthCardDetails authCa
...
+            AuthCardDetails authCardDetails,
+            GatewayResponse<BaseAuthoriseResponse> operationResponse) {
+        return anUpdateChargePostAuthorisation()
+                .withChargeExternalId(chargeExternalId)
+                .withStatus(newStatus)
+                .withTransactionId(transactionId)
+                .withAuth3dsRequiredDetails(auth3dsRequiredEntity)
+                .withSessionIdentifier(sessionIdentifier)
+                .withAuthCardDetails(authCardDetails)
+                .withExemption3ds(operationResponse.getBaseResponse().map(r -> calculateExemption3ds(r, newStatus)).orElse(null))
```
alexbishop1, line 138: Can we use a slightly more descriptive variable name than `r`?

alexbishop1 (changes requested):
I think I gave you bad advice earlier. In the case of `AUTHORISATION ERROR` (and possibly other statuses), we’re probably not going to have enough in the response to determine if we requested an exemption. The more I think about it, the more I feel that we probably don’t want to rely on Worldpay telling us if we requested an exemption. So I think we need to keep track of this ourselves. Not quite sure of the best way to do this. Might be feasible to set it on the `ChargeEntity` but that might get a bit confusing if we’re not persisting it at that point.

PR #2781, opened 2020-12-31
PP-7590 Use Instant for created date in charge test fixtures

alexbishop1:
In test charge fixtures, use `Instant` rather than `ZonedDateTime` for the created date. This brings it into line with `ChargeEntity`.

There are some small behaviour differences here. If a test created a `ZonedDateTime` using `ZonedDateTime.now()` with a default time zone that uses daylight saving time (such as Europe/London) and then subtracted a certain number of days using `minusDays(…)`, the number of hours subtracted might not be an exact multiple of 24 if the period subtracted included the time when daylight saving time started or finished. With the equivalent subtraction on an `Instant`, the number of hours  is always an exact multiple of 24. The previous behaviour was almost certainly unintended so this is OK.


kbottla (approved):
LGTM!

PR #2782, opened 2020-12-31
PP-5642: Set exemption_3ds for a charge to EXEMPTION_NOT_REQUESTED

oswaldquek:
Set exemption_3ds for a charge to EXEMPTION_NOT_REQUESTED if exemption engine
for the gateway account is not enabled. I.e., an authorisation request is made
without an exemption.

Also added an integration test which is useful to verify all the persistence
code has been wired correctly.

A ChargeDao is injected into WorldpayPaymentProvider and is used to update the
ChargeEntity. Unfortunately using a ChargeService caused a circular dependency
error because ChargesApiResource -> ChargeService -> PaymentProviders ->
WorldpayPaymentProvider -> ChargeService.

## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-12-31T16:14:05Z",
+  "generated_at": "2021-01-04T16:41:15Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -184,7 +184,7 @@
         "hashed_secret": "c6b43e7d8e68a66c283fd8760118b89592f6498d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 439,
+        "line_number": 437,
         "type": "Base64 High Entropy String"
       }
     ],
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,8 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-12-17T11:38:55Z",
-  "generated_at": "2020-12-16T14:18:37Z",
+  "generated_at": "2020-12-31T13:18:15Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -185,7 +184,7 @@
         "hashed_secret": "c6b43e7d8e68a66c283fd8760118b89592f6498d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 310,
+        "line_number": 373,
         "type": "Base64 High Entropy String"
       }
     ],
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-12-31T13:18:15Z",
+  "generated_at": "2020-12-31T16:14:05Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
@@ -184,7 +184,7 @@
         "hashed_secret": "c6b43e7d8e68a66c283fd8760118b89592f6498d",
         "is_secret": false,
         "is_verified": false,
-        "line_number": 373,
+        "line_number": 439,
         "type": "Base64 High Entropy String"
       }
     ],
--- a/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
@@ -15,6 +15,7 @@ import uk.gov.pay.connector.charge.model.builder.AbstractChargeResponseBuilder;
 import uk.gov.pay.connector.charge.model.domain.PersistedCard;
 import uk.gov.pay.connector.charge.model.telephone.PaymentOutcome;
 import uk.gov.pay.connector.common.model.api.ExternalTransactionState;
+import uk.gov.pay.connector.paymentprocessor.model.Exemption3ds;
 import uk.gov.pay.connector.util.DateTimeUtils;
 import uk.gov.pay.connector.wallets.WalletType;
 
@@ -131,6 +132,9 @@ public class ChargeResponse {
     @JsonProperty("moto")
     private boolean moto;
 
+    @JsonProperty("exemption_3ds")
+    private Exemption3ds exemption3ds;
+
     ChargeResponse(AbstractChargeResponseBuilder<?, ? extends ChargeResponse> builder) {
         this.dataLinks = builder.getLinks();
         this.chargeId = builder.getChargeId();
@@ -163,6 +167,7 @@ public class ChargeResponse {
         this.walletType = builder.getWalletType();
         this.externalMetadata = builder.getExternalMetadata();
         this.moto = builder.isMoto();
+        this.exemption3ds = builder.getExemption3ds();
     }
 
     public List<Map<String, Object>> getDataLinks() {
--- a/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
@@ -15,7 +15,6 @@ import uk.gov.pay.connector.charge.model.builder.AbstractChargeResponseBuilder;
 import uk.gov.pay.connector.charge.model.domain.PersistedCard;
 import uk.gov.pay.connector.charge.model.telephone.PaymentOutcome;
 import uk.gov.pay.connector.common.model.api.ExternalTransactionState;
-import uk.gov.pay.connector.paymentprocessor.model.Exemption3ds;
 import uk.gov.pay.connector.util.DateTimeUtils;
 import uk.gov.pay.connector.wallets.WalletType;
 
@@ -132,9 +131,6 @@ public class ChargeResponse {
     @JsonProperty("moto")
     private boolean moto;
 
-    @JsonProperty("exemption_3ds")
-    private Exemption3ds exemption3ds;
-
     ChargeResponse(AbstractChargeResponseBuilder<?, ? extends ChargeResponse> builder) {
         this.dataLinks = builder.getLinks();
         this.chargeId = builder.getChargeId();
@@ -167,7 +163,6 @@ public class ChargeResponse {
         this.walletType = builder.getWalletType();
         this.externalMetadata = builder.getExternalMetadata();
         this.moto = builder.isMoto();
-        this.exemption3ds = builder.getExemption3ds();
     }
 
     public List<Map<String, Object>> getDataLinks() {
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -166,6 +166,7 @@ public class ChargeEntity extends AbstractVersionedEntity implements Nettable {
     private boolean moto;
 
     @Column(name = "exemption_3ds")
+    @Enumerated(EnumType.STRING)
     private Exemption3ds exemption3ds;
 
     public ChargeEntity() {
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -23,6 +23,7 @@ import uk.gov.pay.connector.events.model.Event;
 import uk.gov.pay.connector.events.model.UnspecifiedEvent;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.paymentprocessor.model.Exemption3ds;
 import uk.gov.pay.connector.util.RandomIdGenerator;
 import uk.gov.pay.connector.wallets.WalletType;
 
@@ -164,6 +165,9 @@ public class ChargeEntity extends AbstractVersionedEntity implements Nettable {
     @Column(name = "moto")
     private boolean moto;
 
+    @Column(name = "exemption_3ds")
+    private Exemption3ds exemption3ds;
+
     public ChargeEntity() {
         //for jpa
     }
@@ -270,6 +274,14 @@ public class ChargeEntity extends AbstractVersionedEntity implements Nettable {
 
     public boolean isMoto() {
         return moto;
+    }
+
+    public void setExemption3ds(Exemption3ds exemption3ds) {
+        this.exemption3ds = exemption3ds;
+    }
+
+    public Exemption3ds getExemption3ds() {
+        return exemption3ds;
     }
 
     public void setExternalId(String externalId) {
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -154,16 +154,19 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
     public GatewayResponse<WorldpayOrderStatusResponse> authorise(CardAuthorisationGatewayRequest request) {
 
         boolean exemptionEngineEnabled = isExemptionEngineEnabled(request);
+        GatewayResponse<WorldpayOrderStatusResponse> response;
         
         if (!exemptionEngineEnabled) {
             ChargeEntity charge = request.getCharge();
             charge.setExemption3ds(EXEMPTION_NOT_REQUESTED);
             chargeDao.merge(charge);
             LOGGER.info("Updated exemption_3ds of charge to {} - charge_external_id={}", EXEMPTION_NOT_REQUESTED, charge.getExternalId());
+            
+            response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
+        } else {
+            response = worldpayAuthoriseHandler.authoriseWithExemption(request);
         }
-
-        GatewayResponse<WorldpayOrderStatusResponse> response = worldpayAuthoriseHandler.authorise(request, exemptionEngineEnabled);
-
+        
         if (response.getBaseResponse().map(WorldpayOrderStatusResponse::isSoftDecline).orElse(false)) {
             
             var authorisationRequestSummary = generateAuthorisationRequestSummary(request.getCharge(), request.getAuthCardDetails());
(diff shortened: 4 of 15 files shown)


.secrets.baseline
```diff
@@ -3,8 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-12-17T11:38:55Z",
-  "generated_at": "2020-12-16T14:18:37Z",
```
alexbishop1, line 7: Curious we ended up with two of these!

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -132,6 +133,9 @@
     @JsonProperty("moto")
     private boolean moto;
 
+    @JsonProperty("exemption_3ds")
+    private Exemption3ds exemption3ds;
```
alexbishop1: Do we want this to appear in the response at this stage?

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -164,6 +165,9 @@
     @Column(name = "moto")
     private boolean moto;
 
+    @Column(name = "exemption_3ds")
```
alexbishop1, line 168: I think we want `@Enumerated(EnumType.STRING)` here or the value that’s persisted will be the ordinal of the enum.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -145,6 +150,12 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
 
     @Override
     public GatewayResponse<WorldpayOrderStatusResponse> authorise(CardAuthorisationGatewayRequest request) {
+        
+        if (!WorldpayOrderBuilder.isExemptionEngineEnabled(request)) {
```
alexbishop1: There’s a risk here that whether or not the Exemption Engine is enabled may change between when we record in the database and when we build the request. Can we avoid this? Perhaps we could calculate whether or not the Exemption Engine is enabled once and then pass it to `WorldpayAuthoriseHandler` and then down to `WorldpayOrderRequestBuilder`, which would respect that rather than doing its own check to see if the Exemption Engine is enabled.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -145,6 +150,12 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
+            ChargeEntity charge = request.getCharge();
+            charge.setExemption3ds(EXEMPTION_NOT_REQUESTED);
```
alexbishop1, line 161: We might want a test in `ChargeDaoIT` to check this roundtrips (similar to `shouldCreateNewChargeWithParityCheckStatus()`.

alexbishop1 (changes requested):
Looks along the right lines.

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
oswaldquek: Technically this doesn't fulfil any acceptance criteria (even if I think we know we're going to need it in the future). This is here because it's useful to verify the `WorldpayCardResourceIT. should_set_exemption_not_requested_when_request_made_without_an_exemption` test which does pertain to the acceptance criteria. But I see you're in suggesting we create a test in the ChargeDaoIT which would obviate the need for this integration test so I could go with that.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -163,58 +162,24 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
 
         ArgumentCaptor<GatewayOrder> gatewayOrderArgumentCaptor = ArgumentCaptor.forClass(GatewayOrder.class);
         verify(authoriseClient).postRequestFor(eq(WORLDPAY_URL), eq(gatewayAccountEntity), gatewayOrderArgumentCaptor.capture(), anyMap());
         assertXMLEqual(load(WORLDPAY_VALID_AUTHORISE_WORLDPAY_REQUEST_INCLUDING_3DS_WITHOUT_IP_ADDRESS),
                 gatewayOrderArgumentCaptor.getValue().getPayload());
     }
-
-    @Test
-    void should_not_include_exemption_element_if_account_has_no_worldpay_3ds_flex_credentials() throws Exception {
-
```
oswaldquek, line 176: This was moved to WorldpayPaymentProviderTest. should_not_include_exemption_if_account_has_no_worldpay_3ds_flex_credentials

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -163,58 +162,24 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
-        gatewayAccountEntity.setIntegrationVersion3ds(1);
-        gatewayAccountEntity.setWorldpay3dsFlexCredentialsEntity(null);
-        chargeEntityFixture.withGatewayAccountEntity(gatewayAccountEntity);
-        worldpayAuthoriseHandler.authorise(new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), getValidTestCard()));
-
-        verifyNoExemptionRequestInAuthorisationRequest();
-    }
-
-    @Test
-    void should_not_include_exemption_element_if_account_has_exemption_engine_set_to_false() throws Exception {
```
oswaldquek, line 191: This was moved to WorldpayPaymentProviderTest. should_not_include_exemption_if_account_has_exemption_engine_set_to_false

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -229,41 +194,21 @@ void should_include_exemption_element_if_account_has_exemption_engine_set_to_tru
         assertThat(xPath.evaluate("/paymentService/submit/order/exemption/@type", document), is("OP"));
         assertThat(xPath.evaluate("/paymentService/submit/order/exemption/@placement", document), is("AUTHORISATION"));
     }
-    
-    @Test
-    void should_not_include_exemption_element_if_account_has_exemption_engine_set_to_true_but_the_authorisation_is_made_in_response_to_a_soft_decline() throws Exception {
```
oswaldquek, line 234: This is equivalently tested in WorldpayPaymentProviderTest. should_retry_without_exemption_flag_when_authorising_with_exemption_flag_results_in_soft_decline so could be deleted

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -163,58 +162,24 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
-        gatewayAccountEntity.setWorldpay3dsFlexCredentialsEntity(aWorldpay3dsFlexCredentialsEntity().build());
-        chargeEntityFixture.withGatewayAccountEntity(gatewayAccountEntity);
-        worldpayAuthoriseHandler.authorise(new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), getValidTestCard()));
-
-        verifyNoExemptionRequestInAuthorisationRequest();
-    }
-
+    
     @Test
-    void should_include_exemption_element_if_account_has_exemption_engine_set_to_true() throws Exception {
```
oswaldquek, line 207: This was moved to WorldpayPaymentProviderTest .should_include_exemption_if_account_has_exemption_engine_set_to_true.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -229,41 +194,21 @@ void should_include_exemption_element_if_account_has_exemption_engine_set_to_tru
...
-        gatewayAccountEntity.setWorldpay3dsFlexCredentialsEntity(aWorldpay3dsFlexCredentialsEntity().withExemptionEngine(true).build());
-        chargeEntityFixture.withGatewayAccountEntity(gatewayAccountEntity);
-        var cardAuthorisationGatewayRequest = new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), getValidTestCard());
-        worldpayAuthoriseHandler.authorise(cardAuthorisationGatewayRequest, true);
-
-        verifyNoExemptionRequestInAuthorisationRequest();
-    }
 
     @Test
-    void should_not_include_exemption_element_if_account_has_exemption_engine_set_to_true_but_3ds_is_not_enabled() throws Exception {
```
oswaldquek, line 251: Moved to WorldpayPaymentProviderTest .should_not_include_exemption_if_account_has_exemption_engine_set_to_true_but_3ds_is_not_enabled

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: This sounds sensible. However what seemed like a simple idea resulted in a larger commit than expected: https://github.com/alphagov/pay-connector/pull/2782/commits/b5284e0dca74f7ebf70997f5446f272715cd2f87. Happy reviewing!

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
oswaldquek: Fixed

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
oswaldquek: Deleted this as due to the new `ChargeDaoIT.should_update_charge_with_exemption_3ds` test.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: New `ChargeDaoIT.should_update_charge_with_exemption_3ds` test added.

oswaldquek: I think I've never written a DaoIT test during my time in Pay!

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -146,7 +153,16 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
 
-        GatewayResponse<WorldpayOrderStatusResponse> response = worldpayAuthoriseHandler.authorise(request);
+        boolean exemptionEngineEnabled = isExemptionEngineEnabled(request);
+        
+        if (!exemptionEngineEnabled) {
+            ChargeEntity charge = request.getCharge();
+            charge.setExemption3ds(EXEMPTION_NOT_REQUESTED);
+            chargeDao.merge(charge);
+            LOGGER.info("Updated exemption_3ds of charge to {} - charge_external_id={}", EXEMPTION_NOT_REQUESTED, charge.getExternalId());
+        }
```
alexbishop1: I wonder if ultimately we’ll end up doing this after the request has been made because that’s when we’ll be updating the flag if we did request an exemption (because we’ll only know if it’s honoured or not when we have the response). But it’s fine here for now.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -163,58 +162,24 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
-        chargeEntityFixture.withGatewayAccountEntity(gatewayAccountEntity);
-        worldpayAuthoriseHandler.authorise(new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), getValidTestCard()));
-
-        verifyNoExemptionRequestInAuthorisationRequest();
-    }
-
+    
     @Test
-    void should_include_exemption_element_if_account_has_exemption_engine_set_to_true() throws Exception {
+    void should_include_exemption_element() throws Exception {
```
alexbishop1: I think the `should_include_exemption_element_if_account_has_exemption_engine_set_to_true` name was clearer.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -229,41 +194,21 @@ void should_include_exemption_element_if_account_has_exemption_engine_set_to_tru
...
-        chargeEntityFixture.withGatewayAccountEntity(gatewayAccountEntity);
-        var cardAuthorisationGatewayRequest = new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), getValidTestCard());
-        worldpayAuthoriseHandler.authorise(cardAuthorisationGatewayRequest, true);
-
-        verifyNoExemptionRequestInAuthorisationRequest();
-    }
 
     @Test
-    void should_not_include_exemption_element_if_account_has_exemption_engine_set_to_true_but_3ds_is_not_enabled() throws Exception {
+    void should_not_include_exemption_element() throws Exception {
```
alexbishop1, line 195: I’m a bit confused about what’s happening here. We have 3DS and the Exemption Engine enabled but we’re not sending an exemption?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -118,6 +128,70 @@ void setup() {
...
+        var cardAuthRequest = new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), anAuthCardDetails().build());
+        
+        when(worldpayAuthoriseHandler.authorise(cardAuthRequest, false))
+                .thenReturn(getGatewayResponse(WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE));
+
+        worldpayPaymentProvider.authorise(cardAuthRequest);
+    }
+
+    @Test
+    void should_set_exemption_not_requested_when_request_made_without_an_exemption() throws Exception {
```
alexbishop1: I wonder if perhaps this could be combined into the previous tests, so that if we expect an exemption not to be requested we both check that request and that we record that fact.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -118,6 +128,70 @@ void setup() {
...
+        var cardAuthRequest = new CardAuthorisationGatewayRequest(chargeEntity, anAuthCardDetails().build());
+        
+        when(worldpayAuthoriseHandler.authorise(cardAuthRequest, false))
+                .thenReturn(getGatewayResponse(WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE));
+        
+        worldpayPaymentProvider.authorise(cardAuthRequest);
+        
+        ArgumentCaptor<ChargeEntity> chargeDaoArgumentCaptor = ArgumentCaptor.forClass(ChargeEntity.class);
+        verify(chargeDao).merge(chargeDaoArgumentCaptor.capture());
+        assertEquals(chargeDaoArgumentCaptor.getValue().getExemption3ds(), EXEMPTION_NOT_REQUESTED);
```
alexbishop1: Another way to do this might be use in-order verification to check the `setExemption3ds(…)` and `merge(…)` methods are called in that order (but this way is also fine).

alexbishop1 (commented):
This looks pretty good. Left some comments but I’m not sure any of them will actually require changes (hence neither approving nor requesting changes).

One other thing that might be worth considering is changing the `WorldpayAuthoriseHandler.authorise(…)` so that rather than being one method with a boolean parameter to indicate whether the Exemption Engine should be used or not, it’s two methods called `authoriseWithExemption(…)` and `authoriseWIthoutExemption(…)` (say). But that might be something to explore in a follow-up PR.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: This is fine for now. The next few scenarios (which I've already started working on) will force a move of this code block till after the request has been made.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
oswaldquek: Yes it's because of the `false` in 
```
worldpayAuthoriseHandler.authorise(new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), getValidTestCard()), false);
```
I can see this is a bit confusing; I like the suggestion of having the `authoriseWithExemption(…)` and `authoriseWithoutExemption(…)` so will update this.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: I think I'll leave this as is.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
oswaldquek: Done

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
oswaldquek: I've updated the signatures to be `authoriseWithExemption(…)` and `authoriseWithoutExemption(…)` so hopefully this is clearer now.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: Good shout; done.

alexbishop1 (approved):

PR #2783, opened 2020-12-31
PP-7590 Use Instant for authorised_date in API responses

alexbishop1:
In `ChargeResponse` (the Java model that represents the JSON for a charge sent in API responses), change the Java type of the `authorised_date` property from `ZonedDateTime` to `Instant`, which necessitates switching from `ApiResponseDateTimeSerializer` to `ApiResponseInstantSerializer`.

This will not change how the authorised_date for a charge appears as a JSON string in API responses.


oswaldquek (approved):
LGTM

PR #2788, opened 2021-01-06
PP-5642: Exemption honoured, rejected and out-of-scope scenarios

oswaldquek:
Set charge's exemption_3ds=EXEMPTION_HONOURED if exemption was honoured.
Set charge's exemption_3ds=EXEMPTION_REJECTED if exemption was rejected.
Set charge's exemption_3ds=EXEMPTION_OUT_OF_SCOPE if exemption was out of
scope.


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -153,22 +157,25 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
-            
+            exemption3ds = EXEMPTION_NOT_REQUESTED;
             response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
         } else {
             response = worldpayAuthoriseHandler.authoriseWithExemption(request);
         }
         
         if (response.getBaseResponse().map(WorldpayOrderStatusResponse::isSoftDecline).orElse(false)) {
             
+            if (response.getBaseResponse().get().getExemptionResponseResult().orElse("").equals("REJECTED")) {
```
alexbishop1: Might be possible to avoid doing `get()` on the `Optional` by changing the lines above:

```java
response.getBaseResponse().ifPresent(baseResponse -> {
    if (baseResponse.isSoftDecline()) {
       // etc. etc.
   }
});
```

Also, rather than `getExemptionResponseResult().orElse("").equals("REJECTED"))`, we can do `getExemptionResponseResult().map("REJECTED"::equals).orElse(false)`.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -180,12 +187,29 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
                     response,
                     request.getCharge().getChargeStatus(),
                     request.getCharge().getChargeStatus());
-            
+
             response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
+
+        } else if (responseIsHonoured(response)) {
+            exemption3ds = EXEMPTION_HONOURED;
```
alexbishop1: Rather than having `exemption3ds` set on several different lines, it might be feasible to do it once by replacing `responseIsHonoured(…)` with a method that calculates it all cases using a method signature like this:

```java
private Optional<Exemption3ds> determineExemption(boolean exemptionEngineEnabled, GatewayResponse<WorldpayOrderStatusResponse> response) {
    // etc. etc.
}
```

The `Optional` being because we won’t be able get anything if there was an error and there’s nothing meaningful in the response body.

Then we might be able to call it and update the charge in one shot:

```java
determineExemption(exemptionEngineEnabled, response).ifPresent(exemption3ds -> updateChargeWithExemption3ds(exemption3ds, request.getCharge()));
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -180,12 +187,29 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
         }
+
+        updateChargeWithExemption3ds(exemption3ds, request.getCharge());
+
         return response;
     }
 
+    private Boolean responseIsHonoured(GatewayResponse<WorldpayOrderStatusResponse> response) {
```
alexbishop1: I think this name needs “exemption” in it somewhere and it should probably return a primitive `boolean` rather than an autoboxed `Boolean`.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -189,7 +210,48 @@ void should_not_include_exemption_if_account_has_exemption_engine_set_to_true_bu
...
+        
+        gatewayAccountEntity.setRequires3ds(true);
+        gatewayAccountEntity.setIntegrationVersion3ds(1);
+        gatewayAccountEntity.setWorldpay3dsFlexCredentialsEntity(aWorldpay3dsFlexCredentialsEntity().withExemptionEngine(true).build());
+        chargeEntityFixture.withGatewayAccountEntity(gatewayAccountEntity);
+        ChargeEntity chargeEntity = chargeEntityFixture.build();
+        var cardAuthRequest = new CardAuthorisationGatewayRequest(chargeEntity, anAuthCardDetails().build());
+
+        var firstResponse = getGatewayResponse(WORLDPAY_EXEMPTION_REQUEST_SOFT_DECLINE_RESULT_REJECTED_RESPONSE);
+        var secondResponse = getGatewayResponse(WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE);
```
alexbishop1: These might possibly be clearer if they’re inlined to where they’re used.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -267,7 +331,7 @@ void should_retry_without_exemption_flag_when_authorising_with_exemption_flag_re
         assertEquals(secondResponse.getBaseResponse().get(), response.getBaseResponse().get());
         
         ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
-        verify(mockAppender, times(1)).doAppend(loggingEventArgumentCaptor.capture());
+        verify(mockAppender, times(2)).doAppend(loggingEventArgumentCaptor.capture());
```
alexbishop1, line 332: Why did this change?

alexbishop1 (changes requested):
Looks like this should work. I’ve made some suggestions, at least some of which might be feasible.

I’m also becoming a bit concerned that the `authorise(…)` method in `WorldpayPaymentProvider` is getting a bit big (and the class’s test class is now over 500 lines). It might be worth exploring if we can extract some of its functionality into a separate service (which would sit between `WorldpayPaymentProvider` and `WorldpayAuthoriseHandler`).

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I'm ambivalent to using 
```
response.getBaseResponse().ifPresent(baseResponse -> {
    if (baseResponse.isSoftDecline()) {
       // etc. etc.
   }
});
```
as it adds more lines to compared to
```
            if (response.getBaseResponse().get().getExemptionResponseResult().map("REJECTED"::equals).orElse(false)) {
                updateChargeWithExemption3ds(EXEMPTION_REJECTED, request.getCharge());
            } else if (response.getBaseResponse().get().getExemptionResponseResult().map("OUT_OF_SCOPE"::equals).orElse(false)) {
                updateChargeWithExemption3ds(EXEMPTION_OUT_OF_SCOPE, request.getCharge());
            }

```
and I don't think it adds to readability.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I would have liked to have a method that calculates all cases and call it once near the end of `WorldpayPaymentProvider.authorise` too. However I don't think this is feasible (without introducing new variables) as we need to calculate the `exemption_3ds` value straight away after a soft decline and before the retry as the retry assigns the result to the `GatewayResponse<WorldpayOrderStatusResponse> response`.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I've gotten rid of this method.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: Because in addition to the logs that are being tested in this test, `Updated exemption_3ds of charge to EXEMPTION_REJECTED` as well.

oswaldquek: Perhaps I could merge `should_set_exemption_rejected_when_request_made_with_an_exemption` into this test as well?

oswaldquek:
@alexbishop1 I've committed some [changes](https://github.com/alphagov/pay-connector/pull/2788/commits/9c2b4853acad84b00cb316daee8bd1d5997abeb2) that hopefully make the code cleaner. 

I don't think `WorldpayPaymentProvider.authorise` is that bad actually? I don't think it's hard to read and a large chunk of the lines is taken up by the logging:
```
           var authorisationRequestSummary = generateAuthorisationRequestSummary(request.getCharge(), request.getAuthCardDetails());

            authorisationLogger.logChargeAuthorisation(
                    LOGGER,
                    authorisationRequestSummary,
                    request.getCharge(),
                    authorisationService.extractTransactionId(request.getCharge().getExternalId(), response)
                            .orElse("missing transaction ID"),
                    response,
                    request.getCharge().getChargeStatus(),
                    request.getCharge().getChargeStatus());
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
alexbishop1: It’s really the `get()` I’m worried about. It could still be done with `if`/`else` by using `flatMap()`:

```java
if (response.getBaseResponse().flatMap(WorldpayOrderStatusResponse::getExemptionResponseResult).map("REJECTED"::equals).orElse(false)) {
    updateChargeWithExemption3ds(EXEMPTION_REJECTED, request.getCharge());
} else if (response.getBaseResponse().flatMap(WorldpayOrderStatusResponse::getExemptionResponseResult).map("OUT_OF_SCCOPE"::equals).orElse(false)) {
    updateChargeWithExemption3ds(EXEMPTION_OUT_OF_SCOPE, request.getCharge());
}
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -180,12 +181,51 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
                     request.getCharge().getChargeStatus(),
                     request.getCharge().getChargeStatus());
-            
+
             response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
+
+        } else {
+            response.getBaseResponse().ifPresent(worldpayOrderStatusResponse -> 
+                    determineExemption(exemptionEngineEnabled, worldpayOrderStatusResponse)
+                            .ifPresent(exemption3ds -> updateChargeWithExemption3ds(exemption3ds, request.getCharge())));
```
alexbishop1: This could be:

```java
response.getBaseResponse()
        .flatMap(worldpayOrderStatusResponse -> determineExemption(exemptionEngineEnabled, worldpayOrderStatusResponse))
        .ifPresent(exemption3ds -> updateChargeWithExemption3ds(exemption3ds, request.getCharge()));
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -180,12 +181,51 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
+            if (worldpayOrderStatusResponse.getExemptionResponseResult().map("REJECTED"::equals).orElse(false)) {
+                updateChargeWithExemption3ds(EXEMPTION_REJECTED, request.getCharge());
+            }
+        });
+
+        response.getBaseResponse().ifPresent(worldpayOrderStatusResponse -> {
+            if (worldpayOrderStatusResponse.getExemptionResponseResult().map("OUT_OF_SCOPE"::equals).orElse(false)) {
+                updateChargeWithExemption3ds(EXEMPTION_OUT_OF_SCOPE, request.getCharge());
+            }
+        });
```
alexbishop1: We could remove a little bit of duplication by doing something like this:

```java
response.getBaseResponse().flatMap(WorldpayOrderStatusResponse::getExemptionResponseResult).ifPresent(exemption3ds -> {
    if ("REJECTED".equals(exemption3ds)) {
        updateChargeWithExemption3ds(EXEMPTION_REJECTED, request.getCharge());
    } else if ("OUT_OF_SCOPE".equals(exemption3ds)) {
        updateChargeWithExemption3ds(EXEMPTION_OUT_OF_SCOPE, request.getCharge());
    }
});
```

A more radical idea might be to dispense with this method altogether and replace the `updateChargeExemptionForSoftDecline(request, response);` line above (line 171 as I write this) with this:

```java
response.getBaseResponse().flatMap(WorldpayOrderStatusResponse::getExemptionResponseResult).ifPresent(exemption3ds -> {
    if ("REJECTED".equals(exemption3ds)) {
        updateChargeWithExemption3ds(EXEMPTION_REJECTED, request.getCharge());
    } else if ("OUT_OF_SCOPE".equals(exemption3ds)) {
        updateChargeWithExemption3ds(EXEMPTION_OUT_OF_SCOPE, request.getCharge());
    }
});
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -180,12 +181,51 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
+        Optional<Exemption3ds> exemption3ds = Optional.empty();
+
+        if (!exemptionEngineEnabled) {
+            exemption3ds = Optional.of(EXEMPTION_NOT_REQUESTED);
+        } else if (response.getExemptionResponseResult().map("HONOURED"::equals).orElse(false)) {
+            exemption3ds = Optional.of(EXEMPTION_HONOURED);
+        } 
+
+        return exemption3ds;
+    }
```
alexbishop1: It makes me a little sad that this method can’t handle determining the exemption for a soft decline as well. I think we can make it do it by changing the method signature a little bit:

```java
private Optional<Exemption3ds> determineExemption(boolean exemptionEngineEnabled, GatewayResponse<WorldpayOrderStatusResponse> response) {
    if (!exemptionEngineEnabled) {
        return Optional.of(EXEMPTION_NOT_REQUESTED);
    }

    return response.getBaseResponse().flatMap(WorldpayOrderStatusResponse::getExemptionResponseResult).map(exemption3ds -> {
        switch (exemption3ds) {
            case "HONOURED":
                return EXEMPTION_HONOURED;
            case "REJECTED":
                return EXEMPTION_REJECTED;
            case "OUT_OF_SCOPE":
                return EXEMPTION_OUT_OF_SCOPE;
            default:
                return null;
        }
    });
}
```

Then we call it immediately after the `worldpayAuthoriseHandler.authoriseWithExemption(…)` or `worldpayAuthoriseHandler.authoriseWithoutExemption(…)` call (line 168 as I write this):

```java
var exemption3ds = determineExemption(exemptionEngineEnabled, response).orElse(null);
```

Then later on we can just do things like `if (exemption3ds == Exemption3ds.REJECTED)` etc.

alexbishop1: Actually, come to think of it, I don’t think we have a hard requirement to store `REJECTED` or `OUT_OF_SCOPE` only if it’s a soft decline (it might even be desirable to store it in other cases, such as if the payment is `AUTHORISED` but the exemption is `OUT_OF_SCOPE` — if such a thing is even possible). So I think we could actually make this method do the updating itself:

```java
private void storeExemption(boolean exemptionEngineEnabled, CardAuthorisationGatewayRequest request, GatewayResponse<WorldpayOrderStatusResponse> response) {
    if (!exemptionEngineEnabled) {
        updateChargeWithExemption3ds(Exemption3ds.EXEMPTION_NOT_REQUESTED);
    } else {
        response.getBaseResponse().flatMap(WorldpayOrderStatusResponse::getExemptionResponseResult).ifPresent(exemption3ds -> {
            switch (exemption3ds) {
                case "HONOURED":
                    updateChargeWithExemption3ds(Exemption3ds.HONOURED);
                    break;
                case "REJECTED":
                    updateChargeWithExemption3ds(Exemption3ds.EXEMPTION_REJECTED);
                    break;
                case "OUT_OF_SCOPE":
                    updateChargeWithExemption3ds(Exemption3ds.EXEMPTION_OUT_OF_SCOPE);
                    break;
                default:
                    // We asked for an exemption but the response did not say anything about it
             }
      	});
    }
}
```

alexbishop1 (changes requested):
I’ve got a couple of suggestions that I think could simplify this a bit further.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I don't think it's possible to establish a response is a soft decline and having `response.getBaseResponse().get()` to fail.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: Personally, I value readability of code more than using functional idioms. I definitely like seeing how people would do things differently! In this case I find this more difficult to read because
* I think in terms of `if...else...if...else`
* I always have to google what flatMap means whenever I see it being used
* It's not obvious to deduce what the type of `response.getBaseResponse().flatMap(worldpayOrderStatusResponse -> determineExemption(exemptionEngineEnabled, worldpayOrderStatusResponse))` is

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I like this idea.

oswaldquek: I've implemented this method and called it `calculateAndStoreExemption`.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: Well, this code block disappeared with the new `calculateAndStoreExemption` method.

oswaldquek:
@alexbishop1 This is ready for review again. Thanks.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -157,17 +161,14 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
-            LOGGER.info("Updated exemption_3ds of charge to {} - charge_external_id={}", EXEMPTION_NOT_REQUESTED, charge.getExternalId());
-            
             response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
         } else {
             response = worldpayAuthoriseHandler.authoriseWithExemption(request);
         }
         
         if (response.getBaseResponse().map(WorldpayOrderStatusResponse::isSoftDecline).orElse(false)) {
+
+            calculateAndStoreExemption(exemptionEngineEnabled, request.getCharge(), response);
```
alexbishop1: I think we can move this out to immediately after the first authorisation request (line 168 as I write this). That way we wouldn’t need the `else` block.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -180,12 +181,45 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
...
+                        break;
+                    case "REJECTED":
+                        updateChargeWithExemption3ds(EXEMPTION_REJECTED, charge);
+                        break;
+                    case "OUT_OF_SCOPE":
+                        updateChargeWithExemption3ds(EXEMPTION_OUT_OF_SCOPE, charge);
+                        break;
+                    default:
+                        LOGGER.info("Received unrecognised exemption 3ds response result {} from Worldpay - " +
+                                "charge_external_id={}", exemption3ds, charge.getExternalId());
```
alexbishop1: I wonder if we should log this at a higher level because I don’t think we should get any values we don’t expect (as opposed to no value, which won’t reach this line).

alexbishop1 (changes requested):
Looks good: I think centralising storing the exemption makes it more straightforward. Just a couple of a tiny suggestions.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: Good shout, fixed.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: I've logged this at WARN.

alexbishop1 (approved):
Good stuff!

PR #2790, opened 2021-01-06
PP-7589 Add new flag `allow_telephone_payment_notifications` to gateway_accounts

kbottla:
## WHAT YOU DID
- Adds flag `allow_telephone_payment_notifications` to gateway account which indicates if the account is used for telephone payments reporting.
- This flag will also be used to handle Worldpay notifications differently for telephone payments i.e., if the flag is enabled and we receive notifications for charges we are not aware of, worldpay notifications will be rejected (as external service might not have reported payment to us yet). If flag is disabled, we simply ignore worldpay notification (cases where services using same merchant account outside pay and we are receiving notifications about unknown payments).

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1561,4 +1561,20 @@
         </addColumn>
     </changeSet>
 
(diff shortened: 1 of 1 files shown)


SandorArpa (approved):
LOKTM

PR #2791, opened 2021-01-07
PP-7589 Reject Worldpay notifications only for telephone payments accounts

kbottla:
## WHAT YOU DID
- Rejects Worldpay notifications for accounts used for telephone payment reporting and when an associated charge is not found. This is to handle cases where we receive notification from Worldpay before the payment is reported to us.
- For any other accounts, simply ignore notification as notifications could be related to payments outside Pay.

Should set `allow_telephone_payment_notifications` flag for existing accounts before this goes out https://github.com/alphagov/pay-scripts/pull/1051


stephencdaly (approved):
LGTM

PR #2794, opened 2021-01-07
PP-7589 Add pact state to enable telephone payment notifications

stephencdaly:
Add pact state to update an account to enable telephone payment notifications.

The pact state needs to be used by public api before we make the change to check this flag on the gateway account for creating charges for telephone payment notifications.


SandorArpa (approved):
LOKTM

PR #2797, opened 2021-01-08
PP-7589 Restrict use of payment notifications API

stephencdaly:
Check if a gateway account has the `allow_telephone_payment_notifications` flag set to `true` for POST requests to the `v1/api/accounts/{accountId}/telephone-charge` endpoint.

If the flag isn't enabled, return a 403 with the error identifier TELEPHONE_PAYMENT_NOTIFICATIONS_NOT_ALLOWED to be handled by public API to return an appropriate error to the end consumer.

Renamed a few methods to make it clear they're for telephone payments while I was here to make things a bit easier.

## Note

Hiding whitespace changes in the diff will make reviewing a tiny bit easier due to some indentation changes.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
@@ -3,10 +3,14 @@ package uk.gov.pay.connector.charge.resource;
 import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.exception.TelephonePaymentNotificationsNotAllowedException;
 import uk.gov.pay.connector.charge.model.ChargeCreateRequest;
 import uk.gov.pay.connector.charge.model.telephone.TelephoneChargeCreateRequest;
 import uk.gov.pay.connector.charge.service.ChargeExpiryService;
 import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
 
 import javax.inject.Inject;
 import javax.validation.Valid;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -47,11 +50,15 @@
     public static final int MAX_AMOUNT = 10_000_000;
     private final ChargeService chargeService;
     private final ChargeExpiryService chargeExpiryService;
+    private GatewayAccountService gatewayAccountService;
```
alexbishop1: This can probably be `final`.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -86,9 +93,16 @@ public Response createNewTelephoneCharge(
             @NotNull @Valid TelephoneChargeCreateRequest telephoneChargeCreateRequest,
             @Context UriInfo uriInfo
     ) {
-        return chargeService.findCharge(accountId, telephoneChargeCreateRequest)
-                .map(response -> Response.status(200).entity(response).build())
-                .orElseGet(() -> Response.status(201).entity(chargeService.create(telephoneChargeCreateRequest, accountId).get()).build());
+        return gatewayAccountService.getGatewayAccount(accountId).map(gatewayAccount -> {
+            if (!gatewayAccount.isAllowTelephonePaymentNotifications()) {
+                throw new TelephonePaymentNotificationsNotAllowedException(gatewayAccount.getId());
+            }
```
alexbishop1: Should be able to do something with `filter(…)` and `orElseThrow(…)` here.

alexbishop1 (changes requested):
Looks good. Just a couple of small suggestions.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
stephencdaly: Cool thanks, I'll improve it

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
stephencdaly: Good spot. I'm pretty sure I used the intellij auto-generation for this and surprised it didn't make it final 🤔

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
stephencdaly: I guess if I wanted to do this I'd have to throw the `GatewayAccountNotFoundException` somewhere else as I wouldn't be able to differentiate whether gateway account didn't exist or the account was filtered out because it didn't have the flag here?

alexbishop1 (approved):

PR #2798, opened 2021-01-08
PP-5641: Generate a new transaction id when re-submitting an auth request

oswaldquek:
Generate a new transaction id when re-submitting an auth request due to a soft
decline. There's also a manual contract test for this.

Updated the
`contract/WorldpayPaymentProviderTest.submitAuthRequestWithExemptionEngineFlag`
test as we now have the exemption engine enabled on merchant code GBSRECABOFFGDSGBP.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -181,7 +181,8 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
                     response,
                     request.getCharge().getChargeStatus(),
                     request.getCharge().getChargeStatus());
-
+            
+            request.getCharge().setGatewayTransactionId(randomUUID().toString());
             response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
         } 
 
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -113,7 +113,11 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
 
     @Override
     public Optional<String> generateTransactionId() {
-        return Optional.of(randomUUID().toString());
+        return Optional.of(newTransactionId());
+    }
+
+    private String newTransactionId() {
+        return randomUUID().toString();
     }
 
     @Override
@@ -182,7 +186,7 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
                     request.getCharge().getChargeStatus(),
                     request.getCharge().getChargeStatus());
             
-            request.getCharge().setGatewayTransactionId(randomUUID().toString());
+            request.getCharge().setGatewayTransactionId(newTransactionId());
             response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
         } 
 
--- a/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
@@ -44,11 +44,13 @@ import java.util.Map;
 import java.util.Optional;
 
 import static java.lang.String.format;
+import static java.util.UUID.randomUUID;
 import static org.custommonkey.xmlunit.XMLAssert.assertXMLEqual;
 import static org.hamcrest.CoreMatchers.instanceOf;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.Is.is;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyList;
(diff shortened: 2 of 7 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -181,6 +181,7 @@ private GatewayOrder buildQuery(ChargeEntity charge) {
                     request.getCharge().getChargeStatus(),
                     request.getCharge().getChargeStatus());
             
+            request.getCharge().setGatewayTransactionId(randomUUID().toString());
```
alexbishop1: We should probably call the existing `generateTransactionId()` method here.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -265,6 +269,7 @@ void should_retry_without_exemption_flag_when_authorising_with_exemption_flag_re
         
         assertTrue(response.getBaseResponse().isPresent());
         assertEquals(secondResponse.getBaseResponse().get(), response.getBaseResponse().get());
+        assertNotEquals(cardAuthRequest.getCharge().getGatewayTransactionId(), gatewayTransactionId);
```
alexbishop1, line 334: Is it feasible to check the new transaction ID actually gets persisted to the database and replaces the old one?

alexbishop1 (changes requested):
Looks good. Couple of suggestions.

Could you tweak the commit message and PR description for this. It’s not just adding a test, it’s actually changing behaviour.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: That was what I went for originally, until I saw that that method returns an Optional and I know how you dislike doing gets on Optionals. Can you suggest a way forward please.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: I think that would be possible, but it's not a quick fix. It depends how fast you want to get the stories through due to our timescales. The `WorldpayPaymentProvider` doesn't concern itself with persistence. 
I did mention before I did make an integration test for this. It does assert the value in the database is a different one from the original one:
```
    @Test
    public void should_set_new_gateway_transaction_id_when_authorisation_with_exemption_engine_results_in_soft_decline() {
        String gatewayTransactionId = randomUUID().toString();
        String chargeId = createNewChargeWith(ENTERING_CARD_DETAILS, gatewayTransactionId);

        assertEquals(databaseTestHelper.getChargeByExternalId(chargeId).get("gateway_transaction_id"), gatewayTransactionId);
        
        worldpayMockClient.mockResponsesForExemptionEngineSoftDecline();

        givenSetup()
                .body(validAuthorisationDetails)
                .post(authoriseChargeUrlFor(chargeId))
                .then()
                .statusCode(OK.getStatusCode());
        
        assertNotEquals(databaseTestHelper.getChargeByExternalId(chargeId), gatewayTransactionId);
    }
```
However I didn't think it's a great test because we can't really verify it's the new transactionId generated in the `WorldpayPaymentProvider`. The transaction id will simply be "transaction-id" as found in https://github.com/alphagov/pay-connector/blob/master/src/test/resources/templates/worldpay/authorisation-success-response.xml#L6

oswaldquek:
@alexbishop1 Does "It appears that for a soft decline, we must generate a new transaction id when
re-submitting the request" in the commit message not indicate behaviour was changed?

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
alexbishop1: One way would be `generateTransactionId().ifPresent(transactionId -> request.getCharge().setGatewayTransactionId(transactionId));`.

We could instead introduce a private method that returns a non-Optional (because we know `WorldpayPaymentProvider` supports generating gateway transaction IDs), so we’d have something like this:

```java
@Override
public Optional<String> generateTransactionId() {
    return Optional.of(newTransactionId());
}

private String newTransactionId() {
    return randomUUID().toString();
}
```

Then within the `WorldpayPaymentProvider` we’d just call the private method.

Nether of these are ideal (it would probably be better if there was an interface indicating a `PaymentProvider` can generate a transaction ID but that’s obviously a chunky change).

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
alexbishop1: While it doesn’t check the new ID is generated by `WorldpayPaymentProvider`, it does check it replaces the existing one. So I think this test in combination with the unit test would have us covered.

alexbishop1:
“It appears that for a soft decline, we must generate a new transaction id when re-submitting the request” does cover it but I think the behaviour change should be in the first line, which currently just mentions the contract tests.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
oswaldquek: Great, I've added in that test now.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
oswaldquek: 👍 I've gone with using the private method.

oswaldquek:
@alexbishop1 I've made the code changes, updated the commit message and PR description for this so I think this is good to review now.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
```diff
@@ -41,6 +45,24 @@ public WorldpayCardResourceIT() {
         super("worldpay");
     }
 
+    @Test
+    public void should_set_new_gateway_transaction_id_when_authorisation_with_exemption_engine_results_in_soft_decline() {
+        String gatewayTransactionId = randomUUID().toString();
```
alexbishop1, line 50: I suppose this could be something like `"new_gateway_transaction_id"` to make it really obvious.

alexbishop1 (approved):

PR #2801, opened 2021-01-11
PP-7554: Add a new path for enabling/disabling Worldpay 3DS Exemption Engine

adityapahuja:
This pull request
- updates GatewayAccountRequestValidator to include a new path `worldpay_exemption_engine_enabled` to validate that the value in the request is boolean.
- updates the existing integration test to ensure that `exemption_engine_enabled` is present after calling the endpoint `/v1/frontend/accounts/`.
- updates GatewayAccountService to support the new path `worldpay_exemption_engine_enabled` for enabling or disabling Worldpay 3DS Exemption Engine and to throw exception if the gateway account is not Worldpay or the Worldpay gateway account has no 3DS flex credentials entity.
- updates Worldpay3dsFlexCredentialsEntity to include the annotation `@JsonIgnore` on `id`, `gatewayAccountEntity`, `gatewayAccountId` and `jwtMacKey`. `gatewayAccountEntity` should be ignored to avoid throwing StackOverflow exception. `id` and `gatewayAccountId` are not required because they are used for database purposes (e.g. primary key). `jwtMacKey` should be ignored because it is a secret field.
- adds getWorldpay3dsFlexCredentialsEntity() method to GatewayAccountEntity because this method provides an instance of Worldpay3dsFlexCredentialsEntity which is needed for toggling `exemption_engine` in the database. This method is annotated with `@JsonIgnore` because getWorldpay3dsFlexCredentials() method already has `@JsonProperty("worldpay_3ds_flex")` and is used in Responses.
- updates Worldpay3dsFlexCredentialsEntity to allow us to set exemption_engine to true or false in the database.
- adds an integration test to GatewayAccountResource to ensure that Worldpay exemption engine can be enabled for Worldpay gateway account.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
@@ -5,10 +5,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.cardtype.dao.CardTypeDao;
 import uk.gov.pay.connector.common.model.api.jsonpatch.JsonPatchRequest;
-import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
 import uk.gov.pay.connector.gatewayaccount.exception.DigitalWalletNotSupportedGatewayException;
 import uk.gov.pay.connector.gatewayaccount.exception.MerchantIdWithoutCredentialsException;
+import uk.gov.pay.connector.gatewayaccount.exception.MissingWorldpay3dsFlexCredentialsEntityException;
+import uk.gov.pay.connector.gatewayaccount.exception.NotSupportedGatewayAccountException;
 import uk.gov.pay.connector.gatewayaccount.model.EmailCollectionMode;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccount;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
@@ -27,6 +28,7 @@ import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
 import static java.util.Map.entry;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.CREDENTIALS_GATEWAY_MERCHANT_ID;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_ALLOW_APPLE_PAY;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_ALLOW_GOOGLE_PAY;
@@ -43,6 +45,7 @@ import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequest
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_MOTO_MASK_CARD_NUMBER_INPUT;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_MOTO_MASK_CARD_SECURITY_CODE_INPUT;
 import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_NOTIFY_SETTINGS;
+import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED;
 
 public class GatewayAccountService {
 
@@ -179,12 +182,27 @@ public class GatewayAccountService {
             entry(
                 FIELD_ALLOW_TELEPHONE_PAYMENT_NOTIFICATIONS,
                 (gatewayAccountRequest, gatewayAccountEntity) -> gatewayAccountEntity.setAllowTelephonePaymentNotifications(gatewayAccountRequest.valueAsBoolean())
+            ),
+            entry(
+                FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED,
+                (gatewayAccountRequest, gatewayAccountEntity) -> {
+                    throwIfGatewayAccountIsNotWorldpay(gatewayAccountEntity, FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED);
+                    var worldpay3dsFlexCredentialsEntity = gatewayAccountEntity.getWorldpay3dsFlexCredentialsEntity()
+                        .orElseThrow(() -> new MissingWorldpay3dsFlexCredentialsEntityException(gatewayAccountEntity.getId(), FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED));
+                    worldpay3dsFlexCredentialsEntity.setExemptionEngineEnabled(gatewayAccountRequest.valueAsBoolean());
+                }
             )
     );
 
     private void throwIfNotDigitalWalletSupportedGateway(GatewayAccountEntity gatewayAccountEntity) {
-        if (!PaymentGatewayName.WORLDPAY.getName().equals(gatewayAccountEntity.getGatewayName())) {
+        if (!WORLDPAY.getName().equals(gatewayAccountEntity.getGatewayName())) {
             throw new DigitalWalletNotSupportedGatewayException(gatewayAccountEntity.getGatewayName());
+        }
+    }
+
+    private void throwIfGatewayAccountIsNotWorldpay(GatewayAccountEntity gatewayAccountEntity, String path) {
+        if (!WORLDPAY.getName().equals(gatewayAccountEntity.getGatewayName())) {
+            throw new NotSupportedGatewayAccountException(gatewayAccountEntity.getId(), WORLDPAY.getName(), path);
         }
     }
 }
--- a/src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
+++ b/src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
@@ -70,10 +70,12 @@ public class GatewayAccountRequestValidatorTest {
             "replace, moto_mask_card_security_code_input, null, Field [value] is required",
             "replace, moto_mask_card_security_code_input, unfalse, Value [unfalse] must be of type boolean for path [moto_mask_card_security_code_input]",
             "replace, allow_telephone_payment_notifications, null, Field [value] is required",
-            "replace, allow_telephone_payment_notifications, unfalse, Value [unfalse] must be of type boolean for path [allow_telephone_payment_notifications]"
+            "replace, allow_telephone_payment_notifications, unfalse, Value [unfalse] must be of type boolean for path [allow_telephone_payment_notifications]",
+            "replace, worldpay_exemption_engine_enabled, null, Field [value] is required",
+            "replace, worldpay_exemption_engine_enabled, unfalse, Value [unfalse] must be of type boolean for path [worldpay_exemption_engine_enabled]"
     })
     public void shouldThrowWhenRequestsAreInvalid(String op, String path, @Nullable String value, String expectedErrorMessage) {
-        Map<String, String> patch = new HashMap<String, String>() {{
+        Map<String, String> patch = new HashMap<>() {{
             put(FIELD_OPERATION, op);
             put(FIELD_OPERATION_PATH, path);
         }};
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
@@ -26,6 +26,7 @@ import static org.hamcrest.Matchers.not;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
 import static org.hamcrest.collection.IsMapContaining.hasKey;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType.TEST;
 import static uk.gov.pay.connector.util.JsonEncoder.toJson;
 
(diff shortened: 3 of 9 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/exception/Missing3dsFlexCredentialsEntityException.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.pay.connector.gatewayaccount.exception;
+
+import org.apache.commons.lang3.StringUtils;
+import uk.gov.pay.connector.util.ResponseUtil;
+
+import javax.ws.rs.BadRequestException;
+
+public class Missing3dsFlexCredentialsEntityException extends BadRequestException {
```
alexbishop1: Can this be `MissingWorldpay3dsFlexCredentialsEntityException` just to make it a bit clearer?

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/Missing3dsFlexCredentialsEntityException.java
```diff
@@ -0,0 +1,16 @@
...
+    public Missing3dsFlexCredentialsEntityException(Long accountId,
+                                                    String gatewayName,
+                                                    String path){
+        super(ResponseUtil.badRequestResponse(
+                String.format("%s gateway account %s has no 3DS flex credentials entity for path [%s].",
+                        StringUtils.capitalize(gatewayName), accountId, path)));
```
alexbishop1: Not sure it’s worth capitalising the first letter of the gateway name — we don’t bother anywhere else.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/WrongGatewayAccountException.java
```diff
@@ -0,0 +1,16 @@
+package uk.gov.pay.connector.gatewayaccount.exception;
+
+import org.apache.commons.lang3.StringUtils;
+import uk.gov.pay.connector.util.ResponseUtil;
+
+import javax.ws.rs.BadRequestException;
+
+public class WrongGatewayAccountException extends BadRequestException {
```
alexbishop1: Following `DigitalWalletNotSupportedGatewayException`, could we name this something like `JsonPathNotSupportedGatewayException`?

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
```diff
@@ -179,12 +183,29 @@ public boolean isATelephonePaymentNotificationAccount(String merchantCode) {
...
+            ),
+            entry(
+                FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED,
+                (gatewayAccountRequest, gatewayAccountEntity) -> {
+                    throwIfGatewayAccountIsNotWorldpay(gatewayAccountEntity, FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED);
+                    Optional<Worldpay3dsFlexCredentialsEntity> worldpay3dsFlexCredentialsEntity = gatewayAccountEntity.getWorldpay3dsFlexCredentialsEntity();
+                    worldpay3dsFlexCredentialsEntity.ifPresentOrElse(
+                        worldpay3dsFlexCredentials -> worldpay3dsFlexCredentials.setExemptionEngineEnabled(gatewayAccountRequest.valueAsBoolean()),
+                        () -> { throw new Missing3dsFlexCredentialsEntityException(gatewayAccountEntity.getId(), gatewayAccountEntity.getGatewayName(), FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED); }
+                    );
```
alexbishop1: Might be slightly nicer like this?

```java
var worldpay3dsFlexCredentialsEntity = gatewayAccountEntity.getWorldpay3dsFlexCredentialsEntity().orElseThrow(() ->
    new Missing3dsFlexCredentialsEntityException(gatewayAccountEntity.getId(), gatewayAccountEntity.getGatewayName(), FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED));
worldpay3dsFlexCredentialsEntity.setExemptionEngineEnabled(gatewayAccountRequest.valueAsBoolean());
```

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
```diff
@@ -70,10 +70,12 @@ public void before() {
...
             "replace, moto_mask_card_security_code_input, unfalse, Value [unfalse] must be of type boolean for path [moto_mask_card_security_code_input]",
             "replace, allow_telephone_payment_notifications, null, Field [value] is required",
-            "replace, allow_telephone_payment_notifications, unfalse, Value [unfalse] must be of type boolean for path [allow_telephone_payment_notifications]"
+            "replace, allow_telephone_payment_notifications, unfalse, Value [unfalse] must be of type boolean for path [allow_telephone_payment_notifications]",
+            "replace, worldpay_exemption_engine_enabled, null, Field [value] is required",
+            "replace, worldpay_exemption_engine_enabled, unfalse, Value [unfalse] must be of type boolean for path [worldpay_exemption_engine_enabled]"
     })
     public void shouldThrowWhenRequestsAreInvalid(String op, String path, @Nullable String value, String expectedErrorMessage) {
-        Map<String, String> patch = new HashMap<String, String>() {{
+        Map<String, String> patch = new HashMap<>() {{
```
alexbishop1, line 78: Since you’re changing this any way, may as well use `Map.of(…)`.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -480,7 +481,35 @@ public void shouldReturn3dsFlexCredentials_whenGatewayAccountHasCreds() {
...
+                .body(payload)
+                .patch("/v1/api/accounts/" + gatewayAccountId)
+                .then()
+                .statusCode(OK.getStatusCode());
+
+        givenSetup()
+                .get("/v1/frontend/accounts/" + gatewayAccountId)
+                .then()
+                .statusCode(OK.getStatusCode())
+                .body("worldpay_3ds_flex.exemption_engine_enabled", is(true));
```
alexbishop1, line 512: Good that we have this to check the new value is actually persisted.

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
```diff
@@ -398,4 +413,68 @@ public void shouldGetGatewayAccountByExternalId() {
...
+                "value", false)));
+        when(mockGatewayAccountEntity.getGatewayName()).thenReturn(WORLDPAY.getName());
+        when(mockGatewayAccountEntity.getWorldpay3dsFlexCredentialsEntity()).thenReturn(Optional.of(mockWorldpay3dsFlexCredentialsEntity));
+        when(mockGatewayAccountDao.findById(GATEWAY_ACCOUNT_ID)).thenReturn(Optional.of(mockGatewayAccountEntity));
+
+        Optional<GatewayAccount> optionalGatewayAccount = gatewayAccountService.doPatch(GATEWAY_ACCOUNT_ID, request);
+
+        assertThat(optionalGatewayAccount.isPresent(), is(true));
+        verify(mockWorldpay3dsFlexCredentialsEntity).setExemptionEngineEnabled(false);
+        verify(mockGatewayAccountDao).merge(mockGatewayAccountEntity);
```
alexbishop1: Can we use in-order verification for these?

alexbishop1: We’re assuming that persisting the `GatewayAccountEntity` persists the linked `Worldpay3dsFlexCredentialsEntity` but the integration test should ensure we definitely do persist so I think this is OK.

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
```diff
@@ -398,4 +413,68 @@ public void shouldGetGatewayAccountByExternalId() {
...
+
+        Optional<GatewayAccount> optionalGatewayAccount = gatewayAccountService.doPatch(GATEWAY_ACCOUNT_ID, request);
+
+        assertThat(optionalGatewayAccount.isPresent(), is(true));
+        verify(mockWorldpay3dsFlexCredentialsEntity).setExemptionEngineEnabled(true);
+        verify(mockGatewayAccountDao).merge(mockGatewayAccountEntity);
+    }
+
+    @Test
+    public void shouldThrowWrongGatewayAccountExceptionWhenAccountIsNotWorldPay() {
```
alexbishop1: We usually use “Worldpay”.

alexbishop1 (changes requested):
Looks like this will do the job and test coverage looks good. Just a couple of suggestions.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/Missing3dsFlexCredentialsEntityException.java
adityapahuja: I intentionally made this generic exception by including a parameter `gatewayName` rather than a specific exception because another payment service provider might need this exception in the future. Do you still want me to drop the parameter `gatewayName` and rename the exception to `MissingWorldpay3dsFlexCredentialsEntityException`?

alexbishop1: 3DS Flex is a Worldpay trademark, so we’re not going to have it for any other gateway.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/Missing3dsFlexCredentialsEntityException.java
adityapahuja: `gatewayName` returns the gateway name in lowercase e.g. worldpay. Putting it at the beginning of the sentence will look odd. Do you still want me to drop the capitalize() function?

alexbishop1: It does look strange but we do it in plenty of other places.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/WrongGatewayAccountException.java
adityapahuja: `JsonPathNotSupportedGatewayException` suggests that the path is not available or is not supported. How about `NotWorldpayGatewayAccountException`? It tells us that the gateway account we used is not Worldpay gateway account for the path `worldpay_exemption_engine_enabled`. This exception name is more clear and easy to understand.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/Missing3dsFlexCredentialsEntityException.java
adityapahuja: This has been done.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/Missing3dsFlexCredentialsEntityException.java
adityapahuja: This has been done.

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
adityapahuja: This has been done.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
adityapahuja: This has been done.

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
adityapahuja: This has been done.

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
adityapahuja: This has been done.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/WrongGatewayAccountException.java
alexbishop1: I was aiming for consistency with the ending `NotSupportedGatewayException`.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
```diff
@@ -70,13 +70,14 @@ public void before() {
...
+            "replace, worldpay_exemption_engine_enabled, unfalse, Value [unfalse] must be of type boolean for path [worldpay_exemption_engine_enabled]"
     })
     public void shouldThrowWhenRequestsAreInvalid(String op, String path, @Nullable String value, String expectedErrorMessage) {
-        Map<String, String> patch = new HashMap<String, String>() {{
-            put(FIELD_OPERATION, op);
-            put(FIELD_OPERATION_PATH, path);
-        }};
+        Map<String, String> patch = Map.of(
+            FIELD_OPERATION, op,
+            FIELD_OPERATION_PATH, path);
```
alexbishop1: Sorry, I just realised `Map.of(…)` isn’t going to work because we need the map to be mutable!

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
```diff
@@ -398,4 +413,70 @@ public void shouldGetGatewayAccountByExternalId() {
...
+    public void shouldThrowWrongGatewayAccountExceptionWhenAccountIsNotWorldpay() {
+        JsonPatchRequest request = JsonPatchRequest.from(objectMapper.valueToTree(Map.of(
+                "op", "replace",
+                "path", FIELD_WORLDPAY_EXEMPTION_ENGINE_ENABLED,
+                "value", true)));
+        when(mockGatewayAccountEntity.getGatewayName()).thenReturn(EPDQ.getName());
+        when(mockGatewayAccountDao.findById(GATEWAY_ACCOUNT_ID)).thenReturn(Optional.of(mockGatewayAccountEntity));
+
+        exceptionRule.expect(NotWorldpayGatewayAccountException.class);
+        exceptionRule.expectMessage(BAD_REQUEST_MESSAGE);
```
alexbishop1, line 463: I don’t think we need to check the exact message.

alexbishop1 (changes requested):
So it turns out of one the suggestions I made would break stuff!

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountRequestValidatorTest.java
adityapahuja: Yes, I got a test failure. I will fix it.

adityapahuja: Fixed.

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/WrongGatewayAccountException.java
adityapahuja: Thanks. This has been done.

src/test/java/uk/gov/pay/connector/service/GatewayAccountServiceTest.java
adityapahuja: At least, we know that it is returning 400 = Bad request message.

alexbishop1 (approved):

PR #2806, opened 2021-01-12
PP-7609 @PactVerifyProvider for Gateway3dsExemptionResultObtained

alexbishop1:
Add `@PactVerifyProvider` for `GATEWAY_3DS_EXEMPTION_RESULT_OBTAINED` message.


nimalank7 (approved):
I paired with Alex on this and am confident that it needs no further review as per our pairing policy.

PR #2807, opened 2021-01-12
PP-7554: Update Gateway Account Resource DTO to include Worldpay 3DS Flex

adityapahuja:
This pull request has two commits.

1. The first commit adds `worldpay_3ds_flex` JSON field to GatewayAccountResourceDTO. This will enable both GET endpoints `/v1/api/accounts` and `/v1/frontend/accounts` to return GatewayAccountResourceDTO object containing `worldpay_3ds_flex` JSON field. It updates the existing integration test to ensure that the new field is present.
2. The second commit adds `worldpay_3ds_flex` JSON field to an account object returned by calling GET `/v1/api/accounts` endpoint in API specification documentation.

This change is needed for scenario 6 taken from the story.

### DIFF
--- a/docs/api_specification.md
+++ b/docs/api_specification.md
@@ -247,7 +247,8 @@ Content-Type: application/json
             "enabled": true,
             "template_body": null
         }
-    }
+    },
+    "worldpay_3ds_flex": null
 }
 ```
 
@@ -376,6 +377,11 @@ Content-Type: application/json
               "template_body": null
           }
       },
+      "worldpay_3ds_flex": {
+        "issuer": "44992a087a0c4849895cc9a3",
+        "organisational_unit_id", "57992a087a0c4849895ab8a2",
+        "exemption_engine_enabled": true,
+      },
       "_links": {
         "self": {
           "href": "https://connector.example.com/v1/api/accounts/200"
@@ -412,6 +418,11 @@ Content-Type: application/json
               "enabled": true,
               "template_body": null
           }
+      },
+      "worldpay_3ds_flex": {
+        "issuer": "44992a087a0c4849895cc9a3",
+        "organisational_unit_id", "57992a087a0c4849895ab8a2",
+        "exemption_engine_enabled": true,
       },
       "_links": {
         "self": {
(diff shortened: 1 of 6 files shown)


docs/api_specification.md
```diff
@@ -376,6 +376,11 @@ Content-Type: application/json
               "template_body": null
           }
       },
+      "worldpay_3ds_flex": {
+        "issuer": "44992a087a0c4849895cc9a3",
+        "organisational_unit_id", "57992a087a0c4849895ab8a2",
+        "exemption_engine_enabled": true,
+      },
```
oswaldquek, line 384: Should this section be also under [GET /v1/api/accounts/{accountsId}](https://github.com/alphagov/pay-connector/blob/master/docs/api_specification.md#get-v1apiaccountsaccountsid), [GET /v1/frontend/accounts/{accountId}](https://github.com/alphagov/pay-connector/blob/master/docs/api_specification.md#get-v1frontendaccountsaccountid) and [GET /v1/frontend/charges/{chargeId}](https://github.com/alphagov/pay-connector/blob/master/docs/api_specification.md#get-v1frontendchargeschargeid) and their relevant `Response field description` sections too?

oswaldquek: This being said it feels wrong to have to duplicate so many of these responses. I wonder if this api_specification.md is best replaced by swagger instead.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTO.java
```diff
@@ -85,6 +86,10 @@
     @JsonProperty("moto_mask_card_security_code_input")
     private boolean motoMaskCardSecurityCodeInput;
 
+    @JsonInclude(NON_NULL)
+    @JsonProperty("worldpay_3ds_flex")
+    private Optional<Worldpay3dsFlexCredentials> worldpay3dsFlexCredentials;
```
oswaldquek: Not sure about the use of `Optional` here, why can't it just be `Worldpay3dsFlexCredentials`?

oswaldquek (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTO.java
adityapahuja: There are two reasons for using `Optional`.

1. There are gateway accounts that are not Worldpay and they do not have Worldpay 3DS Flex Credentials.
2. The [`getWorldpay3dsFlexCredentials()` method](https://github.com/alphagov/pay-connector/blob/eec188145d13430f1303fe26bb90a1a52cc8c315/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java#L239) in GatewayAccountEntity returns `Optional<Worldpay3dsFlexCredentials>`. This change forces GatewayAccountResourceDTO to use `Optional<Worldpay3dsFlexCredentials>` for consistency.

oswaldquek: Using an `Optional` as a field generally isn't good practice. From https://stackoverflow.com/questions/26327957/should-java-8-getters-return-optional-type/26328555#26328555: "You should almost never use it as a field of something or a method parameter."

Also IntelliJ gives a warning about using Optional fields.

I'd suggest doing this instead 
```
    public Optional<Worldpay3dsFlexCredentials> getWorldpay3dsFlexCredentials() {
        return Optional.ofNullable(worldpay3dsFlexCredentials);
    }
```

oswaldquek: And I think this [constructor](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTO.java#L91-L113) could be reduced to 
```
public GatewayAccountResourceDTO(GatewayAccountEntity gatewayAccountEntity)
```

docs/api_specification.md
adityapahuja: I have made the changes to the endpoints you mentioned and added the missing endpoint (GET /v1/frontend/accounts) to API specification documentation.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTO.java
adityapahuja: I have dropped `Optional`. It would have been good for Java not to allow `Optional` to be used in fields and method parameters. IntelliJ does not give me a warning about using Optional fields. Where do you see the warning message?

oswaldquek (approved):

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountResourceDTO.java
oswaldquek: Screenshot 2021-01-12 at 14 15 45
The yellow highlight

adityapahuja: Thanks. I have fixed the setting in IntelliJ.

PR #2821, opened 2021-01-18
PP-7656 Add frontend get account by external ID endpoint

stephencdaly:
For get account by internal ID, we have 2 endpoints:
- /v1/api/accounts/<id>
- /v1/frontend/accounts/<id>

The /v1/api endpoint is used by public api and the `/v1/frontend`
endpoint is used by selfservice and toolbox. The `/v1/frontend` endpoint
includes the gateway credentials but the `/v1/api` endpoint does not.

For get account by external ID we added the endpoint
`/v1/api/accounts/external-id/<externalId>`

This returns the gateway account without the gateway credentials and
is used by selfservice.

However, to be consistent with the get account endpoint, and as
selfservice needs to get the gateway credentials we should have instead
added a `/v1/frontend/accounts/external-id/<externalId>` endpoint.

This commit adds the new endpoint. The `/v1/api` endpoint will be removed
when it is no longer used by selfservice.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
@@ -127,6 +127,20 @@ public class GatewayAccountResource {
     }
 
     @GET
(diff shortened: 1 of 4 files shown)


SandorArpa (approved):
LOKTM

PR #2822, opened 2021-01-18
PP-7643 Replace assertEquals(…, …) with assertThat(…, is(…))

alexbishop1:


oswaldquek (approved):
This is fine but why the change?

alexbishop1:
Change was suggested in story review.

PR #2824, opened 2021-01-19
PP-7656 Remove old get account by external ID endpoint

stephencdaly:
This endpoint has been replaced by an equivalent endpoint that is `/v1/frontend/` rather than `/v1/api/`


SandorArpa (approved):
LOKTM

PR #2827, opened 2021-01-25
PP-5025: Add secondary Apple Pay payment processing key and certificate

oswaldquek:
Adding a secondary to allow for zero downtime when we create a new payment
processing certificate on the Apple developer portal.


mrlumbu (changes requested):
I'd like to see some tests around the introduction of primary and secondary keys

oswaldquek:
@mrlumbu There are currently two tests in ApplePayDecrypterTest: `should_throw_exception_when_secondary_key_and_cert_are_missing` and `should_decrypt_data_with_secondary_keys_when_primary_private_key_and_public_certificate_are_invalid`.

src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypter.java
```diff
@@ -44,46 +46,86 @@
...
+        byte[] rawData;
+
         try {
-            byte[] data = BASE64_DECODER.decode(applePayAuthRequest.getEncryptedPaymentData().getData().getBytes(UTF_8));
-            byte[] ephemeralPublicKey = BASE64_DECODER.decode(applePayAuthRequest.getEncryptedPaymentData().getHeader().getEphemeralPublicKey().getBytes(UTF_8));
-            byte[] rawData = decrypt(certificate, privateKey, ephemeralPublicKey, data);
-            return objectMapper.readValue(new String(rawData, UTF_8), AppleDecryptedPaymentData.class);
+            rawData = decrypt(primaryCertificate, primaryPrivateKey, ephemeralPublicKey, data);
         } catch (Exception e) {
+            if (secondaryCertificate.isPresent() && secondaryPrivateKey.isPresent()) {
```
mrlumbu, line 97: Not quite getting in what circumstances the secondary cert and key wouldn't be present given the constructor checks. Given we have all the checking in constructor, would it not be best to also check the optional is non-empty there too?
Alternatively you could not require a secondary key in constructor, and only try and retrieve it from config if primary key fails for whatever reason

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
```diff
@@ -1,53 +1,115 @@
...
+        Logger root = (Logger) LoggerFactory.getLogger(ApplePayDecrypter.class);
+        root.setLevel(Level.INFO);
+        root.addAppender(mockAppender);
     }
 
     @Test
-    public void shouldDecryptData_whenPrivateKeyAndPublicCertificateAreValid() {
-        AppleDecryptedPaymentData appleDecryptedPaymentData = applePayDecrypter.performDecryptOperation(applePayAuthRequest);
+    void should_throw_exception_when_secondary_key_and_cert_are_missing() {
+        String primaryKey = "MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgjyo3fzxT7j+CFxC7I4B5iVee2FUyn2vfOSjcgp2/g6qhRANCAARdoBFEtnuapXFKw4DYWsW0yV4bavpdWKszkefi19AhlIRE3WSNWSn25W5tZNFjMWtLISBmqANyufx2xP19oRvy"; //pragma: allowlist secret
```
mrlumbu: maybe call these `invalidPrimaryKey` and `invalidPrimaryCert`?

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
```diff
@@ -1,53 +1,115 @@
...
+
+        Logger root = (Logger) LoggerFactory.getLogger(ApplePayDecrypter.class);
+        root.setLevel(Level.INFO);
+        root.addAppender(mockAppender);
     }
 
     @Test
-    public void shouldDecryptData_whenPrivateKeyAndPublicCertificateAreValid() {
-        AppleDecryptedPaymentData appleDecryptedPaymentData = applePayDecrypter.performDecryptOperation(applePayAuthRequest);
+    void should_throw_exception_when_secondary_key_and_cert_are_missing() {
```
mrlumbu: not sure why snake case? And maybe to make test super clear, include fact that primary things are invalid

src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypter.java
```diff
@@ -44,46 +46,86 @@
...
+                LOGGER.info("Could not decrypt Apple auth request with primary key, trying with secondary key.");
+                try {
+                    rawData = decrypt(secondaryCertificate.get(), secondaryPrivateKey.get(), ephemeralPublicKey, data);
+                } catch (Exception ex) {
+                    LOGGER.error("Error while trying to decrypt apple pay payload: " + ex.getMessage());
+                    throw new InvalidKeyException("Error while trying to decrypt apple pay payload: " + ex.getMessage());
+                }
+            } else {
+                LOGGER.info("Could not decrypt Apple auth request with primary key, and secondary primary key is not " +
```
mrlumbu: do you mean `secondary primary key`? Surely just `secondary`?

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
```diff
@@ -1,53 +1,115 @@
...
+
+        assertThrows(InvalidKeyException.class, () -> 
+                applePayDecrypterWithInvalidPrimaryAndMissingSecondaryKeyCerts.performDecryptOperation(applePayAuthRequest));
+
+        ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
+        verify(mockAppender, times(3)).doAppend(loggingEventArgumentCaptor.capture());
+        List<LoggingEvent> logs = loggingEventArgumentCaptor.getAllValues();
+        assertTrue(logs.stream().anyMatch(loggingEvent ->
+                loggingEvent.getMessage().contains("Could not decrypt Apple auth request with primary key, and secondary " +
+                        "primary key is not present. Throwing an InvalidKeyException.")));
```
mrlumbu: I'm not sure about asserting literal text of log lines - very brittle. Unless there's some critical monitoring that relies on this wording (which would also not be great)?

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
```diff
@@ -1,53 +1,115 @@
...
+
+        ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
+        verify(mockAppender, times(2)).doAppend(loggingEventArgumentCaptor.capture());
+        List<LoggingEvent> logs = loggingEventArgumentCaptor.getAllValues();
+        assertTrue(logs.stream().anyMatch(loggingEvent -> 
+                loggingEvent.getMessage().contains("Could not decrypt Apple auth request with primary key, trying with secondary key.")));
+    }
+    
+    @Test
+    void shouldDecryptData_whenPrivateKeyAndPublicCertificateAreValid() {
```
mrlumbu: Maybe `whenPrimaryPrivateKeyAndPrimaryPublic…`?

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
```diff
@@ -58,31 +120,39 @@ public void shouldDecryptData_whenPrivateKeyAndPublicCertificateAreValid() {
         assertThat(appleDecryptedPaymentData.getPaymentData().getEciIndicator(), is("5"));
     }
 
-    @Test(expected = RuntimeException.class)
-    public void shouldThrowException_whenPublicCertificateIsInvalid() {
-        when(mockApplePayConfig.getPublicCertificate()).thenReturn("nope");
-        applePayDecrypter = new ApplePayDecrypter(mockConfig, objectMapper);
-        applePayDecrypter.performDecryptOperation(applePayAuthRequest);
+    @Test
+    void shouldThrowException_whenPublicCertificateIsInvalid() {
```
mrlumbu: I think it would be clearer to make explicit that there is also no secondary key and cert in these two tests. I was a bit confused here

mrlumbu (changes requested):
Have made some smallish comments on clarifying tests

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
oswaldquek: Because snake case is easier to read than `shouldThrowExceptionWhenPrimaryKeyAndCertAreInvalidAndSecondaryKeyAndCertAreMissing`. There currently aren't any agreed standards on snake vs camel case and we've used a combination of them everywhere.

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
oswaldquek: No there's no monitoring that depends on these log lines. We have asserted on log lines in other test cases but I'm all for having less code and I don't think these assertions add anything so I'll remove them as I think the rest of the test code is sufficient.

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
mrlumbu: as far as I can see we use camel case with underscores to separate clauses. That's certainly what we do in this class, so may be best to stick to this?

oswaldquek: we've already accepted full snake case in other places, e.g. https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java. Don't really think we should get caught up in this cosmetic issue really.

mrlumbu: fair enough. agree it's just cosmetic

mrlumbu (approved):

PR #2828, opened 2021-01-26
BAU Do not log a InvalidStateTransitionException before trying to force

danworth:
We will try to force the state transition if an
`InvalidStateTransitionException` is thrown. We will log an error if this
forcing of the state transition fails. Therefore do not log an error
before trying.

## WHAT ##
This is to fix https://sentry.io/organizations/govuk-pay/issues/2067568955/?project=5480144&referrer=slack

We also log when we're going to try to force a state transition which is why I think this entire log entry can go rather than just down grade it https://github.com/alphagov/pay-connector/blob/6a96fdec7f3d3bb19c9a6b7622815224eb4b9050/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java#L316

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/ChargeNotificationProcessor.java
@@ -48,16 +48,7 @@ public class ChargeNotificationProcessor {
         try {
             chargeService.transitionChargeState(chargeEntity, newStatus, gatewayEventDate);
         } catch (InvalidStateTransitionException e) {
-            logger.error(format("%s (%s) notification '%s' could not be used to update charge: %s",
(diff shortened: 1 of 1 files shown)


stephencdaly (approved):
Seems sensible

alexbishop1:
I presume this won’t affect the Splunk alerts we send to Zendesk that tell us we had to do a forced capture at all? (We use those to contact services to change their settings sometimes.)

danworth:
> I presume this won’t affect the Splunk alerts we send to Zendesk that tell us we had to do a forced capture at all? (We use those to contact services to change their settings sometimes.)

Here's the search and it uses a different log line (the line we log when we actually do the forcing)
https://github.com/alphagov/cyber-security-splunk-apps/blob/7ee92a91197570941c6571962ecf39f141188804/gds-004-pay/default/savedsearches.conf#L1023

PR #2832, opened 2021-01-28
PP-5025: Fix e2e tests

oswaldquek:


src/main/resources/config/config.yaml
```diff
@@ -45,8 +45,8 @@ worldpay:
   applePay:
     primaryPrivateKey: ${APPLE_PAY_PAYMENT_PROCESSING_PRIVATE_KEY:-privateKeyWhichShouldBeBase64Encoded}
     primaryPublicCertificate: ${APPLE_PAY_PAYMENT_PROCESSING_CERTIFICATE:-publicCertificateWhichShouldBeBase64Encoded}
-    secondaryPrivateKey: ${APPLE_PAY_PAYMENT_PROCESSING_PRIVATE_KEY_SECONDARY}
-    secondaryPublicCertificate: ${APPLE_PAY_PAYMENT_PROCESSING_CERTIFICATE_SECONDARY}
+    secondaryPrivateKey: ${APPLE_PAY_PAYMENT_PROCESSING_PRIVATE_KEY_SECONDARY:-}
+    secondaryPublicCertificate: ${APPLE_PAY_PAYMENT_PROCESSING_CERTIFICATE_SECONDARY:-}
```
danworth, line 49: might be nice to be consistent with the default value for the primary keys but not necessary and this might be blocking things 👍

danworth (approved):
👍

src/main/resources/config/config.yaml
oswaldquek: Cheers. The secondary keys are optional so it's ok for them for them to be empty.

PR #2833, opened 2021-01-28
PP-5025: Help reduce key and cert creation errors

oswaldquek:
Creating the Apple private key and cert can be quite gnarly. The openssl
command that generates the private key, as well as the payment processing
certificate received from Apple, have multiple lines. We usually manually edit
these files to make them a single line which is what connector expects. We have
seen recently that in the process of doing this manually the lines are joined
with a space in between them. This commit helps with the fiddliness of creating
the private key and cert by removing any whitespaces in case they exist.

We have also seen that when storing an empty password in
https://github.com/alphagov/pay-infra/blob/master/cli/config/secrets.yml, for
example:
```
APPLE_PAY_PAYMENT_PROCESSING_CERTIFICATE_SECONDARY:       ""
```
the java fields (`secondaryPrivateKey`, `secondaryPublicCertificate`) that the
environment variable gets bound to might be an empty string rather than null.
The secondary Apple payment keycerts being optional, return nulls if the string
is blank.


src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
```diff
@@ -18,11 +21,19 @@
     private String secondaryPublicCertificate;
 
     public Optional<String> getSecondaryPublicCertificate() {
-        return Optional.ofNullable(secondaryPublicCertificate);
+        if (isNotBlank(secondaryPublicCertificate)) {
+            return Optional.of(secondaryPublicCertificate);
+        } else {
+            return Optional.empty();
+        }
```
alexbishop1, line 28: Another way to do this would be:

```java
return Optional.of(secondaryPublicCertificate).filter(not(String::isBlank));
```

(Java’s built in `String.isBlank()` and Apache Commons’ `StringUtils.isNotBlank(…)` use the same definition of white space.)

src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypter.java
```diff
@@ -83,6 +83,10 @@ public ApplePayDecrypter(WorldpayConfig worldpayConfig, ObjectMapper objectMappe
         long daysToExpiry = DAYS.between(Instant.now(), primaryCertificate.getNotAfter().toInstant());
         LOGGER.info("The Apple Pay payment processing cert will expire in {} days", daysToExpiry);
     }
+    
+    private String removeWhitespace(String input) {
+        return input.replaceAll("\\s+","");
```
alexbishop1, line 88: This is uses a different definition of white space to when we’re deciding if the certificate or key is blank in `ApplePayConfig` but that’s probably OK.

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
```diff
@@ -23,8 +23,8 @@
...
+    private static final String ENCODED_PRIMARY_PRIVATE_KEY = "MIGHAgEAMBMGB yqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgOqukXQoQVbg1xvvq/IGLdK0UuJPbbawULTALcuw/Uz2hRANCAAQPjiA1kTEodST2wy5d5kQFrM0D5qBX9Ukry8W6D+vC7OqbMoTm/upRM1GRHeA2LaVTrwAnpGhoO0ETqYF2Nu4V";
```
alexbishop1, line 26: Is there a reason this changed?

alexbishop1 (approved):
Good way to solve a paper cut.

src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
oswaldquek: Ah Alex, you're the king when it comes to this :)

src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayDecrypterTest.java
oswaldquek: I've added whitespace in these to prove that we can still decrypt properly

src/main/java/uk/gov/pay/connector/app/ApplePayConfig.java
oswaldquek: What about `Optional.ofNullable(secondaryPublicCertificate).filter(not(String::isBlank))`. Its null when spinning up the pay local docker card cluster.

oswaldquek: I'll do this in a new PR as we're wanting to test now

alexbishop1: Sorry, I meant `Optional.ofNullable(…)`!

PR #2840, opened 2021-02-02
PP-7672 Send IP address to Smartpay during authorisation

kbottla:
## WHAT YOU DID
- Sends payer IP address, if available, to Smartpay when the flag (send_payer_ip_address_to_gateway) on gateway account is enabled.


stephencdaly (approved):
LGTM

PR #2853, opened 2021-02-19
PP-7572 Use unique WireMock port per test class

stephencdaly:
Avoid port conflicts by using a new port number per integration test class. This should mean the tests running in parallel won't interfere with each other.

Achieve this by creating a `WireMockServer` with a unique port per instance of `DropwizardJUnitRunner` that is tied to the lifecycle of the runner. Access this in tests for setting up stubs by adding it to the `DropwizardTestContext`.


SandorArpa (approved):
Nice work, LOKTM

PR #2854, opened 2021-02-19
PP-7572 Use unique WireMock port when using app rule

stephencdaly:
Have a unique WireMock port per instance of AppWithPostgresRule and use this port when initialising the WireMockRule in test classes.

Move the standard url config overrides which need to be set up to use the WireMock port into the AppWithPostgresRule setup so these don't need to be set up per test instance.


kbottla (approved):
LGTM!

PR #2862, opened 2021-02-25
PP-7836 Log livemode for Stripe account notifications

kbottla:
## WHAT YOU DID
- Log `livemode` field from Stripe account notification so we can exclude test account events from alerts

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
@@ -30,7 +30,8 @@ public class StripeAccountUpdatedHandler {
             LOGGER.info(String.format("Received an account.updated event for stripe account %s", dataObject.getId()),
                     kv("stripe_account_id", dataObject.getId()),
                     kv("payouts_enabled", dataObject.isPayoutsEnabled()),
-                    kv("requirements", dataObject.getRequirements()));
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
```diff
@@ -30,7 +30,8 @@ public void process(StripeNotification notification) {
             LOGGER.info(String.format("Received an account.updated event for stripe account %s", dataObject.getId()),
                     kv("stripe_account_id", dataObject.getId()),
                     kv("payouts_enabled", dataObject.isPayoutsEnabled()),
-                    kv("requirements", dataObject.getRequirements()));
+                    kv("requirements", dataObject.getRequirements()),
+                    kv("livemode", notification.getLivemode()));
```
stephencdaly, line 34: could we just not log this line if the livemode is test to save having to update the alert? or do we think this line is still useful?

stephencdaly (approved):
Seems fine, just questioned whether we need to log the line at all for test

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
kbottla: I think it will be useful to log for test accounts to test changes/new events/searches

PR #2873, opened 2021-03-23
bau: Log gateway account id when sending emails

oswaldquek:
This is useful information to have in the logs. Current log in splunk looks like

```
{
   @timestamp: 2021-03-23T17:10:13.960Z
   @version: 1
   container: connector
   environment: production-2
   level: INFO
   level_value: 20000
   logger_name: uk.gov.pay.connector.usernotification.service.UserNotificationService
   message: Sending PAYMENT_CONFIRMED email, charge_external_id=xxxx
   thread_name: pool-48-thread-29
}
```

### DIFF
--- a/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
@@ -6,6 +6,7 @@ import io.dropwizard.setup.Environment;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
```diff
@@ -91,6 +93,7 @@ public UserNotificationService(NotifyClientFactory notifyClientFactory, Connecto
 
         Stopwatch responseTimeStopwatch = Stopwatch.createStarted();
         return executorService.submit(() -> {
+            MDC.put(GATEWAY_ACCOUNT_ID, gatewayAccountEntity.getId().toString());
```
stephencdaly: Do we need to remove from the MDC after we put it on?
Also, is it worth logging the `payment_external_id` as a structured arg too?

oswaldquek: Actually, maybe better to use structured args here in that case.

oswaldquek: I've actually gone with the `MDC.put` as there is a `logger.error` and unfortunately there is no way to use structure args there.

stephencdaly (approved):

PR #2874, opened 2021-03-24
bau: Use proper xml escaping

oswaldquek:
Prevent a malicious paying user potentially injecting XML into the auth request to Worldpay.

### DIFF
--- a/src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
+++ b/src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
@@ -48,7 +48,7 @@
             </shopper>
             <#if authCardDetails.worldpay3dsFlexDdcResult.isPresent()>
             <additional3DSData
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -48,7 +48,7 @@
             </shopper>
             <#if authCardDetails.worldpay3dsFlexDdcResult.isPresent()>
             <additional3DSData
-                dfReferenceId="${authCardDetails.worldpay3dsFlexDdcResult.get()}"
+                dfReferenceId="${authCardDetails.worldpay3dsFlexDdcResult?xml}"
```
alexbishop1: This still needs `.get()` because `worldpay3dsFlexDdcResult` is an `Optional`. I think it should be `${authCardDetails.worldpay3dsFlexDdcResult.get()?xml}`.

alexbishop1 (changes requested):

alexbishop1 (approved):

PR #2877, opened 2021-03-26
PP-7938 Move Package Coordinates

rorymalcolm:
- As part of the migration from Bintray to Maven Central we must move our coordinates to match the domain we own

- This means that the coordinates of our java packages will now exist under `uk.gov.service.payments` instead of `uk.gov.pay`

- This moves the dependencies of pay-java-commons to maven central from bintray


pom.xml
```diff
@@ -44,19 +44,19 @@
             <snapshots>
                 <enabled>false</enabled>
             </snapshots>
-            <id>bintray-govuk-pay-pay-java-commons</id>
-            <name>bintray</name>
-            <url>https://dl.bintray.com/govuk-pay/pay-java-commons</url>
+            <id>maven-central-govuk-pay-pay-java-commons</id>
+            <name>mavencentral</name>
+            <url>https://repo1.maven.org/maven2/uk/gov/service/payments/</url>
```
oswaldquek: Oh do we actually need this? I thought we could find it in the standard maven repo (https://mvnrepository.com/) but looks like no.

rorymalcolm: I am unsure if that site is actually properly maven central - the links to the central repository they use follow this syntax

oswaldquek: I removed it locally and it works so let's remove it?

pom.xml
```diff
@@ -17,7 +17,7 @@
         <eclipselink.version>2.7.7</eclipselink.version>
         <guice.version>5.0.1</guice.version>
         <jackson.version>2.12.2</jackson.version>
-        <pay-java-commons.version>1.0.20210310094438</pay-java-commons.version>
+        <pay-java-commons.version>1.0.1</pay-java-commons.version>
```
oswaldquek: Let's use the latest `1.0.20210326162042` here?

pom.xml
```diff
@@ -40,23 +40,15 @@
...
-            <url>https://dl.bintray.com/govuk-pay/pay-java-commons</url>
-        </repository>
     </repositories>
     <dependencies>
         <dependency>
-            <groupId>uk.gov.pay</groupId>
+            <groupId>uk.gov.service.payments</groupId>
             <artifactId>logging</artifactId>
-            <version>${pay-java-commons.version}</version>
+            <version>1.0.1</version>
```
oswaldquek: This needs to be ${pay-java-commons.version}

oswaldquek (approved):

PR #2898, opened 2021-04-14
PP-7870 Add Healthcheck Resource Logging

rorymalcolm:
- Add logging for the Healthcheck resource at error level


danworth (approved):
This will help us see why health checks fail which would be useful in the performance test environment right now (hence the change). Error log level makes sense because if a health check fails it's something that should be investigated with priority.

PR #2930, opened 2021-05-17
PP-7919: Create table to test concourse DB migration

katstevens:
## WHAT YOU DID
Similar to how we tested the DB migration job in Ledger, create a dummy table in Connector so we can try out a 'real' DB migration, rather than just a no-op. 

Once this test has been complete, we'll add a second migration to `dropTable` again.

## How to review

This is my first Connector PR and first Connector DB migration - please shout if I've missed something, or if this PR is a terrible idea.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1577,4 +1577,12 @@
             <column name="allow_telephone_payment_notifications" value="false"/>
         </update>
     </changeSet>
+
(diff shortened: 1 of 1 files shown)


katstevens:
Checked this migration locally by running `mvn clean verify` and inspecting the table in the `postgres` Docker container.

oswaldquek (approved):

PR #2935, opened 2021-05-18
PP-7988 Document SQS Queues in README

rorymalcolm:
- Document the queues in a new section away from the environment variables, providing more context on the operation of each and what they're used for.


katstevens (approved):
Love it 👍

PR #2950, opened 2021-05-27
PP-8128 - Add payment_provider column to charges table

nimalank7:
Description:
- PR adds payment_provider column to charges table
- Future JIRA story will update charges with a payment_provider value


kbottla (approved):
LGTM!

PR #2952, opened 2021-06-01
PP-8130 Add service for checking Worldpay credentials

stephencdaly:
- Service sends an inquiry order with an order code that won't be found.
- We know that credentials are valid if a 200 is returned with error code 5, indicating the payment wasn't found.
- If a 401 or a 200 with error code 4 is returned we know that the credentials are invalid.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
@@ -1,0 +1,85 @@
+package uk.gov.pay.connector.gateway.worldpay;
+
+import com.google.inject.Inject;
+import io.dropwizard.setup.Environment;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.gateway.GatewayClient;
+import uk.gov.pay.connector.gateway.GatewayClientFactory;
+import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.GatewayOrder;
+import uk.gov.pay.connector.gateway.worldpay.exception.NotAWorldpayGatewayAccountException;
+import uk.gov.pay.connector.gateway.worldpay.exception.UnexpectedValidateCredentialsResponse;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.model.WorldpayCredentials;
+
+import javax.inject.Named;
+import java.net.URI;
+import java.util.Map;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.gateway.GatewayResponseUnmarshaller.unmarshallResponse;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
+import static uk.gov.pay.connector.gateway.util.AuthUtil.getWorldpayCredentialsCheckAuthHeader;
+import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderRequestBuilder.aWorldpayInquiryRequestBuilder;
+
+public class WorldpayCredentialsValidationService implements WorldpayGatewayResponseGenerator {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(WorldpayCredentialsValidationService.class);
+
+    private final GatewayClient gatewayClient;
+    private final Map<String, URI> gatewayUrlMap;
+
+    @Inject
+    public WorldpayCredentialsValidationService(@Named("WorldpayGatewayUrlMap") Map<String, URI> gatewayUrlMap,
+                                                GatewayClientFactory gatewayClientFactory,
+                                                Environment environment) {
+        this.gatewayUrlMap = gatewayUrlMap;
+        gatewayClient = gatewayClientFactory.createGatewayClient(WORLDPAY, environment.metrics());
+    }
+
+    public boolean validateCredentials(GatewayAccountEntity gatewayAccountEntity, WorldpayCredentials worldpayCredentials) {
+        if (!gatewayAccountEntity.getGatewayName().equals(WORLDPAY.getName())) {
+            throw new NotAWorldpayGatewayAccountException(gatewayAccountEntity.getId());
+        }
+
+        GatewayOrder order = aWorldpayInquiryRequestBuilder()
+                .withTransactionId("an-order-id-that-will-not-exist")
+                .withMerchantCode(worldpayCredentials.getMerchantId())
+                .build();
+
+        try {
+            GatewayClient.Response response = gatewayClient.postRequestFor(
+                    gatewayUrlMap.get(gatewayAccountEntity.getType()),
+                    gatewayAccountEntity,
+                    order,
+                    getWorldpayCredentialsCheckAuthHeader(worldpayCredentials)
+            );
+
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
```diff
@@ -0,0 +1,79 @@
...
+        if (!gatewayAccountEntity.getGatewayName().equals(WORLDPAY.getName())) {
+            throw new NotAWorldpayGatewayAccountException(gatewayAccountEntity.getId());
+        }
+
+        GatewayOrder order = aWorldpayInquiryRequestBuilder()
+                .withTransactionId("an-order-id-that-will-not-exist")
+                .withMerchantCode(worldpayCredentials.getMerchantId())
+                .build();
+
+        GatewayClient.Response response = gatewayClient.postRequestFor(
```
kbottla: Can we catch the exception thrown by `GatewayClient` ? Looks like GatewayClient only returns response if successful otherwise throws exception https://github.com/alphagov/pay-connector/blob/62fae47b55df2447c8abb856a634cc5aaa944948/src/main/java/uk/gov/pay/connector/gateway/GatewayClient.java#L73-L84

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
```diff
@@ -0,0 +1,79 @@
...
+        // - a 200 with error code 4 (security violation) if merchant id is incorrect
+        if (response.getStatus() == 401) {
+            return false;
+        }
+        if (response.getStatus() == 200) {
+            WorldpayQueryResponse worldpayQueryResponse = unmarshallResponse(response, WorldpayQueryResponse.class);
+            switch (worldpayQueryResponse.getErrorCode()) {
+                case "5":
+                    return true;
+                case "4":
```
kbottla, line 68: Shall we make every case (other than "5") `false`. Given we don't have exhaustive list of error codes, we may throw exception for cases we are not aware of.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
stephencdaly: Yep good idea, when coming on to do the resource it probably makes sense to wrap the exception in the service. Will update.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
stephencdaly: I was wondering this but was also thinking if we get a different error we might not be sure if it means that the credentials are invalid and we might want to investigate, hence logging an error.

We do get a list of error codes here and it does seem like 
error code 4 is what we should expect for invalid merchant id https://developer.worldpay.com/docs/wpg/troubleshoot

stephencdaly: Also don't have a strong preference, if something goes wrong we might get told through support before we find out from Sentry

kbottla: Fine to log error and throw exception as we are not expecting anything other than code 4/5 or http status 401.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
stephencdaly: updated

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
```diff
@@ -0,0 +1,83 @@
...
+            );
+
+            // There is no Worldpay endpoint to explicitly check credentials, so we make a query for an non-existent 
+            // transaction and expect:
+            // - a 401 if username/password are incorrect
+            // - a 200 with error code 5 (transaction not found) in the response if all credentials are correct
+            // - a 200 with error code 4 (security violation) if merchant id is incorrect
+            if (response.getStatus() == 401) {
+                return false;
+            }
```
kbottla: should this be in catch block as `GatewayErrorException` is thrown for client errors ? Current handling by AuthoriseHandler https://github.com/alphagov/pay-connector/blob/21fcbe8aa367bd7f238813c324476c2c27dc4c1b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandler.java#L62-L64

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
```diff
@@ -0,0 +1,83 @@
...
+            if (response.getStatus() == 200) {
```
kbottla: We may not need to for 200 status as exception will be thrown for status other than 200

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
stephencdaly: Ah yes good spot, I think you're right

stephencdaly: updated

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
```diff
@@ -0,0 +1,85 @@
...
+                    return true;
+                case "4":
+                    return false;
+                default:
+                    LOGGER.error(format("Unexpected error code %s returned by Worldpay when validating credentials", worldpayQueryResponse.getErrorCode()));
+                    throw new UnexpectedValidateCredentialsResponse();
+            }
+        } catch (GatewayException.GatewayErrorException e) {
+            return e.getStatus().map(status -> {
+                if (status == 401) {
```
kbottla, line 76: (minor)
```suggestion
                if (status == UNAUTHORIZED) {
```

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
```diff
@@ -0,0 +1,85 @@
...
+                    return false;
+                }
+                throw new UnexpectedValidateCredentialsResponse();
+            }).orElseThrow(UnexpectedValidateCredentialsResponse::new);
+        } catch (GatewayException e) {
+            throw new UnexpectedValidateCredentialsResponse();
```
kbottla, line 82: (minor) Not sure if the exceptions are logged in GatewayClient. If not logged already, we could log at info level for support.

kbottla (approved):
Looks good to me! few minor comments on logging or using constant for error code but happy for them to be addressed in separate PR

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCredentialsValidationService.java
stephencdaly: Looking at it, the GatewayClient will already log sufficient detail when there's a non-success response

PR #2955, opened 2021-06-02
PP-8130 Add resource to validate worldpay credentials

stephencdaly:
Rename and use existing GatewayAccount3dsFlexCredentialsResource to add Worldpay credentials validation endpoint.

Register an additional singleton GatewayClient for validating credentials as a unique key is required for the metrics registry.


SandorArpa (approved):
LGTM

PR #2959, opened 2021-06-03
PP-8123 Set active_start_date for new & active credentials

kbottla:
## WHAT YOU DID
- For sandbox or stripe accounts credentials set to ACTIVE state, set `active_start_date` immediately.
- For stripe credentials, if an active credentials record exists, set state to ENTERED - this is specific to switching PSP and the state to be set to ACTIVE when service switches PSP

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -8,12 +8,14 @@ import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCreden
 import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialsEntity;
 
 import javax.inject.Inject;
+import java.time.Instant;
 import java.util.Map;
 
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.SANDBOX;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.STRIPE;
 import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.ACTIVE;
 import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.CREATED;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.ENTERED;
 
 public class GatewayAccountCredentialsService {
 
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -27,20 +29,33 @@ public GatewayAccountCredentialsService(GatewayAccountCredentialsDao gatewayAcco
...
+        GatewayAccountCredentialsEntity gatewayAccountCredentialsEntity;
+
+        if (state == ACTIVE) {
+            gatewayAccountCredentialsEntity
+                    = new GatewayAccountCredentialsEntity(gatewayAccountEntity, paymentProvider, credentials,
+                    state, Instant.now());
+        } else {
+            gatewayAccountCredentialsEntity
+                    = new GatewayAccountCredentialsEntity(gatewayAccountEntity, paymentProvider, credentials, state);
+        }
```
stephencdaly, line 42: maybe using a setter for the active start date would be a tiny bit more concise? but no problem with this

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -27,20 +29,33 @@ public GatewayAccountCredentialsService(GatewayAccountCredentialsDao gatewayAcco
...
         if (paymentGatewayName == SANDBOX) {
             return ACTIVE;
         }
-        // todo: state should be ENTERED if another active credentials record exists (when switching PSP)
         if (paymentGatewayName == STRIPE && !credentials.isEmpty()) {
+            boolean hasActiveCredentials =
+                    gatewayAccountCredentialsDao.hasActiveCredentials(gatewayAccountEntity.getId());
+            if (hasActiveCredentials){
+                return ENTERED;
+            }
```
sfount, line 58: there's something a bit confusing about this logic but I can't quite put my finger on it

maybe splitting it into 
- when to make it immediate active (grouped with the above, stripe with creds or sandbox)
- when to make it entered or created (with creds or without creds)

not precisely sure what that looks like

sfount (approved):
LGTM 👍

stephencdaly (approved):
LGTM

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
kbottla: Could change to something like below if it reads a bit better (linter complains about ternary operator though) ?. With grouping, ending up with about same number of paths. 
```
 PaymentGatewayName paymentGatewayName = PaymentGatewayName.valueFrom(paymentProvider);
  if (paymentGatewayName == SANDBOX) {
      return ACTIVE;
  }
  if (paymentGatewayName == STRIPE && !credentials.isEmpty()) {
      boolean hasAnotherActiveCredential =
              gatewayAccountCredentialsDao.hasActiveCredentials(gatewayAccountEntity.getId());

      return hasAnotherActiveCredential ? ENTERED : ACTIVE;
  }

  return CREATED;
```

kbottla: Grouping looks like this
```
        PaymentGatewayName paymentGatewayName = PaymentGatewayName.valueFrom(paymentProvider);
        if (paymentGatewayName == SANDBOX
                || (paymentGatewayName == STRIPE && !credentials.isEmpty()
                && !gatewayAccountCredentialsDao.hasActiveCredentials(gatewayAccountEntity.getId()))) {
            return ACTIVE;
        }

        if (paymentGatewayName == STRIPE && !credentials.isEmpty()) {
            return ENTERED;
        }

        return CREATED;
```

sfount: Yeah that makes sense, I don't think the second group needs to just be stripe

it's like

Base case, if its a sandbox or the first Stripe creds: `ACTIVE`
If there are credentials: `ENTERED`
Else: `CREATED`

It just so happens that Stripe will be the only one for now that provides credentials with the request -- other Pay provisioned PSPs will do the same so I think we can remove Stripe for the second condition

sfount:
```java 
var isFirstCredentials = !gatewayAccountCredentialsDao.hasActiveCredentials(gatewayAccountEntity.getId());
var credentialsPrePopulated = !credentials.isEmpty();

if ((isFirstCredentials && credentialsPrePopulated) || paymentGatewayName == SANDBOX) {
  return ACTIVE;
}

return credentialsPrePopulated ? ENTERED : CREATED;
```

kbottla: Was trying to avoid querying DB for every create op. Probably fine given it is not resource intensive.

PR #2960, opened 2021-06-03
Update gateway_account_credentials when credentials patched

stephencdaly:
This is to ensure that after back-filling the gateway_account_credentials table, any updates to credentials made by users in the admin tool are propagated to the new table.

Once we have switched over to fully using the new table to get credentials and the new endpoint to update credentials has been added, this code will be discarded.


SandorArpa (approved):
LGTM

PR #2973, opened 2021-06-08
PP-8198 Add common validator for patch requests

stephencdaly:
Validation of JSON patch request body following the JSON patch spec http://jsonpatch.com/:
- Checks the JSON is an array
- Checks that "op", "path" and "value" are present
- Checks that "op" and "path" are strings
- Checks that "op" is one of the allowed values.

Validator takes a map of validators to apply to each allowed op/path combination.
It will check that the op/path combination in the request is one of the allowed combinations and then apply the Consumer provided for the validation of the operation.


SandorArpa (approved):
LOKTM

PR #2974, opened 2021-06-08
PP-8223 add external id to gateway_account_credentials table

SandorArpa:
## WHAT YOU DID
-  add new column, external_id to gateway_account_credentials table


stephencdaly (approved):
LGTM

PR #2980, opened 2021-06-09
PP-8223 add external_id to gateway_account_credentials entity

SandorArpa:
## WHAT YOU DID
- add externalId to GatewayAccountCredentialsEntity
- update GatewayAccountCredentialsService accordingly
- update tests


src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -39,6 +40,7 @@ public void createGatewayAccountCredentials(GatewayAccountEntity gatewayAccountE
         if (state == ACTIVE) {
             gatewayAccountCredentialsEntity.setActiveStartDate(Instant.now());
         }
+        gatewayAccountCredentialsEntity.setExternalId(randomUuid());
```
stephencdaly, line 43: Do we need to set this? Will it not be generated by the default value on the column?

SandorArpa: It seems it won't. Manual insertion triggers the default value, persist does not. If we don't set it then it throws an SQL error that column cannot be empty. Same as GatewayAccount external_id.

stephencdaly (approved):
LGTM

PR #2981, opened 2021-06-09
PP-8198 Add patch gateway account credentials endpoint

stephencdaly:
Add patch endpoint to update gateway account credentials by id.

Add service that allows for updating the credentials column on the entity when a replace operation is included in the patch request.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/exception/GatewayAccountCredentialsNotFoundException.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/exception/GatewayAccountCredentialsNotFoundException.java
@@ -1,0 +1,13 @@
+package uk.gov.pay.connector.gatewayaccount.exception;
+
+import javax.ws.rs.WebApplicationException;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.util.ResponseUtil.notFoundResponse;
+
+public class GatewayAccountCredentialsNotFoundException extends WebApplicationException {
+    
+    public GatewayAccountCredentialsNotFoundException(Long id) {
+        super(notFoundResponse(format("Gateway account credentials with id [%s] not found.", id)));
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -1,7 +1,11 @@
 package uk.gov.pay.connector.gatewayaccountcredentials.service;
 
 import com.google.inject.persist.Transactional;
+import uk.gov.pay.connector.common.model.api.jsonpatch.JsonPatchOp;
+import uk.gov.pay.connector.common.model.api.jsonpatch.JsonPatchRequest;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
+import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountCredentialsNotFoundException;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountCredentials;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.gatewayaccountcredentials.dao.GatewayAccountCredentialsDao;
 import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState;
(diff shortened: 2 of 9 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/exception/GatewayAccountCredentialsNotFoundException.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.pay.connector.gatewayaccount.exception;
+
+import javax.ws.rs.WebApplicationException;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.util.ResponseUtil.notFoundResponse;
+
+public class GatewayAccountCredentialsNotFoundException extends WebApplicationException {
```
kbottla, line 8: Shall we extend `NotFoundException` to return 404 instead of 500 when credential not found

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -86,4 +98,21 @@ public void updateGatewayAccountCredentialsForLegacyEndpoint(GatewayAccountEntit
...
 
+    @Transactional
+    public GatewayAccountCredentials updateGatewayAccountCredentials(long gatewayAccountCredentialsId, List<JsonPatchRequest> updateRequests) {
+        return getGatewayAccountCredentials(gatewayAccountCredentialsId).map(
+                gatewayAccountCredentialsEntity -> {
+                    for (JsonPatchRequest updateRequest : updateRequests) {
+                        if (updateRequest.getPath().equals(FIELD_CREDENTIALS) && updateRequest.getOp() == JsonPatchOp.REPLACE) {
+                            gatewayAccountCredentialsEntity.setCredentials(updateRequest.valueAsObject());
+                        } else {
+                            throw new UnsupportedOperationException(format("Operation [%s] for path [%s] is not supported",
```
kbottla: Can we validate the operations allowed in validator ? or is there any reason this is in service ?

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountCredentialsResourceIT.java
```diff
@@ -446,16 +454,94 @@ public void createGatewayAccountsCredentialsValidatesRequestBusinessLogic_respon
...
+
+        Map<String, Object> updatedGatewayAccountCredentials = databaseTestHelper.getGatewayAccountCredentialsById(credentialsId);
+        Map<String, String> updatedCredentials = new Gson().fromJson(((PGobject)updatedGatewayAccountCredentials.get("credentials")).getValue(), Map.class);
+        assertThat(updatedCredentials, hasEntry("username", "new-username"));
+        assertThat(updatedCredentials, hasEntry("password", "new-password"));
+        assertThat(updatedCredentials, hasEntry("merchant_id", "new-merchant-id"));
+    }
+
+    @Test
+    public void patchGatewayAccountCredentialsInvalidRequestBody_shouldReturn400() {
```
kbottla, line 487: Would be good to add resource tests for different edge cases to reduce the number of IT tests if possible - similar to https://github.com/alphagov/pay-ledger/blob/f914e39ce6e0ce8c40ff6f3e77dcf7e9cae07556/src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceTest.java#L37

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/GatewayAccountCredentialsNotFoundException.java
stephencdaly: it does return 404 by virtue of constructing the `WebApplicationException` using the `notFoundResponse` helper method, which is a pattern we use a lot in connector that I followed.

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
stephencdaly: Yeh we are already validating the operations, this can probably be removed actually. I was doing this for extra safety but we should be able to rely on the tests to ensure no unsupported operation gets through.

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountCredentialsResourceIT.java
stephencdaly: Can I add another pull request to move this test? There might be some other tests in this class that could be moved over

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
stephencdaly: updated

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountCredentialsResourceIT.java
kbottla: yeah happy to be in a followup PR

kbottla (approved):
LGTM!

PR #2982, opened 2021-06-09
PP-8125 - Add method to return payment provider name from gateway_account_credential

nimalank7:
Description:
- If there is only 1 GatewayAccountCredential then use that regardless of status otherwise get the latest ACTIVE GatewayAccountCredential
- To avoid changing too much the getGatewayNameFromGatewayAccountCredential() was added to allow it to be unit tested properly.
The next PR will then rename that method to getGatewayName(). This was done to avoid changing all other tests which rely on getGatewayName() in a single PR.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
@@ -32,16 +32,23 @@ import javax.persistence.OneToMany;
 import javax.persistence.OneToOne;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
+import javax.ws.rs.WebApplicationException;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import static com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL;
 import static com.google.common.collect.Lists.newArrayList;
+import static java.lang.String.format;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType.LIVE;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.ACTIVE;
+import static uk.gov.pay.connector.util.ResponseUtil.serviceErrorResponse;
 
 @Entity
 @Table(name = "gateway_accounts")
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -183,6 +187,22 @@ public String getGatewayName() {
         return gatewayName;
     }
 
+    public String getGatewayNameFromGatewayAccountCredentials() {
+        List<GatewayAccountCredentialsEntity> gatewayAccountCredentialsEntities = getGatewayAccountCredentials();
+        if (gatewayAccountCredentialsEntities.size() == 1) {
+            GatewayAccountCredentialsEntity entity = gatewayAccountCredentialsEntities.get(0);
+            return entity.getPaymentProvider();
```
kbottla, line 198: could simplify a bit 
```suggestion
            return gatewayAccountCredentialsEntities.get(0).getPaymentProvider();
```

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -183,6 +187,22 @@ public String getGatewayName() {
...
+        }
+
+        List<GatewayAccountCredentialsEntity> sortedGatewayAccountCredentialsEntities = gatewayAccountCredentialsEntities.stream()
+                .filter(entity -> entity.getState() == ACTIVE)
+                .sorted(Comparator.comparing(GatewayAccountCredentialsEntity::getActiveStartDate))
+                .collect(Collectors.toList());
+        Collections.reverse(sortedGatewayAccountCredentialsEntities);
+
+        return sortedGatewayAccountCredentialsEntities.get(0).getPaymentProvider();
```
alexbishop1:
```java
return gatewayAccountCredentialsEntities.stream()
        .filter(entity -> entity.getState() == ACTIVE)
        .max(Comparator.comparing(GatewayAccountCredentialsEntity::getActiveStartDate))
        .get() // This could theoretically cause an exception…
        .getPaymentProvider();
```
😀

nimalank7: Ha that is a lot better than my method!

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -183,6 +187,20 @@ public String getGatewayName() {
...
+        if (gatewayAccountCredentialsEntities.size() == 1) {
+            GatewayAccountCredentialsEntity entity = gatewayAccountCredentialsEntities.get(0);
+            return entity.getPaymentProvider();
+        }
+
+        return gatewayAccountCredentialsEntities.stream()
+                .filter(entity -> entity.getState() == ACTIVE)
+                .max(Comparator.comparing(GatewayAccountCredentialsEntity::getActiveStartDate))
+                .get()
+                .getPaymentProvider();
```
kbottla: Can we throw exception instead of `get()` and returning ?
If there are no credentials at all or no creds with ACTIVE state this will throw Index out of bounds exception

```suggestion
Optional<GatewayAccountCredentialsEntity> mayBeGatewayAccountCredential = gatewayAccountCredentialsEntities
                .stream()
                .filter(entity -> entity.getState() == ACTIVE)
                .max(Comparator.comparing(GatewayAccountCredentialsEntity::getActiveStartDate));

        return mayBeGatewayAccountCredential.map(GatewayAccountCredentialsEntity::getPaymentProvider)
                .orElseThrow(() -> new WebApplicationException(serviceErrorResponse(
                        format("Credentials not found for gateway account [%s]", getId()))));
```

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntityTest.java
```diff
@@ -0,0 +1,78 @@
...
+                Map.of(),
+                GatewayAccountCredentialState.ACTIVE
+        ));
+
+        assertThat(gatewayAccountEntity.getGatewayNameFromGatewayAccountCredentials(), is("sandbox"));
+    }
+
+    @Test
+    void shouldReturnStripeAsPaymentProviderNameForLatestGatewayAccountCredentialEntity() {
+        GatewayAccountCredentialsEntity latestActiveGatewayAccountCredential = new GatewayAccountCredentialsEntity(
```
kbottla: Can we use fixtures which comes with most of the values as defaults ? 
```suggestion
        GatewayAccountCredentialsEntity credentialsEntity = aGatewayAccountCredentialsEntity()..withPaymentProvider("stripe").build();
```

nimalank7: Done.

kbottla (approved):

PR #2984, opened 2021-06-10
PP-8191 add switch psp request validator

SandorArpa:
## WHAT YOU DID
- add GatewayAccountSwitchPaymentProviderRequest
- add validator for GatewayAccountSwitchPaymentProviderRequest
- add tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/GatewayAccountSwitchPaymentProviderRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/GatewayAccountSwitchPaymentProviderRequest.java
@@ -1,0 +1,28 @@
+package uk.gov.pay.connector.gatewayaccount;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class GatewayAccountSwitchPaymentProviderRequest {
+
+    public static final String USER_EXTERNAL_ID_FIELD = "user_external_id";
+    public static final String GATEWAY_ACCOUNT_CREDENTIALS_ID_FIELD = "gateway_account_credential_id";
+
+    private String userExternalId;
+    private String gatewayAccountCredentialId;
+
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountSwitchPaymentProviderRequestValidator.java
```diff
@@ -0,0 +1,24 @@
...
+import uk.gov.pay.connector.gatewayaccount.GatewayAccountSwitchPaymentProviderRequest;
+
+import java.util.Collections;
+
+import static java.lang.String.format;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
+public class GatewayAccountSwitchPaymentProviderRequestValidator {
+
+    private static final String USER_EXTERNAL_ID_FIELD = "user_external_id";
```
sfount: If these are defined as static fields on the `GatewayAccountSwitchPaymentProviderRequest` and used for the `@JsonProperty()` annotations then we can guarantee that what we're validating is the same as what's parsed in the request. 

They could then be imported by this class for correctly formatting the validation message.

src/main/java/uk/gov/pay/connector/gatewayaccount/GatewayAccountSwitchPaymentProviderRequest.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.pay.connector.gatewayaccount;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
```
sfount: I don't think this is needed as the request will never be serialised back to JSON.

sfount (approved):

PR #2985, opened 2021-06-10
PP-8198 Improve JSON patch request validator

stephencdaly:
Rename to JsonPatchRequestValidator to make it clearer we're validating patch request that follow the JSON Patch spec.

Refactor validator so that consumers for validating individual operations accept a `JsonPatchRequest` rather than a `JsonNode`, so its clearer what the consumer should be expecting to validate. By doing this improve the error message given when a value is invalid by including the `path` validation failed for.


kbottla (approved):
LGTM!

PR #2986, opened 2021-06-10
PP-8198 Accept patch operation to update last updated by user

stephencdaly:
Accept a `replace` operation for path `last_updated_by_user_external_id` for the patch gateway account credentials by id endpoint.


kbottla (approved):
LGTM!

PR #2988, opened 2021-06-10
PP-8245: Reduce number of hosted graphite metrics

oswaldquek:
We are frequently exceeding our hosted graphite metric limit. We can probably
drop the metrics at the gateway account id level, which is what this commit
does. If we do need granular metrics by gateway account ids this would normally
be done by splunk anyway.


danworth (approved):
Looks good to me. Just a reminder that we don't send application metrics to HG from the test-12 environment so you won't see the effects of these changes until it gets to staging and which point you should be able to confirm.

PR #2996, opened 2021-06-11
PP-8191 add switch payment provider service

SandorArpa:
## WHAT YOU DID
- add a new service that handles switch payment provider service request
- add tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountSwitchPaymentProviderService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountSwitchPaymentProviderService.java
@@ -1,0 +1,70 @@
+package uk.gov.pay.connector.gatewayaccount.service;
+
+import com.google.inject.persist.Transactional;
+import uk.gov.pay.connector.gatewayaccount.GatewayAccountSwitchPaymentProviderRequest;
+import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccountcredentials.dao.GatewayAccountCredentialsDao;
+import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialsEntity;
+
+import javax.inject.Inject;
+import javax.ws.rs.BadRequestException;
+import javax.ws.rs.NotFoundException;
+import java.time.Instant;
+import java.util.List;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.ACTIVE;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.RETIRED;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.VERIFIED_WITH_LIVE_PAYMENT;
+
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/exception/SwitchPaymentProviderException.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.pay.connector.gatewayaccount.exception;
+
+import javax.ws.rs.BadRequestException;
+
+public class SwitchPaymentProviderException extends BadRequestException {
```
stephencdaly: As this is a fairly generic exception, do you think there's much point in having a dedicated exception class? Would it make more sense to just throw a `new WebApplicationException(badRequestResponse("message"))` as we do in various other places?
Or just a `BadRequestException`?

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountSwitchPaymentProviderService.java
```diff
@@ -0,0 +1,70 @@
...
+        this.gatewayAccountCredentialsDao = gatewayAccountCredentialsDao;
+    }
+
+    @Transactional
+    public void switchPaymentProviderForAccount(GatewayAccountEntity gatewayAccountEntity,
+                                                GatewayAccountSwitchPaymentProviderRequest request) {
+
+        List<GatewayAccountCredentialsEntity> credentials = gatewayAccountEntity.getGatewayAccountCredentials();
+
+        if (credentials.size() != 2) {
```
stephencdaly: I don't think we should rely on only having exactly 2 credentials to future proof, I think we should only throw if there are fewer than 2.

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountSwitchPaymentProviderService.java
```diff
@@ -0,0 +1,70 @@
...
+            throw new SwitchPaymentProviderException("Account has no credential to switch to/from");
+        }
+
+        var activeCredentialEntity = credentials
+                .stream()
+                .filter(credential -> ACTIVE.equals(credential.getState())).findFirst()
+                .orElseThrow(() -> new NotFoundException(format("Account credential with id ACTIVE state not found.", request.getGatewayAccountCredentialId())));
```
stephencdaly: I'm not sure a 404 makes sense here, as it's usually used to suggest that either the endpoint wasn't found or a resource specified in the request wasn't found. I think 400 makes more sense.

stephencdaly (commented):
Looks good, just a few small comments about exceptions.

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountSwitchPaymentProviderService.java
SandorArpa: Story -> scenario 6

stephencdaly: The story says 400

stephencdaly: Also I think the error message is a bit wrong, I think " id" should be removed

SandorArpa: 🤦

stephencdaly (approved):
LGTM!

PR #2997, opened 2021-06-11
PP-8198 Fix patch response and add API docs

stephencdaly:
Return the created_date in the response, and modify the test to ensure we are returning all fields.

While modifying the test, created a database fixture for inserting gateway account credentials along with database helper method.

Added basic API docs for endpoint


kbottla (approved):
LGTM!

PR #2998, opened 2021-06-11
PP-8191 refactor to use gateway_account_credential_external_id

SandorArpa:
## WHAT YOU DID
- refactor to reflect that we use the gateway account credential's external id


kbottla (approved):
LGTM!

PR #3003, opened 2021-06-14
PP-8208 Allow patching credentials state

stephencdaly:
Accept "path": "state" in patch requests to update gateway account credentials.

Allow any valid state.

Add validation for the new update operation.


SandorArpa (approved):
LOKTM

PR #3006, opened 2021-06-15
PP-8216 Allow checking credentials for non Worldpay account

stephencdaly:
- Remove restriction that Worldpay credentials can only be checked for a gateway account that already has Worldpay as the payment provider, as we want to be able to call this endpoint when in the process of switching from another provider to Worldpay
- Refactor GatewayClient to allow passing in the gateway name and account type as strings rather than a GatewayAccountEntity, so we log with the correct gateway name when checking Worldpay credentials when the current active provider for the account isn't Worldpay.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/service/GatewayClientTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/GatewayClientTest.java
@@ -29,7 +29,7 @@ import java.net.SocketException;
 import java.net.URI;
 
 import static java.util.Collections.emptyMap;
-import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.anyLong;
 import static org.mockito.Mockito.doAnswer;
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/pay/connector/service/GatewayClientTest.java
```diff
@@ -80,7 +80,7 @@ public void setup() {
         when(mockBuilder.post(Entity.entity(orderPayload, mediaType))).thenReturn(mockResponse);
 
         when(mockGatewayAccountEntity.getGatewayName()).thenReturn("worldpay");
-        when(mockGatewayAccountEntity.getType()).thenReturn("worldpay");
+        when(mockGatewayAccountEntity.getType()).thenReturn("test");
```
sfount, line 83: 👍 :oops:

sfount (approved):
Neat solution and useful cleaning up on the gateway client tests, great stuff 👍

PR #3008, opened 2021-06-15
PP-8216 Improve logic for legacy patch credentials endpoint

stephencdaly:
Previously, we were throwing an error if there was more than one `gateway_account_credentials` record when updating credentials using the legacy patch endpoint.

However, this means that it is not possible to update existing credentials while provider switching is in progress as there will be more than one credential. Instead, look for either a single `gateway_account_credentials` record in any state or else one with state `ACTIVE` if there are multiple records to update.


sfount (approved):
Great solution to meet the requirements 👍 there might eventually be call for this kind of logic to be shared across different getters but we can consider that when refactoring and removing code that is replaced (i.e updating specific credentials by ID)

PR #3009, opened 2021-06-15
PP-8121 Handle empty POST body on create credentials

sfount:
Specify that the request body for a create gateway account credentials
should not be nullable.

In the case of an empty POST body, respond with `422 Unprocessable Entity`.


kbottla (approved):
LGTM!

PR #3010, opened 2021-06-16
PP-8189 Update state for CREATED credentials on PATCH request

stephencdaly:
When the `credentials` path is updated as part of a PATCH credentials API call, update the state of the gateway account credentials, if it is currently in CREATED state to:
- ACTIVE if they are the only credentials for the account
- ENTERED if there are multiple credentials for the account, as this implies that account is PSP switching.


kbottla (approved):
LGTM!

PR #3011, opened 2021-06-16
PP-8189 Only allow PATCHing state to VERIFIED_WITH_LIVE_PAYMENT

stephencdaly:
Amend the validation for the patch gateway account credentials endpoint to only allow the value to be "VERIFIED_WITH_LIVE_PAYMENT". All other state transitions are made internally by connector, so this is the only state that a consumer of the API should be able to request to switch to.


SandorArpa (approved):
LOKTM

PR #3013, opened 2021-06-16
PP-8250 Epdq integration tests - verify request body to PSP

kbottla:
## WHAT YOU DID
- Add new assertion (to epdq integrations tests) to test request body to PSP is as expected.
- Not using test template file, as we need to construct sha signature anyway based on all input params

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceIT.java
@@ -7,16 +7,24 @@ import com.google.common.collect.ImmutableMap;
 import io.restassured.response.ValidatableResponse;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-import uk.gov.service.payments.commons.model.ErrorIdentifier;
 import uk.gov.pay.connector.app.ConnectorApp;
+import uk.gov.pay.connector.common.model.domain.Address;
+import uk.gov.pay.connector.gateway.epdq.payload.EpdqPayloadDefinitionForNewOrder;
+import uk.gov.pay.connector.gateway.model.AuthCardDetails;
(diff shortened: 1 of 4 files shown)


SandorArpa (approved):
LOKTM

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceIT.java
```diff
@@ -172,4 +187,35 @@ public void shouldNotAuthorise_aTransactionInWaitingState() {
...
+        AuthCardDetails authCardDetails = AuthCardDetailsFixture.anAuthCardDetails()
+                .withCardHolder("Scrooge McDuck")
+                .withCardNo("4444333322221111")
+                .withCvc("123")
+                .withEndDate(CardExpiryDate.valueOf("11/99"))
+                .withAddress(address)
+                .build();
+        epdqPayloadDefinition.setAuthCardDetails(authCardDetails);
+
+        return epdqPayloadDefinition.createGatewayOrder().getPayload();
```
stephencdaly, line 219: Probably not any way around using this production code to build the expected payload, this is unit tested so I reckon it's fine

stephencdaly (approved):
Does what we need, which is making sure all the details we expect to be going into the ePDQ request are pulled in from the right places, so should boost our confidence when changing these things 👍

src/test/java/uk/gov/pay/connector/it/resources/epdq/EpdqCardResourceIT.java
kbottla: Yeah we will have to replicate all functionality in test to build payload (which is possible), but as `createGatewayOrder` is unit tested used the same for first pass. IT tests should now detect if payload is not set with expected data.

PR #3015, opened 2021-06-16
PP-8216 Add pact state for PATCHing credentials

stephencdaly:
Add pact state to be used by selfservice contract test to update credentials by id.


kbottla (approved):
LGTM!

PR #3021, opened 2021-06-17
PP-8231 update merchant id on gateway account credentials

SandorArpa:
## WHAT YOU DID
- when patching credentials/gateway_merchant_id also patch the merchant id on gateway account credentials
- add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -13,6 +13,7 @@ import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCreden
 import javax.inject.Inject;
 import javax.ws.rs.WebApplicationException;
 import java.time.Instant;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -31,6 +32,7 @@ import static uk.gov.pay.connector.util.ResponseUtil.serviceErrorResponse;
 
 public class GatewayAccountCredentialsService {
 
+    private static final String GATEWAY_MERCHANT_ID = "gateway_merchant_id";
     private final GatewayAccountCredentialsDao gatewayAccountCredentialsDao;
 
     @Inject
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -95,6 +97,21 @@ public void updateGatewayAccountCredentialsForLegacyEndpoint(GatewayAccountEntit
         }
     }
 
+    @Transactional
+    public void updateGatewayAccountCredentialMerchantId(GatewayAccountEntity gatewayAccountEntity, String merchantAccountId) {
+        List<GatewayAccountCredentialsEntity> credentialsEntities = gatewayAccountEntity.getGatewayAccountCredentials();
+        if (credentialsEntities.isEmpty()) {
+            // Backfill hasn't been run for this gateway account, no need to add to gateway_account_credentials table
+            // as row will be added when backfill is run.
+            return;
```
stephencdaly, line 105: Backfill has been run now, but probably fine to leave like this as this will be removed once the new way of updating this has been added.

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -94,6 +96,21 @@ public void updateGatewayAccountCredentialsForLegacyEndpoint(GatewayAccountEntit
...
+        }
+
+        var credentialsEntity = getSingleOrActiveGatewayAccountCredential(gatewayAccountEntity);
+        Map<String, String> credentials = new HashMap<>(credentialsEntity.getCredentials());
+        credentials.put(GATEWAY_MERCHANT_ID, merchantAccountId);
+        credentialsEntity.setCredentials(credentials);
```
stephencdaly, line 111: Do we not need to call `merge` here because the gateway account credentials are lazily loaded once we're in the `@Transactional` method? Is there any risk this wouldn't get persisted if somehow they were loaded before getting in to the scope of this method (although should never happen?)

stephencdaly (dismissed):
Looks good to me!

stephencdaly (approved):

PR #3022, opened 2021-06-17
PP-8189 Patch credential validation use credentials payment provider

sfount:
Use the credentials payment provider when patching credentials to validate
the request rather than the top level gateway account provider.

Now that the resource code guarantees the existence of the credential
(or handles the missing failure), simplify the service to accept the
entity and no longer handle the DAO missing case.


stephencdaly (approved):
👍

PR #3024, opened 2021-06-18
PP-8221 Allow for payment provider to be specified when creating charge

stephencdaly:
- Optional `payment_provider` field on create charge request accepted.
- If field is present, verify there are credentials with payment provider for gateway account that have state `ENTERED`,
`VERIFIED_WITH_LIVE_PAYMENT` or `ACTIVE`.
- Set provided payment provider on charge to be used for gateway requests for this charge.


kbottla (approved):
LGTM!

PR #3025, opened 2021-06-18
PP-8255 Add send payer email to gateway flag on gateway account

alexbishop1:
Add a `sendPayerEmailToGateway` boolean to `GatewayAccountEntity`, mapped to the existing `send_payer_email_to_gateway` column in the `gateway_accounts` table of the database (defaults to `false`).

Include an appropriate `"send_payer_email_to_gateway"` JSON property in responses to requests made to `/v1/frontend/accounts/ACCOUNT_ID`.

Make it so that the flag can be toggled by sending a `PATCH` request to `/v1/api/accounts/ACCOUNT_ID` with a JSON payload like the following:

```json
{
  "op": "replace",
  "path": "send_payer_email_to_gateway",
  "value": true
}
```

While the flag is exposed and can be toggled, it has no effect on the business logic yet.


kbottla (approved):
LGTM!

PR #3028, opened 2021-06-21
PP-8250 Worldpay tests - verify request body to PSP

kbottla:
## WHAT YOU DID
- Added new assertions (to worldpay integrations tests) to test request body to PSP is as expected.

### DIFF
--- a/src/test/resources/templates/worldpay/valid-cancel-worldpay-request.xml
+++ b/src/test/resources/templates/worldpay/valid-cancel-worldpay-request.xml
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE paymentService PUBLIC "-//WorldPay//DTD WorldPay PaymentService v1//EN"
         "http://dtd.worldpay.com/paymentService_v1.dtd">
-<paymentService version="1.4" merchantCode="MERCHANTCODE">
+<paymentService version="1.4" merchantCode="{{merchantCode}}">
     <modify>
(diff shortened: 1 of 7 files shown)


src/test/resources/templates/worldpay/valid-cancel-worldpay-request.xml
```diff
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE paymentService PUBLIC "-//WorldPay//DTD WorldPay PaymentService v1//EN"
         "http://dtd.worldpay.com/paymentService_v1.dtd">
-<paymentService version="1.4" merchantCode="MERCHANTCODE">
+<paymentService version="1.4" merchantCode="{{merchantCode}}">
```
sfount, line 4: Is there any benefit to making this variable rather than just setting the test up with `merchant_id=MERCHANTCODE` and asserting that the request bodies equal?

I don't think there's anything necessarily wrong with this but potentially more complex and the actual replacing just tests core java utils

sfount (approved):
LGTM 👍

src/test/resources/templates/worldpay/valid-cancel-worldpay-request.xml
kbottla: Just making sure existing unit and integration tests can continue setting and test their own values than hardcoded one. We could do either way (variable or a fixed value across all tests)

PR #3036, opened 2021-06-22
PP-8229 add gateway merchant id patch

SandorArpa:
## WHAT YOU DID
- add new patch path to gateway account resource.
- add validation for new path
- add relevant tests


sfount (approved):
Code looks good 👍 As discussed this can be followed with a refinement to patching credentials to never override existing (non-gateway credential) keys

PR #3040, opened 2021-06-23
PP-8229 refactor patching credentials

SandorArpa:
## WHAT YOU DID
- refactor patch replace `credentials` to not replace all of credentials but to keep existing ones that are not on the top level patch request
- add test


kbottla (approved):
LGTM!

PR #3044, opened 2021-06-24
PP-8125 Get google pay merchant id using new credentials

kbottla:
## WHAT YOU DID
- Look up current or active credential to get `gateway_merchant_id`.


SandorArpa (approved):
LOKTM

PR #3046, opened 2021-06-24
PP-8255 Send payer email address to Worldpay if enabled

alexbishop1:
If a Worldpay gateway account has the setting to send the payer email address to the gateway enabled, and there is an email address for the payer (either entered directly by the user or passed to us by the service when they created the payment), send it to Worldpay in the authorisation request for a regular card payment.

The email address is sent in the XML authorisation request to Worldpay in an element called `<shopperEmailAddress>`, which is in turn inside an element called `<shopper>` (if 3D Secure is enabled, we already send a `<shopper>` element).

The <shopper> element appears after the `<paymentDetails>` element. If 3D Secure is disabled, the email address is sent like this:

```xml
<shopper>
    <shopperEmailAddress>payer@email.example</shopperEmailAddress>
</shopper>
```

If 3D Secure is enabled, the email address is sent before the `<browser>` element (this matches Worldpay’s examples in their documentation and worked when tested):

```xml
<shopper>
    <shopperEmailAddress>payer@email.example</shopperEmailAddress>
    <browser>
        <acceptHeader>*/*</acceptHeader>
        <userAgentHeader>Mozilla/5.0</userAgentHeader>
    </browser>
</shopper>
```

There are no changes if either the setting to send the payer email to the gateway is not enabled, or if there is no payer email address available.

The email address is only sent for regular card payments. It is not sent for Apple Pay or Google Pay payments.

Worldpay have a setting to send payment confirmation emails to paying users if an email address is supplied. To avoid confusing paying users, this setting should not be enabled.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderBuilder.java
@@ -3,6 +3,8 @@ package uk.gov.pay.connector.gateway.worldpay;
 import uk.gov.pay.connector.gateway.GatewayOrder;
 import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
 
+import java.util.Optional;
+
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
 import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderRequestBuilder.aWorldpayAuthoriseOrderRequestBuilder;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_MERCHANT_ID;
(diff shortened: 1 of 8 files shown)


mrlumbu (commented):
Code looks good. Just to confirm: you have tested this against worldpay test gateway? And also, regarding worldpay sending emails, this setting is presumably in the service's control, so do we need to communicate this at all before release?

alexbishop1:
Yes, I’ve tested against a Worldpay test gateway: there’s a new test in `WorldpayPaymentProviderTest`, which is the set of tests that — when enabled (they’re not normally) — actually hit Worldpay.

The setting to send the payer email address to Worldpay is controlled by us (it’s in toolbox not the admin tool) and there’s a note there warning to make sure the service has not enabled sending emails to payers within Worldpay.

mrlumbu (approved):

PR #3049, opened 2021-06-25
PP-8255 Send payer email to Worldpay for Apple Pay and Google Pay

alexbishop1:
When authorising an Apple Pay or Google Pay payment, if the gateway account has the setting to send the payer email address to the gateway enabled, and an email address was supplied for the payer (for Apple Pay, it’s technically the email address of the shipping contact), send it to Worldpay in the authorisation request.

The email address is sent in the XML authorisation request to Worldpay in an element called `<shopperEmailAddress>`, which is in turn inside an element called <shopper> (if 3D Secure is enabled, we already send a `<shopper>` element).

The `<shopper>` element appears after the `<paymentDetails>` element. If it’s an Apple Pay payment, or a Google Pay payment and 3D Secure is disabled, the email address is sent like this:

```xml
<shopper>
    <shopperEmailAddress>name@email.example</shopperEmailAddress>
</shopper>
```

If it’s a Google Pay payment and 3D Secure is enabled, the email address is sent before the `<browser>` element:

```xml
<shopper>
    <shopperEmailAddress>name@email.example</shopperEmailAddress>
    <browser>
        <acceptHeader>*/*</acceptHeader>
        <userAgentHeader>Mozilla/5.0</userAgentHeader>
    </browser>
</shopper>
```

There are no changes if either the setting to send the payer email to the gateway is not enabled, or if there is no email address available.

At the point where we send the authorisation request, we have not yet stored the email address on the charge, so we get it from the `WalletPaymentInfo` (the normalised Apple Pay or Google Pay payload). Although the template has direct access to the `WalletPaymentInfo` object, getting the email from there would bypass the check that the setting to send the payer email address to the gateway is enabled (and we’re not able to check that setting from within the template). Instead, we check if the setting is enabled earlier and if it is we add the email address (if there is one) to the template data and retrieve it from there within the template.

Sending the email address for regular card payments was added in commit 15815a2ce4f1c3a01a7f291b554065b61fcc2788.

Worldpay have a setting to send payment confirmation emails to paying users if an email address is supplied. To avoid confusing paying users, this setting should not be enabled.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/wallets/WorldpayWalletAuthorisationHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/wallets/WorldpayWalletAuthorisationHandler.java
@@ -14,6 +14,7 @@ import uk.gov.pay.connector.wallets.WalletAuthorisationHandler;
 import javax.inject.Inject;
 import java.net.URI;
 import java.util.Map;
+import java.util.Optional;
 
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
 import static uk.gov.pay.connector.gateway.util.AuthUtil.getGatewayAccountCredentialsAsAuthHeader;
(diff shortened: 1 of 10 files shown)


stephencdaly (approved):
Looks good

PR #3054, opened 2021-06-29
PP-8277 add charge exists state with hard coded params

SandorArpa:
## WHAT YOU DID
- add new state with hard coded gateway account and charge external id


kbottla (approved):
LGTM!

PR #3057, opened 2021-06-29
PP-8282 Enable switching PSP only if account as active creds

kbottla:
## WHAT YOU DID
- Allow enabling PSP switching only if the current credential (payment provider) is fully ACTIVE. There should be ac active payment provider to continue taking payments with when adding a new provider to switch to.

- Also added logging when provider switch is enabled and gateway account is switched to new PSP


SandorArpa (approved):
LOKTM

PR #3064, opened 2021-07-01
PP-8282 Get gateway name - return credentials in any state if no ACTIVE credential exist

kbottla:
## WHAT

- Currently getGatewayName returns payment provider for ACTIVE credential if multiple credentials exists. If for any reason, there are multiple credentials but no ACTIVE credential found, an exception is thrown which results in serialisation errors affecting read operations involving these gateway account(s). This could break admin tool access for services, toolbox functionality and even new payments.
       - So fall back to getting first credential if no ACTIVE credentials found.
- All the logic can be moved to service class once we remove serialising GatewayAccountEntity directly and use DTO object for serialisation.


stephencdaly (approved):
Looks good

PR #3065, opened 2021-07-01
PP-8282 Refactor GatewayAccountCredentials patch

kbottla:
## WHAT YOU DID
- Refactor updating gateway account credentials to use switch/case and moved much of the validation to validator
- Added additional unit tests for validation

- Also change logging to warn if a gateway account has multiple but no active credential


stephencdaly (approved):
LGTM

PR #3066, opened 2021-07-02
PP-8214 Add provider_switch_enabled field to response

kbottla:
## WHAT YOU DID
- Currently `provider_switch_enabled` is only available on `/v1/frontend/accounts/` response. Added provider_switch_enabled to GatewayAccountResourceDTO which ensures field is available on all other account endpoints (/v1/api/accounts/*)


SandorArpa (approved):
LGTM

PR #3067, opened 2021-07-02
PP-8316 fix blank prefilled email address when creating a charge

SandorArpa:
## WHAT YOU DID
- no blank value should pass through the system


alexbishop1 (approved):
In an ideal world, we’d just make it a validation error to send an empty-string email address in a request to create a payment but since we know services are already doing that and we want to be kind and not break them, this is a reasonable fix.

PR #3069, opened 2021-07-05
PP-8281 Use hardcoded provider name for GatewayClient

stephencdaly:
When calling GatewayClient methods from payment provider specific classes, pass a hardcoded payment provider name for logging/metrics rather than using the gateway name on the GatewayAccountEntity, as the active gateway might not be the gateway used for the charge.


kbottla (approved):
LGTM!

PR #3070, opened 2021-07-05
PP-8281 Don't pass gateway account to GatewayClient

stephencdaly:
Don't get the gateway name from the GatewayAccountEntity, as the active payment provider might not be the gateway used for the charge. Instead, use a hardcoded gateway name for Stripe requests and also pass the gateway account type.


kbottla (approved):
LGTM!

PR #3072, opened 2021-07-05
PP-8318: Take the last IP address

oswaldquek:
Take the last IP address in the list passed to connector as that is the correct
IP of the client and is set by our infrastructure.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/util/IpAddressMatcherTest.java
+++ b/src/test/java/uk/gov/pay/connector/util/IpAddressMatcherTest.java
@@ -30,14 +30,14 @@ class IpAddressMatcherTest {
     }
 
     @Test
-    void shouldReturnTrueWhenFirstOfForwardedIpAddressesMatches() {
-        String ipAddresses = "1.2.3.4, 102.106.2.1";
+    void shouldReturnTrueWhenLastOfForwardedIpAddressesMatches() {
+        String ipAddresses = "102.106.2.1, 1.2.3.4";
 
         assertTrue(ipAddressMatcher.isMatch(ipAddresses, ALLOWED_IP_ADDRESSES));
     }
 
     @Test
-    void shouldReturnFalseWhenFirstOfForwardedIpAddressesDoesNotMatch() {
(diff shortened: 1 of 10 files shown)


src/test/java/uk/gov/pay/connector/util/IpAddressMatcherTest.java
```diff
@@ -30,15 +30,15 @@ void shouldReturnFalseWhenOnlyOneForwardedIpAddressDoesNotMatch() {
     }
 
     @Test
-    void shouldReturnTrueWhenFirstOfForwardedIpAddressesMatches() {
-        String ipAddresses = "1.2.3.4, 102.106.2.1";
+    void shouldReturnTrueWhenLastOfForwardedIpAddressesMatches() {
+        String ipAddresses = "1.2.3.4, 102.106.2.1, 1.2.3.4";
```
danworth: This might be being picky but because the allowed IP address is both the first and last, this test can pass whether the method takes the first or last (it would pass regardless of the change you've just made I think). I think this goes for the test modified below too.

oswaldquek: Do you mean like if we ever decide to change it so we're taking the first IP address, then the test would still pass?

oswaldquek: I'm not sure I really get what you're saying...

danworth: I mean that the method that you've changed will pass this test whether it takes the first one or the last one. This test is trying to assert that the method being tested is taking the last one but it isn't because it could take either the first one or the last one and pass. If you changed the
```
String ipAddresses = "111.111.111.11, 102.106.2.1, 1.2.3.4";
```
Then it can only pass this test if it is indeed taking the last IP address.

danworth (approved):
Looks good, given the interaction of the various load balancers and proxies between notifications and connector and the way they each modify the `x-forward-for` header we should roll this out very carefully. There is a notifications smoke test we can use but some manual testing via ci and deploy (who's elastic ip address should be in the allowed list for sandbox, and perhaps other PSP) would be good I think 👍

PR #3073, opened 2021-07-05
PP-8281 Use charge payment provider for metrics/logging

stephencdaly:
Use the payment provider stored on the charge, rather than the active payment provider for the gateway account for metrics and logging, as these may not be the same if the account has switched or is in the process of switching providers.


kbottla (approved):
LGTM!

PR #3075, opened 2021-07-06
PP-8281 Use charge payment provider

stephencdaly:
Use charge payment provider, rather than the provider for the active credentials on the account in the following scenarios:
- querying a charge with the gateway
- refunding a charge
- parity checking with ledger


kbottla (approved):
LGTM!

PR #3076, opened 2021-07-06
PP-8270 Move not-null constraint to separate change set

kbottla:
## WHAT YOU DID
- DB migration for 'add send_payer_email_to_gateway column' change set ran without 'non null' constraint in all environments. So the constraint added afterwards is causing liquibase to fail due to incorrect checksum.
- Added a separate change set for non-null constraint on `gateway_accounts.send_payer_email_to_gateway`.


SandorArpa (approved):
LGTM

PR #3082, opened 2021-07-07
PP-8270 set gateway_account_credential_id on charges

kbottla:
## WHAT 

- Sets gateway_account_credential_id on charges (card payments & telephone payments reported)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -238,14 +238,14 @@ public class ChargeService {
                     ? chargeRequest.getLanguage()
                     : SupportedLanguage.ENGLISH;
 
+            GatewayAccountCredentialsEntity gatewayAccountCredential;
             if (chargeRequest.getPaymentProvider() != null) {
                 // check usable credentials exist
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -233,22 +238,23 @@ private ChargeStatus internalChargeStatus(String code) {
                     ? chargeRequest.getLanguage()
                     : SupportedLanguage.ENGLISH;
 
+            GatewayAccountCredentialsEntity gatewayAccountCredential;
             if (chargeRequest.getPaymentProvider() != null) {
                 // check usable credentials exist
```
stephencdaly, line 243: This comment can probably go, I only added it because at the time we were just calling this method to ensure an exception wasn't thrown. But now that we're using the result it's clearer what's happening.

stephencdaly (approved):
LGTM!

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: I will remove in next PR.

PR #3084, opened 2021-07-07
PP-8279 Remove charge email sanitisation at write

sfount:
Remove call to sanitise numbers in strings from the patch email address
endpoint.

This is called any time that the card payments page submits an
email separately from the main card details. The card payments service will
always validate to ensure a correct email format is used which precludes
plain card numbers having been accidentally entered.

Now that valid email addresses consisting of only numbers are no longer
restricted, allow persisting them un-redacted. Relates to https://github.com/alphagov/pay-frontend/pull/2295

Services are able to provide emails with invalid formats through
pre-filled data in a payment request however that code path isn't related
to the card payments page.


kbottla (approved):
LGTM!

PR #3091, opened 2021-07-12
PP-8239 Insert into history table for gateway_account_credentials

stephencdaly:
Use the `@Customizer` annotation on GatewayAccountCredentialsEntity in the same way as used for RefundEntity so that JPA automatically inserts into the gateway_account_credentials_history table when the entity is updated.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/model/domain/HistoryCustomizer.java
+++ b/src/main/java/uk/gov/pay/connector/common/model/domain/HistoryCustomizer.java
@@ -1,4 +1,4 @@
-package uk.gov.pay.connector.refund.model.domain;
+package uk.gov.pay.connector.common.model.domain;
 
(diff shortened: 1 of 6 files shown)


src/main/resources/migrations.xml
```diff
@@ -1661,4 +1661,22 @@
         <addNotNullConstraint tableName="gateway_accounts" columnName="send_payer_email_to_gateway"/>
     </changeSet>
 
+    <changeSet id="build gateway_account_credentials_history table" author="">
```
kbottla: Can we do DB migration in separate PR ?

stephencdaly: oh yeh good point

stephencdaly: migration was moved to another PR and run on all environments

kbottla (approved):
LGTM!

PR #3093, opened 2021-07-12
PP-8218 remove stripe gateway check

SandorArpa:
## WHAT YOU DID
- remove isStripeGatewayAccount check from /stripe-setup endpoint. This check is no longer valid. Gateway is not top level assigned to PPS but rather on a credential level.


nimalank7 (approved):
LGTM! I assume further PR/tickets will untangle the confusion that this endpoint now engenders!

PR #3099, opened 2021-07-15
PP-8209 remove merchant_id from patch endpoint

SandorArpa:
## WHAT YOU DID

- remove the option to patch merchant_id on endpoint used for the "old" credentials
- remove validation for merchant_id
- update tests


stephencdaly (approved):
LGTM

PR #3101, opened 2021-07-15
PP-8333 Add integration tests for PSP switch

stephencdaly:
Add integration tests to ensure
- Active credentials are used for authorisation when there are multiple
- Refunds are made using the payment provider for the charge when credentials are RETIRED


SandorArpa (approved):
LOKTM

PR #3103, opened 2021-07-16
PP-8210 - Add extra validation for gateway account credentials

nimalank7:
Description:
- Covers scenarios 1 and 2. Use the JIRA ticket to verify if PR meets the scenarios.
- Added a test for `gatewayAccountCredentialsService.getCurrentOrActiveCredential(GatewayAccountEntity gatewayAccountEntity)` as there was nothing testing this method previously
- Added a new exception and exception mapper to return 400 with JSON response
- `GatewayAccountEntity.getCurrentOrActiveGatewayAccountCredential()` no longer returns when the list of GatewayAccountCredentials is 1 because it is possible for that it returns in the case of CREATED which we want to disallow
- Instead we now filter if not CREATED and return if present, otherwise throw an exception if the only credential is CREATED and otherwise return an empty Optional
- Scenario 1b is already implemented in `GatewayAccountCredentialsService.getUsableCredentialsForProvider()` as ENTERED is part of USABLE_STATES


src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -240,8 +241,13 @@ public GatewayAccountCredentialsEntity getUsableCredentialsForProvider(GatewayAc
             throw new NoCredentialsExistForProviderException(paymentProvider);
         }
 
-        List<GatewayAccountCredentialsEntity> credentialsInState = credentialsForProvider.stream().filter(gatewayAccountCredentialsEntity ->
-                USABLE_STATES.contains(gatewayAccountCredentialsEntity.getState()))
+        if (credentialsForProvider.size() == 1 && CREATED.equals(credentialsForProvider.get(0).getState())) {
+            throw new GatewayAccountCredentialsNotConfiguredException();
+        }
```
stephencdaly: Doesn't the below already do this logic? It checks whether there is a credential for the payment provider in state `ENTERED`, `VERIFIED_WITH_LIVE_PAYMENT` or `ACTIVE` and if not throws an exception. I think we don't need to change this code, just alter the `NoCredentialsInUsableStateException` to return a response with the new error_identifier

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/exception/GatewayAccountCredentialsNotConfiguredException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.pay.connector.gatewayaccountcredentials.exception;
+
+public class GatewayAccountCredentialsNotConfiguredException extends RuntimeException {
```
stephencdaly: Could we re-use `NoCredentialsInUsableStateException` that was already added and add in the new `error_identifier` to that?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -222,11 +220,20 @@ public String getGatewayName() {
...
-                    .or(() -> getGatewayAccountCredentials().stream().findFirst());
+                    .filter(entity -> entity.getState() != CREATED)
+                    .min(comparing(GatewayAccountCredentialsEntity::getCreatedDate));
+
+            if (mayBeNonCreatedCredential.isPresent()) {
+                return mayBeNonCreatedCredential;
+            }
+
+            if (gatewayAccountCredentialsEntities.size() == 1) {
+                throw new GatewayAccountCredentialsNotConfiguredException();
```
stephencdaly: I think we don't want to throw an exception from this method because it's used for serialising the `payment_provider` field on the gateway account response for some endpoints.

I think also there's a problem that we still want to return the correct `payment_provider` even when the credentials are only in a `CREATED` state. So I think this method needs to stay as it is, but the calling method that gets the credentials to use for the new charge needs to check whether a `CREATED` credential has been returned, and if so throw an exception.

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntityTest.java
```diff
@@ -35,15 +36,6 @@ void setUp() {
                 .build();
     }
 
-    @Test
-    void getGatewayNameShouldReturnPaymentProviderOfOnlyCredentialAvailable() {
```
stephencdaly: I think this test should still apply, following on from the comment on GatewayAccountEntity.

stephencdaly (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/exception/GatewayAccountCredentialsNotConfiguredException.java
nimalank7: Done.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
nimalank7: Done.

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntityTest.java
nimalank7: Done.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateIT.java
```diff
@@ -1002,4 +1007,65 @@ public void shouldReturn400WhenNoCredentialsInUsableStateForProvider() {
...
+        connectorRestApiClient
+                .postCreateCharge(postBody, accountId)
+                .statusCode(Status.BAD_REQUEST.getStatusCode())
+                .contentType(JSON)
+                .body("message", contains("Payment provider details are not configured on this account"))
+                .body("error_identifier", is(ErrorIdentifier.ACCOUNT_NOT_LINKED_WITH_PSP.toString()));
+    }
+
+    @Test
+    public void shouldReturn400WhenCredentialsInCreatedStateForProvider() {
```
stephencdaly: I reckon we don't need this test as it's pretty much a duplicate of `shouldReturn400WhenNoCredentialsInUsableStateForProvider`. As integration tests are expensive in terms of runtime, we should have the minimum possible just to ensure the different components of the system are working together. If we felt we needed to test each individual state (I reckon probably not necessary) it could be tested in the service tests.

stephencdaly (dismissed):
LGTM but I think we could remove one of the ITs

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateIT.java
nimalank7: Yeah I did debate about whether to include this as well. I'll probably remove it.

stephencdaly (dismissed):
LGTM

stephencdaly (dismissed):

stephencdaly (approved):

PR #3109, opened 2021-07-19
PP-8374 Fix search gateway account by payment provider

stephencdaly:
When searching for gateway account by payment provider, use the payment
provider on the gateway account credentials as we will be getting rid
of the `payment_provider` column on the `gateway_accounts` table.

This will either search on the `payment_provider` for the `ACTIVE`
`gateway_account_credentials` if there is one, else the provider for
credentials in another state, preferring `CREATED` over `RETIRED`.
Although practically, if there is no `ACTIVE` credential, there should
only ever be a single `CREATED` credential.

In order to construct this query, had to use `createNativeQuery` to use
raw SQL. There is a limitation of JPA using EclipseLink that a native
query cannot take a list as a parameter, so it is necessary to
dynamically construct the query with each entry in an `IN` statement
having its own named parameter when searching by gateway accounts by a
list of ids.

Add validation for the `accountIds` query parameter using a regex to
ensure that a list of numbers is provided. Remove the
`CommaDelimitedSetParameter` type previously used and deserialize as a
String to make the validation easier.


SandorArpa (approved):
LOKTM

PR #3116, opened 2021-07-21
PP-8373 Use paymentProvider on charge to generate DDC token

kbottla:
## WHAT YOU DID
- Use payment_provider available on charge when generating DDC token instead of `GatewayAccountEntity.getGatewayName()` which could be different if service switches to new payment provider during the payment journey
- Also stops serialising `getCurrentOrActiveGatewayAccountCredential` on GatewayAccountEntity which started when the method access changed to public.


SandorArpa (approved):
LOKTM

PR #3117, opened 2021-07-21
PP-8138 drop credentials column from gateway_accounts

SandorArpa:
## WHAT YOU DID
- drop column `credentials` from `gateway_accounts` table. it's no longer in use.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1679,4 +1679,8 @@
         </createTable>
     </changeSet>
 
+    <changeSet id="drop credentials from gateway_accounts table" runInTransaction="false" author="">
+        <dropColumn tableName="gateway_accounts" columnName="credentials"/>
+    </changeSet>
+
 </databaseChangeLog>


src/main/resources/migrations.xml
```diff
@@ -1679,4 +1679,8 @@
         </createTable>
     </changeSet>
 
+    <changeSet id="drop credentials from gateway_accounts table" runInTransaction="false" author="">
+        <dropColumn tableName="gateway_accounts" columnName="credentials"/>
```
kbottla, line 1683: This will break e2e tests. Need to update https://github.com/alphagov/pay-endtoend/blob/6d25ed392f6bdec3d00ce5c7f8d83d5f279b24bc/src/test/java/uk/gov/pay/endtoend/fixtures/DatabaseFixture.java#L51-L58

SandorArpa: 👍

kbottla (approved):

PR #3120, opened 2021-07-22
Revert "PP-8373 Use paymentProvider on charge to generate DDC token"

kbottla:
Reverts alphagov/pay-connector#3116


SandorArpa (approved):
LOKTM

PR #3128, opened 2021-07-26
PP-8400 Log for a refund failed Stripe notification

stephencdaly:
When we receive a `charge.refund.updated` webhook from Stripe, log at info level if the "status" is failure.

We will add a Splunk alert for this log line as the failure needs to be manually rectified and the service notified.


kbottla (approved):
LGTM!

PR #3134, opened 2021-07-28
PP-7856 - Remove redundant @JsonView annotation

nimalank7:
Description:
- Method returns GatewayAccountResourceDTO not GatewayAccountEntity so the annotation is redundant


SandorArpa (approved):
LOKTM

PR #3135, opened 2021-07-28
PP-8397 Get credentials from ChargeEntity for payment inflight operations

SandorArpa:
## WHAT YOU DID

- For all payment in-flight operations (authorise, cancel, capture, 3ds)
- We currently get credentials from GatewayAccountEntity based on charge’s payment_provider. `charges` table now stores credential ID when payments are created and this can be used everywhere instead.
- `GatewayAccountCredentialsEntity` on `ChargeEntity` is eagerly (default) fetched from database, so should be available all the time `ChargeEntity` is retrieved.
- update/refactor tests accordingly

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
@@ -1,9 +1,12 @@
 package uk.gov.pay.connector.gateway.model.request;
 
+import org.apache.commons.lang3.NotImplementedException;
 import uk.gov.pay.connector.charge.model.domain.Charge;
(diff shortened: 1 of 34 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/request/RefundGatewayRequest.java
```diff
@@ -43,6 +45,11 @@ public GatewayOperation getRequestType() {
         return GatewayOperation.REFUND;
     }
 
+    @Override
+    public Map<String, String> getGatewayCredentials() {
+        return null;
```
stephencdaly: why is this null?

stephencdaly (approved):
LGTM

PR #3141, opened 2021-08-02
PP-8397 get credentials from charge entity for stripe_payments

SandorArpa:
## WHAT YOU DID
- refactor Stripe requests to use credentials from ChargeEntity
- refactor RefundGatewayRequest to use GatewayAccountCredentialsEntity for credentials
- refactor tests accordingly

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
@@ -233,6 +233,16 @@ public class GatewayAccountEntity extends AbstractVersionedEntity {
                                 getId(), paymentProvider))));
     }
 
+    @JsonIgnore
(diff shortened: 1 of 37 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -233,6 +233,16 @@ public String getGatewayName() {
                                 getId(), paymentProvider))));
     }
 
+    @JsonIgnore
+    public GatewayAccountCredentialsEntity getGatewayAccountCredentialsEntity(String paymentProvider) {
```
kbottla, line 237: Would be ideal if this is in GatewayAccountCredentialsService. Fine as it is for now if it make things simple.

kbottla (approved):
LGTM!

PR #3150, opened 2021-08-05
PP-8342 - Add credential_external_id to Charge and LedgerTransaction

nimalank7:
Description:
- credential_external_id is needed to retrieve the gateway account credentials for a charge when performing a refund so needs to be added
 to both Charge and LedgerTransaction model entities
- Ledger currently sends credential_external_id so can just be added to LedgerTransaction
- See RefundResource.submitRefund() which uses chargeService.findCharge() which returns a Charge model that is used to perform the refund

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
@@ -13,6 +13,7 @@ public class Charge {
     private String status;
     private String externalStatus;
     private String gatewayTransactionId;
+    private String credentialExternalId;
     private Long corporateSurcharge;
     private String refundAvailabilityStatus;
     private String reference;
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/client/ledger/model/LedgerTransaction.java
```diff
@@ -21,6 +21,7 @@
     private String transactionId;
     private Long amount;
     private String gatewayAccountId;
+    private String credentialExternalId;
```
kbottla, line 24: Do we have anywhere (may be a pact test) that checks that field is correctly deserialised from ledger ?

src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
```diff
@@ -136,6 +145,30 @@ public String getEmail() {
...
+
+    public String getPaymentGatewayName() {
+        return paymentGatewayName;
+    }
+
+    public void setHistoric(boolean historic) {
+        this.historic = historic;
+    }
+
+    public String getCredentialExternalId() {
```
kbottla, line 164: Worth returning this field as Optional<String> to avoid checking for nulls (in multiple places) later.

kbottla (approved):
LGTM! Added a few minor comments and can be addressed in a separate PR.

src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
nimalank7: Yes.

src/main/java/uk/gov/pay/connector/client/ledger/model/LedgerTransaction.java
nimalank7: No. There is an already existing pact test (connector-ledger-get-payment-transaction.json) which should've been checking for credential_external_id. Upon reflection I should modify this to check for credential_external_id as well since we're now depending on it for refunds.

PR #3151, opened 2021-08-05
PP-8342 - Update Pact test

nimalank7:
Description:
- Modify pact test to check for credential_external_id
- Modify Charge model to return Optional<String> to prevent further null checks

### DIFF
--- a/src/test/resources/pacts/connector-ledger-get-payment-transaction.json
+++ b/src/test/resources/pacts/connector-ledger-get-payment-transaction.json
@@ -44,6 +44,7 @@
           "return_url": "https://service-name.gov.uk/transactions/12345",
           "email": "Joe.Bogs@example.org",
           "payment_provider": "sandbox",
+          "credential_external_id": "credential-external-id",
           "created_date": "2020-02-13T16:26:04.204Z",
           "delayed_capture": false,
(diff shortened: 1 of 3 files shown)


src/test/resources/pacts/connector-ledger-get-payment-transaction.json
```diff
@@ -44,6 +44,7 @@
           "return_url": "https://service-name.gov.uk/transactions/12345",
           "email": "Joe.Bogs@example.org",
           "payment_provider": "sandbox",
+          "credential_external_id": "credential-external-id",
```
kbottla, line 47: Can we assert this field in the test https://github.com/alphagov/pay-connector/blob/8b971c830fc607443026244ae4e4db31d0b8efbb/src/test/java/uk/gov/pay/connector/client/ledger/service/LedgerServiceConsumerTest.java#L47?

nimalank7: Nice catch I completely forgot to update that!

nimalank7: Done.

kbottla (approved):

PR #3157, opened 2021-08-09
PP-8437 Serialise service id on account charge entities

sfount:
Depends on https://github.com/alphagov/pay-connector/pull/3156

Add `serviceId` field to gateway account entities and resource maps. (and charge entity)

Update gateway account test fixtures to reflect service id and assert on resource.


alexbishop1 (approved):
I’d really like us to have a specific type for the service ID like we have `ServicePaymentReference` for the service-supplied payment reference. It can be a bit tricky to have a specific type when it might be null though (have to be careful with the database and JSON converters).

sfount:
> I’d really like us to have a specific type for the service ID like we have `ServicePaymentReference` for the service-supplied payment reference. It can be a bit tricky to have a specific type when it might be null though (have to be careful with the database and JSON converters).

@alexbishop1 Yeah agree it would be good to think this through a bit further in planning, especially as this will start to be the partial index across all services

PR #3163, opened 2021-08-11
PP-8438 Optionally accept `service_id` on create account route

sfount:
POST `/v1/api/accounts/` accepts `service_id` in the request body and
writes this to the corresponding database column.

Parameter is currently nullable in order to not break before consumers
are updated.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/CreateGatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/CreateGatewayAccountResourceIT.java
@@ -51,6 +51,16 @@ public class CreateGatewayAccountResourceIT extends GatewayAccountResourceTestBa
     }
 
     @Test
+    public void createAGatewayAccountWithServiceId() {
+        String gatewayAccountId = createAGatewayAccountWithServiceId("some-special-service-id");
+        givenSetup()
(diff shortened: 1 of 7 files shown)


alexbishop1 (changes requested):
What’s here looks good but feels like there should be some tests in `CreateGatewayAccountResourceIT`.

sfount:
> What’s here looks good but feels like there should be some tests in `CreateGatewayAccountResourceIT`.

🤦  covered in `GatewayAccountResourceIT` but didn't realise they were split up in the test layout, will add those now.

src/test/java/uk/gov/pay/connector/it/resources/CreateGatewayAccountResourceIT.java
```diff
@@ -50,6 +50,16 @@ public void createAGatewayAccount(String provider) {
         assertCorrectCreateResponse(response, TEST, "my test service", "analytics", null);
     }
 
+    @Test
+    public void getAccountShouldReturnServiceId() {
```
alexbishop1: Name doesn’t seem very fitting here: `createAGatewayAccountWithServiceId` or something would be better.

alexbishop1 (changes requested):

alexbishop1 (approved):

PR #3168, opened 2021-08-12
PP-8394 refactor signature comparison to mitigate timing attack

SandorArpa:
## WHAT YOU DID
- String.equals() doesn't implement a constant-time algorithm for comparing inputs. As a result, an attacker may be able to run a timing attack that can forge a valid signature for an arbitrary body.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
@@ -16,7 +16,10 @@ import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
 import uk.gov.pay.connector.refund.model.domain.RefundStatus;
 import uk.gov.pay.connector.util.IpAddressMatcher;
 
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
 import java.util.List;
+import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
```diff
@@ -160,7 +163,8 @@ private boolean isValidNotificationSignature(EpdqNotification notification, Gate
         );
 
         final String expectedShaSignature = getExpectedShaSignature(notification);
-        final boolean verified = actualSignature.equalsIgnoreCase(expectedShaSignature);
+        final boolean verified = MessageDigest.isEqual(actualSignature.toUpperCase().getBytes(),
+                expectedShaSignature.toUpperCase().getBytes(StandardCharsets.UTF_8));
```
alexbishop1: Not sure converting it to upper-case is necessary. 

Also, we should consistently use `getBytes(StandardCharsets.UTF_8)` (`getBytes()` uses the platform default charset, which might be something different).

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/gateway/epdq/EpdqNotificationService.java
SandorArpa: It replaces String.equalsIgnoreCase() and thus the toUpperCase(). MessageDigest has no ignore case method.

alexbishop1: Sorry, missed we were doing a case-insensitive comparison before.

alexbishop1 (approved):

PR #3170, opened 2021-08-13
PP-8446 Allow gateway accounts to be filtered on service ID

sfount:
The `service_id` column on the `gateway_accounts` table has a standard
`btree` index.

Add support for filtering the `/v1/frontend/accounts` route for accounts
that belong to a particular list of service identifiers with the
`serviceIds` parameter.

There are a number of changes that could be made to the param -> filter
methodology here but this is implemented to follow the pattern already
established.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
@@ -14,6 +14,7 @@ import static org.apache.commons.lang3.StringUtils.isBlank;
 public class GatewayAccountSearchParams {
 
     private static final String ACCOUNT_IDS_SQL_FIELD = "accountIds";
+    private static final String SERVICE_IDS_SQL_FIELD = "serviceId";
     private static final String ALLOW_MOTO_SQL_FIELD = "allowMoto";
     private static final String ALLOW_APPLE_PAY_SQL_FIELD = "allowApplePay";
     private static final String ALLOW_GOOGLE_PAY_SQL_FIELD = "allowGooglePay";
@@ -26,6 +27,11 @@ public class GatewayAccountSearchParams {
     @Pattern(regexp = "^[\\d,]+$",
             message = "Parameter [accountIds] must be a comma separated list of numbers")
     private String accountIds;
+
+    @QueryParam("serviceIds")
+    @Pattern(regexp = "^(?:[a-zA-Z0-9]++(\\,+|$))*+$",
+            message = "Parameter [serviceIds] must be a comma separated list of alpha numeric strings")
+    private String serviceIds;
 
     // This is a string value rather than boolean as if the parameter isn't provided, it should not filter by
     // moto enabled/disabled
@@ -96,10 +102,20 @@ public class GatewayAccountSearchParams {
         this.providerSwitchEnabled = providerSwitchEnabled;
     }
 
+    public void setServiceIds(String serviceIds) {
+        this.serviceIds = serviceIds;
+    }
+
     private List<String> getAccountIdsAsList() {
         return isBlank(accountIds)
                 ? new ArrayList<>()
                 : List.of(accountIds.split(","));
+    }
+
+    private List<String> getServiceIdsAsList() {
+        return isBlank(serviceIds)
+                ? new ArrayList<>()
+                : List.of(serviceIds.split(","));
     }
 
     public List<String> getFilterTemplates() {
@@ -111,6 +127,18 @@ public class GatewayAccountSearchParams {
             for (int i = 0; i < accountIdsList.size(); i++) {
                 filter.append("#").append(ACCOUNT_IDS_SQL_FIELD).append(i);
                 if (i != accountIdsList.size() - 1) {
+                    filter.append(",");
+                }
+            }
+            filter.append(")");
+            filters.add(filter.toString());
+        }
+        List<String> serviceIdsList = getServiceIdsAsList();
+        if (!serviceIdsList.isEmpty()) {
+            StringBuilder filter = new StringBuilder(" ga.service_id IN (");
+            for (int i = 0; i < serviceIdsList.size(); i++) {
+                filter.append("#").append(SERVICE_IDS_SQL_FIELD).append(i);
+                if (i != serviceIdsList.size() - 1) {
                     filter.append(",");
                 }
             }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
```diff
@@ -27,6 +28,11 @@
             message = "Parameter [accountIds] must be a comma separated list of numbers")
     private String accountIds;
 
+    @QueryParam("serviceIds")
+    @Pattern(regexp = "^(?:[a-zA-Z0-9]++(\\,+|$))*+$",
```
alexbishop1: The second set of parentheses start a capturing group, which I don’t think we want, and some of the special characters seem to be unnecessary.

I think this is equivalent:

```
^(?:[A-z0-9]+(?:,+|$))+$
```

But even this supports duplicate and trailing commas, which could mean we end up with empty strings in our resulting service IDs list after tokenising, which may do _interesting_ things to the generated SQL. So maybe we just want something like this:

```
^(?:[A-z0-9]+,?)+$
```

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
```diff
@@ -27,6 +28,11 @@
...
+            message = "Parameter [serviceIds] must be a comma separated list of alpha numeric strings")
```
alexbishop1: I’m being picky but “alphanumeric” is all one word

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
```diff
@@ -96,12 +102,22 @@ public void setProviderSwitchEnabled(String providerSwitchEnabled) {
...
+
     private List<String> getAccountIdsAsList() {
         return isBlank(accountIds)
                 ? new ArrayList<>()
                 : List.of(accountIds.split(","));
     }
 
+    private List<String> getServiceIdsAsList() {
+        return isBlank(serviceIds)
+                ? new ArrayList<>()
```
alexbishop1: I know this is just the same as what came before but we should be using something immutable like `Collections.emptyList()` or `List.of()`

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
```diff
@@ -117,6 +133,18 @@ public void setProviderSwitchEnabled(String providerSwitchEnabled) {
...
+            StringBuilder filter = new StringBuilder(" ga.service_id IN (");
+            for (int i = 0; i < serviceIdsList.size(); i++) {
+                filter.append("#").append(SERVICE_IDS_SQL_FIELD).append(i);
+                if (i != serviceIdsList.size() - 1) {
+                    filter.append(",");
+                }
+            }
+            filter.append(")");
+            filters.add(filter.toString());
+        }
```
alexbishop1: I know this just follows the pattern that came before but something like this would be so much nicer:

```java
List<String> serviceIdsList = getServiceIdsAsList();
if (!serviceIdsList.isEmpty()) {
    StringJoiner filter = new StringJoiner(",", " ga.service_id IN (", ")");
    for (int i = 0; i < serviceIdsList.size(); i++) {
        filter.add("#" + SERVICE_IDS_SQL_FIELD + i);
    }
    filters.add(filter.toString());
}
```

Or even this:

```java
List<String> serviceIdsList = getServiceIdsAsList();
if (!serviceIdsList.isEmpty()) {
    String filter = IntStream.range(0, serviceIdsList.size()).boxed()
            .map(i -> "#" + SERVICE_IDS_SQL_FIELD + i)
            .collect(joining(",", " ga.service_id IN (", ")"));
    filters.add(filter);
}
```

Obviously, make the same change in the other one.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceIT.java
```diff
@@ -242,6 +242,35 @@ public void shouldFilterGetGatewayAccountForExistingAccount() {
...
+    public void shouldFilterGetGatewayAccountForExistingAccountByServiceId() {
+        String accountId = createAGatewayAccountFor("worldpay");
+        String serviceId = "someexternalserviceid";
+        GatewayAccountPayload gatewayAccountPayload = GatewayAccountPayload.createDefault().withMerchantId("a-merchant-id");
+        databaseTestHelper.updateCredentialsFor(Long.parseLong(accountId), gson.toJson(gatewayAccountPayload.getCredentials()));
+        databaseTestHelper.updateServiceNameFor(Long.parseLong(accountId), gatewayAccountPayload.getServiceName());
+        databaseTestHelper.updateServiceIdFor(Long.parseLong(accountId), serviceId);
+
+        givenSetup().accept(JSON)
+                .get("/v1/frontend/accounts?serviceIds=" + serviceId)
```
alexbishop1: Quite tempting to have a comma-separated list of IDs because otherwise we’re not testing that the validation allows a list through

alexbishop1 (changes requested):
Looks fine but there’s a couple of things I think we can tidy up

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
sfount: Yeah my reasoning here is that doing it a second way means that a future contributor has to both learn what the existing thing is trying to do and then reason about this other method. It didn't feel like something worth refactoring old code to make more concise. (ref: the commit message)

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
sfount: Much more concise and agreed with the reasoning on edge cases above, will use the final proposed

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountSearchParams.java
alexbishop1: I kind of feel like we should change this now. The pattern was in one place before and now it’ll be two, which will make even more disruptive to fix in the future.

sfount: Happy to do it to get this through - using something as involved as JPA I don't think we should be manually building queries like this without a particularly good reason

alexbishop1 (approved):

PR #3172, opened 2021-08-16
PP-8342 - Method to return GatewayAccountCredentialsEntity for refunds

nimalank7:
Description:
- There are 4 cases here:
(i) credential_external_id is/isn't provided by charge and no gateway account credentials in GatewayAccountEntity -> return an empty optional
(ii) credential_external_id is provided by charge and matched -> return matching credential by finding the first.
(iii) credential_external_id isn't provided by charge and credentials found to use in GatewayAccountEntity -> use payment_provider to search and if none found
return an empty optional
(iv) credential_external_id isn't provided by charge and credentials found to use in GatewayAccountEntity -> use payment_provider to search and if none found
return an empty optional
- Searching by payment_provider involves searching through the credentials in GatewayAccountEntity. If multiple are found then use the credential that is before but closest to the charge
created date. If none are found then return an empty optional.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
@@ -10,6 +10,8 @@ import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import uk.gov.pay.connector.charge.model.domain.Charge;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntityFixture;
 import uk.gov.pay.connector.common.model.api.jsonpatch.JsonPatchRequest;
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountCredentialsNotFoundException;
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
```diff
@@ -373,4 +379,139 @@ void shouldThrowWebApplicationExceptionIfNoCredentialsFound() {
...
+                .build();
+        GatewayAccountEntity gatewayAccountEntity = aGatewayAccountEntity()
+                .withGatewayAccountCredentials(List.of(worldpayGatewayAccountCredentialsEntityOne, smartpayGatewayAccountCredentialsEntityOne))
+                .build();
+        Charge charge = Charge.from(ChargeEntityFixture.aValidChargeEntity().withPaymentProvider(WORLDPAY.getName()).build());
+        String chargeCredentialExternalId = charge.getCredentialExternalId().get();
+        Optional<GatewayAccountCredentialsEntity> gatewayAccountCredentialsEntity = gatewayAccountCredentialsService.findCredentialFromCharge(charge, gatewayAccountEntity);
+
+        assertThat(gatewayAccountCredentialsEntity.isPresent(), is(true));
+        assertThat(gatewayAccountCredentialsEntity.get().getPaymentProvider(), is("worldpay"));
```
stephencdaly, line 430: you should be able to generate an `equals` method on GatewayAccountCredentialsEntity and directly compare equality to make this more robust right?

stephencdaly (approved):
Looks good and well tested. Minor suggestion for improving tests but not a blocker for merging this.

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
nimalank7: Possibly - I think the comparison with the date and the payment provider is sufficient.

PR #3177, opened 2021-08-18
PP-8342 - Update RefundServiceTest

nimalank7:
Description:
- Updated to JUnit 5 however Mockito is strict with throwing UnnecessaryStubbingException as the stubs setup aren't used in some tests
so this has been set to lenient
- Renamed a test as it was testing the same scenario as shouldFailWhenAmountAvailableForRefundMismatchesWithoutCorporateSurcharge
- Ran tests with coverage in IntelliJ and noticed that checkIfChargeIsRefundableOrTerminate and checkIfRefundAmountWithinLimitOrTerminate weren't
being tested. Tests have been added to rectify this

### DIFF
--- a/src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
@@ -1,18 +1,15 @@
 package uk.gov.pay.connector.service;
 
-import org.junit.Before;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.rules.ExpectedException;
-import org.junit.runner.RunWith;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.ArgumentMatcher;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
-import uk.gov.pay.connector.charge.dao.ChargeDao;
+import org.mockito.junit.jupiter.MockitoExtension;
 import uk.gov.pay.connector.charge.model.ChargeResponse;
 import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
-import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.client.ledger.model.LedgerTransaction;
 import uk.gov.pay.connector.client.ledger.service.LedgerService;
 import uk.gov.pay.connector.common.model.api.ExternalRefundStatus;
@@ -51,10 +48,13 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.core.Is.is;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.any;
 import static org.mockito.Mockito.argThat;
 import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
@@ -66,6 +66,7 @@ import static uk.gov.pay.connector.charge.model.domain.ChargeEntityFixture.aVali
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
 import static uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability.EXTERNAL_AVAILABLE;
+import static uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability.EXTERNAL_UNAVAILABLE;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.SANDBOX;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.SMARTPAY;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
@@ -78,21 +79,14 @@ import static uk.gov.pay.connector.model.domain.RefundTransactionsForPaymentFixt
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.CREATED;
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUNDED;
 
-@RunWith(MockitoJUnitRunner.class)
+@ExtendWith(MockitoExtension.class)
 public class RefundServiceTest {
 
     private RefundService refundService;
     private Long refundId;
 
-    @Rule
-    public ExpectedException expectedException = ExpectedException.none();
-
     @Mock
     private GatewayAccountDao mockGatewayAccountDao;
-    @Mock
-    private ChargeDao mockChargeDao;
-    @Mock
-    private ChargeService mockChargeService;
     @Mock
     private RefundDao mockRefundDao;
     @Mock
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -117,18 +110,19 @@
                 .withGatewayAccountCredentials(gatewayAccountCredentialsEntityList)
                 .build();
 
-    @Before
-    public void setUp() {
+    // Mockito is strict with checking for unnecessary stubbings in @BeforeEach so lenient has to be used
```
alexbishop1: I think we can kill this comment: people will be smart enough to assume `lenient()` is being used for a good reason (or find out there’s a good reason when they try to remove it)

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -176,12 +170,10 @@ public void shouldRefundSuccessfully_forWorldpay() {
         verify(mockRefundDao, times(1)).findById(refundId);
         verify(spiedRefundEntity).setStatus(RefundStatus.REFUND_SUBMITTED);
         verify(spiedRefundEntity).setGatewayTransactionId(refundEntity.getExternalId());
-
-        verifyNoMoreInteractions(mockChargeDao);
```
alexbishop1, line 180: Sure we want to remove this? It seems like it might be checking we don’t try to access the charge to do the refund, which needs to work if the charge has been expunged. Using `verifyNoInteractions(mockChargeDao)` would make the intent clearer though.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -508,18 +496,17 @@ public void shouldFailWhenChargeRefundIsNotAvailable() {
                 .build();
 
         when(mockGatewayAccountDao.findById(accountId)).thenReturn(Optional.of(account));
+        when(mockRefundDao.findRefundsByChargeExternalId(externalChargeId)).thenReturn(List.of());
+        when(mockProvider.getExternalChargeRefundAvailability(any(Charge.class), any(List.class))).thenReturn(EXTERNAL_UNAVAILABLE);
```
alexbishop1: Can we not do better than using very generic `any(…)` matchers? Surely we know what the list will contain?

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -508,18 +496,17 @@ public void shouldFailWhenChargeRefundIsNotAvailable() {
...
-        expectedException.expectMessage("HTTP 412 Precondition Failed");
-
-        refundService.doRefund(accountId, Charge.from(chargeEntity), new RefundRequest(100L, 0, userExternalId));
-
-        verify(mockChargeDao).findByExternalIdAndGatewayAccount(externalChargeId, accountId);
-        verifyNoMoreInteractions(mockChargeDao, mockRefundDao, mockChargeService);
+        var thrown = assertThrows(RefundException.class,
+                () -> refundService.doRefund(accountId, Charge.from(chargeEntity), new RefundRequest(100L, 0, userExternalId)));
+        assertThat(thrown.getMessage(), is("HTTP 400 Bad Request"));
+        verifyNoMoreInteractions(mockRefundDao);
```
alexbishop1: We really care that it’s not going to try to persist anything to the `mockRefundDao`, right? Can we be more specific and use `verify(mockRefundDao, never()).persist(…)` or something?

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -623,6 +606,9 @@ public void shouldFailWhenANewRefundHasBeenCreatedSincePreviouslyQueried() {
                 .withStatus(CREATED)
                 .build();
 
+        Charge charge = Charge.from(chargeEntity);
+        charge.setHistoric(true);
```
alexbishop1, line 612: Why is this now needed?

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -635,19 +621,44 @@ public void shouldFailWhenANewRefundHasBeenCreatedSincePreviouslyQueried() {
...
+                .withGatewayAccountEntity(account)
+                .withTransactionId("transactionId")
+                .withStatus(AUTHORISATION_SUCCESS)
+                .build();
+
+        when(mockGatewayAccountDao.findById(accountId)).thenReturn(Optional.of(account));
+        when(mockRefundDao.findRefundsByChargeExternalId(chargeEntity.getExternalId())).thenReturn(List.of());
+
+        var thrown = assertThrows(RefundException.class,
+                () -> refundService.doRefund(accountId, Charge.from(chargeEntity), new RefundRequest(amount + 1, amount, userExternalId)));
```
alexbishop1: I think extracting `amount + 1` to a named local variable could make what’s happening here clearer

alexbishop1 (changes requested):

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: Sure.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: Yes. `RefundEntity` no longer depends on `ChargeEntity` so no need to call `chargeDao`. Similarly the `doRefund` takes in a `Charge` model which means prior to that method call it has already called. Nothing in `RefundService` depends on calling `chargeDao`.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: Yep modified that.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: Yeah made it more specific.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: Those 2 lines were moved up from later down in that same test.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: Done.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
alexbishop1: Yes, so my point was that perhaps that verification was there to ensure that we’re not relying on `ChargeDao` (i.e. the verification should remain)

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -508,18 +496,17 @@ public void shouldFailWhenChargeRefundIsNotAvailable() {
                 .build();
 
         when(mockGatewayAccountDao.findById(accountId)).thenReturn(Optional.of(account));
+        when(mockRefundDao.findRefundsByChargeExternalId(externalChargeId)).thenReturn(List.of());
+        when(mockProvider.getExternalChargeRefundAvailability(any(Charge.class), eq(List.of()))).thenReturn(EXTERNAL_UNAVAILABLE);
```
alexbishop1: I think if we pull the `Charge.from(chargeEntity)` (from a couple of lines below) into a local variable then we should be able to stub using the exact `Charge` object

alexbishop1 (changes requested):

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: The `chargeDao` mock isn’t defined anywhere else in the test so no need to verify it doesn’t interact with it.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
nimalank7: Yes good spot.

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
alexbishop1: OK, I see

src/test/java/uk/gov/pay/connector/service/RefundServiceTest.java
```diff
@@ -507,19 +495,20 @@ public void shouldFailWhenChargeRefundIsNotAvailable() {
...
-        when(mockGatewayAccountDao.findById(accountId)).thenReturn(Optional.of(account));
-
-        expectedException.expect(RefundException.class);
-        expectedException.expectMessage("HTTP 412 Precondition Failed");
+        Charge charge = Charge.from(chargeEntity);
 
-        refundService.doRefund(accountId, Charge.from(chargeEntity), new RefundRequest(100L, 0, userExternalId));
+        when(mockGatewayAccountDao.findById(accountId)).thenReturn(Optional.of(account));
+        when(mockRefundDao.findRefundsByChargeExternalId(externalChargeId)).thenReturn(List.of());
+        when(mockProvider.getExternalChargeRefundAvailability(eq(charge), eq(List.of()))).thenReturn(EXTERNAL_UNAVAILABLE);
```
alexbishop1, line 502: We don’t actually need the `eq(…)` wrappers any more

alexbishop1 (approved):

PR #3178, opened 2021-08-18
PP-8342 - RefundRequest uses GatewayAccountCredentialsEntity

nimalank7:
Description:
- `RefundRequest` now accepts `GatewayAccountCredentialsEntity` so that `PaymentProvider` implementions of refund can call credentials directly rather than via `GatewayAccountEntity`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
+++ b/src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
@@ -12,8 +12,11 @@ import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.GatewayRefundResponse;
 import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
+import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountCredentialsNotFoundException;
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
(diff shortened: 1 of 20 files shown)


src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -57,30 +61,35 @@
...
         this.stateTransitionService = stateTransitionService;
         this.ledgerService = ledgerService;
+        this.gatewayAccountCredentialsService = gatewayAccountCredentialsService;
     }
 
     public ChargeRefundResponse doRefund(Long accountId, Charge charge, RefundRequest refundRequest) {
         GatewayAccountEntity gatewayAccountEntity = gatewayAccountDao.findById(accountId).orElseThrow(
                 () -> new GatewayAccountNotFoundException(accountId));
+        GatewayAccountCredentialsEntity gatewayAccountCredentialsEntity = gatewayAccountCredentialsService.findCredentialFromCharge(charge, gatewayAccountEntity)
+                .orElseThrow(() -> new GatewayAccountCredentialsNotFoundException(format("Gateway account credentials for gateway account with id [%s] not found.", gatewayAccountEntity.getExternalId())));
```
stephencdaly, line 88: The exception message should maybe reference the charge rather than the gateway account just to avoid ambiguity over how we're trying to locate the credentials to use. Both the payment ID and the gateway account ID should be on the log line anyway thanks for the MDC.
Something like "Unable to determine gateway account credentials to use to refund charge" maybe?

stephencdaly (approved):
Looks good, minor comment about the exception message, but could be addressed in the next PR.

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
nimalank7: Yeah makes sense. I'll do that in a follow up PR.

PR #3179, opened 2021-08-19
PP-8342 - PaymentProvider implementation uses GatewayAccountCredentialsEntity

nimalank7:
Description:
- Stripe refund handler already uses request.getGatewayCredentials()
- Sandbox refunds aren't based on RefundRequest so no need to change
- Smartpay, Worldpay and Epdq are now changed
- No need to change tests as #3178 involved changing RefundRequest which required providing GatewayAccountCredentialsEntity.
In short the tests are already modified.
- Exception message in RefundService changed


stephencdaly (approved):
LGTM!

PR #3180, opened 2021-08-19
PP-8450 Add `send_reference_to_gateway` column to gateway_accounts table

kbottla:
## WHAT YOU DID
- We currently send `description` available on charge to Worldpay payment provider (authorisation payload field `description`) and some services require `reference` instead of description.
- Adds new field to gateway_accounts table which will be used to include payment `reference` (if send_reference_to_gateway is true) in authorisation payload or the payment `description`


stephencdaly (approved):
LGTM

PR #3181, opened 2021-08-19
PP-8450 Add send_reference_to_gateway to gateway account response

kbottla:
## WHAT YOU DID
- Added `send_reference_to_gateway` to get/search gateway accounts API responses
- Allows PATCH on new field


stephencdaly (approved):
LGTM!

PR #3185, opened 2021-08-20
PP-8450 Send reference to Worlday if flag is enabled on gateway account

kbottla:
## WHAT YOU DID
- Sends reference available on charge (instead of description) to Worldpay, if the flag `send_reference_to_gateway` is enabled on gateway account.


nimalank7 (approved):
LGTM!

PR #3193, opened 2021-08-24
PP-8409 Emit capture event for historic charges

stephencdaly:
We already emit a capture event for historic ePDQ charges, as we have seen times when we've received a capture success notification after the charge has been expunged.

We don't expect to have Worldpay and Smartpay charges expunged before we receive a capture notification, but there's no reason the behaviour shouldn't be the same if we did.

As we are reducing the time before expunge, this could also potentially make this more likely to happen.


SandorArpa (approved):
LGTM

PR #3206, opened 2021-09-01
PP-8441 add service id and live to event base classes and inheritors

gidsg:
## WHAT YOU DID
add service id and live to event base classes and inheritors.

Note the existing constructors without the new fields are retained where the fields are not available in the model for example `PaymentIncludedInPayout`, `RefundAvailabilityUpdated`, `RefundIncludedInPayout` and `PayoutUpdated`
TODO: in a future PR we should add these fields to models in future so these constructors can be retired.

## How to test

- How should it be reviewed? 
Code compiles and tests pass
- What feedback do you need? 
Do code changes make sense

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationErrorCheckedWithGatewayChargeWasMissing.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationErrorCheckedWithGatewayChargeWasMissing.java
@@ -3,7 +3,7 @@ package uk.gov.pay.connector.events.model.charge;
 import java.time.ZonedDateTime;
 
 public class AuthorisationErrorCheckedWithGatewayChargeWasMissing extends PaymentEventWithoutDetails {
-    public AuthorisationErrorCheckedWithGatewayChargeWasMissing(String resourceExternalId, ZonedDateTime timestamp) {
-        super(resourceExternalId, timestamp);
+    public AuthorisationErrorCheckedWithGatewayChargeWasMissing(String serviceId, boolean live, String resourceExternalId, ZonedDateTime timestamp) {
+        super(serviceId, live, resourceExternalId, timestamp);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/PaymentEvent.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/PaymentEvent.java
@@ -7,13 +7,23 @@ import uk.gov.pay.connector.events.model.ResourceType;
 import java.time.ZonedDateTime;
 
 public class PaymentEvent extends Event {
+    private String serviceId;
+    private boolean live;
     
+    public PaymentEvent(String serviceId, boolean live, String resourceExternalId, EventDetails eventDetails, ZonedDateTime timestamp) {
+        super(resourceExternalId, eventDetails, timestamp);
+        this.serviceId = serviceId;
+        this.live = live;
+    }
+
     public PaymentEvent(String resourceExternalId, EventDetails eventDetails, ZonedDateTime timestamp) {
         super(resourceExternalId, eventDetails, timestamp);
     }
 
-    public PaymentEvent(String resourceExternalId, ZonedDateTime timestamp) {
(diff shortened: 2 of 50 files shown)


src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationErrorCheckedWithGatewayChargeWasMissing.java
```diff
@@ -3,7 +3,7 @@
 import java.time.ZonedDateTime;
 
 public class AuthorisationErrorCheckedWithGatewayChargeWasMissing extends PaymentEventWithoutDetails {
-    public AuthorisationErrorCheckedWithGatewayChargeWasMissing(String resourceExternalId, ZonedDateTime timestamp) {
-        super(resourceExternalId, timestamp);
+    public AuthorisationErrorCheckedWithGatewayChargeWasMissing(String serviceId, boolean isLive, String resourceExternalId, ZonedDateTime timestamp) {
+        super(serviceId, isLive, resourceExternalId, timestamp);
```
alexbishop1: We seem to be inconsistent on whether we call it `live` or `isLive` — can we standardise on one or the other?

src/main/java/uk/gov/pay/connector/events/model/charge/PaymentEvent.java
```diff
@@ -7,13 +7,23 @@
 import java.time.ZonedDateTime;
 
 public class PaymentEvent extends Event {
+    private String serviceId;
+    private boolean live;
```
alexbishop1, line 11: I can’t think of a reason why these shouldn’t be `final`.

src/main/java/uk/gov/pay/connector/events/model/payout/PayoutEvent.java
```diff
@@ -7,6 +7,14 @@
 import java.time.ZonedDateTime;
 
 public class PayoutEvent extends Event {
+    private String serviceId;
+    private boolean live;
```
alexbishop1, line 11: Again, these can probably be `final`.

src/test/java/uk/gov/pay/connector/events/EventServiceTest.java
```diff
@@ -32,37 +32,37 @@
 
     @Test
     public void emitEvent() throws QueueException {
-        Event event = new PaymentEvent("external-id", now());
+        Event event = new PaymentEvent("service-id", true,"external-id", now());
```
alexbishop1: Missing a space.

alexbishop1 (changes requested):
Looks fine. Just a couple of minor of things.

src/test/java/uk/gov/pay/connector/events/EventServiceTest.java
gidsg: fixed in 802f246d8167ee6942c8871367c065a27db808f8

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationErrorCheckedWithGatewayChargeWasMissing.java
gidsg: fair point! Have standardised on `live` now in https://github.com/alphagov/pay-connector/pull/3206/commits/a39d53f0c054f4257ed1fe5f03bdac7bb360df5d.

src/main/java/uk/gov/pay/connector/events/model/payout/PayoutEvent.java
gidsg: Hmm, when I try to do this I get a compilation error:
```
variable live might not have been initialized
```
any ideas?

alexbishop1: OK, that’ll be because there’s the other constructor that doesn’t set them and we don’t explicitly set default values when the fields are declared. That makes perfect sense so leave them as non-`final` for now (we can reconsider if the other constructor goes away).

gidsg:
as suggested @alexbishop1 I have squashed the commits addressing the PR comments.

alexbishop1 (approved):

PR #3218, opened 2021-09-07
PP-8461 add pact test for 3ds info obtained event

SandorArpa:
## WHAT YOU DID
- add a new queue pact test for 3ds info obtained message


kbottla (approved):
LGTM!

PR #3222, opened 2021-09-08
PP-8461 Add events_details to GatewayRequires3dsAuthorisation event

kbottla:
## WHAT
- Added event details to GatewayRequires3dsAuthorisation event which emits 3DS version (if available) and also `requires_3ds` field. Currently ledger doesn't have any info on transaction indicating that payment has gone through 3DS and the new `requires_3ds` field can be used by Ledger to provide following `authorisation_summary` (get one/search payment)
	```
	“authorisation_summary”: {
	  "three_d_secure": {
	      "required": true,
	      "version": "2.2.0/2.1.0/1.0.2"
	  }
	}
	```
- Also updated `Gateway3dsInfoObtained` to use new PaymentEvent constructor (which uses service_id and live flag)


SandorArpa (approved):
LGTM

PR #3225, opened 2021-09-09
PP-8441 add serviceId and live fields to Charge model

gidsg:
## WHAT YOU DID
Follow up to #3206 adding `serviceId` and `live` fields to Charge model. Remove some constructors which are no longer needed as we can use these fields from the model. 

## How to test

Code should compile
Tests should make sense

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/CaptureConfirmedByGatewayNotification.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/CaptureConfirmedByGatewayNotification.java
@@ -3,7 +3,7 @@ package uk.gov.pay.connector.events.model.charge;
 import java.time.ZonedDateTime;
 
 public class CaptureConfirmedByGatewayNotification extends PaymentEventWithoutDetails {
-    public CaptureConfirmedByGatewayNotification(String resourceExternalId, ZonedDateTime timestamp) {
-        super(resourceExternalId, timestamp);
+    public CaptureConfirmedByGatewayNotification(String serviceId, boolean live, String resourceExternalId, ZonedDateTime timestamp) {
+        super(serviceId, live, resourceExternalId, timestamp);
     }
 }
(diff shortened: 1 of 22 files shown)


src/main/java/uk/gov/pay/connector/events/model/charge/CaptureConfirmedByGatewayNotification.java
```diff
@@ -3,7 +3,7 @@
 import java.time.ZonedDateTime;
 
 public class CaptureConfirmedByGatewayNotification extends PaymentEventWithoutDetails {
-    public CaptureConfirmedByGatewayNotification(String resourceExternalId, ZonedDateTime timestamp) {
-        super(resourceExternalId, timestamp);
+    public CaptureConfirmedByGatewayNotification(String serviceId, boolean live, String resourceExternalId, ZonedDateTime timestamp) {
+        super(serviceId, live,resourceExternalId, timestamp);
```
alexbishop1: Missing a space after the comma.

src/test/java/uk/gov/pay/connector/it/base/ChargingITestBase.java
```diff
@@ -70,6 +70,8 @@
     protected static final String EMAIL = randomAlphabetic(242) + "@example.com";
     protected static final long AMOUNT = 6234L;
 
+    protected static final String serviceId = "external-service-id";
```
alexbishop1: Convention dictates that this variable name should be in `SCREAMING_SNAKE_CASE`.

alexbishop1 (changes requested):
Looks good — just a couple of minor things.

gidsg:
@alexbishop1 Thanks, I have addressed your comments by amending commits and force-pushing.

alexbishop1 (approved):

PR #3239, opened 2021-09-15
PP-8440 Set default service id on charge fixture

mrlumbu:
Since we now expect this to be set on Charge creation, seems ok to set a default value in fixture. Also has happy side effect of making all the pacts correct.


alexbishop1 (approved):
That’s so random!

PR #3242, opened 2021-09-20
PP-8628 Send capture success details for forced capture event

stephencdaly:
When we receive a capture success notification, but the charge is in an unexpected state, we ignore the current state and update the status to captured and emit an event to ledger to do this.

However, this event did not include the capture details such as the captured date. We were also not storing the gateway_event_date for the event in connector so the capture event was derived from the updated date for the event rather than the date from the notification.

Fix both this by storing the gateway_event_date and attaching the same event details we attach for the standard CAPTURE_CONFIRMED event to the STATUS_CORRECTED_TO_CAPTURED_TO_MATCH_GATEWAY_STATUS event.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -703,17 +703,17 @@ public class ChargeService {
     @Transactional
     public ChargeEntity forceTransitionChargeState(String chargeExternalId, ChargeStatus targetChargeState) {
         return chargeDao.findByExternalId(chargeExternalId).map(chargeEntity ->
-                forceTransitionChargeState(chargeEntity, targetChargeState))
+                forceTransitionChargeState(chargeEntity, targetChargeState, null))
                 .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeExternalId));
     }
 
     @Transactional
-    public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus targetChargeState) {
+    public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus targetChargeState, ZonedDateTime gatewayEventDate) {
         ChargeStatus fromChargeState = ChargeStatus.fromString(charge.getStatus());
 
         return PaymentGatewayStateTransitions.getEventForForceUpdate(targetChargeState).map(eventClass -> {
             charge.setStatusIgnoringValidTransitions(targetChargeState);
-            ChargeEventEntity chargeEventEntity = chargeEventDao.persistChargeEventOf(charge);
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceTest.java
```diff
@@ -252,14 +251,15 @@ public void forcingChargeToCapturedState_shouldSucceedAndEmitEvent() {
...
-        ChargeEntity updatedCharge = service.forceTransitionChargeState(charge, CAPTURED);
+        ZonedDateTime gatewayEventDate = ZonedDateTime.parse("2021-01-01T01:30:00.000Z");
+        ChargeEntity updatedCharge = service.forceTransitionChargeState(charge, CAPTURED, gatewayEventDate);
 
         ArgumentCaptor<ChargeEntity> chargeEntityArgumentCaptor = ArgumentCaptor.forClass(ChargeEntity.class);
-        verify(mockedChargeEventDao).persistChargeEventOf(chargeEntityArgumentCaptor.capture());
+        verify(mockedChargeEventDao).persistChargeEventOf(chargeEntityArgumentCaptor.capture(), eq(gatewayEventDate));
         assertThat(chargeEntityArgumentCaptor.getValue().getStatus(), is(CAPTURED.getValue()));
-        assertThat(updatedCharge.getStatus(), is(CAPTURED.getValue()));
+        assertThat(updatedCharge.getStatus(), is(CAPTURED.getValue()));;
```
kbottla, line 262: Extra `;`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -703,17 +703,17 @@ public ChargeEntity transitionChargeState(String chargeExternalId, ChargeStatus
...
     }
 
     @Transactional
-    public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus targetChargeState) {
+    public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus targetChargeState, ZonedDateTime gatewayEventDate) {
         ChargeStatus fromChargeState = ChargeStatus.fromString(charge.getStatus());
 
         return PaymentGatewayStateTransitions.getEventForForceUpdate(targetChargeState).map(eventClass -> {
             charge.setStatusIgnoringValidTransitions(targetChargeState);
-            ChargeEventEntity chargeEventEntity = chargeEventDao.persistChargeEventOf(charge);
```
kbottla, line 716: Looks like `ChargeEventDao.persistChargeEventOf(ChargeEntity chargeEntity)` can be removed as it is not being used anywhere. can be done in a separate PR.

kbottla (approved):

PR #3248, opened 2021-09-24
PP-8628 Don't overwrite issuerUrl when processing Stripe 3ds repsonse

stephencdaly:
We were overwriting the issuerUrl when processing the Stripe 3DS response. This meant that post-authorisation the Auth3dsRequiredEntity on the ChargeEntity was null, and so we weren't including the `authorisation_summary` in charge responses from connector for charges that have been through 3DS if 3DS authorisation is rejected.

Delete Stripe3dsSourceAuthorisationResponse.java as it is unused.


SandorArpa (approved):
LGTM

nimalank7 (approved):
LGTM!

PR #3255, opened 2021-09-30
PP-8540 log the remote IP address when authorising a payment

gidsg:
## WHAT YOU DID
Log IP address in authorisation request log message and structured logging

## How to test
Tests should make sense

Application should log IP address in log message and structured logging when authorising a payment

**Question** Is it preferable to return a `null` value in structured logging if no IP present? Currently, the key value is not included in this case.

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/util/AuthorisationRequestSummaryStringifier.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/util/AuthorisationRequestSummaryStringifier.java
@@ -2,6 +2,7 @@ package uk.gov.pay.connector.gateway.util;
 
 import uk.gov.pay.connector.gateway.model.AuthorisationRequestSummary;
 
+import java.util.Optional;
 import java.util.StringJoiner;
 
 public class AuthorisationRequestSummaryStringifier {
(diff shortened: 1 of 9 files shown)


gidsg:
OK @alexbishop1 as discussed off-of-Github, I've removed the Optional field but used it when serialising. If we're happy with this approach I can squash the commits!

src/main/java/uk/gov/pay/connector/gateway/util/AuthorisationRequestSummaryStringifier.java
```diff
@@ -1,7 +1,9 @@
 package uk.gov.pay.connector.gateway.util;
 
+import org.apache.commons.cli.Option;
```
alexbishop1: Suspicious import 🧐

alexbishop1 (changes requested):
I think it’s fine to not include anything if there isn’t an IP address (which I don’t think will happen in practice). Please remove the rogue import and squash the commits.

src/main/java/uk/gov/pay/connector/gateway/util/AuthorisationRequestSummaryStringifier.java
gidsg: @alexbishop1 Good spot 👀  have removed and squashed in 36c3c60

alexbishop1 (approved):

PR #3260, opened 2021-10-07
PP-8574 - Remove legacy Stripe integration

kbottla:
## WHAT
- Our initial integration with Stripe was based on creating tokens and using sources.
  We have switched to using payment methods and payment intents to support 3DS2 so sources based integration is no longer needed.
- Only exception is refunds where we need to support refunding old payments authorised using sources.
- Fixed tests to use payment intent templates instead of sources
- Also removed unused test templates


SandorArpa (approved):
LOKTM

PR #3276, opened 2021-10-26
PP-8732 Migrate events `live` property to boxed Boolean value

sfount:
Allow `null` values for the `live` property on events as some events
still won't be setting the `live` context. This is causing `live` to be
incorrectly coalesced to `false` when serialised by Jackson.

Co-authored-by: Gideon Goldberg <1764158+gidsg@users.noreply.github.com>


gidsg (approved):
👍  🆙 🅱️

PR #3314, opened 2021-11-30
P-8946 add Stripe's Radar and 3ds fees to config

SandorArpa:
## WHAT YOU DID
- With the new contract there are additional fees for successful transactions.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
@@ -52,6 +52,12 @@ public class StripeGatewayConfig extends Configuration {
     @NotNull
     private List<String> credentials;
 
+    @Valid
+    private int radarFeeInPence;
+
+    @Valid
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
```diff
@@ -52,6 +52,12 @@
     @NotNull
     private List<String> credentials;
 
+    @Valid
+    private Double radarFeeInPence;
```
kbottla: Any reason this is a Double rather than int ?

src/main/java/uk/gov/pay/connector/app/StripeGatewayConfig.java
```diff
@@ -52,6 +52,12 @@
...
+
+    @Valid
+    private Double threeDsFeeInPence;
```
kbottla: Same as above

kbottla (approved):

PR #3323, opened 2021-12-06
PP-8947 - Add fee_type field to fees table

nimalank7:
Description:
- Add fee_type field to fees table for Stripe contract changes
- This is so we can start tracking the type of fees


SandorArpa (approved):
LOKTM

PR #3330, opened 2021-12-09
PP-8960 Add task to collect fee for failed payments to queue

stephencdaly:
Upon a charge state transition, add a
"collect_fee_for_stripe_failed_payment" task to the payment task queue
if all of the following are true:
- the charge is in a terminal state
- the charge is failed
- the charge is a Stripe charge
- the charge has a gateway_transaction_id set

Used the `isExpungeable` property of ChargeStatus to determine whether
the status is terminal or not, as this is true only for terminal
statuses.


SandorArpa (approved):
LOKTM

PR #3331, opened 2021-12-09
PP-8960 Do not add task to queue when env var disabled

stephencdaly:
We need to check whether the environment variable is enabled before adding tasks to collect Stripe fees for failed payments to the task queue.


SandorArpa (dismissed):
LOKTM

kbottla (approved):

PR #3332, opened 2021-12-09
PP-8947 - Drop unique constraint on charge_id column in fees

nimalank7:
Description:
- As part of this story we want to allow multiple fees for a single charge so we have to drop the
UNIQUE constraint on charge_id to allow this in the fees table


kbottla (approved):

stephencdaly (approved):
LGTM

PR #3336, opened 2021-12-10
PP-8945 Add GET method to gateway client

stephencdaly:
Add method to make a GET request to the GatewayClient. This will be used to retrieve the payment intent from Stripe when calculating fees for failed payments.

Use a similar metric name as used for POST requests, but add `.get` before the gateway name.


kbottla (approved):
LGTM!

PR #3337, opened 2021-12-10
PP-8947 - Add FeeIncurredEvent

nimalank7:
Description:
- Adds FeeIncurrentEvent

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
@@ -1,0 +1,11 @@
+package uk.gov.pay.connector.events.model.charge;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+import java.time.ZonedDateTime;
+
+public class FeeIncurredEvent extends PaymentEvent {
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.pay.connector.events.model.charge;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+import java.time.ZonedDateTime;
+
+public class FeeIncurredEvent extends PaymentEvent {
+    public FeeIncurredEvent(String resourceExternalId, EventDetails eventDetails, ZonedDateTime timestamp) {
+        super(resourceExternalId, eventDetails, timestamp);
```
kbottla, line 9: Are we going to pass timestamp available on Fee list ? If not we need to get earliest timestamp from fee list on charge entity.

kbottla (approved):

src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
nimalank7: Yeah we'll have to get the earliest timestamp from the list by iterating through etc...

PR #3342, opened 2021-12-13
PP-8945 Use GET request when getting payment intent for refund

stephencdaly:
We have added the ability to make a GET request to the GatewayClient. Use this, rather than a POST request when retrieving the payment intent from Stripe when making a refund.

Have tested this locally, connecting to the Stripe test environment and the refund goes through successfully.


kbottla (approved):
LGTM!

PR #3343, opened 2021-12-13
PP-8947 - Refactors FeeEntity

nimalank7:
Description:
- Uses Instant rather than `ZonedDateTime` in the same way that `ChargeEntity` does for createdDate
- Added a test for `FeeIncurredEventDetails`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
@@ -1,11 +1,33 @@
 package uk.gov.pay.connector.events.model.charge;
 
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.FeeEntity;
 import uk.gov.pay.connector.events.eventdetails.EventDetails;
+import uk.gov.pay.connector.events.eventdetails.charge.FeeIncurredEventDetails;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
```diff
@@ -1,11 +1,32 @@
...
     }
+
+    public static FeeIncurredEvent from(ChargeEntity charge) {
+        Instant earliestInstant = charge.getFees()
+                .stream()
+                .map(FeeEntity::getCreatedDate)
+                .min(Instant::compareTo)
+                .orElse(null);
+
+        ZonedDateTime earliestCreatedDate = earliestInstant != null ? ZonedDateTime.ofInstant(earliestInstant, ZoneId.of("UTC")) : null;
```
stephencdaly: If there are no fees, should we return an event at all? Surely there's never an occurrence where we're want to return the event with no created date, if there are no fees? Should it return null or throw an exception?

stephencdaly (commented):
LGTM, just a question about what we should return for the event if there are no fees available

src/main/java/uk/gov/pay/connector/events/model/charge/FeeIncurredEvent.java
nimalank7: Yep we've refactored it to throw an exception.

stephencdaly (approved):
LGTM

PR #3350, opened 2021-12-14
PP-8947 - Record FEE_INCURRED_EVENT to emitted_events table

nimalank7:
Description:
- This records the FEE_INCURRED_EVENT emission in the emitted_events table
- Previously the FEE_INCURRED_EVENT wasn't being created because there were no fees present on the ChargeEntity due to the @Transactional annotation.
This PR ensures that the fees are added whilst being persisted to the Fees table so allowing for the event to be created
- Verified locally that the event was both created and emitted and that the fees were present for V2 charge
- Paired with @SandorArpa


stephencdaly (approved):
LGTM!

PR #3355, opened 2021-12-16
PP-8945 Calculate and transfer fees for failed payments

stephencdaly:
When a "collect_fee_for_stripe_failed_payment" task is processed on the task queue, calculate the fee due by determining whether or not the payment went through 3DS. If it went through 3DS, both the 3DS and Radar fees apply. If the payment did not go through 3DS, only the Radar fee applies.

Make a request to Stripe to transfer the amount for the failed fee from the connect account to the platform account.

Note that the method used to determine whether the payment has been through 3DS does not currently work for payments that failed 3DS as there will be no charge in Stripe. We'll need to change how we do this when
we get more information from them, but leaving this as a placeholder for now. This methodology is therefore not tested as fully as would be liked as it is expected to change.

This commit does not deal with storing the fees in the database and emitting events to ledger.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
@@ -1,0 +1,95 @@
+package uk.gov.pay.connector.gateway.stripe.handler;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.StripeGatewayConfig;
+import uk.gov.pay.connector.charge.model.domain.Charge;
+import uk.gov.pay.connector.gateway.GatewayClient;
+import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.model.request.GatewayClientRequest;
+import uk.gov.pay.connector.gateway.stripe.json.StripeCharge;
+import uk.gov.pay.connector.gateway.stripe.json.StripePaymentIntent;
+import uk.gov.pay.connector.gateway.stripe.json.StripeTransferResponse;
+import uk.gov.pay.connector.gateway.stripe.request.StripeGetPaymentIntentRequest;
+import uk.gov.pay.connector.gateway.stripe.request.StripeTransferInRequest;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialsEntity;
+import uk.gov.pay.connector.util.JsonObjectMapper;
+
+import java.util.List;
+import java.util.Optional;
+
+public class StripeFailedPaymentFeeCollectionHandler {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(StripeFailedPaymentFeeCollectionHandler.class);
+    
+    private final GatewayClient client;
+    private final StripeGatewayConfig stripeGatewayConfig;
+    private final JsonObjectMapper jsonObjectMapper;
+
+    public StripeFailedPaymentFeeCollectionHandler(
+            GatewayClient gatewayClient,
+            StripeGatewayConfig stripeGatewayConfig,
+            JsonObjectMapper jsonObjectMapper) {
+        this.client = gatewayClient;
+        this.stripeGatewayConfig = stripeGatewayConfig;
+        this.jsonObjectMapper = jsonObjectMapper;
+    }
+
+    public void calculateAndTransferFees(Charge charge, GatewayAccountEntity gatewayAccount, GatewayAccountCredentialsEntity gatewayAccountCredentials)
+            throws GatewayException {
+        // TODO: no Stripe charge exists if the 3DS attempt is failed, so this method of determining whether a payment has been through 3DS will need to change
+        boolean threeDsFeeApplicable = getStripeCharge(charge, gatewayAccount, gatewayAccountCredentials)
+                .map(this::isThreeDsFeeApplicable)
+                .orElse(false);
+        int fee = stripeGatewayConfig.getRadarFeeInPence();
+        if (threeDsFeeApplicable) {
+            fee += stripeGatewayConfig.getThreeDsFeeInPence();
+        }
+
+        transferFeeFromConnectAccount(fee, charge, gatewayAccount, gatewayAccountCredentials);
+    }
+
+    private boolean isThreeDsFeeApplicable(StripeCharge stripeCharge) {
+        return stripeCharge.getPaymentMethodDetails() != null &&
+                stripeCharge.getPaymentMethodDetails().getCard() != null &&
+                stripeCharge.getPaymentMethodDetails().getCard().getThreeDSecure() != null &&
+                stripeCharge.getPaymentMethodDetails().getCard().getThreeDSecure().getAuthenticated();
+    }
+
+    private Optional<StripeCharge> getStripeCharge(Charge charge, GatewayAccountEntity gatewayAccount, GatewayAccountCredentialsEntity gatewayAccountCredentials) throws GatewayException.GatewayErrorException, GatewayException.GenericGatewayException, GatewayException.GatewayConnectionTimeoutException {
+        StripePaymentIntent paymentIntent = getPaymentIntent(charge, gatewayAccount, gatewayAccountCredentials);
+        List<StripeCharge> charges = paymentIntent.getChargesCollection().getCharges();
+        if (charges.size() > 1) {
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
```diff
@@ -31,20 +75,81 @@ public void processMessages() throws QueueException {
...
-                        e.getMessage()),
+                                paymentTaskMessage.getQueueMessageId(),
+                                e.getMessage()),
                         kv(PAYMENT_EXTERNAL_ID, paymentTaskMessage.getPaymentExternalId())
                 );
             }
         }
     }
+
+    private void calculateAndTransferFees(PaymentTaskMessage paymentTaskMessage)
```
kbottla: Can we move transfer handling to StripePaymentProvider or to a separate Stripe transfer handler similar to stripe capture handler (which does capture, calculates fee and a transfer) ? Given we want task queue to be  generic would be good to keep the logic out of TaskQueueMessageHandler.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
```diff
@@ -31,20 +75,81 @@ public void processMessages() throws QueueException {
...
+        int fee = stripeGatewayConfig.getRadarFeeInPence();
+        if (threeDsFeeApplicable) {
+            fee += stripeGatewayConfig.getThreeDsFeeInPence();
+        }
+
+        transferFeeFromConnectAccount(fee, charge, gatewayAccount, gatewayAccountCredentials);
+    }
+
+    private boolean isThreeDsFeeApplicable(StripeCharge stripeCharge) {
+        return stripeCharge.getPaymentMethodDetails().getCard().getThreeDSecure().getAuthenticated();
```
kbottla: This will throw NPE if charge has not gone through 3DS. We need checks similar to this (https://github.com/alphagov/pay-connector/blob/f09fe43f8de74343aa53518e277ad6fc76520df5/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java#L238-L242)

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
stephencdaly: Yep good idea

stephencdaly: Added a new handler class for this that is called through StripePaymentProvider, so we don't have to use a separate GatewayClient. Not 100% sure about that approach though, so open to different suggestions.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
```diff
@@ -0,0 +1,94 @@
...
+            fee += stripeGatewayConfig.getThreeDsFeeInPence();
+        }
+
+        transferFeeFromConnectAccount(fee, charge, gatewayAccount, gatewayAccountCredentials);
+    }
+
+    private boolean isThreeDsFeeApplicable(StripeCharge stripeCharge) {
+        return stripeCharge.getPaymentMethodDetails() != null &&
+                stripeCharge.getPaymentMethodDetails().getCard() != null &&
+                stripeCharge.getPaymentMethodDetails().getCard().getThreeDSecure().getAuthenticated();
```
kbottla, line 57: Can we also check that `getThreeDSecure()` is not null ?

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
```diff
@@ -0,0 +1,94 @@
...
+            GatewayClient gatewayClient,
+            StripeGatewayConfig stripeGatewayConfig,
+            JsonObjectMapper jsonObjectMapper) {
+        this.client = gatewayClient;
+        this.stripeGatewayConfig = stripeGatewayConfig;
+        this.jsonObjectMapper = jsonObjectMapper;
+    }
+
+    public void calculateAndTransferFees(Charge charge, GatewayAccountEntity gatewayAccount, GatewayAccountCredentialsEntity gatewayAccountCredentials)
+            throws GatewayException.GatewayErrorException, GatewayException.GenericGatewayException, GatewayException.GatewayConnectionTimeoutException {
```
kbottla: Any reason we are throwing multiple exceptions instead of only GatewayException ?

src/test/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandlerTest.java
```diff
@@ -0,0 +1,142 @@
...
+    @Test
+    public void shouldTransferFeeForChargeThatHasNotBeenThrough3ds() throws Exception {
+        mockGetPaymentIntentThatHasNotBeenThrough3dsSuccess();
+        setupCommonMocksForCollectFeeSuccess();
+
+        stripeFailedPaymentFeeCollectionHandler.calculateAndTransferFees(charge, gatewayAccountEntity, gatewayAccountCredentialsEntity);
+        verifyTransferRequestPayload(7);
+    }
+
+    private void verifyTransferRequestPayload(int i) throws GatewayException.GenericGatewayException, GatewayException.GatewayErrorException, GatewayException.GatewayConnectionTimeoutException {
```
kbottla:
```suggestion
    private void verifyTransferRequestPayload(int amount) throws GatewayException.GenericGatewayException, GatewayException.GatewayErrorException, GatewayException.GatewayConnectionTimeoutException {
```

src/test/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandlerTest.java
```diff
@@ -0,0 +1,142 @@
...
+import static uk.gov.pay.connector.charge.model.domain.ChargeEntityFixture.aValidChargeEntity;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntityFixture.aGatewayAccountEntity;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialsEntityFixture.aGatewayAccountCredentialsEntity;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.STRIPE_GET_PAYMENT_INTENT_WITH_3DS_AUTHORISED_RESPONSE;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.STRIPE_PAYMENT_INTENT_SUCCESS_RESPONSE;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.STRIPE_TRANSFER_RESPONSE;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.load;
+
+@ExtendWith(MockitoExtension.class)
+class StripeFailedPaymentFeeCollectionHandlerTest {
```
kbottla, line 44: Can we also test exception handling ? ie., when there are multiple charges, no charge on payment intent or if the payment intent is not found.

kbottla (commented):
Looks good to me. Added a few minor comments

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
stephencdaly: Changed this to only throw GatewayException

src/test/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandlerTest.java
stephencdaly: 👍 I think this was intellij's doing when extracting a method and I didn't check.

src/test/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandlerTest.java
stephencdaly: Added one test for when there are multiple charges.
Can't throw an exception when there isn't a charge on a payment intent as this is a valid scenario we will see.
If the payment intent is not found, the GatewayClient will throw a GatewayException, wasn't sure how much use a test would be just that the exception is passed on? Maybe could be added as an integration test later?

kbottla (approved):
LGTM!

PR #3356, opened 2021-12-16
PP-8947 - Fix CaptureConfirmed Event to properly check fees

nimalank7:
Description:
- From locally testing it looks like Google Guice @transactional will persist the charge (i.e. flush the EntityManager changes) so no need to explicitly
 call feeDao.persist(...). Adding CascadeType.Persist and then setting the FeeEntity on the charge will ensure it is flushed to the database (tested locally
 and the fees were correctly persisted)
- Previously the CaptureConfirmed event would always emit the fee because only after the code progressed would the fees actually be persisted. This meant that the fee
check on CaptureConfirmed event would always be 0. Refactoring the code and ensuring the fees are added to the ChargeEntity before emitted the event removes the bug
- Added a service integration test to verify that the fees are persisted into the database
- Paired with @SandorArpa and @nlsteers

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -30,6 +30,7 @@ import uk.gov.service.payments.commons.model.charge.ExternalMetadata;
 
 import javax.persistence.Access;
 import javax.persistence.AccessType;
+import javax.persistence.CascadeType;
 import javax.persistence.Column;
 import javax.persistence.Convert;
 import javax.persistence.Embedded;
@@ -115,7 +116,7 @@ public class ChargeEntity extends AbstractVersionedEntity implements Nettable {
     @JoinColumn(name = "gateway_account_credential_id", updatable = false)
     private GatewayAccountCredentialsEntity gatewayAccountCredentialsEntity;
 
-    @OneToMany(mappedBy = "chargeEntity")
+    @OneToMany(mappedBy = "chargeEntity", cascade = CascadeType.PERSIST)
     private List<FeeEntity> fees = new ArrayList<>();
 
     @OneToMany(mappedBy = "chargeEntity")
@@ -453,7 +454,7 @@ public class ChargeEntity extends AbstractVersionedEntity implements Nettable {
         this.corporateSurcharge = corporateSurcharge;
     }
 
-    public void setFee(FeeEntity fee) {
+    public void addFee(FeeEntity fee) {
         this.fees.add(fee);
     }
 
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
@@ -14,8 +14,6 @@ import uk.gov.pay.connector.events.EventService;
 import uk.gov.pay.connector.events.exception.EventCreationException;
 import uk.gov.pay.connector.events.model.Event;
 import uk.gov.pay.connector.events.model.charge.FeeIncurredEvent;
-import uk.gov.pay.connector.fee.dao.FeeDao;
-import uk.gov.pay.connector.fee.model.Fee;
 import uk.gov.pay.connector.gateway.CaptureResponse;
 import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
@@ -45,7 +43,6 @@ public class CardCaptureService {
     private static final Logger LOG = LoggerFactory.getLogger(CardCaptureService.class);
 
     private final UserNotificationService userNotificationService;
-    private final FeeDao feeDao;
     private final ChargeService chargeService;
     private final PaymentProviders providers;
     protected final Logger logger = LoggerFactory.getLogger(getClass());
@@ -56,7 +53,6 @@ public class CardCaptureService {
 
     @Inject
     public CardCaptureService(ChargeService chargeService,
-                              FeeDao feeDao,
                               PaymentProviders providers,
                               UserNotificationService userNotificationService,
                               Environment environment,
@@ -64,7 +60,6 @@ public class CardCaptureService {
                               CaptureQueue captureQueue,
                               EventService eventService) {
         this.chargeService = chargeService;
-        this.feeDao = feeDao;
         this.providers = providers;
         this.metricRegistry = environment.metrics();
         this.clock = clock;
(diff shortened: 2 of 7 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -127,13 +122,13 @@ public void processGatewayCaptureResponse(String chargeId, String oldStatus, Cap
...
+        ChargeEntity charge = chargeService.findChargeByExternalId(chargeId);
 
         captureResponse.getFeeList().ifPresent(feeList -> {
-                    feeList.forEach(fee ->
-                            persistFee(charge, fee)
-                    );
+                    feeList.forEach(fee -> {
+                        FeeEntity feeEntity = new FeeEntity(charge, clock.instant(), fee.getAmount(), fee.getFeeType());
+                        charge.setFee(feeEntity);
+                    });
```
alexbishop1: This could be:

```java
feeList.stream().map(fee -> new FeeEntity(charge, clock.instant(), fee.getAmount(), fee.getFeeType())).forEach(charge::setFee);
```

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -115,7 +116,7 @@
     @JoinColumn(name = "gateway_account_credential_id", updatable = false)
     private GatewayAccountCredentialsEntity gatewayAccountCredentialsEntity;
 
-    @OneToMany(mappedBy = "chargeEntity")
+    @OneToMany(mappedBy = "chargeEntity", cascade = CascadeType.PERSIST)
```
alexbishop1, line 119: I’m a little concerned that this is the only time in this class where we use `cascade = CascadeType.PERSIST`. It may be that’s legitimate but are we sure we’re not deviating from established patterns?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
alexbishop1: This means we’re now sending the event for the fee before we’ve persisted the fee. This makes me feel a little uncomfortable but I see we already do this with the `Gateway3dsInfoObtained` so maybe it’s OK.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
```diff
@@ -227,7 +224,6 @@ public void doCapture_shouldCaptureAChargeForStripeAccountV1() {
         verify(mockedPaymentProvider, times(1)).capture(request.capture());
         assertThat(request.getValue().getTransactionId(), is(gatewayTxId));
 
-        verify(mockFeeDao, times(1)).persist(any());
```
alexbishop1, line 230: Without something like this it seems we’re not testing that we’re persisting the fees any more. Do we have an integration test that checks this? If not, might we want to consider writing one?

alexbishop1 (changes requested):
Requested a couple of changes. I’m convinced that reordering the code so that we persist the fees before we emit the captured event is the right thing to do but I’m less sure about the changes to how we actually persist fees.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
```diff
@@ -127,13 +122,13 @@ public void processGatewayCaptureResponse(String chargeId, String oldStatus, Cap
...
-        ChargeEntity charge = chargeService.updateChargePostCapture(chargeId, nextStatus);
+        ChargeEntity charge = chargeService.findChargeByExternalId(chargeId);
 
         captureResponse.getFeeList().ifPresent(feeList -> {
-                    feeList.forEach(fee ->
-                            persistFee(charge, fee)
-                    );
+                    feeList.forEach(fee -> {
+                        FeeEntity feeEntity = new FeeEntity(charge, clock.instant(), fee.getAmount(), fee.getFeeType());
+                        charge.setFee(feeEntity);
```
kbottla: we should rename `setFee` on chargeEntity. `setFee` got me thinking that we are setting one fee entity on the charge. May be we rename it to `addFee` or something more clear

kbottla (commented):
I think the fee list are not available when CaptureConfirmed event because of order of events when processing capture response

1. updateChargePostCapture() is called first which will 
     a.  update charge status to CAPTURED 
     b.  adds transition to state transition queue which emits CAPTURE_CONFIRMED event after a certain delay. so possible that CAPTURE_CONFIRMED is emitted before/while fee entities are being persisted. 
2. Fee is persisted and added to charge entity

So changing order - persisting fee , adding feeEntities to charge, and updating charge status (`updateChargePostCapture `) last could work.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
nimalank7: Done.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceTest.java
nimalank7: Written a new service integration test.

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
nimalank7: This is the only way to persist the fees via the ChargeEntity.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java
nimalank7: Made the change.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureServiceIT.java
```diff
@@ -0,0 +1,68 @@
...
+        String transactionId = "transactionId";
+        List<Fee> feeList = List.of(Fee.of(RADAR, 10L), Fee.of(THREE_D_S, 20L), Fee.of(TRANSACTION, 480L));
+        CaptureResponse captureResponse = new CaptureResponse(transactionId, CaptureResponse.ChargeState.COMPLETE, 100L, feeList);
+
+        // Trigger the post gateway capture response programmatically which normally would be invoked by the scheduler.
+        testContext.getInstanceFromGuiceContainer(CardCaptureService.class).processGatewayCaptureResponse(externalChargeId, oldChargeStatus, captureResponse);
+
+        List<Map<String, Object>> listOfFees = databaseTestHelper.getFeesByChargeId(chargeId);
+        List<String> feeTypeColumnValues = listOfFees.stream().map(map -> map.get("fee_type").toString()).collect(Collectors.toList());
+        List<String> feeAmountColumnValues = listOfFees.stream().map(map -> map.get("amount_collected").toString()).collect(Collectors.toList());
```
alexbishop1, line 61: 🗺

alexbishop1 (approved):

PR #3375, opened 2021-12-24
PP-8696: Remove unneeded stages from Jenkinsfile

oswaldquek:
Given Jenkins doesn't do deployment anymore, these stages aren't needed.


oswaldquek:
Will only merge from 4th January 2022 onwards.

katstevens (approved):

PR #3376, opened 2022-01-04
PP-8945 Store fees and send event for failed payments

stephencdaly:
When processing Stripe failed payments, after transferring the applicable amount for Stripe fees from the Stripe connect account, store entries in the `fees` table for each of the fee types that were applied. Send a FEE_INCURRED event to ledger containing the fee breakdown.

Always expect that the charge exists in the connector database, and throw an exception if it has been expunged. We will separately add a check that will stop Stripe charges from being expunged if they have not had the applicable fees taken.

Stop using StripePaymentProvider and call StripeFailedPaymentFeeCollectionHandler directly from
TaskQueueMessageHandler to simplify the code. To allow for this, added a new GatewayOperation enum entry so a separate Stripe client can be created in StripeFailedPaymentFeeCollectionHandler.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
@@ -3,7 +3,10 @@ package uk.gov.pay.connector.gateway.stripe.handler;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
-import uk.gov.pay.connector.charge.model.domain.Charge;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.FeeType;
+import uk.gov.pay.connector.charge.util.StripeFeeCalculator;
+import uk.gov.pay.connector.fee.model.Fee;
 import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.model.request.GatewayClientRequest;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeFailedPaymentFeeCollectionHandler.java
```diff
@@ -1,53 +1,93 @@
...
     private final StripeGatewayConfig stripeGatewayConfig;
     private final JsonObjectMapper jsonObjectMapper;
+    private final Clock clock;
+    private final EventService eventService;
+    private final ChargeDao chargeDao;
 
+    @Inject
     public StripeFailedPaymentFeeCollectionHandler(
-            GatewayClient gatewayClient,
+            GatewayClientFactory gatewayClientFactory,
```
kbottla: Any reason db operations and event emitting is done here ? Looks like StripeFailedPaymentFeeCollectionHandler is responsible for all operations (stripe, db, and emitting events) and doesn't feel like this should be under a gateway specific package gateway/stripe/handler/... . May be a service layer that can use StripePaymentProvider for gateway operation and does rest of the operations ?

stephencdaly: updated to separate out the gateway operations

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeGetPaymentIntentRequest.java
```diff
@@ -2,6 +2,7 @@
 
 import uk.gov.pay.connector.app.StripeGatewayConfig;
 import uk.gov.pay.connector.charge.model.domain.Charge;
```
kbottla, line 4: Few unused imports

kbottla (approved):
LGTM!

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeGetPaymentIntentRequest.java
stephencdaly: Thanks, will remove in next PR

PR #3382, opened 2022-01-05
PP-8945 Simplify fee calculation

stephencdaly:
Simplify the fee calculation for the fees for successful payments to only determine in one place which fees apply. Before we were doing this in two places - when getting the fee list and getting the total fee, which carries the risk of the code getting out of sync in the two places.

To get the total fee, we now sum the amounts of all the fees in the fee list returned from the calculator.

In addition, remove the total fee amount from the CaptureResponse, as this was unused other than in tests.

Also, make CaptureResponse.getFeeList return a `List<Fee>` rather than an `Optional<List<Fee>>`, as in some cases it was returning an empty Optional, and in other cases, an Optional containing an empty List. It is simpler to just always return an empty List in cases where there are no fees.


SandorArpa (approved):
LOKTM

PR #3385, opened 2022-01-07
PP-8945 Alter net amount calculation to account for failed payments

stephencdaly:
Removed the Nettable interface, which was previously implemented by
multiple classes, but is now only implemented by ChargeEntity and moved
the method to get the net amount into ChargeEntity.

Modified the method for calculating the net amount to return a negative
value when the payment has a failed status to account for fees now being
charged to Stripe failed payments.

Co-authored-by: Nathaniel Steers <nathaniel.steers@digital.cabinet-office.gov.uk>


kbottla (approved):
LGTM!

PR #3391, opened 2022-01-10
bau: Use 'test-fargate' when running provider contract test

oswaldquek:
The purpose of running the provider contract test is to ensure the app can be
deployed to the test environment without breaking its contractual obligations
with other apps. Pacts are tagged with 'test-fargate' in the concourse deploy
pipeline for app deployments so this tag needs to be updated.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/pact/QueueMessageContractTest.java
+++ b/src/test/java/uk/gov/pay/connector/pact/QueueMessageContractTest.java
@@ -65,7 +65,7 @@ import static uk.gov.service.payments.commons.model.Source.CARD_EXTERNAL_TELEPHO
 
 @RunWith(PactRunner.class)
 @Provider("connector")
-@PactBroker(scheme = "https", host = "${PACT_BROKER_HOST:pact-broker-test.cloudapps.digital}", tags = {"${PACT_CONSUMER_TAG}", "test"},
+@PactBroker(scheme = "https", host = "${PACT_BROKER_HOST:pact-broker-test.cloudapps.digital}", tags = {"${PACT_CONSUMER_TAG}", "test-fargate"},
         authentication = @PactBrokerAuth(username = "${PACT_BROKER_USERNAME}", password = "${PACT_BROKER_PASSWORD}"),
         consumers = {"ledger"})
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/pact/QueueMessageContractTest.java
```diff
@@ -65,7 +65,7 @@
 
 @RunWith(PactRunner.class)
 @Provider("connector")
-@PactBroker(scheme = "https", host = "${PACT_BROKER_HOST:pact-broker-test.cloudapps.digital}", tags = {"${PACT_CONSUMER_TAG}", "test"},
+@PactBroker(scheme = "https", host = "${PACT_BROKER_HOST:pact-broker-test.cloudapps.digital}", tags = {"${PACT_CONSUMER_TAG}", "test-fargate"},
```
katstevens, line 68: Whereabouts is the tag being set in Concourse (pay-ci)?

oswaldquek: For example, at https://pay-cd.deploy.payments.service.gov.uk/teams/pay-dev/pipelines/deploy-to-test/jobs/adminusers-pact-tag/builds/42. See the "tag-pact" task.

katstevens: Gotcha, thanks - so [this is the config for connector](https://github.com/alphagov/pay-ci/blob/5e005f42911d74305e377c47efd0f5cefe065905/ci/pipelines/deploy-to-test.yml#L1548) 👍

katstevens (dismissed):
Happy with this from the Concourse point of view - recommend getting a review from someone more familiar with Connector/pact...

stephencdaly (dismissed):
This change makes sense to me. If I've understood correctly, we were previously tagging with "test" as part of the master build on Jenkins and then with "test-fargate" as part of the "deploy-to-test" pipeline on Concourse. However we have now removed the step that tags with "test" on Jenkins master builds, so we need to check against tag "test-fargate".

stephencdaly (approved):

PR #3406, opened 2022-01-13
pp-8945 provide payment intent information on stripe authorisation fa…

nlsteers:
…ilures

## WHAT YOU DID
Previously, payments that failed authorisation were storing the stripe charge id for the gateway_transaction_id. This was left over from when our stripe payment integration was not using payment intent.

To be consistent with successful payments, this has been updated to return the payment intent id.

Co-authored-by: @stephencdaly


kbottla (dismissed):
LGTM!

kbottla (approved):

PR #3407, opened 2022-01-13
PP-8951 Add reason to metadata for Stripe transfers

stephencdaly:
Add a "reason" field to the metadata we attach to a Stripe transfer when
it is created. This will allow us to later determine the purpose of the
transfer, as we need to be able to differentiate between transfers made
for refunds and transfers made to collect fees for failed payments when
reconciling payouts.

We also attach a reason to transfers made for payments when they are
captured, but we will not use this for anything other than information
purposes when viewing in the Stripe dashboard.

Co-authored-by: Nathaniel Steers <nathaniel.steers@digital.cabinet-office.gov.uk>

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadata.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadata.java
@@ -1,7 +1,6 @@
 package uk.gov.pay.connector.gateway.stripe.request;
 
 import java.util.Map;
-import java.util.Optional;
 
 public class StripeTransferMetadata {
     public static final String STRIPE_CHARGE_ID_KEY = "stripe_charge_id";
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadata.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadata.java
@@ -1,30 +1,38 @@
 package uk.gov.pay.connector.gateway.stripe.request;
 
 import java.util.Map;
+import java.util.Optional;
 
 public class StripeTransferMetadata {
     public static final String STRIPE_CHARGE_ID_KEY = "stripe_charge_id";
     public static final String GOVUK_PAY_TRANSACTION_EXTERNAL_ID = "govuk_pay_transaction_external_id";
+    public static final String REASON_KEY = "reason";
 
     private final String stripeChargeId;
     private final String govukPayTransactionExternalId;
+    private final StripeTransferMetadataReason reason;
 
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadataReason.java
```diff
@@ -0,0 +1,23 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Arrays;
+import java.util.Locale;
+
+public enum StripeTransferMetadataReason {
+    NOT_DEFINED,
+    TRANSFER_FEE_AMOUNT_FOR_FAILED_PAYMENT,
+    TRANSFER_REFUND_AMOUNT,
+    TRANSFER_PAYMENT_AMOUNT;
```
kbottla, line 12: Any reason we are sending values in full upper case ? Don't see any issues though as `StripeTransferMetadataReason.fromString` is not case sensitive. May look odd

image

vs
image

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadata.java
```diff
@@ -1,30 +1,38 @@
 package uk.gov.pay.connector.gateway.stripe.request;
 
 import java.util.Map;
+import java.util.Optional;
```
kbottla: unused import

kbottla (dismissed):

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadataReason.java
stephencdaly: We decided to send upper case as we thought it wouldn't make a difference and is marginally less code. But if there is a mild preference to use lowercase, happy to do that.

stephencdaly:

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadataReason.java
nlsteers: Overrode the toString to give us a lowercase value of the enum name()

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeTransferMetadata.java
nlsteers: Removed

nlsteers (approved):
I’m _also_ confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #3412, opened 2022-01-14
PP-8951 amend payout processing for fee collection balance tx

nlsteers:
## WHAT YOU DID
Updated the payouts process to emit `PAYMENT_INCLUDED_IN_PAYOUT` events when transfers are tagged with a `TRANSFER_FEE_AMOUNT_FOR_FAILED_PAYMENT` reason.

Previously, the processing loop would incorrectly assume that all transfers were refunds. Now, we check the `reason` to determine if the transfer is a payout for fee collection on failed payments.


stephencdaly (approved):

PR #3415, opened 2022-01-17
PP-8951 only send `total_amount` to ledger when a surcharge is present

nlsteers:
## WHAT YOU DID
Only send `total_amount` to ledger when a surcharge is present.


SandorArpa (approved):
LGTM

PR #3422, opened 2022-01-20
PP-8951 Parity check total_amount only if corporate surcharge exists

kbottla:
## WHAT YOU DID
- Connector now sends total_amount to Ledger only when corporate surcharge exists on a charge. So parity check total_amount only when surcharge is available.

with @nlsteers


nlsteers (approved):
LGTM

PR #3425, opened 2022-01-21
PP-8950 add FeeIncurredEvent to historical emitter

SandorArpa:
## WHAT YOU DID
- Emit FeeIncurredEvent if needed when emitting historical events. We only want to emit the FEE_INCURRED event for charges using the new Stripe pricing. The original Stripe pricing has no FeeType.

with @stephencdaly


stephencdaly (approved):

PR #3427, opened 2022-01-24
PP-8950 Do not expunge Stripe charges missing fees

stephencdaly:
We collect fees asynchronously for failed Stripe payments that we have attempted to authorise with Stripe - and so have a gateway transaction id. We want to prevent these payments from being expunged if we have not yet collected the fee.

As we do for payments not in a terminal state, set the parity check status for these payments to "SKIPPED" so they will be retried later.

With @SandorArpa


kbottla (approved):
LGTM!

PR #3430, opened 2022-01-25
PP-8950 Add pact provider for fee incurred event

stephencdaly:
Add a PactVerifyProvider to generate a FeeIncurred event to test against
the pact produced by ledger.

with @SandorArpa


SandorArpa (approved):

PR #3431, opened 2022-01-25
PP-8950 Skip expunging test Stripe payments without fees

stephencdaly:
Stripe V2 pricing for test accounts, including fees for failed payments, has been enabled on all environments.

If a payment is for a test account, skip expunging it if there are no fees collected when we expect fees to have been collected.


SandorArpa (approved):
LOKTM

PR #3441, opened 2022-02-01
PP-9115 update credential state on notification

SandorArpa:
## WHAT YOU DID
- if account doesn’t have any requirements due, then update gateway_account_credential state to `ACTIVE` (if not already active) when receiving an `account.updated` notification.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
@@ -7,6 +7,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.gateway.stripe.response.StripeNotification;
+import uk.gov.pay.connector.gatewayaccountcredentials.service.GatewayAccountCredentialsService;
 
 import javax.inject.Inject;
 
@@ -18,15 +19,20 @@ public class StripeAccountUpdatedHandler {
     private static final Logger LOGGER = LoggerFactory.getLogger(StripeAccountUpdatedHandler.class);
 
     private final ObjectMapper objectMapper;
+    private final GatewayAccountCredentialsService credentialsService;
 
     @Inject
-    public StripeAccountUpdatedHandler(ObjectMapper objectMapper) {
+    public StripeAccountUpdatedHandler(GatewayAccountCredentialsService credentialsService, ObjectMapper objectMapper) {
+        this.credentialsService = credentialsService;
         this.objectMapper = objectMapper;
     }
 
     public void process(StripeNotification notification) {
         try {
             var dataObject = objectMapper.readValue(notification.getObject(), DataObject.class);
+            if (canBeActivated(dataObject)) {
+                credentialsService.activateCredentialIfNotYetActive(dataObject.getId());
+            }
             LOGGER.info(String.format("Received an account.updated event for stripe account %s", dataObject.getId()),
                     kv("stripe_account_id", dataObject.getId()),
                     kv("payouts_enabled", dataObject.isPayoutsEnabled()),
@@ -37,17 +43,27 @@ public class StripeAccountUpdatedHandler {
         }
     }
 
+    private boolean canBeActivated(DataObject dataObject) {
+        return (!dataObject.getRequirements().hasCurrentlyDue() &&
+                !dataObject.getRequirements().hasPastDue() &&
+                dataObject.isChargesEnabled() &&
+                dataObject.isPayoutsEnabled());
+    }
+
     @JsonIgnoreProperties(ignoreUnknown = true)
     private static class DataObject {
 
         @JsonProperty("id")
         private String id;
 
+        @JsonProperty("charges_enabled")
+        private boolean chargesEnabled;
+
         @JsonProperty("payouts_enabled")
         private boolean payoutsEnabled;
 
         @JsonProperty("requirements")
-        private JsonNode requirements;
+        private Requirements requirements;
 
         private String getId() {
             return id;
@@ -57,8 +73,31 @@ public class StripeAccountUpdatedHandler {
             return payoutsEnabled;
         }
 
-        private String getRequirements() {
-            return requirements.toString();
+        private Requirements getRequirements() {
+            return requirements;
+        }
+
+        public boolean isChargesEnabled() {
+            return chargesEnabled;
+        }
+
+
+    }
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    private static class Requirements {
+        @JsonProperty("currently_due")
+        private JsonNode currentlyDue;
+
+        @JsonProperty("past_due")
+        private JsonNode pastDue;
+
+        public boolean hasCurrentlyDue() {
+            return !currentlyDue.isEmpty();
+        }
+
+        public boolean hasPastDue() {
+            return !pastDue.isEmpty();
         }
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -9,6 +9,7 @@ import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountCredentialsNo
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountCredentials;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.model.StripeCredentials;
 import uk.gov.pay.connector.gatewayaccountcredentials.dao.GatewayAccountCredentialsDao;
 import uk.gov.pay.connector.gatewayaccountcredentials.exception.NoCredentialsExistForProviderException;
 import uk.gov.pay.connector.gatewayaccountcredentials.exception.NoCredentialsInUsableStateException;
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
```diff
@@ -4,29 +4,50 @@
...
+                        .findByCredentialsKeyValue(StripeCredentials.STRIPE_ACCOUNT_ID_KEY, dataObject.getId());
+                credentials.ifPresent(credentialEntity -> {
+                    if (credentialEntity.getState() != ACTIVE) {
+                        credentialEntity.setState(ACTIVE);
+                        credentialsDao.merge(credentialEntity);
+                        LOGGER.info(String.format("Updated stripe account %s to ACTIVE", dataObject.getId()),
+                                kv("stripe_account_id", dataObject.getId()),
+                                kv("requirements", dataObject.getRequirements()));
+                    }
+                });
```
stephencdaly: Reckon this should be moved into GatewayAccountCredentialsService so that everything to do with updating credentials is in one place?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
```diff
@@ -57,8 +87,31 @@ private boolean isPayoutsEnabled() {
...
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    private static class Requirements {
+        @JsonProperty("currently_due")
+        private JsonNode currentlyDue;
+
+        @JsonProperty("past_due")
+        private JsonNode pastDue;
+
+        public boolean isCurrentlyDue() {
```
stephencdaly: The naming of this method isn't very clear. It's kind of the opposite of what it's returning?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAccountUpdatedHandler.java
```diff
@@ -18,15 +19,27 @@
...
         this.objectMapper = objectMapper;
     }
 
     public void process(StripeNotification notification) {
         try {
             var dataObject = objectMapper.readValue(notification.getObject(), DataObject.class);
+            if (canBeActivated(dataObject)) {
+                boolean credentialsUpdated = credentialsService.activateCredentialIfNotYetActive(dataObject.getId());
+                if (credentialsUpdated) {
+                    LOGGER.info(String.format("Received an account.updated event for stripe account %s", dataObject.getId()),
```
stephencdaly: This log line looks the same as the log line below. Do we need to log anything here, and if so could it be done inside the service to simplify things?

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -232,4 +234,19 @@ public boolean hasActiveCredentials(Long gatewayAccountId) {
...
+                .findByCredentialsKeyValue(StripeCredentials.STRIPE_ACCOUNT_ID_KEY, credentialId);
+        AtomicBoolean updated = new AtomicBoolean(false);
+        credentials.ifPresent(credentialEntity -> {
+            if (credentialEntity.getState() != ACTIVE) {
+                credentialEntity.setState(ACTIVE);
+                gatewayAccountCredentialsDao.merge(credentialEntity);
+                updated.set(true);
+            }
+        });
+        return updated.get();
```
stephencdaly: This is possibly a bit better?
```suggestion
        return credentials.map(credentialEntity -> {
            if (credentialEntity.getState() != ACTIVE) {
                credentialEntity.setState(ACTIVE);
                gatewayAccountCredentialsDao.merge(credentialEntity);
                return true;
            }
            return false;
        }).orElse(false);
```

Although we're only returning a boolean to log something. If we need to log, it probably makes more sense to log something here that the credential has been updated to active with the credential external id and the gateway account id?

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -91,14 +99,17 @@
     private PayoutEmitterService mockPayoutEmitterService;
     @Mock
     private Appender<ILoggingEvent> mockAppender;
+    @Mock
+    private GatewayAccountCredentialsDao mockGatewayAccountCredentialsDao;
```
stephencdaly: The class under test doesn't use this mock anymore

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -410,6 +423,47 @@ void shouldReturnFalseWhenForwardedIpAddressIsNotInAllowedIpAddresses() {
...
+    void shouldNotUpdateNonExistentCredentials() {
+        String payload = TestTemplateResourceLoader.load(STRIPE_NOTIFICATION_ACCOUNT_UPDATED);
+        var credentialsEntity = aGatewayAccountCredentialsEntity()
+                .withState(ENTERED)
+                .withPaymentProvider("stripe").build();
+        when(mockGatewayAccountCredentialsDao.findByCredentialsKeyValue(anyString(), anyString()))
+                .thenReturn(Optional.empty());
+        notificationService.handleNotificationFor(payload, signPayload(payload), FORWARDED_IP_ADDRESSES);
+        verify(mockGatewayAccountCredentialsDao, never()).merge(any(GatewayAccountCredentialsEntity.class));
+    }
```
stephencdaly: These tests should be covered in `GatewayAccountCredentialsServiceTest` so should probably be removed from here. It would be good to add some unit tests to make sure we don't try to update if the notification has requirements, or has payouts or charges enabled though.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -108,6 +119,8 @@
 
     @BeforeEach
     void setup() {
+        mockGatewayAccountCredentialsService = new GatewayAccountCredentialsService(mockGatewayAccountCredentialsDao);
```
stephencdaly: This should use the mock rather than creating an instance.

stephencdaly (commented):
A few more comments

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
SandorArpa: Yes, it does, to prove that it was not called.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
SandorArpa: 👍

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
SandorArpa: Removed mock.

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
SandorArpa: 👍

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
stephencdaly: We should just verify that the `activateCredentialIfNotYetActive` method on the mocked GatewayAccountCredentialsService isn't called right?

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -232,4 +234,41 @@ public boolean hasActiveCredentials(Long gatewayAccountId) {
...
         return gatewayAccountCredentials.or(() -> gatewayAccountCredentialsEntityByPaymentProvider);
     }
+
+    @Transactional
+    public void activateCredentialIfNotYetActive(String stripeAccountId) {
+        var credentials = gatewayAccountCredentialsDao
+                .findByCredentialsKeyValue(StripeCredentials.STRIPE_ACCOUNT_ID_KEY, stripeAccountId);
+
+        credentials.ifPresent(credentialEntity -> {
+            if (credentialEntity.getState() == ACTIVE) {
```
stephencdaly: This would be very much an edge case, but should we check that the credential isn't `RETIRED` too? This would prevent against a case where a service switched PSP from Stripe -> Worldpay (or another future provider), but then received an account updated event from Stripe for whatever reason.

stephencdaly: Ah I guess the check that there are no active credentials would do that anyway

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationServiceTest.java
```diff
@@ -410,7 +414,43 @@ void shouldReturnFalseWhenForwardedIpAddressIsNotInAllowedIpAddresses() {
...
+
+    @Test
+    void shouldNotUpdateCredentialsWhenPastDueIsPresent() {
+        String payload = TestTemplateResourceLoader.load(STRIPE_NOTIFICATION_ACCOUNT_UPDATED);
+        payload = payload.replace("\"past_due\": [", "\"past_due\": [\"individual.dob.day\"");
+
+        notificationService.handleNotificationFor(payload, signPayload(payload), FORWARDED_IP_ADDRESSES);
+        verify(mockGatewayAccountCredentialsService, never()).activateCredentialIfNotYetActive(anyString());
+    }
+
```
stephencdaly, line 440: We're just missing the test to check the credentials update is called when there are no requirements.

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
```diff
@@ -514,4 +520,148 @@ void shouldReturnEmptyOptionalIfNoGatewayAccountCredentialsFoundWhenSearchingByP
...
+                .withGatewayAccountEntity(gatewayAccountEntity)
+                .build();
+
+        gatewayAccountEntity.setGatewayAccountCredentials(List.of(gatewayAccountCredentialsEntity));
+        when(mockGatewayAccountCredentialsDao.findByCredentialsKeyValue(StripeCredentials.STRIPE_ACCOUNT_ID_KEY, stripeCredentialId))
+                .thenReturn(Optional.of(gatewayAccountCredentialsEntity));
+        gatewayAccountCredentialsService.activateCredentialIfNotYetActive(stripeCredentialId);
+
+        verify(mockGatewayAccountCredentialsDao, never()).merge(any(GatewayAccountCredentialsEntity.class));
+    }
```
stephencdaly, line 590: Could we use some parameterised tests for the last few, as they look the same apart from a status, just to reduce some repetition? Just a thought though and shouldn't stop this PR from going in.

stephencdaly (commented):
Looks good, I think a test that existed at one point got lost along the way

stephencdaly (approved):
LGTM!

PR #3445, opened 2022-02-03
PP-9155 change default status for created credentials

SandorArpa:
## WHAT YOU DID
- When creating stripe live accounts, set gateway_account_credential state to CREATED.
- When creating stripe test accounts, set gateway_account_credential state to ACTIVE

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -9,6 +9,7 @@ import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountCredentialsNo
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountCredentials;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType;
 import uk.gov.pay.connector.gatewayaccount.model.StripeCredentials;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -84,11 +87,21 @@ public GatewayAccountCredentials createGatewayAccountCredentials(GatewayAccountE
...
+        if (paymentGatewayName == SANDBOX ||
+                (paymentGatewayName == STRIPE && gatewayAccountType == TEST)) {
+            return ACTIVE;
+        }
+
+        if (paymentGatewayName == STRIPE) {
+            return CREATED;
+        }
+
+        if ((isFirstCredentials && credentialsPrePopulated)) {
```
stephencdaly, line 104: Any reason why this isn't part of the first `if`?

stephencdaly (approved):
LGTM

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
SandorArpa: Purely readability, can move it if you think it'd be better. Also, this would be Worldpay, so it kind of separates Stripe.

stephencdaly: Ah I see why it wouldn't be in the flow of logic, but it could instead be

```
if (paymentGatewayName == STRIPE && gatewayAccountType == LIVE) {
    return CREATED;
} 

if (...) {
    return ACTIVE:
}
```

stephencdaly: Agreed that separating out the logic for what we do for various types of test accounts might make it a bit clearer so leaving it as is.

PR #3455, opened 2022-02-10
PP-9265 Increase timeout for queue message receiver shutdown

stephencdaly:
Increase the timeout for waiting for scheduled tasks to finish for the
SQS queue message receivers for the capture queue and the tasks queue.

This is required because the tasks run by these message receivers poll
SQS using long polling with a maximum wait time of 20s before processing
up to a batch of 10 messages.

Increase the timeout to 40s for both queue receivers to decrease the
chance that we have to force-stop the schedulers.

Make the timeout more easily configurable by reading from an
environment variable for each queue receiver.


SandorArpa (approved):
LOKTM

PR #3459, opened 2022-02-15
PP-9257 place Stripe charge disputed webhooks onto connectors task queue

nlsteers:
## WHAT YOU DID
Updated the Stripe Notification Service to handle dispute.created.event webhooks. This change introduces a new queue task handler for webhooks and generifies the Task model for use with multiple event types.

To maintain compatability with existing tasks some handling code and tests remain that backport the previous message format. This can be removed at a later date.

Co-authored-by: Sandor Arpa <sandor.arpa@digital.cabinet-office.gov.uk>


src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueue.java
```diff
@@ -29,14 +30,14 @@ public TaskQueue(SqsQueueService sqsQueueService,
...
     }
 
-    public void addTaskToQueue(PaymentTask paymentTask) throws QueueException, JsonProcessingException {
-        String message = objectMapper.writeValueAsString(paymentTask);
+    public void addTaskToQueue(Task task) throws QueueException, JsonProcessingException {
+        String message = objectMapper.writeValueAsString(task);
         QueueMessage queueMessage = sendMessageToQueue(message);
         LOGGER.info("Task [{}] added to queue. Message ID [{}]",
-                paymentTask.getPaymentExternalId(), queueMessage.getMessageId());
+                task.getData(), queueMessage.getMessageId());
```
stephencdaly: We probably don't want to log out the entire webhook for sending a webhook task. Maybe we could move where we do the logging to be a bit more particular about what useful information we log?

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueService.java
```diff
@@ -55,18 +59,32 @@ public void offerTasksOnStateTransition(ChargeEntity chargeEntity) {
         }
     }
 
+    public void add(Task task) {
+        try {
+            taskQueue.addTaskToQueue(task);
+            logger.info("Added [{}] task to queue", task.getTaskType().getName());
```
stephencdaly: Doesn't look like there'll be any identifying information attached to this log line in many cases, so possibly no point in logging here.

stephencdaly (commented):
Looks good to me, just some minor comments about logging. Maybe good to get another pair of eyes over it to check the approach.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueService.java
stephencdaly: To clarify what I mean by "in many cases". If the thread is processing a HTTP request from another of our apps, e.g. to authorise a payment, there will usually be some data on the MDC - such as the payment_external_id and gateway_account_id, which are automatically included as fields on the log line. 

However if we are processing on a different thread such as a thread processing a webhook notification, or the charge expiry sweep task, there won't be anything on the MDC to be pulled into the log line, so this log line won't be very informative as it doesn't tell us anything about the individual charge or webhook we've added a task for.

One improvement that could be made is to put more things on the MDC for the webhook thread, or the charge expiry sweep thread so useful information is included on all log lines by default. Otherwise, we have to manually include key value pairs when logging to log the identifying information.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueue.java
nlsteers: good 👀, this was supposed to be the task type. Updated.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueService.java
```diff
@@ -55,18 +59,32 @@ public void offerTasksOnStateTransition(ChargeEntity chargeEntity) {
         }
     }
 
+    public void add(Task task) {
+        try {
+            taskQueue.addTaskToQueue(task);
+        } catch (QueueException | JsonProcessingException e) {
+            logger.error("Error adding task to queue",
+                    kv("task_id", task.getTaskId()),
```
stephencdaly: It doesn't look like this `task_id` will tell us anything as looks like it's only used for logging, but this is the only time we log it if we fail to add to the queue.

It might be a good idea to look into adding more data into the MDC so this logging will include useful info - `payment_external_id` or `stripe_webhook_id` etc.. But this can be done in a separate PR.

src/main/java/uk/gov/pay/connector/queue/tasks/model/Task.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Objects;
+import java.util.UUID;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({ "task_id", "data", "task", "payment_external_id"})
+public class Task {
+    
+    @JsonProperty("task_id")
+    private String taskId;
```
stephencdaly: Does having this `task_id` give us anything extra that the queue message id doesn't give us? Looks like it's just used for logging and all the useful places we log it we also have the queue message id.

nlsteers: I think you are right in that this doesn't provide any more detail on the task lifecycle, my thought was it would be easier to associate the de/serialisation points... 

I'll look into adding more detail to the MDC.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueue.java
```diff
@@ -29,14 +32,16 @@ public TaskQueue(SqsQueueService sqsQueueService,
...
     }
 
-    public void addTaskToQueue(PaymentTask paymentTask) throws QueueException, JsonProcessingException {
-        String message = objectMapper.writeValueAsString(paymentTask);
+    public void addTaskToQueue(Task task) throws QueueException, JsonProcessingException {
+        String message = objectMapper.writeValueAsString(task);
         QueueMessage queueMessage = sendMessageToQueue(message);
-        LOGGER.info("Task [{}] added to queue. Message ID [{}]",
-                paymentTask.getPaymentExternalId(), queueMessage.getMessageId());
+        LOGGER.info("Task added to queue. [{}]. [{}]. [{}].",
```
kbottla: We should log payment external ID where ever possible.

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/WebhookTaskHandler.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.pay.connector.queue.tasks.handlers;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WebhookTaskHandler {
```
kbottla: We could name it specific to Stripe. Fine if you prefer to leave it as is.

```suggestion
public class StripeWebhookTaskHandler {
```

src/main/java/uk/gov/pay/connector/queue/tasks/model/Task.java
kbottla: While taskId with randomUUID may not provide more detail, would it still be useful to have payment_external_id / notification_id as top level fields. Placing everything within `data` means we can't log ID until it is deserialised from the queue. 
Might be useful to log notification ID (https://github.com/alphagov/pay-connector/blob/fe77096fdffc216cef329d845dd2f1198dd1e2f8/src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java#L61-L62) similar to payment_external_id here https://github.com/alphagov/pay-connector/blob/fe77096fdffc216cef329d845dd2f1198dd1e2f8/src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java#L55-L57

Also fine if the plan is to log all in WebhookTaskHandler

src/test/java/uk/gov/pay/connector/queue/tasks/TaskQueueServiceTest.java
```diff
@@ -98,9 +95,7 @@ void shouldOfferFeeTask_whenChargeIsTerminallyFailed_andStripe_andGatewayIdPrese
                 .thenReturn(List.of(String.valueOf(chargeEntity.getGatewayAccount().getId())));
 
         taskQueueService.offerTasksOnStateTransition(chargeEntity);
-
-        var expectedPaymentTask = new PaymentTask(chargeEntity.getExternalId(), COLLECT_FEE_FOR_STRIPE_FAILED_PAYMENT_TASK_NAME);
-        verify(mockTaskQueue).addTaskToQueue(eq(expectedPaymentTask));
+        verify(mockTaskQueue).addTaskToQueue(any(Task.class));
```
kbottla: Can we capture arguments and check task data is correctly set. Not sure if another test covers this already but with `any` we won't be able to detect changes to the task name set or data.

src/test/java/uk/gov/pay/connector/queue/tasks/TaskQueueServiceTest.java
```diff
@@ -245,9 +253,7 @@ void shouldOfferFeeTask_whenTestGatewayAccountIsEnabled() throws Exception {
         when(mockStripeGatewayConfig.isEnableTransactionFeeV2ForTestAccounts()).thenReturn(true);
 
         taskQueueService.offerTasksOnStateTransition(chargeEntity);
-
-        var expectedPaymentTask = new PaymentTask(chargeEntity.getExternalId(), COLLECT_FEE_FOR_STRIPE_FAILED_PAYMENT_TASK_NAME);
-        verify(mockTaskQueue).addTaskToQueue(eq(expectedPaymentTask));
+        verify(mockTaskQueue).addTaskToQueue(any(Task.class));
```
kbottla: Same as above. Would be good to check specific values if possible or an integration test which could cover this.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueue.java
nlsteers: This is now in the mapped diagnostic context.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueService.java
nlsteers: I've added the MDC changes 😄

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/WebhookTaskHandler.java
nlsteers: Agreed, have renamed.

src/main/java/uk/gov/pay/connector/queue/tasks/model/Task.java
nlsteers: We've added the `payment_external_id` and the `notification_id` to the MDC so we can log these more effectively without the need to include them as top level fields on the task object.

For webhooks, the plan is to log all in the handler.

src/test/java/uk/gov/pay/connector/queue/tasks/TaskQueueServiceTest.java
nlsteers: Reverted.

src/test/java/uk/gov/pay/connector/queue/tasks/TaskQueueServiceTest.java
nlsteers: Reverted.

stephencdaly (dismissed):
Comments I made have been addressed. Looks good to me!

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
```diff
@@ -1,54 +1,77 @@
...
+                        PaymentTaskData paymentTaskData;
+                        // backport existing task message format - will be removed
+                        if (taskMessage.getTask().getPaymentExternalId() != null) {
+                            paymentTaskData = new PaymentTaskData(taskMessage.getTask().getPaymentExternalId());
+                        } else {
+                            paymentTaskData = objectMapper.readValue(taskMessage.getTask().getData(), PaymentTaskData.class);
+                        }
+                        MDC.put(PAYMENT_EXTERNAL_ID, paymentTaskData.getPaymentExternalId());
+                        LOGGER.info("Processing [{}] task.", taskType.getName());
+                        collectFeesForFailedPaymentsTaskHandler.collectAndPersistFees(paymentTaskData);
```
kbottla, line 58: We also log PAYMENT_EXTERNAL_ID when emitting event in collectAndPersistFees. https://github.com/alphagov/pay-connector/blob/fa986b4f58d0748eb6c22e6bcbc068b418ec6a5c/src/main/java/uk/gov/pay/connector/queue/tasks/CollectFeesForFailedPaymentsTaskHandler.java#L63
This can potentially add PAYMENT_EXTERNAL_ID twice when emitting event or any other logs. Worth checking if this the case and avoid duplicate IDs if possible.

nlsteers: We've removed the additional log statement in the failed payments task handler in favour of adding the `PAYMENT_EXTERNAL_ID` to the MDC. 

FWIW we log information that is already in the MDC elsewhere in connector and it doesn't appear to duplicate the values in the logging output.

kbottla: If it doesn't duplicate should be fine to log PAYMENT_EXTERNAL_ID in both places.

kbottla (approved):

PR #3464, opened 2022-02-17
PP-9260 Use SQS classes from commons

stephencdaly:
Classes for polling SQS queues were moved to commons to share between Java apps. Import them from here.

Also bump AWS SDK version. Not sure why dependabot was not bumping this, as there were versions > 1.12. Bump and remove that ignore from the dependabot file.


nlsteers (approved):
👍

PR #3465, opened 2022-02-17
PP-9005 Log when new stripe pricing is applied

kbottla:
## WHAT YOU DID
- Added a log line when new stripe pricing is applied.


SandorArpa (approved):
LOKTM

PR #3475, opened 2022-02-24
PP-9258 add new event - dispute created

SandorArpa:
## WHAT YOU DID
- Process the webhook data into a DisputeCreated event
- update / add tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/dispute/DisputeCreated.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/dispute/DisputeCreated.java
@@ -1,0 +1,24 @@
+package uk.gov.pay.connector.events.model.dispute;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+import java.time.ZonedDateTime;
+
+public class DisputeCreated extends DisputeEvent{
+    public DisputeCreated(String resourceExternalId, String parentResourceExternalId, String serviceId,
+                          Boolean live, EventDetails eventDetails, ZonedDateTime disputeCreated) {
+        super(resourceExternalId, parentResourceExternalId, serviceId, live, eventDetails, disputeCreated);
+    }
+
+    @Override
+    public String toString() {
+        return "DisputeCreated{" +
+                "resourceExternalId=" + getResourceExternalId() +
+                ", eventDetails=" + getEventDetails() +
+                ", timestamp=" + getTimestamp() +
+                ", parentResourceExternalId= " + getParentResourceExternalId() +
+                ", serviceId=" + getServiceId() +
+                ", live=" + getLive() +
+                '}';
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
+++ b/src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
+import uk.gov.pay.connector.gateway.stripe.response.StripeNotification;
 import uk.gov.service.payments.commons.queue.exception.QueueException;
 import uk.gov.pay.connector.queue.tasks.handlers.CollectFeesForFailedPaymentsTaskHandler;
 import uk.gov.pay.connector.queue.tasks.handlers.StripeWebhookTaskHandler;
@@ -14,6 +15,7 @@ import java.util.List;
 
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static uk.gov.service.payments.logging.LoggingKeys.PAYMENT_EXTERNAL_ID;
+import static uk.gov.service.payments.logging.LoggingKeys.STRIPE_EVENT_ID;
 
 public class TaskQueueMessageHandler {
 
@@ -58,7 +60,10 @@ public class TaskQueueMessageHandler {
                         collectFeesForFailedPaymentsTaskHandler.collectAndPersistFees(paymentTaskData);
                         break;
                     case HANDLE_STRIPE_WEBHOOK_NOTIFICATION:
-                        stripeWebhookTaskHandler.process(taskMessage.getTask().getData());
+                        var stripeNotification = objectMapper.readValue(taskMessage.getTask().getData(), StripeNotification.class);
+                        MDC.put(STRIPE_EVENT_ID, stripeNotification.getId());
+                        LOGGER.info("Processing [{}] task.", taskType.getName());
+                        stripeWebhookTaskHandler.process(stripeNotification);
                         break;
                     default:
                         LOGGER.error("Task [{}] is not supported.", taskType.getName());
@@ -71,6 +76,7 @@ public class TaskQueueMessageHandler {
                 );
             } finally {
                 MDC.remove(PAYMENT_EXTERNAL_ID);
+                MDC.remove(STRIPE_EVENT_ID);
             }
         });
     }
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/connector/events/model/dispute/DisputeCreated.java
```diff
@@ -0,0 +1,27 @@
+package uk.gov.pay.connector.events.model.dispute;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+import java.time.ZonedDateTime;
+
+public class DisputeCreated extends DisputeEvent{
+    private Long eventCreated;
```
stephencdaly: Is this something we do for other events? What does eventCreated mean in this context?

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
```diff
@@ -71,6 +71,7 @@ public void processMessages() throws QueueException {
                 );
             } finally {
                 MDC.remove(PAYMENT_EXTERNAL_ID);
+                MDC.remove("stripe_webhook_id");
```
stephencdaly: Would it be better to move this into [StripeWebhookTaskHandler.java](https://github.com/alphagov/pay-connector/pull/3475/files#diff-4686e0d1bd49bfb944083b37fd579af927c101134758aab612d4b7f59ed7c263) so that we're adding and removing in the same place? Otherwise it's a bit hard to see where this is coming from. We could also use a constant for this string.

I guess the reason why you're doing this is because we don't want to to handle exceptions in StripeWebhookTaskHandler. Would a solution be to catch and rethrow the exception there?

The other option might be to deserialise the StripeNotification here and add the webhook id into the MDC in this class. I think that might actually be the nicest option..

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/StripeWebhookTaskHandler.java
```diff
@@ -1,20 +1,76 @@
...
+        MDC.put("stripe_webhook_id", stripeNotification.getId());
+        if (stripeNotification.getType().equals(DISPUTE_CREATED.getType())) {
+            var stripeDisputeData = mapper.readValue(stripeNotification.getObject(), StripeDisputeData.class);
+            if (stripeDisputeData.getBalanceTransactionList().size() > 1) {
+                throw new RuntimeException("Dispute data has too many balance_transactions");
+            }
+            Optional<LedgerTransaction> ledgerTransaction = ledgerService
+                    .getTransactionForProviderAndGatewayTransactionId(STRIPE.getName(), stripeDisputeData.getPaymentIntentId());
+            if (ledgerTransaction.isPresent()) {
+                var transaction = ledgerTransaction.get();
```
stephencdaly: would it be better to use `.map(..).elseThrow`?

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/StripeWebhookTaskHandler.java
```diff
@@ -1,20 +1,76 @@
...
+                        stripeDisputeData.getPaymentIntentId()));
+            }
+        } else {
+            throw new RuntimeException("Unknown webhook task: " + stripeNotification.getType());
+        }
     }
 
-    public void emitEvent() {
+    public void emitEvent(Event event) {
+        logger.info("Received event for Stripe Webhook Task: " + event.getResourceExternalId());
```
stephencdaly, line 55: would it be good to log what type of webhook we're processing somewhere too?

stephencdaly (commented):
Looks good! Some minor comments, mainly about logging

src/main/java/uk/gov/pay/connector/events/model/dispute/DisputeCreated.java
SandorArpa: Discussed it, we don't need it.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
SandorArpa: 👍

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/StripeWebhookTaskHandler.java
SandorArpa: 👍

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/StripeWebhookTaskHandler.java
SandorArpa: 👍

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
```diff
@@ -58,7 +59,10 @@ public void processMessages() throws QueueException {
                         collectFeesForFailedPaymentsTaskHandler.collectAndPersistFees(paymentTaskData);
                         break;
                     case HANDLE_STRIPE_WEBHOOK_NOTIFICATION:
-                        stripeWebhookTaskHandler.process(taskMessage.getTask().getData());
+                        var stripeNotification = objectMapper.readValue(taskMessage.getTask().getData(), StripeNotification.class);
+                        MDC.put("stripe_webhook_id", stripeNotification.getId());
```
stephencdaly: I noticed that we're adding this to the MDC as `stripe_event_id` in StripeNotificationService. Could we make the 2 the same? It might be good to use a constant for this key that both places use so these don't get out of sync.

stephencdaly (commented):
Looks good, just one last thing

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueMessageHandler.java
nlsteers: Probably should be added to `pay-java-commons` logging keys.

SandorArpa: 👍

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/StripeWebhookTaskHandler.java
```diff
@@ -1,20 +1,71 @@
...
+                    .getTransactionForProviderAndGatewayTransactionId(STRIPE.getName(), stripeDisputeData.getPaymentIntentId());
+            ledgerTransaction.map(transaction -> {
+                var disputeCreatedEvent = createDisputeCreatedEvent(stripeDisputeData, transaction);
+                emitEvent(disputeCreatedEvent);
+                return disputeCreatedEvent;
+            }).orElseThrow(() ->
+                    new RuntimeException(format("LedgerTransaction with gateway transaction id [%s] not found",
+                            stripeDisputeData.getPaymentIntentId())));
+        } else {
+            throw new RuntimeException("Unknown webhook task: " + stripeNotification.getType());
```
nlsteers, line 50: The task type is already validated in the `TaskQueueMessageHandler`'s `switch` statement, do we need to guard against unknown task types here?

stephencdaly (approved):

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/StripeWebhookTaskHandler.java
SandorArpa: It doesn't hurt I think.

nlsteers: Ah, I see you're checking for specific webhook tasks

stephencdaly: This is checking the webhook type rather than the task type, so it does make sense to do this I think.

PR #3492, opened 2022-03-03
Pp 9330 endpoint to create agreement

kundanscorpio:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/agreement/model/AgreementEntity.java
+++ b/src/main/java/uk/gov/pay/connector/agreement/model/AgreementEntity.java
@@ -1,0 +1,168 @@
+package uk.gov.pay.connector.agreement.model;
+
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.util.RandomIdGenerator;
+import uk.gov.service.payments.commons.jpa.InstantToUtcTimestampWithoutTimeZoneConverter;
+
+import javax.persistence.Access;
+import javax.persistence.AccessType;
+import javax.persistence.Entity;
+import javax.persistence.Table;
+import javax.persistence.Id;
+import javax.persistence.Column;
+import javax.persistence.GeneratedValue;
+import javax.persistence.SequenceGenerator;
+import javax.persistence.ManyToOne;
+import javax.persistence.GenerationType;
+import javax.persistence.JoinColumn;
+import javax.persistence.Convert;
+
+import java.time.Instant;
+
+@Entity
+@Table(name = "agreements")
+@Access(AccessType.FIELD)
+@SequenceGenerator(name = "agreements_id_seq",
+        sequenceName = "agreements_id_seq", allocationSize = 1)
+public class AgreementEntity {
+
+    @Id
+    @Column(name = "id")
+    @GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "agreements_id_seq")
+    private Long id;
+
+    @ManyToOne
+    @JoinColumn(name = "gateway_account_id", updatable = false)
+    private GatewayAccountEntity gatewayAccount;
+
+    @Column(name="external_id")
+    private String externalId;
+
+    @Column(name = "created_date")
+    @Convert(converter = InstantToUtcTimestampWithoutTimeZoneConverter.class)
+    private Instant createdDate;
+
+    @Column(name = "reference")
+    private String reference;
+    
+    @Column(name = "service_id")
+    private String serviceId;
+    
+    @Column(name = "live")
+    private boolean live;
+
+    public AgreementEntity() {
+        // For JPA
+    }
+    
+    private AgreementEntity(GatewayAccountEntity gatewayAccount, String serviceId, String reference,
+                            boolean live, Instant createdDate) {
+        this.externalId = RandomIdGenerator.newId();
+        this.gatewayAccount = gatewayAccount;
+        this.serviceId = serviceId;
+        this.reference = reference;
+        this.live = live;
+        this.createdDate = createdDate;
+    }
+
+    public Long getId() {
+        return id;
+    }
+
+    public void setId(Long id) {
+        this.id = id;
+    }
+
+    public GatewayAccountEntity getGatewayAccount() {
+        return gatewayAccount;
+    }
+
+    public void setGatewayAccount(GatewayAccountEntity gatewayAccount) {
+        this.gatewayAccount = gatewayAccount;
+    }
+
+    public String getExternalId() {
+        return externalId;
+    }
+
+    public void setExternalId(String externalId) {
+        this.externalId = externalId;
+    }
+
+    public Instant getCreatedDate() {
+        return createdDate;
+    }
+
+    public void setCreatedDate(Instant createdDate) {
+        this.createdDate = createdDate;
+    }
+
+    public String getReference() {
+        return reference;
+    }
+
+    public void setReference(String reference) {
+        this.reference = reference;
+    }
+
+    public String getServiceId() {
+        return serviceId;
+    }
+
+    public void setServiceId(String serviceId) {
+        this.serviceId = serviceId;
+    }
+
+    public boolean isLive() {
+        return live;
+    }
+
+    public void setLive(boolean live) {
+        this.live = live;
+    }
+    
+    public static class AgreementEntityBuilder {
+        private GatewayAccountEntity gatewayAccount;
+        private Instant createdDate;
+        private String reference;
+        private String serviceId;
+        private boolean live;
+
+        public static AgreementEntityBuilder anAgreementEntity(Instant createdDate) {
+            var agreementEntityBuilder = new AgreementEntityBuilder();
+            agreementEntityBuilder.withCreatedDate(createdDate);
+            return agreementEntityBuilder;
+        }
+        
+        public AgreementEntityBuilder withGatewayAccount(GatewayAccountEntity gatewayAccountEntity) {
+            this.gatewayAccount = gatewayAccountEntity;
+            return this;
+        }
+
+        public AgreementEntityBuilder withCreatedDate(Instant createdDate) {
+            this.createdDate = createdDate;
+            return this;
+        }
+
+        public AgreementEntityBuilder withReference(String reference) {
+            this.reference = reference;
+            return this;
+        }
+
+        public AgreementEntityBuilder withServiceId(String serviceId) {
+            this.serviceId = serviceId;
+            return this;
+        }
+
+        public AgreementEntityBuilder withLive(boolean live) {
+            this.live = live;
+            return this;
+        }
+
+        public AgreementEntity build() {
+            return new AgreementEntity(gatewayAccount, serviceId, reference, live, createdDate);
+        }
+        
+    }
+    
+}
--- a/src/main/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResource.java
@@ -1,0 +1,41 @@
+package uk.gov.pay.connector.agreement.resource;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.agreement.model.AgreementCreateRequest;
+import uk.gov.pay.connector.agreement.model.AgreementResponse;
+import uk.gov.pay.connector.agreement.service.AgreementService;
+
+import javax.inject.Inject;
+import javax.validation.Valid;
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.NotFoundException;
+
+@Path("/")
+public class AgreementsApiResource {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AgreementsApiResource.class);
+
+    private final AgreementService agreementService;
+
+    @Inject
+    public AgreementsApiResource(AgreementService agreementService) {
+        this.agreementService = agreementService;
+    }
+
+    @POST
+    @Path("/v1/api/accounts/{accountId}/agreements")
+    @Produces("application/json")
+    @Consumes("application/json")
+    public AgreementResponse createAgreement(
+            @PathParam("accountId") Long accountId,
+            @Valid AgreementCreateRequest agreementCreateRequest
+    ) {
+        LOGGER.info("Creating new agreement for gateway account ID {}", accountId);
+        return agreementService.create(agreementCreateRequest, accountId).orElseThrow(NotFoundException::new);
+    }
+}
--- a/src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceIT.java
@@ -1,0 +1,133 @@
+package uk.gov.pay.connector.agreement.resource;
+
+
+import com.github.tomakehurst.wiremock.WireMockServer;
+import io.restassured.response.ValidatableResponse;
+import io.restassured.specification.RequestSpecification;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.testcontainers.shaded.com.fasterxml.jackson.core.JsonProcessingException;
+import org.testcontainers.shaded.com.fasterxml.jackson.databind.ObjectMapper;
+import uk.gov.pay.connector.app.ConnectorApp;
+import uk.gov.pay.connector.it.dao.DatabaseFixtures;
+import uk.gov.pay.connector.junit.DropwizardConfig;
+import uk.gov.pay.connector.junit.DropwizardJUnitRunner;
+import uk.gov.pay.connector.junit.DropwizardTestContext;
+import uk.gov.pay.connector.junit.TestContext;
+import uk.gov.pay.connector.rules.WorldpayMockClient;
+import uk.gov.pay.connector.util.DatabaseTestHelper;
+
+import java.util.Map;
+
+import static io.restassured.RestAssured.given;
+import static io.restassured.http.ContentType.JSON;
+import static java.time.temporal.ChronoUnit.SECONDS;
+import static org.apache.http.HttpStatus.SC_OK;
+import static java.lang.String.format;
+import static org.apache.commons.lang3.RandomUtils.nextLong;
+import static org.apache.http.HttpStatus.SC_UNPROCESSABLE_ENTITY;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.matchesPattern;
+import static uk.gov.pay.connector.matcher.ZoneDateTimeAsStringWithinMatcher.isWithin;
+
+@RunWith(DropwizardJUnitRunner.class)
+@DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml")
+public class AgreementsApiResourceIT {
+    private static final String REFERENCE_ID = "1234";
+    private static final String REFERENCE_ID_TOO_LONG = new String(new char[260]).replace('\0', '1');
+    private static final String REFERENCE_ID_EMPTY = "";
+    private DatabaseFixtures.TestAccount testAccount;
+
+    private static final String CREATE_AGREEMENT_URL = "/v1/api/accounts/%s/agreements";
+
+    @DropwizardTestContext
+    private TestContext testContext;
+
+    private DatabaseTestHelper databaseTestHelper;
+    private WireMockServer wireMockServer;
+    private WorldpayMockClient worldpayMockClient;
+    private DatabaseFixtures databaseFixtures;
+    private Long credentialsId;
+    private String credentialsExternalId;
+    private Long accountId;
+    private ObjectMapper objectMapper = new ObjectMapper();
+    
+    public AgreementsApiResourceIT() {
+    }
+    
+    @Before
+    public void setUp() {
+        databaseTestHelper = testContext.getDatabaseTestHelper();
+        wireMockServer = testContext.getWireMockServer();
+        worldpayMockClient = new WorldpayMockClient(wireMockServer);
+        databaseFixtures = DatabaseFixtures.withDatabaseTestHelper(databaseTestHelper);
+
+        testAccount = createTestAccount("worldpay");
+        accountId = testAccount.getAccountId();
+
+        credentialsId = testAccount.getCredentials().get(0).getId();
+        credentialsExternalId = testAccount.getCredentials().get(0).getExternalId();
+    }
+
+    protected RequestSpecification givenSetup() {
+        return given().port(testContext.getPort()).contentType(JSON);
+    }
+    
+    @Test
+    public void shouldCreateAgreement() throws JsonProcessingException {
+        String payload = objectMapper.writeValueAsString(Map.of(
+                "reference", REFERENCE_ID
+        ));
+
+        ValidatableResponse o = givenSetup()
+                .body(payload)
+                .post(format(CREATE_AGREEMENT_URL, accountId))
+
+                .then()
+                .statusCode(SC_OK)
+                .body("reference", equalTo(REFERENCE_ID))
+                .body("service_id", equalTo("valid-external-service-id"))
+                .body("agreement_id", notNullValue())
+                .body("created_date", notNullValue())
+                .body("created_date", matchesPattern("^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(.\\d{1,3})?Z"))
+                .body("created_date", isWithin(10, SECONDS));
+    }
+
+    @Test
+    public void shouldReturn422WhenReferenceIdTooLong() throws JsonProcessingException {
+        String payload = objectMapper.writeValueAsString(Map.of(
+                "reference", REFERENCE_ID_TOO_LONG
+        ));
+
+        givenSetup()
+                .body(payload)
+                .post(format(CREATE_AGREEMENT_URL, accountId))
+                .then()
+                .statusCode(SC_UNPROCESSABLE_ENTITY);
+    }
+
+    @Test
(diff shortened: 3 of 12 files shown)


src/main/java/uk/gov/pay/connector/agreement/model/AgreementEntity.java
```diff
@@ -0,0 +1,157 @@
+package uk.gov.pay.connector.agreement.model;
+
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.util.RandomIdGenerator;
+import uk.gov.service.payments.commons.jpa.InstantToUtcTimestampWithoutTimeZoneConverter;
+
+import javax.persistence.*;
```
gidsg: We try to avoid wildcard imports, see: https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

nlsteers:
Can you fill out the comment and squash some of these commits to tidy up? Commit messages should also include the ticket number so that they are mapped to the Jira story automatically.

src/main/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResource.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.agreement.model.AgreementCreateRequest;
+import uk.gov.pay.connector.agreement.model.AgreementResponse;
+import uk.gov.pay.connector.agreement.service.AgreementService;
+
+import javax.inject.Inject;
+import javax.validation.Valid;
+import javax.ws.rs.*;
```
gidsg: See other comment https://github.com/alphagov/pay-connector/pull/3492/files#r818574925

src/main/java/uk/gov/pay/connector/agreement/model/AgreementEntity.java
alexbishop1: We do not use wildcard imports (see https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports for why).

src/main/java/uk/gov/pay/connector/agreement/model/AgreementEntity.java
```diff
@@ -0,0 +1,157 @@
...
+    @Column(name = "live")
+    private boolean live;
+
+    public AgreementEntity() {
+        // For JPA
+    }
+    
+    private AgreementEntity(GatewayAccountEntity gatewayAccount, String serviceId, String reference,
+                            boolean live, Instant createdDate) {
+        //AgreementEntity agreementEntity = new AgreementEntity();
```
alexbishop1: Delete this please.

src/main/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResource.java
alexbishop1: Another wildcard.

src/main/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResource.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.UriInfo;
+
+@Path("/")
+public class AgreementsApiResource {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AgreementsApiResource.class);
+
+    private static final String ACCOUNT_ID = "accountId";
```
alexbishop1: This constant is only used once so just hardcode `"accountId"` there.

src/main/resources/migrations.xml
```diff
@@ -1738,7 +1738,7 @@
 
     <changeSet id="create agreements table" author="">
         <createTable tableName="agreements">
-            <column name="id" type="bigserial" autoIncrement="true">
+            <column name="agreement_id" type="bigserial" autoIncrement="true">
```
alexbishop1: This does not seem right. We always use `id` for the primary keys of our tables. If there’s something that’s preventing it from working as it currently is, we should figure that out and fix it.

Additionally, this migration has already been run in the test, staging and production environments, so we cannot change it (we can add new migrations if necessary, of course).

Finally, we do not include migrations and code changes in the same pull request because there’s one release per pull request. The usual strategy is to create a PR with the migration, then a separate PR with the code change, which should assume the migration has already been run.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceIT.java
```diff
@@ -0,0 +1,122 @@
...
+import static org.apache.http.HttpStatus.SC_OK;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.notNullValue;
+import static java.lang.String.format;
+import static org.apache.commons.lang3.RandomUtils.nextLong;
+import static uk.gov.pay.connector.util.AddGatewayAccountCredentialsParams.AddGatewayAccountCredentialsParamsBuilder.anAddGatewayAccountCredentialsParams;
+
+@RunWith(DropwizardJUnitRunner.class)
+@DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml")
+public class AgreementResourceIT {
```
alexbishop1: Could you name this `AgreementsApiResourceIT` for consistency please?

Since this is a whole new test file, could you try writing it in JUnit 5? See https://docs.google.com/presentation/d/1Z1VSjvhxU7tffKTx6vCHd3AY5WIyZLE0awOjOiHsJ2Q/edit for a five-minute guide to the differences from JUnit 4 (including the annotations required to do a Dropwizard integration test). Happy to provide further help if you need it.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceIT.java
```diff
@@ -0,0 +1,122 @@
...
+        credentialsId = testAccount.getCredentials().get(0).getId();
+        credentialsExternalId = testAccount.getCredentials().get(0).getExternalId();
+    }
+
+    protected RequestSpecification givenSetup() {
+        return given().port(testContext.getPort()).contentType(JSON);
+    }
+    
+    @Test
+    public void testShouldCreateAgreement() throws JsonProcessingException {
```
alexbishop1: This isn’t JUnit 3, you don’t need to prefix the test name with ‘test’! 😀

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceIT.java
```diff
@@ -0,0 +1,122 @@
...
+        ValidatableResponse o = givenSetup()
+                .body(payload)
+                .post(format(CREATE_AGREEMENT_URL, accountId))
+
+                .then()
+                .statusCode(SC_OK)
+                .body("reference", equalTo(REFERENCE_ID))
+                .body("service_id", equalTo("valid-external-service-id"))
+                .body("agreement_id", notNullValue())
+                .body("created_date", notNullValue());
```
alexbishop1: Our usual pattern for doing this in connector (which is horrible because it relies on the current time) is to do this:

```java
.body("created_date", matchesPattern("^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}(.\\d{1,3})?Z"))
.body("created_date", isWithin(10, SECONDS));
```

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceIT.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    private DatabaseFixtures.TestAccount addGatewayAccountAndCredential(String paymentProvider) {
+        long accountId = nextLong(2, 10000);
+        LocalDateTime createdDate = LocalDate.parse("2021-01-01").atStartOfDay();
+        LocalDateTime activeStartDate = LocalDate.parse("2021-02-01").atStartOfDay();
+        LocalDateTime activeEndDate = LocalDate.parse("2021-03-01").atStartOfDay();
```
alexbishop1: Since these get converted to `Instant`s later, we can just do:

```
var createdDate = Instant.parse("2021-01-01T00:00:00");
var activeStartDate = Instant.parse("2021-02-01T00:00:00");
var activeEndDate = Instant.parse("2021-03-01T00:00:00");
```

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceIT.java
```diff
@@ -0,0 +1,122 @@
...
+                .withPaymentProvider(paymentProvider)
+                .withCreatedDate(createdDate.toInstant(ZoneOffset.UTC))
+                .withActiveStartDate(activeStartDate.toInstant(ZoneOffset.UTC))
+                .withActiveEndDate(activeEndDate.toInstant(ZoneOffset.UTC))
+                .withState(GatewayAccountCredentialState.ACTIVE)
+                .withCredentials(Map.of(
+                        "merchant_id", "a-merchant-id",
+                        "username", "a-username",
+                        "password", "a-password"))
+                .build();
```
alexbishop1: Do we actually need to add the credentials to make the test work? It’s a lot of code for something we don’t actually care about for this test.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceIT.java
```diff
@@ -0,0 +1,122 @@
...
+
+        return databaseFixtures.aTestAccount().withPaymentProvider(paymentProvider)
+                .withIntegrationVersion3ds(2)
+                .withAccountId(accountId)
+                .withGatewayAccountCredentials(Collections.singletonList(credentialsParams))
+                .insert();
+    }
+}
```
alexbishop1: We could have a test in here that checks we do something sensible when we get an invalid reference in the request (too long etc.) or not reference at all.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+    private static AgreementService agreementService = mock(AgreementService.class);
+    
+    private static final ResourceExtension resources = ResourceTestRuleWithCustomExceptionMappersBuilder
+            .getBuilder()
+            .addResource(new AgreementsApiResource(agreementService))
+            .build();
+
+    @Test
+    void shouldReturn200_whenPostToAgreement() {
+        AgreementResponse ar = new AgreementResponseBuilder().withAgreementId("aid").withReference(REFERENCE_ID).withServiceId("serviceid").withCreatedDate(Instant.now()).build();
```
alexbishop1: Best to avoid relying on the current time in tests: since the actual time does not matter, we could just use `Instant.parse("2022-03-03T12:00:00Z")` (feel free to add an air of mystery by choosing a date that’s meaningful to you but nobody else!).

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+        when(agreementService.create(any(AgreementCreateRequest.class), any(Long.class))).thenReturn(Optional.ofNullable(ar));
```
alexbishop1: Not a fan of all the `any`s. For `any(AgreementCreateRequest.class)` we could instead stick an `equals(…)` method on `AgreementCreateRequest` so we can easily compare the expected and actual. For `any(Long.class)` you should be able to use the actual gateway account ID.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+        Map myMap = Map.of("reference", REFERENCE_ID);
```
alexbishop1: `myMap` is not the most descriptive name: I’d just inline this into the `post(…)` line below.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+    private final static String RESOURCE_URL = "/v1/api/accounts/%d/agreements";
+
+    private final static long VALID_ACCOUNT_ID = 123l;
+
+    private final static long NOT_VALID_ACCOUNT_ID = 9876l;
+
+    private static final String REFERENCE_ID = "1234";
+    
+    private static AgreementService agreementService = mock(AgreementService.class);
```
alexbishop1: I don’t think this wants to be static.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+                .post(Entity.json(myMap));
+
+        assertThat(response.getStatus(), is(OK_200));
+    }
+
+    @Test
+    void shouldReturn404_whenPostWithUnknownAccountId() {
+        NotFoundException nfe = new NotFoundException();
+        AgreementResponse ar = new AgreementResponseBuilder().withAgreementId("aid").withReference(REFERENCE_ID).withServiceId("serviceid").withCreatedDate(Instant.now()).build();
+        when(agreementService.create(any(AgreementCreateRequest.class), eq(NOT_VALID_ACCOUNT_ID))).thenThrow(nfe);
```
alexbishop1: This can be:

```java
when(agreementService.create(any(AgreementCreateRequest.class), eq(NOT_VALID_ACCOUNT_ID))).thenThrow(NotFoundException.class);
```

… eliminating the need for the `nfe` local variable.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+        Map myMap = Map.of("reference", REFERENCE_ID);
```
alexbishop1: 🗺️

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+import junitparams.JUnitParamsRunner;
+import uk.gov.pay.connector.agreement.dao.AgreementDao;
+import uk.gov.pay.connector.agreement.model.AgreementCreateRequest;
+import uk.gov.pay.connector.agreement.model.AgreementResponse;
+import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+
+
+@RunWith(JUnitParamsRunner.class)
```
alexbishop1: This does not seem to use parameterised tests, so it can just use  the standard Mockito runner.

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+public class AgreementServiceTest {
```
alexbishop1, line 25: Again, since this is a whole new test class, please use JUnit 5.

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private AgreementDao mockedAgreementDao;
+
+    @Mock
+    private GatewayAccountDao mockedGatewayAccountDao;
+    
+    private AgreementService service;
+    
+    @Before
+    public void setUp() {
+        service = new AgreementService(mockedAgreementDao, mockedGatewayAccountDao, Clock.systemUTC());
```
alexbishop1: Use `Clock.fixed(…)` rather than something that depends on the actual real time.

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @Test
+    public void shouldCreateAnAgreement() { 
+        when(gatewayAccount.getServiceId()).thenReturn(SERVICE_ID);
+        when(gatewayAccount.isLive()).thenReturn(false);
+    	when(mockedGatewayAccountDao.findById(GATEWAY_ACCOUNT_ID)).thenReturn(Optional.of(gatewayAccount));
+
+    	AgreementCreateRequest agreementCreateRequest = new AgreementCreateRequest(REFERENCE_ID);    	
+    	Optional<AgreementResponse> response = service.create(agreementCreateRequest, GATEWAY_ACCOUNT_ID);
+
+    	assertNotNull(response.get());
```
alexbishop1: If there was an empty Optional, the `response.get()` call would actually throw a NullPointerException, so the `assertNotNull(…)` call would never be made. Usually we do something like `assertThat(response.isPresent(), is(true));`.

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    	assertEquals(REFERENCE_ID, response.get().getReference());
+    	assertEquals(SERVICE_ID, response.get().getServiceId());
```
alexbishop1: Usually we prefer `assertThat(actual, is(expected))` to `assertEquals(expected, actual)`.

alexbishop1 (changes requested):
Looks fairly good. I think all the builder stuff to create the domain objects might be a bit overwrought but I encouraged you down that route so I’m not going to complain!

Made a few suggestions, which are mostly minor code changes and some additional tests. We seem to be testing all the substantial new classes except `AgreementDao` but that does so little right now that I think it’s OK.

The sequence of commits is a little circuitous, with lots of in-progress commits and fixes for things that broke along the way. Usually we try to do some squashing to avoid this. It’s OK for a PR to consist of several commits, each delivering a new little bit of value, but we try to avoid commits that reveal the back-and-forth stuff that happened while the code was being written but never got released. In this case, I’d probably just squash all the commits into one.

Also, each commit should start with the story number (so ‘PP-9330’ in this case): it’s useful for future readers of the change history and allows Jira to link the commits to the appropriate story.

In terms of code size, this is about the limit of what we’d want a PR to be.

pom.xml
```diff
@@ -16,7 +16,7 @@
         <wiremock.version>2.32.0</wiremock.version>
         <eclipselink.version>2.7.10</eclipselink.version>
         <guice.version>5.1.0</guice.version>
-        <jackson.version>2.13.2</jackson.version>
+        <jackson.version>2.13.1</jackson.version>
```
gidsg: These look like downgrades, guess this file should be reverted to master.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+            .getBuilder()
+            .addResource(new AgreementsApiResource(agreementService))
+            .build();
+
+    @Test
+    void shouldReturn200_whenPostToAgreement() {
+        AgreementResponse ar = new AgreementResponseBuilder().withAgreementId("aid").withReference(REFERENCE_ID).withServiceId("serviceid").withCreatedDate(Instant.now()).build();
+        when(agreementService.create(any(AgreementCreateRequest.class), any(Long.class))).thenReturn(Optional.ofNullable(ar));
+
+        Map myMap = Map.of("reference", REFERENCE_ID);
```
gidsg: Can we give this a more descriptive name, or just inline it?

pom.xml
```diff
@@ -208,7 +208,7 @@
         <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
-            <version>1.2.11</version>
+            <version>1.2.10</version>
```
alexbishop1: Not sure what happened here but we definitely do not want to be downgrading dependencies.

src/main/java/uk/gov/pay/connector/queue/tasks/PaymentTaskMessage.java
```diff
@@ -0,0 +1 @@
+
```
alexbishop1: 🤨

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceIT.java
```diff
@@ -0,0 +1,122 @@
...
+    private DatabaseTestHelper databaseTestHelper;
+    private WireMockServer wireMockServer;
+    private WorldpayMockClient worldpayMockClient;
+    private DatabaseFixtures databaseFixtures;
+    private Long credentialsId;
+    private String credentialsExternalId;
+    private Long accountId;
+    private ObjectMapper objectMapper = new ObjectMapper();
+    
+    public AgreementsApiResourceIT() {
```
alexbishop1, line 57: I left a few comments about this file in my last review (back when it was called `AgreementResourceIT.java`). As discussed earlier, fine to leave the test as JUnit 4 but my other comments still stand (happy to debate any of them if you disagree, of course).

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceTest.java
```diff
@@ -0,0 +1,81 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.ArgumentMatchers.eq;
+
+import static org.mockito.Mockito.when;
+
+import uk.gov.pay.connector.agreement.model.AgreementCreateRequest;
+
+@ExtendWith(DropwizardExtensionsSupport.class)
+class AgreementsApiResourceTest {
```
alexbishop1, line 34: Again, a few comments from my last review about this file (back when it was called `AgreementResourceTest`), which still need looking at.

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+public class AgreementServiceTest {
+
+    private static final String SERVICE_ID = "TestAgreementServiceID";
+
+    private static final long GATEWAY_ACCOUNT_ID = 10L;
+    
+    private  static final String REFERENCE_ID = "test";
+
+    @Rule
+    public MockitoRule rule = MockitoJUnit.rule();
```
alexbishop1: Is this rule actually being used for anything?

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private AgreementDao mockedAgreementDao = mock(AgreementDao.class);
+
+    private GatewayAccountDao mockedGatewayAccountDao = mock(GatewayAccountDao.class);
+    
+    private AgreementService service;
+    
+    @Before
+    public void setUp() {
+        String instantExpected = "2022-03-03T10:15:30Z";
+        Clock clock = Clock.fixed(Instant.parse(instantExpected), ZoneId.of("UTC"));
```
alexbishop1: `ZoneOffset.UTC` is _slightly_ shorter than `ZoneId.of("UTC")` (and avoids a magic string literal).

src/test/java/uk/gov/pay/connector/agreement/service/AgreementServiceTest.java
```diff
@@ -0,0 +1,67 @@
...
+        when(gatewayAccount.getServiceId()).thenReturn(SERVICE_ID);
+        when(gatewayAccount.isLive()).thenReturn(false);
+    	when(mockedGatewayAccountDao.findById(GATEWAY_ACCOUNT_ID)).thenReturn(Optional.of(gatewayAccount));
+
+    	AgreementCreateRequest agreementCreateRequest = new AgreementCreateRequest(REFERENCE_ID);    	
+    	Optional<AgreementResponse> response = service.create(agreementCreateRequest, GATEWAY_ACCOUNT_ID);
+
+    	assertThat(response.isPresent(), is(true));
+    	assertThat(REFERENCE_ID, is(response.get().getReference()));
+        assertThat(SERVICE_ID, is(response.get().getServiceId()));
```
alexbishop1: Indentation has gone a bit wonky here.

alexbishop1 (changes requested):
Thanks for making the changes but it looks like a few things I commented on last time have not been addressed. Happy to discuss if you disagree with any of them or want more reasoning.

pom.xml
jbz149: the pom.xml is not from us.
we will remove it on the next squash.

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceTest.java
jbz149: All comments reviewed

pom.xml
alexbishop1: I assume we’re still waiting on that squash!

src/main/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResource.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @POST
+    @Path("/v1/api/accounts/{accountId}/agreements")
+    @Produces("application/json")
+    @Consumes("application/json")
+    public AgreementResponse createAgreement(
+            @PathParam("accountId") Long accountId,
+            @Valid AgreementCreateRequest agreementCreateRequest,
+            @Context UriInfo uriInfo
```
alexbishop1: Not spotted this before now but I think we can remove the `UriInfo` because it’s never used.

alexbishop1 (changes requested):
In terms of the code, I’m happy apart from one unused method parameter (noted in a comment below). The mixture of JUnit 4 and 5 is fine for now.

The commit message is not great at current. Could it be reworded? Something like “PP-9930 Endpoint to create agreement” followed by any other pertinent information would be fine.

In addition, the tests seem to be failing, which will need to be fixed before this is merged. Happy to assist with looking into this.

alexbishop1 (approved):

PR #3500, opened 2022-03-07
PP-9370 Fix Stripe account updated Splunk alert

stephencdaly:
The Splunk alert that informs us of whether there is action from support
required when receive an `account.updated` webhook from Stripe relies on
the `requirements` containing an `errors` array. The Splunk alert should
fire if this array is not empty.

As part of a recent change, we started deserialising the `requirements`
object in the webhook, whereas before we were just logging it in its
un-deserialised state. As we were only deserialising the `currently_due`
and `past_due` fields this meant we are no longer logging the `errors`.

Deserialise the `requirements` object in its entirety to fix the alert.
Include all fields, not just `errors` as these can be useful for the
support person to diagnose what needs updating for the account.


SandorArpa (approved):
LOKTM

PR #3511, opened 2022-03-10
PP-9315 Support Worldpay 3DS2 non-JS mode

gidsg:
## WHAT YOU DID
Implement Worldpay 3DS2 flow as described in https://developer.worldpay.com/docs/wpg/directintegration/3ds2#payment-requests-with-javascript-not-enabled

Note: `dfReferenceId=""` is required for Worldpay to accept the payload even though this is not well documented.

## How to test

Tests should make sense. `WorldpayPaymentProviderTest.java` can be run locally if environment variables are set. 

I will test end-to-end including frontend once this is deployed on Test.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
+++ b/src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
@@ -39,6 +39,7 @@ import uk.gov.pay.connector.logging.AuthorisationLogger;
 import uk.gov.pay.connector.paymentprocessor.service.AuthorisationService;
 import uk.gov.pay.connector.paymentprocessor.service.CardExecutorService;
 import uk.gov.pay.connector.refund.model.domain.RefundEntity;
+import uk.gov.pay.connector.util.AcceptLanguageHeaderParser;
 
 import javax.ws.rs.client.ClientBuilder;
 import java.io.IOException;
@@ -53,6 +54,7 @@ import static org.hamcrest.CoreMatchers.not;
 import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.startsWith;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
@@ -166,13 +168,13 @@ public class WorldpayPaymentProviderTest {
         assertTrue(response.getBaseResponse().get().getLastEvent().isPresent());
         assertEquals(response.getBaseResponse().get().getLastEvent().get(), "AUTHORISED");
     }
-    
+
     @Test
     public void submitAuthRequestWithExemptionEngineFlag() {
         validGatewayAccount.setRequires3ds(true);
         validGatewayAccount.setIntegrationVersion3ds(2);
         validGatewayAccount.setWorldpay3dsFlexCredentialsEntity(aWorldpay3dsFlexCredentialsEntity().withExemptionEngine(true).build());
-        
+
         WorldpayPaymentProvider paymentProvider = getValidWorldpayPaymentProvider();
         // card holder of "EE.HONOURED_ISSUER_HONOURED.AUTHORISED" elicits an authorised response, see https://developer.worldpay.com/docs/wpg/scaexemptionservices/exemptionengine#testing-exemption-engine
         var authCardDetails = anAuthCardDetails()
@@ -334,6 +336,41 @@ public class WorldpayPaymentProviderTest {
         });
     }
 
+    @Test
+    public void shouldBeAbleToSendAuthorisationRequestForMerchantUsing3dsFlexWithNonJs() {
+        WorldpayPaymentProvider paymentProvider = getValidWorldpayPaymentProvider();
+
+        validGatewayAccount.setRequires3ds(true);
+        validGatewayAccount.setSendPayerEmailToGateway(true);
+        validGatewayAccount.setIntegrationVersion3ds(2);
+
+        ChargeEntity charge = aValidChargeEntity()
+                .withTransactionId(randomUUID().toString())
+                .withEmail("payer@email.test")
+                .withGatewayAccountEntity(validGatewayAccount)
+                .withGatewayAccountCredentialsEntity(validGatewayAccountCredentialsEntity)
+                .build();
+
+        AuthCardDetails authCardDetails = anAuthCardDetails()
+                .withCardHolder(MAGIC_CARDHOLDER_NAME_FOR_3DS_FLEX_CHALLENGE_REQUIRED_RESPONSE)
+                .withWorldpay3dsFlexDdcResult(null)
+                .build();
+
+        CardAuthorisationGatewayRequest request = new CardAuthorisationGatewayRequest(charge, authCardDetails);
+
+        GatewayResponse<WorldpayOrderStatusResponse> response = paymentProvider.authorise(request);
+
+        assertTrue(response.getSessionIdentifier().isPresent());
+        response.getBaseResponse().ifPresent(res -> {
+            assertThat(res.getGatewayParamsFor3ds().isPresent(), is(true));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getWorldpayChallengeAcsUrl(), is(notNullValue()));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getWorldpayChallengeTransactionId(), is(notNullValue()));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getWorldpayChallengePayload(), is(notNullValue()));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getThreeDsVersion(), startsWith("2."));
+                
+        });
+    }
+
     /**
      * Worldpay does not care about a successful authorization reference to make a capture request.
      * It simply accepts anything as long as the request is well formed. (And ignores it silently)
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/connector/util/LanguageHeaderParser.java
```diff
@@ -0,0 +1,23 @@
+package uk.gov.pay.connector.util;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Optional;
+
+public class LanguageHeaderParser {
+
+    public static String getPreferredLocaleFromAcceptHeader(String acceptLanguageHeader) {
```
alexbishop1: It would be more accurate to call the class `AcceptLanguageHeaderParser` and the method `getPreferredLanguageFromAcceptLanguageHeader(…)`.

I also think I’d prefer it if this was not static and parser were injected where it’s needed.

src/main/java/uk/gov/pay/connector/util/LanguageHeaderParser.java
```diff
@@ -0,0 +1,23 @@
...
+        final var DEFAULT_LOCALE = "en-GB";
```
alexbishop1: This should probably be a `private static final` constant.

src/main/java/uk/gov/pay/connector/util/LanguageHeaderParser.java
```diff
@@ -0,0 +1,23 @@
...
+        Optional<String> maybeAcceptLanguageHeader = Optional.ofNullable(acceptLanguageHeader);
+        return maybeAcceptLanguageHeader.map(alh -> {
+                    try {
+                        List<Locale.LanguageRange> languageRanges = Locale.LanguageRange.parse(alh);
+                        Optional<Locale> locale = Optional.ofNullable(Locale.lookup(languageRanges, List.of(Locale.getAvailableLocales())));
+                        return locale.map(Locale::toLanguageTag).orElse(DEFAULT_LOCALE);
```
alexbishop1: I think this is overkill. I do not think it’s necessary to check the language tag is recognised by Java. The parser will make sure the language tag is well-formed, which I think should be enough. The Worldpay 3DS Flex docs put no constraints on the content of the `<browserLanguage>` element so I assume anything that’s well-formed should be good enough. I doubt it’s going to explode if we send an unrecognised (but well-formed) language tag.

So I think it can be as simple as this (some stuff extracted to a private method to minimise the coverage of the try/catch block):

```java
public static String getPreferredLanguageFromAcceptLanguageHeader(String acceptLanguageHeader) {
    return parseLanguageRanges(acceptLanguageHeader)
                .stream()
                .map(Locale.LanguageRange::getRange)
                .filter(not(languageRange -> languageRange.equals("*")))
                .findFirst()
                .orElse(DEFAULT_LOCALE);
}

private static List<Locale.LanguageRange> parseLanguageRanges(String acceptLanguageHeader) {
    if (acceptLanguageHeader == null) {
        return Collections.emptyList();
    }
    try {
        return Locale.LanguageRange.parse(acceptLanguageHeader);
    } catch (IllegalArgumentException e) {
        return Collections.emptyList();
    }
}

```

src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
```diff
@@ -334,6 +335,41 @@ public void shouldBeAbleToSendAuthorisationRequestForMerchantUsing3dsWithPayerEm
         });
     }
 
+    @Test
+    public void shouldBeAbleToSendAuthorisationRequestForMerchantUsing3dsWithNonJs() {
```
alexbishop1: `Using3dsFlex` really.

src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
```diff
@@ -334,6 +335,41 @@ public void shouldBeAbleToSendAuthorisationRequestForMerchantUsing3dsWithPayerEm
...
+
+        GatewayResponse<WorldpayOrderStatusResponse> response = paymentProvider.authorise(request);
+
+        assertTrue(response.getSessionIdentifier().isPresent());
+        response.getBaseResponse().ifPresent(res -> {
+            assertThat(res.getGatewayParamsFor3ds().isPresent(), is(true));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getWorldpayChallengeAcsUrl(), is(notNullValue()));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getWorldpayChallengeTransactionId(), is(notNullValue()));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getWorldpayChallengePayload(), is(notNullValue()));
+            assertThat(res.getGatewayParamsFor3ds().get().toAuth3dsRequiredEntity().getThreeDsVersion(), startsWith("2"));
```
alexbishop1: `"2."` maybe?

alexbishop1 (changes requested):
Looks good but there’s something I think can be simplified.

src/main/java/uk/gov/pay/connector/util/LanguageHeaderParser.java
gidsg: @alexbishop1 The `toLanguageTag` here is doing some extra work here. Not just checking the locale is recognised. It converts to the `languageTag` which changes e.g. from `en-gb` (which is what is returned by the language range) to `en-GB` which matches what's in Worldpay's documentation.

alexbishop1: I would hope the case would not matter but I see your point about matching Worldpay’s example. I wonder if there’s a way to normalise the language tag without requiring Java to recognise it.

alexbishop1: OK, I’ve been playing with this and it seems that if we call `Locale.forLanguageTag(…).toLanguageTag()` we’ll get normalisation without restricting ourselves to only locales that Java recognises.

For example, `Locale.forLanguageTag("en-gb").toLanguageTag()` gives us `"en-GB"` and `Locale.forLanguageTag("sR-cYRl-Hs").toLanguageTag()` (Serbian in Cyrllic script as used in Serbia) gives us `"sr-Cyrl-HS"`.

The important part is that is seems to be cool about unrecognised language tags: `Locale.forLanguageTag("yy-xy").toLanguageTag()` gives us `"yy-XY"`.

We do have to be careful to filter out `*` as that gives us `"und"` (we are of course assuming that Worldpay does not want us to send `*`).

[`Locale.forLanguageTag(…)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Locale.html#forLanguageTag(java.lang.String)) does do some other conversions but I guess these will be OK.

Of course, this could still be taking us quite far from what the browser gives us but we’re not trying to preserve what the browser sent us, we’re trying to send a useful value to Worldpay. So making sure what the browser gives us is well-formed and then normalising it (with a bit of canonicalisation) without filtering out unrecognised language tags feels like a reasonable approach.

Anyway, I think this will work:

```java
public static String getPreferredLanguageFromAcceptLanguageHeader(String acceptLanguageHeader) {
    return parseLanguageRanges(acceptLanguageHeader)
                .stream()
                .map(Locale.LanguageRange::getRange)
                .filter(not(languageRange -> languageRange.equals("*")))
                .map(Locale::forLanguageTag)
                .map(Locale::toLanguageTag)
                .findFirst()
                .orElse(DEFAULT_LOCALE);
}

private static List<Locale.LanguageRange> parseLanguageRanges(String acceptLanguageHeader) {
    if (acceptLanguageHeader == null) {
        return Collections.emptyList();
    }
    try {
        return Locale.LanguageRange.parse(acceptLanguageHeader);
    } catch (IllegalArgumentException e) {
        return Collections.emptyList();
    }
}
```

gidsg:
@alexbishop1 I believe I've addressed all your comments now. Please may you re-review?

src/test/java/uk/gov/pay/connector/util/AcceptLanguageHeaderParserTest.java
```diff
@@ -0,0 +1,20 @@
...
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+
+class AcceptLanguageHeaderParserTest {
+
+    @Test
+    void convertsLocale() {
+        var acceptLanguageHeaderParser = new AcceptLanguageHeaderParser();
+        assertThat(acceptLanguageHeaderParser.getPreferredLanguageFromAcceptLanguageHeader("en-GB,en-US;q=0.9,en;q=0.8"), is("en-GB"));
```
alexbishop1, line 13: Could we maybe add a test where `en-gb` becomes `en-GB`, just to show that this is behaviour we explicitly want?

alexbishop1 (changes requested):
This looks good. Just one extra assertion and we’re done.

src/test/java/uk/gov/pay/connector/util/AcceptLanguageHeaderParserTest.java
gidsg: @alexbishop1 I don't _think_ that's what happens. Rather Java was returning `en-gb` part way through the transformation process. e.g. 
```
jshell> Locale.LanguageRange.parse("en-GB,en-US;q=0.9,en;q=0.8")
$4 ==> [en-gb, en-us;q=0.9, en;q=0.8]
```
The existing tests should show the returned value using the current code is the expected behaviour, so I think that's sufficient?

gidsg: i.e. we are not claiming to support the behaviour you describe, even if that is the current behaviour, we wouldn't expect user-agents to send a lower case value.

alexbishop1: What I meant is that if the `Accept-Language` header is `en-gb` the eventual output will be `en-GB` i.e. we explicitly expect some normalisation to happen.

gidsg: @alexbishop1 thanks, have added and force-pushed

alexbishop1 (approved):

PR #3516, opened 2022-03-11
PP-9332: Add methods to get agreement by ExternalId from DAO and adde…

jbz149:
…d 2 new fields to ChargeEntity

### DIFF
--- a/src/main/java/uk/gov/pay/connector/agreement/dao/AgreementDao.java
+++ b/src/main/java/uk/gov/pay/connector/agreement/dao/AgreementDao.java
@@ -2,6 +2,7 @@ package uk.gov.pay.connector.agreement.dao;
 
 import com.google.inject.Provider;
 import uk.gov.pay.connector.agreement.model.AgreementEntity;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.common.dao.JpaDao;
 
 import javax.inject.Inject;
@@ -19,4 +20,14 @@ public class AgreementDao extends JpaDao<AgreementEntity> {
         return super.findById(AgreementEntity.class, agreementId);
     }
 
+    public Optional<AgreementEntity> findByExternalId(String externalId) {
+
+        String query = "SELECT ae FROM AgreementEntity ae " +
+                "WHERE ae.externalId = :externalId";
+
+        return entityManager.get()
+                .createQuery(query, AgreementEntity.class)
+                .setParameter("externalId", externalId)
+                .getResultList().stream().findFirst();
+    }
 }
--- a/src/main/java/uk/gov/pay/connector/charge/exception/AgreementIdAndSaveInstrumentMandatoryInputException.java
+++ b/src/main/java/uk/gov/pay/connector/charge/exception/AgreementIdAndSaveInstrumentMandatoryInputException.java
@@ -1,0 +1,11 @@
+package uk.gov.pay.connector.charge.exception;
+
+import javax.ws.rs.WebApplicationException;
+
+import static uk.gov.pay.connector.util.ResponseUtil.badRequestResponse;
+
+public class AgreementIdAndSaveInstrumentMandatoryInputException extends WebApplicationException {
+    public AgreementIdAndSaveInstrumentMandatoryInputException(String message) {
+        super(badRequestResponse(message));
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/charge/exception/AgreementIdAndSaveInstrumentMandatoryInputExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/connector/charge/exception/AgreementIdAndSaveInstrumentMandatoryInputExceptionMapper.java
@@ -1,0 +1,27 @@
+package uk.gov.pay.connector.charge.exception;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.common.model.api.ErrorResponse;
+import uk.gov.service.payments.commons.model.ErrorIdentifier;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+public class AgreementIdAndSaveInstrumentMandatoryInputExceptionMapper implements ExceptionMapper<AgreementIdAndSaveInstrumentMandatoryInputException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AgreementIdAndSaveInstrumentMandatoryInputExceptionMapper.class);
+    
+    @Override
+    public Response toResponse(AgreementIdAndSaveInstrumentMandatoryInputException exception) {
+        LOGGER.info(exception.getMessage());
+
+        ErrorResponse errorResponse = new ErrorResponse(ErrorIdentifier.GENERIC, exception.getMessage());
+
+        return Response.status(Response.Status.BAD_REQUEST)
+                .entity(errorResponse)
+                .type(APPLICATION_JSON)
+                .build();
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
@@ -71,8 +71,14 @@ public class ChargeCreateRequest {
     @ValidPaymentProvider
     private String paymentProvider;
 
+    @JsonProperty("agreement_id")
+    @Length(min = 26, max = 26, message = "Field [agreementId] length must be 26")
+    private String agreementId;
+
+    @JsonProperty("save_payment_instrument_to_agreement")
+    private boolean savePaymentInstrumentToAgreement;
+
     public ChargeCreateRequest() {
-        // for Jackson
     }
 
     ChargeCreateRequest(long amount,
@@ -86,7 +92,9 @@ public class ChargeCreateRequest {
                         ExternalMetadata externalMetadata,
                         Source source,
                         boolean moto,
-                        String paymentProvider) {
+                        String paymentProvider,
+                        String agreementId,
+                        boolean savePaymentInstrumentToAgreement) {
         this.amount = amount;
         this.description = description;
         this.reference = reference;
@@ -99,6 +107,8 @@ public class ChargeCreateRequest {
         this.source = source;
         this.moto = moto;
         this.paymentProvider = paymentProvider;
+        this.agreementId = agreementId;
+        this.savePaymentInstrumentToAgreement = savePaymentInstrumentToAgreement;
     }
 
     public long getAmount() {
@@ -145,6 +155,14 @@ public class ChargeCreateRequest {
         return moto;
     }
 
+    public String getAgreementId() {
+        return agreementId;
+    }
+
+    public boolean getSavePaymentInstrumentToAgreement() {
+        return savePaymentInstrumentToAgreement;
+    }
+
     public String getPaymentProvider() {
         return paymentProvider;
     }
@@ -161,6 +179,8 @@ public class ChargeCreateRequest {
                 ", moto=" + moto +
                 (language != null ? ", language=" + language.toString() : "") +
                 (paymentProvider != null ? ", payment_provider=" + paymentProvider : "") +
+                (agreementId != null ? ", agreementId=" + agreementId.toString() : "") +
+                ", savePaymentInstrumentToAgreement=" + savePaymentInstrumentToAgreement +
                 '}';
     }
 }
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -47,6 +47,7 @@ import javax.persistence.OrderBy;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
 import javax.validation.Valid;
+import javax.validation.constraints.Size;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
@@ -180,6 +181,12 @@ public class ChargeEntity extends AbstractVersionedEntity {
     @Column(name = "service_id")
     private String serviceId;
 
+    @Column(name = "agreement_id")
+    private String agreementId;
+    
+    @Column(name = "save_payment_instrument_to_agreement")
+    private boolean savePaymentInstrumentToAgreement;
+
     public ChargeEntity() {
         //for jpa
     }
@@ -203,7 +210,9 @@ public class ChargeEntity extends AbstractVersionedEntity {
             String gatewayTransactionId,
             CardDetailsEntity cardDetails,
             boolean moto,
-            String serviceId
+            String serviceId,
+            String agreementId,
+            boolean savePaymentInstrumentToAgreement
     ) {
         this.amount = amount;
         this.status = status.getValue();
@@ -224,6 +233,8 @@ public class ChargeEntity extends AbstractVersionedEntity {
         this.cardDetails = cardDetails;
         this.moto = moto;
         this.serviceId = serviceId;
+        this.agreementId = agreementId;
+        this.savePaymentInstrumentToAgreement = savePaymentInstrumentToAgreement;
     }
 
     public Long getId() {
@@ -389,6 +400,14 @@ public class ChargeEntity extends AbstractVersionedEntity {
     public void setWalletType(WalletType walletType) {
         this.walletType = walletType;
     }
+    
+    public void setAgreementId(String agreementId) {
+        this.agreementId = agreementId;
+    }
+
+    public void setSavePaymentInstrumentToAgreement(boolean savePaymentInstrumentToAgreement) {
+        this.savePaymentInstrumentToAgreement = savePaymentInstrumentToAgreement;
+    }
 
     public boolean hasExternalStatus(ExternalChargeState... state) {
         return Arrays.stream(state).anyMatch(s -> fromString(getStatus()).toExternal().equals(s));
@@ -504,6 +523,14 @@ public class ChargeEntity extends AbstractVersionedEntity {
         this.serviceId = serviceId;
     }
 
+    public String getAgreementId() {
+        return agreementId;
+    }
+
+    public boolean isSavePaymentInstrumentToAgreement() {
+        return savePaymentInstrumentToAgreement;
+    }
+    
     public static final class WebChargeEntityBuilder {
         private Long amount;
         private String returnUrl;
@@ -519,6 +546,10 @@ public class ChargeEntity extends AbstractVersionedEntity {
         private Source source;
         private boolean moto;
         private String serviceId;
+        private String agreementId;
+        private boolean savePaymentInstrumentToAgreement;
+
+
 
         private WebChargeEntityBuilder() {
         }
@@ -598,6 +629,16 @@ public class ChargeEntity extends AbstractVersionedEntity {
             return this;
         }
 
+        public WebChargeEntityBuilder withAgreementId(String agreementId) {
+            this.agreementId = agreementId;
+            return this;
+        }
+
+        public WebChargeEntityBuilder withSavePaymentInstrumentToAgreement(boolean savePaymentInstrumentToAgreement) {
+            this.savePaymentInstrumentToAgreement = savePaymentInstrumentToAgreement;
+            return this;
+        }
+
         public ChargeEntity build() {
             return new ChargeEntity(
                     amount,
@@ -617,7 +658,9 @@ public class ChargeEntity extends AbstractVersionedEntity {
                     null,
                     null,
                     moto,
-                    serviceId);
+                    serviceId,
+                    agreementId,
+                    savePaymentInstrumentToAgreement);
         }
     }
 
@@ -633,6 +676,8 @@ public class ChargeEntity extends AbstractVersionedEntity {
         private ServicePaymentReference reference;
         private ExternalMetadata externalMetadata;
         private String serviceId;
+        private String agreementId;
+        private boolean savePaymentInstrumentToAgreement;
 
         private TelephoneChargeEntityBuilder() {
         }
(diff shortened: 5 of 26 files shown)


src/main/java/uk/gov/pay/connector/agreement/dao/AgreementDao.java
```diff
@@ -19,4 +20,14 @@ public AgreementDao(Provider<EntityManager> entityManager) {
         return super.findById(AgreementEntity.class, agreementId);
     }
 
+    public Optional<AgreementEntity> findByExternalId(String externalId) {
+
+        String query = "SELECT c FROM AgreementEntity c " +
+                "WHERE c.externalId = :externalId";
```
alexbishop1: `a` would be a better letter than `c` (assume this was copied from something that retrieved charges).

src/main/java/uk/gov/pay/connector/charge/exception/AgreementIdAndSaveInstrumentMandatoryInputException.java
```diff
@@ -0,0 +1,11 @@
+package uk.gov.pay.connector.charge.exception;
+
+import javax.ws.rs.WebApplicationException;
+
+import static uk.gov.pay.connector.util.ResponseUtil.badRequestResponse;
+
+public class AgreementIdAndSaveInstrumentMandatoryInputException extends WebApplicationException {
```
alexbishop1, line 7: I wonder if this name is a bit too specific but I can’t think of a better one right now and it’s easy enough to rename later.

src/main/java/uk/gov/pay/connector/charge/exception/AgreementIdAndSaveInstrumentMandatoryInputExceptionMapper.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class AgreementIdAndSaveInstrumentMandatoryInputExceptionMapper implements ExceptionMapper<AgreementIdAndSaveInstrumentMandatoryInputException> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AgreementIdAndSaveInstrumentMandatoryInputExceptionMapper.class);
+    
+    @Override
+    public Response toResponse(AgreementIdAndSaveInstrumentMandatoryInputException exception) {
+        LOGGER.info(exception.getMessage());
+
+        ErrorResponse errorResponse = new ErrorResponse(ErrorIdentifier.GENERIC, exception.getMessage());
```
alexbishop1, line 20: I think we will want something more specific than `ErrorIdentifier.GENERIC` so that the eventual caller (public API) will be able to interpret it. However, `ErrorIdentifier` is defined elsewhere (in Pay Java commons) so this is fine for now; we can come back to it.

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -180,6 +181,13 @@
     @Column(name = "service_id")
     private String serviceId;
 
+    @Column(name = "agreement_id")
+    @Size(min = 1, max = 26, message = "Field [agreementId] can have a size between 0 and 26")
```
alexbishop1: Don’t think we want the `@Size` annotation on this: this class does not directly represent a JSON request (it can be a JSON _response_).

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -939,4 +953,20 @@ private void checkIfMotoPaymentsAllowed(boolean moto, GatewayAccountEntity gatew
             throw new MotoPaymentNotAllowedForGatewayAccountException(gatewayAccount.getId());
         }
     }
+
+    private void checkAgreementIdAndSaveInstrumentBothPresent(String agreementId, boolean savePaymentInstrumentToAgreement) {
+        if ((StringUtils.isEmpty(agreementId) && savePaymentInstrumentToAgreement) 
+                || (StringUtils.isNotEmpty(agreementId) && !savePaymentInstrumentToAgreement)) {
+            throw new AgreementIdAndSaveInstrumentMandatoryInputException("One of the mandatory parameters 'agreementID' or 'savePaymentInstrumentToAgreement' is missing.");
+        }
+    }
```
alexbishop1: I think we can just check that `agreementId` is null: any empty strings will have been caught by the length validation on the request property. Also the message isn’t quite right: the parameters aren’t themselves mandatory, it’s just that if one have one you also have the other. Finally, our convention for exceptions like this is to have the field name in snake_case (because that matches the JSON) in square brackets.

I’d do something like this, with two different messages:

```java
private void checkAgreementIdAndSaveInstrumentBothPresent(String agreementId, boolean savePaymentInstrumentToAgreement) {
    if (agreementId != null && !savePaymentInstrumentToAgreement) {
        throw new AgreementIdAndSaveInstrumentMandatoryInputException("If [agreement_id] is present, [save_payment_instrument_to_agreement] must be true");
    } else if (savePaymentInstrumentToAgreement && agreementId == null) {
        throw new AgreementIdAndSaveInstrumentMandatoryInputException("If [save_payment_instrument_to_agreement] is true, [agreement_id] must be specified");
    }
}
```

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -939,4 +953,20 @@ private void checkIfMotoPaymentsAllowed(boolean moto, GatewayAccountEntity gatew
...
+
+    private void checkForUnknownAgreementId(String agreementId) {
+        if (StringUtils.isNotEmpty(agreementId)) {
+            Optional<AgreementEntity> byExternalId = agreementDao.findByExternalId(agreementId);
+            if (byExternalId.isEmpty()) {
+                throw new AgreementNotFoundException(format("Agreement with externalId [%s] not found.", agreementId));
+            }
+        }
+    }
```
alexbishop1: Again, no need to consider the empty string and generally we try to avoid calling `isEmpty()` on optionals in favour of `map(…)` and `ifPresent(…)` etc. (more about that at https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#optionals).

This should do the trick:

```java
if (agreementId != null) {
    agreementDao.findByExternalId(agreementId)
            .orElseThrow(() -> new AgreementNotFoundException(format("Agreement with ID [%s] not found.", agreementId)));
}
```

(Message says “ID” rather than “externalId” because it’s not really external to anything outside of connector.)

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentCreatedEventDetails.java
```diff
@@ -207,6 +213,14 @@ public Source getSource() {
         return source;
     }
 
+    public String getAgreementId() {
+        return agreementId;
+    }
+
+    public boolean getSavePaymentInstrumentToAgreement() {
```
alexbishop1: This is a boolean so it should be `isSavePaymentInstrumentToAgreement()`.

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentCreatedEventDetails.java
```diff
@@ -64,6 +66,8 @@ public PaymentCreatedEventDetails(Builder builder) {
         this.addressCountry = builder.addressCountry;
         this.source = builder.source;
         this.moto = builder.moto;
+        this.agreementId = builder.agreementId;
+        this.savePaymentInstrumentToAgreement = builder.savePaymentInstrumentToAgreement;
```
alexbishop1, line 70: I think these fields need adding to `equals(…)` and `hashCode()`.

src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
```diff
@@ -71,8 +71,14 @@
     @ValidPaymentProvider
     private String paymentProvider;
 
+    @JsonProperty("agreement_id")
+    @Length(min = 1, max = 26, message = "Field [agreementId] can have a size between 0 and 26")
```
alexbishop1: It’s actually always exactly 26 characters.

src/test/java/uk/gov/pay/connector/charge/model/ChargeCreateRequestTest.java
```diff
@@ -26,4 +30,14 @@ public void toStringWithoutEmail() {
         final String stringRepresentation = chargeCreateRequest.toStringWithoutPersonalIdentifiableInformation();
         assertThat(stringRepresentation.contains("test@example.com"), is(false));
     }
+    
+    @Test
+    public void agreementIdShouldNotExceed26characters() {
```
alexbishop1: We don’t seem to have validation tests in this class. I think `ChargesApiResourceCreateIT` would be a more appropriate place.

src/test/java/uk/gov/pay/connector/charge/model/domain/ChargeEntityFixture.java
```diff
@@ -64,6 +64,8 @@
     private Exemption3ds exemption3ds;
     private String paymentProvider = "sandbox";
     private String serviceId = randomUuid();
+    private String agreementId = randomUuid().substring(24);
```
alexbishop1: Should probably be 26.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceCreateTest.java
```diff
@@ -298,6 +299,31 @@ public void shouldThrowExceptionWhenCreateChargeWithZeroAmountIfGatewayAccountDo
         verify(mockedChargeDao, never()).persist(any(ChargeEntity.class));
     }
 
+    @Test(expected = AgreementIdAndSaveInstrumentMandatoryInputException.class)
+    public void shouldThrowExceptionWhenAgreementIdNotProvidedAlongWithSavePaymentInstrumentToAgreement() {
+        final ChargeCreateRequest request = requestBuilder.withAmount(1000).withSavePaymentInstrumentToAgreement(Boolean.TRUE).build();
```
alexbishop1: Really minor but you can just do `true` rather than `Boolean.TRUE` here.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServiceCreateTest.java
```diff
@@ -298,6 +299,31 @@ public void shouldThrowExceptionWhenCreateChargeWithZeroAmountIfGatewayAccountDo
...
+
+        service.create(request, GATEWAY_ACCOUNT_ID, mockedUriInfo);
+
+        verify(mockedChargeDao, never()).persist(any(ChargeEntity.class));
+    }
+
+    @Test(expected = AgreementNotFoundException.class)
+    public void shouldThrowExceptionWhenAgreementIdNotPresentInDb() {
+        String UNKNOWN_AGREEMENT_ID = "unknownId";
+        final ChargeCreateRequest request = requestBuilder.withAmount(1000).withAgreementId(UNKNOWN_AGREEMENT_ID).withSavePaymentInstrumentToAgreement(Boolean.TRUE).build();
```
alexbishop1: And here!

src/test/java/uk/gov/pay/connector/events/model/charge/Gateway3dsExemptionResultObtainedTest.java
```diff
@@ -22,7 +21,7 @@ void serializesEventDetailsWithExemption3dsHonoured() throws JsonProcessingExcep
         var eventDate = ZonedDateTime.now(UTC);
         String actual = Gateway3dsExemptionResultObtained.from(chargeEntity, eventDate).toJsonString();
 
-        assertThat(actual, hasJsonPath("$.timestamp", equalTo(MICROSECOND_FORMATTER.format(eventDate))));
+        assertThat(actual, hasJsonPath("$.timestamp", equalTo(eventDate.toString())));
```
alexbishop1: This looks like it’s been reverted unintentionally.

src/test/java/uk/gov/pay/connector/events/model/charge/Gateway3dsExemptionResultObtainedTest.java
```diff
@@ -35,7 +34,7 @@ void serializesEventDetailsWithExemption3dsNull() throws JsonProcessingException
...
```
alexbishop1: Same.

src/test/java/uk/gov/pay/connector/events/model/charge/Gateway3dsInfoObtainedTest.java
```diff
@@ -24,7 +23,7 @@ void serializesEventDetailsWithVersion3ds() throws JsonProcessingException {
         var eventDate = ZonedDateTime.now(UTC);
         String actual = Gateway3dsInfoObtained.from(chargeEntity, eventDate).toJsonString();
 
-        assertThat(actual, hasJsonPath("$.timestamp", equalTo(MICROSECOND_FORMATTER.format(eventDate))));
+        assertThat(actual, hasJsonPath("$.timestamp", equalTo(eventDate.toString())));
```
alexbishop1: Same.

src/test/java/uk/gov/pay/connector/events/model/charge/Gateway3dsInfoObtainedTest.java
```diff
@@ -37,10 +36,10 @@ void serializesEventDetailsWithVersion3dsNull() throws JsonProcessingException {
...
```
alexbishop1: Same.

src/test/java/uk/gov/pay/connector/events/model/charge/PaymentCreatedTest.java
```diff
@@ -57,13 +63,22 @@ void serializesPayloadForCreatedWithoutCardDetails() throws JsonProcessingExcept
         assertThat(paymentCreatedEvent, hasNoJsonPath("$.event_details.email"));
     }
 
+    @Test
+    void serializesPayloadRecurringCardPaymentDetails() throws JsonProcessingException {
```
alexbishop1: I’m slightly concerned about bagging the name `RecurringCardPayment` for this as it only covers the initial payment and later we’ll have subsequent payments. Maybe just call it `serializesPayloadWithAgreementIdAndSavePaymentInstrumentToAgreement()` (long but precise!).

src/test/java/uk/gov/pay/connector/events/model/charge/PaymentCreatedTest.java
```diff
@@ -57,13 +63,22 @@ void serializesPayloadForCreatedWithoutCardDetails() throws JsonProcessingExcept
...
+        var paymentCreatedEvent = prepareRecurringPaymentCreatedEvent();
```
alexbishop1: Looking at the rest of the file, it seems the conventional way to do this would be:

```java
chargeEntityFixture
        .withSavePaymentInstrumentToAgreement(true)
        .withAgreementId("fdsasdfa");

var paymentCreatedEvent = preparePaymentCreatedEvent();
```

… and not have the `prepareRecurringPaymentCreatedEvent()` method at all.

src/test/java/uk/gov/pay/connector/events/model/charge/PaymentCreatedTest.java
```diff
@@ -188,4 +203,10 @@ private void assertBasePaymentCreatedDetails(String actual) {
         assertThat(actual, hasJsonPath("$.resource_type", equalTo("payment")));
         assertThat(actual, hasJsonPath("$.timestamp", equalTo(time)));
     }
+
+    private void assertRecurringPaymentCreatedDetails(String actual) {
+        assertBasePaymentCreatedDetails(actual);
+        assertThat(actual, hasJsonPath("$.event_details.agreement_id", equalTo(chargeEntity.getAgreementId())));
+        assertThat(actual, hasJsonPath("$.event_details.save_payment_instrument_to_agreement", equalTo(chargeEntity.isSavePaymentInstrumentToAgreement())));
+    }
```
alexbishop1, line 212: This method doesn’t seem to ever be used.

src/test/java/uk/gov/pay/connector/events/model/charge/UserEmailCollectedTest.java
```diff
@@ -21,7 +20,7 @@ public void serializesEventDetailsForChargeAndEventDate() throws JsonProcessingE
         ZonedDateTime eventDate = ZonedDateTime.now(UTC);
         String actual = UserEmailCollected.from(chargeEntity, eventDate).toJsonString();
 
-        assertThat(actual, hasJsonPath("$.timestamp", equalTo(MICROSECOND_FORMATTER.format(eventDate))));
+        assertThat(actual, hasJsonPath("$.timestamp", equalTo(eventDate.toString())));
```
alexbishop1: Another unintentional reversion.

src/test/java/uk/gov/pay/connector/it/contract/WorldpayPaymentProviderTest.java
```diff
@@ -80,16 +80,16 @@
 
     private GatewayAccountEntity validGatewayAccount;
     private GatewayAccountEntity validGatewayAccountFor3ds;
-    private Map<String, String> validCredentials;
-    private Map<String, String> validCredentialsFor3ds;
     private GatewayAccountCredentialsEntity validGatewayAccountCredentialsEntity;
-    private GatewayAccountCredentialsEntity validGatewayAccountCredentialsEntityFor3ds;
+    private Map<String, String> validCredentials;
```
alexbishop1: This file has also been reverted. I’m not sure what’s happened here!

src/test/java/uk/gov/pay/connector/it/dao/AgreementDaoIT.java
```diff
@@ -0,0 +1,79 @@
...
+                .build();
+        gatewayAccountCredentialsEntity.setId(defaultTestAccount.getCredentials().get(0).getId());
+    }
+    
+    @Test
+    public void findByExternalId_shouldFindAnAgreementEntity() {
+        insertTestAgreement();
+        Optional<AgreementEntity> agreement = agreementDao.findByExternalId(defaultTestAgreement.getExternalId());
+        assertThat(agreement.isPresent(), is(true));
+    }
```
alexbishop1: Maybe insert two agreements in here and check that the retrieved one has the right ID.

src/test/java/uk/gov/pay/connector/it/dao/AgreementDaoIT.java
```diff
@@ -0,0 +1,79 @@
...
+
+    @Test
+    public void findByExternalId_shouldNotFindAnAgreementEntity() {
+        Optional<AgreementEntity> agreement = agreementDao.findByExternalId("abcdefg123");
+        assertThat(agreement.isPresent(), is(false));
+    }
```
alexbishop1: Might be worth inserting an agreement just to check the code isn’t grabbing any agreement it can find without considering the ID.

alexbishop1 (changes requested):
Looks pretty good!

There’s a few comments but most of them are fairly minor. Main issues are that the agreement ID is always exactly 26 characters (not up to) and some changes made to master since the branch was created seem to have been unintentionally reverted.

src/test/java/uk/gov/pay/connector/charge/model/ChargeCreateRequestTest.java
```diff
@@ -2,8 +2,12 @@
...
 import org.junit.Test;
 
-import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertThat;
+import javax.validation.Validation;
+import javax.validation.Validator;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+import static org.mockito.ArgumentMatchers.eq;
```
alexbishop1: Doubt these import changes are needed now this file is not being changed.

src/test/java/uk/gov/pay/connector/events/model/charge/PaymentCreatedTest.java
```diff
@@ -48,22 +50,42 @@ private String preparePaymentCreatedEvent() throws JsonProcessingException {
         return paymentCreatedEvent.toJsonString();
     }
 
+    private String prepareRecurringPaymentCreatedEvent() throws JsonProcessingException {
+        chargeEntity = chargeEntityFixture.withSavePaymentInstrumentToAgreement(true).withAgreementId("fdsasdfa").build();
+        var paymentCreatedEvent = PaymentCreated.from(chargeEntity);
+        return paymentCreatedEvent.toJsonString();
+    }
+
```
alexbishop1: This method seems to be unused and can be removed.

src/test/java/uk/gov/pay/connector/it/dao/AgreementDaoIT.java
```diff
@@ -0,0 +1,83 @@
...
+        Optional<AgreementEntity> agreement = agreementDao.findByExternalId(AGREEMENT_EXTERNAL_ID_ONE);
+        assertThat(agreement.isPresent(), is(true));
+        assertThat(agreement.get().getExternalId(), is(AGREEMENT_EXTERNAL_ID_ONE));
+    }
+
+    @Test
+    public void findByExternalId_shouldNotFindAnAgreementEntity() {
+        Optional<AgreementEntity> agreement = agreementDao.findByExternalId("abcdefg123");
+        assertThat(agreement.isPresent(), is(false));
+    }
```
alexbishop1: (Second time writing this comment since GitHub threw it away the first time.)

This still tests with an empty DB. Something like this should work:

```java
@Test
public void findByExternalId_shouldNotFindAnAgreementEntity() {
    insertTestAgreement(AGREEMENT_EXTERNAL_ID_ONE);
    Optional<AgreementEntity> agreement = agreementDao.findByExternalId(AGREEMENT_EXTERNAL_ID_TWO);
    assertThat(agreement.isPresent(), is(false));
}
```

alexbishop1 (changes requested):
Looks good. Just a couple of places where it seems we have some unused code and a small tweak to a test.

alexbishop1 (dismissed):
Hurrah! This all looks good.

PR was a little on the large size. Maybe it could have been two or three: perhaps the agreement DAO stuff in one, then the entity changes etc. in another, and maybe the stuff to add the new fields to the payment created event in a final PR.

alexbishop1 (dismissed):
Curse merge conflicts!

alexbishop1 (dismissed):

alexbishop1 (approved):

PR #3520, opened 2022-03-14
PP-9315 Make it.contract.WorldpayPaymentProviderTest pass again

alexbishop1:
Make the `WorldpayPaymentProviderTest` ‘contract’ test pass again by updating it to reflect the fact that each gateway account credentials are now per-charge.

Also make it so that different credentials can be used for non-3DS and 3DS interactions, which seems to have been lost at some point.


gidsg (approved):
LGTM and tested this passes locally when credentials set 👾 👍

PR #3521, opened 2022-03-14
BAU — Make event tests not fail if run on exact millisecond

alexbishop1:
Some of the event tests check the timestamp in the generated JSON.

They do this by comparing the timestamp in the JSON (formatted using the formatter in `MicrosecondPrecisionDateTimeSerializer`, which outputs an instant with 6 decimal places of second precision) to the stringification of the `ZonedDateTime` that represents the actual timestamp.

This usually works because…

a) the `ZonedDateTime` uses UTC
b) the tests are run on computers that only offer the current time to microsecond precision, meaning when the `ZonedDateTime` is stringified it has at most 6 decimal places for the second
c) it’s unlikely that the test will be run when the current time is an exact millisecond (e.g. 1.123000 seconds) because then the stringification of the `ZonedDateTime` will only have 3 decimal places for the second

However, c) does sometimes happen and it causes the tests to fail. Fix it so that this does not happen.

A better fix would to be use a `Clock` or similar to avoid relying on the current time but that would be more involved.


gidsg (approved):
LGTM, good to improve this flaky test 🎏 👍

PR #3525, opened 2022-03-15
PP-9334: add new table for payment instruments and foreign keys to charges and agreements tables

gidsg:
Co-authored-by: Alan Carter <alan.carter@digital.cabinet-office.gov.uk>

## WHAT YOU DID
Add database migration

## How to test

Check code makes sense, migrations should run successfully when app is run locally

### Logging

### Documentation

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1776,5 +1776,80 @@
 
     </changeSet>
 
+    <changeSet id="add_payment_instruments_table" author="">
+
+        <createTable tableName="payment_instruments">
+            <column name="id" type="bigserial" autoIncrement="true">
+                <constraints primaryKey="true" nullable="false"/>
+            </column>
+            <column name="created_date" type="timestamp without timezone"
+                    defaultValueComputed="(now() at time zone 'utc')">
+             <constraints nullable="false"/>
+            </column>
+            <column name="external_id" type="varchar(32)">
+                <constraints nullable="false" unique="true"/>
+            </column>
+            <column name="last_digits_card_number" type="char(4)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="cardholder_name" type="varchar(255)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="expiry_date" type="varchar(6)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="address_line1" type="varchar(255)">
+                <constraints nullable="true"/>
+            </column>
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1776,5 +1776,80 @@
...
+                <constraints nullable="true"/>
+            </column>
+            <column name="status" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="card_brand" type="text">
+                <constraints nullable="true"/>
+            </column>
+            <column name="start_date" type="timestamp without timezone">
+                <constraints nullable="true"/>
```
alexbishop1, line 1826: Will we ever have a payment instrument without a start date?

src/main/resources/migrations.xml
```diff
@@ -1776,5 +1776,80 @@
...
+            <column name="recurring_auth_token" type="jsonb">
+                <constraints nullable="true"/>
+            </column>
+        </createTable>
+
+    </changeSet>
+
+    <changeSet id="add payment_instrument_id to charges" author="">
+        <addColumn tableName="charges">
+            <column name="payment_instrument_id" type="bigserial">
```
alexbishop1: I think this should be `bigint` (although the column referenced by the foreign key constraint auto-increments, _this_ column does not).

src/main/resources/migrations.xml
```diff
@@ -1776,5 +1776,80 @@
...
+                <constraints foreignKeyName="fk__charges_payment_instruments"
+                             referencedTableName="payment_instruments"
+                             referencedColumnNames="id" nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="add payment_instrument_id to agreements" author="">
+        <addColumn tableName="agreements">
+            <column name="payment_instrument_id" type="bigserial">
```
alexbishop1: `bigint`

alexbishop1 (changes requested):

src/main/resources/migrations.xml
gidsg: Could there in practice be any difference between the `start_date` and `created_date` i.e. could the payment instrument ever be added and the start date added later? if not could we remove it, or default it to `now()` like `created_date`?

alexbishop1: Let’s leave it for now; we can always change it later.

alexbishop1 (approved):

PR #3530, opened 2022-03-17
BAU — Make refund email integration test less flaky

alexbishop1:
The refund email integration tests makes an HTTP call to connector’s notification endpoint and then checks it caused an email to be sent. However, we send emails by submitting a task to an `ExecutorService`, which means the email might not have been sent by the time the HTTP call completes (race condition). Fix this in a shameful way (but not unprecedented for our integration tests) by adding a short delay before checking the email is sent to give the task time to do its thing.


oswaldquek (approved):
🎉 Thanks for fixing this.

PR #3534, opened 2022-03-17
PP-9334: Add entity for Payment Instrument

gidsg:
Co-authored-by: Alan Carter <[alan.carter@digital.cabinet-office.gov.uk](mailto:alan.carter@digital.cabinet-office.gov.uk)>

## WHAT YOU DID
Add entity for Payment Instrument (as added to the DB in https://github.com/alphagov/pay-connector/pull/3525)

## How to test

As this is just the entity not sure if it is usefully tested on its own?

## Code review checklist

Code should match database migration

### Logging

### Documentation

- [ ] Updated [README.md](http://readme.md/) for any of the following ?

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentinstrument/model/PaymentInstrumentEntity.java
+++ b/src/main/java/uk/gov/pay/connector/paymentinstrument/model/PaymentInstrumentEntity.java
@@ -1,0 +1,170 @@
+package uk.gov.pay.connector.paymentinstrument.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import uk.gov.pay.connector.agreement.model.AgreementEntity;
+import uk.gov.pay.connector.charge.model.CardDetailsEntity;
+import uk.gov.pay.connector.common.model.api.ToLowerCaseStringSerializer;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/paymentinstrument/model/PaymentInstrumentEntity.java
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    public void setCardDetails(CardDetailsEntity cardDetails) {
+        this.cardDetails = cardDetails;
+    }
+
+    @Column(name = "status")
+    @Enumerated(EnumType.STRING)
+    @JsonProperty("status")
+    @JsonSerialize(using = ToLowerCaseStringSerializer.class)
```
alexbishop1, line 116: Is lower-case consistent with what we do elsewhere?

gidsg: @alexbishop1 Took this style from https://github.com/alphagov/pay-connector/blob/8b971c830fc607443026244ae4e4db31d0b8efbb/src/main/java/uk/gov/pay/connector/charge/model/CardDetailsEntity.java#L57-L61 so it's at least consistent with what we do in _some_ instances!

alexbishop1: Good enough for me…

alexbishop1 (approved):
Agree there’s not anything of substance to test yet (and no risk from not having tests because this isn’t doing anything yet).

PR #3541, opened 2022-03-21
BAU: Delete PaaS related files

gidsg:
## WHAT YOU DID
Delete unused PaaS deployment related files

It is likely these have now bitrotted and they are confusing for someone trying to decipher the current infra

## How to test
Things should still work


env.sh
```diff
@@ -1,10 +0,0 @@
-#!/bin/bash
-ENV_FILE="$WORKSPACE/pay-scripts/services/connector.env"
-if [ -f $ENV_FILE ]
-then
-  set -a
-  source $ENV_FILE
-  set +a  
-fi
-
-eval "$@"
```
katstevens: This file I'm not sure about deleting - is this definitely PaaS related?

gidsg: @katstevens Hmm, not sure, git history is vague but I can't find anything that invokes it so I presumed it to be some sort of orphan. That said if we're worried I am happy to restore it.

katstevens: My theory is that it's used by another repo for something (eg as part of some mysterious docker-compose file). The file goes a long way back, before any inkling of PaaS.

gidsg: @katstevens OK I suspect it is still redundant but happy to restore it, as it is clearly not PaaS related.

katstevens (approved):

PR #3549, opened 2022-03-23
PP-9331 modify payments endpoint to support setting up payment instruments

jbz149:
- Added agreement_id and save_payment_instrument_to_agreement to Resource, Service, ChargeEntity and ChargeResponse.
- Wrote the unit and integration tests
- Disable an old test due to a timezone change issue (Timestamp is 1 hour late the 21st of March 2022)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
@@ -134,6 +134,9 @@ public class ChargeResponse {
     @JsonProperty("moto")
     private boolean moto;
 
+    @JsonProperty("agreement_id")
+    private String agreementId;
+
     ChargeResponse(AbstractChargeResponseBuilder<?, ? extends ChargeResponse> builder) {
         this.dataLinks = builder.getLinks();
         this.chargeId = builder.getChargeId();
@@ -167,6 +170,7 @@ public class ChargeResponse {
         this.walletType = builder.getWalletType();
         this.externalMetadata = builder.getExternalMetadata();
         this.moto = builder.isMoto();
+        this.agreementId = builder.getAgreementId();
     }
 
     public List<Map<String, Object>> getDataLinks() {
@@ -301,6 +305,10 @@ public class ChargeResponse {
         return moto;
     }
 
+    public String getAgreementId() {
+        return agreementId;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
@@ -373,6 +381,7 @@ public class ChargeResponse {
                 ", totalAmount=" + totalAmount +
                 ", walletType=" + walletType +
                 ", moto=" + moto +
+                ", agreementId=" + agreementId +
                 '}';
     }
 
--- a/src/main/java/uk/gov/pay/connector/charge/model/builder/AbstractChargeResponseBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/builder/AbstractChargeResponseBuilder.java
@@ -49,7 +49,8 @@ public abstract class AbstractChargeResponseBuilder<T extends AbstractChargeResp
     protected WalletType walletType;
     protected ExternalMetadata externalMetadata;
     protected boolean moto;
-
+    private String agreementId;
+    
     protected abstract T thisObject();
 
     public T withChargeId(String chargeId) {
@@ -222,7 +223,12 @@ public abstract class AbstractChargeResponseBuilder<T extends AbstractChargeResp
         this.moto = moto;
         return thisObject();
     }
-
+    
+    public T withAgreementId(String agreementId) {
+        this.agreementId = agreementId;
+        return thisObject();
+    }
+    
     public String getChargeId() {
         return chargeId;
     }
@@ -351,5 +357,9 @@ public abstract class AbstractChargeResponseBuilder<T extends AbstractChargeResp
 
     public boolean isMoto() {
         return moto;
+    }
+
+    public String getAgreementId() {
+        return agreementId;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
@@ -49,7 +49,6 @@ import javax.persistence.OrderBy;
 import javax.persistence.SequenceGenerator;
 import javax.persistence.Table;
 import javax.validation.Valid;
-import javax.validation.constraints.Size;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -500,7 +500,8 @@ public class ChargeService {
                 .withLink("self", GET, selfUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeId))
                 .withLink("refunds", GET, refundsUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeEntity.getExternalId()))
                 .withWalletType(chargeEntity.getWalletType())
-                .withMoto(chargeEntity.isMoto());
+                .withMoto(chargeEntity.isMoto())
+                .withAgreementId(chargeEntity.getAgreementId());
 
         chargeEntity.getFeeAmount().ifPresent(builderOfResponse::withFee);
         chargeEntity.getExternalMetadata().ifPresent(builderOfResponse::withExternalMetadata);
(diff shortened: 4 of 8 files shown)


src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
```diff
@@ -64,6 +64,11 @@ public static Response notFoundResponse(String message) {
         return buildErrorResponse(NOT_FOUND, message);
     }
 
+    public static Response agreementNotFoundResponse(String message) {
```
alexbishop1: I wonder if having an overloaded `notFoundResponse(ErrorIdenifier, String)` method (calling an overloaded `buildErrorResponse(ErrorIdentifier, String, String)` method) might work here. But this is fine for now.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateIT.java
```diff
@@ -1197,7 +1198,40 @@ public void shouldReturn201WhenAgreementIdLengthIsEqualsTo26Characters() {
...
                 .postCreateCharge(postBody, accountId)
                 .statusCode(HttpStatus.SC_CREATED)
-                .contentType(JSON);
+                .contentType(JSON)
+                .body("agreement_id", is(JSON_VALID_AGREEMENT_ID_VALUE))
+                .body("save_payment_instrument_to_agreement", is(true));
     }
 
+    @Test
+    public void shouldReturn400AndErrorIdAgreementNotFoundWhenAgreementIsNotFound() {
```
alexbishop1: 404 (not 400)

alexbishop1 (changes requested):
Looks great apart from one tiny test naming thing.

The description of this PR is a bit generic: something closer to the commit message would be great.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateIT.java
jbz149: We have this from the acceptance criteria in PP-9331 :

If agreement ID does not exist, return this with a 400(?):

{
  "field": "set_up_agreement",
  "code": "P0103",
  "description": "Invalid attribute value: set_up_agreement. Agreement ID does not exist"
}

alexbishop1: Hmm, would have thought we’d have `.statusCode(HttpStatus.SC_BAD_REQUEST)` then.

src/main/java/uk/gov/pay/connector/charge/model/ChargeResponse.java
```diff
@@ -134,6 +134,12 @@
     @JsonProperty("moto")
     private boolean moto;
 
+    @JsonProperty("agreement_id")
+    private String agreementId;
+
+    @JsonProperty("save_payment_instrument_to_agreement")
+    private boolean savePaymentInstrumentToAgreement;
```
sfount: Serialising this will expose the `save_payment_instrument_to_agreement` property to the charge response returned by the API to external calls. 

As far as I'm aware this is an internal flag that will be used during the payment process, it's also not currently documented anywhere so would likely cause confusion as it would start showing up on all create payment requests. Is there part of the spec that called for this in particular or can this be removed?

sfount (commented):
Code change looks good, question about modifying the public response

src/main/java/uk/gov/pay/connector/charge/model/builder/AbstractChargeResponseBuilder.java
```diff
@@ -222,6 +225,16 @@ public T withMoto(boolean moto) {
...
+    
+    public T withAgreementId(String agreementId) {
+        this.agreementId = agreementId;
+        return thisObject();
+    }
+    
+    public T withSavePaymentInstrumentToAgreement(boolean savePaymentInstrumentToAgreement) {
+        this.savePaymentInstrumentToAgreement = savePaymentInstrumentToAgreement;
+        return thisObject();
+    }
```
alexbishop1: I think this can go now we’re not including the property in `ChargeResponse`.

src/main/java/uk/gov/pay/connector/charge/model/builder/AbstractChargeResponseBuilder.java
```diff
@@ -352,4 +365,12 @@ public String getAuthCode() {
...
         return moto;
     }
+
+    public String getAgreementId() {
+        return agreementId;
+    }
+
+    public boolean isSavePaymentInstrumentToAgreement() {
+        return savePaymentInstrumentToAgreement;
+    }
```
alexbishop1: I think this can go now we’re not including the property in `ChargeResponse`.

src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeEntity.java
```diff
@@ -664,7 +667,7 @@ public ChargeEntity build() {
                     gatewayAccountCredentialsEntity,
                     paymentProvider,
                     email,
-                    Instant.now(),
+                    Instant.now(Clock.system(ZoneOffset.UTC)),
```
alexbishop1: I don’t think this change actually makes any difference (because `Instant` has no time zone) so will return the same time from any current-time `Clock`).

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -500,7 +500,9 @@ private CardDetailsEntity createCardDetailsEntity(PrefilledCardHolderDetails pre
                 .withLink("self", GET, selfUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeId))
                 .withLink("refunds", GET, refundsUriFor(uriInfo, chargeEntity.getGatewayAccount().getId(), chargeEntity.getExternalId()))
                 .withWalletType(chargeEntity.getWalletType())
-                .withMoto(chargeEntity.isMoto());
+                .withMoto(chargeEntity.isMoto())
+                .withAgreementId(chargeEntity.getAgreementId())
+                .withSavePaymentInstrumentToAgreement(chargeEntity.isSavePaymentInstrumentToAgreement());
```
alexbishop1: I think this can go now we’re not including the property in `ChargeResponse`.

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
alexbishop1: I think this method is not actually used.

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
```diff
@@ -86,6 +91,15 @@ private static Response buildErrorResponse(Status status, String message) {
         ErrorResponse errorResponse = new ErrorResponse(ErrorIdentifier.GENERIC, List.of(message));
         return responseWithEntity(status, errorResponse);
     }
+    private static Response buildAgreementNotFoundErrorResponse(Status status, String message) {
```
alexbishop1: I think this method is not actually used.

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
```diff
@@ -86,6 +91,15 @@ private static Response buildErrorResponse(Status status, String message) {
...
+        ErrorResponse errorResponse = new ErrorResponse(ErrorIdentifier.AGREEMENT_NOT_FOUND, List.of(message));
+        return responseWithEntity(status, errorResponse);
+    }
+
+    public static Response buildErrorResponse(Status status, ErrorIdentifier errorIdentifier, String message) {
+        ErrorResponse errorResponse = new ErrorResponse(errorIdentifier, List.of(message));
+        return responseWithEntity(status, errorResponse);
+    }
```
alexbishop1: To match the conventions in this file, I think we want a public/private method pair like this:

```java
public static Response notFoundResponse(ErrorIdentifier errorIdentifier, String message) {
    LOGGER.info(message);
    return buildErrorResponse(NOT_FOUND, errorIdentifier, message);
}

private static Response buildErrorResponse(Status status, ErrorIdentifier errorIdentifier, String message) {
    ErrorResponse errorResponse = new ErrorResponse(errorIdentifier, List.of(message));
    return responseWithEntity(status, errorResponse);
}
```

alexbishop1 (changes requested):
Looks fine but I think we can delete some more unused code and the method organisation in `ResponseUtil` needs some tweaks.

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
jbz149: Strangely that method is used.
The exception mapper was never called

alexbishop1 (approved):

alexbishop1 (changes requested):
Actually, please squash the two commits into one.

alexbishop1 (approved):

PR #3582, opened 2022-04-06
BAU: add missing columns to payment instruments table

gidsg:
## WHAT YOU DID
I noticed in manual testing the following exception when trying to make a payment with the save payment instrument flag set to true
```
Internal Exception: org.postgresql.util.PSQLException: ERROR: column "card_type" of relation "payment_instruments" does not exist
```
This adds the migration for the  missing columns.
 
## How to test

Change should make sense. If you want to test end-to-end you can manually set the flag on a test payment e.g.
`update charges set save_payment_instrument_to_agreement=true where external_id='some-external-id';`

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1858,4 +1858,22 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add card type, address_state_province, first_digits_card_number columns to payment_instruments table" author="">
+        <addColumn tableName="payment_instruments">
+            <column name="card_type" type="varchar(20)">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+        <addColumn tableName="payment_instruments">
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1858,4 +1858,12 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add card type column to payment_instruments table" author="">
+        <addColumn tableName="payment_instruments">
+            <column name="card_type" type="varchar(20)">
+                <constraints nullable="true"/>
```
alexbishop1: Not sure about this being nullable. The equivalent column on `charges` can be because a charge does not initially have payment details but I think we should always know the card type when we persist a new payment instrument.

gidsg: OK, I have changed this to `false`.

alexbishop1 (approved):

PR #3601, opened 2022-04-13
PP-9495 Add description and user_identifier to agreements table

sfount:
Adds `description` and `user_reference` columns to the existing
agreements
table.

Note the `description` column will eventually have the constraint
`nullable=false` however this will come in a follow-up migration once
the existing resources and DAOs correctly validate and guarantee this
data is available.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1884,4 +1884,14 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add description, user_identifier to agreements table" author="">
+        <addColumn tableName="agreements">
+            <column name="description" type="varchar(255)">
+                <constraints nullable="true" unique="false"/>
+            </column>
+            <column name="user_identifier" type="varchar(255)">
+                <constraints nullable="true" unique="false"/>
+            </column>
+        </addColumn>
+    </changeSet>
 </databaseChangeLog>


src/main/resources/migrations.xml
```diff
@@ -1884,4 +1884,14 @@
         </addColumn>
     </changeSet>
 
-</databaseChangeLog>
+    <changeSet id="add description, user_reference to agreements table" author="">
+        <addColumn tableName="agreements">
+            <column name="description" type="varchar(255)">
+                <constraints nullable="true" unique="false"/>
+            </column>
+            <column name="user_reference" type="varchar(255)">
```
alexbishop1: Why `user_reference` and not just `reference`?

sfount: `reference` already exists and is the service provided reference for this particular agreement between the service and the paying user.

`user_reference` is there for the service to provide what they refer to the individual paying user as, this could be used
* to group multiple agreements for one paying user together
* was also a core component for the "Stored cards" functionality previously investigated

Neither of those will be realised right away but given the friction of adding fields like this I think the logic is that this could be removed if things are restructured and its use was never realised.

It's also worth noting that `user_reference` was favoured over `user_id` as IDs are generally thing the API feeds back to users and are generated by the platform, references generally denote something the service owns and associates with platform entities.

sfount: Discussed with @alexbishop1 where `user_identifier` was selected as the least confusing internal/ external terminology.

alexbishop1 (approved):

PR #3602, opened 2022-04-13
PP-9495 Agreements includes description and user_identifier in create/ response

sfount:
Reflect `description` and `user_identifier` columns on the `AgreementEntity`.

Fields also included in agreement create/ response pattern.

Ensure these values are appropriately validated, passed along and stored with the agreements service/ resource.

PR #3603, opened 2022-04-13
PP-9495 Require agreements description column

sfount:
Depends on https://github.com/alphagov/pay-connector/pull/3602

The `description` column is required for agreements, now that entitles
and tests populate this field, update the column to required.


alexbishop1 (dismissed):
Since we have no (or very few) agreements this should be quick to execute.

alexbishop1 (approved):

PR #3604, opened 2022-04-13
PP-9142 Change default value for authorisation_mode

stephencdaly:
To be consistent with other enum values, and to make serialization
and deserialization of the enum easier, the `authorisation_mode` should
be stored in the databased capitalized.

Drop the column and re-add with the correct default of "WEB". This is
possible as there are no charges currently in the database that have
been created with a different value.

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -1894,4 +1894,16 @@
             </column>
         </addColumn>
     </changeSet>
+
+    <changeSet id="drop column authorisation_mode from charges table" author="">
+        <dropColumn tableName="charges" columnName="authorisation_mode"/>
+    </changeSet>
+
+    <changeSet id="add authorisation_mode to charges table with default value" author="">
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations.xml
```diff
@@ -1894,4 +1894,16 @@
             </column>
         </addColumn>
     </changeSet>
+
+    <changeSet id="drop column authorisation_mode from charges table" author="">
+        <dropColumn tableName="charges" columnName="authorisation_mode"/>
+    </changeSet>
+
+    <changeSet id="r-add authorisation_mode to charges table with correct default value" author="">
```
kbottla: We can just add it as usual. After db migrations, we can remove change sets (the first one adding column and the drop) not needed from migrations.xml.

```suggestion
    <changeSet id="add authorisation_mode to charges table with default value" author="">
```

kbottla (approved):

PR #3608, opened 2022-04-14
PP-9142 Accept authorisation_mode when creating charge

stephencdaly:
Accept the field `authorisation_mode` in the POST request body to create
a charge. This can be any values for the AuthorisationMode enum in
pay-java-commons - `web`, `moto_api` or `external`. Save this on the
charge, and return in charge responses.

We only ever want the `external` authorisation_mode to be set internally
for payments created by the telephone payment notification API.
Validation to prevent this mode being used when creating a standard
payment through the API will be added in public API.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import org.hibernate.validator.constraints.Length;
 import uk.gov.pay.connector.charge.validation.ValidPaymentProvider;
 import uk.gov.service.payments.commons.api.json.ExternalMetadataDeserialiser;
+import uk.gov.service.payments.commons.model.AuthorisationMode;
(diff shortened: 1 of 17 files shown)


src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
```diff
@@ -77,6 +78,10 @@ public class ChargeCreateRequest {
 
     @JsonProperty("save_payment_instrument_to_agreement")
     private boolean savePaymentInstrumentToAgreement;
+    
+    @JsonProperty("authorisation_mode")
+    @Valid
```
kbottla, line 83: I am assuming validation to allow `moto_api` only when the feature is enabled for gateway_accounts will be in separate PR ?

kbottla (approved):

src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
stephencdaly: Yeh that's a different story. For now, the property is accepted and stored but we do nothing with it.

PR #3621, opened 2022-04-20
PP-9144 Replace 'annotated' validation methods with a single validation method

kbottla:
## WHAT YOU DID
- Use a separate validation method so error responses with `INVALID_ATTRIBUTE_VALUE` identifier can be returned to publicapi.
  Validation exceptions (ValidationExceptionMapper) thrown by using @ValidaitonMethod are currently mapped to GENERIC error identifier and changing this would impact other endpoints.


SandorArpa (approved):
LGTM

PR #3622, opened 2022-04-20
PP-9368 Extract code for marking charges as eligible for capture

alexbishop1:
Extract the code that marks a charge as being eligible for capture into two separate classes: one that’s used for all charges and one that’s used only for capturing delayed capture charges. The code for this was previously split between `CardCaptureService` and `ChargeService`.

There are absolutely no changes to funcionality of the code.

The changes to the tests are more substantial. Previously the unit tests were ostensibly testing CardCaptureService but that `CardCaptureService` used a real `ChargeService`, so some of the code that was being exercised was actually in `ChargeService` (and some relating to state transitions was actually in `ChargeEntity`!). The new tests only exercise the specific code in the new extracted classes and stub out `ChargeService`, with the assumption that any methods being called on `ChargeService` and `ChargeEntity` are already well-tested elsewhere.

This change is being made because we’re about to make marking a charge as being eligible for capture more complicated when an initial payment is made for an agreement and we want it not to be a nightmare to implement and test.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureService.java
@@ -1,0 +1,78 @@
+package uk.gov.pay.connector.charge.service;
+
+import com.google.inject.persist.Transactional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.dao.ChargeDao;
+import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.common.exception.IllegalStateRuntimeException;
+import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
+import uk.gov.pay.connector.queue.capture.CaptureQueue;
+import uk.gov.pay.connector.usernotification.service.UserNotificationService;
+import uk.gov.service.payments.commons.queue.exception.QueueException;
+
+import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AWAITING_CAPTURE_REQUEST;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURE_APPROVED;
+
+public class ChargeEligibleForCaptureService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ChargeEligibleForCaptureService.class);
+
+    private final ChargeService chargeService;
+    private final ChargeDao chargeDao;
+    private final CaptureQueue captureQueue;
+    private final UserNotificationService userNotificationService;
+
+    @Inject
+    public ChargeEligibleForCaptureService(ChargeService chargeService, ChargeDao chargeDao, CaptureQueue captureQueue,
+                                           UserNotificationService userNotificationService) {
+        this.chargeService = chargeService;
+        this.chargeDao = chargeDao;
+        this.captureQueue = captureQueue;
+        this.userNotificationService = userNotificationService;
+    }
+
+    public ChargeEntity markChargeAsEligibleForCapture(String externalId) {
+        ChargeEntity charge = updateToNewStatus(externalId);
+
+        if (!charge.isDelayedCapture()) {
+            addChargeToCaptureQueue(charge);
+            userNotificationService.sendPaymentConfirmedEmail(charge, charge.getGatewayAccount());
+        }
+
+        return charge;
+    }
+
+    @Transactional
+    public ChargeEntity updateToNewStatus(String externalId) {
+        return chargeDao.findByExternalId(externalId).map(charge -> {
+            ChargeStatus targetStatus = charge.isDelayedCapture() ? AWAITING_CAPTURE_REQUEST : CAPTURE_APPROVED;
+
+            try {
+                chargeService.transitionChargeState(charge, targetStatus);
+            } catch (InvalidStateTransitionException e) {
+                throw new IllegalStateRuntimeException(charge.getExternalId());
+            }
+
+            return charge;
+        }).orElseThrow(() -> new ChargeNotFoundRuntimeException(externalId));
+    }
+
+    private void addChargeToCaptureQueue(ChargeEntity charge) {
+        try {
+            captureQueue.sendForCapture(charge);
+        } catch (QueueException e) {
+            LOGGER.error("Exception sending charge [{}] to capture queue", charge.getExternalId());
+            throw new WebApplicationException(format(
+                    "Unable to schedule charge [%s] for capture - %s",
+                    charge.getExternalId(), e.getMessage()));
+        }
+    }
+
+}
--- a/src/main/java/uk/gov/pay/connector/charge/service/DelayedCaptureService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/DelayedCaptureService.java
@@ -1,0 +1,84 @@
+package uk.gov.pay.connector.charge.service;
+
+import com.google.inject.persist.Transactional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.dao.ChargeDao;
+import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
+import uk.gov.pay.connector.common.exception.InvalidStateTransitionException;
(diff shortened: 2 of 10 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureService.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Transactional
+    public ChargeEntity updateToNewStatus(String externalId) {
+        return chargeDao.findByExternalId(externalId).map(charge -> {
+            ChargeStatus targetStatus = charge.isDelayedCapture() ? AWAITING_CAPTURE_REQUEST : CAPTURE_APPROVED;
+
+            try {
+                chargeService.transitionChargeState(charge, targetStatus);
+            } catch (InvalidStateTransitionException e) {
+                throw new IllegalStateRuntimeException(charge.getExternalId());
```
kundanscorpio, line 60: It is better to log error with some meaningful message before we throw exception

src/main/java/uk/gov/pay/connector/charge/service/DelayedCaptureService.java
```diff
@@ -0,0 +1,84 @@
...
+
+                case CAPTURE_APPROVED:
+                case CAPTURE_APPROVED_RETRY:
+                case CAPTURE_READY:
+                case CAPTURE_SUBMITTED:
+                case CAPTURED:
+                    return charge;
+
+                default:
+                    throw new ConflictRuntimeException(charge.getExternalId(),
```
kundanscorpio, line 65: Same as previous: It is good to log error before we throw exception

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -4,6 +4,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.charge.service.ChargeCancelService;
+import uk.gov.pay.connector.charge.service.ChargeEligibleForCaptureService;
+import uk.gov.pay.connector.charge.service.DelayedCaptureService;
 import uk.gov.pay.connector.gateway.model.Auth3dsResult;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 import uk.gov.pay.connector.gateway.model.GatewayError;
```
kundanscorpio, line 11: There is an unused import for CardCaptureService in this class

src/test/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureServiceTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+        assertThrows(WebApplicationException.class,
+                () -> chargeEligibleForCaptureService.markChargeAsEligibleForCapture(chargeEntity.getExternalId()));
+
+        verify(mockChargeService).transitionChargeState(chargeEntity, CAPTURE_APPROVED);
+        verifyNoInteractions(mockUserNotificationService);
+    }
+
+    @Test
+    void shouldThrowExceptionIfChargeNotFound() throws QueueException {
```
kundanscorpio, line 112: This test case does not throw QueueException.  So we can remove it

src/test/java/uk/gov/pay/connector/charge/service/DelayedCaptureServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+    @Test
+    public void shouldThrowExceptionIfChargeIsNotHasIncorrectStatus() {
+        ChargeEntity chargeEntity = ChargeEntityFixture.aValidChargeEntity().withStatus(AUTHORISATION_SUCCESS).build();
+        when(mockChargeDao.findByExternalId(chargeEntity.getExternalId())).thenReturn(Optional.of(chargeEntity));
+
+        var conflictRuntimeException = assertThrows(ConflictRuntimeException.class,
+                () -> delayedCaptureService.markDelayedCaptureChargeAsCaptureApproved(chargeEntity.getExternalId()));
+
+        assertThat(conflictRuntimeException.getMessage(), containsString("HTTP 409 Conflict"));
```
kundanscorpio: It will be good if we can extract this "HTTP 409 Conflict" in a constant since it is used more than once

src/test/java/uk/gov/pay/connector/charge/service/DelayedCaptureServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+    public void shouldThrowExceptionIfChargeCannotBeTransitioned() {
+        ChargeEntity chargeEntity = ChargeEntityFixture.aValidChargeEntity().withStatus(AWAITING_CAPTURE_REQUEST).build();
+        when(mockChargeDao.findByExternalId(chargeEntity.getExternalId())).thenReturn(Optional.of(chargeEntity));
+        doThrow(InvalidStateTransitionException.class).when(mockChargeService).transitionChargeState(chargeEntity, CAPTURE_APPROVED);
+
+
+        var conflictRuntimeException = assertThrows(ConflictRuntimeException.class,
+                () -> delayedCaptureService.markDelayedCaptureChargeAsCaptureApproved(chargeEntity.getExternalId()));
+
+        assertThat(conflictRuntimeException.getMessage(), containsString("HTTP 409 Conflict"));
```
kundanscorpio: It will be good if we can extract this "HTTP 409 Conflict" in a constant since it is used more than once

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceTest.java
```diff
@@ -14,6 +14,8 @@
 import uk.gov.pay.connector.charge.exception.OneTimeTokenUsageInvalidForMotoApiException;
 import uk.gov.pay.connector.charge.exception.OneTimeTokenUsageInvalidForMotoApiExceptionMapper;
 import uk.gov.pay.connector.charge.service.ChargeCancelService;
+import uk.gov.pay.connector.charge.service.ChargeEligibleForCaptureService;
+import uk.gov.pay.connector.charge.service.DelayedCaptureService;
```
kundanscorpio, line 18: Line 23 import uk.gov.pay.connector.paymentprocessor.service.CardCaptureService is an unused import in this class

kundanscorpio (changes requested):
This does not seem to have the functionality of associating payment_instrument_id with agreement. Assumption those changes will be implemented as a part of separate commit. There are few minor comments.

alexbishop1:
> This does not seem to have the functionality of associating payment_instrument_id with agreement. Assumption those changes will be implemented as a part of separate commit.

Correct, this is just moving some code around. I wanted to keep it separate from the functionality change to make it easy to review.

src/main/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureService.java
alexbishop1: This bit of code is just moved from elsewhere; I tried to keep the behaviour exactly the same (does not preclude improving it later, of course). In this case, the `InvalidStateTransitionException` is thrown by `setStatus(…)` method in `ChargeEntity` and that does already log a message (though at warn).

src/main/java/uk/gov/pay/connector/charge/service/DelayedCaptureService.java
alexbishop1: As before: code moved from elsewhere.

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
alexbishop1: Whoops, thought I ran optimise imports. Will remove it.

src/test/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureServiceTest.java
alexbishop1: Thanks for spotting this, will remove.

src/test/java/uk/gov/pay/connector/charge/service/DelayedCaptureServiceTest.java
alexbishop1: Yep, I’ll do that.

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceTest.java
alexbishop1: Consider it gone!

src/test/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureServiceTest.java
alexbishop1: Actually, looks like it is needed: the `inOrder.verify(mockCaptureQueue).sendForCapture(result);` statement gets angsty if it isn’t there.

alexbishop1:
Just amended the commit to do make the following changes:
- remove the two unused imports
- extract “HTTP 409 Conflict” into a constant

I pinky-swear there’s no other changes, so you don’t have to look through all 600 lines again!

kundanscorpio (approved):

PR #3633, opened 2022-04-22
PP-9368 Link payment instrument to agreement on user confirm

alexbishop1:
When a user confirms a payment, marking it eligible for capture (`CAPTURE APPROVED` for most payments, `AWAITING CAPTURE REQUEST` for delayed capture payments), link the payment instrument associated with the payment to the agreement associated with the payment and set the payment instrument to active if and only if the payment has the flag to save the payment instrument to the agreement enabled.

A `@Transactional` annotation makes sure that this linking and status updating happens in the same transaction as the one that updates the status of the payment itself, ensuring consistency.

The code accounts for three errors that can Never Happen™ because they would mean the world has gotten into an inconsistent state:

1. The payment has no payment instrument on it (in which case the flag should not be set)
1. The payment has no agreement ID set on it (in which case the flag should not be set)
1. The payment has an agreement ID but there is no matching agreement in the database (in which case our foreign key constraints may have failed us)

If any of these errors occur, we log an error but continue marking the payment as eligible for capture. This means that this payment will succeed but the service will not be able to take future recurring payments from the agreement.

Other errors (for example, the database connection failing), will cause the entire process of marking the payment eligible for capture to fail, which is consistent with other parts of the system.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementServiceTest.java
@@ -1,0 +1,123 @@
+package uk.gov.pay.connector.charge.service;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.LoggingEvent;
+import ch.qos.logback.core.Appender;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.agreement.dao.AgreementDao;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementServiceTest.java
```diff
@@ -0,0 +1,123 @@
...
+        assertThat(loggingEvent.getLevel(), is(Level.ERROR));
+        assertThat(loggingEvent.getFormattedMessage(),
+                is("Expected charge " + chargeEntity.getExternalId() + " to have an agreement but it does not have one"));
+
+        verifyNoInteractions(mockAgreementEntity);
+        verifyNoInteractions(mockPaymentInstrumentEntity);
+    }
+
+    @Test
+    void logsErrorIfChargeDoesHasPaymentInstrumentAndAgreementIdButAgreementDaoIsLikeNewPhoneWhoDis() {
```
kundanscorpio: name of the test case is not self explanatory.

kundanscorpio (changes requested):
Changes look good. Verified end to end functionality also works fine.. One minor comment added.

src/test/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementServiceTest.java
alexbishop1: I was really hoping I could sneak my whimsical test name past!

alexbishop1:
Some test cases renamed.

kundanscorpio (approved):

PR #3637, opened 2022-04-25
PP-9143 Get card information from Cardid

stephencdaly:
If a card number passes the Luhn check, attempt to look up the card
information from the BIN ranges in Cardid.

If no range containing the card number is found, return an error
response with status 402 and error_identifier CARD_NUMBER_REJECTED.


SandorArpa (approved):
LGTM

PR #3642, opened 2022-04-26
PP-9145 Mark token as used and new state transitions

kbottla:
## WHAT YOU DID
- Marked token as user for authorise api
- Added new state transitions for payments authorised by MOTO api


src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -177,6 +177,9 @@ public Response authorise(@Valid @NotNull AuthoriseRequest authoriseRequest) {
         authoriseRequest.validate();
         ChargeEntity charge = tokenService.validateTokenAndGetChargeForMotoApi(authoriseRequest.getOneTimeToken());
         motoApiCardNumberValidationService.validateCardNumber(charge, authoriseRequest.getCardNumber());
+
+        tokenService.markTokenAsUsed(authoriseRequest.getOneTimeToken());
```
stephencdaly: I think we need to mark as used before validating the card number, as this throws an exception if validation failed, and so the token will not be marked as used. As we transition the state to AUTHORISATION_REJECTED if the card number is invalid, we want to be sure we always mark the token as used to it can't be retried.

stephencdaly (changes requested):

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
kbottla: good point. Updated to mark token as used before validating card number. Also refactored so there is only call to validate and mark token as used https://github.com/alphagov/pay-connector/pull/3642/commits/a04e742c7f82238753ef25505cb4a3d17a8ffd56

stephencdaly (approved):
LGTM!

PR #3649, opened 2022-04-27
PP-9143 Add logging fields to MDC

stephencdaly:
Add logging field for the secure token, the charge and the gateway
account to the MDC for the authorise API.

These are added by default for other requests where the charge id or
gateway account id are included in the request, but need to be added
manually for this request.

Add MDCUtils class to ensure we are adding the same fields to the MDC
wherever they are added from.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
@@ -3,6 +3,7 @@ package uk.gov.pay.connector.paymentprocessor.resource;
 import com.google.common.collect.ImmutableMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 import uk.gov.pay.connector.charge.service.ChargeCancelService;
 import uk.gov.pay.connector.charge.service.ChargeEligibleForCaptureService;
 import uk.gov.pay.connector.charge.service.DelayedCaptureService;
@@ -18,6 +19,7 @@ import uk.gov.pay.connector.paymentprocessor.service.Card3dsResponseAuthService;
 import uk.gov.pay.connector.paymentprocessor.service.CardAuthoriseService;
 import uk.gov.pay.connector.token.TokenService;
 import uk.gov.pay.connector.token.model.domain.TokenEntity;
+import uk.gov.pay.connector.util.MDCUtils;
 import uk.gov.pay.connector.util.ResponseUtil;
 import uk.gov.pay.connector.wallets.applepay.ApplePayService;
 import uk.gov.pay.connector.wallets.applepay.api.ApplePayAuthRequest;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -174,8 +177,11 @@ public Response cancelCharge(@PathParam("accountId") Long accountId, @PathParam(
     @Path("/v1/api/charges/authorise")
     @Produces(APPLICATION_JSON)
     public Response authorise(@Valid @NotNull AuthoriseRequest authoriseRequest) {
+        MDC.put(SECURE_TOKEN, authoriseRequest.getOneTimeToken());
```
kbottla, line 181: Can we remove SECURE_TOKEN from MDC (after the request is done) ?

stephencdaly: I don't think we need to, as each request has it's own thread. I don't think we remove fields from the MDC when they're added for the entire life of the request.
Or do you mean after we've retrieved the charge from the token?

kbottla: These are removed in the response filter https://github.com/alphagov/pay-connector/blob/8b971c830fc607443026244ae4e4db31d0b8efbb/src/main/java/uk/gov/pay/connector/filters/LoggingMDCResponseFilter.java#L21

stephencdaly: Ahh thanks - I guess we don't really need to remove these then... Or would we want to do it explicitly for clarity?

kbottla: I think the threads goes back to pool and can get reused. Context will stay if not removed explicitly.

kbottla: or may be not. I have got a vague recollection that at some point context got mixed up in publicapi because we have not cleared mdc in right order.

stephencdaly: Yeh from what you showed, it looks like we do explicitly clear it. But the LoggingMDCResponseFilter will also run when this API responds and clear everything it adds.

Maybe I could add a comment to say that it is cleared by the LoggingMDCResponseFilter?
In PublicAPI we just do `MDC.clear()` could we maybe do that instead of listing the fields to clear?

kbottla: clear() will remove everything from the context. Not sure if there is anything else that should stay for a while.

kbottla: Yeah we could add a comment so both are in sync.

stephencdaly: Added a comment

kbottla (approved):
LGTM!

PR #3651, opened 2022-04-27
PP-9369 Check agreement is valid before taking payment from it

jbz149:
Update the validation so that if a request is received with `"authorisation_mode": "AGREEMENT"` and an `"agreement_id"`, we check that the agreement exists, has a payment instrument and the payment instrument is in the correct ACTIVE state.

None of the logic after this validation has changed yet so if the conditions above are true, we’ll just process it as a regular user-present payment but that’s fine because public API will not be sending us anything with `"authorisation_mode": "AGREEMENT"` yet.

If the agreement or payment instrument are not found, the error returned is a 400 (bad request). While a 404 (not found) may seem logical, the problem is that the request references a non-existent resource, not that the request is for a non-existent resource. Making a request to save a payment instrument to a non-existent agreement now also returns a 400 for consistency.

Based on code started by @jbz149 and @kundanscorpio, finished by @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -4,6 +4,7 @@ import com.google.inject.persist.Transactional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.agreement.dao.AgreementDao;
+import uk.gov.pay.connector.agreement.model.AgreementEntity;
 import uk.gov.pay.connector.app.CaptureProcessConfig;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.app.LinksConfig;
@@ -11,11 +12,13 @@ import uk.gov.pay.connector.cardtype.dao.CardTypeDao;
 import uk.gov.pay.connector.cardtype.model.domain.CardTypeEntity;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.exception.AgreementIdWithIncompatibleOtherOptionsException;
+import uk.gov.pay.connector.charge.exception.AgreementMissingPaymentInstrumentException;
 import uk.gov.pay.connector.charge.exception.AgreementNotFoundException;
 import uk.gov.pay.connector.charge.exception.AuthorisationModeAgreementRequiresAgreementIdException;
 import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
 import uk.gov.pay.connector.charge.exception.GatewayAccountDisabledException;
 import uk.gov.pay.connector.charge.exception.MotoPaymentNotAllowedForGatewayAccountException;
+import uk.gov.pay.connector.charge.exception.PaymentInstrumentNotActiveException;
 import uk.gov.pay.connector.charge.exception.SavePaymentInstrumentToAgreementRequiresAgreementIdException;
 import uk.gov.pay.connector.charge.exception.SavePaymentInstrumentToAgreementRequiresAgreementModeWebException;
 import uk.gov.pay.connector.charge.exception.ZeroAmountNotAllowedForGatewayAccountException;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -546,6 +550,7 @@ public <T extends AbstractChargeResponseBuilder<T, R>, R> AbstractChargeResponse
     }
 
     private boolean needsNextUrl(ChargeEntity chargeEntity) {
+        if (chargeEntity.getAuthorisationMode() == AuthorisationMode.AGREEMENT) return false;
```
sfount: This is added without any kind of test coverage -- I wonder if this is worth splitting out to its own separate PR. Happy for that to go in here but we should have some kind of assertions about the behaviour.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -930,4 +934,16 @@ private void checkForUnknownAgreementId(String agreementId) {
                     .orElseThrow(() -> new AgreementNotFoundException(format("Agreement with ID [%s] not found.", agreementId)));
         }
     }
+
+    private void checkInstrumentActiveWhenAuthorisationReference(String agreementId, AuthorisationMode authorisation) {
+        if (agreementId != null && authorisation == AuthorisationMode.AGREEMENT) {
+            Optional<AgreementEntity> agreementEntityOptional = agreementDao.findByExternalId(agreementId);
+            AgreementEntity agreementEntity = agreementEntityOptional.get();
```
sfount: If the agreement doesn't exist, this will throw a `NullPointerException`, in favour of this we should chain these optional requests to always result in a uniform `AgreementNotActiveException` (I believe the way it's currently designed the agreement is required to exist as a separate assertion)

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -930,4 +934,16 @@ private void checkForUnknownAgreementId(String agreementId) {
...
+            PaymentInstrumentStatus instrumentStatus = agreementEntity.getPaymentInstrument().getPaymentInstrumentStatus();
```
sfount: Similar with above, payment instruments are much less likely to be in a guaranteed state at the time of this call -- we should be guarding against this not existing to ensure the code ends in a known state

sfount (changes requested):
Concept looks good, a number of places where the code could quite easily throw unknown exceptions that should be guarded against, question on an additional bit of functionalities fit

meta: could the pr name be updated to reflect what it's doing rather than the jira ticket

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Should now be covered by tests.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Sorted.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Also sorted (with some help from https://github.com/alphagov/pay-connector/pull/3747).

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -259,7 +263,14 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
             }
 
             checkAgreementOptions(authorisationMode, chargeRequest.getAgreementId(), chargeRequest.getSavePaymentInstrumentToAgreement());
-            checkForUnknownAgreementId(chargeRequest.getAgreementId());
+
+            if (chargeRequest.getAgreementId() != null) {
+                var agreementEntity = agreementDao.findByExternalId(chargeRequest.getAgreementId())
+                        .orElseThrow(() -> new AgreementNotFoundException("Agreement with ID [" + chargeRequest.getAgreementId() + "] not found."));
+                if (!chargeRequest.getSavePaymentInstrumentToAgreement()) {
+                    checkAgreementHasActivePaymentInstrument(agreementEntity);
```
sfount, line 271: 🙌 avoids two round trips to the db, and code path is clear

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -964,10 +977,24 @@ private void checkAgreementOptions(AuthorisationMode authorisationMode, String a
         }
     }
 
-    private void checkForUnknownAgreementId(String agreementId) {
+    private Optional<AgreementEntity> getAgreementIfAgreementIdSpecified(String agreementId) {
```
sfount: It looks like this method is no longer called by the code above (in favour of what looks like simpler in-lining), I think it can be removed.

sfount (commented):
Looks great! All issues listed above addressed in a very clear and resilient way. Looks like there's a bit of now-redundant code that can be removed, if you'd prefer to do that in a follow up I'm happy to approve this

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Removed.

sfount (approved):

PR #3655, opened 2022-04-28
PP-9145 Authorise MOTO API payment

kbottla:
## WHAT YOU DID
- Authorises MOTO API payment (synchronously) and handles authorisation response
- Returning 500 http status for authorisation errors which seems appropriate instead of 400


SandorArpa (approved):
LGTM

PR #3681, opened 2022-05-05
PP-9406 Add pact tests for cardid

stephencdaly:
Add a pact test that verifies the success response from cardid when
getting card information.

Run cardid pact provider tests as part of the post-merge github actions
pipeline.


SandorArpa (approved):
LOKTM

PR #3685, opened 2022-05-06
PP-9406 Fix flaky test

stephencdaly:
This test was sometimes failing when run by Github actions as the capture hadn't completed during the 100ms wait time.

Instead of increasing the wait time, just check whether the charge is in either CAPTURED or CAPTURE QUEUED states, as this integration test is not intended to test the capture process.


SandorArpa (approved):
LGTM

PR #3688, opened 2022-05-09
PP-9528 Serialise agreement events

sfount:
New base events implementing the `Event` interface.

`AgreementEvent` and `PaymentInstrumentEvent` implement the simplest
possible level of top fields:
* service identifier
* live

Add POJOs to serialise the data that needs to be persisted during the
agreement and payment journey lifecycle. Extend base level agreement and
payment instrument events.

--- 

This data will drive the event store projections and should therefore line up with:

* https://github.com/alphagov/pay-ledger/blob/master/src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementEntity.java
* https://github.com/alphagov/pay-ledger/blob/master/src/main/java/uk/gov/pay/ledger/agreement/entity/PaymentInstrumentEntity.java

---

This pull request doesn't do anything with the events but prepares them to be used (emitted) during the agreement/ recurring payment flow.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/AgreementEvent.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/AgreementEvent.java
@@ -1,0 +1,46 @@
+package uk.gov.pay.connector.events.model.charge;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+import uk.gov.pay.connector.events.model.Event;
+import uk.gov.pay.connector.events.model.ResourceType;
+
+import java.time.ZonedDateTime;
+
+public class AgreementEvent extends Event {
+    private String serviceId;
+    private Boolean live;
+    
+    public AgreementEvent(String serviceId, boolean live, String resourceExternalId, EventDetails eventDetails, ZonedDateTime timestamp) {
+        super(resourceExternalId, eventDetails, timestamp);
+        this.serviceId = serviceId;
+        this.live = live;
+    }
+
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/events/model/charge/AgreementEvent.java
```diff
@@ -0,0 +1,46 @@
...
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+import uk.gov.pay.connector.events.model.Event;
+import uk.gov.pay.connector.events.model.ResourceType;
+
+import java.time.ZonedDateTime;
+
+public class AgreementEvent extends Event {
+    private String serviceId;
+    private Boolean live;
```
alexbishop1, line 11: Looks like this can be primitive and then we wouldn’t have to worry about it possibly being null.

alexbishop1 (changes requested):
I think we can make the `live` `Boolean` fields into `boolean`s but otherwise fine.

src/main/java/uk/gov/pay/connector/events/model/charge/AgreementEvent.java
sfount: we've had some issues with primitive values on event superclasses before
https://github.com/alphagov/pay-connector/pull/3276
https://github.com/alphagov/pay-connector/pull/3279

I'm confident there's a better pattern for this but `null` values not being serialised hasn't bitten us before where default values have.

alexbishop1 (approved):

PR #3689, opened 2022-05-09
PP-9527 Consume ledger events endpoint

sfount:
Depends on https://github.com/alphagov/pay-connector/pull/3688

---

Adds `postEvent` utility method to parse and POST anything extending
`Event` to Ledger's event endpoint.

Expect an `202 Accepted` response or appropriately throw a Ledger
exception to be handled upstream.

The uility is not used in this change. It will be used during the
agreement and recurring payment journeys.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
@@ -16,6 +16,7 @@ import uk.gov.pay.connector.gateway.PaymentGatewayName;
 
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
+import java.time.Duration;
 
 public class ConnectorConfiguration extends Configuration {
 
@@ -157,6 +158,10 @@ public class ConnectorConfiguration extends Configuration {
 
     public Long getLedgerPostEventTimeoutInMillis() {
         return ledgerPostEventTimeoutInMillis;
+    }
+
+    public Duration getLedgerPostEventTimeout() {
+        return Duration.ofMillis(ledgerPostEventTimeoutInMillis);
     }
 
     @JsonProperty("database")
--- a/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
+++ b/src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
@@ -144,12 +144,19 @@ public class ConnectorConfiguration extends Configuration {
     @JsonProperty("cardidBaseURL")
     private String cardidBaseUrl;
 
+    @NotNull
+    private Long ledgerPostEventTimeoutInMillis;
+
     public String getLedgerBaseUrl() {
         return ledgerBaseUrl;
     }
 
     public String getCardidBaseUrl() {
         return cardidBaseUrl;
+    }
+
+    public Long getLedgerPostEventTimeoutInMillis() {
+        return ledgerPostEventTimeoutInMillis;
     }
 
     @JsonProperty("database")
--- a/src/main/java/uk/gov/pay/connector/app/RestClientFactory.java
+++ b/src/main/java/uk/gov/pay/connector/app/RestClientFactory.java
@@ -8,13 +8,14 @@ import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
+import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
 
 public class RestClientFactory {
     private static final String TLSV1_2 = "TLSv1.2";
 
-    public static Client buildClient(RestClientConfig clientConfig) {
+    public static Client buildClient(RestClientConfig clientConfig, Long connectTimeoutInMillis) {
         ClientBuilder clientBuilder = ClientBuilder.newBuilder();
 
         if (!clientConfig.isDisabledSecureConnection()) {
@@ -27,6 +28,9 @@ public class RestClientFactory {
             }
         }
 
+        if (connectTimeoutInMillis != null) {
+            clientBuilder.connectTimeout(connectTimeoutInMillis, TimeUnit.MILLISECONDS);
+        }
         Client client = clientBuilder.build();
         client.register(RestClientLoggingFilter.class);
 
--- a/src/main/java/uk/gov/pay/connector/app/RestClientFactory.java
+++ b/src/main/java/uk/gov/pay/connector/app/RestClientFactory.java
@@ -8,6 +8,7 @@ import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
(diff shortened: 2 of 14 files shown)


src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
```diff
@@ -152,6 +155,10 @@ public String getCardidBaseUrl() {
         return cardidBaseUrl;
     }
 
+    public Long getLedgerPostEventTimeoutInMillis() {
+        return ledgerPostEventTimeoutInMillis;
+    }
```
alexbishop1, line 161: Can we make this return a [`Duration`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Duration.html)? It’s a more appropriate type for what it is and we could just call the method `getLedgerPostEventTimeout()` and not encode the units in its name. If where we actually use it we just need a `long` of milliseconds we can easily use `Duration.toMillis()` to convert.

src/main/java/uk/gov/pay/connector/app/RestClientFactory.java
```diff
@@ -8,13 +8,14 @@
...
 import java.security.NoSuchAlgorithmException;
+import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
 
 public class RestClientFactory {
     private static final String TLSV1_2 = "TLSv1.2";
 
-    public static Client buildClient(RestClientConfig clientConfig) {
+    public static Client buildClient(RestClientConfig clientConfig, Long connectTimeoutInMillis) {
```
alexbishop1: Again, this should ideally take a `Duration`.

src/main/java/uk/gov/pay/connector/client/ledger/service/LedgerService.java
```diff
@@ -88,6 +95,28 @@ public RefundTransactionsForPayment getRefundsForPayment(Long gatewayAccountId,
...
 
+    public Response postEvent(Event event) {
+        return postEventList(List.of(event)); 
+    }
+    
+    public Response postEvent(List<Event> events) {
+        return postEventList(events);
+    }
+    
+    private Response postEventList(List<Event> events) {
```
alexbishop1: This could just be called `postEvents(…)`: it’s obvious from the parameter that it takes a `List`.

src/main/java/uk/gov/pay/connector/client/ledger/service/LedgerService.java
```diff
@@ -88,6 +95,28 @@ public RefundTransactionsForPayment getRefundsForPayment(Long gatewayAccountId,
...
+        var uri = UriBuilder.fromPath(ledgerUrl).path("/v1/event");
```
alexbishop1: Seems the URI is constant so we could just calculate it once in the constructor.

src/main/java/uk/gov/pay/connector/client/ledger/service/LedgerService.java
```diff
@@ -88,6 +95,28 @@ public RefundTransactionsForPayment getRefundsForPayment(Long gatewayAccountId,
...
+        var response = postEventClient.target(uri)
+                .request()
+                .accept(MediaType.APPLICATION_JSON)
+                .post(Entity.json(events));
```
alexbishop1: I assume this can throw exceptions — are we happy to let them bubble up?

src/test/java/uk/gov/pay/connector/client/ledger/service/LedgerServiceTest.java
```diff
@@ -5,46 +5,58 @@
...
         WebTarget mockWebTarget = mock(WebTarget.class);
-        Invocation.Builder mockBuilder = mock(Invocation.Builder.class);
+        mockBuilder = mock(Invocation.Builder.class);
         mockResponse = mock(Response.class);
 
         when(mockConnectorConfiguration.getLedgerBaseUrl()).thenReturn("http://ledgerUrl");
         when(mockClient.target(any(UriBuilder.class))).thenReturn(mockWebTarget);
         when(mockWebTarget.request()).thenReturn(mockBuilder);
         when(mockBuilder.accept(APPLICATION_JSON)).thenReturn(mockBuilder);
+        when(mockBuilder.post(any())).thenReturn(mockResponse);
```
alexbishop1: The `any()` makes this very loose in terms of what it will accept…

src/test/java/uk/gov/pay/connector/client/ledger/service/LedgerServiceTest.java
```diff
@@ -61,4 +73,29 @@ public void getRefundsFromLedgerShouldThrowExceptionIfResponseCannotBeProcessed(
 
         ledgerService.getRefundsForPayment(152L, "external-id");
     }
+
+    @Test
+    public void serialiseAndSendEvent() {
+        var event = new AgreementCreated("service-id", false, "resource-id", null, ZonedDateTime.now());
```
alexbishop1: `ZonedDateTime.now(ZoneOffset.UTC)` or this is going to create different timestamps depending on the time zone of the computer.

src/test/java/uk/gov/pay/connector/client/ledger/service/LedgerServiceTest.java
```diff
@@ -61,4 +73,29 @@ public void getRefundsFromLedgerShouldThrowExceptionIfResponseCannotBeProcessed(
...
+        when(mockResponse.getStatus()).thenReturn(SC_ACCEPTED);
+        ledgerService.postEvent(event);
+        verify(mockBuilder).post(Entity.json(List.of(event)));
```
alexbishop1: Had to scroll up to remind myself what `mockBuilder` — could this be named better?

alexbishop1 (changes requested):
Looks fairly straightforward but it’s network IO and is that ever straightforward? Few comments below. Another thing is that we’re setting the connect timeout but not the read timeout — do we want to configure that as well?

src/main/java/uk/gov/pay/connector/client/ledger/service/LedgerService.java
sfount: I think the main error surface when dealing with events would be in constructing the event itself (incorrect, missing data etc.). 

I think by the time it it's being passed to this endpoint to be processed we're fairly sure we'll get a POST body out if it. If not I can't think of what we'd do other than propagate to the calling code either way. Definitely room for improvement but I think OK for now.

sfount:
> Looks fairly straightforward but it’s network IO and is that ever straightforward? Few comments below. Another thing is that we’re setting the connect timeout but not the read timeout — do we want to configure that as well?

I've added a story to ensure we've got sensible defaults for this client (it may feed in with other work off the back of worldpay investigation) -- the connect timeout at least isolates connector trying to reach this endpoint from other activities which is why it was originally in the spec.

src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
```diff
@@ -152,6 +156,14 @@ public String getCardidBaseUrl() {
         return cardidBaseUrl;
     }
 
+    public Long getLedgerPostEventTimeoutInMillis() {
```
alexbishop1, line 159: Do we need this method now we have `getLedgerPostEventTimeout()`?

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/app/ConnectorConfiguration.java
sfount: At the moment it's needed by jackson to validate the not null constraint, in this ticket I've suggested that all of this is moved out to a separate configuration which can properly use durations etc. https://github.com/alphagov/pay-connector/pull/3689#issuecomment-1134669919

PR #3693, opened 2022-05-10
PP-9503 Add timeout for synchronous authorisation API

stephencdaly:
Time out a request to authorise a payment made through the `/v1/api/charges/authorise` endpoint after a defined period of time, and transition the charge into the AUTHORISATION_TIMEOUT when it is timed out. This API is used for authorising MOTO payments, which needs to be done synchronously.

This differs from authorisations made via the `/v1/frontend/charges/{chargeId}/cards`, where we will respond to the request if the authorisation hasn't finished after a defined period of time, but allow the authorisation to continue on a background thread.

Re-use the CardExecutorService to execute a Future to allow for the request to be timed out after a certain period of time.

The timeout is configured using a new environment variable, SYNCHRONOUS_AUTH_TIMEOUT_IN_MILLISECONDS.

Move the existing timeoutInSeconds configuration option in ExecutorServiceConfig to asynchronousAuthTimeoutInSeconds in AuthorisationConfig to reflect CardExecutorService now having the timeout specified externally. This new configuration option uses the existing environment variable AUTH_READ_TIMEOUT_SECONDS.

Reduce the constant QUEUE_WAIT_WARN_THRESHOLD_MILLIS from 10000 to 1000, so we are able to monitor if any authorisations are waiting longer than 1s on the executor queue - which would be an issue for synchronous authorisations.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
@@ -3,10 +3,14 @@ package uk.gov.pay.connector.paymentprocessor.service;
 import com.codahale.metrics.MetricRegistry;
 import com.google.inject.persist.Transactional;
 import io.dropwizard.setup.Environment;
+import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.cardtype.dao.CardTypeDao;
 import uk.gov.pay.connector.cardtype.model.domain.CardTypeEntity;
+import uk.gov.pay.connector.charge.dao.ChargeDao;
+import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
+import uk.gov.pay.connector.charge.exception.motoapi.AuthorisationTimedOutException;
 import uk.gov.pay.connector.charge.model.domain.Auth3dsRequiredEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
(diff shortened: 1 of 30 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -40,52 +45,79 @@
...
+
+            authorisationResponse.getAuthoriseStatus().ifPresent(authoriseStatus -> {
+                if (authoriseStatus.getMappedChargeStatus() == AUTHORISATION_SUCCESS) {
+                    chargeEligibleForCaptureService.markChargeAsEligibleForCapture(chargeEntity.getExternalId());
+                }
+            });
+
+            return authorisationResponse;
+        } catch (AuthorisationExecutorTimedOutException e) {
+            updateChargeToAuthorisationTimeout(charge.getExternalId());
```
kbottla, line 108: We need to update charge with `authCardDetails`, gateway transaction ID (for worldpay)

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java
```diff
@@ -109,11 +109,11 @@ public <T> Pair<ExecutionStatus, T> execute(Supplier<T> callable) {
         });
 
         try {
-            return Pair.of(COMPLETED, futureObject.get(config.getTimeoutInSeconds(), TimeUnit.SECONDS));
+            return Pair.of(COMPLETED, futureObject.get(timeoutInMilliseconds, TimeUnit.MILLISECONDS));
```
kbottla, line 112: If CardExecutorService is used we need to make sure shutdown hook waits long enough for sync operations https://github.com/alphagov/pay-connector/blob/46c1737e23234cc449bc4f37ee5c30856b4402bb/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java#L77

PR #3696, opened 2022-05-11
PP-9503 Refactor CardExecutorService

stephencdaly:
Refactor CardExecutorService so it can be use to execute synchronous authorisations with a timeout.

Currently, the CardExecutorService executes a Future and waits for a fixed timeout defined by the timeoutInSeconds configuration option in ExecutorServiceConfig and returns an ExecutionStatus of IN_PROGRESS if it has to wait for longer than this amount of time to authorise. AuthorisationService handles this service to throw an exception that maps to a success response, and allows the authorisation to continue.

Refactor CardExecutorService so we can specify a different timeout to wait on the thread that is handling the authorisation API request for the authorisation task to complete.

Move the existing timeoutInSeconds configuration option in ExecutorServiceConfig to asynchronousAuthTimeoutInMilliseconds in AuthorisationConfig to reflect CardExecutorService now having the timeout specified externally. The environment variable used has changed to AUTH_READ_TIMEOUT_MILLISECONDS to reflect the change in time unit. It is fine to change this as it is not set to anything other than the default in our infra anywhere.

Add a method to AuthorisationService that will execute a task on the CardExecutorService synchronously with a timeout specified by a new synchronousAuthTimeoutInMilliseconds configuration option which takes its value from a SYNCHRONOUS_AUTH_TIMEOUT_IN_MILLISECONDS environment variable. This method will throw an exception if the CardExecutorService returns an ExecutionStatus of IN_PROGRESS. This method is not currently used, but will be used for authorising synchronous authorisations.

Reduce the constant QUEUE_WAIT_WARN_THRESHOLD_MILLIS from 10000 to 1000, so we are able to monitor if any authorisations are waiting longer than 1s on the executor queue - which would be an issue for synchronous authorisations.

Increase the timeout for waiting for tasks to finish on the CardExecutorService to 10s so we wait for long enough for synchronous authorisation tasks to expire before shutting down.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
@@ -7,6 +7,8 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.ConnectorConfiguration;
+import uk.gov.pay.connector.app.config.AuthorisationConfig;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.common.exception.OperationAlreadyInProgressRuntimeException;
@@ -14,6 +16,7 @@ import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.exception.GenericGatewayRuntimeException;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.paymentprocessor.exception.AuthorisationExecutorTimedOutException;
 import uk.gov.pay.connector.paymentprocessor.model.OperationType;
 
 import javax.inject.Inject;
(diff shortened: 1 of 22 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
```diff
@@ -26,30 +29,48 @@
...
+    public AuthorisationService(CardExecutorService cardExecutorService, Environment environment, ConnectorConfiguration configuration) {
         this.cardExecutorService = cardExecutorService;
         this.metricRegistry = environment.metrics();
+        this.authorisationConfig = configuration.getAuthorisationConfig();
     }
- 
+
     public <T> T executeAuthorise(String chargeId, Supplier<T> authorisationSupplier) {
-        Pair<ExecutionStatus, T> executeResult = cardExecutorService.execute(authorisationSupplier);
+        int timeoutInMilliseconds = authorisationConfig.getAsynchronousAuthTimeoutInSeconds() * 1000;
```
kbottla: Any reason this is converted to milliseconds instead of down as seconds ?

src/main/resources/config/config.yaml
```diff
@@ -284,3 +283,7 @@ expungeConfig:
 
 authorisation3dsConfig:
   maximumNumberOfTimesToAllowUserToAttempt3ds: ${MAXIMUM_NO_USER_3DS_ATTEMPTS:-1}
+
+authorisationConfig:
+  asynchronousAuthTimeoutInSeconds: ${AUTH_READ_TIMEOUT_SECONDS:-1}
```
kbottla: Can we have one unit (seconds or milliseconds) for both variables ?

stephencdaly: The reason why I wanted milliseconds for the new variable was to make it easier to test more realistically without having to make the test wait a whole second. For the current tests we set `asynchronousAuthTimeoutInSeconds` to 0s to test the timeout, so I could potentially just copy that.

I didn't want to change the unit of the existing environment variable, as that would mean updating where we set it - although it seems we only ever set it in the .env file for connector used by pay local, and set it to 1 there anyway, so I can just use a different environment variable that is milliseconds and remove setting that.

kbottla (approved):

PR #3702, opened 2022-05-13
PP-8608 remove deprecated stripe contract fee handling

SandorArpa:
## WHAT YOU DID
- Stripe contract fee V2 is now applicable to all services which use Stripe. Remove fee break down, fee for test accounts, fee for gateway account list, and V2 start date.


stephencdaly (approved):
LGTM!

PR #3705, opened 2022-05-16
PP-9585: Test validateAfterInactivityPeriod set to 500ms

jfharden:
## WHAT YOU DID

## How to test

- How should it be reviewed? 
- What feedback do you need? 

### Logging

### Documentation

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -188,6 +188,9 @@ jerseyClient:
   userAgent: connector
   gzipEnabledForRequests: false
 
+  # Validates the connection is still alive if it has been unused for this long
+  validateAfterInactivityPeriod: 500ms
+
 customJerseyClient:
   # Sets the read timeout to a specified timeout, in
   # milliseconds. A non-zero value specifies the timeout when


heathd:
Do we know what the recommended value for this setting is? I've done some googling and poking around...

This article was a red-herring.. the date on it claims 2022 but the comments suggest it's 5 years old. https://www.baeldung.com/httpclient-connection-management 

The validate setting was [introduced in 2014](https://github.com/apache/httpcomponents-client/commit/c5c0bc680180c3ccac90430a6380901dc1e968df) (but [not included in a release until 2019](https://github.com/apache/httpcomponents-client/releases/tag/rel%2Fv4.4)).

When introduced it had a default of 1000ms.

[Current default appears to be 2000ms](https://github.com/apache/httpcomponents-client/blob/rel/v4.5.13/httpclient/src/main/java/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.java#L183)

Although it looks like it was also set to [5000ms](https://github.com/apache/httpcomponents-client/commit/6e6b8e5180e94da14efd60f4735416b9b873ae4b) at some point

Can't find anything explaining the rationale for it

heathd:
given that it's defaulted in the `PoolingHttpClientConnectionManager` constructor, I'm also wondering if it's already set, in which case the hypothesis that we're seeing these errors because of failure to revalidate the connection might not stand

heathd:
I've stepped through an integration test in a debugger and it looks like the default config setting of 2000ms is being used if this value is not specified.

So... 🤔

oswaldquek:
> I've stepped through an integration test in a debugger and it looks like the default config setting of 2000ms is being used if this value is not specified.
> 
> So... 🤔

It's probably because you needed to change [test-it-config.yaml](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/CardResourceAuthoriseIT.java#L63). I've tested it and the value is 500ms.

jfharden:
No longer required

PR #3708, opened 2022-05-17
PP-9369 Allow "agreement_id" and "authorisation_mode": "agreement"

alexbishop1:
When creating a charge, update the validation to accept `"agreement_id"` if accompanied by `"authorisation_mode":
"agreement"`. Previously, `"agreement_id"` was only accepted if accompanied by `"save_payment_instrument_to_agreement": true` (which remains allowed). A request with `"agreement_id"`,
`"authorisation_mode": "agreement"` and `"save_payment_instrument_to_agreement": true` is not allowed.

This only affects the validation; there are no changes to how the charge is processed once it is created. If a charge is created with an `"agreement_id"` and `"authorisation_mode": "agreement"`, it will be treated the same as a regular charge with neither of those properties. Only public API makes requests to the connector endpoint altered by this commit, so that will not happen for now.

This commit also expands the test coverage for existing functionality relating to creating charges associated with agreements.


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -282,7 +280,7 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
                     .withGatewayAccountCredentialsEntity(gatewayAccountCredential)
                     .withPaymentProvider(gatewayAccountCredential.getPaymentProvider())
                     .withEmail(isBlank(chargeRequest.getEmail()) ? null : chargeRequest.getEmail())
-                    .withLanguage(language)
+                    .withLanguage(chargeRequest.getLanguage().orElse(SupportedLanguage.ENGLISH))
```
sfount: 👍 neat cleanup

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -922,11 +920,23 @@ private void checkIfMotoPaymentsAllowed(boolean moto, GatewayAccountEntity gatew
...
     }
 
-    private void checkAgreementIdAndSaveInstrumentBothPresent(String agreementId, boolean savePaymentInstrumentToAgreement) {
-        if (agreementId != null && !savePaymentInstrumentToAgreement) {
-            throw new AgreementIdAndSaveInstrumentMandatoryInputException("If [agreement_id] is present, [save_payment_instrument_to_agreement] must be true");
-        } else if (savePaymentInstrumentToAgreement && agreementId == null) {
-            throw new AgreementIdAndSaveInstrumentMandatoryInputException("If [save_payment_instrument_to_agreement] is true, [agreement_id] must be specified");
+    private void checkAgreementOptions(String agreementId, boolean savePaymentInstrumentToAgreement, AuthorisationMode authorisationMode) {
+        if (agreementId != null) {
+            if (savePaymentInstrumentToAgreement && authorisationMode == AuthorisationMode.AGREEMENT) {
```
sfount: I wonder if this logic could be `savePaymentInstrumentToAgreement && authorisationMode != AuthorisationMode.WEB` -- as the payment details can only be persisted with the user present to satisfy a 3DS2 challenge it's unlikely that this would change and is the most specific version of this check

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -922,11 +920,23 @@ private void checkIfMotoPaymentsAllowed(boolean moto, GatewayAccountEntity gatew
         }
     }
 
-    private void checkAgreementIdAndSaveInstrumentBothPresent(String agreementId, boolean savePaymentInstrumentToAgreement) {
-        if (agreementId != null && !savePaymentInstrumentToAgreement) {
-            throw new AgreementIdAndSaveInstrumentMandatoryInputException("If [agreement_id] is present, [save_payment_instrument_to_agreement] must be true");
-        } else if (savePaymentInstrumentToAgreement && agreementId == null) {
-            throw new AgreementIdAndSaveInstrumentMandatoryInputException("If [save_payment_instrument_to_agreement] is true, [agreement_id] must be specified");
+    private void checkAgreementOptions(String agreementId, boolean savePaymentInstrumentToAgreement, AuthorisationMode authorisationMode) {
+        if (agreementId != null) {
```
sfount: No actual issues with this as it does (and has good coverage! to) do what it says on the tin. I suspect these nested conditions might be easier to parse if it was organised by activity, something like

```
- IF auth mode IS agreement
  - THEN agreement_id MUST be specified
  - THEN save_payment_instrument_to_agreement MUST NOT be specified

- IF auth mode IS web
  - IF agreement_id IS specified
    - THEN save_payment_instrument_to_agreement MUST be specified
```

sfount (dismissed):
Looks good, few minor comments but nothing blocking in my opinion

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: Yeah, I can do that.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: I’ll take another look.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: I made this seemingly innocuous change and it broke some tests! It turns out this was because the way we defaulted unspecified values when creating charges didn’t play nicely with the way the tests instantiate `ChargeCreateRequest` objects, leading to surprise `NullPointerException`s. Fixed that now (details in the updated commit message).

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
alexbishop1: OK, done this and changed some of the exceptions and tests to more closely match the revised logic.

sfount (approved):

PR #3714, opened 2022-05-18
PP-9632 Allow specifying reply-to email per gateway account

stephencdaly:
For gateway accounts that have email custom branding enabled, allow specifying the "email_reply_to_id" that we send to Notify in the request to send an email. This will allow for multiple services for the same organisation to use the same Notify service for custom email branding, but with different reply-to email addresses.

Read the value to send to notify from a "email_reply_to_id" field in the JSON object stored in the "notify_settings" database column on the "gateway_accounts" table. If none is set, don't send a value to Notify so that the default reply-to email address for the Notify service is used.

I have tested this locally, hooking connector up to a test account on Notify and tested sending confirmation emails for a gateway account with custom branding both with and without the "email_reply_to_id" set.


SandorArpa (approved):
LGTM

PR #3717, opened 2022-05-19
PP-9503 Don't store updated gateway transaction id on ChargeEntity

stephencdaly:
When making an authorisation for Worldpay, if we request a 3DS exemption and it is soft declined, we need to make the second authorisation request with a new gateway transaction id (order code).

We were updating the gatewayTransactionId field on the ChargeEntity to store this new id for retrieval when building the request to send to Worldpay. This is confusing, as this updated value is never persisted until after the Authorisation is complete, when it is extracted from the BaseGatewayResponse.

Improve this by storing the gateway transaction id in a separate field on the AuthorisationGatewayRequest. When we make the second authorisation request in WorldpayPaymentProvider, construct a new CardAuthorisationGatewayRequest from the original request and a new gatewayTransactionId. This pattern will become more obvious when the AuthorisationGatewayRequest no longer allows retrieval of the ChargeEntity object.

Also remove the overridden `getGatewayCredentials` on CardAuthorisationGatewayRequest and WalletAuthorisationGatewayRequest as these just duplicated the method on the abstract class.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
@@ -9,10 +9,17 @@ import java.util.Map;
 import java.util.Optional;
 
 public abstract class AuthorisationGatewayRequest implements GatewayRequest {
-    protected ChargeEntity charge;
+    private final ChargeEntity charge;
+    private final String gatewayTransactionId;
     
-    public AuthorisationGatewayRequest(ChargeEntity charge) {
+    protected AuthorisationGatewayRequest(ChargeEntity charge) {
         this.charge = charge;
+        this.gatewayTransactionId = charge.getGatewayTransactionId();
+    }
+    
+    protected AuthorisationGatewayRequest(ChargeEntity charge, String gatewayTransactionId) {
+        this.charge = charge;
+        this.gatewayTransactionId = gatewayTransactionId;
     }
 
     public ChargeEntity getCharge() {
@@ -20,7 +27,7 @@ public abstract class AuthorisationGatewayRequest implements GatewayRequest {
     }
 
     public Optional<String> getTransactionId() {
-        return Optional.ofNullable(charge.getGatewayTransactionId());
+        return Optional.ofNullable(gatewayTransactionId);
     }
 
     public String getAmount() {
--- a/src/main/java/uk/gov/pay/connector/gateway/model/request/CardAuthorisationGatewayRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/request/CardAuthorisationGatewayRequest.java
@@ -3,13 +3,16 @@ package uk.gov.pay.connector.gateway.model.request;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 
-import java.util.Map;
-
 public class CardAuthorisationGatewayRequest extends AuthorisationGatewayRequest {
     private AuthCardDetails authCardDetails;
-    
+
     public CardAuthorisationGatewayRequest(ChargeEntity charge, AuthCardDetails authCardDetails) {
         super(charge);
+        this.authCardDetails = authCardDetails;
+    }
+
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
```diff
@@ -10,17 +10,19 @@
...
     }
 
     public ChargeEntity getCharge() {
         return charge;
     }
 
     public Optional<String> getTransactionId() {
-        return Optional.ofNullable(charge.getGatewayTransactionId());
+        return Optional.ofNullable(gatewayTransactionId);
     }
```
alexbishop1, line 31: This is also called from `WorldpayWalletAuthorisationHandler.buildWalletAuthoriseOrder(…)` and it looks like it’s expecting to get back the value set on the `ChargeEntity` using `WalletAuthoriseService.prepareChargeForAuthorisation(…)`.

alexbishop1 (changes requested):
I think this will have an unintended effect on Apple Pay and Google Pay payments.

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
stephencdaly: I think this should be fine, as the initial value of `gatewayTransactionId` is still coming from the ChargeEntity. What's changed here is when we want to update the gatewayTransactionId used to build requests, we update this value rather than the value on the ChargeEntity which is never persisted to the database by the code that updates it in this request.

alexbishop1: You’re right, I’d missed that the `this.gatewayTransactionId = charge.getGatewayTransactionId();` line in the constructor.

That said, I’m still a little uneasy that we’re mutating the request `CardAuthorisationGatewayRequest` and then reusing it for the second, exemption-free request (granted this is not too different from mutating the `ChargeEntity` exposed by the `CardAuthorisationGatewayRequest`).

Could we modify `WorldpayPaymentProvider.authorise(…)` so that if it’s going to make a second request, it creates a whole new `CardAuthorisationGatewayRequest`, calling an overloaded constructor that takes the new transaction ID along with the `ChargeEntity` and the `AuthCardDetails` (which could be retrieved from the original `CardAuthorisationGatewayRequest`)?

Then we could make all the fields in `CardAuthorisationGatewayRequest` and `AuthorisationGatewayRequest` final, which would make it a bit easier to reason about.

stephencdaly: This is a good idea thanks. It will make it clearer that we're making a new request that has different parameters.

stephencdaly: done

src/main/java/uk/gov/pay/connector/gateway/model/request/CardAuthorisationGatewayRequest.java
```diff
@@ -21,8 +24,10 @@ public static CardAuthorisationGatewayRequest valueOf(ChargeEntity charge, AuthC
...
     }
 
-    @Override
-    public Map<String, String> getGatewayCredentials() {
-        return charge.getGatewayAccountCredentialsEntity().getCredentials();
+    public static CardAuthorisationGatewayRequest valueOf(CardAuthorisationGatewayRequest cardAuthorisationGatewayRequest, String gatewayTransactionId) {
+        return new CardAuthorisationGatewayRequest(
+                cardAuthorisationGatewayRequest.getCharge(),
+                cardAuthorisationGatewayRequest.getAuthCardDetails(),
+                gatewayTransactionId);
```
alexbishop1, line 31: Maybe this could be an instance method with the signature `public CardAuthorisationGatewayRequest withNewTransactionId(String newTransactionId)`.

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/gateway/model/request/CardAuthorisationGatewayRequest.java
stephencdaly: Thanks, I'll iterate on this when I come to removing the `getCharge` method. It might change a bit more what it looks like.

PR #3721, opened 2022-05-19
PP-9503 Add ChargeEntity parameter to authorise method

stephencdaly:
Add a ChargeEntity parameter to the PaymentProvider.authorise method.

Previously this was passed inside the AuthorisationGatewayRequest. However, we don't want the code that does the actual authorisation HTTP request to have access to the ChargeEntity as this code should not be able to persist any data on the charge.

The WorldpayPaymentProvider requires the ChargeEntity as it needs to do some updates to charge using the result of a 3DS exemption request.

We are doing this because want to create a separate authorisation method that doesn't need the ChargeEntity and does not attempt to persist anything for use by the MOTO authorisation API code, which will run this code in a separate thread with a timeout before attempting to update the charge itself. This code will  still need to use the AuthorisationGatewayRequest, so removing the  ChargeEntity from this will make it easier for us to discourage introducing persistence to this code path which could cause database conflicts.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
@@ -16,7 +16,6 @@ import uk.gov.pay.connector.common.model.domain.Address;
 import uk.gov.pay.connector.gateway.ClientFactory;
 import uk.gov.pay.connector.gateway.GatewayClientFactory;
 import uk.gov.pay.connector.gateway.GatewayOperation;
-import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.model.Auth3dsResult;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
 import uk.gov.pay.connector.gateway.model.GatewayError;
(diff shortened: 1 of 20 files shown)


src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
```diff
@@ -264,20 +263,12 @@ private GatewayAccountEntity buildTestGatewayAccountWith3ds2Entity() {
         return gatewayAccount;
     }
 
-    CardAuthorisationGatewayRequest buildTestAuthorisation3dsRequest() {
-        return buildTestAuthorisation3dsRequest(buildTestGatewayAccountWith3dsEntity());
+    ChargeEntity buildTest3ds1Charge() {
```
alexbishop1: Since it’s the gateway account configuration that controls whether it’s 3DS1 or 3DS2, rather than any properties of the charge itself, could we call this `buildTestChargeFor3ds1GatewayAccount()` or similar?

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
```diff
@@ -355,36 +329,25 @@ private CardAuthorisationGatewayRequest buildTestAuthorisation3ds2RequestWithPro
...
     }
 
     private RefundGatewayRequest buildTestRefundRequest(GatewayAccountEntity gatewayAccountEntity, GatewayAccountCredentialsEntity credentialsEntity) {
-        ChargeEntity chargeEntity = buildChargeEntity(gatewayAccountEntity);
+        ChargeEntity chargeEntity = buildChargeEntityWithTransactionId(gatewayAccountEntity);
         return buildTestRefundRequest(Charge.from(chargeEntity), gatewayAccountEntity, credentialsEntity);
     }
 
-    protected ChargeEntity buildChargeEntity() {
+    protected ChargeEntity buildChargeEntityWithTransactionId() {
```
alexbishop1: A name like `buildChargeEntityWithTransactionId()` makes me think the transaction ID would be passed in (as in, ‘build charge entity with _this_ transaction ID’). But I can’t think of another name apart from `buildChargeEntityThatHasTransactionId()`, which is a bit clunky.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -191,12 +191,13 @@ void should_not_include_exemption_if_account_has_no_worldpay_3ds_flex_credential
         gatewayAccountEntity.setWorldpay3dsFlexCredentialsEntity(null);
         chargeEntityFixture.withGatewayAccountEntity(gatewayAccountEntity);
 
-        var cardAuthRequest = new CardAuthorisationGatewayRequest(chargeEntityFixture.build(), anAuthCardDetails().build());
+        ChargeEntity charge = chargeEntityFixture.build();
```
alexbishop1: Seems this class mixes `charge` and `chargeEntity` as local variable names — can we standardise on one?

alexbishop1 (changes requested):
Looks fine. Just a couple of minor things.

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
stephencdaly: done

src/test/java/uk/gov/pay/connector/gateway/epdq/BaseEpdqPaymentProviderIT.java
stephencdaly: done

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
stephencdaly: done

alexbishop1 (approved):

PR #3725, opened 2022-05-20
PP-9624: Make connection TTL configurable

oswaldquek:
This will allow for easier experimentation.

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -191,6 +191,7 @@ customJerseyClient:
   # for read, a java.net.SocketTimeoutException is raised. A
   # timeout of zero is interpreted as an infinite timeout.
   readTimeout: 90000ms
+  connectionTTL: ${CONNECTION_TTL:-60s}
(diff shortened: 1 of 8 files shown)


katstevens (commented):
Happy that the `CONNECTION_TTL` setting is correct - someone else should approve the Java bits though 😸

src/main/resources/config/config.yaml
```diff
@@ -191,6 +191,7 @@ customJerseyClient:
   # for read, a java.net.SocketTimeoutException is raised. A
   # timeout of zero is interpreted as an infinite timeout.
   readTimeout: 90000ms
+  connectionTTL: ${CONNECTION_TTL:-60s}
```
katstevens, line 194: Should we also add an explanatory comment for why we've added this setting? Something like 

```
# Used by the HttpClientConnectionManager to ensure stale pool connections are discarded within the stated time
```

stephencdaly (approved):
Looks good!

PR #3726, opened 2022-05-20
PP-9503 Don't store ChargeEntity on AuthorisationGatewayRequest

stephencdaly:
We want to discourage the code that deals with the AuthorisationGatewayRequest to perform the authorisation with the gateway from updating the charge in the database.

To help with this, don't store the entire ChargeEntity on the AuthorisationGatewayRequest.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.connector.gateway.model.request;
 
+import uk.gov.pay.connector.charge.model.ServicePaymentReference;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.util.CorporateCardSurchargeCalculator;
 import uk.gov.pay.connector.gateway.GatewayOperation;
@@ -17,7 +18,7 @@ public abstract class AuthorisationGatewayRequest implements GatewayRequest {
     private final boolean moto;
     private final String amount;
     private final String description;
-    private final String reference;
+    private final ServicePaymentReference reference;
     private final String chargeExternalId;
     private final Map<String, String> credentials;
     private final GatewayAccountEntity gatewayAccount;
@@ -31,7 +32,7 @@ public abstract class AuthorisationGatewayRequest implements GatewayRequest {
         this.moto = charge.isMoto();
         this.amount = String.valueOf(CorporateCardSurchargeCalculator.getTotalAmountFor(charge));
         this.description = charge.getDescription();
-        this.reference = charge.getReference().toString();
+        this.reference = charge.getReference();
         this.chargeExternalId = charge.getExternalId();
         this.credentials = Optional.ofNullable(charge.getGatewayAccountCredentialsEntity()).map(GatewayAccountCredentialsEntity::getCredentials).orElse(null);
         this.gatewayAccount = charge.getGatewayAccount();
@@ -43,7 +44,7 @@ public abstract class AuthorisationGatewayRequest implements GatewayRequest {
                                        boolean moto,
                                        String amount,
                                        String description,
-                                       String reference,
+                                       ServicePaymentReference reference,
                                        String chargeExternalId,
                                        Map<String, String> credentials,
                                        GatewayAccountEntity gatewayAccount) {
@@ -83,7 +84,7 @@ public abstract class AuthorisationGatewayRequest implements GatewayRequest {
         return description;
     }
 
-    public String getReference() {
+    public ServicePaymentReference getReference() {
         return reference;
     }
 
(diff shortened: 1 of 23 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
```diff
@@ -4,56 +4,101 @@
...
 public abstract class AuthorisationGatewayRequest implements GatewayRequest {
-    private final ChargeEntity charge;
     private final String gatewayTransactionId;
-    
+    private final String email;
+    private final SupportedLanguage language;
+    private final boolean moto;
+    private final String amount;
+    private final String description;
+    private final String reference;
```
alexbishop1: We have a `ServicePaymentReference` type for this.

src/main/java/uk/gov/pay/connector/gateway/model/request/CardAuthorisationGatewayRequest.java
```diff
@@ -11,8 +11,21 @@ public CardAuthorisationGatewayRequest(ChargeEntity charge, AuthCardDetails auth
...
+        super(gatewayTransactionId,
+                other.getEmail(),
+                other.getLanguage(),
+                other.isMoto(),
+                other.getAmount(),
+                other.getDescription(),
+                other.getReference(),
+                other.getChargeExternalId(),
+                other.getGatewayCredentials(),
+                other.getGatewayAccount());
```
alexbishop1, line 28: I kinda like how this makes it clear what data is used for the authorisation. Maybe in the future we could similarly deconstruct `authCardDetails` and make authorisation less coupled to the source of the data (especially now we have data coming from API requests and soon stored payment instruments in addition to data entered on the card details page).

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -262,13 +267,17 @@ void should_set_exemption_rejected_when_request_made_with_an_exemption() throws
 
         when(worldpayAuthoriseHandler.authoriseWithExemption(cardAuthRequest))
                 .thenReturn(getGatewayResponse(WORLDPAY_EXEMPTION_REQUEST_SOFT_DECLINE_RESULT_REJECTED_RESPONSE));
-        when(worldpayAuthoriseHandler.authoriseWithoutExemption(cardAuthRequest))
+        when(worldpayAuthoriseHandler.authoriseWithoutExemption(any(CardAuthorisationGatewayRequest.class)))
```
alexbishop1, line 270: `any(…)` makes me sad.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -138,8 +138,8 @@ null, mock(AgreementDao.class), null, mock(ConnectorConfiguration.class), null,
 
         when(mockedProviders.byName(WORLDPAY)).thenReturn(mockedWorldpayPaymentProvider);
         when(mockedWorldpayPaymentProvider.generateTransactionId()).thenReturn(Optional.of(TRANSACTION_ID));
-        when(mockedWorldpayPaymentProvider.generateAuthorisationRequestSummary(charge, authCardDetails))
-                .thenReturn(new WorldpayAuthorisationRequestSummary(charge, authCardDetails));
+        when(mockedWorldpayPaymentProvider.generateAuthorisationRequestSummary(charge.getGatewayAccount(), authCardDetails))
```
alexbishop1: I think this (and similar lines) would be a bit less noisy if there was `gatewayAccount` local variable.

alexbishop1 (dismissed):
No showstoppers.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
stephencdaly: Is there a better solution you know of rather than using `any()` and then `verify` with an argument captor?

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
stephencdaly: I left it as `String` so I wasn't changing how this class looked like from the outside too much. But I guess where we get the reference to use for authorisation we can call the `toString()` there...

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
alexbishop1: An `equals(…)` method on `CardAuthorisationGatewayRequest` might make it easier because then we could construct an expected `CardAuthorisationGatewayRequest` (though that might end up being a bit verbose).

Other alternatives are more drastic, like having a separate dependency-injected factory that takes a `CardAuthorisationGatewayRequest` and returns a new one with a different transaction ID — the interaction with this factory could then be mocked.

I approved the PR so I’m OK with using `any(…)` but it usually indicates there’s something not quite right about the code structure if we have to use it (not that I’m blaming you for that — the code has been evolved by many people over many years!).

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
stephencdaly: done

src/main/java/uk/gov/pay/connector/gateway/model/request/AuthorisationGatewayRequest.java
stephencdaly: changed the signature to use ServicePaymentReference

alexbishop1 (approved):

PR #3727, opened 2022-05-20
BAU add gateway account id to failed refund notification

SandorArpa:
## WHAT YOU DID
- by adding the gateway account id, we are able to check refund notifications that has no gateway transaction id against gateways. This helps us identify certain behaviours, eg service refunding payments through Worldpay portal.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
@@ -16,6 +16,7 @@ import java.util.Optional;
 
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static org.apache.commons.lang3.StringUtils.isBlank;
+import static uk.gov.service.payments.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
 import static uk.gov.service.payments.logging.LoggingKeys.PAYMENT_EXTERNAL_ID;
 import static uk.gov.service.payments.logging.LoggingKeys.PROVIDER;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
```diff
@@ -36,8 +36,8 @@ public class RefundNotificationProcessor {
     public void invoke(PaymentGatewayName gatewayName, RefundStatus newStatus, GatewayAccountEntity gatewayAccountEntity,
                        String gatewayTransactionId, String transactionId, Charge charge) {
         if (isBlank(gatewayTransactionId)) {
-            logger.warn("{} refund notification could not be used to update charge [{}] (missing reference)",
-                    gatewayName, charge.getExternalId());
+            logger.warn("{} refund notification could not be used to update charge [{}] (missing reference), gateway account: [{}]",
+                    gatewayName, charge.getExternalId(), gatewayAccountEntity.getId());
```
stephencdaly: Could we maybe add the charge external id and the gateway account id to the MDC in the notification processor, so all log lines have them when they're available? This would probably have to be done for each to WorldpayNotifcationService, SmartpayNotificationService etc..

There's a helper class called MDCUtils to add all the relevant logging fields. We'd need to make sure we remove them from the MDC after processing the notification though. Possibly another method could be added to MDCUtils to do this.

stephencdaly: Agreed it's a bit too much work to do this for now - so just adding the fields to the log line using `kv` is good enough to do what we want.

stephencdaly (approved):

PR #3728, opened 2022-05-20
PP-9503 Rename chargeExternalId in AuthorisationGatewayRequest

stephencdaly:
Rename chargeExternalId to govUkPayPaymentId, with the intention of discouraging consumers of this class from using this id to look up the charge in the database and updating it, as code making the authorisation request to the gateway should not make database updates.

Remove usages of the chargeExternalId for logging purposes. There is no need to explicitly include the chargeExternalId in log lines, as it is already added to the structured logging by virtue of it being on the MDC.


kbottla (approved):
LGTM!

PR #3743, opened 2022-05-30
PP-9599 implement query stripe payment status

SandorArpa:
## WHAT YOU DID
- implement checking Stripe's payment intent status. Searching by metadata key.
- add tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeQueryPaymentStatusHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeQueryPaymentStatusHandler.java
@@ -1,0 +1,82 @@
+package uk.gov.pay.connector.gateway.stripe.handler;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.app.StripeGatewayConfig;
+import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
+import uk.gov.pay.connector.gateway.ChargeQueryGatewayRequest;
+import uk.gov.pay.connector.gateway.ChargeQueryResponse;
+import uk.gov.pay.connector.gateway.GatewayClient;
+import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.stripe.json.StripeErrorResponse;
+import uk.gov.pay.connector.gateway.stripe.json.StripePaymentIntent;
+import uk.gov.pay.connector.gateway.stripe.json.StripeSearchResponse;
+import uk.gov.pay.connector.gateway.stripe.model.StripeChargeStatus;
+import uk.gov.pay.connector.gateway.stripe.request.StripeQueryPaymentStatusRequest;
+import uk.gov.pay.connector.gateway.stripe.response.StripeQueryResponse;
+import uk.gov.pay.connector.util.JsonObjectMapper;
+
+import java.util.List;
+
+import static java.lang.String.format;
+import static javax.ws.rs.core.Response.Status.Family.CLIENT_ERROR;
+import static javax.ws.rs.core.Response.Status.Family.SERVER_ERROR;
+import static org.apache.http.HttpStatus.SC_UNAUTHORIZED;
+
+public class StripeQueryPaymentStatusHandler {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StripeQueryPaymentStatusHandler.class);
+    private final GatewayClient client;
+    private final StripeGatewayConfig stripeGatewayConfig;
+    private JsonObjectMapper jsonObjectMapper;
+
+    public StripeQueryPaymentStatusHandler(GatewayClient client, StripeGatewayConfig stripeGatewayConfig, JsonObjectMapper jsonObjectMapper) {
+        this.client = client;
+        this.stripeGatewayConfig = stripeGatewayConfig;
+        this.jsonObjectMapper = jsonObjectMapper;
+    }
+
+    public ChargeQueryResponse queryPaymentStatus(ChargeQueryGatewayRequest chargeQueryGatewayRequest) throws GatewayException {
+        LOGGER.info(format("Querying Stripe payment status for [%s]", chargeQueryGatewayRequest.getChargeExternalId()));
+        StripeQueryPaymentStatusRequest request = StripeQueryPaymentStatusRequest.of(chargeQueryGatewayRequest.getGatewayAccount(),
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeQueryPaymentStatusHandler.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public StripeQueryPaymentStatusHandler(GatewayClient client, StripeGatewayConfig stripeGatewayConfig, JsonObjectMapper jsonObjectMapper) {
+        this.client = client;
+        this.stripeGatewayConfig = stripeGatewayConfig;
+        this.jsonObjectMapper = jsonObjectMapper;
+    }
+
+    public ChargeQueryResponse queryPaymentStatus(ChargeQueryGatewayRequest chargeQueryGatewayRequest) throws GatewayException {
+        MDC.put(PAYMENT_EXTERNAL_ID, chargeQueryGatewayRequest.getChargeExternalId());
+        MDC.put(SERVICE_EXTERNAL_ID, chargeQueryGatewayRequest.getGatewayAccount().getExternalId());
```
stephencdaly: I'm not sure about putting on and removing fields from the MDC here. I think we should add/remove from the MDC at the start and end of the scope where we have access to the charge external id.
Adding/removing here means we could end up removing a field from the MDC that was added earlier on in the scope before we want to.

stephencdaly: We could add to the MDC in `EpdqAuthorisationErrorGatewayCleanupService.sweepAndCleanupAuthorisationErrors` inside the `forEach`

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeSearchResponse.java
```diff
@@ -0,0 +1,28 @@
...
+    public boolean hasMore() {
+        return hasMore;
+    }
+
+    public List<StripePaymentIntent> getPaymentIntents() {
+        return paymentIntents;
+    }
+
+    public Optional<StripePaymentIntent> getPaymentIntent() {
+        return paymentIntents != null ? paymentIntents.stream().findFirst() : Optional.empty();
```
stephencdaly: Should we instead throw an exception and log an error to sentry if there is more than one payment intent returned? It should never happen, but would be a cause for concern

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeQueryPaymentStatusHandler.java
```diff
@@ -0,0 +1,92 @@
...
+        LOGGER.info(format("Querying Stripe payment status for [%s]", chargeQueryGatewayRequest.getChargeExternalId()));
+        StripeQueryPaymentStatusRequest request = StripeQueryPaymentStatusRequest.of(chargeQueryGatewayRequest.getGatewayAccount(),
+                stripeGatewayConfig, chargeQueryGatewayRequest.getGatewayCredentials(), chargeQueryGatewayRequest.getChargeExternalId());
+        try {
+            String rawResponse = client.getRequestFor(request).getEntity();
+            StripeSearchResponse queryResponse = jsonObjectMapper.getObject(rawResponse, StripeSearchResponse.class);
+            Optional<StripePaymentIntent> paymentIntent = queryResponse.getPaymentIntent();
+            return paymentIntent.map(payment -> {
+                if (payment.getStatus().equalsIgnoreCase("succeeded")) {
```
stephencdaly: I think for authorised, the status should be `requires_capture`. We might want to share the status mapping with `StripePaymentIntentResponse` so that this is only in one place.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeQueryPaymentStatusHandler.java
```diff
@@ -0,0 +1,92 @@
...
+                    LOGGER.info("Found payment intent with succeeded status");
+                    return new ChargeQueryResponse(ChargeStatus.AUTHORISATION_SUCCESS, new StripeQueryResponse(request.getChargeExternalId()));
+                }
+                if (payment.getStatus().equalsIgnoreCase("canceled")) {
+                    LOGGER.info("Found payment intent with canceled status");
+                    return new ChargeQueryResponse(ChargeStatus.AUTHORISATION_CANCELLED, new StripeQueryResponse(request.getChargeExternalId()));
+                }
```
stephencdaly: Would be good to map authorisation rejected status too. Looks like this will be `requires_payment_method`, as Stripe allows for a new payment method to be provided for a payment intent if authorisation fails.

stephencdaly: It might be good to map all the statuses we know how to map so this query method is reusable in other places. So, also mapping "succeeded" to `CAPTURED`.

stephencdaly (approved):
Looks good!

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripePaymentIntentResponse.java
```diff
@@ -3,15 +3,16 @@
...
 import java.util.Optional;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class StripePaymentIntentResponse {
-    private static Map<String, BaseAuthoriseResponse.AuthoriseStatus> statusMap = Map.of(
-            "requires_capture", BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED,
-            "requires_action", BaseAuthoriseResponse.AuthoriseStatus.REQUIRES_3DS
+    private static Map<StripeChargeStatus, BaseAuthoriseResponse.AuthoriseStatus> statusMap = Map.of(
+            StripeChargeStatus.REQUIRES_CAPTURE, BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED,
+            StripeChargeStatus.REQUIRES_ACTION, BaseAuthoriseResponse.AuthoriseStatus.REQUIRES_3DS
```
stephencdaly, line 15: does this need to exist anymore? can we just use the mapper from StripeChargeStatus?

SandorArpa: I think it's a special case. This piece of code only want to know about these 2 AuthoriseStatus. anything else treated as null

PR #3748, opened 2022-05-31
PP-9599 refactor gateway clean up service

SandorArpa:
## WHAT YOU DID
- refactor `EpdqAuthorisationErrorGatewayCleanupService` to handle other payment providers as well
- refactor/update tests


src/main/java/uk/gov/pay/connector/charge/service/AuthorisationErrorGatewayCleanupService.java
```diff
@@ -43,29 +49,37 @@ public class EpdqAuthorisationErrorGatewayCleanupService {
...
+
+        logger.info("Found {} charges to clean up.", chargesToCleanUp.size());
         
         AtomicInteger successes = new AtomicInteger();
         AtomicInteger failures = new AtomicInteger();
 
         chargesToCleanUp.forEach(chargeEntity -> {
+            MDC.put(PROVIDER, chargeEntity.getPaymentProvider());
+            MDC.put(GATEWAY_ACCOUNT_ID, chargeEntity.getGatewayAccount().getId().toString());
+            MDC.put(PAYMENT_EXTERNAL_ID, chargeEntity.getExternalId());
```
stephencdaly, line 81: We could use `MDCUtils.addChargeAndGatewayAccountDetailsToMDC` to add all the relevant fields to the MDC to make sure we're adding them uniformly.

src/main/java/uk/gov/pay/connector/charge/service/AuthorisationErrorGatewayCleanupService.java
```diff
@@ -78,10 +92,14 @@ public Map<String, Integer> sweepAndCleanupAuthorisationErrors(int limit) {
                 logger.info("Error when querying charge status with gateway: " + e.getMessage(),
                         chargeEntity.getStructuredLoggingArgs());
                 failures.getAndIncrement();
+            } finally {
+                MDC.remove(PROVIDER);
+                MDC.remove(GATEWAY_ACCOUNT_ID);
+                MDC.remove(PAYMENT_EXTERNAL_ID);
```
stephencdaly, line 98: If we use `MDCUtils` to add to the MDC as suggested in my other comment, we could add a method to that class to remove all the relevant MDC fields too. The reason why this doesn't already exist is because for other cases where `MDCUtils` is used, it is in the scope of an API request and all the fields are removed by `LoggingMDCResponseFilter`

stephencdaly (approved):
Looks good to me! Small suggestion about use of MDC but change is good to go in.

PR #3764, opened 2022-06-07
PP-9599 Fix searching Stripe payment intents

stephencdaly:
Alter GatewayClient#getRequestFor to accept a different request
interface named GatewayClientGetRequest, which has a getQueryParams()
method. If there are any query params available, add them to the target
for the request.

StripeQueryPaymentStatusRequest now implements GatewayClientGetRequest,
and provides the metadata value to search by as query params.

The only other GET request we make is to GET a single payment intent
from Stripe, so also changed StripeGetPaymentIntentRequest to implement
GatewayClientGetRequest.

Added StripePaymentProviderIT with a test to check that we successfully
query Stripe with the expected query parameters and map the response.

Rename StripeRequest -> StripePostRequest and
GatewayClientRequest -> GatewayClientPostRequest to reflect the fact
that we have added separate interfaces for the GET requests.


SandorArpa (approved):

PR #3766, opened 2022-06-07
PP-9599 Use newer Stripe API version for searching payment intents

stephencdaly:
The Stripe `/v1/payment_intents/search` is only supported in version
2020-08-27 of their API. Currently in connector we use 2019-05-16.

Specify the version 2020-08-27 in the request headers only for search
payment intent requests to get this working. Keep the version used for
all other requests the same for now and we will look into upgrading
separately.


SandorArpa (approved):

PR #3772, opened 2022-06-09
PP-9644 make token expiry date configurable

nlsteers:
## WHAT YOU DID

Added a configuration option for token expiry, allowing for more a aggressive deletion policy in the performance testing environment

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/ChargeSweepConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/ChargeSweepConfig.java
@@ -2,16 +2,23 @@ package uk.gov.pay.connector.app;
 
 import io.dropwizard.Configuration;
 
+import java.time.Duration;
+
 public class ChargeSweepConfig extends Configuration {
 
     private int defaultChargeExpiryThreshold;
     private int awaitingCaptureExpiryThreshold;
+    private int tokenExpiryThresholdInSeconds;
 
-    public int getDefaultChargeExpiryThreshold() {
-        return defaultChargeExpiryThreshold;
+    public Duration getDefaultChargeExpiryThreshold() {
+        return Duration.ofSeconds(defaultChargeExpiryThreshold);
     }
 
-    public int getAwaitingCaptureExpiryThreshold() {
-        return awaitingCaptureExpiryThreshold;
+    public Duration getAwaitingCaptureExpiryThreshold() {
+        return Duration.ofSeconds(awaitingCaptureExpiryThreshold);
+    }
+    
+    public Duration getTokenExpiryThresholdInSeconds() {
+        return Duration.ofSeconds(tokenExpiryThresholdInSeconds);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
@@ -31,9 +31,10 @@ import uk.gov.pay.connector.token.dao.TokenDao;
 
 import javax.inject.Inject;
 import javax.ws.rs.WebApplicationException;
+import java.time.Clock;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneId;
-import java.time.ZonedDateTime;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
@@ -58,18 +59,15 @@ import static uk.gov.service.payments.logging.LoggingKeys.PROVIDER;
 public class ChargeExpiryService {
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
-
     private static final String EXPIRY_SUCCESS = "expiry-success";
     private static final String EXPIRY_FAILED = "expiry-failed";
-    private static final long TOKEN_EXPIRY_DAYS = 7;
-
     private final ChargeDao chargeDao;
     private final ChargeService chargeService;
     private final TokenDao tokenDao;
     private final PaymentProviders providers;
     private final QueryService queryService;
-
     private final ChargeSweepConfig chargeSweepConfig;
+    private final Clock clock;
 
     @Inject
     public ChargeExpiryService(ChargeDao chargeDao,
@@ -77,13 +75,15 @@ public class ChargeExpiryService {
                                TokenDao tokenDao,
                                PaymentProviders providers,
                                QueryService queryService,
-                               ConnectorConfiguration config) {
+                               ConnectorConfiguration config,
+                               Clock clock) {
         this.chargeDao = chargeDao;
         this.chargeService = chargeService;
         this.tokenDao = tokenDao;
         this.providers = providers;
         this.chargeSweepConfig = config.getChargeSweepConfig();
         this.queryService = queryService;
+        this.clock = clock;
     }
 
     private enum expiryMethod {
@@ -133,17 +133,17 @@ public class ChargeExpiryService {
                 .addAll(getChargesToExpireWithRegularExpiryThreshold())
                 .addAll(getChargesToExpireWithDelayedExpiryThreshold())
                 .build();
-
-        deleteTokensOlderThanSpecifiedDate();
-        logger.info("Charges found for expiry - number_of_charges={}, since_date={}, awaiting_capture_date{}",
+        Instant tokenExpiryThreshold = getExpiryThresholdForTokens();
+        int numberOfTokensDeleted = deleteTokensOlderThanSpecifiedDate(tokenExpiryThreshold);
+        logger.info("Tokens deleted - number_of_tokens={}, since_date={}", numberOfTokensDeleted, tokenExpiryThreshold);
+        logger.info("Charges found for expiry - number_of_charges={}, since_date={}, awaiting_capture_date={}",
                 chargesToExpire.size(), getExpiryDateForRegularCharges(), getExpiryDateForAwaitingCaptureRequest());
 
         return expire(chargesToExpire);
     }
 
-    private int deleteTokensOlderThanSpecifiedDate() {
-        ZonedDateTime cutOffDate = ZonedDateTime.now(ZoneId.of("UTC")).minusDays(TOKEN_EXPIRY_DAYS);
-        return tokenDao.deleteTokensOlderThanSpecifiedDate(cutOffDate);
+    private int deleteTokensOlderThanSpecifiedDate(Instant tokenExpiryDate) {
+        return tokenDao.deleteTokensOlderThanSpecifiedDate(tokenExpiryDate.atZone(ZoneId.of("UTC")));
     }
 
 
@@ -319,15 +319,21 @@ public class ChargeExpiryService {
     }
 
     private Instant getExpiryDateForRegularCharges() {
-        int chargeExpiryWindowSeconds = chargeSweepConfig.getDefaultChargeExpiryThreshold();
-        logger.debug("Charge expiry window size in seconds: [{}]", chargeExpiryWindowSeconds);
-        return Instant.now().minusSeconds(chargeExpiryWindowSeconds);
+        Duration chargeExpiryWindowSeconds = chargeSweepConfig.getDefaultChargeExpiryThreshold();
+        logger.debug("Charge expiry window size in seconds: [{}]", chargeExpiryWindowSeconds.getSeconds());
+        return clock.instant().minus(chargeExpiryWindowSeconds);
     }
 
     private Instant getExpiryDateForAwaitingCaptureRequest() {
-        int chargeExpiryWindowSeconds = chargeSweepConfig.getAwaitingCaptureExpiryThreshold();
-        logger.debug("Charge expiry window size for awaiting_delay_capture in seconds: [{}]", chargeExpiryWindowSeconds);
-        return Instant.now().minusSeconds(chargeExpiryWindowSeconds);
+        Duration chargeExpiryWindowSeconds = chargeSweepConfig.getAwaitingCaptureExpiryThreshold();
+        logger.debug("Charge expiry window size for awaiting_delay_capture in seconds: [{}]", chargeExpiryWindowSeconds.getSeconds());
+        return clock.instant().minus(chargeExpiryWindowSeconds);
+    }
+    
+    private Instant getExpiryThresholdForTokens() {
+        Duration tokenExpiryWindowSeconds = chargeSweepConfig.getTokenExpiryThresholdInSeconds();
+        logger.debug("Token expiry window size in seconds: [{}]", tokenExpiryWindowSeconds.getSeconds());
+        return clock.instant().minus(tokenExpiryWindowSeconds);
     }
 
     private static String getLegalStatusNames(List<ChargeStatus> legalStatuses) {
--- a/src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
@@ -1,5 +1,10 @@
 package uk.gov.pay.connector.charge.service;
 
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.classic.spi.LoggingEvent;
+import ch.qos.logback.core.Appender;
 import com.google.common.collect.ImmutableList;
 import junitparams.JUnitParamsRunner;
 import junitparams.Parameters;
@@ -8,9 +13,11 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
+import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ChargeSweepConfig;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.charge.dao.ChargeDao;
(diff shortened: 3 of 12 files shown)


src/main/java/uk/gov/pay/connector/app/ChargeSweepConfig.java
```diff
@@ -14,4 +15,8 @@ public int getDefaultChargeExpiryThreshold() {
     public int getAwaitingCaptureExpiryThreshold() {
         return awaitingCaptureExpiryThreshold;
     }
+    
+    public int getTokenExpiryDays() {
+        return tokenExpiryDays;
+    }
```
alexbishop1: Make this return a [`Duration`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/Duration.html): it’s a more appropriate type and it will probably make the calculation code simpler.

```java
public Duration getTokenExpiry() {
    return Duration.ofDays(tokenExpiryDays);
}
```

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -133,17 +134,17 @@ public Map<String, Integer> sweepAndExpireChargesAndTokens() {
                 .addAll(getChargesToExpireWithRegularExpiryThreshold())
                 .addAll(getChargesToExpireWithDelayedExpiryThreshold())
                 .build();
-
-        deleteTokensOlderThanSpecifiedDate();
-        logger.info("Charges found for expiry - number_of_charges={}, since_date={}, awaiting_capture_date{}",
+        ZonedDateTime tokenExpiryDate = getExpiryDateForTokens();
```
alexbishop1: Make this an `Instant` rather than a `ZonedDateTime`.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -133,17 +134,17 @@ public Map<String, Integer> sweepAndExpireChargesAndTokens() {
...
                 chargesToExpire.size(), getExpiryDateForRegularCharges(), getExpiryDateForAwaitingCaptureRequest());
 
         return expire(chargesToExpire);
     }
 
-    private int deleteTokensOlderThanSpecifiedDate() {
-        ZonedDateTime cutOffDate = ZonedDateTime.now(ZoneId.of("UTC")).minusDays(TOKEN_EXPIRY_DAYS);
-        return tokenDao.deleteTokensOlderThanSpecifiedDate(cutOffDate);
+    private int deleteTokensOlderThanSpecifiedDate(ZonedDateTime tokenExpiryDate) {
+        return tokenDao.deleteTokensOlderThanSpecifiedDate(tokenExpiryDate);
```
alexbishop1: This can take an `Instant` and convert it to a `ZonedDateTime` before passing to `TokenDao`.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -321,13 +322,19 @@ private Boolean handleCancelResponse(String chargeExternalId, GatewayResponse<Ba
...
     }
 
     private Instant getExpiryDateForAwaitingCaptureRequest() {
         int chargeExpiryWindowSeconds = chargeSweepConfig.getAwaitingCaptureExpiryThreshold();
         logger.debug("Charge expiry window size for awaiting_delay_capture in seconds: [{}]", chargeExpiryWindowSeconds);
-        return Instant.now().minusSeconds(chargeExpiryWindowSeconds);
+        return clock.instant().minusSeconds(chargeExpiryWindowSeconds);
+    }
+    
+    private ZonedDateTime getExpiryDateForTokens() {
```
alexbishop1: This can return an `Instant` rather than a `ZonedDateTime`.

alexbishop1 (changes requested):
Logic and tests look fine but prefer to use `Durations` rather than `int`s and `Instant`s rather than `ZonedDateTime`s. (I have an entire presentation on the latter. No, really. I’ll share it with you.)

src/main/java/uk/gov/pay/connector/app/ChargeSweepConfig.java
nlsteers: Done, also changed to seconds as that's how the other thresholds are measured

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
nlsteers: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
nlsteers: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
nlsteers: Done.

src/main/java/uk/gov/pay/connector/app/ChargeSweepConfig.java
```diff
@@ -2,16 +2,23 @@
 
 import io.dropwizard.Configuration;
 
+import java.time.Duration;
+
 public class ChargeSweepConfig extends Configuration {
 
     private int defaultChargeExpiryThreshold;
     private int awaitingCaptureExpiryThreshold;
+    private int tokenExpiryThreshold;
```
alexbishop1: I think I’d prefer this to be called `tokenExpiryThresholdInSeconds` or sometging (inconsistent with the others but it can serve as an example to them).

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -319,15 +319,21 @@ private Boolean handleCancelResponse(String chargeExternalId, GatewayResponse<Ba
...
+        Duration chargeExpiryWindowSeconds = chargeSweepConfig.getDefaultChargeExpiryThreshold();
         logger.debug("Charge expiry window size in seconds: [{}]", chargeExpiryWindowSeconds);
-        return Instant.now().minusSeconds(chargeExpiryWindowSeconds);
+        return clock.instant().minus(chargeExpiryWindowSeconds);
     }
 
     private Instant getExpiryDateForAwaitingCaptureRequest() {
-        int chargeExpiryWindowSeconds = chargeSweepConfig.getAwaitingCaptureExpiryThreshold();
+        Duration chargeExpiryWindowSeconds = chargeSweepConfig.getAwaitingCaptureExpiryThreshold();
         logger.debug("Charge expiry window size for awaiting_delay_capture in seconds: [{}]", chargeExpiryWindowSeconds);
```
alexbishop1: How is `chargeExpiryWindowSeconds` going to be output in the logs now it’s a `Duration` rather than `int`?

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -109,13 +131,17 @@ public class ChargeExpiryServiceTest {
             AWAITING_CAPTURE_REQUEST
     );
 
+    private static final Duration TOKEN_EXPIRY_WINDOW_SECONDS = Duration.ofSeconds(604800);
```
alexbishop1: You could do something like `Duration.ofSeconds(7 * 24 * 60 * 60)` if you want to make it less mysterious.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -360,12 +393,21 @@ public void shouldSweepAndExpireCharges() throws Exception {
         when(mockPaymentProviders.byName(PaymentGatewayName.WORLDPAY)).thenReturn(mockPaymentProvider);
         when(mockChargeDao.findBeforeDateWithStatusIn(any(Instant.class), eq(EXPIRABLE_AWAITING_CAPTURE_REQUEST_STATUS))).thenReturn(singletonList(chargeEntityAwaitingCapture));
         when(mockChargeDao.findBeforeDateWithStatusIn(any(Instant.class), eq(EXPIRABLE_REGULAR_STATUSES))).thenReturn(singletonList(chargeEntityAuthorisationSuccess));
-
+        when(mockedChargeSweepConfig.getTokenExpiryThreshold()).thenReturn(TOKEN_EXPIRY_WINDOW_SECONDS);
+        when(mockedChargeSweepConfig.getDefaultChargeExpiryThreshold()).thenReturn(CHARGE_EXPIRY_WINDOW_SECONDS);
+        when(mockedChargeSweepConfig.getAwaitingCaptureExpiryThreshold()).thenReturn(AWAITING_DELAY_CAPTURE_EXPIRY_WINDOW);
+        when(mockTokenDao.deleteTokensOlderThanSpecifiedDate(any(ZonedDateTime.class))).thenReturn(1);
```
alexbishop1: Since we have a fixed clock, we should know the exact time that we’re expecting to be passed in.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -360,12 +393,21 @@ public void shouldSweepAndExpireCharges() throws Exception {
...
         when(mockChargeService.transitionChargeState(any(String.class), any())).thenReturn(expiredCharge);
 
         chargeExpiryService.sweepAndExpireChargesAndTokens();
-
+        verify(mockAppender, times(4)).doAppend(loggingEventArgumentCaptor.capture());
+        List<LoggingEvent> loggingEvents = loggingEventArgumentCaptor.getAllValues();
+        assertThat(loggingEvents.stream().map(LoggingEvent::getFormattedMessage).collect(Collectors.toList()),
+                hasItems(
+                        format("Tokens deleted - number_of_tokens=1, since_date=%s", expectedTokenExpiryDate),
+                        format("Charges found for expiry - number_of_charges=2, since_date=%s, awaiting_capture_date=%s", expectedChargeExpiryDate, expectedChargeAwaitingCaptureExpiryDate)
```
alexbishop1: Rather than using string formatting, we could have fixed strings here, which would make sure the timestamps are being output as we expect.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
```diff
@@ -70,7 +71,7 @@ public class Worldpay3dsFlexJwtServiceTest {
     @Rule
     public ExpectedException expectedException = ExpectedException.none();
     
-    private static final int TOKEN_EXPIRY_DURATION_SECONDS = 5400;
+    private static final Duration TOKEN_EXPIRY_DURATION_SECONDS = Duration.ofSeconds(5400);
```
alexbishop1: This name doesn’t need `SECONDS` in it any more.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
```diff
@@ -99,9 +100,9 @@ public void setUp() {
     public void shouldCreateCorrectTokenForWorldpay3dsFlexDdc() {
         var gatewayAccount = new GatewayAccount(1L, WORLDPAY.getName(), TEST);
         var worldpay3dsFlexCredentials = new Worldpay3dsFlexCredentials("me", "myOrg", "fa2daee2-1fbb-45ff-4444-52805d5cd9e0", false);
-        int paymentCreationTimeEpochSeconds19August2029 = 1881821916;
-        int expectedTokenExpirationTimeEpochSeconds = paymentCreationTimeEpochSeconds19August2029 + TOKEN_EXPIRY_DURATION_SECONDS;
-        var paymentCreationInstant = Instant.ofEpochSecond(paymentCreationTimeEpochSeconds19August2029);
+        Duration paymentCreationTimeEpochSeconds19August2029 = Duration.ofSeconds(1881821916);
+        Duration expectedTokenExpirationTimeEpochSeconds = paymentCreationTimeEpochSeconds19August2029.plus(TOKEN_EXPIRY_DURATION_SECONDS);
```
alexbishop1: I don’t think these represent durations: they look like Unix timestamps.

alexbishop1 (changes requested):
I didn’t realise you were going to change all the other expiry stuff to `Duration`s!

There’s a couple of things I think aren’t quite right with the logging and the tests.

src/main/java/uk/gov/pay/connector/app/ChargeSweepConfig.java
nlsteers: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
nlsteers: Good catch, fixed.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
nlsteers: Done.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
nlsteers: Updated.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
nlsteers: No worries, have fixed the clock to a specific moment in time and updated.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
nlsteers: Yep, fixed.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
nlsteers: Done.

src/main/java/uk/gov/pay/connector/charge/service/ChargeExpiryService.java
```diff
@@ -319,15 +319,21 @@ private Boolean handleCancelResponse(String chargeExternalId, GatewayResponse<Ba
     }
 
     private Instant getExpiryDateForRegularCharges() {
-        int chargeExpiryWindowSeconds = chargeSweepConfig.getDefaultChargeExpiryThreshold();
-        logger.debug("Charge expiry window size in seconds: [{}]", chargeExpiryWindowSeconds);
-        return Instant.now().minusSeconds(chargeExpiryWindowSeconds);
+        Duration chargeExpiryWindowSeconds = chargeSweepConfig.getDefaultChargeExpiryThreshold();
```
alexbishop1, line 322: Minor but this variable and similar don’t really need `Seconds` in their names any more.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -66,6 +79,8 @@ public class ChargeExpiryServiceTest {
 
     private ChargeExpiryService chargeExpiryService;
 
+    private final Clock fixedClock = Clock.fixed(Instant.ofEpochSecond(1654732800), ZoneId.of("UTC")); // Thu 9 June 2022 00:00:00
```
alexbishop1, line 82: You could also do `Instant.parse("2022-06-09T00:00:00Z")`.

src/test/java/uk/gov/pay/connector/charge/service/ChargeExpiryServiceTest.java
```diff
@@ -109,13 +130,17 @@ public class ChargeExpiryServiceTest {
             AWAITING_CAPTURE_REQUEST
     );
 
+    private static final Duration TOKEN_EXPIRY_WINDOW = Duration.ofSeconds(7 * 24 * 60 * 60);
+    private static final Duration CHARGE_EXPIRY_WINDOW = Duration.ofSeconds((long) (1.5 * 60 * 60));
```
alexbishop1, line 134: Since floating point arithmetic scares me, I’d probably do `90 * 60`.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
```diff
@@ -113,7 +112,7 @@ public void shouldCreateCorrectTokenForWorldpay3dsFlexDdc() {
         assertThat((Map<String, Object>)jws.getHeader(), hasEntry("typ", "JWT"));
         assertThat(jws.getBody(), hasKey("jti"));
         assertThat(jws.getBody(), hasKey("iat"));
-        assertThat(jws.getBody(), hasEntry("exp", expectedTokenExpirationTimeEpochSeconds));
+        assertThat(jws.getBody(), hasEntry("exp", (int) expectedTokenExpirationTimeEpochSeconds.getEpochSecond()));
```
alexbishop1, line 115: Slightly surprised you need the cast here.

alexbishop1 (approved):
Awesome, looks good.

src/test/java/uk/gov/pay/connector/charge/service/Worldpay3dsFlexJwtServiceTest.java
nlsteers: The assertion fails on account of the postfixed `L` without the cast. `getEpochSecond()` returns a long

PR #3774, opened 2022-06-09
PP-9587: Add disabled property to gateway account

alan-gds:
## WHAT YOU DID
Add disabled and disabled_reason properties to gateway account,
make it appear in JSON responses, make it updatable through the
existing gateway account update API.


stephencdaly (approved):

PR #3783, opened 2022-06-13
PP-9587 Clear disabled reason when account is re-enabled

stephencdaly:
When a PATCH request is made to update a gateway account to set "disabled" to false, clear the stored "disabled_reason".


alan-gds (approved):

PR #3792, opened 2022-06-15
PP-9587: Change status code returned for disabled accounts

alan-gds:
Return a 403, not a 422, for attempts to create charge/refunds on disabled accounts


stephencdaly (approved):

PR #3798, opened 2022-06-17
PP-9587 Log when a gateway account is disabled

stephencdaly:
Log when a gateway account is disabled or re-enabled so we can audit when this is done in Splunk.


SandorArpa (approved):
LGTM

PR #3803, opened 2022-06-20
BAU fix bad logline

nlsteers:
## WHAT YOU DID

Fixed an issue in the auth error gateway cleanup service where some values were being logged in the wrong order


stephencdaly (approved):

SandorArpa (approved):
LOKTM

PR #3816, opened 2022-06-23
PP-9785 Expunge telephone payments in authorisation error states

kbottla:
## WHAT 
- Payments in authorisation error states (AUTHORISATION_ERROR, AUTHORISATION_TIMEOUT, AUTHORISATION_UNEXPECTED_ERROR) are currently **not** expunged immediately. These are processed by the gateway cleanup sweep task which queries and then tries to cancel any payments (if authorised with gateway) and moves charges to a different state.
- External payments errored and reported to us end up in the AUTHORISATION_ERROR state and so are not expunged too. We should just
  expunge these payments. The `provider_id` (gateway_transaction_id) reported for these payments is not exactly the orderCode used for Worldpay


SandorArpa (approved):
LGTM

PR #3820, opened 2022-06-24
PP-9785 Do not throw exceptions for Worldpay order not found query response

kbottla:
## WHAT

- Worldpay connection timed out payments end up in authorisation error states and the requests may never have reached Worldpay. When various tasks (gateway cleanup sweep, discrepancy checker) query Worldpay for the transaction, an error response is returned with code `5` and the message `Could not find payment for order`.
- Currently, WebApplicationException is thrown by the Worldpay query payment method and is causing 500 errors for the discrepancy checker. Gateway cleanup service catches exceptions and continues onto the next payment (without updating charge status). So auth errored charges (for which is there is no Worldpay order) are being retried continuously and new charges are not being cleaned up (as the gateway clean service is attempting to query the same charges for every invoke and also has a limit of 100 charges)
- So returning `ChargeQueryResponse` when Worldpay returns `Could not find payment for order` error.
    - The discrepancy checker will return the message and is displayed in Toolbox
    - Gateway cleanup service will update charge status to `AUTHORISATION_ERROR_CHARGE_MISSING` (which will be expunged by connector)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -71,6 +71,14 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
     static final String WORLDPAY_MACHINE_COOKIE_NAME = "machine";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(WorldpayPaymentProvider.class);
+    private static final String WORLDPAY_ORDER_INQUIRY_ERROR_CODE_FOR_INVALID_CONTENT = "error code: 5";
+    /**
+     * When a payment is not found for the order inquiry, Worldpay returns an error code ‘5’ with the error
+     * message 'Could not find payment for order'.
+     * See https://developer.worldpay.com/docs/wpg/manage/inquiryrequests for details (Getting an error? heading).
+     * For other errors (ex: service unavailable), Worldpay returns a different error code (8) or a different error message
+     */
+    private static final String WORLDPAY_ORDER_INQUIRY_PAYMENT_NOT_FOUND_ERROR_MESSAGE = "Could not find payment for order";
 
     private final GatewayClient authoriseClient;
     private final GatewayClient cancelClient;
@@ -148,6 +156,20 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
 
                     return new ChargeQueryResponse(mappedStatus, worldpayQueryResponse);
                 })
+                .or(() -> {
+                            return worldpayGatewayResponse.getGatewayError()
+                                    .map(gatewayError -> {
+                                                if (gatewayError.getMessage() != null
+                                                        && gatewayError.getMessage().contains(WORLDPAY_ORDER_INQUIRY_ERROR_CODE_FOR_INVALID_CONTENT)
+                                                        && gatewayError.getMessage().contains(WORLDPAY_ORDER_INQUIRY_PAYMENT_NOT_FOUND_ERROR_MESSAGE)
+                                                ) {
+                                                    return new ChargeQueryResponse(gatewayError);
+                                                }
+                                                return null;
+                                            }
+                                    );
+                        }
+                )
                 .orElseThrow(() ->
                         new WebApplicationException(format(
                                 "Unable to query charge %s - an error occurred: %s",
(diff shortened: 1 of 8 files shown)


alexbishop1:
What happens when Worldpay’s order inquiry endpoint is down (which happens quite often)? In those cases retrying again later can actually be useful.

kbottla:
@alexbishop1 I am expecting a different error message than what is handled here https://github.com/alphagov/pay-connector/blob/60b70545a83703c7826b983154081e4bce59fd52/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java#L155 when Worldpay inquiry endpoint is down. 

It would be ideal to have a retrying mechanism and ignore any charges tried n number of times. But that is a bigger piece of work we can think of separately.

alexbishop1:
Yeah, I agree it’s a bigger piece of work. I can’t remember what the standard failure mode is for when the order inquiry endpoint is down — does it do something different to claiming it can’t find the order?

kbottla:
I see `error code: 8, error: Order inquiries have been disabled for all merchants` in the logs for the maintenance window  `12 August 2021 18:00 (UTC) - 21:00 (UTC)`. 

This lines up with Worldpay docs https://developer.worldpay.com/docs/wpg/troubleshoot

Error code no. | Error type | Comments
-- | -- | --
8 | Service not available. | The service is currently unavailable.

alexbishop1:
OK, cool, so this is sufficiently narrowly-targeted.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -148,6 +148,18 @@ public ChargeQueryResponse queryPaymentStatus(ChargeQueryGatewayRequest chargeQu
 
                     return new ChargeQueryResponse(mappedStatus, worldpayQueryResponse);
                 })
+                .or(() -> {
+                            return worldpayGatewayResponse.getGatewayError()
+                                    .map(gatewayError -> {
+                                                if (gatewayError.getMessage() != null
+                                                        && gatewayError.getMessage().contains("Could not find payment for order")) {
```
alexbishop1: Can we…

- … also check the code is 5?
- … extract the error message into a constant and put a comment linking to some appropriate Worldpay documentation (it seems https://developer.worldpay.com/docs/wpg/manage/inquiryrequests is the best available — it describes the string under the ‘Getting an error?’ expand-o-matic)?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -580,14 +582,53 @@ void should_successfully_query_payment_status() throws Exception {
...
+        assertThat(chargeQueryResponse.foundCharge(), is(false));
+        assertThat(chargeQueryResponse.getGatewayError().isPresent(), is(true));
+        assertThat(chargeQueryResponse.getGatewayError().get().getMessage(), is("Worldpay query response (error code: 5, error: Could not find payment for order)"));
+    }
+
+    @Test
+    void query_payment_status_should_throw_error_when_worldpay_returns_error_with_unknown_message() throws Exception {
+        when(response.getEntity()).thenReturn(
+                load(WORLDPAY_ERROR_RESPONSE)
+                        .replace("{{errorDescription}}", "some error message")
```
alexbishop1: We could maybe use the specific message we expect when the order inquiry endpoint is down.

alexbishop1 (changes requested):
Looks fine. Just a couple of things that may help our future selves.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
kbottla: Updated in https://github.com/alphagov/pay-connector/compare/60b70545a83703c7826b983154081e4bce59fd52..063f8416eda0e04efa6f707f103933d8b8de273f

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
kbottla: Updated https://github.com/alphagov/pay-connector/compare/60b70545a83703c7826b983154081e4bce59fd52..063f8416eda0e04efa6f707f103933d8b8de273f

alexbishop1 (approved):

PR #3826, opened 2022-06-28
PP-9796 OpenAPI specs for card types endpoint

kbottla:
## WHAT YOU DID
- OpenAPI specs for card types endpoint

### DIFF
--- a/openapi/connector_spec.yaml
+++ b/openapi/connector_spec.yaml
@@ -36,6 +36,17 @@ paths:
                 $ref: '#/components/schemas/ErrorResponse'
           description: Missing required fields
       summary: Create an agreement
+  /v1/api/card-types:
+    get:
+      operationId: getCardTypes
+      responses:
+        "200":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/CardTypesResponse'
+          description: OK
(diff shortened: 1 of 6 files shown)


openapi/connector_spec.yaml
```diff
@@ -77,6 +88,34 @@ components:
...
+          example: ac8a3abd-bcfd-4fa6-8905-321ce913e7f5
+        label:
+          type: string
+          example: Visa
+        requires3ds:
+          type: boolean
+        type:
+          type: string
+          enum:
+          - CREDIT
```
SandorArpa, line 109: I think we can also have `CREDIT_OR_DEBIT` or null, if `CREDIT_OR_DEBIT`

SandorArpa (changes requested):

openapi/connector_spec.yaml
kbottla: `/v1/api/card-types` endpoint only returns list of card types (for admin tool listing). It is only CREDIT or DEBIT currently. On charges we can have `null` for the card type.

SandorArpa: 👍

SandorArpa (approved):
LGTM

PR #3838, opened 2022-07-01
PP-9796 Fix flaky card resource test

kbottla:
## WHAT YOU DID
- Validation for expiry date checks if the month and date are current or in the future and offsets date to -12 hours to account for timezones.
  However, for the current date & time of 1-Jul-22 10:00 AM, the expiry date `06/22` will be valid due to offset (-12 hours). This is a valid case.

- So fixed test checking for `expiry date one month before current month and year' as it will fail on specific dates/time.


SandorArpa (approved):
LGTM

PR #3845, opened 2022-07-04
PP-9796 Generate OpenAPI specs for all resources

kbottla:
## WHAT YOU DID
- Added all resources to openapi-specs.yaml to generate OpenAPI specs for all repos


stephencdaly (approved):
Too much to check everything, but looks correct from a brief scan

PR #3846, opened 2022-07-04
PP-9796 Remove API specs markdown

kbottla:
## WHAT YOU DID
- Removed `docs/api_specification.md` file as we now have OpenAPI specs for the same (https://github.com/alphagov/pay-connector/blob/master/openapi/connector_spec.yaml).


SandorArpa (approved):
LGTM

PR #3850, opened 2022-07-05
PP-9785 Gateway cleanup sweep - process WEB or MOTO_API payments only

kbottla:
## WHAT YOU DID
- As part of the gateway clean sweep we should only check payments with authorisation mode `WEB` and `MOTO_API` with the gateway and cancel payments if necessary.
  Currently, the task attemtps to cancel `EXTERNAL` changes (telephone payments reported) or charges in any other modes (AGREEMENT).


SandorArpa (approved):
LGTM

PR #3852, opened 2022-07-06
PP-9785 Emit FeeIncurred event only when there is a fee

kbottla:
## WHAT YOU DID
- FeeIncurred event is emitted all the time and causes a warning (https://github.com/alphagov/pay-connector/blob/764bcd716d29f0adf56459785df9b2448522dbd2/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardCaptureService.java#L115) to be logged for all payments (other than stripe) as there won't be any fee available in the capture response.
- Updated to emit FeeIncurred event only when there is a fee list on capture a response.


SandorArpa (approved):
LGTM

PR #3854, opened 2022-07-07
PP-9622 DISPUTE_CREATED event: generate resource_external_id

kbottla:
## WHAT YOU DID
- Generate resource_external_id based on Stripe dispute ID for dispute created event.
  We do it already for other dispute events (dispute_lost, dispute_won, dispute_evidence_submitted)


SandorArpa (approved):
LGTM

PR #3855, opened 2022-07-07
PP-9622 Create dispute objects using static methods

kbottla:
## WHAT YOU DID
- Create dispute event objects using static methods which can be responsible for constructing relevant event details and returning a new dispute event object.
-  StripeWebhookTaskHandler to use these static methods instead of constructing different dispute event objects on its own.


SandorArpa (approved):
LGTM

PR #3856, opened 2022-07-07
PP-9622 Refactor StripeWebhookTaskHandler

kbottla:
## WHAT YOU DID
- Refactored StripeWebhookTaskHandler
   - uses static methods on dispute event classes to create objects
   - extracted common code (get a transaction, deserialise) to separate methods and called once and before a `dispute` event is processed
- Stops throwing exception if there are multiple balance transactions on Stripe dispute
  - Multiple balance_transactions are possible if a dispute is won (one transaction to debit the amount and another to credit it back). One balance transaction can only be excepted for dispute_created or dispute_lost event so the exception is thrown when building related events.


SandorArpa (approved):
LOKTM

PR #3865, opened 2022-07-11
PP-9795 investigate card exec service thread pool exhaustion

nlsteers:
## WHAT YOU DID

During a period of degraded Worldpay service we observed Connector keeping authorisation tasks active for as long as 50 minutes. This was causing some Connector instances to completely exhaust their thread pools and trigger an `unhealthy` warning on fargate, leading to the node being rebooted.

It is still not obvious why some of these tasks were so long lived as there are a number of configured timeouts for network connections to PSPs that should terminate any active sessions long before some of the observed run times we were seeing.

To prevent this from happening in the future, the `CardExecutorService` has been updated to use a timed thread pool executor. This allows us to schedule an additional task alongside any card operation task that will interrupt the thread after a specified time has elapsed. If the task has completed successfully the cancel task is removed and work continues as normal, meaning the added overhead of an additional task per execution is minimal and is handled by a separate executor.

Additionally, the healthcheck has been updated to dump some statistics on the state of the threadpool when the check returns `unhealthy`.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java
@@ -10,6 +10,7 @@ import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.app.ExecutorServiceConfig;
+import uk.gov.pay.connector.util.TimedThreadPoolExecutor;
 
 import javax.ws.rs.WebApplicationException;
 import java.util.Map;
@@ -17,7 +18,6 @@ import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
@@ -42,9 +42,7 @@ public class CardExecutorService {
     private static final int QUEUE_WAIT_WARN_THRESHOLD_MILLIS = 1000;
     public static final int SHUTDOWN_AWAIT_TERMINATION_TIMEOUT_SECONDS = 10;
     private final MetricRegistry metricRegistry;
-
-    private ExecutorServiceConfig config;
-    private ExecutorService executor;
+    private final TimedThreadPoolExecutor timedExectuor;
 
     public enum ExecutionStatus {
         COMPLETED,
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/util/TimedThreadPoolExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /***
+     * pre execution hook that schedules a timeout task at the point the task starts executing
+     * @param t the thread that will run task {@code r}
+     * @param r the task that will be executed
+     */
+    @Override
+    protected void beforeExecute(Thread t, Runnable r) {
+        if(timeout > 0) {
```
oswaldquek: What should the startup behaviour of the app be if timeout if less than or equal to 0? Maybe a warn in the log?

src/main/java/uk/gov/pay/connector/util/TimedThreadPoolExecutor.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+    
+    protected ConcurrentMap<Runnable, ScheduledFuture<?>> getTimeoutTasks() {
+        return timeoutTasks;
+    }
+
+    static class TimedTask implements Runnable {
+        private final Thread thread;
+        
+        public TimedTask(Thread thread) {
```
oswaldquek: is the parameter `thread` the thread executing the Authorisation of a payment? It's kinda hard to follow

nlsteers: Yes, the `thread` here is the thread running the task we want to interrupt. I'll make it a little clearer.

src/main/java/uk/gov/pay/connector/util/TimedThreadPoolExecutor.java
nlsteers: There'd be no point in using this class if you weren't intending to use a timeout so maybe this should be an illegal arg exception.

oswaldquek:
This looks reasonable I think.... is there any way to deploy it to the perf env and run perf tests against it? It'll give us confidence the changes here have no side effects.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardExecutorService.java
```diff
@@ -42,9 +42,7 @@ public class CardExecutorService {
     private static final int QUEUE_WAIT_WARN_THRESHOLD_MILLIS = 1000;
     public static final int SHUTDOWN_AWAIT_TERMINATION_TIMEOUT_SECONDS = 10;
     private final MetricRegistry metricRegistry;
-
-    private ExecutorServiceConfig config;
-    private ExecutorService executor;
+    private final TimedThreadPoolExecutor timedExectuor;
```
katstevens: Picky comment but I feel like fixing the typo `Exectuor` -> `Executor` will save us from Ctrl+F problems in the future...

nlsteers: Good eyes, I'll amend

katstevens:
I agree with @oswaldquek - this feels like a significant change and I'd be more comfortable approving after performance testing. This should be doable by putting the image into the relevant test-perf-1 ECR repo and pinning to it in Concourse?

katstevens:
Going to close this PR for now - the associated ticket was marked as done (and we have [a Grafana dashboard to monitor active threads](https://grafana.monitoring.pay-cd.deploy.payments.service.gov.uk/d/f4c4bf69-fb40-4aae-b003-97582f2c6c79/connector?orgId=1) with associated prometheus alerts)

PR #3869, opened 2022-07-12
PP-8694 Add updated_date to charges table

kbottla:
## WHAT YOU DID
- Add updated_date to charges table which is updated whenever there is an update to charge state.
  This will be used by the expiry sweeper to ignore any charges updated (potentially being authorised) within the last few minutes.

Co-authored-by: @stephencdaly


stephencdaly (approved):

PR #3870, opened 2022-07-12
PP-9790 Make CardAuthoriseService handle auth mode agreement

alexbishop1:
Make `CardAuthoriseService` handle authorisation mode agreement charges by extracting the necessary data from the stored payment instrument and then passing it to the payment provider class.

… which will then throw and exception because we have not implemented that bit yet.

with @sfount

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/AuthCardDetails.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/AuthCardDetails.java
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.ser.std.ToStringSerializer;
 import io.swagger.v3.oas.annotations.media.Schema;
-import uk.gov.pay.connector.charge.model.AddressEntity;
 import uk.gov.pay.connector.charge.model.CardDetailsEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.client.cardid.model.CardInformation;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/AuthCardDetails.java
```diff
@@ -75,19 +74,10 @@ public static AuthCardDetails of(AuthoriseRequest authoriseRequest, ChargeEntity
...
-                            addressEntity.getCity(),
-                            addressEntity.getCounty(),
-                            addressEntity.getCountry()
-                    ));
-            mayBeAddress.ifPresent(authCardDetails::setAddress);
-        }
+        Optional.ofNullable(chargeEntity.getCardDetails())
+                .flatMap(CardDetailsEntity::getBillingAddress)
+                .map(Address::from)
+                .ifPresent(authCardDetails::setAddress);
```
sfount, line 80: 💯

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -88,6 +100,9 @@ private AuthorisationResponse doAuthorise(String chargeId, AuthCardDetails authC
                 case WEB:
                     operationResponse = (GatewayResponse<BaseAuthoriseResponse>) paymentProvider.authorise(request, charge);
                     break;
+                case AGREEMENT:
+                    operationResponse = (GatewayResponse<BaseAuthoriseResponse>) paymentProvider.authoriseUserNotPresent(request, charge);
+                    break;
```
sfount, line 105: 👏

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -957,6 +964,44 @@ void doAuthoriseMotoApi_shouldTransitionToAuthorisationTimeout_andThrowException
...
+        var paymentInstrumentEntity = aPaymentInstrumentEntity(Instant.parse("2022-07-12T10:00:00Z"))
+                .withCardDetails(cardDetailsEntity)
+                .build();
+        var authCardDetails = AuthCardDetailsFixture.anAuthCardDetails().build();
+
+        charge.setAuthorisationMode(AuthorisationMode.AGREEMENT);
+        charge.setPaymentInstrument(paymentInstrumentEntity);
+        when(mockedChargeDao.findByExternalId(charge.getExternalId())).thenReturn(Optional.of(charge));
+        when(mockPaymentInstrumentEntityToAuthCardDetailsConverter.convert(paymentInstrumentEntity)).thenReturn(authCardDetails);
+        when(mockAuthCardDetailsToCardDetailsEntityConverter.convert(authCardDetails)).thenReturn(cardDetailsEntity);
```
sfount, line 981: We might want some additional coverage to make sure that when a charge is auth'd with a payment instrument it gets properties that don't make it to the `AuthCardDetails` object (for example the full card number).

This functionality isn't actually in yet but we should make sure it's covered too as this will mostly test the code path.

sfount (approved):
Added some comments about things we'll need to be sure are covered but this LGTM 👍

PR #3872, opened 2022-07-12
PP-8694 Expiry sweeper - skip expiring updated charges

kbottla:
## WHAT YOU DID
- Excludes charges updated within the last 5 minutes (based on config SKIP_EXPIRING_CHARGES_LAST_UPDATED_IN_SECONDS) from expiring.
  This is to allow any charges in locking (*_READY) state to complete the operation.


stephencdaly (approved):

PR #3873, opened 2022-07-12
PP-8694  Add BackfillerGatewayTransactionIdSet event

kbottla:
## WHAT YOU DID

- GatewayTransactionId is set before a charge is authorised (for Worldpay payments). gateway_transaction_id
  is included in the PAYMENT_ENTERED_DETAILS event. But sometimes, the thread could terminate without updating
  the charge status and the charge will eventually expire. In this case, gateway_transaction_id is not emitted to Ledger.
  Backfill gateway_trasaction_id so the charge can be expunged.


stephencdaly (approved):

PR #3874, opened 2022-07-12
PP-9790 Make sandbox card numbers more flexible

alexbishop1:
`SandboxCardNumbers` is now an interface with two implementations: `SandboxFullCardNumbers` and `SandboxLast4DigitsCardNumbers` (formerly `SandboxWalletCardNumbers`).

Code in `SandboxGatewayResponseGenerator` (which is now a concrete class rather than an interface implemented by `SandboxPaymentProvider` and `SandboxWalletAuthorisationHandler`) can use either implementation of `SandboxCardNumbers` to map between a card number (full or last four digits) and a sandbox gateway response.

The code in `SandboxPaymentProvider` (which knows what kind of authorisation it’s dealing with it) can then choose how it wants `SandboxGatewayResponseGenerator` to do the mapping.

This will be useful for authorising recurring card payments, where we have a payment instrument with the last four digits but not a full card number.


sfount (approved):
Code change looks good to me, inheritance, naming and organisation seem much clearer 👍 .

After discussion I suspect that we want the existing card numbers, described by https://docs.payments.service.gov.uk/testing_govuk_pay/#mock-card-numbers to work for recurring payments -- because of this we may need to add another numbers interface to only check the first 6 and last 4 digits of the existing card numbers.

We've added a follow up to include a card that is successful when setting up an agreement but subsequently declined for recurring payments.

PR #3876, opened 2022-07-12
PP-9790 Sandbox payment provider recurring authorisation

sfount:
Implement user not present interface for the Sandbox payment provider.

Use the new `last4DigitsSandboxResponseGenerator` to return a gateway
response which is matched to information from the payment instrument
available to user not present authorisation.

For now we will use this generator but plan to add a subsequent
generator that will respect the existing test card numbers.

---

The charge operations task queue isn't configured to do recurring
authorisations yet. Add an integration to assert that charges with
agreement authorisation mode are seen through to authorised or declined
depending on sandbox card numbers.

With @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -635,6 +635,16 @@ public class ChargeService {
             Optional.ofNullable(emailAddress).ifPresent(charge::setEmail);
 
             CardDetailsEntity detailsEntity = authCardDetailsToCardDetailsEntityConverter.convert(authCardDetails);
+
+            // propagate details that aren't mapped from payment instrument to auth card details onto the charge
+            // this logic should be removable when payment instruments are modelled and used for all authorisation types
+            charge.getPaymentInstrument()
+                    .ifPresent(paymentInstrument -> {
+                        if (charge.getAuthorisationMode() == AuthorisationMode.AGREEMENT) {
+                            detailsEntity.setFirstDigitsCardNumber(paymentInstrument.getCardDetails().getFirstDigitsCardNumber());
+                            detailsEntity.setFirstDigitsCardNumber(paymentInstrument.getCardDetails().getFirstDigitsCardNumber());
+                        }
+                    });
             charge.setCardDetails(detailsEntity);
 
             if (charge.isSavePaymentInstrumentToAgreement()) {
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -638,13 +638,13 @@ public class ChargeService {
 
             // propagate details that aren't mapped from payment instrument to auth card details onto the charge
             // this logic should be removable when payment instruments are modelled and used for all authorisation types
-            charge.getPaymentInstrument()
-                    .ifPresent(paymentInstrument -> {
-                        if (charge.getAuthorisationMode() == AuthorisationMode.AGREEMENT) {
+            if (charge.getAuthorisationMode() == AuthorisationMode.AGREEMENT) {
+                charge.getPaymentInstrument()
+                        .ifPresent(paymentInstrument -> {
                             detailsEntity.setFirstDigitsCardNumber(paymentInstrument.getCardDetails().getFirstDigitsCardNumber());
-                            detailsEntity.setFirstDigitsCardNumber(paymentInstrument.getCardDetails().getFirstDigitsCardNumber());
-                        }
-                    });
+                            detailsEntity.setLastDigitsCardNumber(paymentInstrument.getCardDetails().getLastDigitsCardNumber());
+                        });
+            }
             charge.setCardDetails(detailsEntity);
 
             if (charge.isSavePaymentInstrumentToAgreement()) {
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -635,6 +635,16 @@ public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
             Optional.ofNullable(emailAddress).ifPresent(charge::setEmail);
 
             CardDetailsEntity detailsEntity = authCardDetailsToCardDetailsEntityConverter.convert(authCardDetails);
+
+            // propagate details that aren't mapped from payment instrument to auth card details onto the charge
+            // this logic should be removable when payment instruments are modelled and used for all authorisation types
+            charge.getPaymentInstrument()
+                    .ifPresent(paymentInstrument -> {
+                        if (charge.getAuthorisationMode() == AuthorisationMode.AGREEMENT) {
```
alexbishop1: Would it make more sense to have the `if` statement wrapping the rest of the code or is there a reason it’s been done like this?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -635,6 +635,16 @@ public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
...
+                            detailsEntity.setFirstDigitsCardNumber(paymentInstrument.getCardDetails().getFirstDigitsCardNumber());
+                            detailsEntity.setFirstDigitsCardNumber(paymentInstrument.getCardDetails().getFirstDigitsCardNumber());
```
alexbishop1: One of these should be `LastDigitsCardNumber`, surely?

src/main/java/uk/gov/pay/connector/charge/util/AuthCardDetailsToCardDetailsEntityConverter.java
```diff
@@ -50,11 +52,17 @@ public CardDetailsEntity convert(AuthCardDetails authCardDetails) {
     }
 
     private static boolean hasFullCardNumber(AuthCardDetails authCardDetails) {
-        return authCardDetails.getCardNo().length() > 6;
+        return Optional.ofNullable(authCardDetails)
```
alexbishop1: I can understand there not being a card number but do we expect callers to pass a null `authCardDetails`?

src/main/java/uk/gov/pay/connector/charge/util/AuthCardDetailsToCardDetailsEntityConverter.java
```diff
@@ -50,11 +52,17 @@ public CardDetailsEntity convert(AuthCardDetails authCardDetails) {
...
+                .map(AuthCardDetails::getCardNo)
+                .map(cardNumber -> cardNumber.length() > 6)
```
alexbishop1, line 56: `.filter(…)` would be more appropriate than `.map(…)`.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceIT.java
```diff
@@ -0,0 +1,82 @@
...
+        var response = testContext.getInstanceFromGuiceContainer(CardAuthoriseService.class)
+                .doAuthoriseWeb("external-charge-id", AuthCardDetailsFixture.anAuthCardDetails().build());
+        assertThat(response.getGatewayError(), is(Optional.empty()));
+        assertThat(response.getAuthoriseStatus(), is(Optional.of(BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED)));
+    }
+    
+    @Test
+    public void shouldAuthoriseSandboxUserNotPresentPayment() {
+        addCharge("success-charge-external-id", "4242");
+        addCharge("decline-charge-external-id", "0002");
```
alexbishop1: I’m lazy so we can add constants for `"4242"` and `"0002"` describing what outcomes they will trigger?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceIT.java
```diff
@@ -0,0 +1,82 @@
...
+
+        var successCharge = testContext.getInstanceFromGuiceContainer(ChargeService.class).findChargeByExternalId("success-charge-external-id");
+        var declineCharge = testContext.getInstanceFromGuiceContainer(ChargeService.class).findChargeByExternalId("decline-charge-external-id");
```
alexbishop1: Might be better to split this into two tests?

alexbishop1 (changes requested):
Looks good, just a couple of things.

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/charge/util/AuthCardDetailsToCardDetailsEntityConverter.java
sfount: I've opted to use `map` for the moment as I think it most helpfully describes taking a optional string value and returning a boolean. `filter` `ifPresent()` would be slick though, I hadn't thought of that

src/main/java/uk/gov/pay/connector/charge/util/AuthCardDetailsToCardDetailsEntityConverter.java
sfount: I've opted to use `map` for the moment as I think it most helpfully describes taking one optional string value and returning a boolean. `filter` `ifPresent()` would be slick though, I hadn't thought of that

PR #3881, opened 2022-07-13
PP-8694 Expunger: Add all submitted states to historic charge exceptions

kbottla:
## WHAT YOU DID
- Charges in a terminal state are expunged after a day (or based on EXPUNGE_CHARGES_OLDER_THAN_DAYS env variables).
  Sometimes we don't receive notifications from payment provider and charges end up in `CAPTURE_SUBMITTED` state. These charges are expunged after a very long time (90 days currently).
- We also have other submitted states (EXPIRE_CANCEL_SUBMITTED, SYSTEM_CANCEL_SUBMITTED, USER_CANCEL_SUBMITTED) for which we may never receive notifications. Added submitted statuses to the list of exceptions so the changes with these statuses can be expunged (after 90 days).


SandorArpa (approved):
LGTM

PR #3882, opened 2022-07-13
PP-8694 Emit RefundAvailableUpdated event for submitted states

kbottla:
## WHAT YOU DID
- Emits RefundAvailabilityUpdated events for all submitted (EXPIRE_CANCEL_SUBMITTED, SYSTEM_CANCEL_SUBMITTED, USER_CANCEL_SUBMITTED) states as these affect refund availability status of payments.
- For payments in submitted (other than CAPTURE_SUMITTED) states, connector calculates refund availability status as `unavailable` but ledger could be returning `pending` status as RefundAvailalbilityUpdated event is not emitted for the charges (if there are no further updates to charges).


SandorArpa (approved):
LGTM

PR #3885, opened 2022-07-14
PP-9851 Rename capture specific queue helpers

sfount:
The capture queue will be extended to perform a number of asynchronous
charge operations.

Update the name of the capture queue helpers (message handler, process,
queue model) to have the more generic "charge async operations" naming
structure.

Anything relating to actually capturing a charge will remain the same
for now as capture will still be a job processed by this queue.


src/main/java/uk/gov/pay/connector/charge/service/ChargeEligibleForCaptureService.java
```diff
@@ -78,9 +78,9 @@ private ChargeStatus deriveCaptureStatusForChargeAuthorisationMode(ChargeEntity
         return charge.getAuthorisationMode() == MOTO_API ? CAPTURE_QUEUED : CAPTURE_APPROVED;
     }
 
-    private void addChargeToCaptureQueue(ChargeEntity charge) {
+    private void addCaptureToChargeAsyncOperationsQueue(ChargeEntity charge) {
```
alexbishop1: I think we could probably get away with just calling this method `addCaptureToQueue(…)`.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -813,7 +813,7 @@ public Optional<ChargeEntity> findByProviderAndTransactionId(String paymentGatew
 
     public boolean isChargeRetriable(String externalId) {
```
alexbishop1: Does this also need renaming now to reflect that it’s whether the capture is retriable (which disappointingly does not seem to be a word)?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/ChargeAsyncOperationsProcess.java
```diff
@@ -7,47 +7,48 @@
...
-    public void handleCaptureMessages() throws QueueException {
-        List<ChargeCaptureMessage> captureMessages = captureQueue.retrieveChargesForCapture();
-        for (ChargeCaptureMessage message : captureMessages) {
+    public void handleChargeAsyncOperationsMessage() throws QueueException {
+        List<ChargeAsyncOperationsMessage> chargeAsyncOperationsMessages = chargeAsyncOperationsQueue.retrieveAsyncOperations();
+        for (ChargeAsyncOperationsMessage message : chargeAsyncOperationsMessages) {
             try {
                 MDC.put(PAYMENT_EXTERNAL_ID, message.getChargeId());
-                LOGGER.info("Charge capture message received - [queueMessageId={}] [queueMessageReceiptHandle={}]",
+                LOGGER.info("Charge async operation message received - [queueMessageId={}] [queueMessageReceiptHandle={}]",
```
alexbishop1, line 43: Might we want to log the type of operation here?

src/main/java/uk/gov/pay/connector/queue/capture/ChargeAsyncOperationsMessage.java
```diff
@@ -0,0 +1,33 @@
+package uk.gov.pay.connector.queue.capture;
+
+import uk.gov.service.payments.commons.queue.model.QueueMessage;
+
+public class ChargeAsyncOperationsMessage {
+    private AsyncChargeOperation asyncChargeOperation;
+    private QueueMessage queueMessage;
```
alexbishop1: These fine fields can be `final`.

src/main/java/uk/gov/pay/connector/queue/managed/ChargeAsyncOperationsMessageReceiver.java
```diff
@@ -0,0 +1,86 @@
...
+                queueSchedulerThreadDelayInSeconds,
+                TimeUnit.SECONDS);
+    }
+
+    @Override
+    public void stop() {
+        LOGGER.info("Shutting down charge async operations service");
+        chargeAsyncOperationMessageExecutorService.shutdown();
+        try {
+            // Wait for existing charges to finish being captured
```
alexbishop1: This will need updating now or fairly soon.

src/main/java/uk/gov/pay/connector/queue/managed/ChargeAsyncOperationsMessageReceiver.java
```diff
@@ -0,0 +1,86 @@
...
+            if (chargeAsyncOperationMessageExecutorService.awaitTermination(queueSchedulerShutdownTimeoutInSeconds, TimeUnit.SECONDS)) {
+                LOGGER.info("charge async operation service shut down cleanly");
+            } else {
+                // If the existing charges being captured didn't terminate within the allowed time then force them to.
+                LOGGER.error("Async operations still being captured after shutdown wait time will now be forcefully stopped");
```
alexbishop1: This doesn’t read quite right.

alexbishop1 (changes requested):
Looks fine but there’s a few strings and things that might need tweaking.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
sfount: Great spot, yeah this specific refers to if the charge should be allowed to capture again. The logic won't be used at all by authorisation so we should make the name more specific

src/main/java/uk/gov/pay/connector/paymentprocessor/service/ChargeAsyncOperationsProcess.java
sfount: Yes but the type of operations don't exist yet! Wanted to start by having something that worked, this to follow in the next pr

alexbishop1 (approved):

sfount:
Although we think it's a good idea for charge operations to be close together, it transpires there already exists a generic task queue which would meet the requirements of something that only needs to guarantee to run once (like authorisation). We'll update acceptance criteria to use that instead.

PR #3890, opened 2022-07-15
PP-9791 Queue charges created with agreement auth mode for authorisation

sfount:
Depends on https://github.com/alphagov/pay-connector/pull/3889

When a charge is created with authorisation mode `agreement` and has
passed all of the validation (returned as created), add that charge to
the generic connector task queue.

* Add `AUTHORISE_USER_NOT_PRESENT` task to the task queue
* Add `AUTHORISATION_USER_NOT_PRESENT_QUEUED` gateway transition state,
  this will be externally represented as "started" and will transition
  from this to the available authorisation states when the job is
  processed


src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -39,6 +39,7 @@
 
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_TIMEOUT;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_USER_NOT_PRESENT_QUEUED;
```
sfount: This likely shouldn't be here

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueService.java
```diff
@@ -47,6 +48,8 @@ public void offerTasksOnStateTransition(ChargeEntity chargeEntity) {
                 !isEmpty(chargeEntity.getGatewayTransactionId()) &&
                 chargeEntity.getFees().isEmpty()) {
             addCollectStripeFeeForFailedPaymentTask(chargeEntity);
+        } else if (chargeEntity.getChargeStatus() == ChargeStatus.AUTHORISATION_USER_NOT_PRESENT_QUEUED) {
+            addAuthoriseUserNotPresentTask(chargeEntity);
```
sfount: Opinion: instead of being included in the `offerTasksOnStateTransition` this could be called from the `markChargeAsEligibleForUserNotPresentAuthorisation` method explicitly -- I think there would be less things to keep on top of when reading the code

src/main/java/uk/gov/pay/connector/events/model/charge/QueuedForAuthorisationUserNotPresent.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.pay.connector.events.model.charge;
+
+import java.time.ZonedDateTime;
+
+public class QueuedForAuthorisationUserNotPresent extends PaymentEventWithoutDetails {
```
alexbishop1: I suppose it’s`QueuedForUserNotPresentAuthorisation` (or maybe `QueuedForAuthorisationWithUserNotPresent`) but meh.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
alexbishop1: Begone!

src/main/java/uk/gov/pay/connector/queue/tasks/TaskType.java
```diff
@@ -4,7 +4,8 @@
 
 public enum TaskType {
     COLLECT_FEE_FOR_STRIPE_FAILED_PAYMENT("collect_fee_for_stripe_failed_payment"),
-    HANDLE_STRIPE_WEBHOOK_NOTIFICATION("handle_stripe_webhook_notification"); 
+    HANDLE_STRIPE_WEBHOOK_NOTIFICATION("handle_stripe_webhook_notification"),
+    AUTHORISE_USER_NOT_PRESENT("authorise_user_not_present");
```
alexbishop1: Maybe `AUTHORISE_WITH_USER_NOT_PRESENT`? I think this is probably understandable as it is to us in the know but my concern is that `AUTHORISE_USER_NOT_PRESENT` could sound like it’s authorising the user and something else is not present, rather than it’s doing an authorisation and the user is not present.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
```diff
@@ -91,11 +131,23 @@ public void shouldCreatePaymentWithAgreementIdAndAuthorisationModeAgreement() {
...
                 .statusCode(SC_CREATED)
                 .body(JSON_AGREEMENT_ID_KEY, is(JSON_VALID_AGREEMENT_ID_VALUE))
-                .contentType(JSON);
+                .contentType(JSON)
+                .extract().path("charge_id");
+
+        assertFrontendChargeStatusIs(chargeId, AUTHORISATION_USER_NOT_PRESENT_QUEUED.getValue());
+        assertApiStateIs(chargeId, EXTERNAL_STARTED.getStatus());
+
+        verify(mockAppender, times(1)).doAppend(loggingEventArgumentCaptor.capture());
```
alexbishop1: `times(1)` is the default.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
```diff
@@ -91,11 +131,23 @@ public void shouldCreatePaymentWithAgreementIdAndAuthorisationModeAgreement() {
...
+        List<LoggingEvent> logEvents = loggingEventArgumentCaptor.getAllValues();
+
+        var log = logEvents.get(0);
```
alexbishop1: Not so sure about using `var` here: it’s not immediately obvious what the type of `log` is (same with line below).

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
```diff
@@ -91,11 +131,23 @@ public void shouldCreatePaymentWithAgreementIdAndAuthorisationModeAgreement() {
...
+        var logArguments= Arrays.stream(log.getArgumentArray()).map(String::valueOf).collect(Collectors.toList());
```
alexbishop1: Missing space before the `=`.

`Collectors.toUnmodifiableList()` would be preferable (though longer).

alexbishop1 (changes requested):
Not that much to it, really. Few things to tidy up (a couple suggested by yourself before it goes in).

src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueueService.java
```diff
@@ -60,6 +60,20 @@ public void add(Task task) {
         }
     }
 
+    public void addAuthoriseUserNotPresentTask(ChargeEntity chargeEntity) {
```
alexbishop1: I suppose this should be `addAuthoriseWithUserNotPresentTask(…)` now.

src/main/java/uk/gov/pay/connector/queue/tasks/TaskType.java
```diff
@@ -4,7 +4,8 @@
 
 public enum TaskType {
     COLLECT_FEE_FOR_STRIPE_FAILED_PAYMENT("collect_fee_for_stripe_failed_payment"),
-    HANDLE_STRIPE_WEBHOOK_NOTIFICATION("handle_stripe_webhook_notification"); 
+    HANDLE_STRIPE_WEBHOOK_NOTIFICATION("handle_stripe_webhook_notification"),
+    AUTHORISE_WITH_USER_NOT_PRESENT("authorise_user_not_present");
```
alexbishop1: Do we want the string to match the name?

alexbishop1 (changes requested):
Thanks for making the changes, maybe two last things.

alexbishop1 (approved):

PR #3891, opened 2022-07-15
PP-9791 Do not allow return_url for agreement auth mode

sfount:
`return_url` is an unexpected attribute for agreement authorisation mode
as the user is not present to be returned to a integrating service.

Add agreement auth mode to a list of unexpected modes to reject the
return url property.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
@@ -35,6 +35,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
+import java.util.EnumSet;
 import java.util.Map;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -127,7 +128,7 @@ public Response createNewCharge(
                         throw new MissingMandatoryAttributeException(RETURN_URL);
                     });
 
-        } else if (authorisationMode == AuthorisationMode.MOTO_API && chargeRequest.getReturnUrl().isPresent()) {
+        } else if (List.of(AuthorisationMode.MOTO_API, AuthorisationMode.AGREEMENT).contains(authorisationMode) && chargeRequest.getReturnUrl().isPresent()) {
```
alexbishop1: Have a constant like this:

```java
private static final AUTHORISATION_MODES_INCOMPATIBLE_WITH_RETURN_URL = EnumSet.of(AuthorisationMode.MOTO_API, AuthorisationMode.AGREEMENT);
```

… then do this:

```java
AUTHORISATION_MODES_INCOMPATIBLE_WITH_RETURN_URL.contains(authorisationMode)
```

alexbishop1 (changes requested):
One tiny thing.

alexbishop1 (approved):

PR #3893, opened 2022-07-18
PP-9791 Handle authorise with user not present task

sfount:
Adds a SQS task service handler to process authorise with user not present
messages.

Ensure the correct handler is used when the SQS message is received.

Ensure the handler `process` method correctly lines charges up for
captures if the authorisation was successful.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/tasks/handlers/AuthoriseWithUserNotPresentHandler.java
+++ b/src/main/java/uk/gov/pay/connector/queue/tasks/handlers/AuthoriseWithUserNotPresentHandler.java
@@ -1,0 +1,33 @@
+package uk.gov.pay.connector.queue.tasks.handlers;
+
+import com.google.inject.Inject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.service.ChargeEligibleForCaptureService;
+import uk.gov.pay.connector.charge.service.ChargeService;
+import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/queue/tasks/handlers/AuthoriseWithUserNotPresentHandler.java
```diff
@@ -0,0 +1,33 @@
...
+
+    public void process(String chargeId) {
+        var charge = chargeService.findChargeByExternalId(chargeId);
+        var response = cardAuthoriseService.doAuthoriseUserNotPresent(charge);
+        response.getAuthoriseStatus()
+                .ifPresent(authoriseStatus -> {
+                    if (authoriseStatus == BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED) {
+                        chargeEligibleForCaptureService.markChargeAsEligibleForCapture(charge.getExternalId()); 
+                    }
+                });
```
alexbishop1, line 31: This would also work:

```java
response.getAuthoriseStatus()
                .filter(BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED::equals)
                .ifPresent(authoriseStatus -> chargeEligibleForCaptureService.markChargeAsEligibleForCapture(charge.getExternalId()));
```

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/AuthoriseWithUserNotPresentHandler.java
sfount: 👍 I like this, will move to this with the next PR

PR #3894, opened 2022-07-18
BAU — FirstDigitsCardNumber/LastDigitsCardNumber: Arabic numerals

alexbishop1:
Make `FirstDigitsCardNumber` and `LastDigitsCardNumber` only accept Arabic numberals (the 0, 1, 2, 3, 4, 5, 6, 7, 8 and 9 used in Latin scripts).

Previously, these two classes used `StringUtils.isNumeric(String)` from Apache Commons, which ultimately uses `Character.isDigit(char)`, which accepts any Unicode digit, including non-Arabic ones.

The code now uses regular expressions.

Since we validate card numbers in many other places beside `FirstDigitsCardNumber` and `LastDigitsCardNumber` we can be fairly confident that no non-Arabic numerals have slipped past, so there should be no observable change to the behaviour of the platform.

But it makes `FirstDigitsCardNumber` and `LastDigitsCardNumber` better maintain their invariants, making our code easier to reason about.

In addition:

- Remove the `FirstDigitsCardNumber.ofNullable(String)` and `LastDigitsCardNumber.ofNullable(String)` methods, which returned `null` if passed `null`. They were only used in some `AttributeConverter`s (where our convention is to have an explicit null check), some code for external telephone payments (where hey were not necessary) and some tests (where small refactorings could obviate the need for them).
- Rather than throwing a `RuntimeException` for invalid input, throw an `IllegalArgumentException` or a `NullPointerException`.
- Update the tests to JUnit 5 and add tests for some more edge cases.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/FirstDigitsCardNumber.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/FirstDigitsCardNumber.java
@@ -4,28 +4,23 @@ import org.apache.commons.lang3.StringUtils;
 import uk.gov.service.payments.commons.model.WrappedStringValue;
 
 import java.util.Objects;
+import java.util.regex.Pattern;
 
 public class FirstDigitsCardNumber extends WrappedStringValue {
+
+    private static final Pattern SIX_DIGITS = Pattern.compile("[0-9]{6}");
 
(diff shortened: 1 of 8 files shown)


SandorArpa (dismissed):
LGTM

SandorArpa (commented):
LOKTM

SandorArpa (approved):

PR #3899, opened 2022-07-19
PP-9788 Update DISPUTE_CREATED event details

kbottla:
## WHAT YOU DID
- Updated DISPUTE_CREATED event for the following
    - Removed `net_amount` and `fee` from dispute_created event as these are included as part of DISPUTE_LOST event and when service is charged for lost dispute
    - Updated evidence_due_date format to ZonedDateTime
    - Removed tests not required

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/dispute/DisputeCreatedEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/dispute/DisputeCreatedEventDetails.java
@@ -1,35 +1,29 @@
 package uk.gov.pay.connector.events.eventdetails.dispute;
 
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeSerializer;
+
+import java.time.ZonedDateTime;
+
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/events/eventdetails/dispute/DisputeCreatedEventDetails.java
```diff
@@ -1,37 +1,31 @@
...
-        this.netAmount = netAmount;
         this.reason = reason;
     }
 
-    public Long getFee() {
-        return fee;
-    }
-
-    public Long getEvidenceDueDate() {
+    public ZonedDateTime getEvidenceDueDate() {
```
stephencdaly, line 21: I think we need to make Adminusers handle this correctly before we update it, otherwise emails will break?

stephencdaly (approved):
LGTM! 

Adminusers changes to follow. If any disputes come in in the meantime, processing the DISPUTE_CREATED event in Adminusers will throw an exception and be retried. If any events end up on the dead letter queue, we can manually add them back onto the queue once the Adminusers change is in.

kbottla:
Adminusers PR https://github.com/alphagov/pay-adminusers/pull/1502 to handle evidenceDueDate as ZonedDateTime

PR #3901, opened 2022-07-19
PP-9328 Cancel agreement endpoint

sfount:
Add `/v1/api/account/:id/agreements/:id/cancel` endpoint.

This will move the payment instrument for a given agreement into the
cancelled state.

If no payment instrument is associated with the agreement, uniformly
throw agreement not active error.

404 if the agreement attempting to be cancelled doesn't exist.

---

Add agreement cancelled by service/ user events to mirror the event
structure of refunds (also processed through the admin tool or the API).

If user details are provided (generally by the admin tool) send the
cancelled by user event. If user details are no provided (generally
through the API) send the cancelled by service event.

with @alexbishop1

### DIFF
--- a/src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceIT.java
@@ -14,8 +14,11 @@ import uk.gov.pay.connector.junit.DropwizardConfig;
 import uk.gov.pay.connector.junit.DropwizardJUnitRunner;
 import uk.gov.pay.connector.junit.DropwizardTestContext;
 import uk.gov.pay.connector.junit.TestContext;
+import uk.gov.pay.connector.paymentinstrument.model.PaymentInstrumentStatus;
 import uk.gov.pay.connector.rules.LedgerStub;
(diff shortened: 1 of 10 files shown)


src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceIT.java
```diff
@@ -146,6 +154,29 @@ public void shouldReturn422WhenDescriptionEmpty() throws JsonProcessingException
...
+        databaseTestHelper.addAgreement(agreementParams);
+
+        givenSetup()
+                .post(format(CANCEL_AGREEMENT_URL, accountId, agreementId))
+                .then()
+                .statusCode(NO_CONTENT_204);
+        var agreementMap = databaseTestHelper.getPaymentInstrument(paymentInstrumentParams.getPaymentInstrumentId());
+        assertThat(agreementMap.get("status"), is("CANCELLED"));
+    }
+
```
alexbishop1, line 179: I wonder if a couple of sad path tests (one for when public API returns a 404 because there’s no agreement and one for when public API returns a 400 because the agreement is in the wrong state) would be useful here.

alexbishop1 (approved):
This looks fine but we may want a couple of more sad path integration tests.

PR #3912, opened 2022-07-21
PP-9677 Process dispute transfers included in payouts

stephencdaly:
Handle Stripe BalanceTransactions which have a source of type Transfer, and metadata "reason" field `transfer_dispute_amount` by emitting a DISPUTE_INCLUDED_IN_PAYOUT event. The event_details for this event contains the gateway_payout_id, which ledger will use to associate the dispute with the payout.

When a Transfer has a "reason" metadata field that is unrecognised or not handled, make the payout processing fail. This is to prevent us from wrongly reconciling any future Transfer types we add as refunds before we add the logic to the payout reconcile process to deal with them.

Use the MDC to add the payout id and connect account id to all log lines for the payout being processed.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
+++ b/src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
@@ -6,17 +6,20 @@ import com.stripe.model.Payout;
 import com.stripe.model.Transfer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
 import uk.gov.pay.connector.events.EventService;
 import uk.gov.pay.connector.events.model.Event;
 import uk.gov.pay.connector.events.model.charge.PaymentIncludedInPayout;
+import uk.gov.pay.connector.events.model.dispute.DisputeIncludedInPayout;
 import uk.gov.pay.connector.events.model.payout.PayoutCreated;
 import uk.gov.pay.connector.events.model.payout.PayoutEvent;
 import uk.gov.pay.connector.events.model.refund.RefundIncludedInPayout;
 import uk.gov.pay.connector.gateway.stripe.json.StripePayout;
 import uk.gov.pay.connector.gateway.stripe.json.StripePayoutStatus;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/payout/PayoutReconcileProcess.java
```diff
@@ -177,10 +177,24 @@ private void reconcileTransfer(PayoutReconcileMessage payoutReconcileMessage, Ba
...
-        } else {
-            emitRefundEvent(payoutReconcileMessage, transactionExternalId);
+
+        StripeTransferMetadataReason reason = stripeTransferMetadata.getReason();
+        switch (reason) {
+            case TRANSFER_FEE_AMOUNT_FOR_FAILED_PAYMENT:
+                emitPaymentEvent(payoutReconcileMessage, transactionExternalId);
+                break;
+            case TRANSFER_REFUND_AMOUNT:
+            case NOT_DEFINED:
```
kbottla, line 187: Am I correct that https://github.com/alphagov/pay-connector/pull/3912/files#diff-2c765e54f465d890c7ac1ceed5a77cb04ac761beea7baded88147e87692244a9R28 will now return `UNCOGNISED` when there is no reason on transfer? If so, is this case still relevant?

stephencdaly: If the reason isn't present, it will resolve as `NOT_DEFINED`, but if there is a reason but the value isn't one of the recognised values, i will return `UNRECOGNISED`.

I did it like this so we don't mistakenly classify any future transfers we make with different reasons as refunds before we've added handling for them here. Possibly we don't need to handle the case where there is no reason as a refund any more if all of those refunds will have now been paid out, but I'm not confident enough of that.

stephencdaly: The diff for [StripeTransferMetadataReason.java](https://github.com/alphagov/pay-connector/pull/3912/files#diff-2c765e54f465d890c7ac1ceed5a77cb04ac761beea7baded88147e87692244a9) doesn't show this for this PR, but we do something different if the reason is null

kbottla: ah I have not seen `if (value == null) {
            return NOT_DEFINED;
        }`. Makes sense now!

kbottla (approved):

PR #3915, opened 2022-07-22
PP-9880 Disallow nonsensical attributes with auth mode agreement

alexbishop1:
When a request to create a charge is made with `"authorisation_mode": "agreement"`, disallow `"moto": true`, `"email"` and `"prefilled_cardholder_details"` because they do not make sense in this context. Respond with an error with an `UNEXPECTED_ATTRIBUTE` error identifier if any of these are included.

Tweak some existing error responses to be more consistent with others. For example, if `"authorisation_mode"` is `"moto_api"`, including `"agreement_id"` now results in an error with an `UNEXPECTED_ATTRIBUTE` error identifier.

When `"save_payment_instrument_to_agreement": true` but `"authorisation_mode"` is not `"web"` (or unspecified), return an error with an `INCORRECT_AUTHORISATION_MODE_FOR_SAVE_PAYMENT_INSTRUMENT_TO_AGREEMENT` error identifier — we cannot simply use `UNEXPECTED_ATTRIBUTE` here because the `"save_payment_instrument_to_agreement"` attribute name needs to be remapped by public API.

Make it so that when `"authorisation_mode": "web"` (or unspecified), including `"save_payment_instrument_to_agreement": true` with no `"agreement_id"` results in an error with a GENERIC error identifier because this indicates a bug in public API.


sfount (dismissed):
Looks good to me, will need to be rebased as it looks like dependabot got there first

sfount (approved):

PR #3923, opened 2022-07-26
PP-9671 Refactor StripeWebhookTaskHandler

stephencdaly:
- Use MDC to store common logging fields
- Generate the dispute external ID in one place in the StripeWebhookTaskHandler rather than in the constructor for each event. This is to make is slightly clearer where this comes from and reduce duplication. This also allows us to add it to the MDC.

## Review notes

- Hiding whitespace diff will make this easier to review


SandorArpa (approved):
LGTM

PR #3937, opened 2022-07-29
PP-9671 Add magic cardholder name to skip submitting dispute evidence

stephencdaly:
Stripe provide 2 test card numbers which will create a dispute as soon as they are captured. On receiving the dispute created webhook, we automatically submit either winning or losing evidence to Stripe for these card numbers.

So that we can internally test the state of a payment when the dispute is still open, add a magic cardholder name: 'skip_evidence', which when used will cause evidence not to be automatically uploaded. This is necessary as there is not a different card number we can use to create the dispute.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeFullTestCardNumbers.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeFullTestCardNumbers.java
@@ -8,8 +8,12 @@ public class StripeFullTestCardNumbers {
 
     private static final String STRIPE_LOSING_EVIDENCE_CARD_NUMBER = "4000000000000259";
     private static final String STRIPE_WINNING_EVIDENCE_CARD_NUMBER = "4000000000002685";
+    private static final String SKIP_SUBMITTING_EVIDENCE_CARDHOLDER_NAME = "skip_evidence";
 
-    public static Optional<String> getSubmitTestDisputeEvidenceText(String firstSixDigits, String lastFourDigits) {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeFullTestCardNumbers.java
```diff
@@ -8,8 +8,12 @@ public class StripeFullTestCardNumbers {
 
     private static final String STRIPE_LOSING_EVIDENCE_CARD_NUMBER = "4000000000000259";
     private static final String STRIPE_WINNING_EVIDENCE_CARD_NUMBER = "4000000000002685";
+    private static final String SKIP_SUBMITTING_EVIDENCE_CARDHOLDER_NAME = "skip_evidence";
 
-    public static Optional<String> getSubmitTestDisputeEvidenceText(String firstSixDigits, String lastFourDigits) {
+    public static Optional<String> getSubmitTestDisputeEvidenceText(String firstSixDigits, String lastFourDigits, String cardholderName) {
+        if (cardholderName.equals(SKIP_SUBMITTING_EVIDENCE_CARDHOLDER_NAME)) {
```
kbottla: cardholder name can be null sometimes (if ledger is yet to process `PAYMENT_DETAILS_ENTERED` event) 
```suggestion
        if (SKIP_SUBMITTING_EVIDENCE_CARDHOLDER_NAME.equals(cardholderName) {
```

stephencdaly: Good spot, updated

kbottla (approved):

PR #3940, opened 2022-08-01
PP-9671 Adjust event date for refund availability event

stephencdaly:
For Stripe test payments, we add on a second to the dispute won and lost
events to ensure the events appear in the correct order in the payment
timeline - as Stripe sends the webhook for this at the same time as the
evidence submitted webhook.

When the dispute is won, we emit a REFUND_AVAILABILITY_UPDATED event.
Make this use the same timestamp as the DISPUTE_WON event so it is
easier to associate these.

Add a test to check that the events we emit when a dispute is won for a
test payment use the expected timestamps.


SandorArpa (approved):
LGTM

PR #3952, opened 2022-08-08
PP-9498 Split AgreementNotFoundException into two

alexbishop1:
Split `AgreementNotFoundException` into `AgreementNotFoundException` and `AgreementNotFoundBadRequestException`.

The former is used with an exception mapper that maps it to a 404 (not found) response and is intended for requests to endpoints like `/v1/api/accounts/account-id/agreements/non-existent-agreement-id/cancel`. The latter maps to a 400 (bad request) and is intended for requests to URLs like `/v1/api/accounts/account-id/charges` with a JSON body containing `"agreement_id": "non-existent-agreement-id"`.


src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -22,20 +22,21 @@
 import io.dropwizard.setup.Bootstrap;
 import io.dropwizard.setup.Environment;
 import org.glassfish.jersey.server.filter.RolesAllowedDynamicFeature;
+import uk.gov.pay.connector.agreement.exception.AgreementNotFoundExceptionMapper;
 import uk.gov.pay.connector.agreement.resource.AgreementsApiResource;
 import uk.gov.pay.connector.cardtype.resource.CardTypesResource;
 import uk.gov.pay.connector.charge.exception.AgreementIdWithIncompatibleOtherOptionsExceptionMapper;
 import uk.gov.pay.connector.charge.exception.AgreementMissingPaymentInstrumentExceptionMapper;
-import uk.gov.pay.connector.charge.exception.AgreementNotFoundExceptionMapper;
+import uk.gov.pay.connector.charge.exception.AgreementNotFoundBadRequestExceptionMapper;
```
sfount, line 30: I've not thought this all the way through but I think I'd generally be in favour of less exception mappers (but with the same amount of exceptions).

So in this case something like:
`AgreementExceptionMapper`

Could map both:
- `AgreementNotFoundException`
- `AgreementNotFoundBadRequestException`
- etc.

This follows precedent so all good for this change request, I just wonder if this is something we can start doing over time

sfount (approved):

PR #3961, opened 2022-08-12
PP-9899: Handle dispute warnings

SandorArpa:
## WHAT YOU DID
Do not process events for dispute notifications which have a warning status; instead, just log.


stephencdaly (approved):
LGTM!

PR #3962, opened 2022-08-12
PP-9906 Better handle retying capture for Stripe

stephencdaly:
A capture can fail if Stripe returns an error response to either the
request to capture the payment intent, or the request to transfer the
funds for the payment to the connect account. If just the transfer fails
we are often able to recover from this due to the idempotency of
Stripe's API.

However, if Stripe returns an error response for the capture attempt but
it actually succeeded on Stripe's end then Stripe will continue to
return an error response on retry attempts until the idempotency expires,
after which time it will return a different error stating that the
payment intent has already been captured.

Make changes to the capture process so that if we are trying to capture
a payment where capture has previously been attempted, first query
Stripe to find out whether the capture was already successful.

If Stripe says the payment has been captured, skip making the capture
request. Then check whether a transfer already exists, and only attempt
the transfer if no transfer exists.


src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -93,17 +84,68 @@ public CaptureResponse capture(CaptureGatewayRequest request) {
...
+                    String.format("Expected exactly one charge associated with payment intent %s, got %s", paymentIntent.getId(), stripeCharges.size()));
         }
+        return stripeCharges.get(0);
+    }
 
-        StripeCharge stripeCharge = stripeCharges.get(0);
+    private List<Fee> transferToConnectAccount(CaptureGatewayRequest request, StripeCharge capturedCharge, boolean checkForExistingTransfer) 
+            throws GatewayException.GenericGatewayException, GatewayErrorException, GatewayException.GatewayConnectionTimeoutException {
+        
+        List<Fee> feeList = generateFeeList(request, capturedCharge.getFee().orElse(0L));
```
stephencdaly: We should probably throw an error rather than returning 0 if there is no fee present on the Stripe charge.
We're now getting the fee from the request to get the payment intent when we retry if the payment has already been captured. For this request we need to make sure we expand the balance_transactions - so logging an error would let us know if this broke somehow.
Thoughts?

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
```diff
@@ -93,17 +84,68 @@ public CaptureResponse capture(CaptureGatewayRequest request) {
...
+
+        if (checkForExistingTransfer) {
+            StripeSearchTransfersRequest searchTransfersRequest = new StripeSearchTransfersRequest(
+                    request.getGatewayAccount(), stripeGatewayConfig, request.getExternalId());
+            
+            String rawResponse = client.getRequestFor(searchTransfersRequest).getEntity();
+            StripeSearchTransfersResponse transfersResponse = jsonObjectMapper.getObject(rawResponse,
+                    StripeSearchTransfersResponse.class);
+            
+            if (!transfersResponse.getTransfers().isEmpty()) {
```
stephencdaly, line 136: Do we think checking for any transfer is good enough at this point, or should we look for a transfer for the amount we expect?
We shouldn't have already made a different transfer at this point, but best to be safe?

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeGetPaymentIntentRequest.java
```diff
@@ -43,4 +50,13 @@ protected String urlPath() {
     public OrderRequestType getOrderRequestType() {
         return OrderRequestType.QUERY;
     }
+
+    @Override
+    public Map<String, String> getQueryParams() {
+        if (expandBalanceTransactions) {
+            return Map.of("expand[]", "charges.data.balance_transaction");
```
kbottla: I think we can always expand balance transaction. Any process other than capture can just ignore `balance_transaction` if not required.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripePaymentIntent.java
```diff
@@ -2,6 +2,8 @@
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
```
kbottla: Unused imports

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
kbottla: I think checking for any transfer is fine. 
At the moment the only other place in connector when transfer happens is when handling failed payments (TaskHandler) which shouldn't trigger unless charge is in CAPTURE ERROR state.

src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCaptureHandler.java
kbottla: Throwing error sounds good instead of setting the fee to `0`. `Fee` should always be present even when the payment amount is too low (1p).

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripeGetPaymentIntentRequest.java
stephencdaly: Thanks, yeh I was initially going to expand for all but then went back on it. Expanding for all is easier so I'll do that.

kbottla (approved):
LGTM!

PR #3971, opened 2022-08-17
PP-9911 Validate gateway account ID when retrieving agreement

alexbishop1:
When retrieving an agreement from the DAO by its external ID, only return agreements that belong to the appropriate gateway account.

with @sfount

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/dao/AgreementDaoIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/dao/AgreementDaoIT.java
@@ -8,83 +8,85 @@ import uk.gov.pay.connector.agreement.model.AgreementEntity;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialsEntity;
 
-import java.util.Map;
 import java.util.Optional;
 
 import static org.apache.commons.lang.math.RandomUtils.nextLong;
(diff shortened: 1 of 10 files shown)


src/test/java/uk/gov/pay/connector/it/dao/AgreementDaoIT.java
```diff
@@ -8,83 +8,85 @@
...
+        insertTestAgreement(AGREEMENT_EXTERNAL_ID_TWO, gatewayAccount1.getId());
+        Optional<AgreementEntity> agreement = agreementDao.findByExternalId(AGREEMENT_EXTERNAL_ID_ONE, gatewayAccount1.getId());
         assertThat(agreement.isPresent(), is(true));
         assertThat(agreement.get().getExternalId(), is(AGREEMENT_EXTERNAL_ID_ONE));
     }
 
+    @Test
+    public void findByExternalId_shouldNotFindAgreementEntityForDifferentGatewayAccount() {
+        insertTestAgreement(AGREEMENT_EXTERNAL_ID_ONE, gatewayAccount1.getId());
+        Optional<AgreementEntity> agreement = agreementDao.findByExternalId(AGREEMENT_EXTERNAL_ID_ONE, gatewayAccount2.getId());
```
sfount, line 57: To have full confidence that it's the newly introduced mechanism that's returning empty we might assert that find with agreement ID + valid gateway account id is present and invalid is not present (to rule out it just not being correctly inserted or something). I think this is covered by the above test so shouldn't be an issue

sfount (approved):

PR #3989, opened 2022-08-31
PP-9947 Update required number of capture retries

kbottla:
- Currently, the capture process retries charges that failed capture for up to 48 hours. This is due to ePDQ behaviour https://github.com/alphagov/pay-connector/commit/e931b4dab25284acedeb6d59d4bfd3d29e454b82. Since we longer have any services with ePDQ this could be reduced. 

- Based on the past 6 months' data, below is the behaviour of capture retries for current payment providers we support (see PP-9947 for details)
  - **Worldpay**: Most live payments are captured immediately. Very few remaining payments are captured after 1 or 2 retry attempts.
  - **Stripe**: Most payments are captured immediately or after 1 or 2 retries. A few exceptions are below
    - **Low-value payments**: Payments with low value cause negative transfer amounts so capture fails. These payments will always fail irrespective of the number of retry attempts
    - **Payments failed due to Stripe's internal error**: When Stripe returns a 5xx error, subsequent requests will return the same 5xx error due to idempotency requests (https://stripe.com/docs/api/idempotent_requests). Note that this is different to connection timeout where requests are successful on Stripe but timed out on our side which is fixed by https://github.com/alphagov/pay-connector/pull/3962. When Stripe responds with internal error / API error, payments won't be captured or transfer won't take place.

For the payments that failed due to idempotency (rare and happened for 33 payments in the last 6 months), Stripe processes requests after 24 hours. We make two requests (capture request + transfer) at the time of capture, so adding some buffer takes the desired number of capture retries config to **26**.

### DIFF
--- a/src/main/resources/config/config.yaml
+++ b/src/main/resources/config/config.yaml
@@ -128,10 +128,9 @@ captureProcessConfig:
 
   # The below effectively get multiplied together. In order to handle how
   # certain payment gateways do things, it is extremely desirable to keep these
-  # values such that we will continue retrying for at least two nights after
-  # the initial attempt. See PP-2627 or commit
-  # e931b4dab25284acedeb6d59d4bfd3d29e454b82 for more details.
-  maximumRetries: ${CAPTURE_PROCESS_MAXIMUM_RETRIES:-48}
+  # values such that we will continue retrying for at least 24 hours after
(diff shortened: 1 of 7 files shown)


SandorArpa (approved):
LGTM

PR #4059, opened 2022-10-04
PP-10173 Make captureSubmitTime in SettlementSummary an Instant

alexbishop1:
In `ChargeResponse.SettlementSummary`, change the type of `captureSubmitTime` from a `ZonedDateTime` to an `Instant`.


stephencdaly:
Does this change anything in the PublicApi response? It looks like PublicApi reads the captureSubmitTime as a string and passes that straight on in the response to the external consumer without manipulating it in any way

alexbishop1:
It does not. The `captureSubmitTime` continues to be serialised using the `ISO_INSTANT_MILLISECOND_PRECISION` formatter from [`ApiResponseDateTimeFormatter`](https://github.com/alphagov/pay-java-commons/blob/1bb6f98026ab79e69aed8a2261a5f973ce063ff1/model/src/main/java/uk/gov/service/payments/commons/model/ApiResponseDateTimeFormatter.java) in pay-java-commons and this serialises `ZonedDateTime`s and `Instant`s identically.

stephencdaly (approved):

PR #4062, opened 2022-10-04
PP-10173 EmittedEventEntity: use Instant for eventDate/emittedDate

alexbishop1:
In `EmittedEventEntity`, use an `Instant` rather than a `ZonedDateTime` for the `eventDate` and `emittedEventDate` fields.

Leave `doNotRetryEmitUntil` as a `ZonedDateTime` for now because it’s persisted into a database column of type `TIMESTAMP WITH TIME ZONE` (as opposed to `TIMESTAMP WITHOUT TIME ZONE`) so will require more tweaks.

This changes how the `EmittedEventBackfill` service logs timestamps when the time is bang on the minute (it will log
`2019-09-20T10:00:00Z` rather than `2019-09-20T10:00Z`) but I’ll eat my hat if anything is relying on that.

To avoid a `NullPointerException`, a change had to made to the `markEventAsEmitted_shouldRecordEventAndEmittedDate` test in `EmittedEventDaoIT`, adding a timestamp to a `RefundSubmitted` event. I believe all `Event`s (of the
`uk.gov.pay.connector.events.model.Event` kind) have a non-null timestamp (with the exception of `UnspecifiedEvent`, which never gets persisted or emitted) so this should be safe.

In `DatabaseTestHelper`, `addEmittedEvent` had to be changed so that the event date is converted to a `LocalDateTime` in UTC rather than a `Timestamp` because it seemed to be being mangled to be an hour ahead when the tests ran on my machine in the BST time zone, which I think is due to JDBI relying on the default time zone. Not entirely sure what change caused this to become a problem now but I’m fairly certain nothing will be altered in production where we run on UTC all year round.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/EmittedEventBatchIterator.java
+++ b/src/main/java/uk/gov/pay/connector/events/EmittedEventBatchIterator.java
@@ -3,6 +3,7 @@ package uk.gov.pay.connector.events;
 import uk.gov.pay.connector.app.config.EmittedEventSweepConfig;
 import uk.gov.pay.connector.events.dao.EmittedEventDao;
 
+import java.time.Instant;
 import java.time.ZonedDateTime;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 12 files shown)


alexbishop1:
Turns out some events in fact do not have timestamps, specifically the ominous-sounding `UnspecifiedEvent`. Fortunately, it never seems to get emitted and nothing ever accesses its timestamp (except `equals(…)`, `hashCode()` and `toString()` so I do not think it will be fall foul of any of the code that assumes events always have a timestamp.

kbottla (approved):

PR #4079, opened 2022-10-07
FIX: Dependabot/maven/com.stripe stripe java 21.10.0

sfount:
A [breaking interface change](https://github.com/stripe/stripe-java/blob/master/CHANGELOG.md#2100---2022-08-02) in release `21.0.0 - 2022-08-02` means
that the request options specified on the list method will no longer
apply to the auto paging request. Pass the request options into the auto
paging request to provide the Stripe API key.


SandorArpa (approved):
LGTM

PR #4102, opened 2022-10-18
PP-10137 Don't allow web payments for MOTO API payments

kbottla:
## WHAT YOU DID
- For MOTO API payments, the `one_time_token` generated should only be used by Auth API (`{publicapi)/v1/auth`). But currently, the token can be used by visiting the `/secure/` redirect URL and then payment can be completed using web pages. So return 404 (for MOTO API payments) for `GET one token/charge for token` endpoints which is used by the Frontend app.
- Also for MOTO_API payments, the `PaymentCreated` event should include prefilled billing details all the time and not just when the charge is in the `CREATED` state. This is because automated tests can create a payment and authorise it immediately. In these cases, it is likely that the charge is in an authorised state at the time PAYMENT_CREATED is emitted and prefilled details will not be included in the payment created event or any other event (for MOTO API payments).


SandorArpa (approved):
LGTM

PR #4126, opened 2022-11-01
PP-10233 include machine cookies generated by us only

SandorArpa:
## WHAT YOU DID

We occasionally get `Could not find order` errors from Worldpay for 3DS authorisation attempts.
As per Worldpay, this is due to invalid machine cookies sent in the request header.
As per Worldpay logs error is Unable to find data in SessionCache for second 3DS request : orderCode=XXX, merchantCode=XXX
We are seeing multiple machine cookies being passed in one string where we would only expect one as HTTP/1.1

- disable automatic cookie setting on clients
- add test to check that we only include our own cookie

    with @james-peacock-gds

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 
 import static java.lang.String.format;
 import static org.glassfish.jersey.apache.connector.ApacheClientProperties.CONNECTION_MANAGER;
+import static org.glassfish.jersey.apache.connector.ApacheClientProperties.DISABLE_COOKIES;
 import static org.glassfish.jersey.client.ClientProperties.READ_TIMEOUT;
 
 public class ClientFactory {
@@ -56,7 +57,8 @@ public class ClientFactory {
                 .using(new ApacheConnectorProvider())
                 .using(conf.getClientConfiguration())
                 .withProperty(READ_TIMEOUT, (int) readTimeout.toMilliseconds())
-                .withProperty(CONNECTION_MANAGER, 
+                .withProperty(DISABLE_COOKIES, true)
+                .withProperty(CONNECTION_MANAGER,
                         createConnectionManager(gateway.getName(), metricName, metricRegistry, conf.getCustomJerseyClient().getConnectionTTL()));
 
         if (System.getProperty(PROXY_HOST_PROPERTY) != null && System.getProperty(PROXY_PORT_PROPERTY) != null) {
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
```diff
@@ -304,4 +309,45 @@ public void shouldReturnStatus500_AWorldpayPaResParseError() {
 
         assertFrontendChargeStatusIs(chargeId, AUTHORISATION_ERROR.getValue());
     }
+
+    @Test
+    public void shouldNotIncludeMachineCookiesMoreThanOne() {
```
stephencdaly: Could you maybe add a code comment on this test to explain the previous behaviour and how this test attempts to reproduce it in order to check that the machine cookie is not added twice? I'm not really able to understand how this was happening from this test.

src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
```diff
@@ -56,7 +57,8 @@ private Client createWithDropwizardClient(PaymentGatewayName gateway, Duration r
                 .using(new ApacheConnectorProvider())
                 .using(conf.getClientConfiguration())
                 .withProperty(READ_TIMEOUT, (int) readTimeout.toMilliseconds())
-                .withProperty(CONNECTION_MANAGER, 
+                .withProperty(DISABLE_COOKIES, true)
```
kbottla, line 60: Would be good to provide a detailed explanation in the commit message (for future selves) about why this is being set (a bit about current behaviour too)

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
kbottla: Don't think the test asserts that it is set exactly once. Instead, it is testing the machine cookie is being set.

```suggestion
    public void shouldIncludeMachineCookieInTheRequestHeaderInThe3dsAuthorisationRequestToWorldpay() {
```

SandorArpa: 👍

src/main/java/uk/gov/pay/connector/gateway/ClientFactory.java
SandorArpa: 👍

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
SandorArpa: 👍

stephencdaly: I was confused about what this was testing. After the test name was updated, I don't think a comment is required anymore, as we discussed.

stephencdaly (approved):
LGTM!

PR #4135, opened 2022-11-03
PP-10198 Make flex credentials mandatory for Worlday non-MOTO live accounts

kbottla:
## WHAT YOU DID

- At present, Worldpay gateway_account_credentials record is set to ACTIVE as soon as Worldpay integration credentials are entered. It should still be the case for MOTO accounts and TEST accounts.
- For LIVE accounts, Worldpay Flex is required.
   - So, for these accounts, gateway_account_credentials state is set to ACTIVE only when both integration credentials and Flex credentials are configured

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
@@ -51,6 +51,7 @@ import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntityFixture.aGatewayAccountEntity;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType.LIVE;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType.TEST;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
```diff
@@ -308,6 +309,145 @@ void shouldNotChangeState_whenCredentialsNotInCreatedState() {
...
+
+            assertThat(credentialsEntity.getState(), is(ACTIVE));
+        }
+
+        @Test
+        void shouldChangeStateForWorldpayLiveMOTOAccount_whenCredentialsAreConfiguredAndNotFlexCredentials() {
+            GatewayAccountEntity gatewayAccountEntity = aGatewayAccountEntity()
+                    .withGatewayName(WORLDPAY.getName())
+                    .withType(LIVE)
+                    .withAllowMoto(false)
```
stephencdaly: The name of the test suggests the `allow_moto` should be true and we should be checking that the state is updated to `ACTIVE`?

kbottla: Both are good spots! I started on it but forgot to update it later. Now updated! https://github.com/alphagov/pay-connector/compare/fb350c2a2ca89bdb097867a57172ed0782bb5498..7beda36fb998cb773639741176b561c4bccadada

stephencdaly (approved):
LGTM!

PR #4145, opened 2022-11-08
PP-8693 Return 402 for gateway errors

james-peacock-ho:
## WHAT YOU DID
- Connector currently returns 500 when there are gateway errors.
- When front end receives 500 from connector this implies the problem is with connector rather than with the PSP, and causes unnecessary alerts.
- Instead 402 will now be returned when there are gateway errors.
- A change will also be made to the front end so that it handles 402. Front end will display the same error page to the user as it does for 500.


src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -50,8 +50,7 @@
 import java.util.Map;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
-import static uk.gov.pay.connector.util.ResponseUtil.badRequestResponse;
-import static uk.gov.pay.connector.util.ResponseUtil.serviceErrorResponse;
+import static uk.gov.pay.connector.util.ResponseUtil.*;
```
kbottla: Think the IDE setting needs to change. We don't use wildcard imports https://gds-way.cloudapps.digital/manuals/programming-languages/java.html#imports

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
```diff
@@ -11,12 +11,7 @@
 
 import static java.lang.String.format;
 import static javax.ws.rs.core.Response.Status;
-import static javax.ws.rs.core.Response.Status.ACCEPTED;
-import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
-import static javax.ws.rs.core.Response.Status.CONFLICT;
-import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
-import static javax.ws.rs.core.Response.Status.NOT_FOUND;
-import static javax.ws.rs.core.Response.Status.OK;
+import static javax.ws.rs.core.Response.Status.*;
```
kbottla: Same here for wild card imports

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
```diff
@@ -20,9 +20,7 @@
 import static java.lang.String.format;
 import static java.util.UUID.randomUUID;
 import static javax.ws.rs.core.HttpHeaders.COOKIE;
-import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
-import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
-import static javax.ws.rs.core.Response.Status.OK;
+import static javax.ws.rs.core.Response.Status.*;
```
kbottla: Wildcard imports

kbottla (changes requested):

kbottla:
depends on https://github.com/alphagov/pay-frontend/pull/3171

src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
james-peacock-ho: thanks, have corrected preferences and made changes

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayCardResourceIT.java
james-peacock-ho: done

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
james-peacock-ho: done

kbottla (approved):
LGTM! We should also update API specs for all endpoints but can be done in a separate PR  https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java#L143

PR #4149, opened 2022-11-09
BAU make Dockerfiles multistage

nlsteers:
## WHAT YOU DID

- update java Dockerfiles to include a `build` stage, this allows us to simplify our CI pipeline as it is no longer necessary to build jars separately to images

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,3 @@
-*/*
-!newrelic/*
-!docker-startup.sh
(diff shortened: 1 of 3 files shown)


katstevens:
@nlsteers Can we close this PR for now?

jfharden:
The dockerfile is multi-stage now and we have the separate ARM dockerfile too, so I'll close this PR for now

PR #4153, opened 2022-11-10
PP-10198 Gateway account credentials: Merge when state is updated

kbottla:
## WHAT YOU DID
- Merge gatway_account_credentials when the `state` is updated.
- When updating the `gatway_account_credentials` state as part of flex credentials changes, GatewayAccountCredentials are derived from GatewayAccountEntity and GatewayAccountCredentials entity is not attached to the persistence context. Thus JPA is not synchronizing changes for state changes.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/resource/GatewayAccountCredentialsResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/resource/GatewayAccountCredentialsResourceIT.java
@@ -44,6 +44,8 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.not;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.ACTIVE;
+import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.CREATED;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/resource/GatewayAccountCredentialsResourceIT.java
```diff
@@ -152,6 +154,29 @@ public void setWorldpay3dsFlexCredentialsWhenThereAreNonExisting() throws JsonPr
         assertThat(result.get("jwt_mac_key"), is(VALID_JWT_MAC_KEY));
     }
 
+    @Test
+    public void updateFlexCredentialsShouldSetGatewayAccountCredentialsStateToActive() throws JsonProcessingException {
```
alexbishop1, line 158: It should do this only if there are standard credentials already. Worth having a test where there are no standard credentials and 3DS Flex credentials are entered and the state does not get updated to `ACTIVE`?

kbottla: There is a unit already covering this scenario https://github.com/alphagov/pay-connector/blob/67dbd55d56733dfd9b0ee958096d2e4885396f4f/src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java#L357 and also above. This test is mainly to test that DB update is happening correctly.

alexbishop1 (approved):

PR #4157, opened 2022-11-15
PP-10198 Fix updating gateway credentials state

kbottla:
## WHAT YOU DID
- When new Flex credentials are created, set the same on the Gateway account so these can be used further to derive and update the Gateway account credentials state.
- When updating the credentials state after the Flex credentials update, look for a Worldpay credentials entity as Flex will only be applicable for the Worldpay credentials entity
- Fixes integration too which was supposed to test for a LIVE account. For test Worldpay accounts, Flex credentials are not mandatory to set the credentials state to ACTIVE.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -164,7 +164,18 @@ public class GatewayAccountCredentialsService {
         updateStateForCredentials(gatewayAccountCredentialsEntity);
     }
 
-    public void updateStateForCredentials(GatewayAccountCredentialsEntity credentialsEntity) {
+    @Transactional
+    public void updateStatePostFlexCredentialsUpdate(GatewayAccountEntity gatewayAccountEntity) {
+        // Flex credentials are currently set at the Gateway account level, and it is not possible to get the Worldpay
+        // credentials entity for which Flex credentials are updated if multiple exist. So get any Worldpay credential entity.
+        // Flex credentials should be moved to the Gateway credentials level to support multiple Flex credentials and
+        // get the correct credential entity here.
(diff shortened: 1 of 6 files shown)


SandorArpa (approved):
LOKTM

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -164,7 +164,18 @@ private void updateCredentials(JsonPatchRequest patchRequest, GatewayAccountCred
...
     }
 
-    public void updateStateForCredentials(GatewayAccountCredentialsEntity credentialsEntity) {
+    @Transactional
+    public void updateStatePostFlexCredentialsUpdate(GatewayAccountEntity gatewayAccountEntity) {
+        // Flex credentials are currently set at the Gateway account level, and it is not possible to get the Worldpay
+        // credentials entity for which Flex credentials are updated if multiple exist. So get any Worldpay credential entity.
+        // Flex credentials should be moved to the Gateway credentials level to support multiple Flex credentials and
+        // get the correct credential entity here.
+        GatewayAccountCredentialsEntity gatewayAccountCredentialsEntity = gatewayAccountEntity.getGatewayAccountCredentialsEntity(WORLDPAY.getName());
```
alexbishop1, line 173: Is there any filtering we can do here? Presumably if we find the credentials are `RETIRED` we don’t want to accept them?

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
kbottla: yeah, good point. we can exclude RETIRED creds if there are any just in case. Will update in a separate PR.

PR #4158, opened 2022-11-15
PP-10198 Exclude RETIRED credentials when deriving state post Flex credentials update

kbottla:
## WHAT YOU DID

- Exclude RETIRED credentials if any exist when getting a gateway credential to update post Flex credentials update.
- Also gets the latest creds, so this works if any service switches from Worldpay to Worldpay and enters Flex credentials as part of switching PSP


SandorArpa (approved):
LGTM

PR #4163, opened 2022-11-18
PP-10314 additional payment intent fields

SandorArpa:
## WHAT YOU DID
- we are processing Strtipe notifications already, but for failed payments we need some additional fields to be parsed for logging


stephencdaly (approved):
LGTM

PR #4211, opened 2022-12-20
PP-10144 Stop replacing gateway_transaction_id with null

SandorArpa:
## WHAT YOU DID
- In certain edge cases, the gateway_transaction_id gets overwritten with a null value when
      authorisations ends up in an unexpected error and the response is missing transaction id.
- For Worldpay, gateway_trasaction_id is set by us on the charge, so this change keeps the gateway_trasaction_id
      set on the charge as is if transaction id is missing in the response

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
@@ -55,7 +55,7 @@ public class AuthorisationService {
     public <T> T executeAuthoriseSync(Supplier<T> authorisationSupplier) throws AuthorisationExecutorTimedOutException {
         int timeoutInMilliseconds = authorisationConfig.getSynchronousAuthTimeoutInMilliseconds();
         return executeAuthorise(authorisationSupplier, timeoutInMilliseconds);
-    } 
+    }
 
     private <T> T executeAuthorise(Supplier<T> authorisationSupplier, int timeoutInMilliseconds)
             throws AuthorisationExecutorTimedOutException {
@@ -71,13 +71,16 @@ public class AuthorisationService {
         }
     }
 
-    public Optional<String> extractTransactionId(String chargeExternalId, GatewayResponse<? extends BaseAuthoriseResponse> operationResponse) {
+    public Optional<String> extractTransactionId(String chargeExternalId,
+                                                 GatewayResponse<? extends BaseAuthoriseResponse> operationResponse,
+                                                 String currentGatewayTransactionId) {
         Optional<String> transactionId = operationResponse.getBaseResponse()
                 .map(BaseAuthoriseResponse::getTransactionId);
 
         if (transactionId.isEmpty() || StringUtils.isBlank(transactionId.get())) {
             logger.warn("AuthCardDetails authorisation response received with no transaction id. -  charge_external_id={}",
                     chargeExternalId);
+            return Optional.ofNullable(currentGatewayTransactionId);
         }
 
         return transactionId;
--- a/src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardServiceTest.java
@@ -9,7 +9,6 @@ import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntityFixture;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.model.domain.FeeType;
-import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.chargeevent.dao.ChargeEventDao;
 import uk.gov.pay.connector.fee.model.Fee;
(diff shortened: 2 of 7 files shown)


src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardServiceTest.java
```diff
@@ -74,12 +74,25 @@ protected ChargeEntity createNewChargeWithFees(String provider, Long chargeId, C
     }
 
     protected ChargeEntity createNewChargeWith(String provider, Long chargeId, ChargeStatus status, String gatewayTransactionId) {
-        GatewayAccountEntity gatewayAccountEntity = ChargeEntityFixture.defaultGatewayAccountEntity();
+        GatewayAccountEntity gatewayAccountEntity = ChargeEntityFixture
```
kbottla: Can we use `GatewayAccountEntityFixture.aGatewayAccountEntity()` instead of separate methods in ChargeEntityFixture to create gatewayAccountEntity?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardServiceTest.java
```diff
@@ -74,12 +74,25 @@ protected ChargeEntity createNewChargeWithFees(String provider, Long chargeId, C
...
                 .aValidChargeEntity()
+                .withPaymentProvider(provider)
                 .withGatewayAccountEntity(gatewayAccountEntity)
                 .withId(chargeId)
                 .withStatus(status)
+                .withEvents(List.of(
+                        aChargeEventEntity().withChargeEntity(new ChargeEntity()).withStatus(ChargeStatus.CREATED).withUpdated(ZonedDateTime.now().minusHours(3)).build(),
+                        aChargeEventEntity().withChargeEntity(new ChargeEntity()).withStatus(ChargeStatus.AUTHORISATION_SUCCESS).withUpdated(ZonedDateTime.now()).build(),
+                        aChargeEventEntity().withChargeEntity(new ChargeEntity()).withStatus(ChargeStatus.ENTERING_CARD_DETAILS).withUpdated(ZonedDateTime.now().minusHours(2)).build(),
+                        aChargeEventEntity().withChargeEntity(new ChargeEntity()).withStatus(ChargeStatus.AUTHORISATION_TIMEOUT).withUpdated(ZonedDateTime.now().minusHours(1)).build(),
```
kbottla: Any reason different authorisation events are added here?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
```diff
@@ -71,13 +71,14 @@ private <T> T executeAuthorise(Supplier<T> authorisationSupplier, int timeoutInM
         }
     }
 
-    public Optional<String> extractTransactionId(String chargeExternalId, GatewayResponse<? extends BaseAuthoriseResponse> operationResponse) {
+    public Optional<String> extractTransactionId(String chargeExternalId, GatewayResponse<? extends BaseAuthoriseResponse> operationResponse, String currentGatewayTransactionId) {
```
kbottla: Can we add unit tests for this method?

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
```diff
@@ -90,6 +90,8 @@
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_UNEXPECTED_ERROR;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.ENTERING_CARD_DETAILS;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.UNDEFINED;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.STRIPE;
```
kbottla: Unused import?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardServiceTest.java
SandorArpa: Yes, without events `PaymentDetailsEntered.from` breaks. I reduced the number of events to minimum required.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
SandorArpa: 👍

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
SandorArpa: 👍

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardServiceTest.java
SandorArpa: 👍

stephencdaly (approved):
I'm not convinced this will solve the issue we're observing as we always set the gateway_transaction_id through ChargeService#setTransactionId which has a null check before overwriting the value.

However, this does improve the logging by including the gateway transaction ID for error responses that don't contain one and makes the code easier to follow.

PR #4212, opened 2022-12-20
PP-10439 Only create agreements when RCP enabled

james-peacock-ho:
- It should only be possible to create agreements when a gateway account has RCP enabled (recurring_enabled = true)
- If account does not have RCP enabled, throw custom exception
- Respond with 422 with error identifier RECURRING_CARD_PAYMENTS_NOT_ALLOWED
- Add unit test for this scenario and update existing
- PACT tests to be added under a separate ticket PP-10443

### DIFF
--- a/src/main/java/uk/gov/pay/connector/agreement/exception/RecurringCardPaymentsNotAllowedExceptionMapper.java
+++ b/src/main/java/uk/gov/pay/connector/agreement/exception/RecurringCardPaymentsNotAllowedExceptionMapper.java
@@ -1,0 +1,31 @@
+package uk.gov.pay.connector.agreement.exception;
+
+import javax.ws.rs.ext.ExceptionMapper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.common.model.api.ErrorResponse;
+import uk.gov.service.payments.commons.model.ErrorIdentifier;
+import javax.ws.rs.core.Response;
+import java.util.List;
+
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+public class RecurringCardPaymentsNotAllowedExceptionMapper 
+        implements ExceptionMapper<RecurringCardPaymentsNotAllowedException> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(RecurringCardPaymentsNotAllowedExceptionMapper.class);
+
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/agreement/exception/RecurringCardPaymentsNotAllowedExceptionMapper.java
```diff
@@ -0,0 +1,31 @@
...
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+public class RecurringCardPaymentsNotAllowedExceptionMapper 
+        implements ExceptionMapper<RecurringCardPaymentsNotAllowedException> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(RecurringCardPaymentsNotAllowedExceptionMapper.class);
+
+    private static final String RESPONSE_ERROR_MESSAGE = "Recurring payment agreements are not enabled on this account";
+
+    @Override
+    public Response toResponse(RecurringCardPaymentsNotAllowedException exception) {
```
kbottla, line 20: Can we have an integration test for this error? ITest makes sure that the mapper is configured correctly and that the response is as expected.

james-peacock-ho: test added. thanks!

src/test/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResourceIT.java
```diff
@@ -154,6 +154,21 @@ public void shouldReturn422WhenDescriptionEmpty() throws JsonProcessingException
...
+        String payload = objectMapper.writeValueAsString(Map.of(
+                "reference", REFERENCE_ID,
+                "description", DESCRIPTION,
+                "user_identifier", USER_IDENTIFIER
+        ));
+        givenSetup()
+                .body(payload)
+                .post(format(CREATE_AGREEMENT_URL, accountId))
+                .then()
+                .statusCode(SC_UNPROCESSABLE_ENTITY);
```
kbottla: Can the error identifier `RECURRING_CARD_PAYMENTS_NOT_ALLOWED ` and the message checked too?

james-peacock-ho: thanks - that's updated now

kbottla (approved):

PR #4216, opened 2022-12-21
PP-10435 Add payment details taken from payment instrument event

sfount:
Adds a new event/ event details to represent when payment details are
extracted from an existing payment instrument rather than input by a
paying user. 

Common fields are taken from the payment details entered event however 
this isn't extended directly as there are a number of fields that have 
been omitted as not relevant to recurring card payments.

When processing a recurring payment (authorisation mode agreement), be
specific about the type of event describing the payment instrument being
used to populate payment details.

Now that recurring charges use a specific event to represent that the
user is not entering details, make sure this is used when the charge is
being backfilled.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
@@ -78,4 +78,60 @@ public class PaymentDetailsTakenFromPaymentInstrumentEventDetails extends EventD
     public int hashCode() {
         return Objects.hash(cardType, cardBrand, firstDigitsCardNumber, lastDigitsCardNumber, cardholderName, expiryDate);
     }
+
+    public String getCardType() {
+        return cardType;
+    }
+
+    public String getCardBrand() {
+        return cardBrand;
+    }
+
+    public String getCardBrandLabel() {
+        return cardBrandLabel;
+    }
+
+    public String getFirstDigitsCardNumber() {
+        return firstDigitsCardNumber;
+    }
+
+    public String getLastDigitsCardNumber() {
+        return lastDigitsCardNumber;
+    }
+
+    public String getCardholderName() {
+        return cardholderName;
+    }
+
+    public String getExpiryDate() {
+        return expiryDate;
+    }
+
+    public String getAddressLine1() {
+        return addressLine1;
+    }
+
+    public String getAddressLine2() {
+        return addressLine2;
+    }
+
+    public String getAddressPostcode() {
+        return addressPostcode;
+    }
+
+    public String getAddressCity() {
+        return addressCity;
+    }
+
+    public String getAddressCounty() {
+        return addressCounty;
+    }
+
+    public String getAddressStateProvince() {
+        return addressStateProvince;
+    }
+
+    public String getAddressCountry() {
+        return addressCountry;
+    }
 }
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
@@ -1,7 +1,6 @@
 package uk.gov.pay.connector.events.eventdetails.charge;
 
 import uk.gov.pay.connector.cardtype.model.domain.CardBrandLabelEntity;
-import uk.gov.pay.connector.charge.model.AddressEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.events.eventdetails.EventDetails;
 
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
```diff
@@ -0,0 +1,137 @@
...
+    private final String expiryDate;
+    private final String addressLine1;
+    private final String addressLine2;
+    private final String addressPostcode;
+    private final String addressCity;
+    private final String addressCounty;
+    private final String addressStateProvince;
+    private final String addressCountry;
+
+    private PaymentDetailsTakenFromPaymentInstrumentEventDetails(String cardType, String cardBrand, String cardBrandLabel, String firstDigitsCardNumber, String lastDigitsCardNumber, String cardholderName, String expiryDate, String addressLine1, String addressLine2, String addressPostcode, String addressCity, String addressCounty, String addressStateProvince, String addressCountry) {
```
kbottla: I think we can use the builder pattern here. There are many arguments to this constructor

src/test/java/uk/gov/pay/connector/events/model/charge/PaymentDetailsTakenFromPaymentInstrumentTest.java
```diff
@@ -0,0 +1,82 @@
...
+        ChargeEventEntity chargeEventEntity = aChargeEventEntity()
+                .withUpdated(ZonedDateTime.parse(time))
+                .withChargeEntity(chargeEntity)
+                .build();
+        String json = PaymentDetailsTakenFromPaymentInstrument.from(chargeEventEntity).toJsonString();
+
+        assertDetails(json, chargeEntity);
+    }
+
+    private void assertDetails(String actual, ChargeEntity chargeEntity) {
```
kbottla, line 86: Can we have another where only minimum info is available? as a lot of fields (like cardType, billing address ..) are not mandatory

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
sfount: In my view the builder pattern used on other events doesn't feel like a good fit, the builder pattern lets you pick and choose which things are relevant to the thing you're instantiating but we then just make objects calling every `withX()` every time at the expense of 2/3x the boilerplate.

I don't foresee this changing much so my view is that it's the lesser of two evils to just have it in the constructor.

I think what these really are are records with no real manipulation or logic, just a method to serialise data similar to https://github.com/alphagov/pay-webhooks/blob/2e1e5f270384f3996cc8dba7cd19af40a11d121f/src/main/java/uk/gov/pay/webhooks/queue/EventMessageDto.java. Unfortunately I don't think Connector is up to date enough to use them

If we feel strongly that this should be a builder I can change it but I think it makes very simple code quite hard to follow

src/test/java/uk/gov/pay/connector/events/model/charge/PaymentDetailsTakenFromPaymentInstrumentTest.java
sfount: Great idea, will add an additional case

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
kbottla: Agree it won't change often. But having too many arguments impacts readability as it makes it hard to reason the order of arguments and if they match constructor args.

Fine to leave it as is for now but we can start with making the constructor definition multiline.

sfount: I've not been using intelijs built in builder tools which I think makes me more reserved with big classes like this. I've given it a go here https://github.com/alphagov/pay-connector/pull/4216/commits/b265b8dd2cb902adbda1ecd8898cdf1a30d8604d

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsTakenFromPaymentInstrumentEventDetails.java
```diff
@@ -0,0 +1,140 @@
...
+        cardDetails.getBillingAddress().ifPresent(billingAddress -> {
+            builder.setAddressLine1(billingAddress.getLine1())
+                    .setAddressLine2(billingAddress.getLine2())
+                    .setAddressPostcode(billingAddress.getPostcode())
+                    .setAddressCity(billingAddress.getCity())
+                    .setAddressCounty(billingAddress.getCounty())
+                    .setAddressStateProvince(billingAddress.getStateOrProvince())
+                    .setAddressCountry(billingAddress.getCountry());
+        });
+        return builder.createPaymentDetailsTakenFromPaymentInstrumentEventDetails();
```
kbottla, line 63: Could pass builder to constructor similar to https://github.com/alphagov/pay-connector/blob/d97c2e860a09385943ea5467c7bf68bb37a4d210/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/PaymentDetailsEnteredEventDetails.java#L37

```suggestion
        return new PaymentDetailsTakenFromPaymentInstrumentEventDetails(builder)
```

sfount: @kbottla 👍 https://github.com/alphagov/pay-connector/commit/80c080175aba25fbbc8680b976d6d82077c4ea6b

kbottla (approved):

PR #4232, opened 2023-01-05
PP-10542: Process balance.available Stripe notification

oswaldquek:


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeNotificationService.java
```diff
@@ -154,6 +158,31 @@ private void processDisputeNotification(StripeNotification notification, String
         taskQueueService.add(new Task(payload, TaskType.HANDLE_STRIPE_WEBHOOK_NOTIFICATION));
     }
 
+    private void processBalanceAvailableNotification(StripeNotification notification) {
+        try {
+            StripeBalance stripeBalance = toStripeBalance(notification.getObject());
+            if (!stripeBalance.getAvailable().isEmpty()) {
+                StripeBalance.Available available = stripeBalance.getAvailable().iterator().next();
+                logger.info("Logging stripe balance",
```
stephencdaly, line 168: Would it be worth adding in a comment about how we intend to use this?  I can see myself coming along and wondering that this is for and what will it affect if I want to change something about it in the future

stephencdaly (commented):
Code changes look good.

It would be good to have a code comment and some detail in the commit message about how this will be used so our future selves know why this was added.

oswaldquek:
> Code changes look good.
> 
> It would be good to have a code comment and some detail in the commit message about how this will be used so our future selves know why this was added.

I thought having the JIRA ticket PP-10542 in the commit message would suffice as someone in the future could go to that story and have a read, but I suppose it's less of faff if I wrote something here so I'll do that.

oswaldquek:
@stephencdaly I've [addressed your review comments which has more detail in the commit message](https://github.com/alphagov/pay-connector/pull/4232/commits/46e671fb709b6bdc2e36fad0f351ad58f47847f2). If you approve I will squash and merge the commits to make it one commit with the message.

stephencdaly (approved):
LGTM!

PR #4259, opened 2023-01-19
PP-10487 Fix task queue error for RCP

james-peacock-ho:
- Context: Occasional delays in database transactions are causing an error whereby connector's task queue message handler picks up a new task before the associated database record has been updated
- Add messages to task queue with 2-second delay
- Update associated unit test

### DIFF
--- a/src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueue.java
+++ b/src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueue.java
@@ -34,7 +34,7 @@ public class TaskQueue extends AbstractQueue {
 
     public void addTaskToQueue(Task task) throws QueueException, JsonProcessingException {
         String message = objectMapper.writeValueAsString(task);
-        QueueMessage queueMessage = sendMessageToQueue(message);
+        QueueMessage queueMessage = sendMessageToQueueWithDelay(message, 2);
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/queue/tasks/TaskQueue.java
```diff
@@ -34,7 +34,7 @@ public TaskQueue(SqsQueueService sqsQueueService,
 
     public void addTaskToQueue(Task task) throws QueueException, JsonProcessingException {
         String message = objectMapper.writeValueAsString(task);
-        QueueMessage queueMessage = sendMessageToQueue(message);
+        QueueMessage queueMessage = sendMessageToQueueWithDelay(message, 2);
```
stephencdaly, line 37: I think we should have the delay time as a Dropwizard app configuration setting.
We have `TaskQueueConfig` for all configurations settings related to the task queue.

We should then set this value in `config.yaml`, reading from an environment variable with a suitable name if it is present, else defaulting to 2 seconds. Make sure to also update `test-config.yaml` and `test-it-config.yaml` when you do this.

We probably don't need to set the environment variable in the terraform, and can just use the default value. But if we ever want to change this delay on a particular environment, it will be easy to do.

stephencdaly (changes requested):

PR #4274, opened 2023-01-30
PP-10639: Remove status from agreement events

alan-gds:
Ledger is no longer going to use this field, so it is now redundant; instead, Ledger itself will define the mapping from event types to `AgreementTypes`.

~https://github.com/alphagov/pay-ledger/pull/2425 needs to be deployed to production before this PR can be merged.~ done


sfount (commented):
Change looks great, really nice cleaning up with the events package move 👍 

This shouldn't be merged until Ledger is ready to map these events here https://github.com/alphagov/pay-ledger/pull/2425

src/main/java/uk/gov/pay/connector/events/model/agreement/AgreementCancelledByService.java
```diff
@@ -17,20 +17,12 @@ public static AgreementCancelledByService from(AgreementEntity agreement, Instan
                 agreement.getServiceId(),
                 agreement.getGatewayAccount().isLive(),
                 agreement.getExternalId(),
-                new AgreementCancelledByServiceEventDetails(AgreementStatus.CANCELLED),
+                new AgreementCancelledByServiceEventDetails(),
                 timestamp
         );
     }
 
     static class AgreementCancelledByServiceEventDetails extends EventDetails {
```
stephencdaly: I don't think we need this nested class if there are no event details. There is an overridden constructer in `AgreementEvent` that doesn't take an EventDetails

alan-gds: Fair enough; I will change this.

stephencdaly (approved):
LGTM!

PR #4276, opened 2023-01-30
PP-10488 Add logging fields for recurring payments

stephencdaly:
For all request contexts where a charge is present, add to the MDC:
- the authorisation_mode
- the agreement_external_id if one is set for the charge

For requests that include an "agreementId" as a path parameter, add the agreement_external_id to the MDC

Where we are logging using the `ChargeEntity$getStructuredLoggingArgs` method, include the authorisation_mode, and the agreement_external_id if one is set for the charge.


SandorArpa (approved):
LGTM

PR #4286, opened 2023-02-03
PP-10486 Check RCP Enabled on setup agreement or recurring payment

james-peacock-ho:
Context: requests to setup an agreement by taking the first payment, or taking further payments, should only be allowed when recurring_enabled=true for the gateway account
- Add check for recurring_enabled on first charge request, when authorisation mode is 'web' because user is present, and the payment details are being saved to the agreement to set it up
- Add check for recurring_enabled for subsequent charges on an existing agreement where authorisation mode is 'agreement'
- In each case, throw RCP Exception if recurring_enabled is false
- Update existing unit and integration tests to reflect that test accounts need to have recurring_enabled in relevant scenarios
- Add new unit and integration tests for the above logic

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
@@ -79,6 +79,7 @@ public class ChargesApiResourceCreateAgreementIT extends ChargingITestBase {
 
     @Test
     public void shouldCreatePaymentWithAgreementIdAndSavePaymentInstrumentToAgreementTrue() {
+        databaseTestHelper.enableRecurring(Long.valueOf(accountId));
         AddAgreementParams agreementParams = anAddAgreementParams()
                 .withGatewayAccountId(accountId)
                 .withExternalAgreementId(JSON_VALID_AGREEMENT_ID_VALUE)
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
```diff
@@ -592,7 +605,64 @@ public void shouldReturn422WhenAgreementIdIdFewerThan26Characters() {
...
+        connectorRestApiClient
+                .postCreateCharge(postBody)
+                .statusCode(SC_UNPROCESSABLE_ENTITY)
+                .contentType(JSON)
+                .body("message", contains("Recurring payment agreements are not enabled on this account"))
+                .body("error_identifier", is(ErrorIdentifier.RECURRING_CARD_PAYMENTS_NOT_ALLOWED.toString()));
+        }
+
+    @Test
+    public void shouldReturn422OnRecurringPaymentRequestWhenRecurringNotEnabledForGatewayAccount() {
```
stephencdaly: This isn't necessarily wrong, but I'd say that it's over testing to have 2 integration tests for the different cases. The integration test only needs to check that the RecurringCardPaymentsNotAllowedException is correctly mapped to the response.

As both of these tests are checking the same exception is mapped correctly, and the different scenarios are tested by the unit tests for the service, we could do without one of them. As integration tests take longer to run, we should really be trying to have the minimum number possible.

stephencdaly (dismissed):

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
james-peacock-ho: ok thanks, that's good to know, I will remove one of them as I agree we don't want to have more than we need! I have struggled to fully understand the distinctions between the different types of test so your explanation is helpful.

stephencdaly (approved):
LGTM!

PR #4291, opened 2023-02-06
PP-10488 Add additional logging for recurring payments

stephencdaly:
- Log when the status of a payment instrument is updated
- Log when we make a HTTP request to ledger to send events for recurring payments
- Log when the agreement is set up with a payment instrument

Remove overridden `toString` method on subclasses of `Event` as they didn't add any additional
useful information, and:
  - Add the `eventType` to the `toString` so this is logged when we make the POST
request to send events to ledger.
  - Remove `eventDetails` as logging the className of the eventDetails associated with an event
isn't particularly useful and throws a NullPointerException unless a null check is done.


SandorArpa (approved):
LGTM

PR #4294, opened 2023-02-07
PP-10498 add authorisation mode to charge

SandorArpa:
## WHAT YOU DID
- add authorisation_mode field to Charge
- add authorisation_mode to ChargeQueryGatewayRequest

with @alan-gds

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.charge.model.domain;
 
 import uk.gov.pay.connector.client.ledger.model.LedgerTransaction;
+import uk.gov.service.payments.commons.model.AuthorisationMode;
 
 import java.time.Instant;
 import java.util.Objects;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
```diff
@@ -110,7 +116,10 @@ public static Charge from(LedgerTransaction transaction) {
                 true,
                 transaction.getServiceId(),
                 transaction.getLive(),
-                transaction.isDisputed()
+                transaction.isDisputed(),
+                Optional.ofNullable(transaction.getAuthorisationMode())
+                        .map(AuthorisationMode::getName)
+                        .orElse(null)
```
stephencdaly: Ledger will always return an authorisation_mode, so this null check isn't necessary.

src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
```diff
@@ -26,11 +27,12 @@ public class Charge {
     private String serviceId;
     private boolean live;
     private Boolean disputed;
+    private String authorisationMode;
```
stephencdaly: We should use the `AuthorisationMode` enum type

src/main/java/uk/gov/pay/connector/charge/model/domain/Charge.java
```diff
@@ -77,7 +80,10 @@ public static Charge from(ChargeEntity chargeEntity) {
                 false,
                 chargeEntity.getServiceId(),
                 chargeEntity.getGatewayAccount().isLive(), 
-                null);
+                null,
+                Optional.ofNullable(chargeEntity.getAuthorisationMode())
+                        .map(AuthorisationMode::getName)
+                        .orElse(null));
```
stephencdaly: This null check isn't necessary - the `authorisation_mode` column in the `charges` table is non-nullable

stephencdaly (changes requested):

stephencdaly (approved):
LGTM!

PR #4295, opened 2023-02-07
PP-10498 add authorisation mode to ledger payment pact test

SandorArpa:
## WHAT YOU DID
- we are now using authorisation_mode from Ledger, thus we need to check if it comes when getting a payment.

with @alan-gds


james-peacock-ho (approved):
LGTM

PR #4300, opened 2023-02-09
PP-10486 refactor to use multilevel credentials

SandorArpa:
## WHAT YOU DID
We are changing credentials from a flat structure to a nested way for RCP. `recurring_merchant_initiated` is now a nested creds.
- refactor crdentials to be able to handle multilevel (Map<String, Object>)

with @alan-gds

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -28,6 +28,7 @@ import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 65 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -75,8 +76,9 @@ public GatewayAccountCredentials createGatewayAccountCredentials(GatewayAccountE
                                                                      String paymentProvider,
                                                                      Map<String, String> credentials) {
         GatewayAccountCredentialState state = calculateStateForNewCredentials(gatewayAccountEntity, paymentProvider, credentials);
+        Map<String, Object> newMap = new HashMap<>(credentials);
```
stephencdaly, line 80: Is it worth adding a short code comment about why we need to do this transformation so we don't forget later?
Hopefully we'll pick this up and change this soon when we play the story that will set the recurring credentials though so probably not a big deal.

stephencdaly (dismissed):
LGTM!

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
SandorArpa: 👍

stephencdaly (approved):

PR #4301, opened 2023-02-09
PP-10498 refactor how we retrieve credentials

SandorArpa:
## WHAT YOU DID
Retrieving Worldpay credentials will depend on the authorisation mode of the charge.
- refactor `GatewayRequest` to include `AuthorisationMode`. It's needed to determine the set of credentials to be used.

with @alan-gds


stephencdaly (approved):
LGTM!

PR #4302, opened 2023-02-09
PP-10429 Update Worldpay XML Request for Setup Agreement

james-peacock-ho:
- For RCP setup agreement requests to Worldpay, we need to include additional elements in the XML request
- Add storedCredentials element to XML request
- Add authenticatedShopperId to XML request
- Add createToken element to XML request
- Add unit test to ensure XML request is built correctly when savePaymentToInstrument is true


stephencdaly (dismissed):

stephencdaly (approved):

PR #4303, opened 2023-02-09
PP-10498 add new method to AuthUtil

SandorArpa:
## WHAT YOU DID
A new method needed that accepts AuthorisationMode with gateway credentials map. If AuthorisationMode is `AGREEMENT`, it should return the nested `recurring_merchant_initiated` creds or throw an excpetion if they are absent.

- implement method
- refactor Worldpay classes to use new method
- add relevant tests

with @alan-gds

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableMap;
 import uk.gov.pay.connector.app.StripeAuthTokens;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
 import uk.gov.pay.connector.gatewayaccount.model.WorldpayCredentials;
+import uk.gov.pay.connector.gatewayaccountcredentials.exception.MissingCredentialsForRecurringPaymentException;
 import uk.gov.service.payments.commons.model.AuthorisationMode;
 
 import java.util.Base64;
@@ -13,11 +14,12 @@ import static java.lang.String.format;
 import static javax.ws.rs.core.HttpHeaders.AUTHORIZATION;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_PASSWORD;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.CREDENTIALS_USERNAME;
+import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccount.RECURRING_MERCHANT_INITIATED;
 
 public class AuthUtil {
     private static final String STRIPE_VERSION_HEADER = "Stripe-Version";
     private static final String STRIPE_API_VERSION = "2019-05-16";
-    
+
     // We are using a separate version as searching payment intents is only supported in newer API versions.
     // This is intended as a temporary solution, and we intend to move all requests to use the same API version.
     public static final String STRIPE_SEARCH_PAYMENT_INTENTS_API_VERSION = "2020-08-27";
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -43,12 +45,24 @@ private static Map<String, String> getStripeAuthHeaderWithApiVersion(StripeGatew
...
     }
-    
+
+    public static Map<String, String> getGatewayAccountCredentialsAsAuthHeader(Map<String, Object> gatewayCredentials, AuthorisationMode authorisationMode) {
+        if (authorisationMode == AuthorisationMode.AGREEMENT) {
+            if (gatewayCredentials.get(RECURRING_MERCHANT_INITIATED) == null) {
+                throw new NoCredentialsInUsableStateException();
+            }
+            Map<String, Object> recurringCreds = (Map<String, Object>) gatewayCredentials.get(RECURRING_MERCHANT_INITIATED);
+            String value = encode(recurringCreds.get(CREDENTIALS_USERNAME).toString(), gatewayCredentials.get(CREDENTIALS_PASSWORD).toString());
```
stephencdaly: This is getting the password from `gatewayCredentials` rather than `recurringCreds`. It seems there is a missing unit test for this method to check that correct credentials are retrieved.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -566,6 +570,70 @@ void assert_authorization_header_is_passed_to_gateway_client() throws Exception
...
+
+        verify(authoriseClient).postRequestFor(
+                eq(WORLDPAY_URL),
+                eq(WORLDPAY), eq("test"),
+                gatewayOrderArgumentCaptor.capture(),
+                anyList(),
+                headers.capture());
+
+        assertThat(headers.getValue().size(), is(1));
+        assertThat(headers.getValue(), is(AuthUtil.getGatewayAccountCredentialsAsAuthHeader(credentials, mockChargeEntity.getAuthorisationMode())));
```
stephencdaly: I don't think we should be using `AuthUtil` in the test - as this test is trying to check that the correct credentials are selected - but by using `AuthUtil` to do this, if `AuthUtil` is doing the wrong thing we wouldn't pick up on this.

We should be asserting against a known value - so encode the username and password we expect to use in the tests itself and check that the value on the headers is equal to this.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -43,12 +45,24 @@ private static Map<String, String> getStripeAuthHeaderWithApiVersion(StripeGatew
...
     public static Map<String, String> getGatewayAccountCredentialsAsAuthHeader(Map<String, Object> gatewayCredentials) {
         String value = encode(gatewayCredentials.get(CREDENTIALS_USERNAME).toString(), gatewayCredentials.get(CREDENTIALS_PASSWORD).toString());
         return ImmutableMap.of(AUTHORIZATION, value);
     }
-    
+
+    public static Map<String, String> getGatewayAccountCredentialsAsAuthHeader(Map<String, Object> gatewayCredentials, AuthorisationMode authorisationMode) {
+        if (authorisationMode == AuthorisationMode.AGREEMENT) {
+            if (gatewayCredentials.get(RECURRING_MERCHANT_INITIATED) == null) {
+                throw new NoCredentialsInUsableStateException();
```
stephencdaly: Could we pass in a specific message to this exception so we know the exact reason why there are no credentials in a usable state. The `NoCredentialsInUsableStateExceptionMapper` logs the exception message.

stephencdaly (changes requested):

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
SandorArpa: 👍

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
stephencdaly: Possibly I'd go for using a completely separate exception. As "NoCredentialsInUsableState" is referring to a credential in the correct state, e.g. "ACVITE", "ENTERED" etc..

We would map the new exception using the same error_identifier though.

SandorArpa: I think the latter is better

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
SandorArpa: 👍

stephencdaly (approved):
LGTM!

PR #4306, opened 2023-02-10
PP-10429 Handle Worldpay Response With Recurring Auth Token

james-peacock-ho:
Context: when setting up an agreement, Worldpay response includes a paymentTokenID and optionally a transactionIdentifier.
- Save above fields to the recurring_auth_token column in payment_instruments
- Add integration test


stephencdaly (approved):

PR #4316, opened 2023-02-14
PP-10498 add new method to select appropriate merchant id

SandorArpa:
## WHAT YOU DID
We need to select the right merchant id for Worldpay.

- add a new method that selects the rigth merchant id for recurring payments

with @james-peacock-gds


stephencdaly (approved):
LGTM!

PR #4317, opened 2023-02-14
PP-10427: Put `AgreementEntity` on `ChargeEntity` directly

alan-gds:
Instead of `ChargeEntity` exposting an `agreementExternalId`, have it contain and expose the `AgreementEntity` directly, so that the `AgreementEntity` doesn't have to keep being looked up by id everywhere else.


stephencdaly (approved):

PR #4321, opened 2023-02-16
PP-10430 add a new gateway request type

SandorArpa:
## WHAT YOU DID
Recurring payments only needs a trimmed down version of CardAuthorisationGatewayRequest.

- add a new implementation of `GatewayRequest` that is to be used by Worldpay's recurring payment order.

with @james-peacock-gds


stephencdaly (approved):
LGTM!

PR #4331, opened 2023-02-17
PP-10427 Send agreement ID as metadata

stephencdaly:
When creating a customer object in Stripe, send the agreement external ID as metadata to store against the customer.


james-peacock-ho (approved):
LGTM

PR #4362, opened 2023-03-06
PP-10761 `card.` frontend support

nlsteers:
## WHAT YOU DID

- introduced new injected values into `LinksConfig`
  - boolean value to enable `card.` `next_url`
  - `card.` frontend url value
- added logic to LinksConfig to determine based on boolean value when to globally enable `card.`


src/main/java/uk/gov/pay/connector/app/LinksConfig.java
```diff
@@ -2,11 +2,27 @@
 
 import io.dropwizard.Configuration;
 
+import java.time.Instant;
```
alexbishop1: Did this import end up getting used in the end?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -285,8 +285,8 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
             }
 
             var agreementEntity = Optional.ofNullable(chargeRequest.getAgreementId()).map(agreementId ->
-                agreementDao.findByExternalId(chargeRequest.getAgreementId(), gatewayAccount.getId())
-                        .orElseThrow(() -> new AgreementNotFoundBadRequestException("Agreement with ID [" + chargeRequest.getAgreementId() + "] not found.")));
+                    agreementDao.findByExternalId(chargeRequest.getAgreementId(), gatewayAccount.getId())
```
alexbishop1, line 288: Looks like none of the whitespace changes in this file are related to this.

alexbishop1 (dismissed):
Looks fine.

src/main/java/uk/gov/pay/connector/app/LinksConfig.java
nlsteers: nope, will remove

nlsteers:
Is related to and requires the following infra pr to be applied in all envs:

https://github.com/alphagov/pay-infra/pull/4252

alexbishop1 (dismissed):

src/main/resources/config/config.yaml
```diff
@@ -31,6 +31,8 @@ logging:
 
 links:
   frontendUrl: ${FRONTEND_URL}
+  cardFrontendUrl: ${CARD_FRONTEND_URL}
+  cardFrontendEnabled: ${CARD_FRONTEND_ENABLED:-false}
```
jfharden, line 35: Will dropwizard convert the string true into a boolean automatically? The env var is the string true/false rather than a boolean, in python for example the string "false" is a truthy value.

nlsteers: We've used this pattern in the past (some existing examples in the `config.yml` as well) and Dropwizard has always been able to interpret the correct boolean value from the string literal.

Additionally, we check this new behaviour is correct in the integration tests which validate and consume the config files.

alexbishop1 (approved):

PR #4367, opened 2023-03-07
PP-10512 Delete Stored Payment Details (Payment Provider method)

james-peacock-ho:
- add default deleteStoredPaymentDetails method to PP interface
- call deleteStoredPaymentDetails method from task handler
- integration test
- Worldpay implementation will follow in separate PR


src/main/java/uk/gov/pay/connector/gateway/model/request/DeleteStoredPaymentDetailsGatewayRequest.java
```diff
@@ -0,0 +1,47 @@
...
+        return GatewayOperation.DELETE_STORED_PAYMENT_DETAILS;
+    }
+
+    @Override
+    public Map<String, Object> getGatewayCredentials() {
+        return agreement.getGatewayAccount().getCredentials(agreement.getGatewayAccount().getGatewayName());
+    }
+
+    @Override
+    public AuthorisationMode getAuthorisationMode() {
```
stephencdaly, line 44: I don't think we need this.

stephencdaly (approved):
LGTM! Just a getter that I don't think will be used which can be removed in next PR.

src/main/java/uk/gov/pay/connector/gateway/model/request/DeleteStoredPaymentDetailsGatewayRequest.java
james-peacock-ho: I agree we shouldn't need it, but I think we have to include it if we are implementing GatewayRequest interface.

stephencdaly: Oh - do we need to implement that interface?

james-peacock-ho: ha ha, good question!

PR #4369, opened 2023-03-07
BAU Attempt to fix flakey test

stephencdaly:
Integration tests fairly often fail with:

```CollectFeesForFailedPaymentsTaskHandlerIT.setUp:69 » UnableToExecuteStatement org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "gateway_accounts_pkey"```

A test is somehow leaving behind some database state with a conflicting gateway account ID. Edit some integration tests to use random longs for the gateway account ID to make this conflict less likely.


james-peacock-ho (approved):
LGTM!

PR #4373, opened 2023-03-08
PP-10512 Correct DeleteStoredPaymentDetailsTaskHandlerIT

james-peacock-ho:
This test was failing in github actions following merge; conflict with same accountId being used in other tests
- add After method to truncate database after test
- set accountId to nextLong rather than hardcoded value to avoid conflicts


stephencdaly (approved):
LGTM!

PR #4383, opened 2023-03-09
PP-10512-Implement-deleteStoredPaymentDetails-for-Worldpay

james-peacock-ho:
Context: When an agreement is cancelled, a delete token request should be sent to Worldpay
- Add Worldpay implementation for deleteStoredPaymentDetails interface method
- Build delete token order, applying agreement and paymentTokenID to template
- Add unit tests and contract test
- Correct template and valid request file

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
@@ -59,6 +59,10 @@ public class AuthUtil {
         return gatewayCredentials.get(CREDENTIALS_MERCHANT_ID).toString();
     }
 
+    public static String getWorldpayMerchantCodeForManagingTokens(Map<String, Object> gatewayCredentials) {
+        return gatewayCredentials.get(CREDENTIALS_MERCHANT_ID).toString();
+    }
+
     public static Map<String, String> getGatewayAccountCredentialsAsAuthHeader(Map<String, Object> gatewayCredentials) {
         String value = encode(gatewayCredentials.get(CREDENTIALS_USERNAME).toString(), gatewayCredentials.get(CREDENTIALS_PASSWORD).toString());
         return ImmutableMap.of(AUTHORIZATION, value);
@@ -77,6 +81,11 @@ public class AuthUtil {
         return ImmutableMap.of(AUTHORIZATION, value);
     }
 
+    public static Map<String, String> getGatewayAccountCredentialsForManagingTokensAsAuthHeader(Map<String, Object> gatewayCredentials) {
+        String value = encode(gatewayCredentials.get(CREDENTIALS_USERNAME).toString(), gatewayCredentials.get(CREDENTIALS_PASSWORD).toString());
+        return ImmutableMap.of(AUTHORIZATION, value);
+    }
+    
     public static Map<String, String> getWorldpayCredentialsCheckAuthHeader(WorldpayCredentials worldpayCredentials) {
         String value = encode(worldpayCredentials.getUsername(), worldpayCredentials.getPassword());
         return ImmutableMap.of(AUTHORIZATION, value);
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -23,6 +23,7 @@ import uk.gov.pay.connector.gateway.model.request.Auth3dsResponseGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
+import uk.gov.pay.connector.gateway.model.request.DeleteStoredPaymentDetailsGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.RecurringPaymentAuthorisationGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
@@ -50,6 +51,7 @@ import java.net.URI;
 import java.time.Instant;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 
 import static java.lang.String.format;
@@ -58,9 +60,12 @@ import static java.util.Collections.singletonList;
 import static java.util.UUID.randomUUID;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
 import static uk.gov.pay.connector.gateway.util.AuthUtil.getGatewayAccountCredentialsAsAuthHeader;
+import static uk.gov.pay.connector.gateway.util.AuthUtil.getGatewayAccountCredentialsForManagingTokensAsAuthHeader;
 import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderRequestBuilder.aWorldpay3dsResponseAuthOrderRequestBuilder;
 import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderRequestBuilder.aWorldpayCancelOrderRequestBuilder;
+import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderRequestBuilder.aWorldpayDeleteTokenOrderRequestBuilder;
 import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderRequestBuilder.aWorldpayInquiryRequestBuilder;
+import static uk.gov.pay.connector.gateway.worldpay.WorldpayOrderStatusResponse.WORLDPAY_RECURRING_AUTH_TOKEN_PAYMENT_TOKEN_ID_KEY;
 import static uk.gov.pay.connector.paymentprocessor.model.Exemption3ds.EXEMPTION_HONOURED;
 import static uk.gov.pay.connector.paymentprocessor.model.Exemption3ds.EXEMPTION_NOT_REQUESTED;
 import static uk.gov.pay.connector.paymentprocessor.model.Exemption3ds.EXEMPTION_OUT_OF_SCOPE;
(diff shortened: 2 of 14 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -369,6 +388,14 @@ private GatewayOrder buildCancelOrder(CancelGatewayRequest request) {
                 .build();
     }
 
+    private GatewayOrder buildDeleteTokenOrder(DeleteStoredPaymentDetailsGatewayRequest request) {
+        var token = request.getPaymentInstrument().getRecurringAuthToken().orElseThrow(NoSuchElementException::new);
+        return aWorldpayDeleteTokenOrderRequestBuilder()
+                .withAgreementId(request.getAgreement().getExternalId())
+                .withPaymentTokenId(token.get(WORLDPAY_RECURRING_AUTH_TOKEN_PAYMENT_TOKEN_ID_KEY))
+                .withMerchantCode(AuthUtil.getWorldpayMerchantCode(request.getGatewayCredentials(), request.getAuthorisationMode()))
```
stephencdaly: This is an interesting one - as there isn't an "authorisation mode" associated with this request.

Possibly we need a method in `AuthUtil` that directly gets the credentials used for managing tokens without providing an authorisation mode and call that here. I think these credentials will be the credentials used to create the token (authorisation mode `WEB`), rather than the credentials used to take recurring payments. Although maybe we can use either?

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/DeleteStoredPaymentDetailsTaskHandler.java
```diff
@@ -29,9 +30,13 @@ public void process(String agreementExternalId, String paymentInstrumentExternal
...
         var paymentInstrument = paymentInstrumentService.findByExternalId(paymentInstrumentExternalId);
         PaymentProvider paymentProvider = providers.byName(PaymentGatewayName.valueFrom(agreement.getGatewayAccount().getGatewayName()));
-
         DeleteStoredPaymentDetailsGatewayRequest request = new DeleteStoredPaymentDetailsGatewayRequest(agreement, paymentInstrument);
-        var response = paymentProvider.deleteStoredPaymentDetails(request);
-        LOGGER.info("Processed deleteStoredPaymentDetails task with response: {}", response.toString());
+        try {
+            var response = paymentProvider.deleteStoredPaymentDetails(request);
+            LOGGER.info("Processed deleteStoredPaymentDetails task with response: {}", response.toString());
+        } catch (GatewayException ex) {
```
stephencdaly: I don't think we should catch this exception as this will mean we silently fail to delete the token and don't retry. We should propagate it up to the TaskQueueMessage handler, where it will be caught and handled appropriately.
If `TaskQueueMessageHandler` handles an exception, it will not mark the SQS message as processed, and so it will be retried some time later. If processing the message repeatedly fails, the message will be sent to a dead letter queue and we will get alerted to Zendesk for support to investigate.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
james-peacock-ho: new method added as discussed

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/DeleteStoredPaymentDetailsTaskHandler.java
james-peacock-ho: changed as requested.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -77,6 +77,11 @@ public static Map<String, String> getGatewayAccountCredentialsAsAuthHeader(Map<S
         return ImmutableMap.of(AUTHORIZATION, value);
     }
 
+    public static Map<String, String> getGatewayAccountCredentialsForManagingTokens(Map<String, Object> gatewayCredentials) {
```
stephencdaly: This is more for constructing an auth header using the appropriate credentials set out of the credentials - so maybe it should be called something like `getAuthHeaderForManagingTokens` or `getGatewayAccountCredentialsForManagingTokensAsAuthHeader` (quite long) if you want to match the other methods.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -99,7 +104,7 @@ public WorldpayPaymentProvider(@Named("WorldpayGatewayUrlMap") Map<String, URI>
                                    @Named("WorldpayAuthoriseGatewayClient") GatewayClient authoriseClient,
                                    @Named("WorldpayCancelGatewayClient") GatewayClient cancelClient,
                                    @Named("WorldpayInquiryGatewayClient") GatewayClient inquiryClient,
-                                   WorldpayWalletAuthorisationHandler worldpayWalletAuthorisationHandler,
+                                   @Named("WorldpayDeleteTokenGatewayClient") GatewayClient deleteTokenClient, WorldpayWalletAuthorisationHandler worldpayWalletAuthorisationHandler,
```
stephencdaly: could you put the WorldpayWalletAuthorisationHandler argument on the next line please?

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -343,6 +349,17 @@ public GatewayResponse<BaseCancelResponse> cancel(CancelGatewayRequest request)
...
     }
 
+    @Override
+    public GatewayResponse<WorldpayDeleteTokenResponse> deleteStoredPaymentDetails(DeleteStoredPaymentDetailsGatewayRequest request) throws GatewayException {
+        GatewayClient.Response response = deleteTokenClient.postRequestFor(
+                gatewayUrlMap.get(request.getGatewayAccount().getType()),
+                WORLDPAY,
+                request.getGatewayAccount().getType(),
+                buildDeleteTokenOrder(request),
+                getGatewayAccountCredentialsAsAuthHeader(request.getGatewayCredentials(), null));
```
stephencdaly: We should use our new method in AuthUtils for getting the authorization header for managing tokens

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -343,6 +349,17 @@ public GatewayResponse<BaseCancelResponse> cancel(CancelGatewayRequest request)
...
+        return getWorldpayDeleteTokenGatewayResponse(response);
```
stephencdaly: I don't think this method actually needs to return anything as we don't use the response. For Stripe we won't return anything from the method that makes the request - we know its successful if no exception is thrown. So I think this method and `paymentProvider#deleteStoredPaymentDetails` can both return `void`

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
stephencdaly: We've added a new method for getting the authorisation header - we need one for getting the merchant code without having to provide and authorisation mode and need to call it from here too.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -343,6 +349,17 @@ public GatewayResponse<BaseCancelResponse> cancel(CancelGatewayRequest request)
         return getWorldpayGatewayResponse(response);
     }
 
+    @Override
+    public GatewayResponse<WorldpayDeleteTokenResponse> deleteStoredPaymentDetails(DeleteStoredPaymentDetailsGatewayRequest request) throws GatewayException {
```
stephencdaly: I think we need a non-"contract" test for this that checks that we build the XML to send to Worldpay correctly. The "contract" test isn't run on CI and is just for locally verifying changes work with the actual Worldpay gateway. We need a test that will run on CI to provide proper coverage.

`src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java` would be an appropriate place for this, and there are other tests there that do assertions on the XML sent to Worldpay

stephencdaly (changes requested):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
james-peacock-ho: this has now been added

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
james-peacock-ho: name updated

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
james-peacock-ho: done

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
james-peacock-ho: done

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
james-peacock-ho: amended as requested

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
james-peacock-ho: tests added for payload and headers

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -615,6 +653,27 @@ void assert_authorization_header_is_passed_to_gateway_client_when_using_recurrin
...
+        
+        verify(deleteTokenClient).postRequestFor(
+                eq(WORLDPAY_URL),
+                eq(WORLDPAY), eq("test"),
+                any(GatewayOrder.class),
+                headers.capture());
+
+        assertThat(headers.getValue().size(), is(1));
+        String expectedHeader = "Basic " + Base64.getEncoder().encodeToString(new String("worldpay-password" + ":" + "password").getBytes());
+        assertThat(headers.getValue().get(AUTHORIZATION), is(expectedHeader));
```
stephencdaly, line 674: As the setup for this test is the same as for the other test you've added - my opinion would be that its better to just add the additional assertions to the same test.

stephencdaly (approved):
LGTM!

PR #4388, opened 2023-03-13
PP-10711 add idempotency comparator util

SandorArpa:
## WHAT YOU DID
- add a new util to compare ChargeCreateRequest to an idempotency request body
- add exhaustive tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/util/ChargeCreateRequestIdempotencyComparatorUtil.java
+++ b/src/main/java/uk/gov/pay/connector/charge/util/ChargeCreateRequestIdempotencyComparatorUtil.java
@@ -1,0 +1,18 @@
+package uk.gov.pay.connector.charge.util;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import uk.gov.pay.connector.charge.model.ChargeCreateRequest;
+
+import java.util.Map;
+
+public class ChargeCreateRequestIdempotencyComparatorUtil {
+    private static final ObjectMapper mapper = new ObjectMapper().registerModule(new Jdk8Module());
+    public static boolean compare(ChargeCreateRequest request, Map<String, Object> idempotencyRequestBody) {
+        JsonNode requestNode = mapper.valueToTree(request);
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/util/ChargeCreateRequestIdempotencyComparatorUtil.java
```diff
@@ -0,0 +1,18 @@
...
+import uk.gov.pay.connector.charge.model.ChargeCreateRequest;
+
+import java.util.Map;
+
+public class ChargeCreateRequestIdempotencyComparatorUtil {
+    private static final ObjectMapper mapper = new ObjectMapper();
+    public static boolean compare(ChargeCreateRequest request, Map<String, Object> idempotencyRequestBody) {
+        JsonNode requestNode = mapper.valueToTree(request);
+        ChargeCreateRequest idempotencyRequest = new ObjectMapper().convertValue(idempotencyRequestBody, ChargeCreateRequest.class);
+        JsonNode idempotencyNode = mapper.valueToTree(idempotencyRequest);
```
stephencdaly:
```suggestion
        JsonNode idempotencyNode = mapper.valueToTree(idempotencyRequestBody);
```
I think we can just do this directly from the map rather than having to read it into a ChargeCreateRequest first

src/test/java/uk/gov/pay/connector/charge/util/ChargeCreateRequestIdempotencyComparatorUtilTest.java
```diff
@@ -0,0 +1,147 @@
...
+                .withAgreementId("an agreement")
+                .withExternalMetadata(new ExternalMetadata(
+                        Map.of("key1", "value1", "key2", 2023)))
+                .build();
+
+        assertThat(ChargeCreateRequestIdempotencyComparatorUtil.compare(chargeCreateRequest, idempotencyMap), is(false));
+    }
+
+    @Test
+    void shouldReturnFalse_whenReferenceIsDifferent() {
```
stephencdaly, line 109: I wonder whether we need quite so many tests for this now that we're using the JsonNode equals method to do the comparison. Maybe we just need tests for:
- a top level field being different
- metadata being different
- fields in a different order
Although maybe even that's overkill as we're kind of testing the JsonNode library by doing that 🤔

stephencdaly (approved):

PR #4391, opened 2023-03-14
PP-10710 rename local vars

SandorArpa:
## WHAT YOU DID
- forgot to rename local vars when renaming class.


james-peacock-ho (commented):
are you sure pGobject, not pgObject?

SandorArpa:
Yes, the class is called `PGobject` :)
See https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/util/JsonToStringStringMapConverter.java

james-peacock-ho (approved):
LGTM

PR #4392, opened 2023-03-14
PP-10513 Refactor DeleteStoredPaymentDetailsGatewayRequest

stephencdaly:
Don't implement the GatewayRequest interface. This interface isn't directly referenced anywhere so there is no need to implement it, it seems it was just created for code organisation purposes. There is no AuthorisationMode associated with this request, so it doesn't make sense implementing the getAuthorisationMode interface method.

Have getters for the individual properties of the Agreement, Payment Instrument and Gateway Account. The intention of the GatewayRequest classes is to pass information to the PaymentProvider code level. It looks like we try to only allow access to the individual bits of information that the PaymentProvider code needs rather than the entire database entities. This is likely to make sure that the PaymentProvider code does not modify the database entities, as this is outside of its concern.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntityFixture.java
+++ b/src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntityFixture.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.gatewayaccount.model;
 
 import uk.gov.pay.connector.cardtype.model.domain.CardTypeEntity;
+import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState;
 import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialsEntity;
 import uk.gov.pay.connector.usernotification.model.domain.EmailNotificationEntity;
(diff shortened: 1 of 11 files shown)


src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntityFixture.java
```diff
@@ -173,6 +175,16 @@ public GatewayAccountEntityFixture withGatewayAccountCredentials(List<GatewayAcc
         this.gatewayAccountCredentialsEntities = credentials;
         return this;
     }
+    
+    public GatewayAccountEntityFixture withActiveStipeGatewayAccountCredentials() {
```
james-peacock-ho: typo - Stripe

james-peacock-ho (commented):
see typo on withActiveStipeGatewayAccountCredentials

otherwise looks great

src/test/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntityFixture.java
stephencdaly: thanks, fixed

james-peacock-ho (approved):
LGTM!

PR #4394, opened 2023-03-14
PP-10798 Add agreement_external_id column to payment_instruments

stephencdaly:
- We are adding `agreement_external_id` rather than `agreement_id` as we have the idea that in future, agreements should live outside of connector if we add more payment method with different "connector" microservices.
- Add foreign key constraint and index.


SandorArpa (approved):
LGTM

PR #4397, opened 2023-03-15
PP-10794 log provider error codes

james-peacock-ho:
- Decline/Refusal codes from providers should be mapped to a common list of rejection reasons
- In this first iteration we are focusing on the most common rejection codes
- Create WorldpayRejectionCodeMapper to map WP refusal codes to a common MappedRejectionReason enum
- Add mapped rejection reason to stringified WP response so that it is included in logs
- Add unit tests
- Stripe decline reasons will be mapped in a separate PR


src/main/java/uk/gov/pay/connector/gateway/model/MappedDeclineReason.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.pay.connector.gateway.model;
+
+public enum MappedDeclineReason {
```
alexbishop1: How attached are we to the term ‘declined’? We tend to use ’rejected’ (as in `AUTHORISATION REJECTED`) in connector code.

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayDeclineCodeMapper.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+    
+    public static WorldpayDeclineCodeMapper fromErrorCode(String worldpayDeclineCode) {
+        for (WorldpayDeclineCodeMapper wep : values()) {
+            if (StringUtils.equals(wep.getErrorCode(), worldpayDeclineCode)) {
+                return wep;
+            }
+        }
+        return WorldpayDeclineCodeMapper.WORLDPAY_UNCATEGORISED;
+    }
```
alexbishop1: We could put create a static `Map<String, String>` (from error code to enum) and then look things up in that.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -250,6 +251,7 @@ public String toString() {
             joiner.add("lastEvent: " + lastEvent);
         }
         if (isNotBlank(getRefusedReturnCode())) {
+            joiner.add("Mapped error description: " + WorldpayDeclineCodeMapper.fromErrorCode(getRefusedReturnCode()).toMappedDeclineReason().name());
```
alexbishop1: Minor point but I think this logically makes sense to be placed after the `ISO8583ReturnCode` (since it’s what the return code is categorised to).

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/gateway/model/MappedDeclineReason.java
james-peacock-ho: Stripe calls them 'decline codes' and Worldpay calls them 'refusal codes'. Originally this ticket was talking about 'error codes' but I think that may be misleading. I'm not sure there is an option that prevent any chance of confusion, but I agree if rejection is common across connector, that would make sense.

james-peacock-ho: renamed as suggested

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayDeclineCodeMapper.java
james-peacock-ho: Done (used <String, enum>)

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
james-peacock-ho: adjusted as suggested.

src/main/java/uk/gov/pay/connector/gateway/model/MappedRejectionReason.java
```diff
@@ -0,0 +1,30 @@
+package uk.gov.pay.connector.gateway.model;
+
+public enum MappedRejectionReason {
```
alexbishop1: Maybe `MappedAuthorisationRejectedReason’? ‘Rejection’ on its own is a little ambiguous.

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayRejectionCodeMapper.java
```diff
@@ -0,0 +1,87 @@
...
+    WORLDPAY_INVALID_ACCOUNT_NUMBER("141", INVALID_ACCOUNT_NUMBER),
+    WORLDPAY_STOP_PAYMENT_ORDER("972", STOP_PAYMENT_ORDER),
+    WORLDPAY_REVOCATION_OF_ALL_AUTHORISATION_ORDER("973", REVOCATION_OF_ALL_AUTHORISATION),
+    WORLDPAY_REVOCATION_OF_AUTHORISATION("975", REVOCATION_OF_AUTHORISATION);
+    
+    private final String errorCode;
+    private final MappedRejectionReason mappedRejectionReason;
+
+    private static Map<String, WorldpayRejectionCodeMapper> worldpayRejectionCodeMap = 
+            Stream.of(values()).collect(toMap(WorldpayRejectionCodeMapper::getErrorCode, e -> e));
```
alexbishop1: Should be able to use `toUnmodifiableMap()`. Also, can we have more descriptive variable names than `e`?

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayRejectionCodeMapper.java
```diff
@@ -0,0 +1,87 @@
...
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.REVOCATION_OF_ALL_AUTHORISATION;
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.REVOCATION_OF_AUTHORISATION;
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.STOLEN_CARD;
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.STOP_PAYMENT_ORDER;
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.SUSPECTED_FRAUD;
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.TRANSACTION_NOT_PERMITTED;
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.TRY_ANOTHER_CARD;
+import static uk.gov.pay.connector.gateway.model.MappedRejectionReason.UNCATEGORISED;
+
+public enum WorldpayRejectionCodeMapper {
```
alexbishop1: Now we have the map, I’m not sure I can see the benefit of having this being an enum. Maybe it could just be a regular class that contains the map. It could even possibly be a map inside the `WorldpayOrderStatusResponse` class if that‘s the only place it’s going to be used and it doesn’t clutter the file too much or make testing difficult.

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayRejectionCodeMapper.java
james-peacock-ho: WorldpayOrderStatusResponse is already 300 lines so I think it may be easier if it's outside. If it is a map in a class, I think we go back to my earlier question of how best to build the map, in a static block inside the WorldpayRejeectionCodeMapper class, or in a separate properties file?

alexbishop1: Just a `private static Map MY_STILL_AMAZING_MAP Map.ofEntries(…)` should be fine. No need for a properties file.

src/main/java/uk/gov/pay/connector/gateway/model/MappedRejectionReason.java
james-peacock-ho: updated

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayRejectionCodeMapper.java
james-peacock-ho: updated

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayAuthorisationRejectedCodeMapper.java
alexbishop1, line 1: This is simpler.

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/gateway/model/WorldpayAuthorisationRejectedCodeMapper.java
james-peacock-ho: it certainly is, thanks for your help.

PR #4401, opened 2023-03-16
PP-10798 Set agreement_external_id on payment instrument

stephencdaly:
When an agreement is set up, set the `agreement_external_id` on the payment instrument. This is done at the same time we record the `payment_instrument_id` on the agreement.

By adding the `agreementExternalId` field to the PaymentInstrumentEntity this means it will be included in API responses that include the PaymentInstrumentEntity_FrontendView schema. There probably isn't an reason this shouldn't be returned, so not adding the @JsonIgnore field.


SandorArpa (approved):
LGTM

PR #4403, opened 2023-03-16
PP-10798 Cancel old payment instruments when replaced on agreement

stephencdaly:
For a charge created to set up an agreement, when marking the charge as eligible for capture:
- Find any payment instruments that have status ACTIVE for the agreement
- Update the state of these payment instruments to CANCELLED
- Add a message to the Task queue to delete the payment details for the payment instrument held with the PSP.


james-peacock-ho (approved):
LGTM!

PR #4404, opened 2023-03-16
PP-10711 implement handling idempotency key

SandorArpa:
## WHAT YOU DID
- add HeaderParam Idempotency-Key
- refactor ChargeService to pass on idempotency key
- refactor createCharge method to check for existing idempotency entry when AuthorisationMode is AGREEMENT and idempotency key is present

- if idempotency entry exists:
	- throw error (409) if request is different
	- return ChargeEntity if request matches

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -1,5 +1,9 @@
 package uk.gov.pay.connector.charge.service;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import com.google.common.collect.Maps;
 import com.google.inject.persist.Transactional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -16,6 +20,7 @@ import uk.gov.pay.connector.charge.exception.AgreementMissingPaymentInstrumentEx
 import uk.gov.pay.connector.charge.exception.AgreementNotFoundBadRequestException;
 import uk.gov.pay.connector.charge.exception.ChargeNotFoundRuntimeException;
 import uk.gov.pay.connector.charge.exception.GatewayAccountDisabledException;
+import uk.gov.pay.connector.charge.exception.IdempotencyKeyUsedException;
 import uk.gov.pay.connector.charge.exception.IncorrectAuthorisationModeForSavePaymentToAgreementException;
 import uk.gov.pay.connector.charge.exception.MissingMandatoryAttributeException;
 import uk.gov.pay.connector.charge.exception.MotoPaymentNotAllowedForGatewayAccountException;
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -322,8 +347,20 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
             }
 
             chargeDao.persist(chargeEntity);
+
+            if (authorisationMode == AGREEMENT && idempotencyKey != null) {
+                IdempotencyEntity idempotencyEntity = new IdempotencyEntity(
+                        idempotencyKey,
+                        gatewayAccount,
+                        chargeEntity.getExternalId(),
+                        mapper.convertValue(chargeRequest, new TypeReference<>() {}),
```
sfount, line 356: Doesn't need to happen here but I wonder if this mapping is something that should happen in the idempotency entity so that each file/ service that uses it doesn't have to know about object mappers and this specific syntax.

Something like 

```java
public static IdempotencyEntity from(String idempotencyKey, ChargeEntity charge, Object request) {
  // ...
}
```

I suspect this would only be worth doing if we need to implement idempotency on any other endpoint and have this instantiation be shared.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -314,6 +326,19 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
                     .ifPresent(chargeEntity::setCardDetails);
 
             if (authorisationMode == AuthorisationMode.AGREEMENT) {
+                if (idempotencyKey != null) {
+                    Optional<IdempotencyEntity> optionalIdempotencyEntity = idempotencyDao.findByGatewayAccountIdAndKey(gatewayAccount.getId(), idempotencyKey);
+                    if (optionalIdempotencyEntity.isPresent()) {
+                        IdempotencyEntity idempotencyEntity = optionalIdempotencyEntity.get();
+                        if (ChargeCreateRequestIdempotencyComparatorUtil.compare(chargeRequest, idempotencyEntity.getRequestBody())) {
+                            LOGGER.info("Idempotency-Key was already used to create a request with matching values {}", idempotencyKey);
```
sfount, line 334: Structured logging with `kv()` would future proof these logs, I suspect there might be places we'd want to include an `idempotency_key: value` as part of the MDC to be logged on the full request.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -314,6 +326,19 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
...
+                            // TODO implement PP-10833, query Ledger if Charge has been expunged
+                            return findChargeByExternalId(idempotencyEntity.getResourceExternalId());
+                        }
+                        LOGGER.info("Idempotency-Key already exist with different values {}", idempotencyKey);
```
sfount, line 338:
```suggestion
                        LOGGER.info("Idempotency-Key already exists with different values {}", idempotencyKey);
```

sfount (approved):
Change looks good to me, some very minor comments but I think they'd mostly be relevant for future changes

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: Yes, that's the next iteration.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: 👍 
will fix in next iteration

PR #4407, opened 2023-03-17
PP-10711 Add logging for idempotency

SandorArpa:
## WHAT YOU DID
- add new method to create structured logs for logging difference between existing and incoming charge requests for a given idempotency key

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -3,8 +3,9 @@ package uk.gov.pay.connector.charge.service;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
-import com.google.common.collect.Maps;
+import com.google.common.collect.MapDifference;
 import com.google.inject.persist.Transactional;
+import net.logstash.logback.argument.StructuredArgument;
 import org.slf4j.Logger;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -330,12 +336,14 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
                     Optional<IdempotencyEntity> optionalIdempotencyEntity = idempotencyDao.findByGatewayAccountIdAndKey(gatewayAccount.getId(), idempotencyKey);
                     if (optionalIdempotencyEntity.isPresent()) {
                         IdempotencyEntity idempotencyEntity = optionalIdempotencyEntity.get();
-                        if (ChargeCreateRequestIdempotencyComparatorUtil.compare(chargeRequest, idempotencyEntity.getRequestBody())) {
+                        Map<String, MapDifference.ValueDifference<Object>> diffMap = ChargeCreateRequestIdempotencyComparatorUtil.diff(chargeRequest, idempotencyEntity.getRequestBody());
```
sfount: I think a clear compare method which returns a boolean is much easier to follow here, the flow of the code is a bit harder to follow where a diff/ map is being constructed. Maybe this could be exposed as a separate accessor on the utility if we wanted to record that in the logs

sfount (commented):
Using structured logging around this flow looks like a good change, a few questions about the interface

sfount (approved):
`equals` and `diff` interface looks really clear, code looks good to me 👍

PR #4408, opened 2023-03-17
PP-10711 refactor creation of idempotency entity

SandorArpa:
## WHAT YOU DID
- add a `from` method that converts ChargeCreateRequest to a map.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/idempotency/model/IdempotencyEntity.java
+++ b/src/main/java/uk/gov/pay/connector/idempotency/model/IdempotencyEntity.java
@@ -1,5 +1,9 @@
 package uk.gov.pay.connector.idempotency.model;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import uk.gov.pay.connector.charge.model.ChargeCreateRequest;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/idempotency/model/IdempotencyEntity.java
```diff
@@ -25,6 +29,7 @@
 @SequenceGenerator(name = "idempotency_id_seq",
         sequenceName = "idempotency_id_seq", allocationSize = 1)
 public class IdempotencyEntity {
+    private static final ObjectMapper mapper = new ObjectMapper().registerModule(new Jdk8Module());
```
sfount, line 32: As I understand it this mapper will only ever be instantiated once and then shared between any instances of the idempotency entity class. I've not seen this pattern used too commonly with other classes either using dependency injection or pulling in a shared dropwizard mapper

sfount (approved):
Encapsulating the logic further down so that each controller/ service that uses idempotency doesn't have to figure out object mappers feels like a step in the right direction. Not 100% sure on the pattern but happy to chat about it

src/main/java/uk/gov/pay/connector/idempotency/model/IdempotencyEntity.java
stephencdaly: The ObjectMapper needs to register the `Jdk8Module` so it serialises Optionals how we need it to. Otherwise it serialises them as `{"present":true}`.

We might be fine registering this module on the shared ObjectMapper. But if we're not confident - we might need to create this separate ObjectMapper for use here. I'm not sure too much about the implications of having multiple ObjectMappers...

PR #4411, opened 2023-03-20
PP-10794 Map Stripe error codes

james-peacock-ho:
When authorisation is rejected by a payment provider, logs should include a non-payment-provider-specific rejection reason
- Add mapper for Stripe decline codes
- Update mapped authorisation rejected reasons
- Sorted mapped authorisation rejected enum values for readability
- Add unit tests


stephencdaly (approved):
LGTM!

PR #4413, opened 2023-03-21
PP-10711 Refactor idempotency check

stephencdaly:
- Do idempotency check in ChargesApiResource before attempting to create a charge
- If the idempotency check finds an entry with the same key and matching request body, return the charge response with a 200 status code rather that a 201
- Add `@JsonIgnore` annotations on the getters of ChargeCreateRequest and `@JsonInclude(JsonInclude.Include.NON_NULL)` on the class. This means that we will not serialise nullable properties on the request that have default values returned by the getters. This will mean that if we add new optional properties to the ChargeCreateRequest, the equality check between the request stored in the `idempotency` table before the property is added will still be deemed to be equal to a newly made request.


src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
```diff
@@ -141,34 +146,42 @@ public ChargeCreateRequest() {
         this.authorisationMode = authorisationMode;
     }
 
+    @JsonIgnore
```
kbottla, line 149: Is there any need to add `@JsonIgnore` to getters given properties are annotated with `@JsonProperty`?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
```diff
@@ -150,6 +152,140 @@ public void shouldCreatePaymentWithAgreementIdAndAuthorisationModeAgreement() {
...
+        AddAgreementParams agreementParams = anAddAgreementParams()
+                .withGatewayAccountId(accountId)
+                .withExternalAgreementId(JSON_VALID_AGREEMENT_ID_VALUE)
+                .withPaymentInstrumentId(paymentInstrumentId)
+                .build();
+        databaseTestHelper.addAgreement(agreementParams);
+
+        String existingChargeExternalId = addCharge(CREATED);
+
+        // IMPORTANT: Do not modify this request body if the test fails. This is used to verify that when making a
```
kbottla: Should this comment be here https://github.com/alphagov/pay-connector/pull/4413/files?diff=unified&w=1#diff-83f5646a82f5199686bcf25941ee09f764dfa5b9b9baea7169fc111caaf5db12R219 given it has all properties.

src/main/java/uk/gov/pay/connector/charge/model/ChargeCreateRequest.java
stephencdaly: Yeh, Jackson will use the getters by default. Without the`@JsonIgnore` it will include all the defaults we set in the getters, and also serialise Optionals as `{present: true}` unless we register the Jdk8Module on the ObjectMapper (which we were doing previously but it not necessary with `@JsonIgnore`.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateAgreementIT.java
stephencdaly: Yeh probably should actually, will add something

stephencdaly: Have added a comment

kbottla (approved):

PR #4417, opened 2023-03-24
ZD-5266603 Remove obsolete sandbox test card number

alexbishop1:
Remove obsolete sandbox test card number since now we have a new card number that identifies in the same way.


nlsteers (approved):
👍

sfount (approved):

PR #4420, opened 2023-03-27
PP-10865 remove smartpay part 1

SandorArpa:
## WHAT YOU DID
- remove Smartpay related code. Smartpay is no longer supported and it's unnecessary waste of time to refactor its classes every time we do a change to other PPS.


stephencdaly (approved):
🗑

PR #4454, opened 2023-04-13
PP-10715 Add idempotency key expiry to test config

james-peacock-ho:
- Idempotency keys expire after a set time (currently 24 hours)
- Add the corresponding environment variable to test configs
- Update unit test to add additional verification


SandorArpa (approved):
LGTM

PR #4460, opened 2023-04-14
PP-10712 Handle race condition for idempotency

stephencdaly:
Handle the potential race condition when two requests to create a charge with the same Idempotency Key are made at about the same time, resulting in an entry in the idempotency table not existing when the check is made prior to creating a charge. In this case, one request will commit the database transaction first, resulting in a RollbackException being thrown when the other request attempts to commit its transaction to insert an entry in the `idempotency` table which violates the UNIQUE criteria.

Handle the RollbackException by responding to the request with a 409 and "error_identifier": "IDEMPOTENCY_KEY_USED".

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -94,6 +94,7 @@ import uk.gov.service.payments.commons.model.AuthorisationMode;
 import uk.gov.service.payments.commons.model.charge.ExternalMetadata;
 
 import javax.inject.Inject;
+import javax.persistence.RollbackException;
 import javax.ws.rs.core.UriBuilder;
 import javax.ws.rs.core.UriInfo;
 import java.net.URI;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -259,10 +260,20 @@ private ChargeStatus internalChargeStatus(String code) {
...
-                );
+        try {
+            return createCharge(chargeRequest, accountId, idempotencyKey)
+                    .map(charge ->
+                            populateResponseBuilderWith(aChargeResponseBuilder(), uriInfo, charge).build()
+                    );
+        } catch (RollbackException e) {
+            // This handles the case when there is a race between two requests attempting to create a charge with the same 
+            // idempotency key
+            if (idempotencyKey != null && idempotencyDao.findByGatewayAccountIdAndKey(accountId, idempotencyKey).isPresent()) {
```
alexbishop1: Maybe extract the `idempotencyKey != null && idempotencyDao.findByGatewayAccountIdAndKey(accountId, idempotencyKey).isPresent()` into a private boolean-returning method called `idempotencyKeyAlreadyUsedToCreateCharge()` or something so we can avoid the comment?

alexbishop1 (changes requested):
Looks good! One suggestion to maybe avoid needing a comment.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
stephencdaly: The reason I thought a comment was appropriate is that it explains how we can get into this situation - but maybe with that method name it is self explanatory. Happy to go with your suggestion

stephencdaly: Actually, the log line explains what's happening. I added the comment before the log line. I think I'll just remove the comment..

stephencdaly: comment removed

alexbishop1 (approved):

PR #4466, opened 2023-04-17
PP-10794 Add more common Worldpay refusal codes to map

james-peacock-ho:
- Context: Where possible we are mapping the payment processor decline codes to a common MappedAuthorisationRejectionReason
- Map some more frequently-occurring worldpay codes


src/main/java/uk/gov/pay/connector/gateway/model/WorldpayAuthorisationRejectedCodeMapper.java
```diff
@@ -48,7 +50,10 @@ public class WorldpayAuthorisationRejectedCodeMapper {
                 entry("59", SUSPECTED_FRAUD),
                 entry("61", EXCEEDS_WITHDRAWAL_AMOUNT_LIMIT),
                 entry("65", AUTHENTICATION_REQUESTED),
-                entry("141", INVALID_ACCOUNT_NUMBER));
+                entry("78", INVALID_ACCOUNT_NUMBER),
+                entry("91", ISSUER_TEMPORARILY_UNAVAILABLE),
+                entry("141", INVALID_ACCOUNT_NUMBER),
```
alan-gds: INVALID_ACCOUNT_NUMBER is now mapped twice; is this correct?  I couldn't see anything for code 141 in https://developer.worldpay.com/docs/access-worldpay/reference/useful-tables, although that's actually the mapping that we already had (and this commit has added a mapping for 78).  Is it correct that we now have two mappings for this rejection reason?

james-peacock-ho: well spotted. I've looked back and discovered that 141 was added incorrectly due to a typo in the notes of a conversation with Worldpay. I have now removed it. In theory we could have several payment provider codes mapped to the same MappedAuthorisationRejectedReason if their meaning and effect is the same for our purposes, but having removed the erroneous one this is no longer relevant.

alan-gds (approved):
LGTM

PR #4468, opened 2023-04-18
PP-10831 Support cancelling payment instruments for sandbox

james-peacock-ho:
- Context: Deleting stored payment details tasks, such as when an agreement is cancelled, should not throw an exception when the payment provider is Sandbox, and no further action needs to be taken.
- Add empty implementation of deleteStoredPaymentDetails in SandboxPaymentProvider (to override the default implementation that throws an exception)
- Delete the DeleteStoredPaymentDetailsTaskHandlerIT test class, which contained a single test to ensure an exception is thrown when the above method has not been implemented for a particular payment provider - this is now redundant as the method has been implemented for Worldpay, Stripe and Sandbox.


alan-gds (approved):
Personally I would be in favour of keeping the test, assuming no extra work would be needed.  That's really not a big deal though, and I'm happy to go along with what you and Stphen D agreed.  Also, if retaining the test *would* mean extra work was needed, then it definitely makes sense to delete it.

PR #4472, opened 2023-04-20
PP-10830 Add agreement_external_id column to charges

james-peacock-ho:
The agreement_id column on the charges table stores the agreement external_id rather than the internal ID. We need to rename agreement_id → agreement_external_id to make this clearer:
- Add agreement_external_id column to charges table.
- Subsequent PRs will address the code changes and migrations necessary to safely transfer values to the new column.


SandorArpa (approved):
LGTM

PR #4477, opened 2023-04-21
PP-10830 Save agreement_external_id in charges.agreement_external_id

james-peacock-ho:
The agreement_id column on the charges table stores the agreement external_id rather than the internal ID.
We need to rename agreement_id → agreement_external_id to make this clearer.
In the previous PR we added an agreement_external_id column to the charges table.

When a charge is created for an agreement, we need to save the agreement external id to the new agreement_external_id column, in addition to the agreement_id column:
- Add agreementExternalEntity to ChargeEntity.
- Unit test and integration test to cover.

Subsequent PRs will deal with copying pre-existing agreement ids to the new agreement_external_id column, code changes to read from the new column, and dropping the old column.


SandorArpa (approved):
LOKTM

PR #4483, opened 2023-04-25
PP-10830 Stop using agreement_id column in charges

james-peacock-ho:
Context: We are renaming charges.agreement_id to charges.agreement_external_id.
We have already created the agreement_external_id column and copied values across.
We have already changed code so that the external id is saved to both columns.
We have already changed code so that it reads the external id from the new column

- Stop saving agreement external id to the old agreement_id column
- In ChargeEntity and associated tests, refer to the agreement external id as agreementExternalId
- In ChargeEntity and associated tests, refer to the agreement entity as agreementEntity
- Remove redundant tests that existed solely for the short-term situation where agreement external id was being saved to both columns

Subsequent PR will deal with deletion of the old agreement_id column.


SandorArpa (approved):
LGTM

PR #4487, opened 2023-04-26
PP-10830 Drop agreement_id column from charges

james-peacock-ho:
Context: We are renaming charges.agreement_id to charges.agreement_external_id.
We have already created the new column and copied agreement external id values to it.
We have already changed code so that it reads from and writes to the new column.
We have already removed references in code to the old column.

- Drop agreement_id column


SandorArpa (approved):
LGTM

PR #4493, opened 2023-04-28
PP-10958 Add credential_id to charge create request

kbottla:
## WHAT YOU DID
- Allows `credential_id` on create charge request so credential_id can be specified to create a charge when service is switching from and to the same provider.


stephencdaly (approved):
LGTM!

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -363,6 +359,31 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
...
+                    chargeRequest.getCredentialId(), gatewayAccount.getId());
+
+            if (chargeRequest.getPaymentProvider() != null
+                    && !chargeRequest.getPaymentProvider().equals(gatewayAccountCredential.getPaymentProvider())) {
+                throw new CredentialsNotFoundBadRequestException(
+                        format("Credentials not found for credential_id [%s] and payment_provider [%s]",
+                                chargeRequest.getCredentialId(),
+                                chargeRequest.getPaymentProvider())
+                );
+            }
```
alexbishop1, line 375: Maybe it’s a bit early in the morning but this makes my head hurt! Could find method on `gatewayAccountCredentialsService` take a credential ID, gateway account ID and the expected payment provider? Then we’d just need to check if the returned optional contains anything. That might then make it possible to have some early returns, reducing the amount of nested if statements.

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
```diff
@@ -1004,4 +1004,38 @@ void shouldNotUpdateNonExistentCredentials() {
...
+            assertThat(credentialExternalId, is(credentialsEntity.getExternalId()));
+        }
+
+        @Test
+        void shouldThrowExceptionIfCredentialNotFoundForExternalId() {
+            String credentialExternalId = "credential-external-id";
+            when(mockGatewayAccountCredentialsDao.findByExternalIdAndGatewayAccountId(credentialExternalId, 1L))
+                    .thenReturn(Optional.empty());
+
+            CredentialsNotFoundBadRequestException exception = assertThrows(CredentialsNotFoundBadRequestException.class, () -> {
```
alexbishop1, line 1034: If there was ever a time for `var`, I think this is it!

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateProviderCredentialsIT.java
```diff
@@ -67,7 +70,145 @@ public void shouldCreateChargeForProvidedPaymentProvider() {
...
+                .withAccountId(accountId)
+                .build();
+        databaseTestHelper.addGatewayAccount(gatewayAccountParams);
+
+        String postBody = toJson(Map.of(
+                JSON_AMOUNT_KEY, AMOUNT,
+                JSON_REFERENCE_KEY, JSON_REFERENCE_VALUE,
+                JSON_DESCRIPTION_KEY, JSON_DESCRIPTION_VALUE,
+                JSON_RETURN_URL_KEY, RETURN_URL,
+                JSON_CREDENTIAL_ID_KEY, credentialsToUse.getExternalId()
```
alexbishop1, line 96: Under what circumstances do we expect a credential ID to be specified but a payment provider not to be?

alexbishop1 (changes requested):
Just a couple of things.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateProviderCredentialsIT.java
kbottla: There is no use case right now and as there is now a credential_id I am going to remove payment_provider from the endpoint

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
kbottla: I am going to remove payment_provider so this would be simplified.

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsServiceTest.java
kbottla: Not a big fan of `var`. It is not clear what that the type will be if not using IDEs

kbottla: Although in this case it should be fine to use var

PR #4497, opened 2023-05-02
PP-10958 Add pact state to create charge by credential_id

kbottla:
## WHAT YOU DID
- Add pact state to create gateway account credential with specific external_id so that a charge can be created by consumer pact test with credential_id


james-peacock-ho (approved):
LGTM

PR #4500, opened 2023-05-03
PP-10948 Augment MappedAuthorisationRejectedReason with can retry

alexbishop1:
Update the `MappedAuthorisationRejectedReason` enum so that it has a new boolean-bearing `canRetry()` method, which states whether or not the payment can be retried (that is, if there is some chance that an attempt to collect another payment using the same stored token may succeed).

The data comes from Worldpay and Stripe, who told us which of their decline codes indicate that a payment cannot be retried. For `PICKUP_CARD` (Worldpay 7, Stripe `pickup_card`) and `TRANSACTION_NOT_PERMITTED` (Worldpay 57, Stripe `transaction_not_allowed`), Worldpay told us the payment cannot be retried but Stripe did not: we allow these to be retried for now.

`BaseAuthoriseResponse` and its implementers now have an optional-bearing `getMappedAuthorisationRejectedReason()` method. If the `AuthoriseStatus` for the `BaseAuthoriseResponse` object is `AuthoriseStatus.REJECTED`, this new method will return a `MappedAuthorisationRejectedReason` calculated by mapping the decline code from the payment gateway, or `MappedAuthorisationRejectedReason.UNCATEGORISED` if there is no mapping. If the `AuthoriseStatus` is anything else, it will return an empty optional. Thus, `getMappedAuthorisationRejectedReason()` returns a non-empty optional if and only if the authorisation was rejected.

In line with this, the post-authorisation logging for Worldpay payments has now changed such that a
`MappedAuthorisationRejectedReason` will only be logged if the authorisation was rejected. Previously, if the Worldpay response included an ISO 8583 return code (there are codes for non-rejected authorisations; it is unclear if Worldpay ever return them), the logging would always include a `MappedAuthorisationRejectedReason` (which may have been
`MappedAuthorisationRejectedReason.UNCATEGORISED`).

The Stripe logging already only logs a `MappedAuthorisationRejectedReason` if the Stripe response includes a decline code and Stripe’s docs indicate a decline code will only be present for authorisations that failed due to a card decline, so no equivalent change had to be made here.

The `BaseAuthoriseResponse` object is available to `ChargeAuthorisationService.updateChargePostAuthorisation(…)`, which will allow whether or not the payment can be retried to be persisted etc. in the future.


SandorArpa (approved):
LGTM

PR #4507, opened 2023-05-09
PP-10979 Add `can_retry` column to charges table

kbottla:
## WHAT YOU DID
- Added `can_retry` column to charges table to store retry status(true/false) for recurring payments declined.


SandorArpa (approved):
LGTM

PR #4508, opened 2023-05-09
PP-10949 add agreement inactivated event model

SandorArpa:
## WHAT YOU DID
- add a new event model, `AgreementInactivated` and its EventDetails implementation
- extend a test to include the new event model

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/agreement/AgreementInactivated.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/agreement/AgreementInactivated.java
@@ -1,0 +1,45 @@
+package uk.gov.pay.connector.events.model.agreement;
+
+import uk.gov.pay.connector.agreement.model.AgreementEntity;
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+import uk.gov.pay.connector.gateway.model.MappedAuthorisationRejectedReason;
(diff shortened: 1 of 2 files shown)


stephencdaly (approved):

src/main/java/uk/gov/pay/connector/events/model/agreement/AgreementInactivated.java
```diff
@@ -0,0 +1,45 @@
...
+                agreement.getServiceId(),
+                agreement.getGatewayAccount().isLive(),
+                agreement.getExternalId(),
+                new AgreementInactivated.AgreementInactivatedEventDetails(agreement.getPaymentInstrument().orElse(null), mappedReason),
+                timestamp
+        );
+    }
+
+    static class AgreementInactivatedEventDetails extends EventDetails {
+        private String paymentInstrumentExternalId;
```
kbottla, line 27: Don't think we need `paymentInstrumentExternalId` as it is already included in the AgreementSetUp event.

kbottla: Actually, ignore this. It is possible to have multiple payment instruments.

SandorArpa: 👍

PR #4511, opened 2023-05-10
PP-10979 Persist `can_retry` field post-authorisation

kbottla:
## WHAT YOU DID
- for card payments failed, payment providers return the reason which we map to mapped rejected reason and if the payment can be retried.
- persist this info onto the charge for recurring payments


SandorArpa (approved):
LGTM

PR #4512, opened 2023-05-10
PP-10978 Map rejected reason for sandox payments

kbottla:
## WHAT YOU DID
- Map sandbox rejected reason so when user not present payments are declined, agreements can be inactivated.


SandorArpa (approved):
LOKTM

PR #4515, opened 2023-05-11
PP-10983- Remove JUnit 4 unit tests and replace it with JUnit 5

AbdirizakIdris:
- Updated JUnit4 tests to JUnit 5 equivalent

### DIFF
--- a/src/test/java/uk/gov/pay/connector/tasks/ParityCheckServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/tasks/ParityCheckServiceTest.java
@@ -1,11 +1,11 @@
 package uk.gov.pay.connector.tasks;
 
-import org.junit.Before;
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/pay/connector/tasks/ParityCheckServiceTest.java
```diff
@@ -55,9 +52,9 @@
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUNDED;
 import static uk.gov.pay.connector.refund.model.domain.RefundStatus.REFUND_ERROR;
 import static uk.gov.pay.connector.wallets.WalletType.APPLE_PAY;
+import static uk.gov.service.payments.commons.model.Source.CARD_PAYMENT_LINK;
 
-@RunWith(MockitoJUnitRunner.class)
-public class ParityCheckServiceTest {
+@ExtendWith(MockitoExtension.class)public class ParityCheckServiceTest {
```
alexbishop1: Want a line break in here.

alexbishop1 (changes requested):
Looks good!

There’s a few places where it looks like `public class` has been replaced by `class` and `public void` has been replaced by `void` but spaces have been left before `class` and `void`, which messes up the indentation (most notable where there’s `void` immediately under `@Test`). Could you fix that please?

src/test/java/uk/gov/pay/connector/tasks/ParityCheckServiceTest.java
AbdirizakIdris: Yes good spot Alex.

alexbishop1 (approved):
The alignment now aligns with expectations!

PR #4516, opened 2023-05-11
PP-10949 implement inactivating agreement

SandorArpa:
## WHAT YOU DID
When PSP returns a decline code which cannot be retried, we should inactivate the agreement to indicate that it is no longer valid. The service needs to set up a new payment or create a new agreement in this case.

- implement de-activating the matching payment instrument
- post an `AgreementInactivated` event to Ledger
- add/update relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -64,6 +64,7 @@ import uk.gov.pay.connector.common.model.domain.PrefilledAddress;
 import uk.gov.pay.connector.common.service.PatchRequestBuilder;
 import uk.gov.pay.connector.events.EventService;
 import uk.gov.pay.connector.events.eventdetails.charge.RefundAvailabilityUpdatedEventDetails;
+import uk.gov.pay.connector.events.model.agreement.AgreementInactivated;
 import uk.gov.pay.connector.events.model.charge.Gateway3dsInfoObtained;
 import uk.gov.pay.connector.events.model.charge.PaymentDetailsEntered;
 import uk.gov.pay.connector.events.model.charge.PaymentDetailsSubmittedByAPI;
@@ -73,6 +74,7 @@ import uk.gov.pay.connector.events.model.charge.UserEmailCollected;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
+import uk.gov.pay.connector.gateway.model.MappedAuthorisationRejectedReason;
 import uk.gov.pay.connector.gateway.model.ProviderSessionIdentifier;
 import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
@@ -654,9 +656,11 @@ public class ChargeService {
                                                           Auth3dsRequiredEntity auth3dsRequiredDetails,
                                                           ProviderSessionIdentifier sessionIdentifier,
                                                           AuthCardDetails authCardDetails,
-                                                          Map<String, String> recurringAuthToken, Boolean canRetry) {
+                                                          Map<String, String> recurringAuthToken,
+                                                          Boolean canRetry,
+                                                          String rejectedReason) {
         return updateChargeAndEmitEventPostAuthorisation(chargeExternalId, newStatus, authCardDetails, transactionId, auth3dsRequiredDetails, sessionIdentifier,
-                null, null, recurringAuthToken, canRetry);
+                null, null, recurringAuthToken, canRetry, rejectedReason);
 
     }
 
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -696,6 +701,10 @@ private ChargeEntity updateChargeAndEmitEventPostAuthorisation(String chargeExte
         return chargeEntity;
     }
 
+    public void inactivateAgreement(String chargeExternalId) {
```
kbottla: Empty method?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -738,6 +749,22 @@ public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
             LOGGER.info("Stored confirmation details for charge - charge_external_id={}",
                     chargeExternalId);
 
+            if (canRetry != null && !canRetry) {
```
kbottla, line 748: Can this be extracted into a separate method?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -738,6 +749,22 @@ public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
...
+                charge.getAgreement()
+                        .map(agreementEntity -> agreementEntity.getPaymentInstrument()
+                                .map(paymentInstrumentEntity -> {
+                    paymentInstrumentEntity.setStatus(PaymentInstrumentStatus.INACTIVE);
+                    return paymentInstrumentEntity;
+                }));
+                charge.getAgreement()
+                        .map(agreementEntity -> {
+                            AgreementInactivated inactivatedEvent = AgreementInactivated
```
kbottla: Can the `AgreementInactivated` event take in reason as String so we can avoid converting reason into `MappedAuthorisationRejectedReason` and then take the reason from it?

src/test/java/uk/gov/pay/connector/charge/service/ChargeServicePostAuthorisationTest.java
```diff
@@ -163,8 +183,53 @@ void updateChargePostAuthorisationUpdatesChargeAndEmitsEvent_forAuthorisationMod
...
+        assertThat(chargeEntity.getWalletType(), is(nullValue()));
+        assertThat(chargeEntity.getCardDetails(), is(cardDetails));
+        assertThat(chargeEntity.getCanRetry(), is(false));
+
+        assertThat(paymentInstrument.getStatus(), is(PaymentInstrumentStatus.INACTIVE));
+
+        verify(mockStateTransitionService).offerPaymentStateTransition(EXTERNAL_ID, AUTHORISATION_READY, AUTHORISATION_SUCCESS, mockChargeEventEntity);
+        verify(mockEventService).emitAndRecordEvent(PaymentDetailsTakenFromPaymentInstrument.from(chargeEntity));
+
+        verify(mockLedgerService).postEvent(any(AgreementInactivated.class));
```
kbottla: Would be good to assert the actual values instead of `any`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: 👍

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: 👍

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: 👍

SandorArpa: 👍

src/test/java/uk/gov/pay/connector/charge/service/ChargeServicePostAuthorisationTest.java
SandorArpa: 👍

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -738,10 +745,30 @@ public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
...
+            if (canRetry != null && !canRetry) {
+                inactivateAgreement(charge, rejectedReason);
+            }
+
             return charge;
         }).orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeExternalId));
     }
 
+    private void inactivateAgreement(ChargeEntity charge, String rejectedReason) {
+        charge.getAgreement()
```
kbottla: Any reason setting payment instrument status and posting event is in 2 separate maps? Also can `ifPresent` be used since the value returned by the map is not used?

SandorArpa: 👍

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -738,10 +745,26 @@ public ChargeEntity updateChargePostAuthorisation(String chargeExternalId,
...
+                .map(agreementEntity -> agreementEntity.getPaymentInstrument()
+                        .map(paymentInstrumentEntity -> {
+                            paymentInstrumentEntity.setStatus(PaymentInstrumentStatus.INACTIVE);
+                            AgreementInactivated inactivatedEvent = AgreementInactivated
+                                    .from(agreementEntity, rejectedReason, Instant.now());
+                            ledgerService.postEvent(inactivatedEvent);
+            return agreementEntity;
+        }));
```
kbottla: Can we use `ifPresent` as we are not returning any or does it have to be a map()?

also, agreement specific func can be outside the payment instrument mapping function

SandorArpa: 👍

kbottla (approved):

PR #4525, opened 2023-05-16
PP-11019 Add gateway account internal ID to event schema

sfount:
Adds `gateway_account_internal_id` to the top level paymetns and refund events schema. For the time being, this will allow dependent services to process events based on the gateway account ID (used by the data model across Pay services).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationCancelled.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationCancelled.java
@@ -4,7 +4,7 @@ import java.time.Instant;
 
 // Semantically same as auth rejected
 public class AuthorisationCancelled extends PaymentEventWithoutDetails {
-    public AuthorisationCancelled(String serviceId, boolean live, Long gatewayAccountInternalId, String resourceExternalId, Instant timestamp) {
-        super(serviceId, live, gatewayAccountInternalId, resourceExternalId, timestamp);
+    public AuthorisationCancelled(String serviceId, boolean live, Long gatewayAccountId, String resourceExternalId, Instant timestamp) {
+        super(serviceId, live, gatewayAccountId, resourceExternalId, timestamp);
     }
 }
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationCancelled.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationCancelled.java
@@ -4,7 +4,7 @@ import java.time.Instant;
 
 // Semantically same as auth rejected
 public class AuthorisationCancelled extends PaymentEventWithoutDetails {
-    public AuthorisationCancelled(String serviceId, boolean live, String resourceExternalId, Instant timestamp) {
-        super(serviceId, live, resourceExternalId, timestamp);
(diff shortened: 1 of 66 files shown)


stephencdaly (dismissed):

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationCancelled.java
```diff
@@ -4,7 +4,7 @@
 
 // Semantically same as auth rejected
 public class AuthorisationCancelled extends PaymentEventWithoutDetails {
-    public AuthorisationCancelled(String serviceId, boolean live, String resourceExternalId, Instant timestamp) {
-        super(serviceId, live, resourceExternalId, timestamp);
+    public AuthorisationCancelled(String serviceId, boolean live, Long gatewayAccountInternalId, String resourceExternalId, Instant timestamp) {
```
kbottla: I think using `internal` is going to confuse anyone without context and is only internal to connector.
don't think it will convey what it is internal for if looking in ledger/webhooks for events.

and we will endup with another naming system.

stephencdaly: Yeh I agree with this - might be best to refer to it in the same way we refer to it in all other apps

sfount: I think I agree, I'm happy to change it -- the problem this was trying to address is that all of the other apps refer to things like `service_id` and `webhooks_id` etc. which of course refers to the external ID (as they wouldn't know about primary key identifiers). This is definitely different but it is true that other services like admin users currently call it `gateway_account_id` so we can line up with that

src/main/java/uk/gov/pay/connector/events/model/refund/RefundEvent.java
```diff
@@ -42,7 +42,7 @@ public String getServiceId() {
         return serviceId;
     }
 
-    public Long getGatewayAccountInternalId() {
-        return gatewayAccountInternalId;
+    public Long getgatewayAccountId() {
```
stephencdaly: Capitalisation is off

src/main/java/uk/gov/pay/connector/events/model/charge/PaymentEvent.java
```diff
@@ -42,7 +42,7 @@ public String getServiceId() {
         return serviceId;
     }
 
-    public Long getGatewayAccountInternalId() {
-        return gatewayAccountInternalId;
+    public Long getgatewayAccountId() {
```
stephencdaly: Capitalisation is off

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationCancelled.java
stephencdaly: Maybe something that could be addressed as we de-gateway-account-ify. The gateway_account_id will be used in less places and will be removed from here anyway when we do that

stephencdaly (dismissed):

stephencdaly (approved):

kbottla (approved):

PR #4529, opened 2023-05-19
PP-10981 Add can_retry inside state for get charge response

alexbishop1:
When getting a single charge (including for frontend), include a `can_retry` property inside the `state` object only if the following are both true:

- the charge has an authorisation mode of `AGREEMENT`
- the charge’s `canRetry` field is explicitly set to `true` or `false` (not left as `null`)

`ChargeService` is so large that adding many more lines to it runs the risk that it may cause it to achieve sentience and take over the world. Therefore, the code for creating an `ExternalTransactionState` (the Java representation of the JSON `state` object) has been moved into a factory, which contains the new logic for deciding when to include the `can_retry` property. This also makes unit testing a little easier.

`ChargesApiResourceIT` contains some new tests to check can_retry is serialised correctly when we want it.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -171,70 +171,70 @@
         "filename": "openapi/connector_spec.yaml",
         "hashed_secret": "7aaebcf1d5950305be73fbdda70fa0c55c6c8542",
         "is_verified": false,
-        "line_number": 3756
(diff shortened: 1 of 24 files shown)


SandorArpa (approved):
LGTM

PR #4535, opened 2023-05-23
PP-10983 Update util and wallet tests to JUnit5

kbottla:
## WHAT YOU DID
- Updated util and wallet related tests to JUnit5.
- use lenient() in `WalletAuthoriseServiceTest` as there are a lot of tests that need to be refactored to remove unnecessary stubbing


SandorArpa (approved):
LGTM

PR #4550, opened 2023-05-26
PP-11021 Some decline codes cannot be retried for recurring payments

alexbishop1:
In `MappedAuthorisationRejectedReason`, rename the boolean flag `canRetry` to `recurringPaymentCanBeRetried` to reflect that it only indicates whether a recurring payment can be retried (which is different to a payment where the user is present and able to do something).

In line with this, update the following reasons to indicate that they cannot be retried:

`AUTHENTICATION_REQUIRED`
At https://stripe.com/docs/declines/codes Stripe say: “The customer should try again and authenticate their card when prompted during the transaction.” Since the customer will not be around to do this, we cannot retry.

`EXPIRED_CARD`
Pages like https://www.tidalcommerce.com/learn/card-decline-codes say this means the expiration date is in the past. The user is not around to provide us with a new one. We may wish to review this when we have automatic card updating in a better place but even  then we’d probably know the new card details before the card expired.

`INVALID_CVV2`
There is no user around to provide the correct card security code. Seems unlikely this would ever happen in practice because how would the initial payment succeed?

`NO_SUCH_ISSUER`
Pages like https://paymentcloudinc.com/blog/credit-card-decline-codes/error-code-15/ say this means the card number has been entered incorrectly. Again, hard to see how this would ever happen in practice.

`STOLEN_CARD`
The ‘Refusal advice codes’ section of https://developer.worldpay.com/docs/access-worldpay/reference/useful-tables says we should not retry this. Although we do not use Access Worldpay and refusal advice codes, it’s fairly unambiguous what a stolen card means so we should follow this advice and not retry. This also gives us parity with `LOST_CARD`, which has similar advice.

There are also some reasons that may on the face of it look like they should not be retried but which we continue to allow to be retried:

`AUTHENTICATION_REQUESTED`
We map this from Worldpay decline code 65, which is defined on https://developer.worldpay.com/docs/access-worldpay/reference/useful-tables as “Exceeds withdrawal count limit OR Authentication requested”. Pages like https://paymentcloudinc.com/blog/credit-card-decline-codes/code-65/ say this can mean the user has “reached the maximum number of permitted transactions for a given period”. Therefore, it makes sense to try again.

`INVALID_ACCOUNT_NUMBER`
At Stripe https://stripe.com/docs/declines/codes Stripe say: “The customer needs to contact their card issuer to check that the card is working correctly.” If all the user’s payments are failing, it’s possible they may have sorted this before we retry.

`REFER_TO_ISSUER`
We map this from Worldpay decline code 1, which https://paymentcloudinc.com/blog/credit-card-decline-codes/code-01/ says can also mean insufficient funds.

No reasons that were marked as not being able to be retried were changed to being marked as being able to be retried as part of this.

Also reorder `MappedAuthorisationRejectedReason` so that the reasons that can be retried are above those that cannot. This is in a separate commit to aid reviewing.


SandorArpa (approved):
LGTM

PR #4554, opened 2023-05-26
PP-110601 extend cancel agreement integration test

SandorArpa:
## WHAT YOU DID
- check that the cancelled date on the agreement was set


james-peacock-ho (approved):
LGTM

PR #4573, opened 2023-06-13
PP-10833 Consider idempotency key records when expunging charges

AbdirizakIdris:
## Context:
-  Currently we are expunging charges and idempotency keys after 24 hours
   - It is possible that the charge is expunged from the connector but the Idempotency key stays. For a new request with the same idempotency key but expunged charge, Connector will throw ChargeNotFoundException.
   -  So we need to check if the idempotency key record exists for charge before expunging from the connector database
### WHAT
- Added ```idempotencyExistsByResourceExternalId```  method to find any idempotency records for ```resourceExternalId```

### DIFF
--- a/src/main/java/uk/gov/pay/connector/idempotency/dao/IdempotencyDao.java
+++ b/src/main/java/uk/gov/pay/connector/idempotency/dao/IdempotencyDao.java
@@ -34,4 +34,16 @@ public class IdempotencyDao extends JpaDao<IdempotencyEntity> {
                 .setParameter("expiryDate", expiryDate)
                 .executeUpdate();
     }
+
+    public boolean idempotencyExistsByResourceExternalId(String resourceExternalId) {
+        String query = "SELECT COUNT(ie) FROM IdempotencyEntity ie WHERE ie.resourceExternalId = :resourceExternalId";
+
+        long count = (long) entityManager
+                .get()
+                .createQuery(query)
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/idempotency/dao/IdempotencyDao.java
```diff
@@ -34,4 +34,14 @@ public int deleteIdempotencyKeysOlderThanSpecifiedDateTime(Instant expiryDate) {
                 .setParameter("expiryDate", expiryDate)
                 .executeUpdate();
     }
+
+    public Optional<IdempotencyEntity> findByResourceExternalId(String resourceExternalId) {
```
kbottla: Do we need the IdempotencyEntity? 

We can also return true/false if there are any idempotency records for the resource external ID similar to https://github.com/alphagov/pay-connector/blob/8cedaa3acde96012d1e0378b0203cbc86f4e6429/src/main/java/uk/gov/pay/connector/gatewayaccount/dao/StripeAccountSetupDao.java#L30

When expunging the charge (recurring payment), we can simply check if this method returns true/false and delete the charge if any Idempotency records exist.

AbdirizakIdris: Good suggestion - updated it now thanks

kbottla (dismissed):

kbottla (approved):

PR #4577, opened 2023-06-15
BAU: Add a check to make sure pacts were generated

jfharden:
## WHAT YOU DID
Add a check after the unit tests which checks to see if the pact files were generated.

If the pact files aren't generated then pacts will not be published and the build will fail after merge and passed e2e tests which it's trying to deploy into test.

## How to test

I pushed the check which passed
In a second commit i updated dropwizard to 2.1.6 which we know breaks the pact file generation, you can see the unit test failed
I reverted the dropwizard change


nlsteers (approved):
💀

PR #4582, opened 2023-06-19
PP-10598 add new agreement state

SandorArpa:
## WHAT YOU DID
- add a new agreement state that also creates a gateway account
- add an active agreement
- add default values to variables for selfservice use

with @james-peacock-gds


james-peacock-ho (approved):

PR #4583, opened 2023-06-19
PP-10598 fix pact state

SandorArpa:
## WHAT YOU DID
- a cancel agreement is going to trigger a post request to Ledger. Stub out this request.

with @james-peacock-gds


james-peacock-ho (approved):

PR #4584, opened 2023-06-19
PP-10598 add new state for cancelled payment instrument

SandorArpa:
## WHAT YOU DID
- add a new state for agreement pact tests where the payment instrument status is cancelled

with @james-peacock-gds


james-peacock-ho (approved):

PR #4595, opened 2023-06-27
PP-10600 fix pact test

SandorArpa:
## WHAT YOU DID
- add default card details and default gateway account to pact test


james-peacock-ho (approved):

PR #4597, opened 2023-06-28
PP-10599 add a new State for pact tests

SandorArpa:
## WHAT YOU DID
- add a new State for a newly created agreement without payment instrument

with @james-peacock-gds


james-peacock-ho (approved):

PR #4600, opened 2023-06-30
PP-11157 Move known credentials keys from config to code

alexbishop1:
Move the known credentials keys for each payment gateway from the `config.yaml` file into the code since it is highly unlikely we will ever change these without also changing the code.

This allows us to support more than one set of credentials keys for a particular gateway, allowing us to migrate some types of Worldpay credentials from `merchant_id` to the more correct `merchant_code`.


stephencdaly (approved):
LGTM!

PR #4602, opened 2023-07-03
PP-11244 Stop using ChargeEntity for building JSON responses

stephencdaly:
The only place we still use ChargeEntity as an API response model is for the GET `/v1/frontend/tokens/{chargeTokenId}` endpoint.

Instead use the `FrontendChargeResponse` model that we use for other charge responses consumed by the card connector app.

https://github.com/alphagov/pay-frontend/pull/3551 made changes to frontend to support this change.

The diff for openapi/connector_spec.yaml is very difficult to visually parse, there's lots of things about `_FrontendView`s which exist due to the JsonViews used to serialise the correct response from a GatewayAccountEntity. This will all be sorted out in a later commit that will stop using GatewayAccountEntity to serialise JSON responses. You might just need to trust that is is all okay.

You can view the updated API spec in Swagger editor here https://editor.swagger.io/?url=https://raw.githubusercontent.com/alphagov/pay-connector/24eacd2335bda7025970adc2ce58c2d249fa14c5/openapi/connector_spec.yaml

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -153,88 +153,81 @@
         "line_number": 787
       },
       {
-        "type": "Hex High Entropy String",
-        "filename": "openapi/connector_spec.yaml",
-        "hashed_secret": "48ca4a65970b54002939b02e8d8c1b0167de7e4b",
-        "is_verified": false,
-        "line_number": 3028
-      },
-      {
         "type": "Base64 High Entropy String",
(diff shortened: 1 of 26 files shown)


oswaldquek (commented):
LGTM! Nice one.

oswaldquek (approved):

PR #4604, opened 2023-07-04
PP-11167 Reject payment links payments with card number in reference

kbottla:
## WHAT YOU DID
- Rejects payment link payments with a card number in the reference. Currently behind a feature flag so this can be enabled when all microservices are ready to handle new error identifier.


stephencdaly (approved):
LGTM!

PR #4616, opened 2023-07-07
PP-11167 Reject agent initiated moto payments for card number in reference

kbottla:
## WHAT YOU DID

- Rejects agent initiated moto payments too for the card number in reference
- And logs the first and last digits of the card number which will be useful when investigating


SandorArpa (approved):
LGTM

PR #4617, opened 2023-07-07
PP-11244 Remove OpenApi tag

stephencdaly:
Remove the "Gateway accounts - frontend" OpenApi tag, and instead tag all gateway account endpoints with "Gateway accounts".

There is no meaningful reason why some endpoints are `/v1/frontend/` and others are `/v1/api/` as all of these endpoints are consumed by selfservice and toolbox. Using the same tag will group them all together when viewed in an API browser.


SandorArpa (approved):
LGTM

PR #4620, opened 2023-07-11
PP-11152 Convert credentials map into POJO

stephencdaly:
Add a GatewayCredentials interface, which the types representing the credentials structure for each payment gateway inherit. Add implementations for Worldpay, Stripe and ePDQ. Also add an implementation for Sandbox to avoid returning null/Optional when getting the credentials for a Sandbox account, even though there are no credentials that can be entered for Sandbox.

Add a getter to GatewayAccountCredentialsEntity to convert the credentials JSON map to a POJO that represents the credentials structure for the payment gateway the credentials are for.

Add a setter to GatewayAccountCredentialsEntity to allow serializing implementations of GatewayCredentials to a JSON map to insert into the database.

Start using the GatewayCredentials representation rather than the credentials map when updating Worldpay credentials.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
@@ -2,7 +2,6 @@ package uk.gov.pay.connector.gatewayaccount.model;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
-import com.google.common.collect.ImmutableMap;
 
 import java.util.Map;
 
@@ -14,8 +13,8 @@ public class StripeCredentials implements GatewayCredentials {
     @JsonProperty(STRIPE_ACCOUNT_ID_KEY)
     private String stripeAccountId;
 
-    public StripeCredentials(@JsonProperty(STRIPE_ACCOUNT_ID_KEY) String stripeAccountId) {
-        this.stripeAccountId = stripeAccountId;
+    public StripeCredentials() {
+        // for Jackson
     }
 
     public String getStripeAccountId() {
@@ -23,7 +22,7 @@ public class StripeCredentials implements GatewayCredentials {
     }
 
     public Map<String, String> toMap() {
-        return ImmutableMap.of(STRIPE_ACCOUNT_ID_KEY, stripeAccountId);
+        return Map.of(STRIPE_ACCOUNT_ID_KEY, stripeAccountId);
     }
     
 }
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
@@ -1,20 +1,29 @@
 package uk.gov.pay.connector.gatewayaccount.model;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableMap;
 
 import java.util.Map;
 
-public class StripeCredentials {
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class StripeCredentials implements GatewayCredentials {
 
     public static final String STRIPE_ACCOUNT_ID_KEY = "stripe_account_id";
-    private String accountId;
+    
+    @JsonProperty(STRIPE_ACCOUNT_ID_KEY)
+    private String stripeAccountId;
 
-    public StripeCredentials(@JsonProperty(STRIPE_ACCOUNT_ID_KEY) String accountId) {
-        this.accountId = accountId;
+    public StripeCredentials(@JsonProperty(STRIPE_ACCOUNT_ID_KEY) String stripeAccountId) {
+        this.stripeAccountId = stripeAccountId;
+    }
+
+    public String getStripeAccountId() {
+        return stripeAccountId;
     }
 
     public Map<String, String> toMap() {
-        return ImmutableMap.of(STRIPE_ACCOUNT_ID_KEY, accountId);
+        return ImmutableMap.of(STRIPE_ACCOUNT_ID_KEY, stripeAccountId);
     }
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/credentials/SandboxCredentials.java
```diff
@@ -0,0 +1,10 @@
+package uk.gov.pay.connector.gatewayaccount.model.credentials;
+
+public class SandboxCredentials implements GatewayCredentials {
```
stephencdaly: The other option if we didn't want a Sandbox implementation would be to return an Optional from the getCredentialsAsPojo method on the GatewayAccountCredentialsEntity. But it would be nice to avoid having to handle optionals all over the place.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
```diff
@@ -1,20 +1,29 @@
...
 
     public static final String STRIPE_ACCOUNT_ID_KEY = "stripe_account_id";
-    private String accountId;
+    
+    @JsonProperty(STRIPE_ACCOUNT_ID_KEY)
+    private String stripeAccountId;
 
-    public StripeCredentials(@JsonProperty(STRIPE_ACCOUNT_ID_KEY) String accountId) {
-        this.accountId = accountId;
+    public StripeCredentials(@JsonProperty(STRIPE_ACCOUNT_ID_KEY) String stripeAccountId) {
```
alexbishop1: Is this annotation necessary if the field is annotated?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
```diff
@@ -1,20 +1,29 @@
...
+        this.stripeAccountId = stripeAccountId;
+    }
+
+    public String getStripeAccountId() {
+        return stripeAccountId;
     }
 
     public Map<String, String> toMap() {
-        return ImmutableMap.of(STRIPE_ACCOUNT_ID_KEY, accountId);
+        return ImmutableMap.of(STRIPE_ACCOUNT_ID_KEY, stripeAccountId);
```
alexbishop1: Since we’re changing this line anyway, we should make this `Map.of(…)`.

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialsEntity.java
```diff
@@ -29,6 +40,15 @@
...
 public class GatewayAccountCredentialsEntity extends AbstractVersionedEntity {
 
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+    
+    private static final Map<PaymentGatewayName, Class<? extends GatewayCredentials>> GATEWAY_CREDENTIALS_TYPE_MAP = Map.ofEntries(
+            entry(PaymentGatewayName.WORLDPAY, WorldpayCredentials.class),
+            entry(PaymentGatewayName.STRIPE, StripeCredentials.class),
+            entry(PaymentGatewayName.EPDQ, EpdqCredentials.class),
+            entry(PaymentGatewayName.SANDBOX, SandboxCredentials.class)
+    );
```
alexbishop1: There’s so few entries that there’s an overloaded `Map.of(…)` method that could be used.

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialsEntity.java
```diff
@@ -90,6 +110,13 @@ public Map<String, Object> getCredentials() {
         return credentials;
     }
 
+    // This will replace `getCredentials()` when all usages have been updated
+    public GatewayCredentials getCredentialsObject() {
+        return Optional.ofNullable(GATEWAY_CREDENTIALS_TYPE_MAP.get(PaymentGatewayName.valueFrom(paymentProvider)))
+                .map(type -> objectMapper.convertValue(this.getCredentials(), type))
+                .orElseThrow(() -> new IllegalArgumentException("Unsupported payment provider: " + paymentProvider));
```
alexbishop1: Would a switch statement maybe be a little nicer?

```java
switch (PaymentGatewayName.valueFrom(paymentProvider)) {
    case WORLDPAY:
         return objectMapper.convertValue(this.getCredentials(), WorldpayCredentials.class);
    case STRIPE:
         return objectMapper.convertValue(this.getCredentials(), StripeCredentials.class);
    case EPDQ:
         return objectMapper.convertValue(this.getCredentials(), EpdqCredentials.class);
    case SANDBOX:
         return objectMapper.convertValue(this.getCredentials(), SandboxCredentials.class);
    default:
       throw new IllegalArgumentException("Unsupported payment provider: " + paymentProvider;
}
```

I can’t decide so I’ll leave it up to you.

Aside — in an alternative reality where we’re using a Java version that supports switch expressions, we could do this:

```java
return objectMapper.convertValue(this.getCredentials(), switch (PaymentGatewayName.valueFrom(paymentProvider)) {
    case WORLDPAY -> WorldpayCredentials.class;
    case STRIPE -> StripeCredentials.class;
    case EPDQ -> EpdqCredentials.class;
    case SANDBOX -> SandboxCredentials.class;
});
```

alexbishop1 (dismissed):

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialsEntity.java
stephencdaly: Yeh maybe, I did consider this. It could make it a bit easier to follow

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeCredentials.java
stephencdaly: Looks like its a way to avoid needing an empty constructor

alexbishop1 (approved):

PR #4622, opened 2023-07-13
PP-10506 Inactivate agreements only when recurring payments fail

kbottla:
## WHAT YOU DID
- Inactivate agreements only when charges with AUTHORISATION_MODE=agreement fail with a non-retriable PSP error code.
- Removed comment `TODO implement PP-10833, query Ledger if Charge has been expunged` as we no longer need it. Charges are expunged only after the idempotency key record is expunged.


SandorArpa (approved):
LGTM

PR #4629, opened 2023-07-14
PP-11152 Add isForRecurringPayment to GatewayRequest

stephencdaly:
We are adding in `recurring_customer_initiated` nested credentials to the credentials JSON for Worldpay accounts, to use instead of the credentials at the root level for payments to set up a recurring payment agreement.

In order to determine whether to get the one-off or recurring customer initiated credentials, we need to know whether a payment is recurring or not. Add `isForRecurringPayment` method to GatewayRequest interface to tell us this.

We determine whether a payment is recurring or not by whether it has an agreement ID set. This means that for query and refund requests, where we get the payment from ledger, we need to consume the `agreement_id` field on returned transactions. Add a Pact test for getting a recurring payment from ledger to test the contract for this.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/ChargeQueryGatewayRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/ChargeQueryGatewayRequest.java
@@ -15,14 +15,16 @@ public class ChargeQueryGatewayRequest implements GatewayRequest {
     private final String chargeExternalId;
     private final String transactionId;
     private final AuthorisationMode authorisationMode;
+    private boolean isForRecurringPayment;
 
     public ChargeQueryGatewayRequest(GatewayAccountEntity gatewayAccountEntity, GatewayAccountCredentialsEntity gatewayAccountCredentialsEntity,
-                                     String chargeExternalId, String transactionId, AuthorisationMode authorisationMode) {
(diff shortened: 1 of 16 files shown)


src/main/java/uk/gov/pay/connector/gateway/ChargeQueryGatewayRequest.java
```diff
@@ -51,14 +53,20 @@ public AuthorisationMode getAuthorisationMode() {
     public Map<String, Object> getGatewayCredentials() {
         return gatewayAccountCredentialsEntity.getCredentials();
     }
-    
+
+    @Override
+    public boolean isRecurring() {
+        return isRecurring;
+    }
```
alexbishop1: Perhaps I’m being a little too pedantic here but could this be `isForRecurringPayment()` or similar? I worry that when on, say, `CaptureGatewayRequest` it could be taken to be saying the _request_ is recurring rather than the payment it relates to.

alexbishop1 (commented):
Changes look fine but I’ve taken the opportunity to bikeshed the name.

src/main/java/uk/gov/pay/connector/gateway/ChargeQueryGatewayRequest.java
stephencdaly: I agree with your reasoning, have renamed

alexbishop1 (approved):

PR #4630, opened 2023-07-17
PP-11152 Use credentials POJO for building gateway requests

stephencdaly:
Use the GatewayCredentials representation of the credentials when retrieving the credentials for use when making requests to the payment gateways.

As part of this, make the change to get the credentials for a customer initiated recurring Worldpay payment from the
`recurring_customer_initiated` nested credentials, rather than the root credentials. Credentials for accounts used for Worldpay recurring payments have already been updated on all environments to include `recurring_customer_initiated` credentials.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
@@ -11,6 +11,7 @@ import uk.gov.pay.connector.gatewayaccountcredentials.exception.MissingCredentia
 import uk.gov.pay.connector.gatewayaccountcredentials.exception.NoCredentialsInUsableStateException;
 import uk.gov.service.payments.commons.model.AuthorisationMode;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.Map;
 
@@ -45,14 +46,8 @@ public class AuthUtil {
     public static String getWorldpayMerchantCode(GatewayCredentials credentials, AuthorisationMode authorisationMode, boolean isRecurring) {
         WorldpayCredentials worldpayCredentials = castGatewayCredentialsToWorldpayCredentials(credentials);
         if (isRecurring) {
-            if (authorisationMode == AuthorisationMode.AGREEMENT) {
-                return worldpayCredentials.getRecurringMerchantInitiatedCredentials()
-                        .map(WorldpayMerchantCodeCredentials::getMerchantCode)
-                        .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
-            }
-            return worldpayCredentials.getRecurringCustomerInitiatedCredentials()
-                    .map(WorldpayMerchantCodeCredentials::getMerchantCode)
-                    .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
+            WorldpayMerchantCodeCredentials credentialsForAuthType = getWorldpayRecurringCredentialsForAuthType(worldpayCredentials, authorisationMode);
+            return credentialsForAuthType.getMerchantCode();
         }
 
         return worldpayCredentials.getOneOffCustomerInitiatedCredentials()
@@ -65,14 +60,8 @@ public class AuthUtil {
         WorldpayCredentials worldpayCredentials = castGatewayCredentialsToWorldpayCredentials(credentials);
 
         if (isRecurring) {
-            if (authorisationMode == AuthorisationMode.AGREEMENT) {
-                return worldpayCredentials.getRecurringMerchantInitiatedCredentials()
-                        .map(AuthUtil::getWorldpayAuthHeader)
-                        .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
-            }
-            return worldpayCredentials.getRecurringCustomerInitiatedCredentials()
-                    .map(AuthUtil::getWorldpayAuthHeader)
-                    .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
+            WorldpayMerchantCodeCredentials credentialsForAuthType = getWorldpayRecurringCredentialsForAuthType(worldpayCredentials, authorisationMode);
+            return getWorldpayAuthHeader(credentialsForAuthType);
         }
 
         return worldpayCredentials.getOneOffCustomerInitiatedCredentials()
@@ -83,12 +72,21 @@ public class AuthUtil {
                     return getAuthHeader(username, password);
                 });
     }
+    
+    private static WorldpayMerchantCodeCredentials getWorldpayRecurringCredentialsForAuthType(WorldpayCredentials worldpayCredentials, AuthorisationMode authorisationMode) {
+        if (authorisationMode == AuthorisationMode.AGREEMENT) {
+            return worldpayCredentials.getRecurringMerchantInitiatedCredentials()
+                    .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
+        }
+        return worldpayCredentials.getRecurringCustomerInitiatedCredentials()
+                .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
+    }
 
     private static Map<String, String> getWorldpayAuthHeader(WorldpayMerchantCodeCredentials creds) {
         return getAuthHeader(creds.getUsername(), creds.getPassword());
     }
 
-    public static Map<String, String> getGatewayAccountCredentialsForManagingTokensAsAuthHeader(GatewayCredentials credentials) {
+    public static Map<String, String> getWorldpayAuthHeaderForManagingRecurringAuthTokens(GatewayCredentials credentials) {
         WorldpayCredentials worldpayCredentials = castGatewayCredentialsToWorldpayCredentials(credentials);
         WorldpayMerchantCodeCredentials recurringCustomerInitiatedCredentials = worldpayCredentials.getRecurringCustomerInitiatedCredentials()
                 .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
@@ -115,7 +113,7 @@ public class AuthUtil {
     }
 
     private static Map<String, String> getAuthHeader(String username, String password) {
-        String value = "Basic " + Base64.getEncoder().encodeToString(new String(username + ":" + password).getBytes());
+        String value = "Basic " + Base64.getEncoder().encodeToString((username + ":" + password).getBytes(StandardCharsets.UTF_8));
         return Map.of(AUTHORIZATION, value);
     }
 
--- a/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
@@ -1,10 +1,14 @@
 package uk.gov.pay.connector.gateway.util;
 
-import com.google.common.collect.ImmutableMap;
 import uk.gov.pay.connector.app.StripeAuthTokens;
 import uk.gov.pay.connector.app.StripeGatewayConfig;
+import uk.gov.pay.connector.gatewayaccount.model.EpdqCredentials;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayCredentials;
+import uk.gov.pay.connector.gatewayaccount.model.WorldpayCredentials;
+import uk.gov.pay.connector.gatewayaccount.model.WorldpayMerchantCodeCredentials;
 import uk.gov.pay.connector.gatewayaccount.model.WorldpayValidatableCredentials;
 import uk.gov.pay.connector.gatewayaccountcredentials.exception.MissingCredentialsForRecurringPaymentException;
+import uk.gov.pay.connector.gatewayaccountcredentials.exception.NoCredentialsInUsableStateException;
 import uk.gov.service.payments.commons.model.AuthorisationMode;
 
 import java.util.Base64;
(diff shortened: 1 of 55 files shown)


src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -41,53 +36,87 @@ public static Map<String, String> getStripeAuthHeader(StripeGatewayConfig stripe
...
+            return worldpayCredentials.getRecurringCustomerInitiatedCredentials()
+                    .map(WorldpayMerchantCodeCredentials::getMerchantCode)
+                    .orElseThrow(MissingCredentialsForRecurringPaymentException::new);
+        }
+
+        return worldpayCredentials.getOneOffCustomerInitiatedCredentials()
+                .map(WorldpayMerchantCodeCredentials::getMerchantCode)
+                .orElseGet(() -> worldpayCredentials.getLegacyOneOffCustomerInitiatedMerchantCode()
+                        .orElseThrow(NoCredentialsInUsableStateException::new));
+    }
```
alexbishop1: Maybe having a separate method rather than an `isRecurring` flag might make this a bit simpler?

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -41,53 +36,87 @@ public static Map<String, String> getStripeAuthHeader(StripeGatewayConfig stripe
...
         }
-        return gatewayCredentials.get(CREDENTIALS_MERCHANT_ID).toString();
+
+        return worldpayCredentials.getOneOffCustomerInitiatedCredentials()
+                .map(AuthUtil::getWorldpayAuthHeader)
+                .orElseGet(() -> {
+                    String username = worldpayCredentials.getLegacyOneOffCustomerInitiatedUsername().orElseThrow(NoCredentialsInUsableStateException::new);
+                    String password = worldpayCredentials.getLegacyOneOffCustomerInitiatedPassword().orElseThrow(NoCredentialsInUsableStateException::new);
+                    return getAuthHeader(username, password);
+                });
```
alexbishop1: Feels like this should be able to reuse some of the logic in the method above.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -41,53 +36,87 @@ public static Map<String, String> getStripeAuthHeader(StripeGatewayConfig stripe
...
-    public static String getWorldpayMerchantCodeForManagingTokens(Map<String, Object> gatewayCredentials) {
-        return gatewayCredentials.get(CREDENTIALS_MERCHANT_ID).toString();
+    private static Map<String, String> getWorldpayAuthHeader(WorldpayMerchantCodeCredentials creds) {
+        return getAuthHeader(creds.getUsername(), creds.getPassword());
     }
 
-    public static Map<String, String> getGatewayAccountCredentialsAsAuthHeader(Map<String, Object> gatewayCredentials) {
-        String value = encode(gatewayCredentials.get(CREDENTIALS_USERNAME).toString(), gatewayCredentials.get(CREDENTIALS_PASSWORD).toString());
-        return ImmutableMap.of(AUTHORIZATION, value);
+    public static Map<String, String> getGatewayAccountCredentialsForManagingTokensAsAuthHeader(GatewayCredentials credentials) {
```
alexbishop1: It’s not too clear that `Tokens` means tokens for recurring payments here.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -41,53 +36,87 @@ public static Map<String, String> getStripeAuthHeader(StripeGatewayConfig stripe
...
+    public static Map<String, String> getEpdqAuthHeader(GatewayCredentials credentials) {
+        if (!(credentials instanceof EpdqCredentials)) {
+            throw new IllegalArgumentException("Expected provided GatewayCredentials to be of type EpdqCredentials");
+        }
+        var epdqCredentials = (EpdqCredentials) credentials;
+        return getAuthHeader(epdqCredentials.getUsername(), epdqCredentials.getPassword());
+    }
+
+    private static Map<String, String> getAuthHeader(String username, String password) {
+        String value = "Basic " + Base64.getEncoder().encodeToString(new String(username + ":" + password).getBytes());
```
alexbishop1: `new String(username + ":" + password).getBytes()` can just be `(username + ":" + password).getBytes()` but it should really be `(username + ":" + password).getBytes(StandardCharsets.UTF_8)` to make it less platform-dependent.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -219,23 +219,11 @@ public Optional<GatewayAccountCredentialsEntity> getCurrentOrActiveGatewayAccoun
...
-                .orElseThrow(() -> new WebApplicationException(serviceErrorResponse(
-                        format("Credentials not found for gateway account [%s] and payment_provider [%s] ",
-                                getId(), paymentProvider))));
-    }
     
     public GatewayAccountCredentialsEntity getGatewayAccountCredentialsEntity(String paymentProvider) {
         return gatewayAccountCredentials.stream()
                 .filter(entity -> entity.getPaymentProvider().equals(paymentProvider))
-                .findFirst()
+                .max(comparing(GatewayAccountCredentialsEntity::getActiveStartDate))
```
alexbishop1, line 226: Why was this change made?

src/test/java/uk/gov/pay/connector/gateway/util/AuthUtilTest.java
```diff
@@ -35,65 +37,126 @@ class AuthUtilTest {
             ));
 
     @Test
-    void shouldThrowException_whenAuthModeAgreement_andNoCredentials() {
+    void getWorldpayAuthHeader_shouldThrowExceptionWhenNoCredentials_forRecurringCustomerInitiated() {
```
alexbishop1, line 41: Can we try to be consistent with the naming convention for tests in this class?

src/test/java/uk/gov/pay/connector/gateway/worldpay/wallets/WorldpayWalletAuthorisationHandlerTest.java
```diff
@@ -117,7 +121,8 @@ void shouldSendApplePayRequestWhenApplePayDetailsArePresent() throws Exception {
             assertXMLEqual(TestTemplateResourceLoader.load(WORLDPAY_VALID_AUTHORISE_WORLDPAY_APPLE_PAY_REQUEST),
                     gatewayOrderArgumentCaptor.getValue().getPayload());
             assertThat(headers.getValue().size(), is(1));
-            assertThat(headers.getValue(), is(AuthUtil.getGatewayAccountCredentialsAsAuthHeader(gatewayAccountCredentials)));
+            String expectedHeader = "Basic " + Base64.getEncoder().encodeToString(new String( username + ":" + password).getBytes());
```
alexbishop1: Please always specify a charset when using `String.getBytes()`.

src/test/java/uk/gov/pay/connector/it/resources/CreateGatewayAccountResourceIT.java
```diff
@@ -83,11 +86,12 @@ public void createStripeGatewayAccountWithoutCredentials() {
 
     @Test
     public void createStripeGatewayAccountWithCredentials() {
+        String StripeAccountId = "abc";
```
alexbishop1: Capital letter at the start of a local variable name?

alexbishop1 (changes requested):
Overall this change makes the code much nicer. Just a few small qualms.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
stephencdaly: If we do that, we would be moving the logic to a different place, and repeating it in multiple places.
The logic could be made simpler if we were to remove the legacy credentials at some point.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
stephencdaly: I did briefly consider that, however I feel that by trying to re-use the logic we would end up having to apply more mental gymnastics to understanding it.
Re-using the logic would be much easier if we were able to get rid of the legacy credentials - as then we're just dealing with WorldpayMerchantCodeCredentials.

Possibly we could re-use just the recurring credentials part..

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
stephencdaly: This was already the name of the method, I could rename it.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
stephencdaly: Could improve that, but it was existing behaviour.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
stephencdaly: Ah yes I forgot about this change. This was because I removed and replaced the usages of the above `public Map<String, Object> getCredentials(String paymentProvider) {` method, which had this logic.

It feels like better behaviour to get the most recently active credential for the provider (if there is one) rather than just the first. I could potentially make the effort to separate this out to a different PR if you think that is worthwhile.

src/test/java/uk/gov/pay/connector/gateway/util/AuthUtilTest.java
stephencdaly: As long as it's clear what an individual test is testing, does it matter that naming isn't consistent? I don't want to end up renaming all the tests to make them consistent...

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
alexbishop1: Perhaps I was not clear. I meant we could have a `getWorldpayMerchantCode(GatewayCredentials credentials)` method containing everything that’s currently outside the `if` block and a `getWorldpayMerchantCodeForRecurring(GatewayCredentials credentials, AuthorisationMode authorisationMode)` method containing everything that’s currently inside the `if` block. It would just make the body of each method a little shorter. Each call site would call the appropriate version of the method, rather than passing the `isRecurring` flag.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
alexbishop1: It’s fine to stay in this PR now it’s here.

src/test/java/uk/gov/pay/connector/gateway/util/AuthUtilTest.java
alexbishop1: I suppose so.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
```diff
@@ -41,53 +37,84 @@ public static Map<String, String> getStripeAuthHeader(StripeGatewayConfig stripe
...
+            WorldpayMerchantCodeCredentials credentialsForAuthType = getWorldpayRecurringCredentialsForAuthType(worldpayCredentials, authorisationMode);
+            return getWorldpayAuthHeader(credentialsForAuthType);
         }
-        return gatewayCredentials.get(CREDENTIALS_MERCHANT_ID).toString();
+
+        return worldpayCredentials.getOneOffCustomerInitiatedCredentials()
+                .map(AuthUtil::getWorldpayAuthHeader)
+                .orElseGet(() -> {
+                    String username = worldpayCredentials.getLegacyOneOffCustomerInitiatedUsername().orElseThrow(NoCredentialsInUsableStateException::new);
+                    String password = worldpayCredentials.getLegacyOneOffCustomerInitiatedPassword().orElseThrow(NoCredentialsInUsableStateException::new);
```
alexbishop1, line 71: Probably not for now but I’m almost tempted to suggest we add a `getLegacyCredentials()` method that returns a `WorldpayMerchantCodeCredentials` object created on the fly, so we can harmonise how we treat legacy and new credentials.

alexbishop1 (approved):
As my childhood dentist used to say: “You’ll do!”

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
stephencdaly: Yeh maybe that would help simplify some things, if it were done in a way that doesn't make it less clear.

src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java
stephencdaly: That's what I understood by your first comment. But the call sites are often for code that could be either for a recurring payment or one-off payment. There's one or 2 places where it only applies to one or the other. But for authorise (customer present), authorise 3ds, capture, query, refund it's all shared code between one-off and recurring.

PR #4632, opened 2023-07-18
PP-10833 consider idempotency key records when expunging charges

AbdirizakIdris:
## WHAT 
- Currently we are expunging charges and idempotency keys after 24 hours * It is possible that the charge is expunged from the connector but the Idempotency key stays. For a new request with the same idempotency key but expunged charge, Connector will throw ChargeNotFoundException. So we need to check if the idempotency key record exists for charge before expunging from the connector database

  - extended conditions to check if the idempotency key record exists for charge, If so, don’t expunge the charge

### DIFF
--- a/src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
+++ b/src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
@@ -11,6 +11,7 @@ import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
+import uk.gov.pay.connector.idempotency.dao.IdempotencyDao;
 import uk.gov.pay.connector.tasks.service.ParityCheckService;
 import uk.gov.service.payments.commons.model.AuthorisationMode;
 
@@ -43,6 +44,7 @@ public class ChargeExpungeService {
     private final ExpungeConfig expungeConfig;
     private final ParityCheckService parityCheckService;
     private final ChargeService chargeService;
+    private final IdempotencyDao idempotencyDao;
     private final List<PaymentGatewayName> expungeExemptedGateways = List.of(
             PaymentGatewayName.EPDQ,
             PaymentGatewayName.WORLDPAY,
@@ -62,11 +64,12 @@ public class ChargeExpungeService {
     @Inject
     public ChargeExpungeService(ChargeDao chargeDao, ConnectorConfiguration connectorConfiguration,
                                 ParityCheckService parityCheckService,
-                                ChargeService chargeService) {
+                                ChargeService chargeService, IdempotencyDao idempotencyDao) {
         this.chargeDao = chargeDao;
         expungeConfig = connectorConfiguration.getExpungeConfig();
         this.parityCheckService = parityCheckService;
         this.chargeService = chargeService;
+        this.idempotencyDao = idempotencyDao;
     }
 
     private boolean inTerminalState(ChargeEntity chargeEntity) {
@@ -126,6 +129,10 @@ public class ChargeExpungeService {
             logger.info("Charge not expunged because it is a Stripe payment that requires fees to be collected, but " +
                             "fees have not yet been processed.",
                     kv(PAYMENT_EXTERNAL_ID, chargeEntity.getExternalId()));
+        } else if (chargeEntity.getAuthorisationMode() == AuthorisationMode.AGREEMENT &&
+                idempotencyDao.idempotencyExistsByResourceExternalId(chargeEntity.getExternalId())) {
+                logger.info("Charge not expunged from connector because Idempotency record exists for charge",
+                        kv(PAYMENT_EXTERNAL_ID, chargeEntity.getExternalId()));
         } else if (parityCheckService.parityCheckChargeForExpunger(chargeEntity)) {
             expungeCharge(chargeEntity);
             logger.info("Charge expunged from connector {}", kv(PAYMENT_EXTERNAL_ID, chargeEntity.getExternalId()));
--- a/src/test/java/uk/gov/pay/connector/expunge/service/ChargeExpungeServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/expunge/service/ChargeExpungeServiceTest.java
@@ -18,6 +18,7 @@ import uk.gov.pay.connector.charge.service.ChargeService;
 import uk.gov.pay.connector.fee.model.Fee;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType;
+import uk.gov.pay.connector.idempotency.dao.IdempotencyDao;
 import uk.gov.pay.connector.tasks.service.ParityCheckService;
 
 import java.time.Duration;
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
```diff
@@ -98,13 +102,36 @@ public void expunge(Integer noOfChargesToExpunge) {
...
-                    } catch (OptimisticLockException error) {
-                        logger.info("Expunging process conflicted with an already running process, exit");
-                        MDC.remove(MDC_REQUEST_ID_KEY);
-                        MDC.remove(PAYMENT_EXTERNAL_ID);
-                        throw error;
+
+                    IdempotencyEntity idempotencyEntity = new IdempotencyEntity();
+                    String resourceId = idempotencyEntity.getResourceExternalId();
+
+                    if (chargeEntity.getAuthorisationMode() == AuthorisationMode.AGREEMENT) {
```
kbottla: Would probably be better if the condition is in method `parityCheckAndExpungeIfMet`.

In `parityCheckAndExpungeIfMet` method, check for agreement mode and idempotency record can be last one condition

```
if (!inTerminalState(chargeEntity)) {
...
} else if (chargeEntity.getPaymentGatewayName() == STRIPE && isStripePaymentMissingFees(chargeEntity)) {
....
} else if (......idempotency-record-exists){
   ///log..
}
 else if (parityCheckService.parityCheckChargeForExpunger(chargeEntity)) {
            expungeCharge(chargeEntity);
```

src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
```diff
@@ -98,13 +102,36 @@ public void expunge(Integer noOfChargesToExpunge) {
...
+                            logger.info("Idempotency exists for charge for AGREEMENT mode, skipping expunge");
+                        } else {
+                            try {
+                                parityCheckAndExpungeIfMet(chargeEntity);
+                            } catch (OptimisticLockException error) {
+                                logger.info("Expunging process conflicted with an already running process, exit");
+                                MDC.remove(MDC_REQUEST_ID_KEY);
+                                MDC.remove(PAYMENT_EXTERNAL_ID);
+                                throw error;
+                            }
```
kbottla: See the above comment. 

Ideally, if the condition is added to this method it can be as below before the try/catch block. We can also avoid duplicating code across if/else blocks

```
if ( charge.authorisation_mode == AGREEMENT  and idempotency_record_exists) {
   // log
  continue;
}
```

src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
kbottla: Can test(s) be added for this change?

kbottla (changes requested):

src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
AbdirizakIdris: updated into this method now

src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
```diff
@@ -126,6 +129,11 @@ private void parityCheckAndExpungeIfMet(ChargeEntity chargeEntity) {
             logger.info("Charge not expunged because it is a Stripe payment that requires fees to be collected, but " +
                             "fees have not yet been processed.",
                     kv(PAYMENT_EXTERNAL_ID, chargeEntity.getExternalId()));
+        } else if (chargeEntity.getAuthorisationMode() == AuthorisationMode.AGREEMENT &&
+                idempotencyDao.idempotencyExistsByResourceExternalId(chargeEntity.getExternalId())) {
+                chargeService.updateChargeParityStatus(chargeEntity.getExternalId(), SKIPPED);
```
kbottla: We can remove this line. If we skip the charge it won't be tried for expunging for about 24 hours. 

In this case, charges can be tried again sooner.

src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
```diff
@@ -126,6 +129,11 @@ private void parityCheckAndExpungeIfMet(ChargeEntity chargeEntity) {
...
+                logger.info("Charge not expunged from connector because Idempotency exists for charge in AGREEMENT mode {}",
```
kbottla:
```suggestion
                logger.info("Charge not expunged from connector because Idempotency record exists for charge",
```

src/test/java/uk/gov/pay/connector/expunge/service/ChargeExpungeServiceTest.java
```diff
@@ -239,6 +245,67 @@ void expunge_shouldNotExpungeCharge_ifIsTestStripePaymentMissingFees() {
         verify(mockChargeDao, never()).expungeCharge(any(), any());
     }
 
+    @Test
+    void expunge_shouldNotExpungeChargeIfAgreementModeWithExistingIdempotency() {
```
kbottla:
```suggestion
    void expunge_shouldNotExpungeChargeIfAuthorisationModeIsAgreementModeAndIdempotencyRecordExists() {
```

src/test/java/uk/gov/pay/connector/expunge/service/ChargeExpungeServiceTest.java
```diff
@@ -239,6 +245,67 @@ void expunge_shouldNotExpungeCharge_ifIsTestStripePaymentMissingFees() {
...
+        when(mockIdempotencyDao.idempotencyExistsByResourceExternalId(resourceId)).thenReturn(true);
+
+        chargeExpungeService.expunge(1);
+        verify(mockChargeService).updateChargeParityStatus(chargeEntity.getExternalId(), SKIPPED);
+        verify(mockChargeDao, never()).expungeCharge(any(), any());
+        verify(mockIdempotencyDao).idempotencyExistsByResourceExternalId(resourceId);
+    }
+
+    @Test
+    void expunge_shouldExpungeChargeIfAgreementModeWithNOExistingIdempotency() {
```
kbottla:
```suggestion
    void expunge_shouldExpungeChargeIfAuthorisationModeIsAgreementAndIdempotencyRecordDoesNotExist() {
```

src/test/java/uk/gov/pay/connector/expunge/service/ChargeExpungeServiceTest.java
```diff
@@ -239,6 +245,67 @@ void expunge_shouldNotExpungeCharge_ifIsTestStripePaymentMissingFees() {
...
+
+        chargeExpungeService.expunge(1);
+
+        verify(mockChargeDao).expungeCharge(chargeEntity.getId(), chargeEntity.getExternalId());
+        verify(mockChargeService, never()).updateChargeParityStatus(any(),any());
+    }
+
+
+    @Test
+    void expunge_shouldExpungeChargeIfNotAgreementModeAndNOExistingIdempotency() {
```
kbottla:
```suggestion
    void expunge_shouldExpungeChargeIfAuthorisationModeIsNotAgreement() {
```

src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
AbdirizakIdris: Great spot, thanks!

kbottla (approved):

PR #4647, opened 2023-07-24
PP-11152 Don't serialize nulls for WorldpayCredentials

stephencdaly:
Using the Jdk8Module to serialize the Optional getters sensibly caused did not obey the `@JsonInclude(JsonInclude.Include.NON_NULL)` annotation for fields with getters that return Optionals.

We don't want to include null fields, particularly when serializing the JSON for writing to the database. We also prefer this for API responses to maintain consistency.

Instead of registering the Jdk8Module on the ObjectMapper, add additional getters to WorldpayCredentials that don't return Optionals for Optional fields. Add `@JsonIgnore` to the Optional getters so the other getters are used instead.


alexbishop1 (approved):
It makes me sad that you now need a PhD in Jackson to understand how `WorldpayCredentials` is serialised but this gives us the behaviour we want.

PR #4651, opened 2023-07-25
PP-11152 Stop including legacy Worldpay creds in API response

stephencdaly:
We now always return `one_off_customer_initated` in the credentials JSON in API responses, if the credentials are stored using the new or legacy structure for one-off Worldpay credentials.

All consumers of the API now use the `one_off_customer_initated` from the API response. Stop returning `merchant_id` and `username` at the root of the credentials object if the credentials are stored with the legacy structure in the database.


src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/resource/GatewayAccountCredentialsResourceIT.java
```diff
@@ -187,18 +187,15 @@ public void patchGatewayAccountCredentialsGatewayMerchantId() {
                 .then()
                 .statusCode(200)
                 .body("$", hasKey("credentials"))
-                .body("credentials", hasKey("username"))
-                .body("credentials", hasKey("merchant_id"))
-                .body("credentials", not(hasKey("password")))
+                .body("credentials", hasKey("one_off_customer_initiated"))
+                .body("credentials.one_off_customer_initiated", hasKey("username"))
+                .body("credentials.one_off_customer_initiated", hasKey("merchant_code"))
```
alexbishop1, line 192: Maybe `.body("credentials.one_off_customer_initiated", not(hasKey("password")))` if it’s not tested elsewhere?

sfount (approved):
Do we still need the properties if they're never serialised? I'm guessing there's still the option they can be used by the Connector code itself?

alexbishop1 (approved):
This should make things simpler.

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/resource/GatewayAccountCredentialsResourceIT.java
stephencdaly: It's already tested in multiple places so I left it out here

PR #4655, opened 2023-07-26
PP-11157 Finish adding support for patching Worldpay credentials

alexbishop1:
- Add a missing break statement that could have caused a nasty overwriting bug
- Make credentials for the different gateways serialise consistently, excluding null values
- Add integration tests for Worldpay credentials, splitting them out from `GatewayAccountCredentialsResourceIT` to a new `GatewayAccountCredentialsResourceWorldpayIT` class (which also includes some existing Worldpay-specific tests), which now lives alongside a `GatewayAccountCredentialsResourceWorldpay3dsFlexIT` class — while a lot of tests have moved around, only `GatewayAccountCredentialsResourceWorldpayIT` contains new tests.


stephencdaly (approved):
LGTM!

PR #4659, opened 2023-07-28
PP-11246 PACT test for failed post agreement events

james-peacock-ho:
- Add PACT test covering a request with ledger with 422 response due to missing mandatory field


SandorArpa (approved):
LGTM

PR #4660, opened 2023-07-28
PP-11152 Don't allow updating legacy credentials for Worldpay

stephencdaly:
Selfservice now uses the new PATCH paths for updating Worldpay credentials:
- `credentials/worldpay/one_off_customer_initiated`
- `credentials/worldpay/recurring_customer_initiated` 
- `credentials/worldpay/recurring_merchant_initiated`

Attempting to update credentials using the `credentials` path in a PATCH request would produce unexpected results.

Return a 400 if a PATCH request to update the credentials for a Worldpay account has `"path": "credentials"`.

Also return a 400 if POST `/v1/api/accounts/{accountId}/credentials` is called for a Worldpay account and the `credentials` field is set. Credentials are only currently included for a Stripe account when we enable PSP switch. For Worldpay accounts, we do not set credentials at this point as they are entered by the service user. Ensure an appropriate error is returned in this case to ensure we don't allow setting credentials in the legacy format.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -602,14 +602,14 @@
         "filename": "src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/resource/GatewayAccountCredentialsRequestValidatorTest.java",
         "hashed_secret": "e472e4dc23e4530d4fddeb6571b1c23447ab4c24",
         "is_verified": false,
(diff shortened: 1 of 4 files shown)


SandorArpa (approved):
LGTM

PR #4661, opened 2023-07-28
PP-11152 Update credentials state for new PATCH path

stephencdaly:
We have added a new PATCH path used for worldpay accounts to update one-off credentials - `credentials/worldpay/one_off_customer_initiated`.

Ensure that the same rules are applied for changing the state of the GatewayAccountCredentialsEntity after updating the credentials as for the `credential` PATCH path. Don't apply any specific logic for when credentials for recurring payments should change state at this point.

Duplicate the tests for this logic for requests with path `credentials/worldpay/one_off_customer_initiated`. For logic that only applies for Worldpay accounts around 3DS flex and MOTO, only have tests for when the new path is used, as the old path is no longer used for updating Worldpay credentials.

Remove some tests that tested behaviour when credentials in the requests are empty, as this can never happen because we validate that the required fields are present.


sfount (approved):
LGTM 👍 I assume coverage for the recurring channels will come with the follow-up ticket

PR #4669, opened 2023-08-03
PP-11252  Include recurring credentials for pact state

stephencdaly:
We want the pact state "a gateway account 333 with Worldpay 3DS Flex credentials exists" to include recurring Worldpay credentials as well as one-off credentials for a selfservice Pact test.

The easiest way to do this is to update the "default" credentials we insert into the database for integration and Pact tests when we don't need particular credentials to be present.


alexbishop1 (approved):

SandorArpa (approved):
LGTM

PR #4672, opened 2023-08-04
PP-11206 Add tests for setting active state

stephencdaly:
The behaviour for when we set credentials to ACTIVE for Worldpay accounts which have recurring credentials was already correct.

Add tests for this behaviour.


SandorArpa (approved):
LGTM

PR #4674, opened 2023-08-07
PP-11152 Improve AuthUtilTest

stephencdaly:
Improve the tests in AuthUtilTest to make it so that for tests where we are checking we retrieve the correct credentials for the payment type - recurring/non-recurring and authorisation mode web/agreement for recurring payments, the credentials has all 3 types of credentials set.

This improves the tests by making definitely sure we're selecting the correct credential when they are all set.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -509,6 +509,29 @@
         "line_number": 46
       }
     ],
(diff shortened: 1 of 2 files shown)


SandorArpa (approved):
LOKTM

PR #4679, opened 2023-08-09
PP-10528: New endpoint to resend charge confirmation emails

alan-gds:
`/v1/api/accounts/{accountId}/charges/{chargeId}/resend-confirmation-email`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
+++ b/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
@@ -24,6 +24,7 @@ import uk.gov.pay.connector.common.model.api.ErrorResponse;
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
+import uk.gov.pay.connector.usernotification.service.UserNotificationService;
 import uk.gov.service.payments.commons.model.AuthorisationMode;
 
 import javax.annotation.Nullable;
@@ -46,6 +47,8 @@ import java.util.Optional;
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static javax.ws.rs.core.Response.created;
 import static javax.ws.rs.core.Response.ok;
+import static uk.gov.pay.connector.util.ResponseUtil.buildErrorResponse;
+import static uk.gov.pay.connector.util.ResponseUtil.noContentResponse;
 import static uk.gov.pay.connector.util.ResponseUtil.notFoundResponse;
 import static uk.gov.pay.connector.util.ResponseUtil.responseWithChargeNotFound;
 import static uk.gov.pay.connector.util.ResponseUtil.responseWithGatewayTransactionNotFound;
@@ -73,14 +76,17 @@ public class ChargesApiResource {
     private final ChargeService chargeService;
     private final ChargeExpiryService chargeExpiryService;
     private final GatewayAccountService gatewayAccountService;
+    private final UserNotificationService userNotificationService;
 
     @Inject
     public ChargesApiResource(ChargeService chargeService,
                               ChargeExpiryService chargeExpiryService,
-                              GatewayAccountService gatewayAccountService) {
+                              GatewayAccountService gatewayAccountService,
+                              UserNotificationService userNotificationService) {
         this.chargeService = chargeService;
         this.chargeExpiryService = chargeExpiryService;
         this.gatewayAccountService = gatewayAccountService;
+        this.userNotificationService = userNotificationService;
     }
 
     @GET
@@ -197,6 +203,33 @@ public class ChargesApiResource {
         return chargeService.findCharge(accountId, telephoneChargeCreateRequest)
                 .map(response -> Response.status(200).entity(response).build())
                 .orElseGet(() -> Response.status(201).entity(chargeService.createFromTelephonePaymentNotification(telephoneChargeCreateRequest, gatewayAccount)).build());
+    }
+
+    @POST
+    @Path("/v1/api/accounts/{accountId}/charges/{chargeId}/resend-confirmation-email")
+    @Produces(APPLICATION_JSON)
+    @Operation(
+            summary = "Resend confirmation email for a charge",
+            responses = {
+                    @ApiResponse(responseCode = "204", description = "No content"),
+                    @ApiResponse(responseCode = "402", description = "Could not send email"),
+                    @ApiResponse(responseCode = "404", description = "Not found - charge not found"),
+                    @ApiResponse(responseCode = "500", description = "Internal server error")
+            }
+    )
+    public Response resendConfirmationEmail(
+            @Parameter(example = "1", description = "Gateway account ID")
+            @PathParam(ACCOUNT_ID) Long accountId,
+            @Parameter(example = "spmh0fb7rbi1lebv1j3f7hc3m9", description = "Charge external ID")
+            @PathParam("chargeId") String chargeId) {
+        var account = gatewayAccountService.getGatewayAccount(accountId)
+                .orElseThrow(() -> new GatewayAccountNotFoundException(accountId));
+
+        return chargeService.findCharge(chargeId, accountId)
+                            .map(charge -> userNotificationService.sendPaymentConfirmedEmailSynchronously(charge, account)
+                            .map((reference) -> noContentResponse())
+                            .orElseGet(() -> buildErrorResponse(Response.Status.PAYMENT_REQUIRED, "Failed to send email")))
+                .orElseGet(() -> responseWithChargeNotFound(chargeId));
     }
 
     @POST
--- a/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
@@ -74,46 +74,77 @@ public class UserNotificationService {
     }
 
     public Future<Optional<String>> sendPaymentConfirmedEmail(ChargeEntity chargeEntity, GatewayAccountEntity gatewayAccountEntity) {
-        return sendEmail(EmailNotificationType.PAYMENT_CONFIRMED, Charge.from(chargeEntity), gatewayAccountEntity,
-                buildConfirmationEmailPersonalisationFrom(chargeEntity));
+        var charge = Charge.from(chargeEntity);
+        return sendEmail(EmailNotificationType.PAYMENT_CONFIRMED, charge, gatewayAccountEntity,
+                buildConfirmationEmailPersonalisationFrom(charge, gatewayAccountEntity));
+    }
+
+    public Optional<String> sendPaymentConfirmedEmailSynchronously(Charge charge, GatewayAccountEntity gatewayAccountEntity) {
+        return sendEmailSynchronously(EmailNotificationType.PAYMENT_CONFIRMED, charge, gatewayAccountEntity,
+                buildConfirmationEmailPersonalisationFrom(charge, gatewayAccountEntity));
     }
 
     private Future<Optional<String>> sendEmail(EmailNotificationType emailNotificationType, Charge charge, GatewayAccountEntity gatewayAccountEntity, HashMap<String, String> personalisation) {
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -199,6 +209,48 @@ public Response createNewTelephoneCharge(
...
+            @Parameter(example = "1", description = "Gateway account ID")
+            @PathParam(ACCOUNT_ID) Long accountId,
+            @Parameter(example = "spmh0fb7rbi1lebv1j3f7hc3m9", description = "Charge external ID")
+            @PathParam("chargeId") String chargeId) {
+        var account = gatewayAccountService.getGatewayAccount(accountId)
+                .orElseThrow(() -> new GatewayAccountNotFoundException(accountId));
+
+        return chargeService.findCharge(chargeId, accountId)
+                .map(charge -> {
+                    var chargeEntity = chargeService.findChargeByExternalId(chargeId);
```
stephencdaly: We're getting a `Charge` using `findCharge`  and then getting a `ChargeEntity` using `findChargeByExternalId`.

I think the right thing to do here is to just get the `Charge` and use that to send the email. If you look at the logic of what these 2 methods do, `findCharge` first tries look for the charge in connector's database, and if it doesn't find it there, makes a request to ledger to get it. This is to deal with the case where charges have been expunged from connector, which currently happens after a day.

`findChargeByExternalId` however, only looks in connector's database.

When resending an email, it might be quite likely the charge has been expunged, so we need to deal with this.

My suggestion would be to start by updating `UserNotificationService$sendPaymentConfirmedEmail` to accept a `Charge` rather than a `ChargeEntity`, or adding a separate method that accepts a `Charge` and does any extra steps to get all the info needed for the email. The difference between these 2 types is that a `ChargeEntity` represents a charge from connector's database but a `Charge` represents a charge obtained from either connector or ledger. Hopefully the `Charge` already has all the information we need to send the email.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -199,6 +209,48 @@ public Response createNewTelephoneCharge(
...
+
+                    try {
+                        var success = userNotificationService.sendPaymentConfirmedEmail(chargeEntity, account)
+                                .get(5, TimeUnit.SECONDS)
```
stephencdaly: I think that this method currently returns a `Future` because we don't want to wait around for the email to send when we send as part of the capture process.

I think for this case, we're happy waiting around for a response - and the timeout should probably just be the standard timeout for the Notify client. What I'd probably suggest here is creating a separate method in `UserNotificationService` that synchronously makes the request to Notify, making sure to share the code that makes the request.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -199,6 +209,48 @@ public Response createNewTelephoneCharge(
...
+                                .isPresent();
+                        if (success) {
+                            return noContentResponse();
+                        }
+                        else {
+                            return gatewayErrorResponse("Failed to send email");
```
stephencdaly: This will return the correct thing, as we want to return a 402, and `gatewayErrorResponse` returns a 402. However, it's slightly confusing as a "gateway error" would usually refer to an error contacting a payment gateway - Worldpay, Stripe etc.

I'd be inclined to directly call `buildErrorResponse(PAYMENT_REQUIRED, message);` or rename the `gatewayErrorResponse` method if you can think of a suitable name to avoid this confusion.

src/test/java/uk/gov/pay/connector/charge/resource/ChargesApiResourceTest.java
```diff
@@ -102,4 +111,84 @@ void createCharge_idempotencyKeyEmpty_shouldReturn422() {
...
+    void resendConfirmationEmail_accountNotFound_shouldReturn404() {
+        var accountId = 1234L;
+        var chargeId = "charge-id";
+
+        when(gatewayAccountService.getGatewayAccount(accountId)).thenReturn(Optional.empty());
+
+        Response response = resources
+                .target(String.format("/v1/api/accounts/%d/charges/%s/resend-confirmation-email", accountId, chargeId))
+                .request()
+                .header("Idempotency-Key", "")
```
stephencdaly: The `.header("Idempotency-Key", "")` won't mean anything for this request, best to leave it out to avoid confusion.

src/test/java/uk/gov/pay/connector/charge/resource/ChargesApiResourceTest.java
```diff
@@ -102,4 +111,84 @@ void createCharge_idempotencyKeyEmpty_shouldReturn422() {
...
+                .thenReturn(CompletableFuture.completedFuture(Optional.of("Email sent")));
+
+        Response response = resources
+                .target(String.format("/v1/api/accounts/%d/charges/%s/resend-confirmation-email", accountId, chargeId))
+                .request()
+                .header("Idempotency-Key", "")
+                .post(Entity.json(Collections.emptyMap()));
+
+        assertThat(response.getStatus(), is(204));
+    }
```
stephencdaly, line 183: Would be good to add one integration test in addition to these unit tests.

stephencdaly (dismissed):
Logic all looks good, we just need to consider expunged charges and it's probably better if we get rid of the Future stuff with the timeout.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -199,6 +209,40 @@ public Response createNewTelephoneCharge(
...
+                .map(charge -> {
+                    var success = userNotificationService.sendPaymentConfirmedEmailSynchronously(charge, account)
+                            .isPresent();
+                    if (success) {
+                        return noContentResponse();
+                    }
+                    else {
+                        return gatewayErrorResponse("Failed to send email");
+                    }
+                })
```
stephencdaly: A more Java 8+ way of doing this would be something like
```suggestion
                                .map(charge -> userNotificationService.sendPaymentConfirmedEmailSynchronously(charge, account)
                            .map((reference) -> noContentResponse())
                            .orElseGet(() -> gatewayErrorResponse("Failed to send email")))
```

src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
```diff
@@ -74,8 +74,18 @@ public Future<Optional<String>> sendRefundIssuedEmail(RefundEntity refundEntity,
     }
 
     public Future<Optional<String>> sendPaymentConfirmedEmail(ChargeEntity chargeEntity, GatewayAccountEntity gatewayAccountEntity) {
-        return sendEmail(EmailNotificationType.PAYMENT_CONFIRMED, Charge.from(chargeEntity), gatewayAccountEntity,
-                buildConfirmationEmailPersonalisationFrom(chargeEntity));
+        var charge = Charge.from(chargeEntity);
+        return sendPaymentConfirmedEmail(charge, gatewayAccountEntity);
+    }
+
+    public Future<Optional<String>> sendPaymentConfirmedEmail(Charge charge, GatewayAccountEntity gatewayAccountEntity) {
```
stephencdaly: I think this is only called by the method above, so it could be inlined or made private?

src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
```diff
@@ -74,8 +74,18 @@ public Future<Optional<String>> sendRefundIssuedEmail(RefundEntity refundEntity,
...
+        return sendEmail(EmailNotificationType.PAYMENT_CONFIRMED, charge, gatewayAccountEntity,
+                buildConfirmationEmailPersonalisationFrom(charge, gatewayAccountEntity));
+    }
+
+    public Optional<String> sendPaymentConfirmedEmailSynchronously(Charge charge, GatewayAccountEntity gatewayAccountEntity) {
```
stephencdaly, line 82: We should probably add unit tests for this new method

src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
```diff
@@ -94,26 +104,47 @@ private Future<Optional<String>> sendEmail(EmailNotificationType emailNotificati
...
+
+        if (!emailNotifyGloballyEnabled || !isEmailEnabled || gatewayAccountEntity.getEmailCollectionMode().equals(OFF) ||
+                gatewayAccountEntity.getEmailCollectionMode().equals(OPTIONAL) && ofNullable(charge.getEmail()).isEmpty()) {
+            return Optional.empty();
+        }
+
+        if (charge.getEmail() == null) {
+            logger.warn("Cannot send email for charge_external_id = {} because the charge does not have an email address", charge.getExternalId());
+            return Optional.empty();
+        }
```
stephencdaly, line 126: Yeh should probably try to de-duplicate the code up until here? It looks shared between the synchronous and non-synchronous methods?

src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
alan-gds: Good spot.  This overload stopped being used once I added an additional overload for sending the email synchronously.

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
alan-gds: Yeah, it felt clunky when I wrote it; definitely prefer the suggestion.  I use a functional style instinctually when I'm writing Scala, but somehow it doesn't come so naturally to me in Java. 😅

src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java
```diff
@@ -199,6 +205,33 @@ public Response createNewTelephoneCharge(
...
+    public Response resendConfirmationEmail(
+            @Parameter(example = "1", description = "Gateway account ID")
+            @PathParam(ACCOUNT_ID) Long accountId,
+            @Parameter(example = "spmh0fb7rbi1lebv1j3f7hc3m9", description = "Charge external ID")
+            @PathParam("chargeId") String chargeId) {
+        var account = gatewayAccountService.getGatewayAccount(accountId)
+                .orElseThrow(() -> new GatewayAccountNotFoundException(accountId));
+
+        return chargeService.findCharge(chargeId, accountId)
+                                .map(charge -> userNotificationService.sendPaymentConfirmedEmailSynchronously(charge, account)
```
stephencdaly: Indentation looks a little off here

src/test/java/uk/gov/pay/connector/charge/resource/ChargesApiResourceResendConfirmationEmailIT.java
```diff
@@ -0,0 +1,121 @@
...
+    public void setup() {
+        databaseTestHelper = testContext.getDatabaseTestHelper();
+        accountId = String.valueOf(RandomUtils.nextInt());
+        addGatewayAccount();
+    }   
+
+    @Test
+    public void shouldReturn204WhenEmailSuccessfullySent() throws Exception {
+        String transactionId = String.valueOf(RandomUtils.nextInt());
+        String paymentProvider = "epdq";
```
stephencdaly: I'd just select a different payment provider for this test as epdq is currently being retired - sandbox should do

src/test/java/uk/gov/pay/connector/usernotification/service/UserNotificationServiceTest.java
```diff
@@ -167,6 +167,30 @@ void shouldSendPaymentConfirmationEmailWithReferenceInCustomParagraphIfEmailNoti
         assertThat(maybeNotificationId.get(), is(notificationId.toString()));
     }
 
+    @Test
+    void shouldSendPaymentConfirmationEmailSynchronouslyIfEmailNotifyIsEnabled() throws Exception {
```
stephencdaly, line 171: We probably just need some additional unit tests for the cases when the email won't be sent. It would just be duplicating all the tests like `shouldNotSendEmail_IfEmailEnabledButChargeDoesNotHaveAnEmailAddress` for this method.
We should do this because although the code for this is shared in a private method for this class, if any modifications are made to the public method under test to change how this code is called, we might break this functionality and not be able to tell.

kbottla (approved):

PR #4684, opened 2023-08-15
PP-11319 Allow enabling wallets for stripe

stephencdaly:
Previously, we only allowed updating the `allow_apple_pay` and `allow_google_pay` flags for Worldpay accounts. Allow this for Stripe accounts too.


SandorArpa (approved):
LOKTM

PR #4685, opened 2023-08-15
PP-11319 Return allow_google_pay flag correctly for Stripe

stephencdaly:
For Worldpay accounts, we require a `gateway_merchant_id` to be set on the credentials for Google Pay to be enabled, and so only return `"allow_google_pay"=true`in API responses for a gateway account if the `allow_google_pay` flag is set to true AND the `gateway_merchant_id` is set.

Stripe does not require the credentials to be configured in a particular way, so only look at the `allow_google_pay` flag on the gateway account when returning this flag in API responses.


SandorArpa (approved):
LOKM

PR #4690, opened 2023-08-18
PP-11272: Remove EPDQ config

oswaldquek:
As part of PP-11272, we want to remove EPDQ-related environment variables. This means removing the EPDQ config which depends on EPDQ environment variables. This PR deletes the minimum required to get PP-11272 unblocked; there are certainly more that can be deleted but this can be done at a later time.

TODO: check we don't have EPDQ smoke and e2e tests before deploying.


src/test/java/uk/gov/pay/connector/it/resources/DiscrepancyResourceIT.java
```diff
@@ -1,172 +0,0 @@
...
-import static java.time.temporal.ChronoUnit.HOURS;
-import static java.util.Collections.singletonList;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static uk.gov.pay.connector.util.JsonEncoder.toJson;
-
-@RunWith(DropwizardJUnitRunner.class)
-@DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml")
-public class DiscrepancyResourceIT extends ChargingITestBase {
```
stephencdaly, line 30: The discrepancy checker is also useful for Worldpay, so I think we should keep these tests but make them for Worldpay?

stephencdaly (changes requested):
Looks good, we should just make sure we test things on test-12 before deploying to prod to make sure we're happy that if anyone tries to create an ePDQ payment or refund that connector doesn't get in an odd state.

Do we also want to make sure there are no ePDQ payments in the connector database on all environments before deploying, just to make sure we don't inadvertently run into the situation that occurred when we tried to remove Smartpay code? I don't remember the exact cause of that.

Just a comment about some tests that have been removed that I think we want to keep.

src/test/java/uk/gov/pay/connector/it/resources/GatewayCleanupResourceIT.java
```diff
@@ -1,130 +0,0 @@
...
-import static uk.gov.pay.connector.gateway.PaymentGatewayName.SANDBOX;
-import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
-import static uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialState.RETIRED;
-import static uk.gov.pay.connector.it.dao.DatabaseFixtures.withDatabaseTestHelper;
-import static uk.gov.pay.connector.util.AddChargeParams.AddChargeParamsBuilder.anAddChargeParams;
-import static uk.gov.pay.connector.util.AddGatewayAccountCredentialsParams.AddGatewayAccountCredentialsParamsBuilder.anAddGatewayAccountCredentialsParams;
-
-@RunWith(DropwizardJUnitRunner.class)
-@DropwizardConfig(app = ConnectorApp.class, config = "config/test-it-config.yaml")
-public class GatewayCleanupResourceIT extends ChargingITestBase {
```
stephencdaly, line 34: Likewise for this, the gateway cleanup is now also used for Worldpay and Stripe, so I think we should just change the payment provider used for these tests.

src/test/java/uk/gov/pay/connector/it/resources/DiscrepancyResourceIT.java
```diff
@@ -105,37 +115,19 @@ public void shouldReportOnChargesThatAreInErrorStatesInGatewayAccount() {
...
+        assertEquals("EXPIRED", statusComparisonForChargeId.get("payStatus").asText());
+        assertEquals(chargeId, statusComparisonForChargeId.get("chargeId").asText());
+        assertEquals("Worldpay query response (orderCode: MyUniqueTransactionId!12, lastEvent: REFUSED, ISO8583ReturnCode code: 5, ISO8583ReturnCode description: REFUSED)", 
+                statusComparisonForChargeId.get("rawGatewayResponse").asText());
+        assertEquals("EXTERNAL_FAILED_REJECTED", statusComparisonForChargeId.get("gatewayExternalStatus").asText());
+        assertEquals("EXTERNAL_FAILED_EXPIRED", statusComparisonForChargeId.get("payExternalStatus").asText());
     }
-
-    @Test
-    public void shouldHandleCaseWhereChargeIsInUnknownState() {
```
oswaldquek, line 117: I have removed this test as there's no worldpay equivalent for `epdqMockClient.mockUnknown()` two lines below.

src/test/java/uk/gov/pay/connector/it/resources/DiscrepancyResourceIT.java
```diff
@@ -146,27 +138,37 @@ public void should404_whenAChargeIdDoesntExist() {
     @Test
     public void shouldProcessDiscrepanciesWherePayStateIsExpiredAndGatewayStateIsAuthorised() {
         String chargeId = addCharge(ChargeStatus.EXPIRED, "ref", Instant.now().minus(8, DAYS), "irrelevant");
-        String chargeId2 = addCharge(ChargeStatus.EXPIRED, "ref", Instant.now().minus(8, DAYS), "irrelevant");
```
oswaldquek, line 149: I removed chargeId2 because:
- it is exactly the same as chargeId1
- [mockAuthorisationQueryAndCancel](https://github.com/alphagov/pay-connector/pull/4690/files#diff-df1f6d7a0271ccf4b4021db42324138fdb2b2383a90ca7529c9d5a01dbe364bcR162) would be more complicated as I'd have to set the scenario state for two charge ids: STARTED -> "payment status queried" -> STARTED -> "payment status queried"

stephencdaly (approved):
LGTM! Just be careful when deploying

PR #4692, opened 2023-08-23
PP-11364 deserialise additional applay pay fields

SandorArpa:
## WHAT YOU DID

Stripe requires us to send some additional data from PKPaymentToken when we create a token for Apple Pay.

- deserialise additional fields coming from frontend


stephencdaly (approved):
LGTM!

PR #4697, opened 2023-08-29
PP-11365 Use `payment_data` string when decrypting Apple Pay data

kbottla:
## WHAT YOU DID
- Currently, connector uses `encrypted_payment_data` values when decrypting Apple Pay data. Frontend is sending both JSON (in `encrypted_payment_data`) and JSON as a string (`payment_data`). See https://github.com/alphagov/pay-frontend/pull/3635.
- We are going to use `payment_data` (String) to avoid any issues when deserialising encrypted_payment_data and then serialising data to send to Stripe.
-  `encrypted_payment_data` field is to be removed in a separate PR.


james-peacock-ho (approved):
LGTM

PR #4699, opened 2023-08-30
PP-11365 Remove `encrypted_payment_data` field

kbottla:
## WHAT YOU DID
- Removed `encrypted_payment_data` field as Connector now uses `payment_data` (JSON as string) to decrypt Apple Pay data.


SandorArpa (approved):
LGTM

PR #4704, opened 2023-09-01
PP-11394 removed unused google pay endpoint

SandorArpa:
## WHAT YOU DID

PP-11393 adds a separate endpoint for authorising Stripe goole pay payments. Rename the existing endpoint to make it clear it is for Worldpay.

- remove unused endpoint
- refactor tests to use new endpoint


stephencdaly (approved):
LGTM!

PR #4708, opened 2023-09-07
PP-11465 move apple pay decryption to worldpay code

james-peacock-ho:
Context: ApplePay wallet payments authorisations via Worldpay require the payment data to be decrypted in connector prior to sending the request to Worldpay. As this step is not required for Stripe wallet payment authorisations, it should be moved into the Worldpay-specific code in connector, to simplify subsequent ApplePay/Stripe implementation.

- Postpone the ApplePay decryption step until building the Worldpay request as decryption isn’t required for Stripe. This means WalletAuthorisationData object isn’t required until the Worldpay-specific code.
- Add wallet type to the WalletAuthorisationRequest to determine whether decryption is needed (decryption isn’t required for Google Pay).
- Extract and store card expiry date from Worldpay response, when it is available and in expected format (it will only be returned in response to successful authorisations).
- Update tests accordingly


src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayService.java
```diff
@@ -15,20 +15,8 @@ public class ApplePayService extends WalletService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ApplePayService.class);
 
-    private ApplePayDecrypter applePayDecrypter;
-
     @Inject
-    public ApplePayService(ApplePayDecrypter applePayDecrypter, WalletAuthoriseService authoriseService) {
+    public ApplePayService(WalletAuthoriseService authoriseService) {
```
stephencdaly: Can we just delete the ApplePayService and GooglePayService subclasses now? Looks like all they do is set the `walletType` for the WalletService, which we can get from the `WalletAuthorisationRequest`

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -165,6 +177,19 @@ public String getRefusedReturnCode() {
...
+    @Override
+    public Optional<CardExpiryDate> getCardExpiryDate() {
+        if (expiryDateMonth != null && expiryDateYear != null) {
+            try {
+                return Optional.of(CardExpiryDate.valueOf(expiryDateMonth + "/" + expiryDateYear.substring(2, 4)));
+            } catch (IndexOutOfBoundsException ex) {
+                LOGGER.error(format("Expiry date year in Worldpay response for transaction {} is in an unexpected format." + transactionId));
+                return Optional.empty();
+            }
+        }
```
alexbishop1: Oh, yikes, this is a bit 😬!

This is what I would do:

1. Add a new static `valueOf(YearMonth)` method to `CardExpiryDate`
2. Use regexes (or similar) to check that `expiryDateMonth` and `expiryDateYear` are in the expected formats
3. Assuming they are, convert them to `ints` and use `YearMonth.of(int, int)` to get a `YearMonth`
5. Call the new `CardExpiryDate,valueOf(YearMonth)` method to get a `CardExpiryDate`

src/main/java/uk/gov/pay/connector/gateway/worldpay/wallets/WorldpayWalletAuthorisationHandler.java
```diff
@@ -52,26 +65,44 @@ private GatewayOrder buildWalletAuthoriseOrder(WalletAuthorisationGatewayRequest
...
                 .withDescription(request.getDescription())
                 .withAmount(request.getAmount())
                 .build();
     }
+
+    private WalletAuthorisationData extractWalletAuthorisationData(WalletAuthorisationGatewayRequest request) {
+        if (request.getWalletAuthorisationRequest().getWalletType() == WalletType.APPLE_PAY) {
+            return decryptApplePaymentData(request.getGovUkPayPaymentId(), (ApplePayAuthRequest) request.getWalletAuthorisationRequest());
+        }
+        return (GooglePayAuthRequest) request.getWalletAuthorisationRequest();
```
alexbishop1: A `switch` statement might work better here. It does almost feel like `WalletAuthorisationRequest` should be an interface with Apple Pay- and Google Pay-specific subtypes.

src/main/java/uk/gov/pay/connector/gateway/worldpay/wallets/WorldpayWalletAuthorisationHandler.java
```diff
@@ -52,26 +65,44 @@ private GatewayOrder buildWalletAuthoriseOrder(WalletAuthorisationGatewayRequest
...
+    }
+    private WalletAuthorisationData decryptApplePaymentData(String chargeId, ApplePayAuthRequest applePayAuthRequest) {
+        LOGGER.info("Decrypting apple pay payload for charge with id {}", chargeId);
```
alexbishop1: “Apple Pay”

src/main/java/uk/gov/pay/connector/gateway/worldpay/wallets/WorldpayWalletAuthorisationHandler.java
```diff
@@ -52,26 +65,44 @@ private GatewayOrder buildWalletAuthoriseOrder(WalletAuthorisationGatewayRequest
...
+        AppleDecryptedPaymentData result = applePayDecrypter.performDecryptOperation(applePayAuthRequest);
+        result.setPaymentInfo(applePayAuthRequest.getPaymentInfo());
+        LOGGER.info("Finished decryption for id {}", chargeId);
```
alexbishop1: Do we really need a log line saying we finished decrypting?

src/main/java/uk/gov/pay/connector/wallets/WalletAuthorisationRequestToAuthCardDetailsConverter.java
```diff
@@ -0,0 +1,21 @@
+package uk.gov.pay.connector.wallets;
+
+import uk.gov.pay.connector.gateway.model.AuthCardDetails;
+import uk.gov.service.payments.commons.model.CardExpiryDate;
+
+import java.time.format.DateTimeFormatter;
```
alexbishop1: Is this ever used?

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -86,16 +86,18 @@ public GatewayResponse<BaseAuthoriseResponse> doAuthorise(String chargeId, Walle
             Optional<ProviderSessionIdentifier> sessionIdentifier = operationResponse.getSessionIdentifier();
             Optional<Auth3dsRequiredEntity> auth3dsDetailsEntity =
                     operationResponse.getBaseResponse().flatMap(BaseAuthoriseResponse::extractAuth3dsRequiredDetails);
-
-            logMetrics(charge, operationResponse, requestStatus, walletAuthorisationData.getWalletType());
+            Optional<CardExpiryDate> cardExpiryDate = operationResponse.getBaseResponse().flatMap(BaseAuthoriseResponse::getCardExpiryDate);
```
alexbishop1: Some would argue we shouldn’t have variables of type `Optional` and this should either be a `CardExpiryDate` or null.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponseTest.java
```diff
@@ -91,4 +93,25 @@ void WorldpayAuthorisationRejectedCode_should_be_mapped_to_MappedAuthorisationRe
         assertTrue(worldpayOrderStatusResponse.toString().contains("Mapped rejection reason: DO_NOT_HONOUR"));
     }
 
+    @Test
+    void get_expiry_date_should_return_correctly_when_expiry_date_fields_are_present() throws Exception {
+        String response = load(WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE);
+        WorldpayOrderStatusResponse worldpayOrderStatusResponse = XMLUnmarshaller.unmarshall(response, WorldpayOrderStatusResponse.class);
+        assertThat(worldpayOrderStatusResponse.getCardExpiryDate().isPresent(), is(true));
+        assertThat(worldpayOrderStatusResponse.getCardExpiryDate().get(), is(CardExpiryDate.valueOf("11/99")));
```
alexbishop1: This is really minor but can we pick an expiry date in this millennium?

alexbishop1 (changes requested):
Fundamentally this just moves the decryption code somewhere else and looks fine. Some suggestions for a few specifics.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
james-peacock-ho: Agreed, I was 😬 too. Are you suggesting that 2,3,4 above should all be done within this getter method? If expiryDateMonth and expiryDateYear aren't in expected format, would you agree catching and logging as an error and returning Optional.empty() is appropriate or is there a better approach?

alexbishop1: Yes, something like this:

```java
// Way up near the beginning of the file in constantland
private static final Pattern TWO_DIGIT_MONTH = Pattern.compile("[0-9]{2}");
private static final Pattern FOUR_DIGIT_YEAR = Pattern.compile("[1-9][0-0}{3}");

// Back down at the getter
public Optional<CardExpiryDate> getCardExpiryDate() {
    if (expiryDateMonth == null || !TWO_DIGIT_MONTH.matcher(expiryDateMonth).matches()) {
        LOGGER.error(format("Expiry date month in Worldpay response for transaction {} is in an unexpected format.", transactionId));
        return Optional.empty();
    }
    
    if (expiryDateYear == null || !FOUR_DIGIT_YEAR.matcher(expiryDateYear).matches()) {
        LOGGER.error(format("Expiry date year in Worldpay response for transaction {} is in an unexpected format.", transactionId));
        return Optional.empty();
    }

    var expiryDateYearMonth = YearMonth.of(Integer.parseInt(expiryDateYear), Integer.parseInt(expiryDateMonth));
    return Optional.of(CardExpiryDate.valueOf(expiryDateYearMonth));
}

File: src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayService.java (hunk 1 above)
Comment (james-peacock-ho) [reply src/main/java/uk/gov/pay/connector/wallets/applepay/ApplePayService.java]: done!

File: src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java (hunk 2 above)
Comment (james-peacock-ho) [reply src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java]: done!

File: src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java [hunk 9]
```diff
@@ -26,11 +32,13 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
     private static final String WORLDPAY_AUTHORISED_EVENT = "AUTHORISED";
     private static final String WORLDPAY_REFUSED_EVENT = "REFUSED";
     private static final String WORLDPAY_CANCELLED_EVENT = "CANCELLED";
-
     public static final String WORLDPAY_RECURRING_AUTH_TOKEN_PAYMENT_TOKEN_ID_KEY = "paymentTokenID";
-
     public static final String WORLDPAY_RECURRING_AUTH_TOKEN_TRANSACTION_IDENTIFIER_KEY = "schemeTransactionIdentifier";
+    private static final Pattern TWO_DIGIT_MONTH = Pattern.compile("[0-9]{2}");
+    private static final Pattern FOUR_DIGIT_YEAR = Pattern.compile("[1-9][0-9]{3}");
```
alexbishop1: Since `CardExpiryDate` only supports years in the range 2000–2099, this should probably be `20[0-9]{2}`.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -165,6 +179,29 @@ public String getRefusedReturnCode() {
...
+        }
+        
+        var expiryDateYearMonth = YearMonth.of(Integer.parseInt(expiryDateYear), Integer.parseInt(expiryDateMonth));
+        CardExpiryDate expiryDate;
+        try {
+            expiryDate = CardExpiryDate.valueOf(expiryDateYearMonth);
+        } catch (IllegalArgumentException ex) {
+            LOGGER.error(format("Expiry date in Worldpay response for transaction {} is outside of the expected range. {}", transactionId, ex.getMessage()));
+            return Optional.empty();
+        }
```
alexbishop1: If the validation regex is good enough, this try/catch should be unnecessary.

src/main/java/uk/gov/pay/connector/gateway/worldpay/wallets/WorldpayWalletAuthorisationHandler.java
```diff
@@ -52,26 +65,48 @@ private GatewayOrder buildWalletAuthoriseOrder(WalletAuthorisationGatewayRequest
...
     }
+
+    private WalletAuthorisationData extractWalletAuthorisationData(WalletAuthorisationGatewayRequest request) {
+        switch (request.getWalletAuthorisationRequest().getWalletType()) {
+            case APPLE_PAY:
+                return decryptApplePaymentData(request.getGovUkPayPaymentId(), (ApplePayAuthRequest) request.getWalletAuthorisationRequest());
+            case GOOGLE_PAY:
+                return (GooglePayAuthRequest) request.getWalletAuthorisationRequest();
+            default:
+                throw new IllegalArgumentException("Wallet Type not recognised");
```
alexbishop1: We should include what `request.getWalletAuthorisationRequest().getWalletType()` in the exception so we know what unrecognised value we’ve been given.

src/test/java/uk/gov/pay/connector/wallets/WalletServiceTest.java
```diff
@@ -0,0 +1,147 @@
...
+                .withSessionIdentifier(ProviderSessionIdentifier.of("234"))
+                .build();
+        when(worldpayResponse.authoriseStatus()).thenReturn(BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED);
+        when(mockWalletAuthoriseService.doAuthorise(externalChargeId, applePayAuthRequest)).thenReturn(gatewayResponse);
+
+        Response authorisationResponse = walletService.authorise(externalChargeId, applePayAuthRequest);
+
+        verify(mockWalletAuthoriseService).doAuthorise(externalChargeId, applePayAuthRequest);
+        assertThat(authorisationResponse.getStatus(), is(200));
+        assertThat(authorisationResponse.getEntity(), is(ImmutableMap.of("status", "AUTHORISATION SUCCESS")));
```
alexbishop1: We can use `Map.of(…)` here.

alexbishop1 (changes requested):
Almost there!

alexbishop1 (approved):

alexbishop1:
Please squash the commits when merging.

PR #4713, opened 2023-09-15
PP-11465 Correct logging for expiry dates in Worldpay wallet responses

james-peacock-ho:
Context: Currently, a Sentry alert is triggered for missing expiry dates in Worldpay responses to wallet authorisations.
This is overly sensitive as the expiry date is not expected in decline or error responses, only in some success responses.

- Correct conditional statements so that:
  - log at info level when expiry date is missing from a Worldpay wallet authorisation success response
  - log at error level when expiry date is present but in an unexpected format for any Worldpay wallet authorisation response
- Update/add tests accordingly

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -181,20 +181,25 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
 
     @Override
     public Optional<CardExpiryDate> getCardExpiryDate() {
-        if (expiryDateMonth == null || !TWO_DIGIT_MONTH.matcher(expiryDateMonth).matches()) {
-            LOGGER.error(format("Expiry date month in Worldpay response for transaction {} is in an unexpected format.", transactionId));
+        if (expiryDateMonth == null && expiryDateYear == null) {
+            if (AuthoriseStatus.AUTHORISED.toString().equals(lastEvent)) {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -181,20 +181,25 @@ public String getRefusedReturnCode() {
...
+                LOGGER.info("Expiry date is not included in Worldpay wallet payment authorisation success response.");
+            }
             return Optional.empty();
         }
-
-        if (expiryDateYear == null || !FOUR_DIGIT_YEAR.matcher(expiryDateYear).matches()) {
-            LOGGER.error(format("Expiry date year in Worldpay response for transaction {} is in an unexpected format.", transactionId));
+        if (expiryDateMonth == null || !TWO_DIGIT_MONTH.matcher(expiryDateMonth).matches()
+                || expiryDateYear == null || !FOUR_DIGIT_YEAR.matcher(expiryDateYear).matches()) {
+            int monthDigits = (expiryDateMonth == null) ? 0 : expiryDateMonth.length();
```
stephencdaly, line 192: Maybe a bit overkill here with the logic for if only one of expiry date/month is present? Don't see an issue with it, but also wouldn't see an issue with just logging that the expiry date isn't there if either the month/year aren't present.

So changing 
`if (expiryDateMonth == null && expiryDateYear == null) {`
to
`if (expiryDateMonth == null || expiryDateYear == null) {`

stephencdaly (approved):
LGTM! Probably just ignore my comment and get this in though

PR #4714, opened 2023-09-15
PP-11548: Use our own abstraction of prometheus counters

jfharden:
## WHAT YOU DID
1. Now the ADOT sidecar is adding the default labels we need, stop using our custom implementation and revert to a simpler setup.
2. Add state-transition prometheus metrics
3. Update the local build script to take account for arm computers

A note on the metric. There are 3 dropwizard metrics:
```java
        incrementPerGatewayStateTransitionCounter(targetChargeState, chargeEventEntity);
        incrementPerGatewayStateTransitionMeter(targetChargeState, chargeEventEntity);
        incrementPerGatewayAccountStateTransitionCounter(targetChargeState, chargeEventEntity);
```
incrementPerGatewayStateTransitionMeter is a rate, we don't need that, promql will calculate that for us no problem.
The other two both create the same increment as far as I can see, which would mean our metric is double counted currently:

https://github.com/alphagov/pay-connector/blob/1975f9a31c1e80be13454c6e5fc27867485ffe60/src/main/java/uk/gov/pay/connector/queue/statetransition/StateTransitionService.java#L92-L97

https://github.com/alphagov/pay-connector/blob/1975f9a31c1e80be13454c6e5fc27867485ffe60/src/main/java/uk/gov/pay/connector/queue/statetransition/StateTransitionService.java#L108-L113

### Logging

### Documentation


src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -144,7 +143,8 @@ public void run(ConnectorConfiguration configuration, Environment environment) {
         environment.jersey().register(new ConstraintViolationExceptionMapper());
         environment.jersey().register(new ValidationExceptionMapper());
         environment.jersey().register(new UnsupportedOperationExceptionMapper());
-        environment.jersey().register(new LoggingExceptionMapper<>() {});
+        environment.jersey().register(new LoggingExceptionMapper<>() {
```
jfharden, line 146: This was the auto formatter so I'm going to leave it in.

kbottla (approved):

PR #4716, opened 2023-09-18
PP-11385 Extract expiry date from Stripe auth response

stephencdaly:
For Stripe wallet payments, we are not able to obtain the expiry date for the card before authorising. The payment method, included in the payment intent response includes the expiry date. Extract the expiry date from the authorisation response so we can store it on the charge for wallet payments.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
@@ -1,10 +1,15 @@
 package uk.gov.pay.connector.gateway.stripe;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.gateway.model.Gateway3dsRequiredParams;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
 import uk.gov.pay.connector.gateway.stripe.response.Stripe3dsRequiredParams;
 import uk.gov.pay.connector.gateway.stripe.response.StripePaymentIntentResponse;
+import uk.gov.service.payments.commons.model.CardExpiryDate;
 
+import java.time.DateTimeException;
+import java.time.YearMonth;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -16,6 +21,8 @@ public class StripeAuthorisationResponse implements BaseAuthoriseResponse {
     public static final String STRIPE_RECURRING_AUTH_TOKEN_CUSTOMER_ID_KEY = "customerId";
     public static final String STRIPE_RECURRING_AUTH_TOKEN_PAYMENT_METHOD_ID_KEY = "paymentMethodId";
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(StripeAuthorisationResponse.class);
+
     private final String transactionId;
     private final AuthoriseStatus authoriseStatus;
     private final String redirectUrl;
@@ -23,13 +30,15 @@ public class StripeAuthorisationResponse implements BaseAuthoriseResponse {
     private final String stringifiedResponse;
     private final String customerId;
     private final String paymentMethodId;
+    private final CardExpiryDate cardExpiryDate;
 
     private StripeAuthorisationResponse(String transactionId,
                                         AuthoriseStatus authoriseStatus,
                                         String redirectUrl,
                                         String stringifiedResponse,
                                         String customerId,
-                                        String paymentMethodId) {
+                                        String paymentMethodId,
+                                        CardExpiryDate cardExpiryDate) {
         this.customerId = customerId;
         this.paymentMethodId = paymentMethodId;
         if (Objects.isNull(authoriseStatus)) {
@@ -39,6 +48,7 @@ public class StripeAuthorisationResponse implements BaseAuthoriseResponse {
         this.authoriseStatus = authoriseStatus;
         this.redirectUrl = redirectUrl;
         this.stringifiedResponse = stringifiedResponse;
+        this.cardExpiryDate = cardExpiryDate;
     }
 
     public static StripeAuthorisationResponse of(StripePaymentIntentResponse stripePaymentIntent) {
@@ -48,8 +58,9 @@ public class StripeAuthorisationResponse implements BaseAuthoriseResponse {
                 stripePaymentIntent.getRedirectUrl().orElse(null),
                 stripePaymentIntent.getStringifiedOutcome(),
                 stripePaymentIntent.getCustomerId(),
-                stripePaymentIntent.getPaymentMethodId()
-                );
+                stripePaymentIntent.getPaymentMethod().getId(),
+                extractCardExpiryDate(stripePaymentIntent)
+        );
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
```diff
@@ -90,4 +101,21 @@ public Optional<Map<String, String>> getGatewayRecurringAuthToken() {
                 entry(STRIPE_RECURRING_AUTH_TOKEN_PAYMENT_METHOD_ID_KEY, paymentMethodId)
         ));
     }
+
+    @Override
+    public Optional<CardExpiryDate> getCardExpiryDate() {
+        return Optional.of(cardExpiryDate);
+    }
+
+    private static CardExpiryDate extractCardExpiryDate(StripePaymentIntentResponse stripePaymentIntent) {
```
alexbishop1: Maybe this should return an `Optional<CardExpiryDate>` and then do `extractCardExpiryDate(stripePaymentIntent).orElse(null)` when it’s used in the earlier line?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
```diff
@@ -90,4 +101,21 @@ public Optional<Map<String, String>> getGatewayRecurringAuthToken() {
...
+        return stripePaymentIntent.getPaymentMethod().getCard().map(card -> {
+            try {
+                YearMonth yearMonth = YearMonth.of(card.getCardExpiryYear(), card.getCardExpiryMonth());
```
alexbishop1, line 117: I assume we’re sure that `card.getCardExpiryYear()` will always return a four-digit year and `card.getCardExpiryMonth()` will always return a 1-indexed month?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
```diff
@@ -90,4 +101,21 @@ public Optional<Map<String, String>> getGatewayRecurringAuthToken() {
...
+                return CardExpiryDate.valueOf(yearMonth);
```
alexbishop1, line 118: This will throw an `IllegalArgumentException` if `yearMonth` is before 2000-01 or after 2099-12… worth handling?

src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
```diff
@@ -171,4 +173,9 @@ protected Map<String, String> params() {
     protected String idempotencyKeyType() {
         return "payment_intent";
     }
+
+    @Override
+    protected List<String> expansionFields() {
+        return Collections.singletonList("payment_method.card");
```
alexbishop1: `List.of(…)` is cooler 😎

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeCard.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.pay.connector.gateway.stripe.response;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class StripeCard {
+    
+    @JsonProperty("exp_month")
+    private Integer cardExpiryMonth;
```
alexbishop1, line 10: Is it `Integer` (as opposed to `int`) because this could be null? If so, do we need to consider the possibility of a `NullPointerException` when it’s fed into `YearMonth.of(…)`?

alexbishop1 (commented):
This is likely fine but I’ve got a couple of things to check.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
stephencdaly: Can Sentry just get that

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
stephencdaly: Yeh I'm going to trust the Stripe API docs on this one..

src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripeCard.java
stephencdaly: I have added in some handling for these being null.

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
stephencdaly: Yeh good idea, changed

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
alexbishop1: The equivalent Worldpay code in `WorldpayOrderStatusResponse` checks the range (because it gets the year as a string and checks it matches the regex `20[0-9]{2}`). Do we know if Stripe is doing any checks on the year before sending it back to us or is there a risk it’s just echoing back what the wallet gave us (which means it’s essentially user input)? I think I’d feel happier if we did catch this.

(Slight aside: wondering if `CardExpiryDate` should have a static `isValid(YearMonth)` method to encapsulate the valid range in one place and avoid it leaking all over the code — one of the problems `CardExpiryDate` was supposed to fix in the first place.)

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
```diff
@@ -90,4 +101,25 @@ public Optional<Map<String, String>> getGatewayRecurringAuthToken() {
...
+        return Optional.of(cardExpiryDate);
+    }
+
+    private static Optional<CardExpiryDate> extractCardExpiryDate(StripePaymentIntentResponse stripePaymentIntent) {
+        return stripePaymentIntent.getPaymentMethod().getCard().map(card -> {
+            try {
+                if (card.getCardExpiryYear() == null || card.getCardExpiryMonth() == null) {
+                    LOGGER.info("Missing card expiry date on payment method");
+                    return null;
+                }
```
alexbishop1: Does this need to be inside the try/catch?

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
alexbishop1: Would it be hideous amounts of work to make `card.getCardExpiryYear()` return `Optional<Year>` and `card.getCardExpiryMonth()` return `Optional<Month>`?

alexbishop1 (changes requested):
In addition to the comments with this review, there are also some separate replies to your responses to comments from my previous review (sorry, didn’t quite think it through).

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
stephencdaly: Done

src/main/java/uk/gov/pay/connector/gateway/stripe/StripeAuthorisationResponse.java
stephencdaly: Moved

alexbishop1 (approved):

PR #4717, opened 2023-09-19
PP-11385 Authorise Stripe Apple Pay payment

stephencdaly:
Authorise Apple Pay payment by:

- Making a POST request to /v1/tokens on Stripe with the data from Apple Pay to create a token.
- Making a POST request to /v1/payment_intents to create a Payment intent and authorise the payment by providing the token ID.
    
When an authorisation is rejected, we should be able to extract the card expiry date from the create payment intent response. This will be done in a follow-up PR.


james-peacock-ho (approved):
Discussed with Stephen & looks good to me.

PR #4719, opened 2023-09-20
PP-11393 Stripe requests for Google Pay

james-peacock-ho:
- Create specific Google Pay auth request classes for Stripe and Worldpay as the structure is different
- Use token provided by Google Pay via frontend to build the request to Stripe
- Add appropriate unit tests and remove redundant test

The corresponding “Authorise Google Pay via Stripe” endpoint will be added in a subsequent PR


stephencdaly (approved):

PR #4722, opened 2023-09-20
PP-11393 Implement expandable fields for Stripe responses

stephencdaly:
When making a request to Stripe, it is possible to specify fields to be "expanded" in the response. If these fields are not expanded, Stripe will return the ID for the resource the field refers to. If they are expanded, the resource will be embedded in the response.

Implement a method to handle the response containing either the expanded object or the ID by first trying to deserialize the response in the getter for the expandable field, and if this throws an exception assuming the value is an ID and returning this.

For now, just use this for getting the payment method from a payment intent response.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripePaymentIntentResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripePaymentIntentResponse.java
@@ -2,8 +2,12 @@ package uk.gov.pay.connector.gateway.stripe.response;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
+import uk.gov.pay.connector.gateway.stripe.json.StripeExpandableField;
 import uk.gov.pay.connector.gateway.stripe.json.StripeCharge;
+import uk.gov.pay.connector.gateway.stripe.json.StripeExpandableFieldDeserializer;
 import uk.gov.pay.connector.gateway.stripe.json.StripePaymentIntent;
 import uk.gov.pay.connector.gateway.stripe.model.StripeChargeStatus;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/response/StripePaymentIntentResponse.java
```diff
@@ -34,7 +38,7 @@ public class StripePaymentIntentResponse {
     private String customerId;
     
     @JsonProperty("payment_method")
-    private StripePaymentMethodResponse paymentMethod;
+    private StripeExpandableField<StripePaymentMethodResponse> paymentMethod;
```
kbottla, line 41: Can we not expand the payment method for all requests so we don't need to have the custom logic? Or is it the Stripe behaviour that the payment method is not expanded all the time?

stephencdaly: Yeh to be honest we could do that...
I was just doing this to handle the fact that we weren't expanding when retrieving payment intents, but we now are when creating a payment intent. If we don't see a use for this logic, I could just discard it and make sure we always expand.

I guess the only thing that we need to be careful of when expanding is that we mock out Stripe in all tests apart from smoke test - so we have to be very careful that we aren't expecting something to be expanded when it isn't and vice versa as out tests wouldn't catch that. Although that problem still might exist even if we implement this logic.

kbottla: `paymentMethodId` on payment intent is not required for one-off payments and looks like a recent addition so there will probably be a few tests relying on paymentMethodId. And yeah smoke tests should catch any issues with one-off and recurring payments.

stephencdaly: Ok, probably the best way to go. Oh well, I had some fun figuring this one out 😆

kbottla (approved):
Discussed this with @stephencdaly and as StripePaymentIntent is a shared class between StripeNotificationService that processes webhooks without expanded fields (like payment method) and StripePaymentProvider functionality which requests expanded fields from Stripe, this approach of using separate Expandable field is required. 

The other option is to use SDK only for webhooks processes but might need significant effort.

PR #4725, opened 2023-09-21
PP-11393 Save card expiry date for auth rejected

stephencdaly:
When Stripe returns a response indicating the authorisation was rejected when creating the payment_intent for a wallet payment, extract the card expiry date from the response and save it on the charge.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeErrorResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeErrorResponse.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import org.apache.commons.lang3.StringUtils;
 import uk.gov.pay.connector.gateway.stripe.util.PaymentIntentStringifier;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeErrorResponse.java
```diff
@@ -22,6 +28,8 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class StripeErrorResponse {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(StripeErrorResponse.class);
```
SandorArpa: Is it used anywhere? I can't see it in the changes.

SandorArpa (changes requested):

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeErrorResponse.java
stephencdaly: Ah yes it isn't, will remove

stephencdaly: removed

SandorArpa (approved):
LOKTM

PR #4726, opened 2023-09-21
PP-11393 Move Stripe model classes

stephencdaly:
We have models that represent objects in Stripe API responses in 2 different packages.

Move all the models to the `gateway/stripe/json` package so they are in the same place, and make the naming consistent.


james-peacock-ho (approved):
looks ok to me !

PR #4728, opened 2023-09-22
PP-11563 Separate Apple and Google Pay code paths

stephencdaly:
We pretend that the Apple Pay and Google Pay requests are the same for most of the code path, which makes the code difficult to understand.

Separate out the code paths at the point of calling methods on the PaymentProvider interface, adding separate `authoriseApplePay` and `authoriseGooglePay` methods. Make it clearer in the Worldpay code that we use different XML templates for the gateway request for the different wallet types, and use different data to build them.

For now, keep the shared code path in WalletService and WalletAuthoriseService as this allows us to share some code in a fairly sensible way. This can be refactored further if there is a good way to share this code whilst making it easier to follow.


src/main/java/uk/gov/pay/connector/wallets/WalletAuthorisationRequestToAuthCardDetailsConverter.java
```diff
@@ -1,16 +1,17 @@
 package uk.gov.pay.connector.wallets;
 
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
+import uk.gov.pay.connector.wallets.model.WalletPaymentInfo;
 import uk.gov.service.payments.commons.model.CardExpiryDate;
 
 public class WalletAuthorisationRequestToAuthCardDetailsConverter {
```
alexbishop1: This should probably be renamed to `WalletPaymentInfoToAuthCardDetailsConverter`.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -53,15 +57,26 @@ public WalletAuthoriseService(PaymentProviders paymentProviders,
...
+                switch (walletAuthorisationRequest.getWalletType()) {
+                    case APPLE_PAY:
+                        operationResponse = authoriseApplePay(charge, walletAuthorisationRequest);
+                        break;
+                    case GOOGLE_PAY:
+                        operationResponse = authoriseGooglePay(charge, walletAuthorisationRequest);
+                        break;
+                    default:
+                        throw new UnsupportedOperationException(format("Wallet type %s not recognised", walletAuthorisationRequest.getWalletType()));
+                }
```
alexbishop1, line 79: We should pause for a moment here to reflect on how we could make this code nicer with switch expressions (Java 14), sealed classes (Java 17) and pattern matching for switch (Java 21).

src/main/java/uk/gov/pay/connector/wallets/googlepay/api/GooglePayAuthRequest.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.pay.connector.wallets.googlepay.api;
+
+import uk.gov.pay.connector.wallets.model.WalletPaymentInfo;
+
+public interface GooglePayAuthRequest {
```
alexbishop1: Should this extend `WalletAuthorisationRequest`?

alexbishop1 (changes requested):
This is good. I like the way we now have `ApplePayAuthorisationGatewayRequest` and `GooglePayAuthorisationGatewayRequest`. The fact the tests have not had to change much is reassuring. Just a couple of small requests.

src/main/java/uk/gov/pay/connector/wallets/googlepay/api/GooglePayAuthRequest.java
stephencdaly: Yes, I think it should. I did make it extend WalletAuthorisationRequest in some further refactoring that I then abandoned - so makes sense to update this PR to do that.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
stephencdaly: Agreed

alexbishop1 (changes requested):
After the renaming of `WalletAuthorisationDataToAuthCardDetailsConverter` to `WalletPaymentInfoToAuthCardDetailsConverter`, there’s still some local variables named `walletAuthorisationDataToAuthCardDetailsConverter` or `mockWalletAuthorisationDataToAuthCardDetailsConverter` or similar. Can those be renamed please?

alexbishop1 (approved):

PR #4732, opened 2023-09-26
PP-11470 Return 402 for all authorisation errors

stephencdaly:
Frontend expected connector to return 402 for an authorisation error, and 400 if the authorisation is rejected. For one type of error we were returning 402 rather than 400, which looks like a mistake.

Make it so that we return a 402 for all authorisation errors.


SandorArpa (approved):
LGTM

PR #4734, opened 2023-09-28
PP-11393 Fix status code for auth errors

stephencdaly:
When the Stripe API returns an error, the GatewayResponse returned by the authorise service does not have a GatewayError set, but the BaseAuthoriseResponse has an AuthoriseStatus of Error.

We were treating this the same as a decline and returning a 400 response code. PP-8693 made some changes so that authorisation errors return a 402 response rather than a 500 response. It makes sense that all authorisation errors return a 402 rather than a 400. The frontend code for wallet payments also now expects a 400 response to always be a decline.

Remove handling for AuthoriseStatus.SUBMITTED, as this was only possible for ePDQ which has now been retired.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeCardResourceAuthoriseIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeCardResourceAuthoriseIT.java
@@ -48,6 +48,7 @@ import static org.hamcrest.Matchers.hasKey;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.nullValue;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ERROR;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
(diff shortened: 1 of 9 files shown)


src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeCardResourceAuthoriseIT.java
```diff
@@ -150,6 +151,27 @@ public void cardAuthorisationWithPaymentIntentsFailureShouldReturnBadRequest() {
...
+        String externalChargeId = addCharge();
+
+        given().port(testContext.getPort())
+                .contentType(JSON)
+                .body(validAuthorisationDetails)
+                .post(authoriseChargeUrlFor(externalChargeId))
+                .then()
+                .statusCode(402)
+                .body("message", contains("There was an error authorising the transaction."))
+                .body("error_identifier", is(ErrorIdentifier.GENERIC.toString()));
```
kbottla, line 170: This one is dependent on the error_code from Stripe. Only a few 402s from Stripe result in an error https://github.com/alphagov/pay-connector/blob/28faec6fbe588727dea78b8c125ed9d25b084730/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java#L47-L53

src/test/java/uk/gov/pay/connector/it/resources/stripe/StripeCardResourceAuthoriseIT.java
```diff
@@ -150,6 +151,27 @@ public void cardAuthorisationWithPaymentIntentsFailureShouldReturnBadRequest() {
         assertFrontendChargeStatusIs(externalChargeId, AUTHORISATION_REJECTED.toString());
     }
 
+    @Test
+    public void shouldReturn402_whenStripeReturnsAnError() {
```
kbottla: Can this be explicit about the type of error Stripe returns
402 - with known error identifier - results in declined payments
402 - with unknown error code - results in auth errored payments
500 - server error

stephencdaly: So just update the test name?

kbottla: yeah and possibly the mock if it is not clear

kbottla (approved):

PR #4761, opened 2023-10-16
PP-11651:  Use Stripe SDK classes

oswaldquek:
We have generally made our own classes to deserialise JSON from Stripe. However there doesn't seem to be any gains from doing this. As we integrate more with Stripe, duplicating these classes would be unnecessary work.

This PR aims to revert [expandable fields](https://github.com/alphagov/pay-connector/pull/4722) which impacts classes StripeErrorResponse and StripePaymentIntent, amongst others. More could be deleted but we want to keep this PR as small as possible.

There's a manual [StripePaymentProviderTest](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/contract/StripePaymentProviderTest.java) integration test that I've run locally to give confidence that using the Stripe SDK classes work:

Screenshot 2023-10-19 at 10 47 22

## Notes

We are using [STRIPE_API_VERSION 2019-05-16](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/util/AuthUtil.java#L23). The corresponding stripe-java version would be [v10.15.0](https://github.com/stripe/stripe-java/blob/v10.15.0/CHANGELOG.md). We are currently using stripe-java version 21.10.0 which corresponds to [STRIPE_API_VERSION 2022-08-01](https://github.com/stripe/stripe-java/blob/v21.10.0/CHANGELOG.md#2100---2022-08-02).


oswaldquek:
Closing as we probably want to use the Stripe SDK itself to make requests, rather than our custom jersey client provider. This would bring the Stripe API version we use up to date too. This should be prioritised non-firebreak work.

PR #4762, opened 2023-10-16
PP-11567 Add method to send refund email synchronously

kbottla:
## WHAT YOU DID
- Added a method to send refund emails synchronously similar to payment emails


stephencdaly (approved):
Looks good to me!

PR #4763, opened 2023-10-16
PP-11567 Add method to find refund by externalId

kbottla:
## WHAT YOU DID
- Added service method to find refund by external_id


stephencdaly (approved):
Looks good to me!

PR #4773, opened 2023-10-17
PP-11567 Add retry payment/refund email task to the task queue

kbottla:
## WHAT YOU DID
- When a payment/refund email fails, adds a message to the task queue for retrying (doesn't send failures metric)
- Doesn't retry sending email, if the task handler attempts emails. Instead sends a failure metric and logs an error.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
@@ -12,6 +12,7 @@ import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.util.CorporateCardSurchargeCalculator;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
+import uk.gov.pay.connector.queue.tasks.TaskQueueService;
 import uk.gov.pay.connector.refund.model.domain.RefundEntity;
 import uk.gov.pay.connector.usernotification.govuknotify.NotifyClientFactory;
 import uk.gov.pay.connector.usernotification.model.domain.EmailNotificationEntity;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
```diff
@@ -120,32 +134,43 @@ private boolean shouldSendEmail(EmailNotificationType emailNotificationType, Cha
...
             return false;
-        }
-        else {
+        } else {
             return true;
         }
     }
 
-    private Optional<String> sendEmailInner(EmailNotificationType emailNotificationType, Charge charge, GatewayAccountEntity gatewayAccountEntity, HashMap<String, String> personalisation, Stopwatch responseTimeStopwatch) {
+    private Optional<String> sendEmailInner(EmailNotificationType emailNotificationType, Charge charge,
```
oswaldquek, line 145: Why is the word "inner" in this method name? I don't know what that conveys.

oswaldquek (commented):
Why do we have methods to send email asynchronously and synchronously in UserNotificationService? The calling method can always wait on a Future to make it synchronous.

kbottla:
asynchronous methods are used during the payment/refund journey and are completed in the background. 

the synchronous method has been added for the new endpoint added recently https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/charge/resource/ChargesApiResource.java#L229 so the endpoint can return 200 or non 2xx response immediately.

src/main/java/uk/gov/pay/connector/usernotification/service/UserNotificationService.java
kbottla: don't think it is adding anything. Maybe the intention was to make it clear that it is internal/private. It's been refactored recently https://github.com/alphagov/pay-connector/commit/df6972f00b3a1bb171805938bc749e1f60e06a8b#diff-ac2ffab9f16a5bb65352f6fe53148764ab2d804839ae50023e775db7950b67e9R129

oswaldquek: I think it can simply be renamed to `private Optional<String> sendEmail`. Happy for that to be done in another PR though.

oswaldquek (approved):

PR #4774, opened 2023-10-18
PP-11567 Refactor RetryPaymentOrRefundEmailTaskData

kbottla:
## WHAT YOU DID
- Refactored `RetryPaymentOrRefundEmailTaskData` class to add a failed timestamp. Failed timestamp will be used (https://github.com/alphagov/pay-connector/pull/4775) to delay retrying failed emails for the required duration (1 hour or as per config). SQS only allows a maximum delivery delay of 15 minutes so we have to use failed_timestamp to defer (add message back to SQS queue) retrying if longer than 15 mins.
- Uses the system clock in UserNotificaitonService so fixed dates can be used for testing
- Also added maximum delivery delay to config

### DIFF
--- a/src/main/java/uk/gov/pay/connector/app/SqsConfig.java
+++ b/src/main/java/uk/gov/pay/connector/app/SqsConfig.java
@@ -26,7 +26,7 @@ public class SqsConfig extends Configuration {
     private int messageMaximumBatchSize;
 
     @Max(900)
-    private int maxAllowedDeliveryDelay;
+    private int maxAllowedDeliveryDelayInSeconds;
 
     private String eventQueueUrl;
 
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/connector/app/SqsConfig.java
```diff
@@ -24,6 +24,10 @@ public class SqsConfig extends Configuration {
     private int messageMaximumWaitTimeInSeconds;
     @Max(10)
     private int messageMaximumBatchSize;
+
+    @Max(900)
+    private int maxAllowedDeliveryDelay;
```
alexbishop1: What’s the units here? If it’s seconds, could this be named `maxAllowedDeliveryDelayInSeconds`? Even better, make it a `Duration` (we do this in some other places like `ChargeSweepConfig`).

alexbishop1 (changes requested):
Looks straightforward enough and good use of `Clock`s. Just one thing about making the units of the maximum delivery delay clearer.

src/main/java/uk/gov/pay/connector/app/SqsConfig.java
kbottla: Thanks, @alexbishop1. Renamed variable to `maxAllowedDeliveryDelayInSeconds`.  We need to pass this value as is so will leave it as `int` for now and will update it in the next PR if makes a difference.

alexbishop1 (approved):

PR #4775, opened 2023-10-18
PP-11567 Delay retrying failed emails as per config

kbottla:
## WHAT YOU DID
 - Adds RetryPaymentOrRefundEmail task to SQS queue if the time elapsed since failure is less than the config.
      Message is added back to SQS queue with maximum delivery delay allowed (15 mins) for SQS queue.
- Always delays message delivery by max_delivery_delay and doesn't calculate time left which can be done separately if required.


james-peacock-ho (approved):
LGTM

PR #4788, opened 2023-10-26
PP-11641 Throw UnsupportedOperationException for unknown pp

james-peacock-ho:
Currently a RuntimeException is thrown if wallet authorisation is requested for a payment provider other than Stripe or Worldpay. This causes a 500 response when it should be 400. This was identified in frontend consumer PACT test.

- Throw UnsupportedOperationException instead; this maps to a 400 response.


stephencdaly (approved):
This should be just very temporary behaviour as we shouldn't need to do this check once we've removed the provider specific request classes. But looks fine to me!

PR #4798, opened 2023-10-31
PP-11641 Stub Payment Provider Responses for PACT

james-peacock-ho:
Context: Need to update provider states for frontend-connector PACT to correctly reflect interactions for wallet payments

- Add new provider states for Worldpay and Stripe charges
- Stub the interaction with Stripe/Worldpay in these new states

Subsequent PR will point frontend consumer tests to these new states


stephencdaly (approved):
LGTM!

PR #4799, opened 2023-10-31
PP-11279: Persistence tests

oswaldquek:
As per the acceptance criteria of this story, when a service is archived, for it's related gateway accounts, we want to verify we can:
- Disable the gateway account(s)
- Remove notification credentials from a gateway account(s)
- Redact sensitive gateway account(s) credentials information

This PR doesn't include removing rows from the `gateway_account_credentials_history` table.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
@@ -126,7 +126,7 @@ public class GatewayAccountEntity extends AbstractVersionedEntity {
     @Enumerated(EnumType.STRING)
     private EmailCollectionMode emailCollectionMode = EmailCollectionMode.MANDATORY;
 
-    @OneToOne(mappedBy = "accountEntity", cascade = CascadeType.PERSIST)
+    @OneToOne(mappedBy = "accountEntity", cascade = CascadeType.PERSIST, orphanRemoval = true)
     private NotificationCredentials notificationCredentials;
 
     @OneToOne(mappedBy = "gatewayAccountEntity", cascade = CascadeType.PERSIST)
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/WorldpayMerchantCodeCredentials.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/WorldpayMerchantCodeCredentials.java
@@ -14,6 +14,8 @@ import java.util.Objects;
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class WorldpayMerchantCodeCredentials {
 
+    public static final String DELETED = "<DELETED>";
+    
     @JsonView({GatewayCredentials.Views.Api.class})
     private String merchantCode;
     @JsonView({GatewayCredentials.Views.Api.class})
@@ -29,6 +31,11 @@ public class WorldpayMerchantCodeCredentials {
         this.merchantCode = merchantCode;
         this.username = username;
         this.password = password;
+    }
+    
+    public void redactSensitiveInformation() {
+        username = DELETED;
+        password = DELETED;
     }
 
     public String getMerchantCode() {
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialsEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialsEntity.java
@@ -180,7 +180,8 @@ public class GatewayAccountCredentialsEntity extends AbstractVersionedEntity {
     }
 
     public void setCredentials(GatewayCredentials credentials) {
-        this.credentials = objectMapper.convertValue(credentials, new TypeReference<Map<String, Object>>() {});
+        this.credentials = objectMapper.convertValue(credentials, new TypeReference<>() {
+        });
     }
 
     public void setState(GatewayAccountCredentialState state) {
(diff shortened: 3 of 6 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
```diff
@@ -126,7 +126,7 @@ public class GatewayAccountEntity extends AbstractVersionedEntity {
     @Enumerated(EnumType.STRING)
     private EmailCollectionMode emailCollectionMode = EmailCollectionMode.MANDATORY;
 
-    @OneToOne(mappedBy = "accountEntity", cascade = CascadeType.PERSIST)
+    @OneToOne(mappedBy = "accountEntity", cascade = CascadeType.PERSIST, orphanRemoval = true)
```
kbottla, line 129: What does `orphanRemoval` do here?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/WorldpayMerchantCodeCredentials.java
```diff
@@ -30,6 +32,11 @@ public WorldpayMerchantCodeCredentials(String merchantCode, String username, Str
         this.username = username;
         this.password = password;
     }
+    
+    public void redactSensitiveInformation() {
```
kbottla, line 36: Would it help to have a generic method to redact info from credentials i.e., look for username & password anywhere in JSON and redact them? Otherwise, we might have to have a specific code for each payment provider and type of credentials

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialsEntity.java
```diff
@@ -180,7 +180,8 @@ public void setCredentials(Map<String, Object> credentials) {
     }
 
     public void setCredentials(GatewayCredentials credentials) {
-        this.credentials = objectMapper.convertValue(credentials, new TypeReference<Map<String, Object>>() {});
+        this.credentials = objectMapper.convertValue(credentials, new TypeReference<>() {
```
kbottla, line 183: Any reason type `Map<String, Object>` is removed?

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/dao/GatewayAccountCredentialsDaoIT.java
```diff
@@ -171,17 +186,31 @@ public void shouldPersistHistory() {
         assertThat(historyRows.get(0).get("state"), is("ACTIVE"));
         assertThat(historyRows.get(0).get("history_start_date"), not(nullValue()));
         assertThat(historyRows.get(0).get("history_end_date"), is(nullValue()));
+        assertEquals(objectMapper.readValue(historyRows.get(0).get("credentials").toString(), Map.class), credentials);
         
         gatewayAccountCredentialsEntity.setState(RETIRED);
+        WorldpayCredentials worldpayCredentials = (WorldpayCredentials) gatewayAccountCredentialsEntity.getCredentialsObject();
+        worldpayCredentials.getRecurringCustomerInitiatedCredentials().ifPresent(WorldpayMerchantCodeCredentials::redactSensitiveInformation);
```
kbottla, line 193: Is this testing `redactSensitiveInformation`?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountEntity.java
oswaldquek: If `orphanRemoval` is not present, this [test](https://github.com/alphagov/pay-connector/pull/4799/files#diff-281a56ddd7ca079374b8b95fe57ec44d924c23d3cc1f32841a666473d1c97e50R102) fails. If it is not there the notification credential will still be present even if we set it to null in the gateway account, i.e. `gatewayAccount.setNotificationCredentials(null);`

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/model/GatewayAccountCredentialsEntity.java
oswaldquek: It's because Intellij said I could as the type can be inferred. Admittedly it's got nothing to do with the acceptance criteria of this story!

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/dao/GatewayAccountCredentialsDaoIT.java
oswaldquek: Not sure what you mean here but I amended/enhanced the `shouldPersistHistory` test to include adding a row to the gateway account creds history table when the creds are amended. Calling `redactSensitiveInformation` here is a means to updating the creds.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/WorldpayMerchantCodeCredentials.java
oswaldquek: I think the generic would help but the need for it will become clearer when I add tests for redacting epdq credentials. I plan to have another PR for redacting from epdq so I think the generic method should go there (in the future) if that's ok. Let's leave this current PR for Worldpay just now.

kbottla (approved):

src/main/java/uk/gov/pay/connector/gatewayaccount/model/WorldpayMerchantCodeCredentials.java
oswaldquek: I've been thinking about this and I think looking for credentials anywhere in JSON might be inefficient as technically JSON can have an unlimited nested tree of objects. It will be easier to look for creds based on on our strongly typed [WorldpayCredentials](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/model/WorldpayCredentials.java) and [EpdqCredentials](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/model/EpdqCredentials.java) classes.

PR #4809, opened 2023-11-02
PP-11641 Remove old Google Pay endpoints

james-peacock-ho:
Context: Worldpay- and Stripe-specific endpoints have been replaced with a single Google Pay authorisation endpoint. Tests have already been replicated for the new endpoint. Frontend is now pointing to the new endpoint.

- Remove the google/worldpay and google/stripe endpoints
- Remove redundant tests


stephencdaly (commented):
Not sure whether this will be part of another PR, but do you think we should remove StripeGooglePayAuthRequest and WorldpayGooglePayAuthRequest as this would probably simplify the code somewhat?

stephencdaly (approved):
LGTM!

PR #4811, opened 2023-11-02
PP-11641 Refactor Google Pay Auth Requests

james-peacock-ho:
- Remove the payment provider-specific Google Pay auth request classes
- Make GooglePayAuthRequest a concrete class for use in all Google Pay auth requests.
- Remove redundant method to convert to payment-provider specific class
- Throw ValidationException if fields required by Payment Provider are missing


src/main/java/uk/gov/pay/connector/wallets/WalletService.java
```diff
@@ -43,13 +42,12 @@ public Response authorise(String chargeId, WalletAuthorisationRequest walletAuth
 
     }
 
-    public Response convertAndAuthorise(String chargeId, GenericGooglePayAuthRequest genericGooglePayAuthRequest) {
+    public Response convertAndAuthorise(String chargeId, GooglePayAuthRequest googlePayAuthRequest) {
```
stephencdaly: We can remove this method now and just call authorise directly

src/main/resources/templates/worldpay/WorldpayAuthoriseGooglePayOrderTemplate.xml
```diff
@@ -8,9 +8,11 @@
             <amount currencyCode="GBP" exponent="2" value="${amount}"/>
             <paymentDetails>
                 <PAYWITHGOOGLE-SSL>
-                    <protocolVersion>${googlePayPaymentData.encryptedPaymentData.protocolVersion?xml}</protocolVersion>
-                    <signature>${googlePayPaymentData.encryptedPaymentData.signature?xml}</signature>
-                    <signedMessage>${googlePayPaymentData.encryptedPaymentData.signedMessage?xml}</signedMessage>
+                    <#if googlePayPaymentData.encryptedPaymentData.isPresent()>
```
stephencdaly: What happens if we didn't do this `isPresent` check and it wasn't there - would it throw an exception that's caught and mapped appropriately in the code?
If so, that might be preferable to doing the `isPresent` check as this makes it look like it's optional to include this in the request. If not, then we'll just have to live with it.

james-peacock-ho: The template failures throw `java.lang.RuntimeException: Could not render template worldpay/WorldpayAuthoriseGooglePayOrderTemplate.xml` which would map to a 500 error.

On closer inspection it seems that regardless of whether the `isPresent` check is there, if `encryptedPaymentData` is null, the template will fail to build.

However, I don't see how `GooglePayAuthRequest` could be constructed without `encryptedPaymentData` and follow this path in the code because an empty `encryptedPaymentData` will cause a ValidationException to be thrown in `WorldpayWalletAuthorisationHandler`.

On that basis, I think we should remove the `isPresent` check as it serves no purpose.

stephencdaly (approved):
LGTM! Probably wise to test Google Pay for at least Worldpay on the test environment due to the request template changes.

PR #4816, opened 2023-11-03
PP-11711 upgrade notify client 3.19.1 -> 4.1.0

nlsteers:
## WHAT YOU DID

Upgrade notify client 3.19.1 -> 4.1.0

Noteworthy changes in this version:

- Joda-Time has been removed and replaced with java.time
  - The client now uses [ZonedDateTime instead of DateTime](https://blog.joda.org/2014/11/converting-from-joda-time-to-javatime.html)


SandorArpa (approved):
LOKTM

PR #4828, opened 2023-11-09
PP-11279: Handle a "service archived" task

oswaldquek:
Handle a "service archived" task. Also handle logging and metrics.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
@@ -32,6 +32,7 @@ import java.util.Optional;
 import java.util.function.BiConsumer;
 import java.util.stream.Collectors;
 
+import static java.lang.String.format;
 import static java.util.Map.entry;
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
```diff
@@ -259,7 +260,12 @@ private void throwIfGatewayAccountIsNotWorldpay(GatewayAccountEntity gatewayAcco
 
     @Transactional
     public void disableAccountsAndRedactOrDeleteCredentials(String serviceId) {
-        gatewayAccountDao.findByServiceId(serviceId).forEach(ga -> {
+        List<GatewayAccountEntity> gatewayAccounts = gatewayAccountDao.findByServiceId(serviceId);
+        
+        LOGGER.info(format("Disabling gateway accounts %s for service.", gatewayAccounts.stream().map(GatewayAccountEntity::getExternalId).collect(Collectors.joining(","))),
```
kbottla, line 265: Can gateway account IDs be logged as kv pair? kv("gateway_account_ids", list)

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/ServiceArchivedTaskHandler.java
```diff
@@ -0,0 +1,32 @@
...
+import io.prometheus.client.Histogram;
+import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
+import uk.gov.pay.connector.queue.tasks.model.ServiceArchivedTaskData;
+
+import javax.inject.Inject;
+
+public class ServiceArchivedTaskHandler {
+    private final GatewayAccountService gatewayAccountService;
+
+    private static final Histogram duration = Histogram.build()
```
kbottla: Do we need metrics for this given it is only dealing with one service at a time?

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
oswaldquek: They could be, but I'm not sure how useful it would be. Json fields are useful to find out, for example, the path of a http request (based on request id), or payment (based on payment id) in splunk. It would be extremely rare to search based on a list of gateway account ids.

src/main/java/uk/gov/pay/connector/queue/tasks/handlers/ServiceArchivedTaskHandler.java
oswaldquek: Yeah I think you're right here. I'll remove this :)

kbottla (approved):

PR #4846, opened 2023-11-20
PP-11791-add-new-CA-bundles

rfisher-gds:
## WHAT YOU DID

Add to the Docker images the current RDS CA bundle for eu-west-1 and eu-central-1.

I had to bump the base container version as the old one is no longer visible.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM eclipse-temurin:11-jre-alpine@sha256:c033de8d3135b954d230eed6b9bd14dc3dbf32550e448705222f1c907ceebe08
+FROM eclipse-temurin:11.0.21_9-jre-alpine@sha256:1017b1b26f9b92921a93a730fa33327d83cdfa0cc2f7fa486b1348ad4eda755c
```
jfharden: You need to leave the version number as just 11-jre-alpine otherwise dependabot can't keep the image up to date (see https://github.com/alphagov/pay-architecture/blob/cea7f1df695bc1cc74360eb6a3362cf842f49399/adr/024-java-base-image-pinning-strategy.md?plain=1#L15 for a detailed set of reasoning)

jfharden (approved):

PR #4857, opened 2023-11-24
PP-11591 Do not set 3ds flex in connector

james-peacock-ho:
Context: 3DS integration version is already set to 2 in self-service when flex credentials are entered

- Don't set 3DS integration version in connector when creating gateway accounts or switching to Worldpay.


stephencdaly (approved):
Changes look good - the commit message could probably be made a bit more accurate as there might be some confusion between setting 3DS flex credentials and setting the `integration_version_3ds` which enables flex for Worldpay.

stephencdaly (approved):
New commit message looks good

PR #4893, opened 2023-12-15
PP-11613: Return 404 if trying to refund an EPDQ payment

oswaldquek:
The 404 returned  here by connector will be [handled by publicapi](https://github.com/alphagov/pay-publicapi/blob/master/src/main/java/uk/gov/pay/api/service/CreateRefundService.java#L57-L63) which will return a [P0603 API error
message](https://github.com/alphagov/pay-publicapi/blob/master/src/main/java/uk/gov/pay/api/exception/mapper/CreateRefundExceptionMapper.java#L35-L37) which fulfills the acceptance criteria of this story.

Also for some java fun, although there aren't enough payment gateways to make any noticeable difference, I amended the `valueFrom` method to use a `Set.contains` (which has O(1) time complexity) rather than iterating through the values (O(n) time complexity).

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/PaymentGatewayName.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/PaymentGatewayName.java
@@ -1,13 +1,17 @@
 package uk.gov.pay.connector.gateway;
 
-import java.util.List;
+import java.util.EnumSet;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 public enum PaymentGatewayName {
     SANDBOX("sandbox"), SMARTPAY("smartpay"), WORLDPAY("worldpay"), EPDQ("epdq"), STRIPE("stripe");
 
     private final String gatewayName;
     
-    private static List<String> unsupported = List.of(SMARTPAY.gatewayName, EPDQ.gatewayName);
+    private static final Set<String> UNSUPPORTED = Set.of(SMARTPAY.gatewayName, EPDQ.gatewayName);
+
+    private static final Set<String> PAYMENT_GATEWAY_NAMES = EnumSet.allOf(PaymentGatewayName.class).stream().map(x -> x.gatewayName).collect(Collectors.toSet());
 
     PaymentGatewayName(String gatewayName) {
         this.gatewayName = gatewayName;
@@ -18,7 +22,7 @@ public enum PaymentGatewayName {
     }
     
     public static boolean isUnsupported(String gatewayName) {
-        return unsupported.contains(gatewayName);
+        return UNSUPPORTED.contains(gatewayName);
     }
 
     public static class Unsupported extends RuntimeException {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/PaymentGatewayName.java
```diff
@@ -1,9 +1,13 @@
 package uk.gov.pay.connector.gateway;
 
+import java.util.List;
+
 public enum PaymentGatewayName {
     SANDBOX("sandbox"), SMARTPAY("smartpay"), WORLDPAY("worldpay"), EPDQ("epdq"), STRIPE("stripe");
 
     private final String gatewayName;
+    
+    private static List<String> unsupported = List.of(SMARTPAY.gatewayName, EPDQ.gatewayName);
```
alexbishop1: This could be a `Set` (maybe even an `EnumSet`) and should probably be `final` with a `SCREAMING_SNAKE_CASE` name.

alexbishop1 (changes requested):
Looks fine apart from a stylistic issue.

src/main/java/uk/gov/pay/connector/gateway/PaymentGatewayName.java
oswaldquek: Fixed. I also made a fun java change (see updated PR description).

src/main/java/uk/gov/pay/connector/gateway/PaymentGatewayName.java
```diff
@@ -1,9 +1,17 @@
...
+import java.util.stream.Collectors;
+
 public enum PaymentGatewayName {
     SANDBOX("sandbox"), SMARTPAY("smartpay"), WORLDPAY("worldpay"), EPDQ("epdq"), STRIPE("stripe");
 
     private final String gatewayName;
+    
+    private static final Set<String> UNSUPPORTED = Set.of(SMARTPAY.gatewayName, EPDQ.gatewayName);
+
+    private static final Set<String> PAYMENT_GATEWAY_NAMES = EnumSet.allOf(PaymentGatewayName.class).stream().map(x -> x.gatewayName).collect(Collectors.toSet());
```
alexbishop1, line 14: I’m being picky but `Collectors.toSet()` should probably be `Collectors.toUnmodifiableSet()`.

I suppose another way of doing this would be to have a `Map<String, PaymentGatewayName>`, with the key being the canonical (lower-case) name, which would avoid the need to do the upper-case conversion when looking up…

alexbishop1 (changes requested):
Nice optimisations but we should use an unmodifiable set and it would be nice to avoid the upper-casing…

src/main/java/uk/gov/pay/connector/gateway/PaymentGatewayName.java
oswaldquek: You are right but are you really gonna make me wait for another unit/integration/e2e build???

oswaldquek: It's a private field so the outside world shouldn't be able to access it so I don't think it really matters actually.

alexbishop1: OK, you win.

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/gateway/PaymentGatewayName.java
oswaldquek: Thanks 🙂

PR #4894, opened 2023-12-15
PP-11613: Make refund status for EPDQ always be unavailable

oswaldquek:
This is because we don't support EPDQ anymore.


james-peacock-ho (approved):
LGTM!

PR #4919, opened 2024-01-05
  PP-11554 make request to stripe to check refund status

AbdirizakIdris:
## WHAT YOU DID
 - Add service method to check status of refund from Stripe

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
+++ b/src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
@@ -1,24 +1,54 @@
 package uk.gov.pay.connector.refund.service;
+
+import com.stripe.exception.StripeException;
 import uk.gov.pay.connector.client.ledger.service.LedgerService;
+import uk.gov.pay.connector.gateway.stripe.StripeSdkClient;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.refund.dao.RefundDao;
 import uk.gov.pay.connector.refund.model.domain.Refund;
 
 import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
 import java.util.Optional;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.util.ResponseUtil.badRequestResponse;
 
 public class RefundReversalService {
     private final LedgerService ledgerService;
     private final RefundDao refundDao;
+    private final StripeSdkClient stripeClient;
 
     @Inject
-    public RefundReversalService(LedgerService ledgerService, RefundDao refundDao) {
-            this.ledgerService = ledgerService;
-            this.refundDao = refundDao;
+    public RefundReversalService(LedgerService ledgerService, RefundDao refundDao, StripeSdkClient stripeClient) {
+        this.ledgerService = ledgerService;
+        this.refundDao = refundDao;
+        this.stripeClient = stripeClient;
     }
 
     public Optional<Refund> findMaybeHistoricRefundByRefundId(String refundExternalId) {
-           return refundDao.findByExternalId(refundExternalId)
-                   .map(Refund::from)
-                   .or(() -> ledgerService.getTransaction(refundExternalId).map(Refund::from));
+        return refundDao.findByExternalId(refundExternalId)
+                .map(Refund::from)
+                .or(() -> ledgerService.getTransaction(refundExternalId).map(Refund::from));
+    }
+
+    public void reverseFailedRefund(GatewayAccountEntity gatewayAccount, Refund refund) {
+        String stripeRefundId = refund.getGatewayTransactionId();
+        boolean isLiveGatewayAccount = gatewayAccount.isLive();
+        String refundExternalId = refund.getExternalId();
+
+        try {
+            com.stripe.model.Refund refundFromStripe = stripeClient.getRefund(stripeRefundId, isLiveGatewayAccount);
+            String refundStatus = refundFromStripe.getStatus();
+
+            if ("failed".equals(refundStatus)) {
+                // TODO: 04/01/2024  (PP-11555)
+            } else {
+                throw new WebApplicationException(badRequestResponse(
+                        format("Refund with Refund ID: %s and Stripe ID: %s is not in a failed state", refundExternalId, stripeRefundId)));
+            }
+        } catch (StripeException e) {
+            throw new WebApplicationException(format("Unexpected error trying to get refund with ID:%s from Stripe", refundExternalId));
+        }
     }
 }
--- a/src/test/java/uk/gov/pay/connector/service/RefundReversalServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/service/RefundReversalServiceTest.java
@@ -1,24 +1,38 @@
 package uk.gov.pay.connector.service;
 
+import com.stripe.exception.StripeException;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -1,24 +1,50 @@
...
     }
+
+    public void reverseFailedRefund(GatewayAccountEntity gatewayAccount, Refund refund) throws StripeException {
+        String stripeRefundId = refund.getGatewayTransactionId();
+        boolean isLiveGatewayAccount = gatewayAccount.isLive();
+
+        com.stripe.model.Refund refundFromStripe = stripeClient.getRefund(stripeRefundId, isLiveGatewayAccount);
+        String refundStatus = refundFromStripe.getStatus();
+        
+        if ("fail".equals(refundStatus)) {
```
stephencdaly: It doesn't look like this is a possible status for a Stripe refund. Statuses can be found here https://stripe.com/docs/api/refunds/object#refund_object-status

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -1,24 +1,50 @@
...
+            // TODO: 04/01/2024  (PP-11555)
+        } else if ("success".equals(refundStatus)) {
+            throw new WebApplicationException(format("Refund with refund id: %s is not in a failed state", stripeRefundId));
```
stephencdaly: From the story, it looks like this error should be returned if the refund is in any status other than "failed". So I think this "else if" should be removed, and this error should just be thrown from an "else" case.

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -1,24 +1,50 @@
...
+            this.stripeClient = stripeClient;
     }
 
     public Optional<Refund> findMaybeHistoricRefundByRefundId(String refundExternalId) {
            return refundDao.findByExternalId(refundExternalId)
                    .map(Refund::from)
                    .or(() -> ledgerService.getTransaction(refundExternalId).map(Refund::from));
     }
+
+    public void reverseFailedRefund(GatewayAccountEntity gatewayAccount, Refund refund) throws StripeException {
```
stephencdaly: The StripeException indicates that there was an error getting the refund from Stripe. This is "Scenario 3" in the JIRA ticket. I think this method should catch the StripeException and throw a WebApplicationException with the error message from scenario 3.

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
stephencdaly: You'll also need to make sure that the WebApplicationException thrown will cause an error response with a 400 status code to be returned from the API request. Look at other instances where we throw WebApplicationExceptions to see how we achieve this. I think currently this exception would be mapped to a 500 error.

Here is an example of where we throw a WebApplicationException that will explicitly map to a 500 response https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/util/JsonObjectMapper.java#L28

You'll need to do something similar but make it map to a 400 response.

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -1,24 +1,50 @@
 package uk.gov.pay.connector.refund.service;
+import com.stripe.exception.StripeException;
 import uk.gov.pay.connector.client.ledger.service.LedgerService;
+import uk.gov.pay.connector.gateway.stripe.StripeSdkClient;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.refund.dao.RefundDao;
 import uk.gov.pay.connector.refund.model.domain.Refund;
+import uk.gov.pay.connector.refund.model.domain.RefundEntity;
```
stephencdaly: I think there are some unused imports. Use the ⌃⌥O keyboard shortcut in intellJ to optimize the imports.

stephencdaly (changes requested):
Overall, it's doing the right thing but some things don't quite match up to the acceptance criteria for the story.

src/main/java/uk/gov/pay/connector/util/ResponseUtil.java
```diff
@@ -78,6 +78,10 @@ public static Response serviceErrorResponse(String message) {
         LOGGER.info(message);
         return buildErrorResponse(INTERNAL_SERVER_ERROR, message);
     }
+    public static Response refundErrorResponse(String message) {
```
stephencdaly: We can just use the existing `badRequestResponse` method as this doesn't do anything different.

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
stephencdaly: We can remove `throws StripeException` from the method signature now, as the exception is caught and handled within the method.

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -1,24 +1,53 @@
...
+
+        try {
+            com.stripe.model.Refund refundFromStripe = stripeClient.getRefund(stripeRefundId, isLiveGatewayAccount);
+            String refundStatus = refundFromStripe.getStatus();
+
+            if ("failed".equals(refundStatus)) {
+                // TODO: 04/01/2024  (PP-11555)
+            } else {
+                throw new WebApplicationException(badRequestResponse(
+                        format("Refund with refund id: %s is not in a failed state", stripeRefundId)));
```
stephencdaly: The error message includes the Stripe ID. The story says we should include Pay's refund external ID. If we are including the Stripe ID, we should make it clear this is the Stripe ID. Maybe we could include both?

E.g.

"Refund with ID abc123asdar3q221q and Stripe ID ref_234kjhgqkead23 is not in a failed state"

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -1,24 +1,53 @@
...
+            }
+        } catch (StripeException e) {
+            throw new WebApplicationException("Unexpected error trying to get refund with id: %s from Stripe");
```
stephencdaly: You need to template the ID into the error message.

src/test/java/uk/gov/pay/connector/service/RefundReversalServiceTest.java
```diff
@@ -75,4 +89,65 @@ void shouldNotFindRefund() {
...
+
+        String stripeRefundId = refund.getGatewayTransactionId();
+        boolean isLiveGatewayAccount = gatewayAccountEntity.isLive();
+        StripeException mockStripeException = Mockito.mock(StripeException.class);
+
+        when(mockStripeSDKClient.getRefund(stripeRefundId, isLiveGatewayAccount))
+                .thenThrow(mockStripeException);
+        
+        assertThrows(WebApplicationException.class, () -> refundReversalService.reverseFailedRefund(gatewayAccountEntity, refund),
+                "Unexpected error trying to get refund with id: a-transaction-id");
```
stephencdaly: It doesn't look like this is actually checking the error message for the exception, as this should have failed. `assertThrows` will return the thrown exception, so you can assign it to a variable and then do an assertion on the error message. Look for other examples where we use `assertThrows` to see how this is done.

stephencdaly (approved):
LGTM!

PR #4937, opened 2024-01-22
PP-11878 Improve StateTransition Failure Logging

james-peacock-ho:
- Provide specific detail in logs for EventCreationExceptions caused in StateTransition event creation.


james-peacock-ho:
starting again with a fresh branch due to an odd error in concourse

PR #4938, opened 2024-01-22
PP-11878 Improve state transition failure logging

james-peacock-ho:
- Log specific cause of EventCreationExceptions in state transitions
- Add unit tests


stephencdaly (approved):
LGTM!

PR #4951, opened 2024-01-31
PP-11861 enable applepay for sandbox accounts

SandorArpa:
## WHAT YOU DID

Enable Apple Pay by default when a new gateway account is created and the payment provider is sandbox.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
@@ -35,6 +35,7 @@ import java.util.stream.Collectors;
 import static java.lang.String.format;
 import static java.util.Map.entry;
 import static net.logstash.logback.argument.StructuredArguments.kv;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.SANDBOX;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
```diff
@@ -117,6 +118,10 @@ public CreateGatewayAccountResponse createGatewayAccount(GatewayAccountRequest g
 
         gatewayAccountEntity.setCardTypes(cardTypeDao.findAllNon3ds());
 
+        if (gatewayAccountRequest.getPaymentProvider().equals(SANDBOX.getName())) {
```
alexbishop1: Probably safer to do `SANDBOX.getName().equals(gatewayAccountRequest.getPaymentProvider())`.

alexbishop1 (changes requested):
One small thing…

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
SandorArpa: My bad, I have done that change yesterday, but forgot to push it 🤦

alexbishop1 (approved):

PR #4979, opened 2024-02-16
bau: StateTransitionQueue javadoc

oswaldquek:
Javadoc to clarify the existence of this class and why state transition events are treated differently from other events.


stephencdaly (approved):
Hopefully should make it easier to understand the purpose of this.

PR #5016, opened 2024-03-07
PP-12349 Calculate amounts for DISPUTE_LOST event

stephencdaly:
Handle the case where a dispute lost Stripe event has multiple balance_transactions associated with it for calculating the net amount and fee fields to send in the DISPUTE_LOST event.

Move calculating the total net amount and fee for a dispute lost  event to a new StripeDisputeCalculator, which is called from  StripeWebhookTaskHandler in order to calculate the total net amount once to be used for both transferring the money in Stripe and creating the DISPUTE_LOST event to send to ledger.


src/test/java/uk/gov/pay/connector/events/model/dispute/DisputeLostTest.java
james-peacock-ho, line 1: unused imports

james-peacock-ho (commented):
looks great, there's a couple of unused imports where you removed a test.

james-peacock-ho (approved):
LGTM!

PR #5019, opened 2024-03-08
PP-12242 Fix search by worldpay merchant id

james-peacock-ho:
Context: Searching for a gateway account by Worldpay merchant id doesn't work in toolbox since `merchant_id` was moved inside a nested object in the `credentials` field of the `gateway_account_credentials` table.

- Update GatewayAccountSearchParams to check for the search term within the `one_off_customer_initiated` object in the `credentials` field.
- Add test for this scenario.

Subsequent PRs will check for `merchant_id` within the `recurring_customer_initiated` and `recurring_agent_initiated` objects as well.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
@@ -376,6 +376,43 @@ public class GatewayAccountResourceIT extends NewGatewayAccountResourceTestBase 
     }
 
     @Test
+    public void shouldReturnAccountInformationWhenSearchingByWorldpayMerchantId() {
+        long accountId = RandomUtils.nextInt();
+        AddGatewayAccountCredentialsParams credentialsParams = anAddGatewayAccountCredentialsParams()
+                .withPaymentProvider(WORLDPAY.getName())
+                .withGatewayAccountId(accountId)
+                .withState(ACTIVE)
+                .withCredentials(Map.of(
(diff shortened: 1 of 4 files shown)


oswaldquek:
Would it give us confidence the search works by adding an integration test to search for a gateway account by merchant id? We're testing the [search with no params](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java#L352) and I think it'd be beneficial to test this at the integration test level too.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -375,6 +375,42 @@ public void shouldGetAllGatewayAccountsWhenSearchWithNoParams() {
                 .body("accounts[1].worldpay_3ds_flex", not(hasKey("jwt_mac_key")));
     }
 
+    @Test
+    public void shouldReturnAccountInformationWhenSearchingByWorldpayMerchantId() {
```
oswaldquek, line 379: I don't want to block the PR because of the build time, but in the next PR if you could rename this to "should....WorldpayMerchantCode", that'd be fab, to keep namings consistent.

oswaldquek (approved):

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
james-peacock-ho: good point, thanks

PR #5033, opened 2024-03-15
PP-12441 interim ignore failing tests

nlsteers:
## WHAT YOU DID

- ignore failing tests whilst we diagnose the cause of the failure, we're confident that the reduced coverage won't result in destabilisation 
- the problem has probably existed for a while


nlsteers:
closing, duplicate of https://github.com/alphagov/pay-connector/pull/5030

PR #5077, opened 2024-04-03
PP-12302 fix flaky test

SandorArpa:
## WHAT YOU DID

GatewayAccountDaoIT.persist_shouldCreateAnAccount and findById_shouldUpdateAccountCardTypes tests are flaky due to the List<Map> order is not guaranteed.

- sort list before asserting test.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoIT.java
@@ -15,6 +15,7 @@ import uk.gov.pay.connector.usernotification.model.domain.NotificationCredential
 import uk.gov.pay.connector.util.AddGatewayAccountCredentialsParams;
 
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -25,7 +26,6 @@ import static org.hamcrest.CoreMatchers.allOf;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.contains;
 import static org.hamcrest.Matchers.containsInAnyOrder;
-import static org.hamcrest.Matchers.hasEntry;
 import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.Matchers.is;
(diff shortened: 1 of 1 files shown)


oswaldquek:
Interesting. Shouldn't the `containsInAnyOrder` work?

SandorArpa:
shoulda, woulda, coulda, it doesn't 🤷

DomBelcher:
@SandorArpa is there any way to reliably reproduce the flaky test behaviour?

SandorArpa:
@DomBelcher 
https://github.com/alphagov/pay-connector/actions/runs/8539261738/job/23393590036
https://github.com/alphagov/pay-connector/actions/runs/8539261738/job/23394080910
and a few more during merge

oswaldquek:
> @DomBelcher https://github.com/alphagov/pay-connector/actions/runs/8539261738/job/23393590036 https://github.com/alphagov/pay-connector/actions/runs/8539261738/job/23394080910 and a few more during merge

It looks like these tests are failing on [findById_shouldFindGatewayAccount](https://github.com/alphagov/pay-connector/blob/740a60ac492cfd2d8ce53ee3ddf82393cd690756/src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoIT.java#L198) which uses `contains` rather than `containsInAnyOrder`.

DomBelcher:
Is there any way to be sure in the failing tests that `gatewayAccount.getCardTypes()` did actually return a list of all the correct card types, and that the test failed because the matcher failed to properly test for that? Is it possible that only the Visa card type was returned (or would that have given a different error)?

DomBelcher:
> Is there any way to be sure in the failing tests that `gatewayAccount.getCardTypes()` did actually return a list of all the correct card types, and that the test failed because the matcher failed to properly test for that? Is it possible that only the Visa card type was returned (or would that have given a different error)?

This does cause the same error, so I think it's not a problem with the matcher, but with the gateway account only having Visa as an accepted card type. FWIW the default test account created in `ITestBaseExtension` only has Visa as an accepted card type, however I can't see any reason that the test account created in `findById_shouldFindGatewayAccount` would be conflicting with it. 

Changing `contains` to `containsInAnyOrder` gives a more accurate error in this instance, correctly identifying that there is no Mastercard card type in the list.

DomBelcher:
> This does cause the same error, so I think it's not a problem with the matcher, but with the gateway account only having Visa as an accepted card type

Not sure about this actually, as `contains` is order dependent, and having both card types returned but with Visa appearing before Mastercard in the List causes the same error. It's strange if this has never shown up before though. 

Whatever the cause is, changing `contains` to `containsInAnyOrder` should help diagnose it if the error ever shows up again. (unless that also throws up issues, I can't find any tests failing due to that, but it doesn't mean there aren't any)

PR #5102, opened 2024-04-08
PP-12302 Use WireMock dynamicPort & localhost binding

DomBelcher:
## WHAT YOU DID
 - Change setup for integration tests to use Wiremock's `dynamicPort` config instead of `Portfactory.findFreePort`, and bind WireMock servers to `localhost`
 - This prevents occasional flaky tests caused by choosing ports with PortFactory and by binding to the default 0.0.0.0 address
 - This can cause tests to fail where a port is free on localhost but not on another address
 - This also saves time on the setup for each integration test, as dyanmicPort is faster than PortFactory.findFreePort

## How to test
Run integration tests


SandorArpa (approved):
LOKTM

PR #5139, opened 2024-04-19
 PP-11555 create transfer to send back to connect account

AbdirizakIdris:
## WHAT YOU DID
  - Implement method for reversal of a refund when Stripe refund is in failed state

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
+++ b/src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
@@ -9,11 +9,10 @@ import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountEntity;
 import uk.gov.pay.connector.refund.dao.RefundDao;
 import uk.gov.pay.connector.refund.model.domain.Refund;
 
-
 import javax.inject.Inject;
 import javax.ws.rs.WebApplicationException;
+import java.util.Map;
 import java.util.Optional;
-
 
 import static java.lang.String.format;
 import static uk.gov.pay.connector.util.ResponseUtil.badRequestResponse;
@@ -23,13 +22,15 @@ public class RefundReversalService {
     private final LedgerService ledgerService;
     private final RefundDao refundDao;
     private final StripeSdkClientFactory stripeSdkClientFactory;
+    private final RefundReversalStripeConnectTransferRequestBuilder refundRequest;
 
 
     @Inject
-    public RefundReversalService(LedgerService ledgerService, RefundDao refundDao, StripeSdkClientFactory stripeSdkClientFactory) {
+    public RefundReversalService(LedgerService ledgerService, RefundDao refundDao, StripeSdkClientFactory stripeSdkClientFactory, RefundReversalStripeConnectTransferRequestBuilder refundRequest) {
         this.ledgerService = ledgerService;
         this.refundDao = refundDao;
         this.stripeSdkClientFactory = stripeSdkClientFactory;
+        this.refundRequest = refundRequest;
     }
 
 
@@ -53,7 +54,10 @@ public class RefundReversalService {
 
 
             if ("failed".equals(refundStatus)) {
-                // TODO: 04/01/2024  (PP-11555)
+
+                Map<String, Object> refundRequestBody = refundRequest.createRequest(refundFromStripe);
+                stripeClient.createTransfer(refundRequestBody, isLiveGatewayAccount);
+
             } else {
                 throw new WebApplicationException(badRequestResponse(
                         format("Refund with Refund ID: %s and Stripe ID: %s is not in a failed state", refundExternalId, stripeRefundId)));
--- a/src/test/java/uk/gov/pay/connector/gateway/stripe/StripeSdkClientTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/stripe/StripeSdkClientTest.java
@@ -49,6 +49,9 @@ class StripeSdkClientTest {
     public static final String STRIPE_CONNECT_ACCOUNT_ID = "stripe-account-id";
     public static final String CUSTOMER_ID = "customer-id";
     public static final String STRIPE_REFUND_ID = "a-stripe-refund-id";
+    public static final Map<String, Object> transferRequest = Map.of(
+            "destination", "destination",
+            "amount", "amount");
 
     @BeforeEach
     public void setUp() {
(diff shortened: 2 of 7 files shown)


src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -53,7 +53,13 @@ public void reverseFailedRefund(GatewayAccountEntity gatewayAccount, Refund refu
 
 
             if ("failed".equals(refundStatus)) {
-                // TODO: 04/01/2024  (PP-11555)
+                
+                RandomIdGenerator randomIdGenerator = new RandomIdGenerator();
```
alexbishop1: We should be able to have one `RandomIdGenerator` as a `private static final` field (because it should be threadsafe) rather than creating a new one each time.

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -53,7 +53,13 @@ public void reverseFailedRefund(GatewayAccountEntity gatewayAccount, Refund refu
...
+                RefundReversalStripeConnectTransferRequestBuilder refundRequest = new RefundReversalStripeConnectTransferRequestBuilder(randomIdGenerator);
+
+                Map<String, Object> refundRequestBody = refundRequest.createRequest(refundFromStripe);
+                stripeClient.createTransfer(refundRequestBody, isLiveGatewayAccount);
```
alexbishop1, line 59: We need to make sure we do something sensible if this call causes a `StripeException` to be thrown. It might be fine just to let it bubble up as we have here or we may want to catch it and handle it that way. At a minimum we probably want an integration test to make sure connector returns an appropriate response if a `StripeException` happens. This can wait until a follow-up pull request.

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeSdkClientTest.java
```diff
@@ -120,9 +123,33 @@ void getRefund_shouldUseTestApiKey() throws Exception {
...
+
         Map<String, Object> params = paramsArgumentCaptor.getValue();
         assertThat(params, hasEntry("expand", List.of("charge")));
         assertThat(requestOptionsArgumentCaptor.getValue().getApiKey(), is(TEST_API_KEY));
     }
+
+    @Test
+    void createTransfer_shouldUseLiveAPIKey() throws Exception {
+        when(stripeAuthTokens.getLive()).thenReturn(LIVE_API_KEY);
+        stripeSDKClient.createTransfer(transferRequest, true);
```
alexbishop1, line 139: Put a blank line before this line please. This is the hey-look-this-is-what-we’re-testing bit so it’s good to make it stand out!

src/test/java/uk/gov/pay/connector/gateway/stripe/StripeSdkClientTest.java
```diff
@@ -120,9 +123,33 @@ void getRefund_shouldUseTestApiKey() throws Exception {
...
+
+        verify(stripeSDKWrapper).createTransfer(paramsArgumentCaptor.capture(), requestOptionsArgumentCaptor.capture());
+
+        assertThat(requestOptionsArgumentCaptor.getValue().getApiKey(), is(LIVE_API_KEY));
+
+        assertThat(transferRequest, hasEntry("destination", "destination"));
```
alexbishop1: This is not quite right because we’re checking the map passed to `stripeSDKClient`. We want to check the map passed to `stripeSDKWrapper`. So get the map from the `paramsArgumentCaptor` and check it. Since it’s just passed straight through, we can just assert that the map is the same as `transferRequest`.

src/test/java/uk/gov/pay/connector/it/resources/stripe/RefundReversalResourceIT.java
```diff
@@ -70,14 +71,27 @@ public void setUp() {
     }
 
     @Test
-    public void shouldSuccessfullyGetRefundFromStripe() throws JsonProcessingException, StripeException {
+    void shouldSuccessfullyGetRefundFromStripe() throws JsonProcessingException, StripeException {
```
alexbishop1: This is now testing it actually does the transfer so it should be renamed.

src/test/java/uk/gov/pay/connector/it/resources/stripe/RefundReversalResourceIT.java
```diff
@@ -70,14 +71,27 @@ public void setUp() {
...
         app.getLedgerStub().returnLedgerTransaction(CHARGE_EXTERNAL_ID, charge);
         app.getLedgerStub().returnLedgerTransaction(REFUND_EXTERNAL_ID, refund);
 
         when(mockStripeSdkClient.getRefund(eq(STRIPE_REFUND_ID), anyBoolean())).thenReturn(mockStripeRefund);
         app.getStripeMockClient().mockRefund();
         
         when(mockStripeRefund.getStatus()).thenReturn("failed");
+
+        com.stripe.model.Charge mockedStripeCharge = Mockito.mock(com.stripe.model.Charge.class);
+        com.stripe.model.Account mockedStripeAccount = Mockito.mock(com.stripe.model.Account.class);
```
alexbishop1: Usually we’d have mocks like this as `@Mock` fields.

src/test/java/uk/gov/pay/connector/service/RefundReversalServiceTest.java
```diff
@@ -114,6 +114,18 @@ void shouldNotThrowExceptionWhenRefundIsInFailedState() throws StripeException {
...
+
+        when(mockedStripeRefund.getChargeObject()).thenReturn(mockedStripeCharge);
+
+        when(mockedStripeCharge.getId()).thenReturn("ch_sdkhdg887s");
+        when(mockedStripeCharge.getOnBehalfOfObject()).thenReturn(mockedStripeAccount);
+        when(mockedStripeAccount.getId()).thenReturn("acct_jdsa7789d");
+        when(mockedStripeCharge.getTransferGroup()).thenReturn("abc");
+        when(mockedStripeRefund.getAmount()).thenReturn(100L);
+        when(mockedStripeRefund.getCurrency()).thenReturn("GBP");
+
```
alexbishop1, line 133: We want to make sure we actually translate the above into an appropriate request to Stripe. So at the end of this test, add a `verify(mockStripeSDKClient).createTransfer(…)` line that checks we’re passing in a map with the right values.

alexbishop1 (changes requested):
Looking good! I think we want to make some of the tests a bit more rigorous in terms of what they verify and I’ve made a few stylistic suggestions too.

I’ve also put a comment about making sure we do the handle the transfer failing, which can wait until a follow-up pull request (but don’t forget it!).

src/test/java/uk/gov/pay/connector/it/resources/stripe/RefundReversalResourceIT.java
alexbishop1: Still needs to be renamed.

alexbishop1 (changes requested):
One test to rename and this is good to merge!

alexbishop1 (approved):

PR #5141, opened 2024-04-19
WIP - PP-12354 Resolve multiple gateway accounts

james-peacock-ho:
This is for discussion in mobbing, no requirement to merge.
- return the correct gateway account when there are multiple gateway accounts for a particular serviceId and account type, according to strategy set out in PP-12352.
- where there are both Stripe and Worldpay accounts (as can be the case for test accounts), return Stripe. We may need to address this unlikely scenario differently, e.g. return the most recently created.


james-peacock-ho:
closing this for now as we're not going to attempt to implement this logic until requirements are clearer

PR #5148, opened 2024-04-22
PP-12354 using service id instead of gateway

james-peacock-ho:
- Add v1/frontend endpoint to get card types for a gateway account by serviceId and accountType (test|live)
- Add v1/api endpoint to get gateway account by serviceId and accountType
- Add unit tests
- Add integration tests and removed some redundant integration tests

NB: where multiple test gateway accounts exist for a service, the one most recently added to the database will be returned. This logic will need refinement in a future story.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
@@ -23,6 +23,7 @@ import uk.gov.pay.connector.gatewayaccount.model.CreateGatewayAccountResponse;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountRequest;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountResponse;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountSearchParams;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountWithCredentialsResponse;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountsListDTO;
 import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
@@ -48,7 +49,6 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 import javax.ws.rs.core.UriInfo;
 import java.net.URI;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -70,9 +70,6 @@ import static uk.gov.pay.connector.util.ResponseUtil.successResponseWithEntity;
 public class GatewayAccountResource {
 
     private static final Logger logger = LoggerFactory.getLogger(GatewayAccountResource.class);
-
-    private static final String DESCRIPTION_FIELD_NAME = "description";
-    private static final String ANALYTICS_ID_FIELD_NAME = "analytics_id";
     private static final String SERVICE_NAME_FIELD_NAME = "service_name";
     private static final String REQUIRES_3DS_FIELD_NAME = "toggle_3ds";
     private static final String CARD_TYPES_FIELD_NAME = "card_types";
@@ -120,6 +117,26 @@ public class GatewayAccountResource {
         return gatewayAccountService.getGatewayAccount(gatewayAccountId)
                 .map(GatewayAccountWithCredentialsResponse::new)
                 .orElseThrow(() -> new GatewayAccountNotFoundException(gatewayAccountId));
+    }
+
+    @GET
+    @Path("/v1/api/service/{serviceId}/{accountType}/account")
+    @Produces(APPLICATION_JSON)
+    @Operation(
+            summary = "Find gateway account by service external ID and account type (test|live)",
+            description = "Get gateway account by service external ID and account type (test|live). Returns notifications credentials, gateway account credentials (without password). Doesn't include card_types or gateway_merchant_id",
+            tags = {"Gateway accounts"},
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "OK",
+                            content = @Content(schema = @Schema(implementation = GatewayAccountWithCredentialsResponse.class))),
+                    @ApiResponse(responseCode = "404", description = "Not found")
+            }
+    )
+    public GatewayAccountWithCredentialsResponse getGatewayAccountByServiceIdAndAccountType(@PathParam("serviceId") String serviceId, @PathParam("accountType") GatewayAccountType accountType) {
+        GatewayAccountWithCredentialsResponse response = gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, accountType)
+                .map(GatewayAccountWithCredentialsResponse::new)
+                .orElseThrow(() -> new GatewayAccountNotFoundException(format("Gateway account for service external id %s and account type %s not found.", serviceId, accountType)));
+        return response;
     }
 
     @GET
@@ -210,6 +227,36 @@ public class GatewayAccountResource {
                 .orElseGet(() -> notFoundResponse(format("Account with id %s not found.", accountId)));
     }
 
+    @GET
+    @Path("/v1/frontend/service/{serviceId}/{accountType}/card-types")
+    @Produces(APPLICATION_JSON)
+    @Operation(
+            summary = "Get card types for gateway account by service external ID and account type",
+            tags = {"Gateway accounts"},
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "OK",
+                            content = @Content(schema = @Schema(example = "{" +
+                                    "    \"card_types\": [" +
+                                    "        {" +
+                                    "            \"id\": \"ab8a3abd-bcfd-4fa6-8905-321ce913e7f5\"," +
+                                    "            \"brand\": \"visa\"," +
+                                    "            \"label\": \"Visa\"," +
+                                    "            \"type\": \"DEBIT\"," +
+                                    "            \"requires3ds\": false" +
+                                    "        }" +
+                                    "    ]" +
+                                    "}"))),
+                    @ApiResponse(responseCode = "404", description = "Not found")
+            }
+    )
+    public Response getAcceptedCardTypesByServiceIdAndAccountType(@PathParam("serviceId") String serviceId, @PathParam("accountType") GatewayAccountType accountType) {
+        logger.info("Getting accepted card types for service id {}, account type {}", serviceId, accountType.toString());
+        System.out.println("Account type: " + accountType);
+        return gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, accountType)
+                .map(gatewayAccount -> successResponseWithEntity(ImmutableMap.of(CARD_TYPES_FIELD_NAME, gatewayAccount.getCardTypes())))
+                .orElseGet(() -> notFoundResponse(format("Gateway account for service external id %s and account type %s not found.", serviceId, accountType)));
+    }
+    
     @POST
     @Path("/v1/api/accounts")
     @Consumes(APPLICATION_JSON)
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
@@ -253,7 +253,7 @@ public class GatewayAccountResource {
         logger.info("Getting accepted card types for service id {}, account type {}", serviceId, accountType.toString());
         System.out.println("Account type: " + accountType);
         return gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, accountType)
-                .map(gatewayAccount -> successResponseWithEntity(ImmutableMap.of(CARD_TYPES_FIELD_NAME, gatewayAccount.getCardTypes())))
+                .map(gatewayAccount -> successResponseWithEntity(Map.of(CARD_TYPES_FIELD_NAME, gatewayAccount.getCardTypes())))
                 .orElseGet(() -> notFoundResponse(format("Gateway account for service external id %s and account type %s not found.", serviceId, accountType)));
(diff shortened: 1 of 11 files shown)


src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -90,6 +112,18 @@ void getAccountShouldNotReturnCardTypes() {
...
+    @Test
+    void getAccountByServiceIdAndAccountTypeShouldNotReturnCredentialsOrCardTypes() {
+        app.createAGatewayAccountWithServiceId("a-valid-service-id");
+        String url = ACCOUNTS_API_SERVICE_ID_URL.replace("{serviceId}", "a-valid-service-id").replace("{accountType}", GatewayAccountType.TEST.name());
+        app.givenSetup()
+                .get(url)
+                .then()
+                .statusCode(200)
+                .body("card_types", is(nullValue()))
+                .body("credentials", is(nullValue()));
```
DomBelcher: This is more a problem with the obfuscation of creating the gateway accounts, but it's not clear whether the card_types and credentials are null because they're not returned, or they just don't exist on the account - I think it's actually both in this case, there are no credentials created on the gateway account, but if there were they wouldn't be returned

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -133,6 +168,20 @@ void getAccountShouldReturn3dsSetting() {
...
+        String gatewayAccountId = app.createAGatewayAccountFor("stripe", "desc", "id", "true", "test");
+        app.getDatabaseTestHelper().updateServiceIdFor(Long.parseLong(gatewayAccountId), "a-valid-service-id");
+        String url = ACCOUNTS_API_SERVICE_ID_URL.replace("{serviceId}", "a-valid-service-id").replace("{accountType}", GatewayAccountType.TEST.name());
+        app.givenSetup()
+                .get(url)
+                .then()
+                .statusCode(200)
+                .body("analytics_id", is("id"))
+                .body("description", is("desc"))
+                .body("requires3ds", is(true));
```
DomBelcher: I think this test could be combined with `getAccountByServiceIdAndAccountTypeShouldNotReturnCredentialsOrCardTypes`

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -168,6 +217,30 @@ void getAccountShouldReturnCorporatePrepaidDebitCardSurchargeAmount() {
...
+                .insert();
+
+        String url = ACCOUNTS_API_SERVICE_ID_URL.replace("{serviceId}", "a-valid-service-id").replace("{accountType}", GatewayAccountType.TEST.name());
+        app.givenSetup()
+                .get(url)
+                .then()
+                .statusCode(200)
+                .body("corporate_credit_card_surcharge_amount", is(corporateCreditCardSurchargeAmount))
+                .body("corporate_debit_card_surcharge_amount", is(corporateDebitCardSurchargeAmount))
+                .body("corporate_prepaid_debit_card_surcharge_amount", is(corporatePrepaidDebitCardSurchargeAmount));
```
DomBelcher: This could also be combined with `getAccountByServiceIdAndAccountTypeShouldNotReturnCredentialsOrCardTypes`

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
DomBelcher: I've dug into this and it seems that this is an outdated test. The `credentials` field on `gateway_accounts` was dropped (#3117) but this test persisted. I assume it can safely be deleted

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceIT.java
```diff
@@ -136,6 +138,28 @@ void shouldAcceptAllCardTypesNotRequiring3DSForNewlyCreatedAccountAs3dsIsDisable
         validateNon3dsCardType(response, "unionpay", "Union Pay", "CREDIT");
     }
 
+    @Test
+    void shouldGetCardTypesByServiceIdAndAccountType() {
+        app.createAGatewayAccountWithServiceId("a-service-id");
```
DomBelcher: My feeling is we shouldn't use this method of creating accounts, as it works by doing a POST request to the `/v1/api/accounts` endpoint, rather than inserting a record directly into the database, which makes the test depend on that endpoint functioning correctly as well.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -269,6 +342,109 @@ void shouldReturnAccountInformationForGetAccountById_withWorldpayCredentials() {
                 .body("gateway_account_credentials[0].credentials.recurring_merchant_initiated", not(hasKey("password")));
     }
 
+    @Test
+    void shouldReturnAccountInformationForGetAccountByServiceIdAndAccountType_withWorldpayCredentials() {
```
DomBelcher: The naming of the tests is a bit inconsistent, and they could be broken up a bit to increase readability - e.g. 
`getAccountByServiceIdAndAccountType_withWorldpayCredentials_shouldReturnAccountInformation`

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -269,6 +342,109 @@ void shouldReturnAccountInformationForGetAccountById_withWorldpayCredentials() {
...
+                                CREDENTIALS_USERNAME, "cit-username",
+                                CREDENTIALS_PASSWORD, "cit-password"),
+                        RECURRING_MERCHANT_INITIATED, Map.of(
+                                CREDENTIALS_MERCHANT_CODE, "mit-merchant-code",
+                                CREDENTIALS_USERNAME, "mit-username",
+                                CREDENTIALS_PASSWORD, "mit-password")))
+                .build();
+
+        this.defaultTestAccount = DatabaseFixtures
+                .withDatabaseTestHelper(app.getDatabaseTestHelper())
```
DomBelcher, line 400: I think this method of creating a test account is better as it works directly with the database. It is also possible to use `app.getDatabaseFixtures()` instead of using the static `DatabaseFixtures`

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
DomBelcher: Another option would be to use the `@Nested` annotation in JUnit5 to create groups of tests for each method, e.g. having 

```
@Nested
class GetAccountByServiceIdAndAccountType {
  @Test
  void withWorldpayCredentials_shouldReturnAccountInformation() {
    // test body
  }
  // rest of tests...
}
```

which would make it easier to separate the tests for different methods. Could also use the `@DisplayName` annoation to show the API endpoint instead of the class name, e.g. `@DisplayName("GET /v1/services/{serviceId}/{accountType}/account")`

This would be a break from current convention but could be a nice convention to establish going forward - maybe worth further discussion

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -349,6 +558,18 @@ void shouldNotReturn3dsFlexCredentials_whenGatewayAccountHasNoCreds() {
                 .body("worldpay_3ds_flex", nullValue());
     }
 
+    @Test
+    void shouldNotReturn3dsFlexCredentials_whenGatewayAccountHasNoCreds_byServiceIdAndAccountType_byServiceIdAndAccountType() {
```
DomBelcher: Method name duplicates `byServiceIdAndAccountType`

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -749,6 +970,25 @@ void shouldReturn3dsFlexCredentials_whenGatewayAccountHasCreds() {
                 .body("worldpay_3ds_flex.exemption_engine_enabled", is(false));
     }
 
+    @Test
+    void shouldReturn3dsFlexCredentials_whenGatewayAccountHasCreds_byServiceIdAndAccountType() {
```
DomBelcher, line 956: Could be combined with `shouldReturnAccountInformationForGetAccountByServiceIdAndAccountType_withWorldpayCredentials` - I think we should look to combine the happy path tests into a few which test important cases, e.g. `worldpay` and `stripe` payment providers, and possibly `test` and `live` account types.

DomBelcher (dismissed):
A few things worth discussing/thinking about further

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
DomBelcher: The `card_types` parameter also appears to be outdated. The `getCardTypes()` method on `GatewayAccountEntity` was previously annotated with `@JsonProperty(card_types)`, so this tested that wasn't returned, but this has since changed. Not sure it's worth keeping around, it certainly feels overkill to have a test dedicated to it

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -210,6 +227,36 @@ public Response getGatewayAccountAcceptedCardTypes(@Parameter(example = "1", des
...
+                                    "    ]" +
+                                    "}"))),
+                    @ApiResponse(responseCode = "404", description = "Not found")
+            }
+    )
+    public Response getAcceptedCardTypesByServiceIdAndAccountType(@PathParam("serviceId") String serviceId, @PathParam("accountType") GatewayAccountType accountType) {
+        logger.info("Getting accepted card types for service id {}, account type {}", serviceId, accountType.toString());
+        System.out.println("Account type: " + accountType);
+        return gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, accountType)
+                .map(gatewayAccount -> successResponseWithEntity(ImmutableMap.of(CARD_TYPES_FIELD_NAME, gatewayAccount.getCardTypes())))
```
oswaldquek: Could use the standard Java Map instead of ImmutableMap here...

src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoIT.java
```diff
@@ -646,6 +647,23 @@ void findByExternalId_shouldFindGatewayAccount() {
...
+    void findByServiceIdAndAccountType_shouldFindGatewayAccount() {
+        Long id = nextLong();
+        String externalId = randomUuid();
+        String serviceExternalId = randomUuid();
+        databaseFixtures
+                .aTestAccount()
+                .withAccountId(id)
+                .withExternalId(externalId)
+                .withServiceId(serviceExternalId)
+                .insert();
```
oswaldquek: Can we add another gateway account to make sure the code won't fall over if there are two accounts?

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -55,41 +56,198 @@ public class GatewayAccountResourceIT {
...
-                .get(ACCOUNTS_API_URL + unknownAccountId)
+                .get(url)
                 .then()
                 .statusCode(404);
     }
 
     @Test
-    void getAccountShouldNotReturnCredentials() {
-        String gatewayAccountId = testBaseExtension.createAGatewayAccountFor("worldpay");
+    void shouldReturnAccountInformation_withWorldpayCredentials_whenGetByServiceIdAndAccountType() {
```
oswaldquek: Could shorten this to shouldReturnWorldpayAccountInformation_whenGetByServiceIdAndAccountType. The test tests all the account information exists, not just the worldpay credentials so we can remove this from the test name.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -55,41 +56,198 @@ public class GatewayAccountResourceIT {
...
+                .body("allow_authorisation_api", is(true))
+                .body("recurring_enabled", is(true))
+                .body("requires3ds", is(true))
+                .body("block_prepaid_cards", is(true))
+                .body("disabled", is(true))
+                .body("disabled_reason", is("Disabled because reasons"))
+                .body("gateway_account_credentials.size()", is(1))
+                .body("gateway_account_credentials[0].payment_provider", is("worldpay"))
+                .body("gateway_account_credentials[0].state", is("ACTIVE"))
+                .body("gateway_account_credentials[0].gateway_account_id", is(accountIdAsInt))
```
oswaldquek: `accountIdAsInt` is only used here, so it could inlined as `Math.toIntExact(accountId)` and line 135 can go.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -55,41 +56,198 @@ public class GatewayAccountResourceIT {
...
+                .body("worldpay_3ds_flex", not(hasKey("jwt_mac_key")))
+                .body("worldpay_3ds_flex", not(hasKey("version")))
+                .body("worldpay_3ds_flex", not(hasKey("gateway_account_id")))
+                .body("worldpay_3ds_flex.exemption_engine_enabled", is(false));
     }
 
     @Test
-    void getAccountShouldNotReturnCardTypes() {
-        String gatewayAccountId = testBaseExtension.createAGatewayAccountFor("worldpay");
+    void shouldReturnAccountInformation_withStripeCredentials_whenGetByServiceIdAndAccountType() {
```
oswaldquek: Like the test above, do we want to test for all the returned parameters, rather than just the credentials?

oswaldquek (changes requested):
For the future, a PR for one endpoint would be better :)

oswaldquek (approved):

PR #5182, opened 2024-05-01
PP-12374 Pass Wallet Payments authorisation success to metrics and alerts

marcotranchino:
With a previous PR[1], we have clarified the logs written by Connector when a
Google Pay or an Apple Pay payment is rejected, by indicating the charge
status more clearly, and then that the connection request had been
successful, despite the payment didn't go through.

With this change, we are now correcting the `successOrFailure` value being
passed on to metrics and alerts.

Previously we were passing success or failure depending on the connection to
the Payment Gateway being successful or not.

With this change, we are now passing `success` only in two cases:
- if the payment has gone through, i.e. only for a charge status of
  `AUTHORISATION SUCCESS`
- if 3D Secure authorisation is required, i.e. for a charge status of
  `AUTHORISATION_3DS_REQUIRED`

The latter happens only in a small proportion of Google Pay payments, and we
have agreed to count these as success(close enough and there will not be many
of them).

For all other charge status values, we will from now on be passing on the
`failure` value.

Further information in JIRA[2].

[1]
https://github.com/alphagov/pay-connector/pull/5176

[2]
https://payments-platform.atlassian.net/browse/PP-12374

## REQUEST FOR FEEDBACK

Are we happy to count success just for AUTHORISATION_SUCCESS ?
In src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeStatus.java we have other options that are not failures, e.g. AUTHORISATION_SUBMITTED or CAPTURE_APPROVED etc.

Would we want to pass something other than success/failure for those cases?

## How to test

Manually, locally
Searching in Splunk after deployment

### Logging

- [ ] ~~only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert~~.

### Documentation

- [ ] ~~Updated README.md for any of the following ?~~

* ~~Introduced any new environment variables / removed existing environment variable~~
* ~~Added new API / updated existing API definition~~

### DIFF
--- a/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
@@ -31,6 +31,8 @@ import uk.gov.service.payments.commons.model.CardExpiryDate;
 import java.util.Optional;
 
 import static java.lang.String.format;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
 
 public class WalletAuthoriseService {
     
@@ -135,12 +137,14 @@ public class WalletAuthoriseService {
 
     private void logMetrics(ChargeEntity chargeEntity,
                             GatewayResponse<BaseAuthoriseResponse> operationResponse,
-                            String successOrFailure,
+                            String requestStatus,
                             ChargeStatus chargeStatus,
                             WalletType walletType) {
-
+        String successOrFailure = (chargeStatus == AUTHORISATION_SUCCESS || chargeStatus == AUTHORISATION_3DS_REQUIRED)
+                ? "success"
+                : "failure";
         LOGGER.info("{} authorisation - charge status={}, request status={}, charge_external_id={}, payment provider response={}",
-                walletType.toString(), chargeStatus, successOrFailure, chargeEntity.getExternalId(), operationResponse.toString());
+                walletType.toString(), chargeStatus, requestStatus, chargeEntity.getExternalId(), operationResponse.toString());
         metricRegistry.counter(format("gateway-operations.%s.%s.authorise.%s.result.%s",
                 chargeEntity.getPaymentProvider(),
                 chargeEntity.getGatewayAccount().getType(),
--- a/src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
@@ -83,6 +83,7 @@ import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.isNull;
 import static org.mockito.ArgumentMatchers.nullable;
 import static org.mockito.Mockito.anyString;
+import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.lenient;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
@@ -234,6 +235,14 @@ class WalletAuthoriseServiceTest extends CardServiceTest {
         assertThat(loggingEvents.stream().anyMatch(le -> le.getFormattedMessage().contains(
                 format("APPLE_PAY authorisation - charge status=AUTHORISATION SUCCESS, request status=success, charge_external_id=%s, payment provider response=%s", charge.getExternalId(), gatewayResponse.toString()))
         ), is(true));
+
+        String expectedMetric = format("gateway-operations.%s.%s.authorise.%s.result.%s",
+                "sandbox",
+                "test",
+                "apple-pay",
+                "success");
+
+        verify(mockMetricRegistry).counter(expectedMetric);
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


marcotranchino:
I have discussed with @jfharden and ascertained that if this PR is approved as is, there is no need to change any metrics or dashboards, given that this PR is only changing what counts as success and what counts as failure, and there's no change to the possible values. [Further details in JIRA](https://payments-platform.atlassian.net/browse/PP-12374).

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -135,12 +135,15 @@ public GatewayResponse<BaseAuthoriseResponse> authorise(String chargeId, WalletA
...
                             GatewayResponse<BaseAuthoriseResponse> operationResponse,
-                            String successOrFailure,
+                            String requestStatus,
                             ChargeStatus chargeStatus,
                             WalletType walletType) {
-
+        String successOrFailure = "failure";
+        if (chargeStatus == ChargeStatus.AUTHORISATION_SUCCESS) {
+            successOrFailure = "success";
+        }
```
alexbishop1: This might be a good opportunity for a ternary operator.

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
```diff
@@ -346,6 +354,30 @@ void doAuthorise_shouldRespondAuthorisationRejected_whenProviderAuthorisationIsR
...
+    void verifyLogging_applePay_shouldLogFailure_WhenAuthorisationIsRejected() throws Exception {
+        providerWillRejectApplePay();
+
+        GatewayResponse response = walletAuthoriseService.authorise(charge.getExternalId(), validApplePayDetails);
+
+        assertThat(response.isSuccessful(), is(true));
+        assertThat(charge.getStatus(), is(AUTHORISATION_REJECTED.getValue()));
+
+        ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
+        verify(mockAppender, times(4)).doAppend(loggingEventArgumentCaptor.capture());
```
alexbishop1: I’d say `atLeast(1)` rather than `times(4)` to make this less brittle.

alexbishop1 (changes requested):
Looks good! Just a few minor suggestions.

> Are we happy to count success just for AUTHORISATION_SUCCESS ?
In src/main/java/uk/gov/pay/connector/charge/model/domain/ChargeStatus.java we have other options that are not failures, e.g. AUTHORISATION_SUBMITTED or CAPTURE_APPROVED etc.
>
> Would we want to pass something other than success/failure for those cases?

Most of those statuses cannot happen at this point. Almost all of the time it’s going to be `AUTHORISATION_SUCCESS`, `AUTHORISATION_REJECTED` or (more rarely) `AUTHORISATION_ERROR`.

For Google Pay, a small number of payments may go `AUTHORISATION_3DS_REQUIRED`, which is neither success nor failure (it’s saying 3D Secure authorisation is required). We can probably count these as success (close enough and there will not be many of them).

Anything else either is not going to happen at this point or can be classed as a failure.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
marcotranchino: Amended, thanks! :)

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
marcotranchino: Amended, thanks! :)

marcotranchino:
Hi @alexbishop1 
Thanks for the review! 
I have made the suggested amendments, so this is now ready for re-review.
Cheers :)

marcotranchino:
~~[Concourse-ci integration tests are failing ](https://pay-cd.deploy.payments.service.gov.uk/builds/122348060) but the issue seems to be unrelated to this change ([I have verified that they fail with a copy from main too](https://github.com/alphagov/pay-connector/pull/5193)).~~

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
```diff
@@ -82,14 +82,7 @@
...
 import static org.mockito.ArgumentMatchers.nullable;
-import static org.mockito.Mockito.anyString;
-import static org.mockito.Mockito.lenient;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
alexbishop1: Please do not use wildcard imports.

alexbishop1 (changes requested):
Change the wildcard imports and this is good to go.

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
marcotranchino: Amended, thanks

marcotranchino:
Amended, thanks!

alexbishop1 (approved):

PR #5216, opened 2024-05-10
PP-12354 Get Stripe Setup by serviceId and accountType

james-peacock-ho:
- add endpoint to get Stripe setup data by serviceId and accountType
- add integration tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
@@ -10,6 +10,8 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.parameters.RequestBody;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.tags.Tag;
+import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType;
 import uk.gov.pay.connector.gatewayaccount.model.StripeAccountSetup;
 import uk.gov.pay.connector.gatewayaccount.model.StripeAccountSetupUpdateRequest;
 import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/it/resources/StripeAccountSetupResourceIT.java
```diff
@@ -19,59 +20,134 @@ public class StripeAccountSetupResourceIT {
...
+                    .body("message[0]", is("Gateway account not found for service ID [unknown-service-id] and account type [test]"));
+        }
+
+        @Test
+        public void returnsSuccessfulResponseWhenGatewayAccountIsNotAStripeAccount() {
+            testBaseExtension.createAGatewayAccountWithServiceId("a-valid-service-id","worldpay");
+            app.givenSetup()
+                    .get("/v1/api/service/a-valid-service-id/TEST/stripe-setup")
+                    .then()
+                    .statusCode(200);
```
DomBelcher: While this is how the existing endpoint works, I'm not sure a `200` response makes sense in this case - there is no stripe setup for a non-stripe account, so I would think it should be a `404` - something we may need to think about?

src/test/java/uk/gov/pay/connector/it/resources/StripeAccountSetupResourceIT.java
```diff
@@ -19,59 +20,134 @@ public class StripeAccountSetupResourceIT {
...
+                            ImmutableMap.of(
+                                    "op", "replace",
+                                    "path", "vat_number",
+                                    "value", true),
+                            ImmutableMap.of(
+                                    "op", "replace",
+                                    "path", "director",
+                                    "value", true)
+                    )))
+                    .patch("/v1/api/accounts/" + gatewayAccountId + "/stripe-setup")
```
DomBelcher, line 118: Good choice to make use of the existing `patch` endpoint 👍

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
```diff
@@ -64,6 +66,25 @@ public StripeAccountSetup getStripeAccountSetup(
...
+                    @ApiResponse(responseCode = "200", description = "OK",
+                            content = @Content(schema = @Schema(implementation = StripeAccountSetup.class))),
+                    @ApiResponse(responseCode = "404", description = "Not found")
+            }
+    )
+    public StripeAccountSetup getStripeAccountSetupByServiceIdAndAccountType(
+            @Parameter(example = "46eb1b601348499196c99de90482ee68", description = "Service ID") @PathParam("serviceId") String serviceId, 
+            @Parameter(example = "test", description = "Account type") @PathParam("accountType") GatewayAccountType accountType) {
+        return gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, accountType)
+                .map(gatewayAccountEntity -> stripeAccountSetupService.getCompletedTasks(gatewayAccountEntity.getId()))
```
DomBelcher, line 84: This is definitely the most straightforward way of retrieving the stripe account setup, but it might be possible to do it on the database layer (e.g. using a `JOIN`) which may improve performance, but I'm not sure it's worth the extra effort

src/test/java/uk/gov/pay/connector/it/resources/StripeAccountSetupResourceIT.java
```diff
@@ -19,59 +20,134 @@ public class StripeAccountSetupResourceIT {
     public static AppWithPostgresAndSqsExtension app = new AppWithPostgresAndSqsExtension();
     public static GatewayAccountResourceITBaseExtensions testBaseExtension = new GatewayAccountResourceITBaseExtensions("sandbox", app.getLocalPort());
 
-    @Test
-    public void getStripeSetupWithNoTasksCompletedReturnsFalseFlags() {
-        String gatewayAccountId = testBaseExtension.createAGatewayAccountFor("stripe");
+    @Nested
+    class getStripeSetupByGatewayAccountId {
+        @Test
+        public void withNoTasksCompletedReturnsFalseFlags() {
```
DomBelcher: Could remove the `public` modifiers

DomBelcher (commented):
Nice work 👍 - no essential changes but some things which may be worth us thinking about

src/test/java/uk/gov/pay/connector/it/resources/StripeAccountSetupResourceIT.java
james-peacock-ho: This is to cater for PSP switch scenario where e.g. a service moves from Worldpay to Stripe - the same gateway account is used but just a new set of credentials - so in the process of the switch you have a 'worldpay' gateway account with stripe setup data.

DomBelcher: Makes sense. It's probably worth adding some assertions that the body is what is expected - it seems to be populated with all the tasks as false, which I assume is correct

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
DomBelcher: While this may be a slight performance improvent, it's highly unlikely this endpoint is used enough to warrant it

DomBelcher (approved):
Looks good 👍

PR #5245, opened 2024-05-16
PP-12173 use apache5 jersey connector

nlsteers:
## WHAT YOU DID

- use `jersey-apache5-connector` when creating payment provider jersey clients


SandorArpa (approved):
LOKTM

PR #5255, opened 2024-05-20
PP-12355: Unhappy paths for updating a service name

oswaldquek:


src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -338,7 +338,15 @@ public Response updateGatewayAccountServiceNameByServiceId(
...
 
+        if (missingServiceNameField(payload)) {
+            return fieldsMissingResponse(Collections.singletonList(SERVICE_NAME_FIELD_NAME));
+        }
+        
         String serviceName = payload.get(SERVICE_NAME_FIELD_NAME);
+
+        if (invalidServiceNameLength(serviceName)) {
+            return fieldsInvalidSizeResponse(Collections.singletonList(SERVICE_NAME_FIELD_NAME));
+        }
```
DomBelcher: Rather than doing validaiton this way, would it be better to use an `UpdateServiceNameRequest` (or similar) class with `@Valid` to validate the payload? It would mean the errors were returned as `422` instead, which is more appropriate for a syntactically correct but unprocessable request, e.g. due to missing fields

oswaldquek: Sounds good. I'll do it for the `/v1/frontend/service/{serviceId}/{accountType}/servicename` endpoint but will leave `/v1/frontend/accounts/{accountId}/servicename` as is as it will be deprecated at some point in the future.

oswaldquek: If a `UpdateServiceNameRequest` is used, and a `@NotNull` annotation is on the serviceName field, the missing service name check would also go and a 422 also returned if the `service_name` is not provided.

src/test/java/uk/gov/pay/connector/gatewayaccountcredentials/resource/GatewayAccountCredentialsResourceWorldpay3dsFlexIT.java
```diff
@@ -368,15 +368,14 @@ void forInvalidCredentials_shouldReturn200_withResultInvalid() {
...
+            app.givenSetup()
                     .body(toJson(Map.of(
                             "payment_provider", "worldpay",
                             "service_id", SERVICE_ID,
                             "service_name", SERVICE_NAME,
                             "type", "test"
                     )))
-                    .post("/v1/api/accounts")
-                    .then().extract().path("gateway_account_id");
+                    .post("/v1/api/accounts");
```
oswaldquek, line 378: Not related to this PR, but I thought I'd do some cleanup here anyway.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceIT.java
```diff
@@ -215,8 +224,43 @@ void updateGatewayAccountServiceNameSuccessfully() {
...
+                    .statusCode(422)
+                    .body("message", contains("Field [service_name] can have a size between 1 and 50"))
+                    .body("error_identifier", is(ErrorIdentifier.GENERIC.toString()));
+        }
+        
+        @Test
+        void assertNotFoundForNonExistentServiceId() {
+            app.givenSetup().accept(JSON)
+                    .body(Map.of("service_name", "New Service Name"))
+                    .patch(format("/v1/frontend/service/nexiste-pas/test/servicename", serviceId))
```
DomBelcher: This use of format is redundant if the `serviceId` and `accountType` are hardcoded into the URL

DomBelcher (changes requested):
Nice work, just one small change

DomBelcher (approved):
LGTM

PR #5265, opened 2024-05-21
PP-12355: Add update endpoint for a gateway account using Service Id and Account Type

oswaldquek:
Adds the endpoint `PATCH /v1/api/service/{serviceId}/[test|live]`.

This PR contains only happy path tests.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
@@ -1029,13 +1029,12 @@ public class GatewayAccountResourceIT {
     @Nested
     class PatchByServiceIdAndAccountType {
 
-        private Map<String, String> gatewayAccountRequest;
         private String serviceId;
 
         @BeforeEach
         void before() {
             serviceId = RandomIdGenerator.newId();
-            gatewayAccountRequest = Map.of(
+            Map<String, String> gatewayAccountRequest = Map.of(
                     "payment_provider", "worldpay",
                     "service_id", serviceId,
                     "service_name", "Service Name",
@@ -1209,10 +1208,10 @@ public class GatewayAccountResourceIT {
         }
         
         @Test
-        void updateAllowTelephonePaymentNotificationsSuccessfull() {
+        void updateAllowTelephonePaymentNotificationsSuccessfully() {
             app.givenSetup()
                     .get(format("/v1/api/service/%s/test/account", serviceId))
-                    .then().log().body()
+                    .then()
                     .body("allow_telephone_payment_notifications", is(false));
 
             Map<String, Object> payload = Map.of("op", "replace",
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
@@ -1065,7 +1065,7 @@ public class GatewayAccountResourceIT {
                     .statusCode(OK.getStatusCode());
 
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
                     .statusCode(OK.getStatusCode())
                     .body("worldpay_3ds_flex.exemption_engine_enabled", is(true));
@@ -1082,7 +1082,7 @@ public class GatewayAccountResourceIT {
                     .statusCode(OK.getStatusCode());
 
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
                     .statusCode(OK.getStatusCode())
                     .body("worldpay_3ds_flex.exemption_engine_enabled", is(false));
@@ -1091,8 +1091,9 @@ public class GatewayAccountResourceIT {
         @Test
         void updateNotifySettingsSuccessfully() {
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
+                    .statusCode(OK.getStatusCode())
                     .body("notifySettings", nullValue());
 
             Map<String, Object> payload = Map.of("op", "replace",
@@ -1108,7 +1109,7 @@ public class GatewayAccountResourceIT {
                     .statusCode(OK.getStatusCode());
 
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
                     .statusCode(OK.getStatusCode())
                     .body("notifySettings.api_token", is("anapitoken"))
@@ -1125,16 +1126,18 @@ public class GatewayAccountResourceIT {
                     .statusCode(OK.getStatusCode());
 
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
+                    .statusCode(OK.getStatusCode())
                     .body("notifySettings", nullValue());
         }
         
         @Test
         void updateBlockPrepaidCardsSuccessfully() {
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
+                    .statusCode(OK.getStatusCode())
                     .body("block_prepaid_cards", is(false));
 
             Map<String, Object> payload = Map.of("op", "replace",
@@ -1148,16 +1151,18 @@ public class GatewayAccountResourceIT {
                     .statusCode(OK.getStatusCode());
 
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
+                    .statusCode(OK.getStatusCode())
                     .body("block_prepaid_cards", is(true));
         }
         
         @Test
         void updateEmailCollectionModeSuccessfully() {
             app.givenSetup()
-                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .get(format("/v1/api/service/%s/test", serviceId))
                     .then()
+                    .statusCode(OK.getStatusCode())
                     .body("email_collection_mode", is("MANDATORY"));
 
             Map<String, Object> payload = Map.of("op", "replace",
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
-                .body(payload)
-                .patch("/v1/api/accounts/" + gatewayAccountId)
-                .then()
-                .statusCode(OK.getStatusCode());
+            payload = Map.of("op", "remove",
+                    "path", "notify_settings");
 
-        payload = objectMapper.writeValueAsString(Map.of("op", "remove",
-                "path", "notify_settings"));
+            app.givenSetup().body(toJson(payload)).patch(format("/v1/api/service/%s/test/", serviceId));
```
DomBelcher: It would be good to have an assertion for the status code of this API call

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
-        app.givenSetup()
-                .body(payload)
-                .patch("/v1/api/accounts/" + gatewayAccountId)
-                .then()
-                .statusCode(OK.getStatusCode());
-    }
+            app.givenSetup()
+                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .then()
+                    .body("notifySettings", nullValue());
```
DomBelcher, line 1132: Also wouldn't hurt to have an assertion for the status code here - while we can infer that there should be an account returned, checking only for a null value of `notifySettings` doesn't confirm that

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+        }
+        
+        @Test
+        void updateBlockPrepaidCardsSuccessfully() {
+            app.givenSetup()
+                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .then()
+                    .body("block_prepaid_cards", is(false));
```
DomBelcher, line 1141: Less important because the body is inspected here, but again no harm in having an assertion for the status code

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
-                .get("/v1/api/accounts/" + gatewayAccountId)
-                .then()
-                .body("corporate_credit_card_surcharge_amount", is(0))
-                .body("corporate_debit_card_surcharge_amount", is(200))
-                .body("corporate_prepaid_debit_card_surcharge_amount", is(0));
-    }
+            app.givenSetup()
+                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .then()
+                    .body("block_prepaid_cards", is(true));
```
DomBelcher, line 1157: Status code, as above

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+        }
+        
+        @Test
+        void updateEmailCollectionModeSuccessfully() {
```
DomBelcher, line 1161: Is it necessary to test all these operations individually? Could it not be adequately tested with only a couple of happy path tests which test a combination of operations?

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+            app.givenSetup()
+                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .then()
+                    .body("corporate_credit_card_surcharge_amount", is(100))
+                    .body("corporate_debit_card_surcharge_amount", is(200))
+                    .body("corporate_prepaid_debit_card_surcharge_amount", is(400));
+        }
+        
+        @Test
+        void updateAllowTelephonePaymentNotificationsSuccessfull() {
```
DomBelcher: Typo in test name, missing a letter y on the end - although maybe this test could be rolled into one which tests several operations together

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+                    "path", "corporate_debit_card_surcharge_amount",
+                    "value", 200);
+
+            app.givenSetup().body(toJson(payload)).patch(format("/v1/api/service/%s/test/", serviceId));
+
+            payload = Map.of("op", "replace",
+                    "path", "corporate_prepaid_debit_card_surcharge_amount",
+                    "value", 400);
+
+            app.givenSetup().body(toJson(payload)).patch(format("/v1/api/service/%s/test/", serviceId));
```
DomBelcher: Status code, as above

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+                    "path", "corporate_credit_card_surcharge_amount",
+                    "value", 100);
+
+            app.givenSetup().body(toJson(payload)).patch(format("/v1/api/service/%s/test/", serviceId));
+
+            payload = Map.of("op", "replace",
+                    "path", "corporate_debit_card_surcharge_amount",
+                    "value", 200);
+
+            app.givenSetup().body(toJson(payload)).patch(format("/v1/api/service/%s/test/", serviceId));
```
DomBelcher: Status code, as above

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+                    .then()
+                    .body("corporate_credit_card_surcharge_amount", is(0))
+                    .body("corporate_debit_card_surcharge_amount", is(0))
+                    .body("corporate_prepaid_debit_card_surcharge_amount", is(0));
+
+            Map<String, Object> payload = Map.of("op", "replace",
+                    "path", "corporate_credit_card_surcharge_amount",
+                    "value", 100);
+
+            app.givenSetup().body(toJson(payload)).patch(format("/v1/api/service/%s/test/", serviceId));
```
DomBelcher: Status code, as above

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+            app.givenSetup()
+                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .then().log().body()
+                    .body("allow_telephone_payment_notifications", is(false));
+
+            Map<String, Object> payload = Map.of("op", "replace",
+                    "path", "allow_telephone_payment_notifications",
+                    "value", true);
+
+            app.givenSetup().body(toJson(payload)).patch(format("/v1/api/service/%s/test/", serviceId));
```
DomBelcher: Status code, as above

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
```diff
@@ -1051,266 +1026,498 @@ void whenNotificationCredentialsInvalidValues_shouldReturn400() {
...
+                    .body("corporate_credit_card_surcharge_amount", is(100))
+                    .body("corporate_debit_card_surcharge_amount", is(200))
+                    .body("corporate_prepaid_debit_card_surcharge_amount", is(400));
+        }
+        
+        @Test
+        void updateAllowTelephonePaymentNotificationsSuccessfull() {
+            app.givenSetup()
+                    .get(format("/v1/api/service/%s/test/account", serviceId))
+                    .then().log().body()
```
DomBelcher: Delete `.log().body()`

DomBelcher (changes requested):
I think it's worth trying to combine some of the tests together - if the `PATCH` endpoint is supposed to accept multiple updates together we might as well make use of that. 

EDIT: I should have checked before writing this comment that the endpoint did actually accept multiple updates - it seems it doesn't, although other `PATCH` endpoints do - is this something we want to support, or is it better limited to one update per API call?

Status code assertions likely aren't essential everywhere, but I think we should check for `200`s on the `PATCH` endpoint at least, since that's what we're testing.

Also maybe an issue for a separate PR, but `GatewayAccountResourceIT` is becoming quite a large file, maybe worth separating parts of it out into other test files.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
oswaldquek: Yes the reason I haven't done so in a number of places is because I'm already asserting the status code in many many other places already. I didn't think it was worthwhile asserting it on every call to `PATCH /v1/api/service/{serviceId}/[test|live]`.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
oswaldquek: Yeah again I didn't bother with it here because there are many other places where we assert this. And the integration test for `GET /v1/api/service/serviceId/accountType/account` will have covered this off.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
oswaldquek: I did think about this but I decided against it as each of these operations test different json bodies. For example in this test we have the assertions `.body("email_collection_mode", is("MANDATORY"))` and `.body("email_collection_mode", is("OFF"))` which can be unwieldy to parameterise for other operations. 

Where I could reduce the old tests, I have. For example the test `updateNotifySettingsSuccessfully` is a combination of [shouldReturn200_whenNotifySettingsIsUpdated](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java#L1055) and [shouldReturn200_whenNotifySettingsIsRemoved](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java#L1144).

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceIT.java
oswaldquek: Especially for this the status code for the PATCH call has already been tested a few lines above. I'm happy with being doubly sure so I'll make the assertion for the status code for all PATCH calls in this PR.

DomBelcher (approved):
Nice work 👍

PR #5268, opened 2024-05-22
PP-12352 Correct get Gateway Account endpoint

james-peacock-ho:
- remove the `/account` from the get gateway account by service id endpoint


DomBelcher (approved):
LGTM

PR #5287, opened 2024-05-28
PP-12177 refactor imports

SandorArpa:
## WHAT YOU DID

Some imports were transitive, taken from Spotify's docker-client.

- update pay-java-commons to `1.0.20240528083746`
- refactor given imports to use org.apache.commons.lang3


alexbishop1 (approved):
`RandomUtils` is deprecated in recent Commons Lang 3 releases. We may want to heed their advice to use Apache Commons RNG at some point.

PR #5295, opened 2024-05-29
PP-12355: Serialise the POST body of /v1/api/accounts/{accountId}/email-notification to a java class

oswaldquek:
It's easier to use a java class as we can leverage the hibernate validators instead of manually validating the incoming Map<String,String>. This is also in line with how we do validation in other places.

### DIFF
--- a/openapi/connector_spec.yaml
+++ b/openapi/connector_spec.yaml
@@ -3364,6 +3364,19 @@ components:
         version:
           type: integer
           format: int64
+    EmailNotificationPatchRequest:
+      type: object
+      properties:
+        op:
+          type: string
+        path:
+          type: string
+        pathTokens:
+          type: array
+          items:
+            type: string
+        value:
+          type: string
     EpdqCredentials:
       type: object
       properties:
--- a/openapi/connector_spec.yaml
+++ b/openapi/connector_spec.yaml
@@ -3371,10 +3371,6 @@ components:
           type: string
         path:
           type: string
-        pathTokens:
-          type: array
-          items:
-            type: string
         value:
           type: string
     EpdqCredentials:
--- a/src/main/java/uk/gov/pay/connector/usernotification/resource/EmailNotificationResource.java
+++ b/src/main/java/uk/gov/pay/connector/usernotification/resource/EmailNotificationResource.java
@@ -96,7 +96,7 @@ public class EmailNotificationResource {
     }
 
     private NotificationPatchInfo getNotificationInfoFromPath(EmailNotificationPatchRequest emailPatchRequest) {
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/pay/connector/usernotification/resource/EmailNotificationResource.java
```diff
@@ -107,7 +95,7 @@ public Response enableEmailNotification(@Parameter(example = "1", description =
                 .orElseGet(() -> notFoundResponse(format("The gateway account id '%s' does not exist", gatewayAccountId)));
     }
 
-    private NotificationPatchInfo getNotificationInfoFromPath(PatchRequestBuilder.PatchRequest emailPatchRequest) {
+    private NotificationPatchInfo getNotificationInfoFromPath(@Valid EmailNotificationPatchRequest emailPatchRequest) {
```
DomBelcher: Is the `@Valid` annotation here redundant since it's already applied on the `enableEmailNotification` method parameter?

src/test/java/uk/gov/pay/connector/it/resources/EmailNotificationResourceIT.java
```diff
@@ -49,7 +50,7 @@ void patchEmailNotification_shouldNotUpdateIfAccountIdDoesNotExist() {
             app.givenSetup().accept(JSON)
                     .body(getPatchRequestBody("/payment_confirmed/" + EmailNotificationResource.EMAIL_NOTIFICATION_TEMPLATE_BODY, templateBody))
                     .patch(ACCOUNTS_API_URL + nonExistingAccountId + "/email-notification")
-                    .then()
+                    .then().log().body()
```
DomBelcher: Remove the `.log().body()` unless there's a reason for leaving it in

DomBelcher (changes requested):
Really nice, I like this a lot - just a few minor things and it's good

openapi/connector_spec.yaml
```diff
@@ -3364,6 +3364,19 @@ components:
...
+      type: object
+      properties:
+        op:
+          type: string
+        path:
+          type: string
+        pathTokens:
+          type: array
+          items:
+            type: string
```
DomBelcher: I assume `pathTokens` isn't supposed to be part of the request, and this is generated in the spec because of the `getPathTokens` method on `EmailNotificationPatchRequest` - if so is it possible to suppress that?

DomBelcher (changes requested):

DomBelcher (approved):
Nice work 👍

PR #5296, opened 2024-05-29
PP-12678 Handle null postcode when inferring North American region

alexbishop1:
When inferring a US state or Canadian province or territory for a cardholder address, handle there being no postal code. This can happen if a service is using the authorisation API and prefills a billing address without including a postal code (and would previously cause an error if the country was the US or Canada).

Also add some tests to ensure we normalise postal codes (by removing spaces and upper-casing them) correctly.


SandorArpa (approved):
LOKTM

PR #5302, opened 2024-05-31
PP-12355 Refactor AllowedStrings validator with auto-generated error messages

DomBelcher:
## WHAT YOU DID
 - Move `AllowedStrings` validator to utils
 - Add code to generate validation error messages from list of allowed strings and field name


src/main/java/uk/gov/pay/connector/util/validation/AllowedStringsValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    
+    @Override
+    public boolean isValid(String value, ConstraintValidatorContext context) {
+        if (useCustomErrorMessage) {
+            context.disableDefaultConstraintViolation();
+            context.buildConstraintViolationWithTemplate(errorMessage).addConstraintViolation();
+        }
+        
+        if (value == null) return false;
+        return allowedStrings.contains(value);
```
nlsteers, line 43: is `contains()` appropriate for this type of validation? 

image

src/main/java/uk/gov/pay/connector/util/validation/AllowedStringsValidator.java
```diff
@@ -0,0 +1,45 @@
...
+    private String errorMessage;
+    private boolean useCustomErrorMessage = false;
+    @Override
+    public void initialize(AllowedStrings parameters) {
+        allowedStrings = Set.of(parameters.allowed());
+
+        if (parameters.message().isEmpty()) {
+            useCustomErrorMessage = true;
+            switch(allowedStrings.size()) {
+                case 0: errorMessage = String.format("The '%s' field must be empty", parameters.fieldName()); break;
```
nlsteers, line 19: `@AllowedStrings` implies there should be a value, there shouldn't be a case for enforcing empty string. 

Given that `String[] allowed();` does not have a default, would this case ever be satisfied?

nlsteers (changes requested):
honestly, i think you might be overcomplicating this

it's a common pattern with validation annotations to have a `message` element conveying the specifics of the constraint violation. 

no need to try and guess what the message should be if one isn't supplied, i'd remove the default and go back to making `message` required

happy to be talked out of this opinion but it was easier to understand what this validation was doing before

DomBelcher:
Closing as this is an unncessary change

PR #5312, opened 2024-06-04
PP-12355 Clean up StripeAccountSetupResource

DomBelcher:
## WHAT YOU DID
 - Clean up unused imports and fix error messages in StripeAccountSetupResource

## How to test
 - Run tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
@@ -1,6 +1,5 @@
 package uk.gov.pay.connector.gatewayaccount.resource;
 
-import com.fasterxml.jackson.databind.JsonNode;
 import com.google.inject.Inject;
 import io.dropwizard.jersey.PATCH;
 import io.swagger.v3.oas.annotations.Operation;
@@ -17,7 +16,6 @@ import uk.gov.pay.connector.gatewayaccount.model.StripeAccountSetupUpdateRequest
 import uk.gov.pay.connector.gatewayaccount.model.StripeSetupPatchRequest;
 import uk.gov.pay.connector.gatewayaccount.service.GatewayAccountService;
 import uk.gov.pay.connector.gatewayaccount.service.StripeAccountSetupService;
-import uk.gov.service.payments.commons.model.jsonpatch.JsonPatchRequest;
 
 import javax.validation.Valid;
 import javax.ws.rs.Consumes;
@@ -29,7 +27,6 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.core.Response;
 import java.util.List;
 import java.util.stream.Collectors;
-import java.util.stream.StreamSupport;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
```diff
@@ -83,13 +77,15 @@ public StripeAccountSetup getStripeAccountSetupByServiceIdAndAccountType(
             @Parameter(example = "46eb1b601348499196c99de90482ee68", description = "Service ID") @PathParam("serviceId") String serviceId, 
             @Parameter(example = "test", description = "Account type") @PathParam("accountType") GatewayAccountType accountType) {
         return gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, accountType)
+                .or(() -> { throw new GatewayAccountNotFoundException(serviceId, accountType); })
                 .map(gatewayAccountEntity -> stripeAccountSetupService.getCompletedTasks(gatewayAccountEntity.getId()))
-                .orElseThrow(() -> new GatewayAccountNotFoundException(serviceId, accountType));
+                .orElseThrow(() -> new IllegalStateException("Internal Server Error"));
```
james-peacock-ho, line 82: is this line actually necessary? I'm not sure how it could be triggered.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
```diff
@@ -130,6 +126,7 @@ public Response patchStripeAccountSetup(
     @PATCH
     @Path("/v1/api/service/{serviceId}/account/{accountType}/stripe-setup")
     @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
```
james-peacock-ho, line 129: do we need this when the response is `Response.ok()`?

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
```diff
@@ -148,23 +145,25 @@ public Response patchStripeAccountSetup(
...
                 .map(gatewayAccountEntity -> {
                     List<StripeAccountSetupUpdateRequest> updateRequests = request.stream()
                             .map(StripeAccountSetupUpdateRequest::from)
                             .collect(Collectors.toList());
 
                     stripeAccountSetupService.update(gatewayAccountEntity, updateRequests);
                     return Response.ok().build();
-                }).orElseThrow(NotFoundException::new);
+                })
+                .orElseThrow(() -> new IllegalStateException("Internal Server Error"));
```
james-peacock-ho, line 166: again I'm not sure about the scenario that causes this exception. Does this cover the possibility of the gateway account not actually being a stripe account?

james-peacock-ho (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
DomBelcher: It causes an issue with the `GatewayAccountNotFoundException` if it's not there

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
DomBelcher: It shouldn't ever happen, it's just to keep the compiler happy that the possibility of the `Optional` being empty is handled and to return the response from the `Optional`. It may be clearer to put the `GatewayAccountNotFoundException` there instead, the reason I put it directly after `getGatewayAccountByServiceIdAndAccountType` was so it was clear that was the only way it could be triggered. There's more complex flows where this is necessary, so I thought it would be sensible to be consistent with those even if it's not required

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountSetupResource.java
DomBelcher: Same as above - it's only to keep the compiler happy and isn't strictly necessary to do it this way.

james-peacock-ho (approved):
LGTM!

PR #5341, opened 2024-06-14
PP-12358 Add degateway endpoint for charge capture

james-peacock-ho:
- Add endpoint to capture charge without gateway account id

### DIFF
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
@@ -256,6 +256,29 @@ public class CardResource {
     }
 
     @POST
+    @Path("/v1/api/charges/{chargeId}/capture")
+    @Produces(APPLICATION_JSON)
+    @Operation(
+            summary = "Mark delayed capture charge as eligible for capture and adds charge to capture queue",
+            description = "This endpoint should be called to capture a delayed capture charge. The charge needs to have been previously marked as AWAITING CAPTURE REQUEST for this call to succeed. " +
+                    "When a charge is in any of the states CAPTURED, CAPTURE APPROVED, CAPTURE APPROVED RETRY, CAPTURE READY, CAPTURE SUBMITTED then nothing happens and the response will be a 204. " +
+                    "When a charge is in a status that cannot transition (eg. none of the above) then 409 response is returned. ",
+            responses = {
+                    @ApiResponse(responseCode = "204", description = "No content"),
+                    @ApiResponse(responseCode = "409", description = "Conflict - if charge is not in correct state"),
+                    @ApiResponse(responseCode = "404", description = "Not found - charge not found"),
+                    @ApiResponse(responseCode = "500", description = "Internal server error")
+            }
+    )
+    public Response markChargeAsCaptureApproved(@PathParam("chargeId") String chargeId,
+                                                @Context UriInfo uriInfo) {
+        logger.info("Mark charge as CAPTURE APPROVED [charge_external_id={}]", chargeId);
+        delayedCaptureService.markDelayedCaptureChargeAsCaptureApproved(chargeId);
+        return ResponseUtil.noContentResponse();
+    }
+
+
+    @POST
     @Path("/v1/api/accounts/{accountId}/charges/{chargeId}/cancel")
     @Produces(APPLICATION_JSON)
     @Operation(
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
@@ -270,7 +270,7 @@ public class CardResource {
                     @ApiResponse(responseCode = "500", description = "Internal server error")
             }
     )
-    public Response markChargeAsCaptureApproved(@PathParam("chargeId") String chargeId,
+    public Response markChargeAsCaptureApprovedByChargeId(@PathParam("chargeId") String chargeId,
                                                 @Context UriInfo uriInfo) {
         logger.info("Mark charge as CAPTURE APPROVED [charge_external_id={}]", chargeId);
         delayedCaptureService.markDelayedCaptureChargeAsCaptureApproved(chargeId);
--- a/src/test/java/uk/gov/pay/connector/it/resources/CardResourceCaptureWithSqsQueueIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/CardResourceCaptureWithSqsQueueIT.java
@@ -38,7 +38,6 @@ public class CardResourceCaptureWithSqsQueueIT {
 
     private Appender<ILoggingEvent> mockAppender = mock(Appender.class);
     private ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
-    private String captureApproveUrl;
     
     @BeforeEach
     void setUpLogger() {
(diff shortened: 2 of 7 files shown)


src/main/java/uk/gov/pay/connector/paymentprocessor/resource/CardResource.java
```diff
@@ -255,6 +255,29 @@ public Response markChargeAsCaptureApproved(@PathParam("accountId") Long account
...
+                    "When a charge is in any of the states CAPTURED, CAPTURE APPROVED, CAPTURE APPROVED RETRY, CAPTURE READY, CAPTURE SUBMITTED then nothing happens and the response will be a 204. " +
+                    "When a charge is in a status that cannot transition (eg. none of the above) then 409 response is returned. ",
+            responses = {
+                    @ApiResponse(responseCode = "204", description = "No content"),
+                    @ApiResponse(responseCode = "409", description = "Conflict - if charge is not in correct state"),
+                    @ApiResponse(responseCode = "404", description = "Not found - charge not found"),
+                    @ApiResponse(responseCode = "500", description = "Internal server error")
+            }
+    )
+    public Response markChargeAsCaptureApproved(@PathParam("chargeId") String chargeId,
```
nlsteers: if you renamed this to something like 
```suggestion
    public Response markChargeAsCaptureApprovedByChargeId(@PathParam("chargeId") String chargeId,
```
you can avoid the `_1` in the openapi spec for the operation ID

src/test/java/uk/gov/pay/connector/it/resources/CardResourceCaptureWithSqsQueueIT.java
```diff
@@ -65,13 +64,33 @@ void shouldAddChargeToQueueAndSubmitForCapture_IfChargeWasPreviouslyAuthorised()
     }
 
     @Test
-    void shouldAddChargeToQueueAndSubmitForCapture_IfChargeWasAwaitingCapture() {
+    void shouldAddChargeToQueueAndSubmitForCapture_IfChargeWasAwaitingCapture_whenCaptureByChargeIdAndAccountId() {
```
nlsteers: should this be nested / parameterised if we're sticking with established test convention for 'degatewayed' endpoints?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceIT.java
```diff
@@ -712,6 +713,79 @@ void shouldNotReturnCanRetryWhenChargeHasAuthorisationModeWeb() {
...
     }
+    
+    @Nested
+    class CaptureByChargeIdOnly {
+        @Test
+        void shouldGetNoContentForMarkChargeAsCaptureApproved_withStatus_awaitingCaptureRequest() {
+            String extChargeId = testBaseExtension.addChargeAndCardDetails(AWAITING_CAPTURE_REQUEST,
+                    ServicePaymentReference.of("ref"), Instant.now().minus(90, MINUTES));
+
+            testBaseExtension.getConnectorRestApiClient()
```
nlsteers, line 624: in other ITests we've been using `app.givenSetup().body().[operation]().then()` rather than the rest assured client in the ITestBaseExtension, is there any appreciable difference?

nlsteers (commented):
looks mostly fine to me, few questions and one suggestion

src/test/java/uk/gov/pay/connector/it/resources/CardResourceCaptureWithSqsQueueIT.java
james-peacock-ho: yes I did wonder. I thought `@Nested` was probably more appropriate when you have multiple tests in each nested category. In this case there is only one test for each of three endpoints. But could change it to nested to avoid having such a long method name?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceIT.java
james-peacock-ho: as far as I can tell there's no difference, I just thought on balance it made sense to be consistent with the other tests in this class.

src/test/java/uk/gov/pay/connector/it/resources/CardResourceCaptureWithSqsQueueIT.java
james-peacock-ho: have made this change now.

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceIT.java
james-peacock-ho: I've made the change as suggested.

src/test/java/uk/gov/pay/connector/it/resources/CardResourceCaptureWithSqsQueueIT.java
```diff
@@ -47,40 +47,69 @@ void setUpLogger() {
...
-    void shouldAddChargeToQueueAndSubmitForCapture_IfChargeWasAwaitingCapture() {
-        String chargeId = testBaseExtension.addCharge(AWAITING_CAPTURE_REQUEST, "ref", Instant.now().minus(48, HOURS).plus(1, MINUTES), RandomIdGenerator.newId());
+    @Nested
+    class DelayedCaptureApproveByChargeIdAndAccountId {
+        @Test
+        void shouldAddChargeToQueueAndSubmitForCapture_IfChargeWasAwaitingCapture_whenCaptureByChargeIdAndAccountId() {
+            String chargeId = testBaseExtension.addCharge(AWAITING_CAPTURE_REQUEST, "ref", Instant.now().minus(48, HOURS).plus(1, MINUTES), RandomIdGenerator.newId());
 
-        captureApproveUrl = ITestBaseExtension.captureUrlForAwaitingCaptureCharge(testBaseExtension.getAccountId(), chargeId);
+            String captureApproveByChargeIdAndAccountIdUrl = ITestBaseExtension.captureUrlByChargeIdAndAccountIdForAwaitingCaptureCharge(testBaseExtension.getAccountId(), chargeId);
```
nlsteers: given this is a static method and is only used once in this test, i'd just qualify the static import over assigning to a local var

src/test/java/uk/gov/pay/connector/util/RestAssuredClient.java
```diff
@@ -210,6 +210,14 @@ public ValidatableResponse postMarkChargeAsCaptureApproved() {
                 .then();
     }
 
+    public ValidatableResponse postMarkChargeAsCaptureApprovedByChargeId() {
```
nlsteers: is this used anywhere?

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceIT.java
```diff
@@ -457,7 +459,7 @@ void shouldGetNoContentForMarkChargeAsCaptureApproved_withStatus_awaitingCapture
         testBaseExtension.getConnectorRestApiClient()
                 .withAccountId(accountId)
                 .withChargeId(extChargeId)
-                .postMarkChargeAsCaptureApproved()
+                .postMarkChargeAsCaptureApprovedByChargeIdAndAccountId()
```
nlsteers: how much work would it be to put these under their own nested block for `ChargeIdAndGatewayAccountId` in addition to matching the style you've introduced in the `DelayedCaptureApproveByChargeIdOnly` tests?

src/test/java/uk/gov/pay/connector/util/RestAssuredClient.java
james-peacock-ho: not any more, good spot

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceIT.java
james-peacock-ho: Nested block added.

nlsteers (approved):

PR #5342, opened 2024-06-14
PP-12703: Delete /v1/api/account/{accountId}/notification-credentials

oswaldquek:
During the process of migrating this API to a serviceId and accountType variant, it was discovered that [NotificationCredentials are for Smartpay accounts
only](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountWithCredentialsResponse.java#L22). The associated classes and endpoint for this NotificationCredentials were created from https://payments-platform.atlassian.net/browse/PP-719 which shows that these classes were created for Smartpay functionality. Given we no longer have any Smartpay gateway accounts the `POST
/v1/api/account/{accountId}/notification-credentials` can be deleted.


james-peacock-ho (changes requested):
I think the commit and PR name has a typo in the ticket number, should be PP-12703 - might be worth fixing just for the audit trail

james-peacock-ho (approved):

PR #5347, opened 2024-06-17
PP-12703: Move existing switch psp integration tests

oswaldquek:
This is to make later PR reviewing (for switching psp by service id and account type) easier.


james-peacock-ho (approved):
All looks good.

PR #5350, opened 2024-06-18
PP-12703: Test for successfully switching PSP by service id and account type

oswaldquek:
This commit migrates the old ByGatewayAccountId.shouldSwitchPaymentProvider test to switch by service id and account type. The old test asserts the integration_version_3ds is 2; however existing code in the GatewayAccountSwitchPaymentProviderService does not set the integration_version_3ds to 2. Historically, the integration_version_3ds should be set to 2 when switching from Worldpay to Stripe (see acceptance criteria on PP-8191); however this has been superseded by PP-11591 where a Worldpay account will always have integration_version_3ds set to 2. Therefore the old assertion is no longer necessary.

Also unlike many other tests, the new
ByServiceIdAndAccountType.switchPaymentProviderFromWorldpayToStripeSuccessfully test sets up test data using APIs only (as opposed to setting up state by directly inserting data into the database). I'm not sure using APIs only should always be done, but at least we now have one place where anyone can easily understand how the state for a switching psp scenario is set up from an API point of view.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -76,6 +76,10 @@
       "path": "detect_secrets.filters.allowlist.is_line_allowlisted"
     },
     {
+      "path": "detect_secrets.filters.common.is_baseline_file",
+      "filename": ".secrets.baseline"
+    },
+    {
       "path": "detect_secrets.filters.common.is_ignored_due_to_verification_policies",
       "min_level": 2
     },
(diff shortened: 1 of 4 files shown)


james-peacock-ho (approved):
Looks good to me.
I assume there will also be some unhappy path testing for this endpoint in a subsequent PR.

PR #5351, opened 2024-06-18
PP-12358 degateway cancel charge endpoint

nlsteers:
## WHAT YOU DID

- de-gateway cancel charge endpoint
  - add gateway account service as injectable dependency to card resource class
- add missing cancel charge resource unit tests

### DIFF
--- a/src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
+++ b/src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
@@ -30,7 +30,7 @@ public class CardResourceCancelChargeIT {
     private static final AppWithPostgresAndSqsExtension app = new AppWithPostgresAndSqsExtension();
 
     private static final String SERVICE_ID = RandomIdGenerator.randomUuid();
-    private static final String SERVICE_NAME = "";
+    private static final String SERVICE_NAME = "Buy a Civil Servant a Coffee";
     private static final GatewayAccountType GATEWAY_ACCOUNT_TYPE = GatewayAccountType.TEST;
     private static final PaymentGatewayName PAYMENT_GATEWAY_NAME = PaymentGatewayName.STRIPE;
 
--- a/src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
+++ b/src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
@@ -1,0 +1,204 @@
+package uk.gov.pay.connector.paymentprocessor.resource;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.extension.AppWithPostgresAndSqsExtension;
+import uk.gov.pay.connector.gateway.PaymentGatewayName;
+import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType;
+import uk.gov.pay.connector.it.util.ChargeUtils;
+import uk.gov.pay.connector.util.RandomIdGenerator;
+
+import java.util.Map;
+import java.time.Duration;
+
+import static java.lang.Integer.MAX_VALUE;
+import static java.lang.String.format;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.is;
+import static org.testcontainers.shaded.org.awaitility.Awaitility.await;
+import static uk.gov.pay.connector.it.util.ChargeUtils.createChargePostBody;
+import static uk.gov.pay.connector.util.JsonEncoder.toJson;
+
+public class CardResourceCancelChargeIT {
+
+    @RegisterExtension
+    private static final AppWithPostgresAndSqsExtension app = new AppWithPostgresAndSqsExtension();
+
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
```diff
@@ -0,0 +1,204 @@
...
+import static uk.gov.pay.connector.it.util.ChargeUtils.createChargePostBody;
+import static uk.gov.pay.connector.util.JsonEncoder.toJson;
+
+public class CardResourceCancelChargeIT {
+
+    @RegisterExtension
+    private static final AppWithPostgresAndSqsExtension app = new AppWithPostgresAndSqsExtension();
+
+    private static final String SERVICE_ID = RandomIdGenerator.randomUuid();
+    private static final String SERVICE_NAME = "";
```
james-peacock-ho: is it necessary to pass in a service name? If so, maybe we should use something like "A service name" ? If not necessary, maybe we can leave it out?

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
```diff
@@ -0,0 +1,204 @@
...
+            @DisplayName("Should return 204 when successful")
+            void success_shouldReturn204() {
+                var chargeId = createNewCharge();
+                app.givenSetup()
+                        .body("")
+                        .post(String.format("/v1/api/accounts/%s/charges/%s/cancel", gatewayAccountId, chargeId))
+                        .then().statusCode(204);
+
+                await()
+                        .pollDelay(Duration.ofSeconds(1))
```
james-peacock-ho, line 70: What are we waiting for here? If it's just a case of needing to give time for the db to be updated, i'm wondering if this is missing from tests that I've done previously. Is there something particular about the cancel function that means it will take more time than other changes? Just wondering if 1 second is more than necessary in this context?

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
```diff
@@ -0,0 +1,204 @@
...
+            @DisplayName("Should return 204 when successful")
+            void success_shouldReturn204() {
+                var chargeId = createNewCharge();
+                app.givenSetup()
+                        .body("")
+                        .post(String.format("/v1/api/service/%s/account/%s/charges/%s/cancel", SERVICE_ID, GATEWAY_ACCOUNT_TYPE, chargeId))
+                        .then().statusCode(204);
+
+                await()
+                        .pollDelay(Duration.ofSeconds(1))
```
james-peacock-ho, line 143: as above

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeTest.java
```diff
@@ -0,0 +1,295 @@
...
+    @DisplayName("Given a gateway account id")
+    class ByAccountId {
+
+        @Nested
+        @DisplayName("Then cancel charge")
+        class CancelCharge {
+
+            @Test
+            @DisplayName("Should return 404 if charge is not found")
+            void chargeNotFound_shouldReturn404() {
```
james-peacock-ho, line 104: with these tests, what coverage they are giving us beyond what is already in the integration tests?

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
```diff
@@ -0,0 +1,204 @@
...
+                        "type", GATEWAY_ACCOUNT_TYPE,
+                        "payment_provider", PAYMENT_GATEWAY_NAME.getName(),
+                        "service_name", SERVICE_NAME
+                )))
+                .post("/v1/api/accounts")
+                .then().extract().path("gateway_account_id");
+    }
+
+    @Nested
+    @DisplayName("Given a gateway account id")
```
james-peacock-ho, line 53: If this is valuable for clarity then we should consider doing it for all tests. Would be interested to know what it's giving us that isn't clear from the class names and test names?

james-peacock-ho (changes requested):

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
nlsteers: yes good spot, was trying to think of something witty and forgot to populate the string 😅

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
nlsteers: 1 second is enough I think, I'm usually against having any kind of wait in a test but this is an effort to reduce flakiness in ITests that test endpoints where some async processing happens in the background (e.g. charge state transition)

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeIT.java
nlsteers: I'd argue that human readable natural language is always clearer than `shouldDoX_whenY`

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceCancelChargeTest.java
nlsteers: not a whole lot in terms of coverage (aside from testing database interactions) but there's a benefit to having both given unit tests are much faster to run. Improves devex by upping feedback time, often ITests are only run on CI at the point of PR

james-peacock-ho (approved):
LGTM

PR #5355, opened 2024-06-19
PP-12703:  Unsuccessful scenario tests for switching psp

oswaldquek:
This commit creates a mirror of existing switch PSP tests by account id so there are equivalent tests for service id and account type. It also makes the GatewayAccountSwitchPaymentProviderRequest use validation constraints prevalent in other java classes representing POST bodies. This means we don't need the   GatewayAccountSwitchPaymentProviderRequestValidator to do validations anymore.


src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceSwitchPspValidationTest.java
```diff
@@ -70,15 +74,180 @@ void shouldReturn400WhenNoActiveCredentialFound() throws JsonProcessingException
...
+            Response response = resources.client()
+                    .target(format("/v1/api/service/%s/account/test/switch-psp", serviceId))
+                    .request()
+                    .post(Entity.json(Map.of("user_external_id", "some-user-external-id",
+                            "gateway_account_credential_external_id", switchToExtId)));
+
+            assertThat(response.getStatus(), is(400));
+
+            String errorMessage = response.readEntity(JsonNode.class).get("message").get(0).textValue();
+            assertThat(errorMessage, is("Credential with id [" + switchToExtId + "] is not in correct state."));
```
DomBelcher: Would it be useful for this error message to say what the correct state should be? From the code it seems to be `VERIFIED_WITH_LIVE_PAYMENT` is the only acceptable state

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
```diff
@@ -641,7 +641,7 @@ private List<String> extractNotFoundCardTypeIds(List<UUID> cardTypeIds, List<Car
     public Response switchPaymentProvider(
```
DomBelcher: I know it's from an earlier PR but overloading the `switchPaymentProvider` method causes a `switchPaymentProvider_1` method to be generated in the spec for one of the two. It's a minor point but it's a little easier to find in the code if the new method is has a different name, e.g. `switchPaymentProviderByServiceIdAndType` - it is more verbose but generally this is in line with the other degatewayed endpoints

DomBelcher (commented):
Nice work, using the validation API over a validator class makes sense

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResource.java
oswaldquek: Yes agreed

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceSwitchPspValidationTest.java
oswaldquek: This rings a bell from what I [discovered earlier this week](https://gds.slack.com/archives/CAD6G150R/p1718633645550179?thread_ts=1718630373.406099&cid=CAD6G150R). I'll amend the error message to say "Credential with id [the-id] is not in the VERIFIED_WITH_LIVE_PAYMENT state."

src/test/java/uk/gov/pay/connector/gatewayaccount/resource/GatewayAccountResourceSwitchPspValidationTest.java
```diff
@@ -112,7 +279,7 @@ void shouldReturn400WhenCredentialIsMissing() throws JsonProcessingException {
         }
 
         @Test
-        void shouldReturn400WhenCredentialsNonExistent() throws JsonProcessingException {
+        void shouldReturn400WhenCredentialsNonExistent() {
```
DomBelcher: I know this is old code, but this actually returns a 404

DomBelcher (approved):
Nice work 👍

PR #5358, opened 2024-06-19
PP-12359 Add degatewayed endpoint for charge events

james-peacock-ho:
- Add new endpoint to get charge events by service id and account type
- Add integration tests and refactor some existing tests to improve readability


src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -38,163 +39,272 @@ public class ChargeEventsResourceIT {
...
+            DatabaseFixtures.TestChargeEvent enteringCardDetailsTestChargeEvent = createTestChargeEvent(testCharge)
+                    .withChargeStatus(ENTERING_CARD_DETAILS).withDate(createdDate.plusSeconds(1)).insert();
+            createTestChargeEvent(testCharge)
+                    .withChargeStatus(AUTHORISATION_READY).withDate(createdDate.plusSeconds(2)).insert();
+            DatabaseFixtures.TestChargeEvent captureApprovedTestChargeEvent = createTestChargeEvent(testCharge)
+                    .withChargeStatus(CAPTURE_APPROVED).withDate(createdDate.plusSeconds(3)).insert();
+            createCaptureEvents(testCharge, createdDate);
+
+            app.givenSetup()
+                    .get(format(CHARGE_EVENTS_BY_SERVICE_ID_URL, SERVICE_ID, GatewayAccountType.TEST, CHARGE_ID))
```
oswaldquek: No biggie but I personally it easier to read if "/v1/api/accounts/%s/charges/%s/events" were inlined here.

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -38,163 +39,272 @@ public class ChargeEventsResourceIT {
...
+            DatabaseFixtures.TestCharge testCharge = createTestCharge(CHARGE_ID).insert();
+            DatabaseFixtures.TestChargeEvent createdTestChargeEvent = createTestChargeEvent(testCharge)
+                    .withChargeStatus(CREATED).withDate(createdDate).insert();
+            DatabaseFixtures.TestChargeEvent enteringCardDetailsTestChargeEvent = createTestChargeEvent(testCharge)
+                    .withChargeStatus(ENTERING_CARD_DETAILS).withDate(createdDate.plusSeconds(1)).insert();
+            createTestChargeEvent(testCharge)
+                    .withChargeStatus(AUTHORISATION_READY).withDate(createdDate.plusSeconds(2)).insert();
+            DatabaseFixtures.TestChargeEvent captureApprovedTestChargeEvent = createTestChargeEvent(testCharge)
+                    .withChargeStatus(CAPTURE_APPROVED).withDate(createdDate.plusSeconds(3)).insert();
+            createCaptureEvents(testCharge, createdDate);
```
oswaldquek: I think it'd make it easier to understand if the events created in this method, and the AUTHORISATION_READY event, were all put into one method called `addFutureEvents`, so it's easily understood those wouldn't be returned by the API call. Because presumably future events shouldn't be returned.

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -38,163 +39,272 @@ public class ChargeEventsResourceIT {
...
```
oswaldquek: same comment as above

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -38,163 +39,272 @@ public class ChargeEventsResourceIT {
...
+        public void shouldGetAllEventsForAGivenChargeWithRefunds() {
+            ZonedDateTime createdDate = ZonedDateTime.now();
+            app.getDatabaseTestHelper().addGatewayAccount(gatewayAccountParams);
+            DatabaseFixtures.TestCharge testCharge = createTestCharge(CHARGE_ID).insert();
+            DatabaseFixtures.TestChargeEvent createdTestChargeEvent = createTestChargeEvent(testCharge)
+                    .withChargeStatus(CREATED).withDate(createdDate).insert();
+            DatabaseFixtures.TestChargeEvent enteringCardDetailsTestChargeEvent = createTestChargeEvent(testCharge)
+                    .withChargeStatus(ENTERING_CARD_DETAILS).withDate(createdDate.plusSeconds(1)).insert();
+            createTestChargeEvent(testCharge)
+                    .withChargeStatus(AUTHORISATION_READY).withDate(createdDate.plusSeconds(2)).insert();
```
oswaldquek: There's a massive amount of test setup here; I'd get rid of this line as it's already tested in shouldGetAllEventsForAGivenChargeWithoutRefunds and they are not events that will be returned by a call to the API.

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -38,163 +39,272 @@ public class ChargeEventsResourceIT {
...
+                    .withChargeExternalId(CHARGE_ID)
+                    .insert();
+    
+            ZonedDateTime historyRefund1SubmittedStartDate = createdDate.plusSeconds(8);
+            createTestRefundHistory(testRefund1)
+                    .insert(RefundStatus.CREATED, createdDate.plusSeconds(7), historyRefund1SubmittedStartDate, SUBMITTED_BY, USER_EMAIL)
+                    .insert(RefundStatus.REFUND_SUBMITTED, gatewayTransactionIdForRefund1, historyRefund1SubmittedStartDate, testRefund1CreatedDate, SUBMITTED_BY, USER_EMAIL)
+                    .insert(RefundStatus.REFUNDED, gatewayTransactionIdForRefund1, testRefund1CreatedDate, SUBMITTED_BY, testCharge.getExternalChargeId());
+    
+            ZonedDateTime historyRefund2SubmittedStartDate = createdDate.plusSeconds(11);
```
oswaldquek: Hmm why are all these refunds returned, given their start dates are in the future, i.e. `createdDate.plusSeconds()`?

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -38,163 +39,272 @@ public class ChargeEventsResourceIT {
...
+                    .contentType(JSON)
+                    .statusCode(NOT_FOUND.getStatusCode())
+                    .body("message", contains(format("Charge with id [%s] not found.", CHARGE_ID)))
+                    .body("error_identifier", is(ErrorIdentifier.GENERIC.toString()));
+        }
+    
+        @Test
+        public void shouldReturn404WhenChargeIdDoesNotExist() {
+            app.givenSetup()
+                    .get(format(CHARGE_EVENTS_BY_SERVICE_ID_URL, SERVICE_ID, GatewayAccountType.TEST, "non-existent-charge"))
```
oswaldquek: Hmm where is a gateway account with SERVICE_ID added in the database?

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -38,163 +39,272 @@ public class ChargeEventsResourceIT {
...
+                    .body("events[2]", new TransactionEventMatcher("PAYMENT", withState("success", "true"), "100", captureApprovedTestChargeEvent.getUpdated()))
+                    .body("events[3]", new TransactionEventMatcher("REFUND", withState("submitted", "false"), "10", historyRefund1SubmittedStartDate, gatewayTransactionIdForRefund1, SUBMITTED_BY))
+                    .body("events[4]", new TransactionEventMatcher("REFUND", withState("success", "true"), "10", testRefund1CreatedDate, gatewayTransactionIdForRefund1, SUBMITTED_BY))
+                    .body("events[5]", new TransactionEventMatcher("REFUND", withState("submitted", "false"), "90", historyRefund2SubmittedStartDate, gatewayTransactionIdForRefund2, null))
+                    .body("events[6]", new TransactionEventMatcher("REFUND", withState("success", "true"), "90", testRefund2CreatedDate, gatewayTransactionIdForRefund2, null));
+        }
+
+        @Test
+        public void shouldReturn404WhenServiceIdIsIncorrect() {
+            app.getDatabaseTestHelper().addGatewayAccount(gatewayAccountParams);
```
oswaldquek: Since at a high level we're moving away from gateway accounts, I think it'd be better to follow a set up like [this](https://github.com/alphagov/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceUpdateIT.java#L38-L49). That way it's easily seen a gateway account was created with a service Id.

oswaldquek (changes requested):
I think the tests could be structured in a way that makes understanding easier.

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
james-peacock-ho: I think the fact that the events are in the future is a red herring, because all of the relevant events will be returned regardless of date. For realism I could set the start time to yesterday so all of them are in the past.

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
james-peacock-ho: updated

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
james-peacock-ho: there is no filter on dates so in fact future-dated events would be returned but to avoid any confusion i've set the charge creation and associated events to yesterday's date so all events are in the past.

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
james-peacock-ho: I've done what I can to reduce the word count here by having shared methods for creating the required test entities, and comments and make it clear what is happening.

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
james-peacock-ho: as far as I can tell there is nothing to prevent future dates being returned

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
james-peacock-ho: this is now added in the API call to create the account

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
james-peacock-ho: done

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -30,189 +35,292 @@
...
-
-        ZonedDateTime refundTest2RefundedDate = createdDate.plusSeconds(12);
-        DatabaseFixtures.TestRefund refundedTestRefund2 = createTestRefund(testCharge)
-                .withAmount(90L)
-                .withType(RefundStatus.REFUNDED)
-                .withGatewayTransactionId(testGatewayReferenceRefund2)
-                .withCreatedDate(refundTest2RefundedDate)
-                .withChargeExternalId(testCharge.getExternalChargeId())
+    void setup() {
+        //create the gateway account via the API
```
oswaldquek, line 58: This comment probably isn't needed

src/test/java/uk/gov/pay/connector/it/resources/ChargeEventsResourceIT.java
```diff
@@ -30,189 +35,292 @@
...
+                        "payment_provider", PaymentGatewayName.STRIPE.getName(),
+                        "service_name", "MyTestService"
+                )))
+                .post("/v1/api/accounts")
+                .then()
+                .statusCode(201)
+                .extract().path("gateway_account_id");
+
+
+        //create a £1 payment for that account
```
oswaldquek, line 72: This comment probably isn't needed

oswaldquek (approved):
Some comments aren't needed, but it's not worth waiting another build for. 
The other comments are really useful and helps me understand how the system works better!

PR #5380, opened 2024-06-26
bau: Code cleanup

oswaldquek:
Delete unused code and use more succinct java 21 switch statement.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
@@ -32,13 +32,11 @@ import uk.gov.pay.connector.events.model.refund.RefundCreatedByService;
 import uk.gov.pay.connector.events.model.refund.RefundCreatedByUser;
 import uk.gov.pay.connector.events.model.refund.RefundError;
 import uk.gov.pay.connector.events.model.refund.RefundEvent;
-import uk.gov.pay.connector.gateway.PaymentProviders;
 import uk.gov.pay.connector.queue.statetransition.PaymentStateTransition;
 import uk.gov.pay.connector.queue.statetransition.RefundStateTransition;
 import uk.gov.pay.connector.queue.statetransition.StateTransition;
 import uk.gov.pay.connector.refund.dao.RefundDao;
 import uk.gov.pay.connector.refund.model.domain.RefundHistory;
-import uk.gov.pay.connector.refund.service.RefundService;
 
 import javax.inject.Inject;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
```diff
@@ -75,24 +71,19 @@ public class EventFactory {
...
-            PaymentStateTransition paymentStateTransition = (PaymentStateTransition) stateTransition;
-            return createPaymentEvents(paymentStateTransition);
-        } else if (stateTransition instanceof RefundStateTransition) {
-            RefundStateTransition refundStateTransition = (RefundStateTransition) stateTransition;
-            return createRefundEvents(refundStateTransition);
-        } else {
-            throw new EventCreationException(stateTransition.getIdentifier(), "Failed to create StateTransition event because event is not an instance of PaymentStateTransition or RefundStateTransition");
-        }
+        return switch (stateTransition) {
+            case PaymentStateTransition ignored -> createPaymentEvents((PaymentStateTransition) stateTransition);
```
jamesdp10, line 82: I was a bit thrown by 'ignored' but I guess this just means that you don't need to use this particular reference to the stateTransition. I wonder if there is a convention for naming these unused references

src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
```diff
@@ -75,24 +71,19 @@ public class EventFactory {
...
+            case RefundStateTransition ignored -> createRefundEvents((RefundStateTransition) stateTransition);
+            default -> throw new EventCreationException(stateTransition.getIdentifier(), 
+                    "Failed to create StateTransition event because event is not an instance of PaymentStateTransition or RefundStateTransition");
```
jamesdp10, line 85: perhaps trivial, but I don't think this logic is tested anywhere, I couldn't find a reference to this exception statement in tests.

jamesdp10 (approved):
a couple of minor comments but looks good to me

src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
oswaldquek: I would use underscore here but it's still in preview feature 😢

src/main/java/uk/gov/pay/connector/events/model/EventFactory.java
oswaldquek: This is here because the compiler forces us to deal with a default case here, and as such I don't think it needs to be tested. Even with the previous `if...else` statement the compiler forces us to have a final `else` block.

james-peacock-ho (approved):
LGTM

PR #5384, opened 2024-06-27
PP-12819 Increase capture readTimeout duration

kbottla:
## WHAT YOU DID
- We are seeing an increase in the number of `gateway connection timeout` errors for all Worldpay operations other than authorise (since upgrading to Dropwizard 3). This is because all ops have `readTimeout` configured between 1 and 2 seconds whereas for authorisation it is 50 seconds.

- Increased `readTimeout` config for capture operations and if this addresses the issue, the same can be done for cancel & refund
     - There should be no impact as captures are done by background processes with a retry mechanism.


marcotranchino (approved):
LGTM

PR #5385, opened 2024-06-27
PP-12819 Increase readTimeout for refund and cancel ops

kbottla:
## WHAT
- Increased the `readTimeout` for Worldpay refund and cancel ops to address the increase in `gateway connection timeout` errors.
- Most operations complete in under a second and very few (about 10 refunds & under 10 cancels - a day) which have been timing out will take a couple of seconds longer to complete the op. so user impact is small.


alexbishop1 (approved):
⏱️

SandorArpa (approved):
LGTM

PR #5400, opened 2024-07-01
PP-12359 Correct get events by service id path

james-peacock-ho:
- Correct endpoint for get events by serviceId


DomBelcher (approved):
👍

PR #5410, opened 2024-07-03
PP-12358: Refactor ITestBaseExtension's addCharge* methods

oswaldquek:
There are too many ITestBaseExtension.addCharge* methods which make for difficult reading. Use the standard java builder pattern to rectify this. A new AddChargeParameters class has been added for use with the addCharge* methods. There is another separate AddChargeParams class which could be confusing but reducing these two classes to one requires some thought and should go into another PR.


james-peacock-ho (approved):
LGTM!

PR #5415, opened 2024-07-05
PP-12844 Add get charge by service ID endpoint

james-peacock-ho:
- Add new endpoint to get charge by service ID and account type
- Add integration tests
- Use degatewayed endpoints in the degatewayed delayed capture tests


nataliecarey (approved):
LGTM

PR #5416, opened 2024-07-05
PP-12358: Refactor adding charges in test classes

oswaldquek:
- `ITestBaseExtension.addCharge` should really take an AddChargeParameters rather than it's Builder.
- Change AddChargeParams to be a record class.

There is potential to merge AddChargeParameters and AddChargeParams but that's quite a bit of work for now; not priority and is better left for the next firebreak.


james-peacock-ho (approved):
All looks fine to me. I'm unsure what is gained by making AddChargeParams a record, aside from being able to lose a load of getter methods, so if you could explain this to me sometime that would be great.

oswaldquek:
> All looks fine to me. I'm unsure what is gained by making AddChargeParams a record, aside from being able to lose a load of getter methods, so if you could explain this to me sometime that would be great.

You've explained what we gain from moving to a record 😆 Losing getter methods -> less boilerplate code!

PR #5421, opened 2024-07-08
PP-12830 remove spotify

SandorArpa:
## WHAT YOU DID

When we upgraded to Dropwizard 3, a few scattered code was left behind that refers to com.spotify.


alexbishop1 (approved):
🔇

PR #5424, opened 2024-07-09
PP-12844 Add and refactor tests for create charge endpoint

james-peacock-ho:
Context: There is already a large number of integration tests for the create charge endpoint, and replicating each of these for the degatewayed endpoint would add significantly to the test running time. Some of these tests are for validation only so these are being moved into a single test class. This will minimise the impact on performance through reduced overall startup time and each test will run quicker as a unit test.

- Move validation tests for create charge endpoint into ChargesApiResourceValidationTest
- Use parameterised tests to avoid duplication for the gateway and degatewayed endpoint
- Inline fields so to make the inputs and outputs clearer
- Create accounts via API rather than in the database as per agreed team approach
- Use degatewayed urls in the setup for tests of degatewayed endpoints where possible.
- Subsequent PRs will add coverage for the degatewayed endpoint in respect of the remaining permutations of metadata, Moto, prefilled cardholder details, provider credentials, and zero amount.


src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateLanguageIT.java
```diff
@@ -1,79 +0,0 @@
...
-                JSON_REFERENCE_KEY, JSON_REFERENCE_VALUE,
-                JSON_DESCRIPTION_KEY, JSON_DESCRIPTION_VALUE,
-                JSON_RETURN_URL_KEY, RETURN_URL,
-                JSON_EMAIL_KEY, EMAIL
-        ));
-
-        ValidatableResponse response = testBaseExtension.getConnectorRestApiClient()
-                .postCreateCharge(postBody)
-                .statusCode(Response.Status.CREATED.getStatusCode())
-                .body(JSON_LANGUAGE_KEY, is("en"))
```
oswaldquek, line 48: This assertion `.body("language", is("en"))` seems to have been deleted completely and could go into one of the successful create charge tests.

oswaldquek (changes requested):
Sorry...a minor assertion to put back in!

src/test/java/uk/gov/pay/connector/it/resources/ChargesApiResourceCreateLanguageIT.java
james-peacock-ho: Very well spotted. I thought I had put that back in but clearly not.

oswaldquek (approved):

PR #5430, opened 2024-07-11
PP-12619: Apply strategy for returning gateway account by service id and account type

oswaldquek:
If a service has multiple gateway accounts,
* If a Stripe test account is found, return it
* Otherwise return the sandbox test account
* Obviously, if there’s only one test account found, return that one, regardless of whether it is a sandbox, Stripe, or Worldpay test account
* The only way for a Worldpay test account to be returned is if that’s the only test gateway account found on that service

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoIT.java
@@ -2,9 +2,9 @@ package uk.gov.pay.connector.it.dao;
 
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
-import org.junit.jupiter.api.Nested;
 import uk.gov.pay.connector.cardtype.model.domain.CardTypeEntity;
 import uk.gov.pay.connector.extension.AppWithPostgresAndSqsExtension;
 import uk.gov.pay.connector.gatewayaccount.dao.GatewayAccountDao;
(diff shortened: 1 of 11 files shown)


src/test/java/uk/gov/pay/connector/service/GetGatewayAccountByServiceIdAndAccountTypeTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        Optional<GatewayAccountEntity> gatewayAccount =
+                gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(SERVICE_ID, GatewayAccountType.TEST);
+        assertTrue(gatewayAccount.isPresent());
+        assertTrue(gatewayAccount.get().isWorldpayGatewayAccount());
+    }
+    
+    @Test
+    void shouldLogAndThrowErrorIfThereAreMultipleLiveGatewayAccounts() {
+        //TODO
```
oswaldquek, line 129: This will be implemented in next PR

src/test/java/uk/gov/pay/connector/it/dao/GatewayAccountDaoIT.java
```diff
@@ -660,8 +660,7 @@ void findByExternalId_shouldFindGatewayAccount() {
 
     @Test
     void findByServiceIdAndAccountType_shouldReturnMostRecentlyAddedAccount_whenMultipleAccountsExist() {
```
james-peacock-ho: Maybe the name of this test should change as the behaviour is different

james-peacock-ho (changes requested):
One small change to a test name.

As well as the exception test that you are doing in the next PR, maybe also:
- include a test scenario where there are obscure accounts like Smartpay? (or do these no longer exist for test or live?)
- integration test for the get /v1/api/service/{serviceId}/account/{accountType} endpoint?

james-peacock-ho (approved):
LGTM!

PR #5437, opened 2024-07-12
PP-12619: Throw error if there are multiple live gateway accounts

oswaldquek:
Throw error if there are multiple live gateway accounts associated with a service.
It should also not be possible to create another live gateway account associated with a service.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MultipleLiveGatewayAccountsException.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MultipleLiveGatewayAccountsException.java
@@ -6,8 +6,16 @@ import static java.lang.String.format;
 import static uk.gov.pay.connector.util.ResponseUtil.conflictErrorResponse;
 
 public class MultipleLiveGatewayAccountsException extends WebApplicationException {
+
+    private MultipleLiveGatewayAccountsException(String message) {
+        super(conflictErrorResponse(message));
+    }
     
-    public MultipleLiveGatewayAccountsException(String serviceId) {
-        super(conflictErrorResponse(format("Multiple live gateway accounts found for service [%s]", serviceId)));
+    public static MultipleLiveGatewayAccountsException multipleLiveGatewayAccounts(String serviceId) {
+        return new MultipleLiveGatewayAccountsException(format("Multiple live gateway accounts found for service [%s]", serviceId));
+    }
+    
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
```diff
@@ -148,6 +149,11 @@ public Optional<GatewayAccountEntity> getGatewayAccountByExternal(String gateway
 
     public Optional<GatewayAccountEntity> getGatewayAccountByServiceIdAndAccountType(String serviceId, GatewayAccountType accountType) {
         List<GatewayAccountEntity> gatewayAccounts = gatewayAccountDao.findByServiceIdAndAccountType(serviceId, accountType);
+        
+        if (gatewayAccounts.stream().filter(GatewayAccountEntity::isLive).toList().size() > 1) {
+            throw new MultipleLiveGatewayAccountsException(serviceId);
+        }
```
DomBelcher, line 161: Since the `gatewayAccounts` list is the result of finding the accounts for that service ID/type combo, every account in the list should be either test or live, in which case you can just do `if(accountType.equals(LIVE) && gatewayAccounts.size() > 1)` - although it's probably only the tiniest fraction more perfomant, it does save doing the check when querying for test accounts

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MultipleLiveGatewayAccountsException.java
```diff
@@ -0,0 +1,13 @@
...
+
+import javax.ws.rs.WebApplicationException;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.util.ResponseUtil.conflictErrorResponse;
+
+public class MultipleLiveGatewayAccountsException extends WebApplicationException {
+    
+    public MultipleLiveGatewayAccountsException(String serviceId) {
+        super(conflictErrorResponse(format("Multiple live gateway accounts found for service [%s]", serviceId)));
```
james-peacock-ho: This exception wording makes sense when the existence of multiple live accounts has been identified. But in the case where there is an attempt to create a live account when one already exists, shouldn't the wording be different, e.g. "There is already a live gateway account for this service."?
Alternatively, could make the wording a bit broader to cover both scenarios - e.g. "Services should only have a single live gateway account".

james-peacock-ho (changes requested):

james-peacock-ho (approved):

PR #5443, opened 2024-07-16
PP-12360 Fix issues in Worldpay SubmitRefund tests

DomBelcher:
## WHAT YOU DID
 - Fix `SubmitRefund` tests not using test account object created in test setup
   - This was occurring due to a missing `.insert()` call when creating the default test account
   - These tests were passing due to the default test account using the account ID from `ITestBaseExtension`, which is inserted separately, so calling `.insert()` would have caused the tests to fail due to a duplicate primary key.
   - This is fixed by removing the dependecy on `ITestBaseExtension`
 - Fix single test not running due to missing `@Test` annotations

## How to test
 - Run tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/util/RandomIdGenerator.java
+++ b/src/main/java/uk/gov/pay/connector/util/RandomIdGenerator.java
@@ -48,4 +48,8 @@ public class RandomIdGenerator {
         }
         return sb.toString();
     }
+    
+    public static long randomLong() {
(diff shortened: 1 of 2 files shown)


oswaldquek (approved):
I think using the `ITestBaseExtension` has its merits (although I think it'd be better renamed as `SetupGatewayAccountExtension`) as setting up accounts and credentials are arguably not the concern of making refunds but if you've chosen to set these up manually here that's fine too.

PR #5446, opened 2024-07-18
PP-12938: Some code improvements

oswaldquek:
Add javadoc to explain why the caller of `POST
/v1/api/accounts/{accountId}/charges/{chargeId}/refunds` needs to specify the refund amount available.

The `PaymentGatewayStateTransitions.getInstance` was not thread safe. The solution here would be use the double-checked locking solution, but since the connector will pretty much always need to transition a charge state, let's eagerly create the instance.


james-peacock-ho (approved):
LGTM

PR #5460, opened 2024-07-25
PP-12360 Some refactoring of Refund processing code

DomBelcher:
## WHAT YOU DID
 - Refactor the code for submitting a refund in `RefundService` to improve readability
   - Inline code from `processRefund`
   - Rename `updateRefund` to `updateRefundWithTransactionIdAndState`
   - Refactor calls to re-fetch refund entity from database into standalone method, with explanation of why re-fetching from the DB is necessary
   - Rename `doRefund` to `submitRefund` and update to accept `GatewayAccountEntity` instead of account ID, to allow this to be used in both gatewayed and de-gatewayed endpoints

## HOW TO TEST
 - Run tests


src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -106,8 +106,25 @@ public ChargeRefundResponse doRefundFor(GatewayAccountEntity gatewayAccountEntit
...
+//            setRefundStatus(refundEntity.getId(), gatewayAccountEntity, REFUND_SUBMITTED, charge);
+//        }
+        RefundStatus refundStatus = determineRefundStatus(gatewayRefundResponse);
+
+        if (refundStatus == REFUNDED) {
+            // If the gateway confirms refunds immediately, the refund status needs
+            // to be set to REFUND_SUBMITTED and then REFUNDED. This will  help
+            // services to view refund history in detail in self service.
+            // see Javadoc (RefundHistory) for details on how history is handled
+            setRefundStatus(refundEntity.getId(), gatewayAccountEntity, REFUND_SUBMITTED, charge)
```
oswaldquek: Why does `setRefundStatus` return an Optional? It's clear by this point a RefundEntity already exists. Why can't we inline the `setRefundStatus` code in here?

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -84,13 +84,13 @@ public RefundService(RefundDao refundDao,
         this.gatewayAccountCredentialsService = gatewayAccountCredentialsService;
     }
 
-    public ChargeRefundResponse doRefund(Long accountId, Charge charge, RefundRequest refundRequest) {
+    public Optional<ChargeRefundResponse> doRefund(Long accountId, Charge charge, RefundRequest refundRequest) {
```
oswaldquek: Not sure why an Optional has to be returned here?

src/main/java/uk/gov/pay/connector/refund/resource/RefundsResource.java
```diff
@@ -94,8 +94,12 @@ public Response submitRefund(@Parameter(example = "1", description = "Gateway ac
         validateRefundRequest(refundRequest.getAmount());
 
         ChargeRefundResponse refundServiceResponse = chargeService.findCharge(chargeExternalId, accountId)
-                .map(charge -> refundService.doRefund(accountId, charge, refundRequest))
-                .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeExternalId));
+                .or(() -> {
+                    throw new ChargeNotFoundRuntimeException(chargeExternalId);
+                })
+                .flatMap(charge -> refundService.doRefund(accountId, charge, refundRequest))
+                .orElseThrow(() -> new RuntimeException(""));
```
oswaldquek: What's the value of re-jigging these lines?

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -371,4 +362,10 @@ public void updateRefundParityStatus(String externalId, ParityCheckStatus parity
     public Optional<RefundEntity> findRefundByExternalId(String externalId) {
         return refundDao.findByExternalId(externalId);
     }
+    
+    // This is necessary in transactional methods to ensure the refund entity exists within the persistence context
+    // if called after a refund entity has been created, this should never cause an error
+    public RefundEntity reloadRefundFromDatabase(Long refundId) {
```
DomBelcher: I guess I could rename this something like `reloadRefundWithinPersistenceContext` which may remove the need for the comment, but I think the comment explains it better

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -106,8 +107,19 @@ public ChargeRefundResponse doRefundFor(GatewayAccountEntity gatewayAccountEntit
...
-        return new ChargeRefundResponse(gatewayRefundResponse, refund);
+        RefundStatus refundStatus = determineRefundStatus(gatewayRefundResponse);
+
+        if (refundStatus == REFUNDED) {
+            // If the gateway confirms refunds immediately, the refund status needs
+            // to be set to REFUND_SUBMITTED and then REFUNDED. This will  help
+            // services to view refund history in detail in self service.
+            // see Javadoc (RefundHistory) for details on how history is handled
+             RefundEntity refundWithSubmittedState = setRefundStatus(refundEntity.getId(), gatewayAccountEntity, REFUND_SUBMITTED, charge);
+             userNotificationService.sendRefundIssuedEmail(refundWithSubmittedState, charge, gatewayAccountEntity);
```
DomBelcher, line 105: I think moving this call to `sendRefundIssuedEmail` is the only significant change to the actual flow, as it used to be called in `updateRefund`, but since it only runs if the `refundStatus` is `REFUNDED` there, this shouldn't actually change anything. It also doesn't matter whether an updated `RefundEntity` is passed in as the only data it uses from the refund is the external ID and the refund amount, which shouldn't change.

DomBelcher: This could be changed to pass in the original `RefundEntity` returned by `createRefund`, and removing the assignment to `refundWithSubmittedState`. This might make it clearer that `sendRefundIssuedEmail` doesn't depend on the refund state being updated - it could even be called before updating the refund state - although maybe it's better not to send the email before updating the state in case that transaction fails

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
DomBelcher: I suppose I could change the method name and leave the comment in, so the purpose behind the method is clear, and the comment explains why it's necessary - I could change the comment to something like 
`This is necessary in transactional methods to ensure updates to the refund are persisted once the transaction completes`

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -84,13 +85,13 @@ public RefundService(RefundDao refundDao,
...
-        return doRefundFor(gatewayAccountEntity, charge, refundRequest);
-    }
+//    public ChargeRefundResponse doRefund(Long accountId, Charge charge, RefundRequest refundRequest) {
+//        GatewayAccountEntity gatewayAccountEntity = gatewayAccountDao.findById(accountId).orElseThrow(
+//                () -> new GatewayAccountNotFoundException(accountId));
+//        return doRefundFor(gatewayAccountEntity, charge, refundRequest);
+//    }
     
-    public ChargeRefundResponse doRefundFor(GatewayAccountEntity gatewayAccountEntity, Charge charge, RefundRequest refundRequest) {
+    public ChargeRefundResponse doRefund(GatewayAccountEntity gatewayAccountEntity, Charge charge, RefundRequest refundRequest) {
```
DomBelcher: Maybe we can improve on `doRefund` as the method name? Even `submitRefund` I think is better, but maybe we can do better than that too

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
oswaldquek: If it's only the external id and amount that are used, perhaps passing in those values to the `sendRefundIssuedEmail` is the right thing to do. However I don't think what we have here is particularly an issue.

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
```diff
@@ -371,4 +362,10 @@ public void updateRefundParityStatus(String externalId, ParityCheckStatus parity
     public Optional<RefundEntity> findRefundByExternalId(String externalId) {
         return refundDao.findByExternalId(externalId);
     }
+    
+    // This is necessary in transactional methods to ensure the refund entity exists within the persistence context
+    // if called after a refund entity has been created, this should never cause an error
+    public RefundEntity reloadRefundFromDatabase(Long refundId) {
+       return refundDao.findById(refundId).orElseThrow(() -> new RuntimeException(format("Refund %s no longer exists in database", refundId)));
```
oswaldquek, line 353: 🎉

src/main/java/uk/gov/pay/connector/refund/service/RefundService.java
oswaldquek: I think this is fine. Method could be `private` though?

oswaldquek (approved):

PR #5464, opened 2024-07-26
PP-12915: Complete "request stripe test account" functionality

oswaldquek:
`POST /v1/service/{serviceId}/request-stripe-test-account` already creates a test account on Stripe's end. This commit additionally:
* Creates a gateway account in connector with payment provider "stripe" and type "test"
* Disables the old sandbox account
* Handles error scenarios


src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -9,60 +11,105 @@
...
-//                    @ApiResponse(responseCode = "400", description = "Stripe Connect Account already exists"),
-//                    @ApiResponse(responseCode = "404", description = "Not found - Account with serviceId does not exist")
-//            }
-//    )
+    @Operation(
+            summary = "1) Create a Stripe Connect Account 2) Create gateway account in connector 3) Disables the old " +
+                    "sandbox account",
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "OK"),
+                    @ApiResponse(responseCode = "400", description = "Stripe Connect Account already exists, or existing test account is not a Sandbox one"),
```
DomBelcher: I think a `409` makes more sense in this case

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountRequest.java
```diff
@@ -139,4 +139,72 @@ public boolean isAllowApplePay() {
     public boolean isAllowGooglePay() {
         return allowGooglePay;
     }
+    
+    public static final class GatewayAccountRequestBuilder {
```
DomBelcher: If this is an inner class on `GatewayAccountRequest` you could just call it `Builder`, and even have `builder()` method on `GatewayAccountRequest` so it can be instantiated as `GatewayAccountRequest.builder()`

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountRequest.java
```diff
@@ -139,4 +139,72 @@ public boolean isAllowApplePay() {
...
+        private @NotBlank String providerAccountType;
```
DomBelcher: Does the `@NotBlank` annotation do anything here, as there's no use of `@Valid` when creating the builder? (although I admit my understand of the Javax validation API is patchy)

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -9,60 +11,105 @@
...
+                .withDescription(sandboxGatewayAccount.getDescription())
+                .withAnalyticsId(sandboxGatewayAccount.getAnalyticsId())
+                .withPaymentProvider("stripe")
+                .withRequires3ds(sandboxGatewayAccount.isRequires3ds())
+                .withAllowApplePay(sandboxGatewayAccount.isAllowApplePay())
+                .withAllowGooglePay(sandboxGatewayAccount.isAllowGooglePay())
+                .build();
+        gatewayAccountService.createGatewayAccount(gatewayAccountRequest, uriInfo);
+
+        JsonNode patchRequest = objectMapper.valueToTree(
```
DomBelcher: I think it would be simpler & clearer to just add a `GatewayAccountService.disableAccount` method

src/test/java/uk/gov/pay/connector/extension/AppWithPostgresAndSqsExtension.java
```diff
@@ -313,13 +314,15 @@ public void purgeEventQueue() {
         sqsClient.purgeQueue(new PurgeQueueRequest(getEventQueueUrl()));
     }
 
-    public void setupSandboxGatewayAccount(String serviceId, String serviceName) {
-        givenSetup().body(toJson(Map.of(
+    public CreateGatewayAccountResponse setupSandboxGatewayAccount(String serviceId, String serviceName) {
```
DomBelcher: I think this should be in `ITestBaseExtension`

DomBelcher (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccount/model/GatewayAccountRequest.java
oswaldquek: Not sure how the `@NotBlank` got in there

src/test/java/uk/gov/pay/connector/extension/AppWithPostgresAndSqsExtension.java
oswaldquek: I'll move this method back into a private method in StripeAccountResourceIT. I feel ITestBaseExtension is a quagmire and I don't want to exacerbate it by adding a new method with a name that's close to other method names in that class thus making that class more bloated.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -1,68 +1,111 @@
...
-//                    @ApiResponse(responseCode = "404", description = "Not found - Account with serviceId does not exist")
-//            }
-//    )
+    @Operation(
+            summary = "1) Create a Stripe Connect Account 2) Create gateway account in connector 3) Disables the old " +
+                    "sandbox account",
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "OK"),
+                    @ApiResponse(responseCode = "400", description = "Stripe Connect Account already exists, or existing test account is not a Sandbox one"),
+                    @ApiResponse(responseCode = "409", description = "Not found - Account with serviceId does not exist")
```
DomBelcher: Sorry I should have been clearer with my earlier comment - I think a `409` makes sense if there is already a Stripe account or the current test account isn't a sandbox account. If there's no test account with the specified service ID I'm a bit torn on whether a `404` or `409` makes more sense. The [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) would suggest that maybe a `409` is correct, given the example of creating `/a/b/c/d/` requiring `/a/b/c/` to exist, but whether this is to be taken as gospel I don't know.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -1,68 +1,111 @@
...
+        var maybeSandboxTestAccount = gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, TEST)
                 .orElseThrow(() -> new GatewayAccountNotFoundException(serviceId, TEST));
+
+        if (maybeSandboxTestAccount.getGatewayName().equalsIgnoreCase("stripe")) {
+            throw new ConflictWebApplicationException("Cannot request Stripe test account because a Stripe test account already exists.");
+        }
+
+        if (!maybeSandboxTestAccount.getGatewayName().equalsIgnoreCase("sandbox")) {
+            throw new BadRequestException("Cannot request Stripe test account because existing test account is not a Sandbox one.");
+        }
```
DomBelcher: As above re. `400` vs `404` vs `409`

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -1,68 +1,111 @@
...
+    private final ObjectMapper objectMapper;
 
     @Inject
     public StripeAccountResource(StripeAccountService stripeAccountService,
-                                 GatewayAccountService gatewayAccountService) {
+                                 GatewayAccountService gatewayAccountService, 
+                                 ObjectMapper objectMapper) {
         this.stripeAccountService = stripeAccountService;
         this.gatewayAccountService = gatewayAccountService;
+        this.objectMapper = objectMapper;
```
DomBelcher: The object mapper isn't needed if not disabling the account via `PATCH` request

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -1,68 +1,111 @@
...
-//                    @ApiResponse(responseCode = "200", description = "OK"),
-//                    @ApiResponse(responseCode = "400", description = "Stripe Connect Account already exists"),
-//                    @ApiResponse(responseCode = "404", description = "Not found - Account with serviceId does not exist")
-//            }
-//    )
+    @Operation(
+            summary = "1) Create a Stripe Connect Account 2) Create gateway account in connector 3) Disables the old " +
+                    "sandbox account",
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "OK"),
```
DomBelcher: Maybe a `201` is more appropriate if the account is created successfully

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -1,68 +1,111 @@
...
     ) {
-        return gatewayAccountService.getGatewayAccountByServiceIdAndAccountType(serviceId, TEST)
-                .map(gatewayAccountEntity -> {
-                    Account testAccount = stripeAccountService.createTestAccount(gatewayAccountEntity.getServiceName());
-                    stripeAccountService.createDefaultPersonForAccount(testAccount.getId());
-                    return Response.ok().build();
-                })
+        var sandboxGatewayAccount = getSandboxGatewayAccount(serviceId);
+
+        Account testAccount = stripeAccountService.createTestAccount(sandboxGatewayAccount.getServiceName());
```
DomBelcher, line 66: This still throws a `RuntimeException` if creating the account fails for any reason - does this need changing? It would result in the API returning a `500`, but I think a `502` would be correct

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -1,68 +1,111 @@
...
+                .withServiceId(sandboxGatewayAccount.getServiceId())
+                .withDescription(sandboxGatewayAccount.getDescription())
+                .withAnalyticsId(sandboxGatewayAccount.getAnalyticsId())
+                .withPaymentProvider("stripe")
+                .withRequires3ds(sandboxGatewayAccount.isRequires3ds())
+                .withAllowApplePay(sandboxGatewayAccount.isAllowApplePay())
+                .withAllowGooglePay(sandboxGatewayAccount.isAllowGooglePay())
+                .build();
+        gatewayAccountService.createGatewayAccount(gatewayAccountRequest, uriInfo);
+        gatewayAccountService.disableAccount(sandboxGatewayAccount.getId());
```
DomBelcher, line 84: Do we need to give any thought to how we handle potential errors with creating the new gateway account or disabling the old one, or any other potential sad paths? Is it for example possible for two `request-stripe-test-account` requests to come in before the first has time to complete, and two Connect accounts or Gateway accounts to be created as a result? 

As a side note, should we add in some check on `GatewayAccountService.createGatewayAccount` that creating the new account won't duplicate an existing account, e.g. creating a second Stripe test account - I know this already exists as a check for creating a live account, but not for test accounts

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -1,68 +1,111 @@
...
 import javax.ws.rs.core.Response;
+import javax.ws.rs.core.UriInfo;
+import java.util.Map;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 import static uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType.TEST;
+import static uk.gov.pay.connector.gatewayaccount.resource.GatewayAccountRequestValidator.FIELD_DISABLED;
+import static uk.gov.service.payments.commons.model.jsonpatch.JsonPatchKeys.FIELD_OPERATION;
+import static uk.gov.service.payments.commons.model.jsonpatch.JsonPatchKeys.FIELD_OPERATION_PATH;
+import static uk.gov.service.payments.commons.model.jsonpatch.JsonPatchKeys.FIELD_VALUE;
```
DomBelcher: A lot of these imports can be removed

DomBelcher (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
oswaldquek: Ugh I was going to fix the RuntimeException stuff but slipped my mind...

src/test/java/uk/gov/pay/connector/extension/AppWithPostgresAndSqsExtension.java
DomBelcher: I agree on `ITestBaseExtension` being a bit bloated, it's part of why I think we should keep `AppWithPostgesAndSqsExtension` as lean as possible. I think it highlights a potential issue of doing things via the API - if we use the test helpers and the associated builders then we wouldn't need a whole host of `createXwithY` methods, etc.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
oswaldquek: I'm going to go for a `500` here. The runtime exception wraps a StripeException which could be a number of things - e.g. AuthenticationException, InvalidRequestException, RateLimitException - which encompasses a number of scenarios, some of which could be a client rather than a server fault, so it seems better to return a generic `500` rather than `502`.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
oswaldquek: It's not that you weren't clear - I changed the wrong line 🤦

oswaldquek: Given that throughout our code we return a 404 when a resource doesn't exist, we should keep it that way here.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
oswaldquek: I think it should be a 409 if the existing test account is not a sandbox one, because 1) it indicates the current state isn't correct and 2) it's nearly the same case for if a Stripe test account already exists.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
oswaldquek: My opinion is that a 200 should be returned because the resource "Stripe test account" is created on Stripe's server; it's not a resource we control or can direct the user to. A location URL is usually returned with a 201 and this wouldn't be applicable here.

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
oswaldquek: > Is it for example possible for two request-stripe-test-account requests to come in before the first has time to complete, and two Connect accounts or Gateway accounts to be created as a result?

Hmm, good thinking. It has occurred to me I haven't associated the Stripe connect account id (beginning with "acct_") with the newly created gateway account in connector. I'll need to add a gateway account credential for this. 

We can't prevent multiple connect accounts from being created in the scenario you're proposing, but once a Stripe gateway account has been associated to a connector gateway account, then we can essentially check for that existence and leave any subsequent Stripe connect accounts as orphans. If you have a look in an example [Stripe test account on toolbox](https://toolbox.payments.service.gov.uk/gateway_accounts/7180) you can see it's mapped to acct_1PVrIHRbaqVLkZV5. As long as that's fixed we don't need to worry about multiple `request-stripe-test-account`s creating multiple Stripe connect accounts.

>As a side note, should we add in some check on GatewayAccountService.createGatewayAccount that creating the new account won't duplicate an existing account, e.g. creating a second Stripe test account - I know this already exists as a check for creating a live account, but not for test accounts

I think this is very unlikely to happen but this is something we should do.

Considering all the above, it feels like lots of thinking and a new set of tests so is probably best to go in a next PR?

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
DomBelcher: Yeah I think 409 for both makes sense

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/StripeAccountResource.java
```diff
@@ -43,26 +48,53 @@ public StripeAccountResource(StripeAccountService stripeAccountService,
...
-//                    @ApiResponse(responseCode = "404", description = "Not found - Account with serviceId does not exist")
-//            }
-//    )
+    @Operation(
+            summary = "1) Create a Stripe Connect Account 2) Create gateway account in connector 3) Disables the old " +
+                    "sandbox account",
+            responses = {
+                    @ApiResponse(responseCode = "201", description = "OK"),
+                    @ApiResponse(responseCode = "409", description = "Stripe Connect Account already exists, or existing test account is not a Sandbox one"),
+                    @ApiResponse(responseCode = "404", description = "Not found - Account with serviceId does not exist")
```
DomBelcher: Might as well reorder these so they're in ascending order

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeGatewayAccountRequest.java
```diff
@@ -24,17 +24,90 @@ public StripeGatewayAccountRequest(@JsonProperty("type") String providerAccountT
...
     ) {
-
         super(providerAccountType, paymentProvider, serviceName, serviceId, description, analyticsId, requires3ds, allowApplePay, allowGooglePay);
-        this.credentials = Optional.ofNullable(credentials);
+        this.credentials = credentials;
     }
 
     public Optional<StripeCredentials> getCredentials() {
-        return credentials;
+        return Optional.ofNullable(credentials);
```
DomBelcher: I'm not sure this method is even atually used, only `getCredentialsAsMap` seems to be. Maybe it can just be removed?

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeGatewayAccountRequest.java
```diff
@@ -24,17 +24,90 @@ public StripeGatewayAccountRequest(@JsonProperty("type") String providerAccountT
...
     }
 
     @Override
     public Map<String, String> getCredentialsAsMap() {
-        return credentials.isPresent() ? credentials.get().toMap() : newHashMap();
+        return Optional.ofNullable(credentials).map(StripeCredentials::toMap).orElse(newHashMap());
```
DomBelcher: Not that it matters much but I think we could use `Map.of()` instead of `newHashMap()` as the Map seems to never be mutated, but it may be a virtually pointless change

DomBelcher (commented):
Great work! Just a few very minor points which may not even need addressing

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeGatewayAccountRequest.java
oswaldquek: Strange, IntelliJ didn't flag this up for me.

src/main/java/uk/gov/pay/connector/gatewayaccount/model/StripeGatewayAccountRequest.java
oswaldquek: Yeah it's a pointless change and not worth waiting on a new build for, but I'm going to push a new commit anyway and it's better to use standard JDK classes.

DomBelcher: > it's better to use standard JDK classes.

Not entirely pointless then, only almost :smile:

DomBelcher (approved):
Nice work!

PR #5468, opened 2024-07-30
PP-12360 Worldpay integration tests for degatewayed submitRefund

DomBelcher:
## WHAT YOU DID
 - Add integration tests for `POST /v1/api/service/{serviceId}/account/{accountType}/charges/{chargeId}/refunds` for Worldpay

## How to test
 - Run tests

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/resource/RefundsResource.java
+++ b/src/main/java/uk/gov/pay/connector/refund/resource/RefundsResource.java
@@ -83,6 +83,7 @@ public class RefundsResource {
                                     "}"))),
                     @ApiResponse(responseCode = "400", description = "Bad request - Invalid fields or not sufficient amount available for refund"),
                     @ApiResponse(responseCode = "404", description = "Not found - gateway account or charge not found"),
+                    @ApiResponse(responseCode = "412", description = "Precondition failed - Refund amount available does not match"),
                     @ApiResponse(responseCode = "500", description = "Internal server error")
             }
     )
@@ -130,6 +131,7 @@ public class RefundsResource {
                                     "}"))),
                     @ApiResponse(responseCode = "400", description = "Bad request - Invalid fields or not sufficient amount available for refund"),
                     @ApiResponse(responseCode = "404", description = "Not found - gateway account or charge not found"),
+                    @ApiResponse(responseCode = "412", description = "Precondition failed - Refund amount available does not match"),
                     @ApiResponse(responseCode = "500", description = "Internal server error")
             }
     )
--- a/src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
@@ -4,9 +4,13 @@ import com.google.gson.Gson;
 import io.restassured.http.ContentType;
 import io.restassured.response.ValidatableResponse;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
 import uk.gov.pay.connector.extension.AppWithPostgresAndSqsExtension;
 import uk.gov.pay.connector.it.dao.DatabaseFixtures;
 import uk.gov.pay.connector.refund.model.domain.RefundStatus;
@@ -14,9 +18,11 @@ import uk.gov.pay.connector.util.AddGatewayAccountCredentialsParams;
 import uk.gov.pay.connector.util.TestTemplateResourceLoader;
 import uk.gov.service.payments.commons.model.ErrorIdentifier;
 
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.equalTo;
 import static com.github.tomakehurst.wiremock.client.WireMock.equalToXml;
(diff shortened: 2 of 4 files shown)


src/main/java/uk/gov/pay/connector/refund/resource/RefundsResource.java
```diff
@@ -83,6 +83,7 @@ public RefundsResource(RefundService refundService, ChargeService chargeService,
                                     "}"))),
                     @ApiResponse(responseCode = "400", description = "Bad request - Invalid fields or not sufficient amount available for refund"),
                     @ApiResponse(responseCode = "404", description = "Not found - gateway account or charge not found"),
+                    @ApiResponse(responseCode = "412", description = "Precondition failed - Refund amount available does not match"),
```
oswaldquek, line 86: I don't think 412 is correct here as it seems to involve the "If-Unmodified-Since" or "If-Match" request headers which we're not using.. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412. I'd go for 409 here.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
```diff
@@ -69,6 +74,8 @@ public class WorldpayRefundSubmitResourceIT {
                     CREDENTIALS_USERNAME, "test-user",
                     CREDENTIALS_PASSWORD, "test-password")
     );
+    
+    private static final List<ChargeStatus> NON_REFUNDABLE_CHARGE_STATUSES = Arrays.stream(ChargeStatus.values()).filter(chargeStatus -> !chargeStatus.equals(CAPTURED)).toList();
```
oswaldquek: This 1) could be kept as a Stream since you're [streaming it anyway](https://github.com/alphagov/pay-connector/pull/5468/files#diff-b4651ff932a3899958dec246cc905507983891515024db7cfd8de40d93221a5aR639) and 2) moved into a field under `class ByServiceIdAndType` as that's where it's only being used.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
```diff
@@ -374,40 +377,43 @@ void shouldBeAbleToRequestARefund_partialAmount() {
...
-
-            verifyRequestBodyToWorldpay(WORLDPAY_URL, expectedRequestBody);
+            verifyRequestBodyToWorldpay(expectedRequestBody);
         }
 
         @Test
+        @DisplayName("Should create refund, send refund request to Worldpay, and update refund state to 'REFUND SUBMITTED' for full refund")
         void shouldBeAbleToRequestARefund_fullAmount() {
-            Long refundAmount = defaultTestCharge.getAmount();
+            int refundAmount = (int) defaultTestCharge.getAmount();
```
oswaldquek, line 411: This can just be the primitive `long` type.

oswaldquek (changes requested):
Nice one. Some easy changes requested.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
DomBelcher: This was just to keep the linter happy, the assertion `.body("amount", is(refundAmount))` expects an int meaning the refund amount has to be one of:

 - A boxed `Long` so you can call `Long.intValue()`
 - A primitive `long` cast to an `int` in `.body("amount", is((int) refundAmount))`
 - A primitive `int` as above

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
DomBelcher: Is it worth adding the same parameterised test to the `ByAccountId` tests? If so I think this needs to stay as a `List` so it can be streamed twice. If not I'll make it a `Stream` and move it

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
oswaldquek: Oh hmm... the `is` matcher should be to take a `long` or a `Long`, have you tried it? It works for me. The `is` method looks like `public static <T> Matcher<T> is(T value)` so can pass anything in.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
oswaldquek: I think it's fine not adding it to the `ByAccountId` tests. It'd be nice to but it's not technically in the scope of the story and I'm keen for the story to move along.

src/main/java/uk/gov/pay/connector/refund/resource/RefundsResource.java
DomBelcher: Our use here is similar to the `If-Match` header, although we're using the `amount_available` body parameter in place of an etag on the header. Happy to change it to a `409` though, I didn't give any thought to why it was a `412`, I just added this in because that's the code set in `RefundException.refundAmountAvailableMismatchException`

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
DomBelcher: It considers ints and longs to be different, e.g. `100` != `100L`

src/main/java/uk/gov/pay/connector/refund/resource/RefundsResource.java
oswaldquek: Ah I didn't realise that. That's fine then.

src/test/java/uk/gov/pay/connector/it/resources/worldpay/WorldpayRefundSubmitResourceIT.java
oswaldquek: 👌

oswaldquek (approved):

PR #5469, opened 2024-07-30
PP-12413: Prevent creation of duplicate Stripe test accounts

oswaldquek:
Because there is a possibility of concurrent requests to `POST /v1/service/{serviceId}/request-stripe-test-account`, do a check to prevent a duplicate test Stripe gateway account from being created.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MultipleStripeTestGatewayAccountsException.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MultipleStripeTestGatewayAccountsException.java
@@ -1,0 +1,16 @@
+package uk.gov.pay.connector.gatewayaccount.exception;
+
+import javax.ws.rs.WebApplicationException;
+
+import static java.lang.String.format;
+import static uk.gov.pay.connector.util.ResponseUtil.conflictErrorResponse;
+
+public class MultipleStripeTestGatewayAccountsException extends WebApplicationException {
+    
+    public MultipleStripeTestGatewayAccountsException(String serviceId, 
+                                                      String gatewayAccountExternalId, 
+                                                      String gatewayAccountCredentialExternalId) {
+        super(conflictErrorResponse(format("Service '%s' already has an active Stripe gateway account with external id " +
+                "'%s' with gateway account credential is '%s' ", serviceId, gatewayAccountExternalId, gatewayAccountCredentialExternalId)));
+    }
+}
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MultipleStripeTestGatewayAccountsException.java
```diff
@@ -0,0 +1,16 @@
...
+import static java.lang.String.format;
+import static uk.gov.pay.connector.util.ResponseUtil.conflictErrorResponse;
+
+public class MultipleStripeTestGatewayAccountsException extends WebApplicationException {
+    
+    public MultipleStripeTestGatewayAccountsException(String serviceId, 
+                                                      String gatewayAccountExternalId, 
+                                                      String gatewayAccountCredentialExternalId) {
+        super(conflictErrorResponse(format("Service '%s' already has an active Stripe gateway account with external id " +
+                "'%s' with gateway account credential is '%s' ", serviceId, gatewayAccountExternalId, gatewayAccountCredentialExternalId)));
```
DomBelcher, line 14: The String concatenation here is redundant

src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
```diff
@@ -138,6 +140,8 @@ public CreateGatewayAccountResponse createGatewayAccount(GatewayAccountRequest g
                 getGatewayAccountByServiceIdAndAccountType(gatewayAccountRequest.getServiceId(), LIVE).isPresent()) {
             throw MultipleLiveGatewayAccountsException.liveGatewayAccountAlreadyExists(gatewayAccountEntity.getServiceId());
         }
+        
+        throwIfRequestIsForStripeTestAccountButItAlreadyExists(gatewayAccountEntity, gatewayAccountRequest.getServiceId());
```
DomBelcher: This could be moved up to the start of the method if the requested account type is passed in via `gatewayAccountRequest.getProviderAccountType()` in place of `gatewayAccountEntity` - it makes a bit more sense I think if it just requires the account type rather than the new account entity - not that significant though

src/main/java/uk/gov/pay/connector/gatewayaccount/exception/MultipleStripeTestGatewayAccountsException.java
oswaldquek: It's deliberate as I generally don't want lines roughly more than 120 chars long. It just makes things easier to read.

DomBelcher: Ah ok, makes sense

DomBelcher (approved):
Nice work 👍

PR #5474, opened 2024-08-01
PP-13031 add new columns

SandorArpa:
## WHAT YOU DID

Add exemption_3ds_requested column of type VARCHAR(32) to the charges table of the connector database with all existing rows having the value NULL


marcotranchino (approved):
LGTM 👍

PR #5480, opened 2024-08-05
PP-12360 Clean up sandbox integration tests for submit refund endpoint

DomBelcher:
## WHAT YOU DID
 - Minor cleanup of sandbox integration tests for submit refund endpoint
 - Because `defaultTestAccount` was never inserted into the database, it was not actually used in the tests, instead the tests would fall back to the default account inserted in `ITestBaseExtension`
 - As the `defaultTestAccount` was already included in the test code, and nothing else from `ITestBaseExtension` was used, it was easiest to simply remove the use of `ITestBaseExtension`

## How to test
 - Run tests

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundsResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundsResourceIT.java
@@ -11,7 +11,6 @@ import uk.gov.pay.connector.charge.model.ChargeResponse;
 import uk.gov.pay.connector.client.ledger.model.LedgerTransaction;
 import uk.gov.pay.connector.common.model.api.ExternalRefundStatus;
 import uk.gov.pay.connector.extension.AppWithPostgresAndSqsExtension;
-import uk.gov.pay.connector.it.base.ITestBaseExtension;
 import uk.gov.pay.connector.it.dao.DatabaseFixtures;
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundsResourceIT.java
```diff
@@ -39,39 +38,51 @@
...
     
     @BeforeEach
     void setUpCharge() {
         defaultTestAccount = DatabaseFixtures
                 .withDatabaseTestHelper(app.getDatabaseTestHelper())
                 .aTestAccount()
-                .withAccountId(Long.parseLong(testBaseExtension.getAccountId()))
-                .withPaymentProvider(testBaseExtension.getPaymentProvider())
-                .withGatewayAccountCredentials(List.of(testBaseExtension.getCredentialParams()))
-                .withCredentials(testBaseExtension.getCredentials());
```
oswaldquek, line 65: Never actually inserted. Lol.

oswaldquek (approved):

src/test/java/uk/gov/pay/connector/it/resources/sandbox/SandboxRefundsResourceIT.java
DomBelcher: If it was it would have caused an error as it causes a primary key conflict with the account created in `ITestBaseExtension`

PR #5481, opened 2024-08-05
PP-12413: Set up pact provider state

oswaldquek:
Sets up the [state from this PR](https://github.com/alphagov/pay-selfservice/pull/4244/files#diff-25be690c9966a02113560ca2ba8e5acd2b14c4d209657eb9231d9f59ca93112bR48).

Unfortunately setting up two separate states - one for the gateway account and one of the stripe mocking - is not possible. They all need to be in one state 😢


alexbishop1 (approved):
What a state.

PR #5484, opened 2024-08-06
PP-13031 add 3dsRequested to charge entity

SandorArpa:
## WHAT YOU DID

Add field `exemption_3ds_requested` to ChargeEntity.
Add process to set the value when exemption is requested.

Add relevant tests.


marcotranchino (approved):
LGTM 👍

PR #5496, opened 2024-08-09
PP-12876 Make serviceId mandatory when creating accounts

james-peacock-ho:
- Ensure serviceId cannot be null or blank in a GatewayAccountRequest
- Update tests to meet this requirement
- GatewayAccountResourceITBaseExtensions has a lot of helper methods to create gateway accounts, so many that it undermines their usefulness and the class is quite cluttered. I've created a CreateGatewayAccountRequestBuilder to provide flexibility for creating the request payload with different parameters, a method for creating an account using this class, and I've refactored the tests to use that method and deleted the old helper methods.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/AgreementsApiResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/AgreementsApiResourceIT.java
@@ -2,8 +2,8 @@ package uk.gov.pay.connector.it.resources;
 
 import io.restassured.response.ValidatableResponse;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import uk.gov.pay.connector.extension.AppWithPostgresAndSqsExtension;
 import uk.gov.pay.connector.gatewayaccount.model.GatewayAccountType;
@@ -28,6 +28,7 @@ import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.matchesPattern;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
+import static uk.gov.pay.connector.it.resources.GatewayAccountResourceITBaseExtensions.CreateGatewayAccountPayloadBuilder.aCreateGatewayAccountPayloadBuilder;
 import static uk.gov.pay.connector.matcher.ZoneDateTimeAsStringWithinMatcher.isWithin;
 import static uk.gov.pay.connector.util.AddAgreementParams.AddAgreementParamsBuilder.anAddAgreementParams;
 import static uk.gov.pay.connector.util.AddPaymentInstrumentParams.AddPaymentInstrumentParamsBuilder.anAddPaymentInstrumentParams;
@@ -37,7 +38,7 @@ public class AgreementsApiResourceIT {
     
     @RegisterExtension
     public static AppWithPostgresAndSqsExtension app = new AppWithPostgresAndSqsExtension();
-    public static GatewayAccountResourceITBaseExtensions testBaseExtension = new GatewayAccountResourceITBaseExtensions("stripe", app.getLocalPort());
+    public static GatewayAccountResourceITBaseExtensions testBaseExtension = new GatewayAccountResourceITBaseExtensions(app.getLocalPort());
     
     public static final String VALID_SERVICE_ID = "a-valid-service-id";
     private static final String REFERENCE_ID = "1234";
@@ -182,9 +183,18 @@ public class AgreementsApiResourceIT {
 
         @Nested
         class CreateAgreement {
+            private String gatewayAccountId;
+            
+            @BeforeEach
+            void setUp() {
+                gatewayAccountId = testBaseExtension.createAGatewayAccountAndExtractAccountId(
+                        aCreateGatewayAccountPayloadBuilder()
+                        .withServiceId(VALID_SERVICE_ID)
+                        .build());
+            }
+            
             @Test
             void shouldCreateAgreement_forValidRequest() {
-                String gatewayAccountId = testBaseExtension.createAGatewayAccountWithServiceId(VALID_SERVICE_ID);
                 testBaseExtension.updateGatewayAccount(gatewayAccountId, "recurring_enabled", true);
 
                 String createAgreementPayload = toJson(Map.of(
@@ -216,8 +226,6 @@ public class AgreementsApiResourceIT {
 
             @Test
             void shouldReturn422WhenReferenceIdTooLong() {
-                testBaseExtension.createAGatewayAccountWithServiceId(VALID_SERVICE_ID);
-
                 String payload = toJson(Map.of(
                         "reference", REFERENCE_ID_TOO_LONG,
                         "description", DESCRIPTION
@@ -233,8 +241,6 @@ public class AgreementsApiResourceIT {
 
             @Test
             void shouldReturn422WhenReferenceIdEmpty() {
-                testBaseExtension.createAGatewayAccountWithServiceId(VALID_SERVICE_ID);
-
                 String payload = toJson(Map.of(
                         "reference", REFERENCE_ID_EMPTY,
                         "description", DESCRIPTION
@@ -250,8 +256,6 @@ public class AgreementsApiResourceIT {
 
             @Test
             void shouldReturn422WhenReferenceIdNotProvided() {
-                testBaseExtension.createAGatewayAccountWithServiceId(VALID_SERVICE_ID);
-
                 String payload = toJson(Map.of(
                         "description", DESCRIPTION
                 ));
@@ -267,8 +271,6 @@ public class AgreementsApiResourceIT {
 
             @Test
             void shouldReturn422WhenDescriptionNotProvided() {
-                testBaseExtension.createAGatewayAccountWithServiceId(VALID_SERVICE_ID);
-
                 String payload = toJson(Map.of(
                         "reference", REFERENCE_ID
                 ));
@@ -283,8 +285,6 @@ public class AgreementsApiResourceIT {
 
             @Test
             void shouldReturn422WhenDescriptionEmpty() {
-                testBaseExtension.createAGatewayAccountWithServiceId(VALID_SERVICE_ID);
-
                 String payload = toJson(Map.of(
                         "reference", REFERENCE_ID,
                         "description", ""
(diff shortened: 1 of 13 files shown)


src/test/java/uk/gov/pay/connector/it/resources/AgreementsApiResourceIT.java
```diff
@@ -37,7 +38,7 @@ public class AgreementsApiResourceIT {
     
     @RegisterExtension
     public static AppWithPostgresAndSqsExtension app = new AppWithPostgresAndSqsExtension();
-    public static GatewayAccountResourceITBaseExtensions testBaseExtension = new GatewayAccountResourceITBaseExtensions("stripe", app.getLocalPort());
+    public static GatewayAccountResourceITBaseExtensions testBaseExtension = new GatewayAccountResourceITBaseExtensions(app.getLocalPort());
```
DomBelcher: While we're here I wonder if we should rename `GatewayAccountResourceITBaseExtensions` - it's not actually a JUnit extension any more, as the lifecycle methods have all been removed

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
```diff
@@ -1,28 +1,28 @@
...
 import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.hamcrest.Matchers.nullValue;
 import static uk.gov.pay.connector.util.JsonEncoder.toJson;
 
 public class GatewayAccountResourceITBaseExtensions {
-    private final String paymentProvider;
+    @RegisterExtension
+    public static AppWithPostgresAndSqsExtension app = new AppWithPostgresAndSqsExtension();
```
DomBelcher: You shouldn't need the app extension, if the constructor just takes the app port you can set up requests with `given().port(...)`

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
```diff
@@ -31,75 +31,18 @@ public GatewayAccountResourceITBaseExtensions(String paymentProvider, int appLoc
...
-
-    protected String createAGatewayAccountWithServiceId(String serviceId, String testProvider) {
-        return extractGatewayAccountId(createAGatewayAccountFor(appLocalPort, testProvider, "description", "analytics-id", "", "test", serviceId));
-    }
-
-    public static ValidatableResponse createAGatewayAccountFor(int port, String testProvider, String description, String analyticsId, String requires3ds, String type, String serviceId) {
-        RequestSpecification requestSpecification = getRequestSpecification(port, testProvider, description, analyticsId, requires3ds, type, serviceId);
-        return requestSpecification
+    
+    protected ValidatableResponse createAGatewayAccount(Map<String, String> createGatewayAccountPayload) {
```
DomBelcher: Very minor but I prefer `createGatewayAccount` to `createAGatewayAccount`

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
```diff
@@ -141,82 +72,60 @@ public static void assertCorrectCreateResponse(ValidatableResponse response, Gat
...
-
-        static GatewayAccountPayload createDefault() {
-            return new GatewayAccountPayload()
-                    .withUsername("a-username")
-                    .withPassword("a-password")
-                    .withMerchantId("a-merchant-id")
-                    .withServiceName("a-service-name");
+    
+    public static class CreateGatewayAccountPayloadBuilder {
+        Map<String, String> payload = new HashMap<String, String> (Map.ofEntries(
```
DomBelcher: I don't think you need to use `Map.ofEntries`, `Map.of(k1, v1, k2, v2, ...)` should work

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
```diff
@@ -31,75 +31,18 @@ public GatewayAccountResourceITBaseExtensions(String paymentProvider, int appLoc
...
-        }
-        if (serviceId != null) {
-            payload.put("service_id", serviceId);
-        }
-        RequestSpecification requestSpecification = given().port(port)
-                .contentType(JSON)
-                .body(toJson(payload));
-        return requestSpecification;
+    
+    protected String createAGatewayAccountAndExtractAccountId(Map<String, String> createGatewayAccountPayload) {
```
DomBelcher: Is there ever a need to return create the gateway account and extract something other than the account ID? If so `createAGatewayAccount` could just return the account ID by default

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
```diff
@@ -141,82 +72,60 @@ public static void assertCorrectCreateResponse(ValidatableResponse response, Gat
...
+        public static CreateGatewayAccountPayloadBuilder aCreateGatewayAccountPayloadBuilder() {
+            return new CreateGatewayAccountPayloadBuilder();
         }
-
-        public GatewayAccountPayload withServiceName(String serviceName) {
-            this.serviceName = serviceName;
-            return this;
+        
+        public Map<String, String> build() {
+            return payload;
```
DomBelcher: If you return `Map.copyOf(payload)` you can make the resulting map unmodifiable, which motivates proper use of the builder

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
DomBelcher: Also make the `payload` private to ensure it's only accessible via the builder

DomBelcher (changes requested):
Nice work, just a couple of suggestions

src/test/java/uk/gov/pay/connector/it/resources/AgreementsApiResourceIT.java
james-peacock-ho: done

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
james-peacock-ho: I've given that a try but I think some setup is done in the app extension, because it seems no gateway accounts are created without it (maybe it's the db setup?)

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
james-peacock-ho: agree

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
james-peacock-ho: there used to be.. but not any more. thanks

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITBaseExtensions.java
james-peacock-ho: updated as discussed

src/test/java/uk/gov/pay/connector/it/resources/AgreementsApiResourceIT.java
```diff
@@ -323,8 +321,11 @@ class CancelAgreement {
             
             @Test
             void shouldReturn204AndCancelAgreement() {
-                String gatewayAccountId = testBaseExtension.createAGatewayAccountWithServiceId(VALID_SERVICE_ID);
-                testBaseExtension.updateGatewayAccount(gatewayAccountId, "recurring_enabled", true);
+                String gatewayAccountId = testHelpers.createGatewayAccount(
```
DomBelcher: Does this do anything different than the `createGatewayAccount` call in the `@BeforeEach` block?

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceCreateIT.java
```diff
@@ -76,37 +86,56 @@ class PreDegateway {
...
                     .then()
                     .statusCode(201)
                     .contentType(JSON);
         }
         
         @Test
         public void createASandboxGatewayAccount() {
-            ValidatableResponse response = GatewayAccountResourceITBaseExtensions.createAGatewayAccountFor(app.getLocalPort(), "sandbox", "my test service", "analytics");
-            GatewayAccountResourceITBaseExtensions.assertCorrectCreateResponse(response, TEST, "my test service", "analytics", null);
+            //ValidatableResponse response = GatewayAccountResourceITBaseExtensions.createAGatewayAccountFor(app.getLocalPort(), "sandbox", "my test service", "analytics");
```
DomBelcher: Delete this commented out line

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITHelpers.java
```diff
@@ -0,0 +1,126 @@
...
+    public GatewayAccountResourceITHelpers(int appLocalPort) {
+        this.appLocalPort = appLocalPort;
+    }
+
+    public static final String ACCOUNTS_API_URL = "/v1/api/accounts/";
+    public static final String ACCOUNTS_API_SERVICE_ID_URL = "/v1/api/service/{serviceId}/account/{accountType}";
+    public static final String ACCOUNTS_FRONTEND_URL = "/v1/frontend/accounts/";
+    public static final String ACCOUNT_FRONTEND_EXTERNAL_ID_URL = "/v1/frontend/accounts/external-id/";
+    
+    protected String createGatewayAccount(Map<String, String> createGatewayAccountPayload) {
```
DomBelcher: It probably doesn't make much difference but I don't think this needs to be `protected`, `public` should be fine

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceCreateIT.java
james-peacock-ho: oops thanks

src/test/java/uk/gov/pay/connector/it/resources/AgreementsApiResourceIT.java
james-peacock-ho: good point, moving the beforeEach up a level

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceITHelpers.java
james-peacock-ho: ok done

src/test/java/uk/gov/pay/connector/it/resources/AgreementsApiResourceIT.java
DomBelcher: Ah the diff hid that it was in a separate block, I think it works where it is now though

DomBelcher (approved):
Nice work

PR #5507, opened 2024-08-15
PP-13062 handle exception

SandorArpa:
## WHAT YOU DID

An unhandled case on our side: we do not anticipate the response to an order inquiry request will give us a lastEvent for an authorisation error payment but we need to because we could query the order inquiry endpoint if there was a timeout.

Handle <lastEvent>null</lastEvent> Worldpay response case.


marcotranchino (approved):
LGTM 👍

alexbishop1 (changes requested):
Should there be a test for this? (Perhaps not a unit test for this class but a test where it’s used.)

Please change the commit message to not say “exception” since we’re not handling an exception in the Java sense.

PR #5513, opened 2024-08-19
PP-13038 Support toggling corporate exemptions on Connector via REST

marcotranchino:
With this change, we are introducing the ability to toggle corporate exemptions for Worldpay accounts in Connector via REST patch requests.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-13038

## How to test

How should it be reviewed?  What feedback do you need? 

- The new implementation should be very similar to the way we toggle the Worldpay exemption engine flag.
- The second commit refactors a switch statement in order to use the new style with arrow syntax.

### Logging

No changes to logs as we are re-using an existing Patch method.

- [X] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### Documentation

- [X] Updated README.md for any of the following ? NO

* Introduced any new environment variables / removed existing environment variable --> NO
* Added new API / updated existing API definition --> NO


SandorArpa (approved):
LOKTM

PR #5517, opened 2024-08-20
PP-13029 log reason of exemption result

alexbishop1:
When requesting a 3ds exemption, log the `reason` attribute of the `exemptionResponse` if exemption was requested and reason is available.

If no exemption was requested log should be as is.

Add tests.

with @alexbishop1


SandorArpa (approved):
LOKTM

PR #5533, opened 2024-09-03
PP-13154 Stop sending DDC result for samsung browsers

kbottla:
## WHAT YOU DID
- Temporary fix to not send Worldpay 3DS Flex device data collection result when a user is using Samsung Internet 26


SandorArpa (approved):
LOKTM

PR #5534, opened 2024-09-04
Revert "PP-13154 Stop sending DDC result for samsung browsers"

kbottla:
Reverts alphagov/pay-connector#5533

(to merge after CardinalCommerce confirms that we can revert our change)


alexbishop1 (approved):
CardinalCommerce have confirmed their fix is in and working so — with the approval of partner services — we can deploy this revert.

PR #5535, opened 2024-09-04
PP-13051 Send corporate exemption with Worldpay authorisation request for corporate cards if enabled

hjvoid:
## WHAT YOU DID
Ensure that if corporate exemptions are enabled and a corporate card is used, then the charges tabel of the connector database is updatd to the value of optimised and the authorisation request to Worldpay includs `<exemption type="OP" placement="OPTIMISED"/>` element inside the order element

## How to test

- How should it be reviewed? 
    - Check the code.
- What feedback do you need? 
    - I've made the test on Line 223 of `src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java` a parameterised test for readability but realise this may be better as two separate tests?
    - I have read the ticket as 3D secure enabled meaning the `requires_3ds` column of the `gateway_accounts` table.
    - I suspect there's a better way of structuring my changes in [WorldpayAuthoriseOrderTemplate.xml](https://github.com/alphagov/pay-connector/pull/5535/files#diff-a18066368d3d8d932d82af7fd4458986a47b0d41cf495b7432c8497767664976)
    - I'm not sure if the naming of the tests in [WorldpayPaymentProviderTest.java](https://github.com/alphagov/pay-connector/pull/5535/files#diff-d64f06829c2303c7615e41d361e4db01df11be2bc86524770d875079b99d79fd) could be made more explicit.
 
- Steps to test
    - Checkout this branch as well as `pp-13038-toggle-corporate-exemptions-button` pay-toolbox branch (this may not be needed if this branch has been merged by the time of writing). 
    - In `pay-connector` you will need to change the test Worldpay URLs and test threeDsFlexDdcUrls in `config.yaml` to the worldpay test urls. You can find theses in `pay-magic-urls` in pay-infra.
    - Run connector from local branch:
       `pay local launch --cluster endtoend --local connector`
    - Create a user:
       `pay local user`
    - Run toolbox locally and go to the gateway account for the service just created.
    - Go to ‘Manage PSP’ and switch to Worldpay.
    - This will create a worldpay credentials record (in gateway_account_credentials) but the status needs to be changed 
    to `ACTIVE` so you can use APIs or a SQL update. Don't forget to add a`active_start_date`. Alternatively you can run the following query `update gateway_account_credentials set state = 'ACTIVE', active_start_date = now()  where id = 2;`. 
    - Remember to set `requires_3ds` to true in the DB on the gateway account.
    - Now set the `worldpay_3ds_flex_credentials` using the following query `insert into worldpay_3ds_flex_credentials(id, gateway_account_id, issuer, organisational_unit_id, jwt_mac_key, version, exemption_engine, corporate_exemptions) values (1,1, 'issuer', 'organisational_unit_id', 'jwt_mac_key', 1,  false, false);` . The correct values can be found on the Worldpay developer docs.
    - Now if you `select * from worldpay_3ds_flex_credentials` you can toggle the exemption_engine and corporate_exemption properties as necessary. 
    - Now set your account credentials using the selfservice UI by going first to _setttings_ and then to _Your PSP - Worldpay_ menu. You will need to obtain these from someone with access to pay-low-pass. 
    - Now create a new payment link in order to make some test transactions.
    - You can check if `CORPORATE` or `OPTIMISED` has been set if you `select * from charges`
    - You can use the mock card numbers in the card numbers tech docs to test corporate/non-corporate cards https://docs.payments.service.gov.uk/testing_govuk_pay/#mock-card-numbers
    - You should also check that each transaction has a `REQUESTED_3DS_EXEMPTION` event that is either set to `OPTIMISED` or `CORPORATE` dependant upon how you set `exemption_enging` , `corporate_exemptions` and whether you use a corporate card or not.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
@@ -31,6 +31,7 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
         private String acceptHeader;
         private String userAgentHeader;
         private boolean requires3ds;
+        private boolean corporateExemptionEnabled;
         private String paResponse3ds;
         private String payerIpAddress;
         private String payerEmail;
@@ -106,11 +107,17 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
         public boolean isRequires3ds() {
             return requires3ds;
         }
+        
+        public boolean isCorporateExemptionEnabled() { return corporateExemptionEnabled; }
 
         public void setRequires3ds(boolean requires3ds) {
             this.requires3ds = requires3ds;
         }
 
+        public void setCorporateExemptionEnabled(boolean corporateExemptionEnabled) {
+            this.corporateExemptionEnabled = corporateExemptionEnabled;
+        }
+        
         public Optional<String> getPaResponse3ds() {
             return Optional.ofNullable(paResponse3ds);
         }
@@ -293,6 +300,12 @@ public class WorldpayOrderRequestBuilder extends OrderRequestBuilder {
     public WorldpayOrderRequestBuilder with3dsRequired(boolean requires3ds) {
         logger.info("3DS requirement is: " + requires3ds + " for " + worldpayTemplateData.sessionId);
         worldpayTemplateData.setRequires3ds(requires3ds);
+        return this;
+    }
+    
+    public WorldpayOrderRequestBuilder withCorporateExemptionEnabled(boolean corporateExemptionEnabled) {
+        logger.info("Corporate Exemption is: " + corporateExemptionEnabled + " for " + worldpayTemplateData.sessionId);
+        worldpayTemplateData.setCorporateExemptionEnabled(corporateExemptionEnabled);
         return this;
     }
 
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
@@ -7,6 +7,7 @@ import uk.gov.pay.connector.charge.dao.ChargeDao;
 import uk.gov.pay.connector.charge.model.domain.Charge;
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.charge.model.domain.ChargeStatus;
+import uk.gov.pay.connector.charge.model.domain.Exemption3dsType;
 import uk.gov.pay.connector.common.model.api.ExternalChargeRefundAvailability;
 import uk.gov.pay.connector.events.EventService;
 import uk.gov.pay.connector.events.model.charge.Requested3dsExemption;
@@ -62,6 +63,7 @@ import static java.lang.String.format;
 import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 import static java.util.UUID.randomUUID;
+import static uk.gov.pay.connector.charge.model.domain.Exemption3dsType.CORPORATE;
 import static uk.gov.pay.connector.charge.model.domain.Exemption3dsType.OPTIMISED;
 import static uk.gov.pay.connector.gateway.PaymentGatewayName.WORLDPAY;
 import static uk.gov.pay.connector.gateway.util.AuthUtil.getWorldpayAuthHeader;
@@ -208,14 +210,20 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
     public GatewayResponse<WorldpayOrderStatusResponse> authorise(CardAuthorisationGatewayRequest request, ChargeEntity charge) {
 
         boolean exemptionEngineEnabled = isExemptionEngineEnabled(request);
+        boolean corporateExemptionsEnabled = isCorporateExemptionsEnabled(request);
+        boolean cardIsCorporate = isCorporateCard(request);
+        
         GatewayResponse<WorldpayOrderStatusResponse> response;
+        Exemption3dsType exemptionType = corporateExemptionsEnabled && cardIsCorporate ? CORPORATE : OPTIMISED;
 
-        if (!exemptionEngineEnabled) {
-            response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
-        } else {
-            charge = updateChargeWithRequested3dsExemption(charge);
-            response = worldpayAuthoriseHandler.authoriseWithExemption(request);
+        if (exemptionEngineEnabled || corporateExemptionsEnabled) {
+            charge = updateChargeWithRequested3dsExemption(charge, exemptionType);
         }
+        
+        response = exemptionEngineEnabled
+                ? worldpayAuthoriseHandler.authoriseWithExemption(request)
+                : worldpayAuthoriseHandler.authoriseWithoutExemption(request);
+        
 
         calculateAndStoreExemption(exemptionEngineEnabled, charge, response);
 
@@ -292,9 +300,9 @@ public class WorldpayPaymentProvider implements PaymentProvider, WorldpayGateway
     }
 
     @Transactional
-    public ChargeEntity updateChargeWithRequested3dsExemption(ChargeEntity chargeEntity) {
-        chargeEntity.setExemption3dsRequested(OPTIMISED);
-        LOGGER.info("Requesting {} exemption - charge_external_id={}", "OPTIMISED", chargeEntity.getExternalId());
+    public ChargeEntity updateChargeWithRequested3dsExemption(ChargeEntity chargeEntity, Exemption3dsType exemption3dsType) {
+        chargeEntity.setExemption3dsRequested(exemption3dsType);
+        LOGGER.info("Requesting {} exemption - charge_external_id={}", exemption3dsType, chargeEntity.getExternalId());
         eventService.emitAndRecordEvent(Requested3dsExemption.from(chargeEntity, Instant.now()));
 
         return chargeDao.merge(chargeEntity);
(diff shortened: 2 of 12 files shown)


src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -80,10 +80,16 @@
                 challengeWindowSize="390x400" challengePreference="noPreference"
             />
             </#if>
-            <#if exemptionEngineEnabled>
+            <#if authCardDetails.corporateCard>
+            <#if (corporateExemptionEnabled && !exemptionEngineEnabled) || (corporateExemptionEnabled && exemptionEngineEnabled)>
```
marcotranchino: This expression 

```javascript

<#if (corporateExemptionEnabled && !exemptionEngineEnabled) || (corporateExemptionEnabled && exemptionEngineEnabled)>
```

can be simplified with:

```javascript
<#if corporateExemptionEnabled>
```

Unless the business logic should be doing something different?

hjvoid: Good spot!

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
```diff
@@ -295,6 +302,12 @@ public WorldpayOrderRequestBuilder with3dsRequired(boolean requires3ds) {
         worldpayTemplateData.setRequires3ds(requires3ds);
         return this;
     }
+    
+    public WorldpayOrderRequestBuilder withCorporateExemptionEnabled(boolean corporateExemptionEnabled) {
+        logger.info("Corporate Exemption is: " + corporateExemptionEnabled + " for " + worldpayTemplateData.sessionId);
```
marcotranchino, line 307: Hugo, 
Sandor and I are wondering if you meant this log to remain, if we need it, 
or whether it was needed just for testing. What do you think?

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -208,14 +210,20 @@ private GatewayOrder buildQuery(ChargeQueryGatewayRequest request) {
...
-        if (!exemptionEngineEnabled) {
-            response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
-        } else {
-            charge = updateChargeWithRequested3dsExemption(charge);
-            response = worldpayAuthoriseHandler.authoriseWithExemption(request);
+        if (exemptionEngineEnabled || corporateExemptionsEnabled) {
+            charge = updateChargeWithRequested3dsExemption(charge, exemptionType);
         }
+        
+        response = exemptionEngineEnabled
```
marcotranchino: The test should be with **(exemptionEngineEnabled || corporateExemptionsEnabled)** instead of just **exemptionEngineEnabled**.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
hjvoid: Yeah it's part of the AC's:

> And the line Requesting CORPORATE exemption - charge_external_id=abcdefghijklmnopqrstuvwxyz is logged

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -80,10 +80,16 @@
...
             />
             </#if>
-            <#if exemptionEngineEnabled>
+            <#if authCardDetails.corporateCard>
+            <#if (corporateExemptionEnabled)>
+            <exemption type="CP" placement="AUTHORISATION"/>
+            <#elseif !corporateExemptionEnabled && exemptionEngineEnabled>
             <exemption type="OP" placement="OPTIMISED"/>
             </#if>
+            <#elseif !authCardDetails.corporateCard && corporateExemptionEnabled && exemptionEngineEnabled>
```
marcotranchino: We can remove the following from this elseif: **&& corporateExemptionEnabled**
Because we don't care about it because the card is not corporate.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -215,16 +220,25 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
                 gatewayOrderArgumentCaptor.getValue().getPayload());
     }
 
-    @Test
-    void should_include_exemption_element_if_account_has_exemption_engine_set_to_true() throws Exception {
+    @ParameterizedTest
```
marcotranchino, line 223: We think that this is great. No need to separate the tests! 🙌🏼

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
```diff
@@ -477,6 +484,42 @@ void shouldGenerateValidDeleteTokenRequest() throws Exception {
...
+        assertEquals(OrderRequestType.AUTHORISE, actualRequest.getOrderRequestType());
+    }
+
+    private static Stream<Arguments> exemptionValues() {
+        return Stream.of(
+                arguments(true, false, WORLDPAY_VALID_AUTHORISE_3DS_REQUEST_EXEMPTION_AUTHORISATION, true),
+                arguments(true, true, WORLDPAY_VALID_AUTHORISE_3DS_REQUEST_EXEMPTION_AUTHORISATION, true),
+                arguments(false, true, WORLDPAY_VALID_AUTHORISE_3DS_REQUEST_EXEMPTION_OPTIMISED, true),
+                arguments(true, true, WORLDPAY_VALID_AUTHORISE_3DS_REQUEST_EXEMPTION_OPTIMISED, false)
+        );
```
marcotranchino, line 526: Hugo, 
We are thinking that it would be beneficial to have another test to check what happens when it's 

arguments(false, false, NEED_TO_FIGURE_OUT_WHAT_GOES_HERE, true).

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -355,11 +371,66 @@ void should_set_exemption_rejected_when_request_made_with_an_exemption() throws
...
         assertThat(secondRequest.getTransactionId(), not(nullValue()));
         assertThat(secondRequest.getTransactionId(), not(chargeEntity.getGatewayTransactionId()));
-        verifyChargeUpdatedWith3dsAndExemptionRequested(EXEMPTION_REJECTED);
+        verifyChargeUpdatedWith3dsAndExemptionRequested(EXEMPTION_REJECTED, OPTIMISED);
         verifyLoggingWithExemptionReason(EXEMPTION_REJECTED, "HIGH_RISK", chargeEntity.getExternalId(), 3);
         verify3dsExemptionEventsEmitted(chargeEntity, EXEMPTION_REJECTED, OPTIMISED);
     }
 
+    @Test
+    void should_send_worldpay_authorisation_without_exemption_if_exemption_engine_false_and_corporate_card_is_used() throws Exception {
```
marcotranchino: This does not make it clear whether corporate exemption is enabled.
Alex B has confirmed that corporate exemption could be enabled and working even if the exemption engine is disabled.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -355,11 +371,66 @@ void should_set_exemption_rejected_when_request_made_with_an_exemption() throws
...
+        verifyChargeUpdatedWith3dsAndExemptionRequested(EXEMPTION_HONOURED, exemption3dsType);
+        verify3dsExemptionEventsEmitted(chargeEntity, EXEMPTION_HONOURED, exemption3dsType);
+        verifyLoggingTypeOf3dsExemption(exemption3dsType, chargeEntity.getExternalId());
+    }
+
+    private static Stream<Arguments> exemptionValues() {
+        return Stream.of(
+                arguments(true, CORPORATE, true),
+                arguments(false, OPTIMISED, true),
+                arguments(true, OPTIMISED, false)
```
marcotranchino: This is good! 👍🏼

marcotranchino (changes requested):
Great work, @hjvoid ! 🙌 
However we believe that we should consider the case when the corporate exemption is enabled but the exemption engine is disabled.
Alex B has confirmed that, in that case, corporate exemptions should still work.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
marcotranchino: Ah I see, thanks for clarifying! :)

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
hjvoid: This test is being removed as it describes incorrect logic. I have added a new arguments to the `exemptionValues` function in the soon to be renamed `should_send_worldpay_authorisation_with_exemption_if_exemption_engine_or_corporate_exemptions_enabled` test instead.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
hjvoid: I've added the those values to the test above (see line 525) and used a previously used template `WORLDPAY_VALID_AUTHORISE_WORLDPAY_3DS_REQUEST_MIN_ADDRESS` which tallies with the template comparison we need. Alternatively I could make a new template and call it something more descriptive, however, we would end up with 3 templates which are identical.

marcotranchino (approved):
Great work, @hjvoid ! 🙌🏼
All the points raised in the review with @SandorArpa have been addressed. 👍🏼

PR #5547, opened 2024-09-13
PP-12863 Parity check - ignore refund_summary for disputed transactions

kbottla:
## WHAT YOU DID
- For disputed payment, Ledger returns `refund_summary.status` as `unavailable`. But if the charge is not expunged, Connector returns `refund_summary.status` as `available` as the connector doesn't know about disputes and is causing parity check failures.
- So removed checking refund_summary for disputed payments.


marcotranchino (approved):
LGTM 👍

PR #5549, opened 2024-09-17
PP-11559 after async refund transfer send events to ledger

AbdirizakIdris:
## Background
- A Stripe refund succeeded initially but later failed asynchronously. We do not handle this failure case automatically
- Funds for a refund are taken from our Stripe platform account as soon as the request to make the refund succeeds. We then make a transfer from the service’s connect account to our platform account. If the refund later fails, Stripe returns the money to our platform account.
- To correct this, we need to make a new transfer from our platform account to the connect account.
- We then add a new payment into Pay with required details

## WHAT YOU DID
- Added events for `RefundFailureFundsSentToConnectAccount`, `RefundStatusCorrectedToErrorByAdmin` and `PaymentStatusCorrectedToSuccessByAdmin` 
- Called these events in `RefundReversalService` to be posted to Ledger

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/refund/PaymentStatusCorrectedToSuccessByAdminEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/refund/PaymentStatusCorrectedToSuccessByAdminEventDetails.java
@@ -1,0 +1,101 @@
+package uk.gov.pay.connector.events.eventdetails.refund;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import uk.gov.pay.connector.common.model.api.ExternalRefundStatus;
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+import uk.gov.service.payments.commons.api.json.IsoInstantMicrosecondSerializer;
+
+import java.time.Instant;
+
+public class PaymentStatusCorrectedToSuccessByAdminEventDetails extends EventDetails {
+    private final long fee;
+    private final long netAmount;
+    @JsonSerialize(using = IsoInstantMicrosecondSerializer.class)
+    private final Instant capturedDate;
+    private final ExternalRefundStatus refundStatus;
+    private final String updatedReason;
+    private final String adminGithubId;
+    private final long gatewayAccountId;
+    @JsonSerialize(using = IsoInstantMicrosecondSerializer.class)
+    private final Instant captureSubmittedDate;
+    private final long refundAmountRefunded;
+    private final long refundAmountAvailable;
+    private final String reference;
+    private final String zendeskId;
+
+    public PaymentStatusCorrectedToSuccessByAdminEventDetails(
+            long fee,
+            long netAmount,
+            Instant capturedDate,
+            ExternalRefundStatus refundStatus,
+            String updatedReason,
+            String adminGithubId,
+            long gatewayAccountId,
+            Instant captureSubmittedDate,
+            long refundAmountRefunded,
+            long refundAmountAvailable,
+            String reference,
+            String zendeskId) {
+        this.fee = fee;  
+        this.netAmount = netAmount;
+        this.capturedDate = capturedDate;
+        this.refundStatus = refundStatus;
+        this.updatedReason = updatedReason;
+        this.adminGithubId = adminGithubId;
+        this.gatewayAccountId = gatewayAccountId;
+        this.captureSubmittedDate = captureSubmittedDate;
+        this.refundAmountRefunded = refundAmountRefunded;
+        this.refundAmountAvailable = refundAmountAvailable;
+        this.reference = reference;
+        this.zendeskId = zendeskId;
+    }
+
+    public long getFee() {
+        return fee;
+    }
+
+    public long getNetAmount() {
+        return netAmount;
+    }
+
+    public Instant getCapturedDate() {
+        return capturedDate;
+    }
+
+    public ExternalRefundStatus getRefundStatus() {
+        return refundStatus;
+    }
+
+    public String getUpdatedReason() {
+        return updatedReason;
+    }
+
+    public String getAdminGithubId() {
+        return adminGithubId;
+    }
+
+    public long getGatewayAccountId() {
+        return gatewayAccountId;
+    }
+
+    public Instant getCaptureSubmittedDate() {
+        return captureSubmittedDate;
+    }
+
+    public long getRefundAmountRefunded() {
+        return refundAmountRefunded;
+    }
+
+    public long getRefundAmountAvailable() {
+        return refundAmountAvailable;
+    }
+
+
+    public String getReference() {
+        return reference;
+    }
+
+    public String getZendeskId() {
+        return zendeskId;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundFailureFundsSentToConnectAccountEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundFailureFundsSentToConnectAccountEventDetails.java
@@ -1,0 +1,74 @@
+package uk.gov.pay.connector.events.eventdetails.refund;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+public class RefundFailureFundsSentToConnectAccountEventDetails extends EventDetails {
+
+    private final long amount;
+    private final String reference;
+    private final String description;
+    private final String adminGithubId;
+    private final String updatedReason;
+    private final String paymentProvider;
+    private final long gatewayAccountId;
+    private final String gatewayTransactionId;
+    private final String zendeskId;
+
+
+    public RefundFailureFundsSentToConnectAccountEventDetails(
+            long amount,
+            String reference,
+            String description,
+            String adminGithubId,
+            String updatedReason,
+            String paymentProvider,
+            long gatewayAccountId,
+            String gatewayTransactionId,
+            String zendeskId) {
+        this.amount = amount;
+        this.reference = reference;
+        this.description = description;
+        this.adminGithubId = adminGithubId;
+        this.updatedReason = updatedReason;
+        this.paymentProvider = paymentProvider;
+        this.gatewayAccountId = gatewayAccountId;
+        this.gatewayTransactionId = gatewayTransactionId;
+        this.zendeskId = zendeskId;
+    }
+
+    public long getAmount() {
+        return amount;
+    }
+
+    public String getReference() {
+        return reference;
+    }
+
+    public String getDescription() {
+        return description;
+    }
+
+    public String getAdminGithubId() {
+        return adminGithubId;
+    }
+
+    public String getUpdatedReason() {
+        return updatedReason;
+    }
+
+    public String getPaymentProvider() {
+        return paymentProvider;
+    }
+
+    public long getGatewayAccountId() {
+        return gatewayAccountId;
+    }
+
+    public String getGatewayTransactionId() {
+        return gatewayTransactionId;
+    }
+
+    public String getZendeskId() {
+        return zendeskId;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundStatusCorrectedToErrorByAdminEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundStatusCorrectedToErrorByAdminEventDetails.java
@@ -1,0 +1,40 @@
+package uk.gov.pay.connector.events.eventdetails.refund;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+public class RefundStatusCorrectedToErrorByAdminEventDetails extends EventDetails {
+    
+    String updatedReason;
+    String adminGithubId;
+    String zendeskId;
+
+    public RefundStatusCorrectedToErrorByAdminEventDetails( String updatedReason, String adminGithubId, String zendeskId) {
+        this.updatedReason = updatedReason;
+        this.adminGithubId = adminGithubId;
+        this.zendeskId = zendeskId;
+    }
+
+    public String getUpdatedReason() {
+        return updatedReason;
+    }
+
+    public void setUpdatedReason(String updatedReason) {
+        this.updatedReason = updatedReason;
+    }
+
+    public String getAdminGithubId() {
+        return adminGithubId;
+    }
+
+    public void setAdminGithubId(String adminGithubId) {
+        this.adminGithubId = adminGithubId;
+    }
+
+    public String getZendeskId() {
+        return zendeskId;
+    }
+
+    public void setZendeskId(String zendeskId) {
+        this.zendeskId = zendeskId;
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
@@ -1,0 +1,44 @@
+package uk.gov.pay.connector.events.model.refund;
+
+import uk.gov.pay.connector.charge.model.domain.Charge;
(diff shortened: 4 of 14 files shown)


src/main/java/uk/gov/pay/connector/events/eventdetails/refund/PaymentStatusCorrectedToSuccessByAdminEventDetails.java
```diff
@@ -0,0 +1,97 @@
...
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeDeserializer;
+import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeSerializer;
+
+import java.time.Instant;
+import java.time.ZonedDateTime;
+
+public class PaymentStatusCorrectedToSuccessByAdminEventDetails extends EventDetails {
+    private final Long fee;
+    private final Long netAmount;
```
alexbishop1: It does not look like any of these `Long`s will ever be be null, so I think they can be primitive `long`s (this applies to the other `Long`s in the rest of the class).

src/main/java/uk/gov/pay/connector/events/eventdetails/refund/PaymentStatusCorrectedToSuccessByAdminEventDetails.java
```diff
@@ -0,0 +1,97 @@
...
+    @JsonSerialize(using = MicrosecondPrecisionDateTimeSerializer.class)
+    private final ZonedDateTime capturedDate;
```
alexbishop1: It pains me that this is a `ZonedDateTime` and not an `Instant` but every other `EventDetails` class is the same so it’s fine.

src/main/java/uk/gov/pay/connector/events/eventdetails/refund/PaymentStatusCorrectedToSuccessByAdminEventDetails.java
```diff
@@ -0,0 +1,97 @@
...
+    private final ExternalRefundStatus refundStatus;
+    private final String updatedReason;
+    private final String adminGithubId;
+    private final Long gatewayAccountId;
+    @JsonSerialize(using = MicrosecondPrecisionDateTimeSerializer.class)
+    private final ZonedDateTime captureSubmittedDate;
+    private final Long refundAmountRefunded;
+    private final Long refundAmountAvailable;
+    private final String transactionId;
```
alexbishop1: I don’t think the `transactionId` is required: it’s covered by the `resourceExternalId` in the associated event.

It does look like `reference` has been missed though.

src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundFailureFundsSentToConnectAccountEventDetails.java
```diff
@@ -0,0 +1,74 @@
+package uk.gov.pay.connector.events.eventdetails.refund;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+public class RefundFailureFundsSentToConnectAccountEventDetails extends EventDetails {
+
+    private final Long amount;
```
alexbishop1: Again, the `Long`s in this class can be `long`s.

src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundFailureFundsSentToConnectAccountEventDetails.java
```diff
@@ -0,0 +1,74 @@
...
+    private final String reference;
+    private final String description;
+    private final String adminGithubId;
+    private final String updatedReason;
+    private final String paymentProvider;
+    private final Long gatewayAccountId;
+    private final String gatewayTransactionId;
+    private final String transactionId;
```
alexbishop1: Like `PaymentStatusCorrectedToSuccessByAdminEventDetails`, I think this can go.

src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundStatusCorrectedToErrorByAdminEventDetails.java
```diff
@@ -0,0 +1,51 @@
+package uk.gov.pay.connector.events.eventdetails.refund;
+
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+public class RefundStatusCorrectedToErrorByAdminEventDetails extends EventDetails {
+
+    String parentTransactionId;
+    Enum transactionType;
+    String updatedReason;
+    String transactionId;
```
alexbishop1: I think these `parentTransactionId`, `transactionType` and `transactionId` are unnecessary because they will be inferred from the `resourceExternalId` in the associated event.

I think an `adminGithubId` needs to be added though.

src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
```diff
@@ -0,0 +1,43 @@
...
+
+import uk.gov.pay.connector.charge.model.domain.Charge;
+import uk.gov.pay.connector.events.eventdetails.refund.PaymentStatusCorrectedToSuccessByAdminEventDetails;
+import uk.gov.pay.connector.refund.model.domain.Refund;
+
+import java.time.Instant;
+import java.time.ZoneId;
+import java.time.ZonedDateTime;
+
+public class PaymentStatusCorrectedToSuccessByAdmin extends RefundEvent {
```
alexbishop1: Although this even will be sent in the context of managing a refund, it’s actually a payment event so it should extend `PaymentEvent`.

src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
```diff
@@ -0,0 +1,43 @@
...
+    private PaymentStatusCorrectedToSuccessByAdmin(String serviceId, boolean live, Long gatewayAccountId,
+                                                   String resourceExternalId, String parentResourceExternalId,
+                                                   PaymentStatusCorrectedToSuccessByAdminEventDetails eventDetails,
+                                                   Instant timestamp) {
+        super(serviceId, live, gatewayAccountId, resourceExternalId, parentResourceExternalId, eventDetails, timestamp);
+    }
+
+    public static PaymentStatusCorrectedToSuccessByAdmin from(Refund refund, Charge charge, String randomIdGenerator, ZonedDateTime timestamp) {
```
alexbishop1: The `randomIdGenerator` is actually the random ID generated by the generator but by this point it’s the ID of the new payment that represents the transfer. We conventionally refer to a payment ID (or refund ID) in the events code as the `resourceExternalId`.

And please make `timestamp` be an `Instant` rather than a `ZonedDateTime`.

src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
```diff
@@ -0,0 +1,43 @@
...
+                charge.getGatewayAccountId(),
+                refund.getExternalId(),
+                refund.getChargeExternalId(),
+                new PaymentStatusCorrectedToSuccessByAdminEventDetails(
+                        0L,
+                        refund.getAmount(),
+                        timestamp,
+                        refund.getExternalStatus(),
+                        "A refund failed and we returned the recovered funds to the service",
+                        "AdminGithubId (nameGds)",
```
alexbishop1: This probably wants to be the actual GitHub user, which I think we get from toolbox when it calls the endpoint.

src/main/java/uk/gov/pay/connector/events/model/refund/RefundFailureFundsSentToConnectAccount.java
```diff
@@ -0,0 +1,40 @@
+package uk.gov.pay.connector.events.model.refund;
+
+import uk.gov.pay.connector.charge.model.domain.Charge;
+import uk.gov.pay.connector.events.eventdetails.refund.RefundFailureFundsSentToConnectAccountEventDetails;
+import uk.gov.pay.connector.refund.model.domain.Refund;
+
+import java.time.Instant;
+
+public class RefundFailureFundsSentToConnectAccount extends RefundEvent {
```
alexbishop1: Again, this is actually a `PaymentEvent`.

src/main/java/uk/gov/pay/connector/events/model/refund/RefundStatusCorrectedToErrorByAdmin.java
```diff
@@ -0,0 +1,35 @@
...
+                charge.getGatewayAccountId(),
+                refund.getExternalId(),
+                refund.getChargeExternalId(),
+                new RefundStatusCorrectedToErrorByAdminEventDetails(
+                        charge.getGatewayTransactionId(),
+                        ResourceType.PAYMENT,
+                        "Correct refund status to match Stripe",
+                        randomIdGenerator
+                ),
+                ZonedDateTime.now().toInstant()
```
alexbishop1: `Instant.now()` will do what you want.

src/main/java/uk/gov/pay/connector/refund/service/RefundReversalService.java
```diff
@@ -78,5 +84,15 @@ public void reverseFailedRefund(GatewayAccountEntity gatewayAccount, Refund refu
                         format("There was an error trying to create transfer with id: %s from Stripe: %s", refundExternalId, e.getMessage()));
             }
         }
+
+        RandomIdGenerator randomIdGenerator = new RandomIdGenerator();
```
alexbishop1: The `RandomIdGenerator` should be injected into the constructor of this class.

src/test/java/uk/gov/pay/connector/service/RefundReversalServiceTest.java
```diff
@@ -182,11 +200,38 @@ void shouldThrowWebApplicationExceptionForUnexpectedError() throws StripeExcepti
...
+                () -> refundReversalService.reverseFailedRefund(gatewayAccountEntity, refund, charge));
 
         assertThat(thrown.getMessage(),
                 is("There was an error trying to get refund from Stripe with refund id: " + refund.getExternalId()));
 
         verify(mockStripeSDKClient).getRefund(stripeRefundId, isLiveGatewayAccount);
     }
+
+    @Test
+    void shouldCreateCorrectEventsAfterRefundReversalToPostToLedger() throws StripeException {
```
alexbishop1, line 216: There’s an argument to be made that this should part of a complete happy path test but let’s ponder that later.

src/test/java/uk/gov/pay/connector/service/RefundReversalServiceTest.java
```diff
@@ -182,11 +200,38 @@ void shouldThrowWebApplicationExceptionForUnexpectedError() throws StripeExcepti
...
+        refundReversalService.reverseFailedRefund(gatewayAccountEntity, refund, charge);
+
+        ArgumentCaptor<List<Event>> eventsCaptor = ArgumentCaptor.forClass(List.class);
+        verify(mockLedgerService).postEvent(eventsCaptor.capture());
+        List<Event> postedEvents = eventsCaptor.getValue();
+
+        assertThat(postedEvents, hasSize(3));
+        assertTrue(postedEvents.get(0) instanceof RefundFailureFundsSentToConnectAccount);
+        assertTrue(postedEvents.get(1) instanceof PaymentStatusCorrectedToSuccessByAdmin);
+        assertTrue(postedEvents.get(2) instanceof RefundStatusCorrectedToErrorByAdmin);
```
alexbishop1: Could maybe assert on some of the important parts of these events (the IDs, messages etc.) to make sure we’re passing in the right data.

alexbishop1 (changes requested):
This is along the right lines. There’s a few bits where I think we’re extending from the wrong class so fix those first because they’ll have the biggest ripple effects.

src/main/java/uk/gov/pay/connector/events/eventdetails/refund/RefundFailureFundsSentToConnectAccountEventDetails.java
alexbishop1: Looks like these are still `Long` objects rather than `long` primitives.

src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
alexbishop1: This still extends `RefundEvent`, not `PaymentEvent`.

src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
AbdirizakIdris: Not sure how to retrieve GitHub User from toolbox

src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static PaymentStatusCorrectedToSuccessByAdmin from(Refund refund, Charge charge, ZonedDateTime timestamp, GithubAndZendeskCredential githubId) {
+
+        return new PaymentStatusCorrectedToSuccessByAdmin(
+                charge.getServiceId(),
+                charge.isLive(),
+                charge.getGatewayAccountId(),
+                charge.getExternalId(),
+                charge.getExternalId(),
```
alexbishop1: Careful here: one of these should be the refund ID and one should be the charge ID.

src/main/java/uk/gov/pay/connector/refund/resource/RefundReversalResource.java
```diff
@@ -84,14 +88,14 @@ public Response reverseRefund(@Parameter(example = "1", description = "Gateway a
...
+        Refund refund = refundReversalService.findMaybeHistoricRefundByRefundId(refundExternalId)
                 .orElseThrow(() -> new RefundNotFoundRuntimeException(refundExternalId));
 
         if (!refund.getChargeExternalId().equals(chargeExternalId)) {
             throw new RefundNotFoundForChargeException(refundExternalId, chargeExternalId, gatewayAccountId);
         }
-        
-        refundReversalService.reverseFailedRefund(gatewayAccount,refund);
+
+        refundReversalService.reverseFailedRefund(gatewayAccount, refund, charge, githubAndZendeskCredential);
```
alexbishop1: I think might want to pass the GitHub user and the Zendesk ticket ID as separate parameters — they’re only initially together because of how they get sent to the endpoint so we can split them after that.

src/main/java/uk/gov/pay/connector/events/model/refund/PaymentStatusCorrectedToSuccessByAdmin.java
```diff
@@ -0,0 +1,44 @@
...
+public class PaymentStatusCorrectedToSuccessByAdmin extends RefundEvent {
+    private PaymentStatusCorrectedToSuccessByAdmin(String serviceId, boolean live, Long gatewayAccountId,
+                                                   String resourceExternalId,
+                                                   String parentResourceExternalId,
+                                                   PaymentStatusCorrectedToSuccessByAdminEventDetails eventDetails,
+                                                   Instant timestamp) {
+        super(serviceId, live, gatewayAccountId, resourceExternalId, parentResourceExternalId, eventDetails, timestamp);
+    }
+
+    public static PaymentStatusCorrectedToSuccessByAdmin from(Refund refund, Charge charge, ZonedDateTime timestamp, GithubAndZendeskCredential githubId) {
```
alexbishop1: This should take an `Instant`, not a `ZonedDateTime`.

alexbishop1 (changes requested):
I think this is almost there!

alexbishop1 (approved):

PR #5564, opened 2024-10-02
PP-13217 Expunge 0p payments in capture submitted state

kbottla:
## WHAT YOU DID
- For `0p` payments we don't get CAPTURED notifications from Worldpay as payments are never captured. So these charges stay in the connector and are expunged after 90 days which isn't necessary.

  Expunge 0p payments in CAPTURED_SUBMITTED state (after a day) along with other terminal payments.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
+++ b/src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
@@ -61,6 +61,8 @@ public class ChargeExpungeService {
             SYSTEM_CANCEL_SUBMITTED,
             USER_CANCEL_SUBMITTED);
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/expunge/service/ChargeExpungeService.java
```diff
@@ -61,6 +61,8 @@ public class ChargeExpungeService {
             SYSTEM_CANCEL_SUBMITTED,
             USER_CANCEL_SUBMITTED);
 
+    private final List<ChargeStatus> terminalStatesFor0pPayments = List.of(CAPTURE_SUBMITTED);
```
alexbishop1, line 64: All these lists should really be `EnumSet`s.

alexbishop1 (approved):

PR #5568, opened 2024-10-03
PP-13220 Use Instant instead of ZonedDateTime in payout code

alexbishop1:
Use `Instant` instead of `ZonedDateTime` in the (Stripe) payout code because they’re only used as timestamps and time zone support is unnecessary.

This introduces a small behaviour change in that the code will no longer tolerate the `"arrival_date"` or `"created"` properties in the Stripe Payout object being null. However, [Stripe’s docs](https://docs.stripe.com/api/payouts/object) say they should never be null so let’s hope they’re telling the truth.


marcotranchino (approved):
LGTM 👍

PR #5570, opened 2024-10-03
PP-13217 Handler for payments stuck in submitted state

kbottla:
## WHAT
- Added a new task handler to query payment status with the gateway (for charges stuck in the submitted state) and update the charge state to CAPTURED if the gateway returns status as CAPTURED. 
- Any errors returned by the gateway are ignored as for Worldpay payments we get error responses during the maintenance window.  These charges will be added back to the task queue by ChargeExpungeService.

- Mapped Worldpay SETTLED_BY_MERCHANT status to CAPTURED in Connector as we don't have an equivalent for SETTLED status and by the time we query the order, it may be settled.

PR that adds charges to the TaskQueue https://github.com/alphagov/pay-connector/pull/5571


marcotranchino (approved):
LGTM 👍

PR #5573, opened 2024-10-03
PP-13220 Use Instant instead of ZonedDateTime in EventDetails

alexbishop1:
Use `Instant` instead of `ZonedDateTime` in the `EventDetails` subclasses because they’re only used as timestamps and time zone support is unnecessary.


marcotranchino (approved):
LGTM 👍

PR #5575, opened 2024-10-04
PP-13220 Add builder for PaymentInstrumentCreatedDetails

alexbishop1:
Add a builder for `PaymentInstrumentCreatedDetails` so its fields no longer need to be mutable.


marcotranchino (approved):
👍

PR #5578, opened 2024-10-07
PP-13217 Fixes for QueryAndUpdatePaymentInSubmittedStateTaskHandler

kbottla:
## WHAT YOU DID
- `QueryAndUpdatePaymentInSubmittedStateTaskHandler` ignores `WebApplicationException` as this exception is returned when Worldpay is down for maintenance
- Added `@Transactional` annotation to the `process` method to bring ChargeEntity into transactional scope. At present, charge status is **not** being updated to CAPTURED but a new charge event for CAPTURED state is being inserted.

- Also maps Worldpay status `SETTLED` to `CAPTURED` as for some Worldpay payments we are getting `SETTLED` status instead of SETTLED_BY_MERCHANT.


SandorArpa (approved):
LOKTM

PR #5582, opened 2024-10-08
PP-13051 corporate exemption with Worldpay authorisation request

hjvoid:
## WHAT YOU DID
Ensure that if corporate exemptions are enabled and a corporate card is used, then the charges tabel of the connector database is updatd to the value of optimised and the authorisation request to Worldpay includs <exemption type="OP" placement="OPTIMISED"/> element inside the order element

## How to test

- How should it be reviewed?

    - Check the code and test manually if possible.

* What feedback do you need

    - I've made the test on Line 223 of src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java a parameterised test for readability but realise this may be better as two separate tests?

    - I have read the ticket as 3D secure enabled meaning the requires_3ds column of the gateway_accounts table.
    - I suspect there's a better way of structuring my changes in [WorldpayAuthoriseOrderTemplate.xml](https://github.com/alphagov/pay-connector/pull/5535/files#diff-a18066368d3d8d932d82af7fd4458986a47b0d41cf495b7432c8497767664976)
    - I'm not sure if the naming of the tests in [WorldpayPaymentProviderTest.java](https://github.com/alphagov/pay-connector/pull/5535/files#diff-d64f06829c2303c7615e41d361e4db01df11be2bc86524770d875079b99d79fd) could be made more explicit.
    - Added a logger within `WorldpayPaymentProvider.java` for whether a corporate card has been used as per Krishna's suggestion.

* Updates since previous merge
   -  Clean up src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml . Sadly we need to keep <#if corporateExemptionEnabled || exemptionEngineEnabled> on line 83 otherwise the next if statement doesn't kick in.
   -  Added a logger that registers if a corporate card has been used and includes the `external_id`.
   -  `WorldpayOrderBuilder` cannot be tested directly as it’s an interface and technically has no logic to test. I've added two more tests to `WorldpayAuthoriseTestHandlerTest.java` instead to include checks for the `isCorporateExemptionEnabled` boolean value instead.

* Steps to test

    - Checkout this branch as well as pp-13038-toggle-corporate-exemptions-button pay-toolbox branch (this may not be needed if this branch has been merged by the time of writing).
    - In pay-connector you will need to change the test Worldpay URLs and test threeDsFlexDdcUrls in config.yaml to the worldpay test urls. You can find theses in pay-magic-urls in pay-infra.
    - Run connector from local branch:
`pay local launch --cluster endtoend --local connector`
    - Create a user:
`pay local user`
    - Run toolbox locally and go to the gateway account for the service just created.
    - Go to ‘Manage PSP’ and switch to Worldpay.
    - This will create a worldpay credentials record (in gateway_account_credentials) but the status needs to be changed
to ACTIVE so you can use APIs or a SQL update. Don't forget to add an `active_start_date`. Alternatively you can run the following query `update gateway_account_credentials set state = 'ACTIVE', active_start_date = now()  where id = 2;`.
    - Remember to set `requires_3ds` to true in the DB on the gateway account.
    - Now set the `worldpay_3ds_flex_credentials` using the following query insert into `worldpay_3ds_flex_credentials(id, gateway_account_id, issuer, organisational_unit_id, jwt_mac_key, version, exemption_engine, corporate_exemptions) values (1,1, 'issuer', 'organisational_unit_id', 'jwt_mac_key', 1,  false, false); `. The correct values can be found on the Worldpay developer docs.
    - Now if you `select * from worldpay_3ds_flex_credentials` you can toggle the exemption_engine and corporate_exemption properties as necessary.
    - Now set your account credentials using the selfservice UI by going first to settings and then to "Your PSP - Worldpay" menu. You will need to obtain these from someone with access to pay-low-pass.
    - Now create a new payment link in order to make some test transactions.
    - You can check if CORPORATE or OPTIMISED has been set if you `select * from charges`.
    - You can use the mock card numbers in the card numbers tech docs to test corporate/non-corporate cards https://docs.payments.service.gov.uk/testing_govuk_pay/#mock-card-numbers
    - You should also check that each transaction has a `REQUESTED_3DS_EXEMPTION` event that is either set to `OPTIMISED` or `CORPORATE` dependant upon how you set `exemption_engine`, `corporate_exemptions` and whether you use a corporate card or not.


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -208,14 +210,21 @@ private GatewayOrder buildQuery(ChargeQueryGatewayRequest request) {
...
-        if (!exemptionEngineEnabled) {
-            response = worldpayAuthoriseHandler.authoriseWithoutExemption(request);
-        } else {
-            charge = updateChargeWithRequested3dsExemption(charge);
-            response = worldpayAuthoriseHandler.authoriseWithExemption(request);
+        if (exemptionEngineEnabled || corporateExemptionsEnabled) {
+            charge = updateChargeWithRequested3dsExemption(charge, exemptionType);
         }
+        
+        response = exemptionEngineEnabled
```
SandorArpa: How about corporate exemption? Shouldn't this variable be corporateExemptionsOrExemptionEngineEnabled?

SandorArpa (dismissed):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
marcotranchino: I have made the change. I need to look into failing tests now.

marcotranchino: The changes discussed in this conversation have now been removed after an additional chat in video call between me, @SandorArpa and @alexbishop1

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderBuilder.java
```diff
@@ -43,10 +44,16 @@ static WorldpayOrderRequestBuilder buildAuthoriseOrderWithoutExemptionEngine(Wor
 
         boolean is3dsRequired = request.getAuthCardDetails().getWorldpay3dsFlexDdcResult().isPresent() ||
                 request.getGatewayAccount().isRequires3ds();
+        
+        boolean isCorporateExemptionEnabled = request.getAuthCardDetails().getWorldpay3dsFlexDdcResult().isPresent() &&
+                request.getGatewayAccount().getWorldpay3dsFlexCredentials().map(Worldpay3dsFlexCredentials::isCorporateExemptionsEnabled)
+                        .orElse(false);
```
alexbishop1: I’m not sure this logic is quite right. If the gateway account has corporate exemptions enabled (the right side of the expression) then that should be enough. The DDC result only comes into play when we think the request might result in 3DS being required, which it won’t be if we’ve requested an exemption (we’ll get a soft decline instead).

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilder.java
```diff
@@ -295,6 +302,12 @@ public WorldpayOrderRequestBuilder with3dsRequired(boolean requires3ds) {
         worldpayTemplateData.setRequires3ds(requires3ds);
         return this;
     }
+    
+    public WorldpayOrderRequestBuilder withCorporateExemptionEnabled(boolean corporateExemptionEnabled) {
+        logger.info("Corporate Exemption is: " + corporateExemptionEnabled + " for " + worldpayTemplateData.sessionId);
```
alexbishop1, line 309: I assume `sessionId` is the GOV.UK Pay payment ID? It’s not great to rely on that but I see we’re doing it elsewhere in this class as well.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -208,14 +210,20 @@ private GatewayOrder buildQuery(ChargeQueryGatewayRequest request) {
     public GatewayResponse<WorldpayOrderStatusResponse> authorise(CardAuthorisationGatewayRequest request, ChargeEntity charge) {
 
         boolean exemptionEngineEnabled = isExemptionEngineEnabled(request);
+        boolean corporateExemptionsEnabled = isCorporateExemptionsEnabled(request);
+        boolean cardIsCorporate = isCorporateCard(request, charge);
+        
         GatewayResponse<WorldpayOrderStatusResponse> response;
+        Exemption3dsType exemptionType = corporateExemptionsEnabled && cardIsCorporate ? CORPORATE : OPTIMISED;
```
alexbishop1: Although this will work, it feels a bit not quite right because it will set the `exemptionType` to `OPTIMISED` if no exemption is being requested.

I think we might want to do something like this instead (this code basically replaces the entire first part of the method):

```java
boolean corporateExemptionsEnabledAndCorporateCardUsed = isCorporateExemptionsEnabled(request) && isCorporateCard(request, charge);
boolean exemptionEngineEnabled = isExemptionEngineEnabled(request);

if (corporateExemptionsEnabledAndCorporateCardUsed) {
    charge = updateChargeWithRequested3dsExemption(charge, Exemption3dsType.CORPORATE);
} else if (exemptionEngineEnabled) {
    charge = updateChargeWithRequested3dsExemption(charge, Exemption3dsType.OPTIMISED);
}

GatewayResponse<WorldpayOrderStatusResponse> response = …
```

A further advantage of this is we now have a `corporateExemptionsEnabledAndCorporateCardUsed` boolean we could potentially pass down through the `WorldpayOrderRequestBuilder` into the template, meaning we would not have to repeat logic and reducing the chances of our code that records the exemption type differing from the code that actually sends the exemption.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -306,6 +314,19 @@ private boolean isExemptionEngineEnabled(CardAuthorisationGatewayRequest request
...
+    private boolean isCorporateExemptionsEnabled(CardAuthorisationGatewayRequest request) {
+        GatewayAccountEntity gatewayAccount = request.getGatewayAccount();
+        return gatewayAccount.isRequires3ds() && gatewayAccount.getWorldpay3dsFlexCredentials()
+                .map(Worldpay3dsFlexCredentials::isCorporateExemptionsEnabled)
+                .orElse(false);
+    }
+    
+    private boolean isCorporateCard(CardAuthorisationGatewayRequest request, ChargeEntity charge) {
+        AuthCardDetails authCardDetails = request.getAuthCardDetails();
+        LOGGER.info("Card is corporate card: {} for charge_external_id={}", authCardDetails.isCorporateCard(), charge.getExternalId());
```
alexbishop1: Not sure the logging should happen here; perhaps after this method is called.

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -80,10 +80,14 @@
...
             </#if>
-            <#if exemptionEngineEnabled>
+            <#if corporateExemptionEnabled || exemptionEngineEnabled>
+            <#if corporateExemptionEnabled && authCardDetails.corporateCard>
+            <exemption type="CP" placement="AUTHORISATION"/>
+            <#else>
             <exemption type="OP" placement="OPTIMISED"/>
             </#if>
             </#if>
+        </#if>
```
alexbishop1: Seems like this could just be:

```
<#if corporateExemptionEnabled && authCardDetails.corporateCard>
<exemption type="CP" placement="AUTHORISATION"/>
<#elseif exemptionEngineEnabled>
<exemption type="OP" placement="OPTIMISED"/>
</#if>
```

(The first `#if` could be `<#if corporateExemptionsEnabledAndCorporateCardUsed>` if my suggestion above is implemented.)

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -215,16 +220,87 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
 
-    @Test
-    void should_include_exemption_element_if_account_has_exemption_engine_set_to_true() throws Exception {
+    @ParameterizedTest
+    @MethodSource("exemptionValues")
+    void should_include_exemption_element_if_account_has_exemption_engine_set_to_true(
+            boolean corporateExemption,
+            String exemptionType,
+            String exemptionPlacement,
+            boolean corporateCard
```
alexbishop1: Can these be named to make it clearer with are inputs and which are outputs? So `corporateExemption` would be `corporateExemptionEnabled` and `exemptionType` would be `expectedExemptionType` etc. Maybe also reorder them so the inputs are first followed by the outputs.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -215,16 +220,87 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
+
+        Document document = XPathUtils.getDocumentXmlString(gatewayOrderArgumentCaptor.getValue().getPayload());
+        XPath xPath = XPathFactory.newInstance().newXPath();
+        assertThat(xPath.evaluate("/paymentService/submit/order/exemption/@type", document), is(exemptionType));
+        assertThat(xPath.evaluate("/paymentService/submit/order/exemption/@placement", document), is(exemptionPlacement));
+    }
+
+    @ParameterizedTest
+    @MethodSource("flex3dsCredentialsForCardAndAccount")
+    void should_not_set_corporate_exemption_unless_gateway_account_has_3dsFlex_credentials_and_card_details_has_3ds_flex_Ddc_result(
```
alexbishop1: As discussed above, the logic of requiring a DDC result does not seem quite right.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -745,6 +821,21 @@ void should_throw_illegal_argument_exception_when_no_payment_instrument_present(
...
+        );
+    }
+
+    private static Stream<Arguments> exemptionValues() {
+        return Stream.of(
+                arguments(true, "CP", "AUTHORISATION", true),
+                arguments(true, "OP", "OPTIMISED", false),
+                arguments(false, "OP", "OPTIMISED", true)
+        );
+    }
```
alexbishop1: Since these are only used for one test each, I wonder if [`@CsvSource`](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests-sources-CsvSource) may be more readable, due to keeping the arguments close to the tests.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
```diff
@@ -477,6 +484,42 @@ void shouldGenerateValidDeleteTokenRequest() throws Exception {
         assertEquals(OrderRequestType.DELETE_STORED_PAYMENT_DETAILS, actualRequest.getOrderRequestType());
     }
 
+    @ParameterizedTest
+    @MethodSource("exemptionValues")
+    void shouldGenerateValidAuthoriseOrderRequestAndIncludeCorrectExemptionResponse(Boolean corporateExemptions, Boolean exemptionEngine, String testTemplate, Boolean corporateCard) throws Exception {
```
alexbishop1: Again, I think this could be rejigged to make the inputs and outputs clearer.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -355,11 +381,68 @@ void should_set_exemption_rejected_when_request_made_with_an_exemption() throws
...
         assertThat(secondRequest.getTransactionId(), not(chargeEntity.getGatewayTransactionId()));
-        verifyChargeUpdatedWith3dsAndExemptionRequested(EXEMPTION_REJECTED);
-        verifyLoggingWithExemptionReason(EXEMPTION_REJECTED, "HIGH_RISK", chargeEntity.getExternalId(), 3);
+        verifyChargeUpdatedWith3dsAndExemptionRequested(EXEMPTION_REJECTED, OPTIMISED);
+        verifyLoggingWithExemptionReason(EXEMPTION_REJECTED, "HIGH_RISK", chargeEntity.getExternalId(), 4);
         verify3dsExemptionEventsEmitted(chargeEntity, EXEMPTION_REJECTED, OPTIMISED);
     }
 
+    @Test
+    void should_send_corporate_exemption_with_worldpay_authorisation_if_corporate_exemptions_disabled_and_corporate_card_is_used() throws Exception {
```
alexbishop1: This seems wrong: we should never be requesting a corporate exemption if corporate exemptions are disabled!

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
SandorArpa: Do we want to always log it or if corporate card only? Or if corporate engine enabled and corporate card?

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
SandorArpa: I think we should implement it in PP-13276. That's where we are simplifying things and planned to pass something like `corporateExemptionsEnabledAndCorporateCardUsed`

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
SandorArpa: It shall be replaced by CsvSource

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
alexbishop1: Probably the latter but the former could also be useful.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProvider.java
```diff
@@ -208,15 +210,24 @@ private GatewayOrder buildQuery(ChargeQueryGatewayRequest request) {
     public GatewayResponse<WorldpayOrderStatusResponse> authorise(CardAuthorisationGatewayRequest request, ChargeEntity charge) {
 
         boolean exemptionEngineEnabled = isExemptionEngineEnabled(request);
-        GatewayResponse<WorldpayOrderStatusResponse> response;
+        boolean corporateExemptionsEnabled = isCorporateExemptionsEnabled(request);
+        boolean cardIsCorporate = isCorporateCard(request, charge);
+        if (cardIsCorporate) {
+            LOGGER.info("Card is corporate card and corporate exemption is enabled {}: charge_external_id={}", corporateExemptionsEnabled, charge.getExternalId());
+        }
```
alexbishop1: I think outputting `corporate exemption is enabled false` might be a little unclear — can we reword it?

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
alexbishop1: Actually, I think we should change this now. Currently it looks like if corporate exemptions are enabled and the Exemption Engine is not enabled but it is not a corporate card it will still request an Exemption Engine exemption!

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -215,16 +221,34 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
                 gatewayOrderArgumentCaptor.getValue().getPayload());
     }
 
-    @Test
-    void should_include_exemption_element_if_account_has_exemption_engine_set_to_true() throws Exception {
+    @ParameterizedTest
+    @CsvSource(textBlock =
+            """
```
alexbishop1: You can actually do this:

```java
@CsvSource(textBlock = """
        true, true, CP, AUTHORISATION,
        …
        """)
```

Not that it really matters.

You can also add `useHeadersInDisplayName = true` and then add a header row, which could be useful to make it clearer what the boolean values mean (the headers also appear in the test output when it lists the permutations).

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -215,16 +221,34 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
+                true, true, CP, AUTHORISATION,
+                true, false, OP, OPTIMISED,
+                false, true, OP, OPTIMISED,
+                false, false, OP, OPTIMISED
+            """
+    )
+    void should_include_the_correct_exemption_element_when_account_exemption_engine_is_enabled(
+            boolean corporateExemptionEnabled,
+            boolean corporateCard,
```
alexbishop1: Calling this `corporateCardNumberEntered` might be a tad clearer.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -215,16 +221,34 @@ void should_include_3ds_elements_without_ip_address() throws Exception {
...
+            String expectedExemptionType,
+            String exemptionPlacement) throws Exception {
```
alexbishop1: This could be `expectedExemptionPlacement`.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
alexbishop1: It does not appear to have been replaced by `CsvSource`.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -355,11 +382,70 @@ void should_set_exemption_rejected_when_request_made_with_an_exemption() throws
...
+        verifyEventEmitted(chargeEntity, EXEMPTION_NOT_REQUESTED);
+    }
+
+    @ParameterizedTest
+    @CsvSource(textBlock =
+            """
+               true, true, CORPORATE, true,
+               false, true, OPTIMISED, true
+            """
+    )
```
alexbishop1: Since there’s only two permutations, I think this might be best done as two separate tests.

alexbishop1 (changes requested):
Getting there…

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -355,11 +382,87 @@ void should_set_exemption_rejected_when_request_made_with_an_exemption() throws
...
+        when(worldpayAuthoriseHandler.authoriseWithExemption(cardAuthRequest))
+                .thenReturn(getGatewayResponse(WORLDPAY_EXEMPTION_REQUEST_HONOURED_RESPONSE));
+
+        worldpayPaymentProvider.authorise(cardAuthRequest, chargeEntity);
+
+        assertThat(chargeEntity.getExemption3dsRequested(), is(CORPORATE));
+
+        verifyChargeUpdatedWith3dsAndExemptionRequested(EXEMPTION_HONOURED, CORPORATE);
+        verify3dsExemptionEventsEmitted(chargeEntity, EXEMPTION_HONOURED, CORPORATE);
+        verifyLoggingTypeOf3dsExemption(CORPORATE, chargeEntity.getExternalId());
```
alexbishop1, line 433: Is there anything we can do here to check it’s not sending the Exemption Engine exemption?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -355,11 +382,87 @@ void should_set_exemption_rejected_when_request_made_with_an_exemption() throws
...
+    }
+
+    @Test
+    void should_not_send_corporate_exemption_with_worldpay_authorisation_if_corporate_exemption_is_not_enabled_and_exemption_engine_enabled_and_corporate_card_used() throws Exception {
```
alexbishop1: Although `should_not_send_corporate_exemption` is true, it’s more objectively `should_send_optimised_exemption`.

src/test/java/uk/gov/pay/connector/util/TestTemplateResourceLoader.java
```diff
@@ -53,6 +53,8 @@ public class TestTemplateResourceLoader {
     public static final String WORLDPAY_VALID_AUTHORISE_WORLDPAY_GOOGLE_PAY_3DS_REQUEST_WITH_IP_ADDRESS = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-request-google-pay-3ds-with-ip-address.xml";
     public static final String WORLDPAY_VALID_AUTHORISE_WORLDPAY_GOOGLE_PAY_3DS_REQUEST_WITH_EMAIL = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-request-google-pay-3ds-with-email.xml";
     public static final String WORLDPAY_VALID_AUTHORISE_REQUEST_3DS_FLEX_NON_JS = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-request-3ds-flex-non-js.xml";
+    public static final String WORLDPAY_VALID_AUTHORISE_3DS_REQUEST_EXEMPTION_OPTIMISED = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-3ds-request-exemption-optimised.xml";
+    public static final String WORLDPAY_VALID_AUTHORISE_3DS_REQUEST_EXEMPTION_AUTHORISATION = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-3ds-request-exemption-authorisation.xml";
```
alexbishop1: Both this constant and the filename should have ‘corporate’ in there somewhere.

alexbishop1 (changes requested):
Looking good. Just a couple of final things.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
SandorArpa: Not really, we cannot stub WorldpayOrderBuilder where the use is determined.

alexbishop1 (approved):

PR #5590, opened 2024-10-15
PP-13202 fix json include

SandorArpa:
## WHAT YOU DID

`AuthorisationSummary`'s @JsonInclude(Include.NON_NULL) does not exclude version from the JSON object when it's null.
- add @JsonInclude(Include.NON_EMPTY) on variable level.


hjvoid (dismissed):
LGTM😎

hjvoid (approved):
Wow - you wrote them quickly 😂. Looks good to me!

PR #5593, opened 2024-10-18
PP-13247 Set idempotency key for async failed refund fix and other improvements

AbdirizakIdris:
## Background

- A Stripe refund succeeded initially but later failed asynchronously. We do not handle this failure case automatically
Funds for a refund are taken from our Stripe platform account as soon as the request to make the refund succeeds. We then make a transfer from the service’s connect account to our platform account. If the refund later fails, Stripe returns the money to our platform account.
- To correct this, we need to make a new transfer from our platform account to the connect account.
We then add a new payment into Pay with required details

## WHAT YOU DID
 - Set the idempotency key for each request to correct async failed refund
 - Added try-catch to catch exceptions if new payment and refund events can't be posted to Ledger
 - Added loggging for posting refunds and new payment events to Ledger


marcotranchino (approved):
These changes only affect Idris's new functionality.
We paired on this and are happy that it's safe to merge.

PR #5594, opened 2024-10-18
PP-13264 Only require service ID in degatewayed updateServiceName endpoint 

DomBelcher:
## WHAT YOU DID
 - Change `PATCH /v1/frontend/service/{serviceId}/account/{accountType}/servicename` endpoint to `PATCH /v1/frontend/service/{serviceId}/servicename`
 - Endpoint will change service name in both Test and Live accounts (if a live account exists)
 - Will 404 if no test account exists, will do nothing if no live account exists

## How to test
 - Run tests

### DIFF
--- a/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceIT.java
+++ b/src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceIT.java
@@ -191,7 +191,6 @@ public class GatewayAccountFrontendResourceIT {
 
     @Nested
     class UpdateServiceNameByServiceId {
-        private Map<String, String> gatewayAccountRequest = Map.of();
         private Map<String, String> testGatewayAccountRequest;
         private Map<String, String> liveGatewayAccountRequest;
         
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountFrontendResourceIT.java
```diff
@@ -190,43 +190,127 @@ void shouldReturn404IfGatewayAccountIsNotNumeric() {
...
+            // attempt to update service name
+            app.givenSetup().accept(JSON)
+                    .body(Map.of("service_name", "New Service Name"))
+                    .patch(format("/v1/frontend/service/%s/servicename", serviceId))
+                    .then()
+                    .statusCode(404);
+        }
+
+        @Test
+        void shouldReturn404_whenNoTestAccountExists()
```
oswaldquek: I'm not sure about this. The endpoint is saying to update service name for all gateway accounts relating to that service. If there's no test service, I think the right thing to do is to update the service name of the live service anyway.

DomBelcher: My reasoning was there shouldn't ever be a service without a test account, although that may be an incorrect assumption. I could change it to only `404` if there's neither a test or live account though

DomBelcher: I've changed this to only return a `404` if no gateway accounts exist with the requested service ID

oswaldquek (approved):

PR #5629, opened 2024-11-07
PP-13240 set requires_3ds to false

SandorArpa:
## WHAT YOU DID

requires_3ds set to false and GATEWAY_DOES_NOT_REQUIRE_3DS_AUTHORISATION event emitted when transitioning to AUTHORISATION SUCCESS or AUTHORISATION REJECTED if if requires_3ds is not already true

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -71,6 +71,7 @@ import uk.gov.pay.connector.events.EventService;
 import uk.gov.pay.connector.events.eventdetails.charge.RefundAvailabilityUpdatedEventDetails;
 import uk.gov.pay.connector.events.model.agreement.AgreementInactivated;
 import uk.gov.pay.connector.events.model.charge.Gateway3dsInfoObtained;
+import uk.gov.pay.connector.events.model.charge.GatewayDoesNotRequire3dsAuthorisation;
 import uk.gov.pay.connector.events.model.charge.PaymentDetailsEntered;
 import uk.gov.pay.connector.events.model.charge.PaymentDetailsSubmittedByAPI;
 import uk.gov.pay.connector.events.model.charge.PaymentDetailsTakenFromPaymentInstrument;
@@ -108,6 +109,7 @@ import javax.ws.rs.core.UriInfo;
 import java.net.URI;
 import java.time.Duration;
 import java.time.Instant;
+import java.time.InstantSource;
 import java.time.ZonedDateTime;
 import java.util.HashMap;
 import java.util.List;
@@ -131,6 +133,7 @@ import static uk.gov.pay.connector.charge.model.domain.ChargeEntity.TelephoneCha
 import static uk.gov.pay.connector.charge.model.domain.ChargeEntity.WebChargeEntityBuilder.aWebChargeEntity;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_ERROR;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_REJECTED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_USER_NOT_PRESENT_QUEUED;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AWAITING_CAPTURE_REQUEST;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.CAPTURED;
@@ -178,6 +181,7 @@ public class ChargeService {
 
     private final Boolean rejectPaymentLinkPaymentsWithCardNumberInReference;
     private final CardidService cardidService;
+    private final InstantSource instantSource;
 
     @Inject
     public ChargeService(TokenDao tokenDao,
@@ -200,7 +204,8 @@ public class ChargeService {
                          IdempotencyDao idempotencyDao,
                          ExternalTransactionStateFactory externalTransactionStateFactory,
                          ObjectMapper objectMapper,
-                         CardidService cardidService) {
+                         CardidService cardidService,
+                         InstantSource instantSource) {
         this.tokenDao = tokenDao;
         this.chargeDao = chargeDao;
         this.chargeEventDao = chargeEventDao;
@@ -225,6 +230,7 @@ public class ChargeService {
         this.worldpay3dsFlexJwtService = worldpay3dsFlexJwtService;
         this.rejectPaymentLinkPaymentsWithCardNumberInReference = config.getRejectPaymentLinkPaymentsWithCardNumberInReference();
         this.cardidService = cardidService;
+        this.instantSource = instantSource;
     }
 
     @Transactional
@@ -1008,6 +1014,20 @@ public class ChargeService {
         ChargeEntity charge = findChargeByExternalId(externalId);
         ChargeStatus status = ChargeStatus.fromString(charge.getStatus());
         return status == CAPTURED || status == CAPTURE_SUBMITTED;
+    }
+
+    @Transactional
+    public ChargeEntity updateRequires3dsPostAuthorisation(String externalId) {
+        return chargeDao.findByExternalId(externalId)
+                .map(chargeEntity -> {
+                    if (chargeEntity.getChargeStatus() == AUTHORISATION_SUCCESS || chargeEntity.getChargeStatus() == AUTHORISATION_REJECTED) {
+                        if (!Boolean.TRUE.equals(chargeEntity.getRequires3ds())) {
+                            chargeEntity.setRequires3ds(false);
+                            eventService.emitAndRecordEvent(GatewayDoesNotRequire3dsAuthorisation.from(chargeEntity, instantSource.instant()));
+                        }
+                    }
+                    return chargeEntity;
+                }).orElseThrow(() -> new ChargeNotFoundRuntimeException(externalId));
     }
 
     public int count3dsRequiredEvents(String externalId) {
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/GatewayDoesNotRequire3dsAuthorisationEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/GatewayDoesNotRequire3dsAuthorisationEventDetails.java
@@ -1,0 +1,27 @@
+package uk.gov.pay.connector.events.eventdetails.charge;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+import java.util.Objects;
+
+public class GatewayDoesNotRequire3dsAuthorisationEventDetails extends EventDetails {
+
+    private final boolean requires3DS = false;
+
+    @JsonProperty("requires_3ds")
+    public boolean isRequires3DS() {
+        return requires3DS;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return o instanceof GatewayDoesNotRequire3dsAuthorisationEventDetails &&
+                this.requires3DS == ((GatewayDoesNotRequire3dsAuthorisationEventDetails) o).requires3DS;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(requires3DS);
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
@@ -221,6 +221,8 @@ public class CardAuthoriseService {
                 mayBeCanRetry.orElse(null),
                 mayBeRejectedReason.orElse(null));
 
+        updatedCharge = chargeService.updateRequires3dsPostAuthorisation(updatedCharge.getExternalId());
+
         var authorisationRequestSummary = generateAuthorisationRequestSummary(charge, authCardDetails);
 
         authorisationLogger.logChargeAuthorisation(
--- a/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
@@ -31,8 +31,8 @@ import uk.gov.service.payments.commons.model.CardExpiryDate;
 import java.util.Optional;
 
 import static java.lang.String.format;
-import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_3DS_REQUIRED;
+import static uk.gov.pay.connector.charge.model.domain.ChargeStatus.AUTHORISATION_SUCCESS;
 
 public class WalletAuthoriseService {
     
@@ -75,16 +75,10 @@ public class WalletAuthoriseService {
             try {
 
                 LOGGER.info("Authorising charge for {}", walletAuthorisationRequest.getWalletType().toString());
-                switch (walletAuthorisationRequest.getWalletType()) {
-                    case APPLE_PAY:
-                        operationResponse = authoriseApplePay(charge, walletAuthorisationRequest);
-                        break;
-                    case GOOGLE_PAY:
-                        operationResponse = authoriseGooglePay(charge, walletAuthorisationRequest);
-                        break;
-                    default:
-                        throw new UnsupportedOperationException(format("Wallet type %s not recognised", walletAuthorisationRequest.getWalletType()));
-                }
+                operationResponse = switch (walletAuthorisationRequest.getWalletType()) {
+                    case APPLE_PAY -> authoriseApplePay(charge, walletAuthorisationRequest);
+                    case GOOGLE_PAY -> authoriseGooglePay(charge, walletAuthorisationRequest);
+                };
 
                 if (operationResponse.getBaseResponse().isPresent()) {
                     requestStatus = "success";
@@ -189,6 +183,8 @@ public class WalletAuthoriseService {
                 walletAuthorisationRequest.getWalletType(),
                 walletAuthorisationRequest.getPaymentInfo().getEmail(),
                 auth3dsRequiredDetails);
+
+        updatedCharge = chargeService.updateRequires3dsPostAuthorisation(updatedCharge.getExternalId());
 
         metricRegistry.counter(String.format(
                 "gateway-operations.%s.%s.%s.authorise.result.%s",
--- a/src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
@@ -48,6 +48,7 @@ import uk.gov.pay.connector.common.model.api.ErrorResponse;
 import uk.gov.pay.connector.common.model.api.ExternalTransactionStateFactory;
 import uk.gov.pay.connector.events.EventService;
 import uk.gov.pay.connector.events.model.Event;
+import uk.gov.pay.connector.events.model.charge.GatewayDoesNotRequire3dsAuthorisation;
 import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gateway.model.AuthCardDetails;
@@ -81,6 +82,7 @@ import uk.gov.service.payments.commons.model.AuthorisationMode;
 import uk.gov.service.payments.commons.model.CardExpiryDate;
 
 import java.time.Instant;
+import java.time.InstantSource;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Supplier;
@@ -106,6 +108,7 @@ import static org.mockito.ArgumentMatchers.isNull;
 import static org.mockito.Mockito.anyString;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
@@ -211,6 +214,8 @@ class CardAuthoriseServiceTest extends CardServiceTest {
     @Captor
     private ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor;
 
+    private final InstantSource fixedInstantSource = InstantSource.fixed(Instant.parse("2024-11-11T10:07:00Z"));
+
     private CardAuthoriseService cardAuthorisationService;
     private final CardDetailsEntity cardDetailsEntity = new CardDetailsEntity(
             LastDigitsCardNumber.of("1234"),
@@ -232,7 +237,8 @@ class CardAuthoriseServiceTest extends CardServiceTest {
                 null, null, null, mockConfiguration, null,
                 stateTransitionService, ledgerService, mockRefundService, mockEventService, mockPaymentInstrumentService,
                 mockGatewayAccountCredentialsService, mockAuthCardDetailsToCardDetailsEntityConverter,
-                mockTaskQueueService, mockWorldpay3dsFlexJwtService, mockIdempotencyDao, mockExternalTransactionStateFactory, objectMapper, null);
+                mockTaskQueueService, mockWorldpay3dsFlexJwtService, mockIdempotencyDao, mockExternalTransactionStateFactory,
+                objectMapper, null, fixedInstantSource);
 
         LinkPaymentInstrumentToAgreementService linkPaymentInstrumentToAgreementService = mock(LinkPaymentInstrumentToAgreementService.class);
         CaptureQueue captureQueue = mock(CaptureQueue.class);
@@ -301,9 +307,11 @@ class CardAuthoriseServiceTest extends CardServiceTest {
         assertThat(response.getAuthoriseStatus().get(), is(AuthoriseStatus.AUTHORISED));
 
         ArgumentCaptor<Event> eventCaptor = ArgumentCaptor.forClass(Event.class);
-        verify(mockEventService, times(1)).emitAndRecordEvent(eventCaptor.capture());
-        assertThat(eventCaptor.getValue().getResourceExternalId(), is(charge.getExternalId()));
-        assertThat(eventCaptor.getValue().getEventType(), is("PAYMENT_DETAILS_ENTERED"));
+        verify(mockEventService, times(2)).emitAndRecordEvent(eventCaptor.capture());
+        assertThat(eventCaptor.getAllValues().get(0).getResourceExternalId(), is(charge.getExternalId()));
+        assertThat(eventCaptor.getAllValues().get(0).getEventType(), is("PAYMENT_DETAILS_ENTERED"));
+
+        verifyGatewayDoesNotRequire3dsEventWasEmitted(charge);
     }
 
     @Test
@@ -326,6 +334,8 @@ class CardAuthoriseServiceTest extends CardServiceTest {
         assertThat(charge.get3dsRequiredDetails(), is(nullValue()));
         assertThat(charge.getCardDetails(), is(cardDetailsEntity));
         assertThat(charge.getCorporateSurcharge().isPresent(), is(false));
+
+        verifyGatewayDoesNotRequire3dsEventWasEmitted(charge);
     }
 
     @Test
@@ -354,6 +364,9 @@ class CardAuthoriseServiceTest extends CardServiceTest {
         assertThat(charge.get3dsRequiredDetails(), is(nullValue()));
         assertThat(charge.getCardDetails(), is(cardDetailsEntity));
         assertThat(charge.getCorporateSurcharge().isPresent(), is(false));
+        assertThat(charge.getRequires3ds(), is(false));
+
+        verifyGatewayDoesNotRequire3dsEventWasEmitted(charge);
     }
 
     @Test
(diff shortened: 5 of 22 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -995,6 +995,16 @@ public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus
...
     }
 
+    @Transactional
+    public void updateRequires3ds(String chargeExternalId, Boolean requires3ds) {
+        Optional<ChargeEntity> charge = chargeDao.findByExternalId(chargeExternalId);
+        charge.map(chargeEntity -> {
+            chargeEntity.setRequires3ds(requires3ds);
+            chargeDao.merge(chargeEntity);
+            return chargeEntity;
+        });
```
alexbishop1:
```java
chargeDao.findByExternalId(chargeExternalId).ifPresent(chargeEntity -> {
    chargeEntity.setRequires3ds(requires3ds);
    chargeDao.merge(chargeEntity);
});
```

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/GatewayDoesNotRequires3dsAuthorisationEventDetails.java
```diff
@@ -0,0 +1,26 @@
+package uk.gov.pay.connector.events.eventdetails.charge;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import uk.gov.pay.connector.charge.model.domain.Auth3dsRequiredEntity;
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.events.eventdetails.EventDetails;
+
+import java.util.Optional;
+
+public class GatewayDoesNotRequires3dsAuthorisationEventDetails extends EventDetails {
```
alexbishop1: “Requires” should be “Require”.

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/GatewayDoesNotRequires3dsAuthorisationEventDetails.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private final boolean requires3DS;
+
+    private GatewayDoesNotRequires3dsAuthorisationEventDetails() {
+        this.requires3DS = false;
+    }
+
+    public static GatewayDoesNotRequires3dsAuthorisationEventDetails from() {
+        return new GatewayDoesNotRequires3dsAuthorisationEventDetails();
+    }
```
alexbishop1: You know, I’d just have a public no-args constructor and be done with it.

src/main/java/uk/gov/pay/connector/events/model/charge/GatewayDoesNotRequires3dsAuthorisation.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.pay.connector.events.model.charge;
+
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
+import uk.gov.pay.connector.events.eventdetails.charge.GatewayDoesNotRequires3dsAuthorisationEventDetails;
+
+import java.time.Instant;
+
+public class GatewayDoesNotRequires3dsAuthorisation extends PaymentEvent {
```
alexbishop1: “Requires” should be “Require”.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -221,6 +228,14 @@ private AuthorisationResponse updateChargePostAuthorisation(AuthCardDetails auth
                 mayBeCanRetry.orElse(null),
                 mayBeRejectedReason.orElse(null));
 
+        if (newStatus == AUTHORISATION_SUCCESS || newStatus == AUTHORISATION_REJECTED) {
+            if (updatedCharge.getRequires3ds() == null || !updatedCharge.getRequires3ds()) {
```
alexbishop1: Perhaps I’m overthinking this but I wonder if some cannily-named private methods might make the semantics clearer:

```java
private boolean canDefinitelyDetermineIf3dsWasRequired(ChargeStatus status) {
    return status == AUTHORISATION_SUCCESS || status == AUTHORISATION_REJECTED;
}

private boolean alreadyRequired3ds(ChargeEntity chargeEntity) {
    return Boolean.TRUE.equals(chargeEntity.getRequires3ds());
}

// …

if (canDefinitelyDetermineIf3dsWasRequired(newStatus) && !alreadyRequired3ds(chargeEntity)) {
    // …
}
```

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -221,6 +228,14 @@ private AuthorisationResponse updateChargePostAuthorisation(AuthCardDetails auth
...
+                updatedCharge.setRequires3ds(false);
+                chargeService.updateRequires3ds(updatedCharge.getExternalId(), false);
```
alexbishop1: This feels odd because we’re essentially setting it twice.

Can we make `updateRequires3ds(…)` return the updated `ChargeEntity` and then use that for the rest of the method?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -221,6 +228,14 @@ private AuthorisationResponse updateChargePostAuthorisation(AuthCardDetails auth
...
+                eventService.emitAndRecordEvent(GatewayDoesNotRequires3dsAuthorisation.from(updatedCharge, Instant.now()));
```
alexbishop1: If we inject an `InstantSource` (just add a parameter to the constructor) then we can use `instantSource.instant()` here, which will allow us to control the time in tests.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -208,6 +212,9 @@ class CardAuthoriseServiceTest extends CardServiceTest {
     @Mock
     private ExternalTransactionStateFactory mockExternalTransactionStateFactory;
 
+    @Mock
+    private EventService mockedEventServiceForCardAuthService;
```
alexbishop1: It seems a bit odd to have separate mock `EventService`s when in reality they would be the same one. I assume this was done to make it easier to assert on what events are and are not being sent.

I see a lot of `verifyNoInteractions(mockedEventServiceForCardAuthService)` — with a single mock `EventService`, something like `verifyNoMoreInteractions(…)` or even `verify(mockEventService, never()).emitAndRecordEvent(any(GatewayDoesNotRequires3dsEvent.class))` could be used.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -800,9 +860,20 @@ void doAuthoriseMotoApi_shouldPublishEvent() throws Exception {
...
         assertThat(eventCaptor.getValue().getEventType(), is("PAYMENT_DETAILS_SUBMITTED_BY_API"));
+
+        assertThat(charge.getRequires3ds(), is(false));
+        eventCaptor = ArgumentCaptor.forClass(Event.class);
+        verify(mockedEventServiceForCardAuthService).emitAndRecordEvent(eventCaptor.capture());
+        GatewayDoesNotRequires3dsAuthorisation event = (GatewayDoesNotRequires3dsAuthorisation) eventCaptor.getAllValues().get(0);
+        assertThat(event.getResourceExternalId(), is(charge.getExternalId()));
+        assertThat(event.getEventType(), is("GATEWAY_DOES_NOT_REQUIRES_3DS_AUTHORISATION"));
+        GatewayDoesNotRequires3dsAuthorisationEventDetails eventDetails = (GatewayDoesNotRequires3dsAuthorisationEventDetails) event.getEventDetails();
+        assertThat(eventDetails.isRequires3DS(), is(false));
```
alexbishop1: If `GatewayDoesNotRequires3dsAuthorisation` and `GatewayDoesNotRequires3dsAuthorisationEventDetails` had `equals(…)` methods on them, we could create a matching `GatewayDoesNotRequires3dsAuthorisation` object and then do `verify(mockedEventServiceForCardAuthService).emitAndRecordEvent(gatewayDoesNotRequires3dsAuthorisation)`.

alexbishop1 (changes requested):
Logic looks fine but I think there‘s some improvements we could make.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
SandorArpa: In reality EventService would be the same, however using Mockito only registers use in class. `verify(mockEventService, never()).emitAndRecordEvent(any(GatewayDoesNotRequires3dsEvent.class))` would always be true as the mock does not register any interactions from the second class when used in two mock constructors 🤷

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/GatewayDoesNotRequire3dsAuthorisationEventDetails.java
```diff
@@ -0,0 +1,26 @@
...
+import uk.gov.pay.connector.events.model.charge.GatewayDoesNotRequire3dsAuthorisation;
+
+import java.util.Objects;
+
+public class GatewayDoesNotRequire3dsAuthorisationEventDetails extends EventDetails {
+
+    private final boolean requires3DS = false;
+
+    public GatewayDoesNotRequire3dsAuthorisationEventDetails() {
+    }
```
alexbishop1: Since there are no other constructors, this can be omitted and the compiler will give us it for free.

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/GatewayDoesNotRequire3dsAuthorisationEventDetails.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @JsonProperty("requires_3ds")
+    public boolean isRequires3DS() {
+        return requires3DS;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        return o instanceof GatewayDoesNotRequire3dsAuthorisationEventDetails &&
+                this.requires3DS == ((GatewayDoesNotRequire3dsAuthorisationEventDetails) o).requires3DS;
```
alexbishop1, line 20: Should really override `hashCode()` here too.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -221,6 +231,13 @@ private AuthorisationResponse updateChargePostAuthorisation(AuthCardDetails auth
                 mayBeCanRetry.orElse(null),
                 mayBeRejectedReason.orElse(null));
 
+        if (newStatus == AUTHORISATION_SUCCESS || newStatus == AUTHORISATION_REJECTED) {
+            if (!Boolean.TRUE.equals(updatedCharge.getRequires3ds())) {
+                updatedCharge = chargeService.updateRequires3ds(updatedCharge.getExternalId(), false).get();
```
alexbishop1: Rather than just `get()`, maybe have an `orElseThrow(…)` with an `IllegalStateException` with a message saying the charge has vanished from the database or something, just so we have more to go on if this assumption ever does cease to hold.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -161,7 +167,7 @@ class CardAuthoriseServiceTest extends CardServiceTest {
     private LedgerService ledgerService;
 
     @Mock
-    private EventService mockEventService;
+    private EventService mockedEventService;
```
alexbishop1: Minor but it’s a little inconsistent to use the `mocked` prefix when everything else in this class uses `mock`.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -301,9 +311,13 @@ void doAuthoriseWeb_shouldPublishEvent() throws Exception {
...
-        verify(mockEventService, times(1)).emitAndRecordEvent(eventCaptor.capture());
-        assertThat(eventCaptor.getValue().getResourceExternalId(), is(charge.getExternalId()));
-        assertThat(eventCaptor.getValue().getEventType(), is("PAYMENT_DETAILS_ENTERED"));
+        verify(mockedEventService, times(2)).emitAndRecordEvent(eventCaptor.capture());
+        assertThat(eventCaptor.getAllValues().get(0).getResourceExternalId(), is(charge.getExternalId()));
+        assertThat(eventCaptor.getAllValues().get(0).getEventType(), is("PAYMENT_DETAILS_ENTERED"));
+
+        GatewayDoesNotRequire3dsAuthorisation event = GatewayDoesNotRequire3dsAuthorisation.from(charge, fixedInstantSource.instant());
+        GatewayDoesNotRequire3dsAuthorisation emittedEvent = (GatewayDoesNotRequire3dsAuthorisation) eventCaptor.getAllValues().get(1);
+        assertThat(event, equalTo(emittedEvent));
```
alexbishop1: Strictly speaking you could make an equal `PaymentDetailsEvent` event like you do with the `GatewayDoesNotRequire3dsAuthorisation` and then just do the standard `verify(mockedEventService).emitAndRecordEvent(event)` but instantiating a `PaymentDetailsEvent` requires a monstrous number of arguments so maybe it’s best it’s done this way.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -1039,6 +1142,30 @@ void doAuthoriseMotoApi_shouldReportUnexpectedError_whenProviderError() throws E
         assertTrue(response.getGatewayError().isPresent());
         assertThat(response.getGatewayError().get().getErrorType(), is(GATEWAY_ERROR));
         assertThat(charge.getStatus(), is(AUTHORISATION_UNEXPECTED_ERROR.getValue()));
+        assertThat(charge.getRequires3ds(), is(nullValue()));
+
+        verify(mockedEventService, never()).emitAndRecordEvent(any(GatewayDoesNotRequire3dsAuthorisation.class));
+    }
+
+    @Test
+    void doAuthoriseApi_shouldNotOverrideRequires3ds_IfAlreadyTrue() throws Exception {
```
alexbishop1, line 1141: Good spot.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -800,9 +864,18 @@ void doAuthoriseMotoApi_shouldPublishEvent() throws Exception {
...
+        verify(mockedEventService, times(2)).emitAndRecordEvent(eventCaptor.capture());
+        assertThat(eventCaptor.getAllValues().get(0).getResourceExternalId(), is(charge.getExternalId()));
+        assertThat(eventCaptor.getAllValues().get(0).getEventType(), is("PAYMENT_DETAILS_SUBMITTED_BY_API"));
+
+        assertThat(charge.getRequires3ds(), is(false));
+        GatewayDoesNotRequire3dsAuthorisation event = (GatewayDoesNotRequire3dsAuthorisation) eventCaptor.getAllValues().get(1);
+        assertThat(event.getResourceExternalId(), is(charge.getExternalId()));
+        assertThat(event.getEventType(), is("GATEWAY_DOES_NOT_REQUIRE_3DS_AUTHORISATION"));
+        GatewayDoesNotRequire3dsAuthorisationEventDetails eventDetails = (GatewayDoesNotRequire3dsAuthorisationEventDetails) event.getEventDetails();
+        assertThat(eventDetails.isRequires3DS(), is(false));
```
alexbishop1: How is this different to what `verifyGatewayDoesNotRequire3dsEventWasEmitted(charge)` does?

alexbishop1 (changes requested):
Looks good but I think there’s a couple more optimisations that can be made to the tests.

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -190,6 +201,14 @@ private void processGatewayAuthorisationResponse(
                 walletAuthorisationRequest.getPaymentInfo().getEmail(),
                 auth3dsRequiredDetails);
 
+        if (status == AUTHORISATION_SUCCESS || status == AUTHORISATION_REJECTED) {
+            if (!Boolean.TRUE.equals(updatedCharge.getRequires3ds())) {
+                updatedCharge = chargeService.updateRequires3ds(updatedCharge.getExternalId(), false)
+                        .orElseThrow(() -> new ChargeNotFoundRuntimeException(chargeExternalId));
+                eventService.emitAndRecordEvent(GatewayDoesNotRequire3dsAuthorisation.from(updatedCharge, instantSource.instant()));
+            }
+        }
```
alexbishop1: Can we someone how extract this so the same logic is not in two different places?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -301,9 +310,13 @@ void doAuthoriseWeb_shouldPublishEvent() throws Exception {
...
-        verify(mockEventService, times(1)).emitAndRecordEvent(eventCaptor.capture());
-        assertThat(eventCaptor.getValue().getResourceExternalId(), is(charge.getExternalId()));
-        assertThat(eventCaptor.getValue().getEventType(), is("PAYMENT_DETAILS_ENTERED"));
+        verify(mockEventService, times(2)).emitAndRecordEvent(eventCaptor.capture());
+        assertThat(eventCaptor.getAllValues().get(0).getResourceExternalId(), is(charge.getExternalId()));
+        assertThat(eventCaptor.getAllValues().get(0).getEventType(), is("PAYMENT_DETAILS_ENTERED"));
+
+        GatewayDoesNotRequire3dsAuthorisation event = GatewayDoesNotRequire3dsAuthorisation.from(charge, fixedInstantSource.instant());
+        GatewayDoesNotRequire3dsAuthorisation emittedEvent = (GatewayDoesNotRequire3dsAuthorisation) eventCaptor.getAllValues().get(1);
+        assertThat(event, equalTo(emittedEvent));
```
alexbishop1: I think this can call the `verifyGatewayDoesNotRequire3dsEventWasEmitted(…)` method instead.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -994,6 +1083,8 @@ void doAuthoriseMotoApi_shouldStoreCardDetails_ForAuthorisationError() throws Ex
 
         CardDetailsEntity cardDetails = charge.getCardDetails();
         assertThat(cardDetails, is(notNullValue()));
+
+        verify(mockEventService, never()).emitAndRecordEvent(any(GatewayDoesNotRequire3dsAuthorisation.class));
```
alexbishop1, line 1083: For consistency, should we also have `assertThat(charge.getRequires3ds(), is(nullValue())`?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -1197,6 +1330,14 @@ void doAuthoriseUserNotPresent_shouldThrowExceptionIfNoPaymentInstrument() throw
         assertThrows(IllegalArgumentException.class, () -> cardAuthorisationService.doAuthoriseUserNotPresent(charge));
     }
 
+    private void verifyGatewayDoesNotRequire3dsEventWasEmitted(ChargeEntity chargeEntity) {
+        ArgumentCaptor<Event> eventCaptor = ArgumentCaptor.forClass(PaymentEvent.class);
+        verify(mockEventService, times(2)).emitAndRecordEvent(eventCaptor.capture());
+        GatewayDoesNotRequire3dsAuthorisation event = GatewayDoesNotRequire3dsAuthorisation.from(chargeEntity, fixedInstantSource.instant());
+        GatewayDoesNotRequire3dsAuthorisation emittedEvent = (GatewayDoesNotRequire3dsAuthorisation) eventCaptor.getAllValues().get(1);
+        assertThat(event, equalTo(emittedEvent));
```
alexbishop1: The `verify(mockEventService, times(2))` is a bit brittle.

I think just doing `verify(mockEventService).emitAndRecordEvent(event)` (where `event` is the `GatewayDoesNotRequire3dsAuthorisation` object) should do the trick.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -109,6 +116,11 @@ class WorldpayCardAuthoriseServiceTest extends CardServiceTest {
     
     @Mock
     private AuthorisationConfig mockAuthorisationConfig;
+    
+    @Mock
+    private EventService mockedEventService;
```
alexbishop1: Another `mocked` versus `mock` inconsistency.

alexbishop1 (changes requested):
Looking good, just a few bits where I think we can remove duplication and simplify a little.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -995,6 +1001,15 @@ public ChargeEntity forceTransitionChargeState(ChargeEntity charge, ChargeStatus
...
     }
 
+    @Transactional
+    public Optional<ChargeEntity> updateRequires3ds(String chargeExternalId, Boolean requires3ds) {
+        return chargeDao.findByExternalId(chargeExternalId).map(chargeEntity -> {
+            chargeEntity.setRequires3ds(requires3ds);
+            chargeDao.merge(chargeEntity);
+            return chargeEntity;
+        });
+    }
```
alexbishop1: Can we inline this into the other new method now?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -1010,6 +1025,18 @@ public boolean isChargeCaptureSuccess(String externalId) {
         return status == CAPTURED || status == CAPTURE_SUBMITTED;
     }
 
+    public ChargeEntity updateRequires3ds(ChargeEntity chargeEntity) {
```
alexbishop1: I wonder if we could make the naming clearer here? I’m having trouble thinking what it could be though. Maybe `updateRequires3dsPostAuthorisation(…)` or something?

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -221,6 +221,8 @@ private AuthorisationResponse updateChargePostAuthorisation(AuthCardDetails auth
                 mayBeCanRetry.orElse(null),
                 mayBeRejectedReason.orElse(null));
 
+        updatedCharge = chargeService.updateRequires3ds(updatedCharge);
```
alexbishop1: Now I’m wondering if we could move this into `ChargeService`’s `updateChargePostCardAuthorisation(…)` (and `updateChargePostWalletAuthorisation(…)`)…

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -193,12 +209,19 @@ void authorise_with_exemption_flag_results_in_issuer_honouring_exemption_request
...
         verify(mockAppender, times(1)).doAppend(loggingEventArgumentCaptor.capture());
         String log = loggingEventArgumentCaptor.getAllValues().get(0).getMessage();
         assertTrue(log.contains("Authorisation with billing address and with 3DS data and without device data collection result"));
         assertTrue(log.contains("Worldpay authorisation response (orderCode: transaction-id, lastEvent: AUTHORISED, exemptionResponse result: HONOURED, exemptionResponse reason: ISSUER_HONOURED)"));
+
+        ArgumentCaptor<Event> eventCaptor = ArgumentCaptor.forClass(PaymentEvent.class);
+        verify(mockEventService, times(2)).emitAndRecordEvent(eventCaptor.capture());
+        GatewayDoesNotRequire3dsAuthorisation event = GatewayDoesNotRequire3dsAuthorisation.from(charge, fixedInstantSource.instant());
+        GatewayDoesNotRequire3dsAuthorisation emittedEvent = (GatewayDoesNotRequire3dsAuthorisation) eventCaptor.getAllValues().get(1);
+        assertThat(event, equalTo(emittedEvent));
```
alexbishop1: I assume this could also be replaced with a call to `verifyGatewayDoesNotRequire3dsEventWasEmitted(…)`.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -274,6 +308,17 @@ void do_authorise_should_respond_with_3ds_response_for_3ds_orders_with_worldpay_
...
+
+        verify(mockEventService, never()).emitAndRecordEvent(any(GatewayDoesNotRequire3dsAuthorisation.class));
+    }
+
+    private void verifyGatewayDoesNotRequire3dsEventWasEmitted(ChargeEntity chargeEntity) {
+        ArgumentCaptor<Event> eventCaptor = ArgumentCaptor.forClass(PaymentEvent.class);
+        verify(mockEventService, times(2)).emitAndRecordEvent(eventCaptor.capture());
+        GatewayDoesNotRequire3dsAuthorisation event = GatewayDoesNotRequire3dsAuthorisation.from(chargeEntity, fixedInstantSource.instant());
+        GatewayDoesNotRequire3dsAuthorisation emittedEvent = (GatewayDoesNotRequire3dsAuthorisation) eventCaptor.getAllValues().get(1);
+        assertThat(event, equalTo(emittedEvent));
```
alexbishop1: I think this could be replaced with:

```java
GatewayDoesNotRequire3dsAuthorisation event = GatewayDoesNotRequire3dsAuthorisation.from(chargeEntity, fixedInstantSource.instant());
verify(mockEventService).emitAndRecordEvent(event);
```

alexbishop1 (changes requested):
Looking good! Just a couple of final things.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
SandorArpa: Why not do this kind of refactoring once we delivered the project?

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: I believe keep as little as possible in methods marked as `@Transactional`.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -1010,6 +1016,23 @@ public boolean isChargeCaptureSuccess(String externalId) {
...
+            if (!Boolean.TRUE.equals(chargeEntity.getRequires3ds())) {
+                String externalId = chargeEntity.getExternalId();
+                chargeEntity = chargeDao.findByExternalId(externalId)
+                        .map(chargeToBeUpdated -> {
+                    chargeToBeUpdated.setRequires3ds(false);
+                    chargeDao.merge(chargeToBeUpdated);
+                    return chargeToBeUpdated;
+                }).orElseThrow(() -> new ChargeNotFoundRuntimeException(externalId));
+                eventService.emitAndRecordEvent(GatewayDoesNotRequire3dsAuthorisation.from(chargeEntity, instantSource.instant()));
+            }
```
alexbishop1: Do we actually need to get the `ChargeEntity` from the database if we already have it? Can we not just modify the passed-in one?

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -76,14 +76,9 @@ public GatewayResponse<BaseAuthoriseResponse> authorise(String chargeId, WalletA
...
-                        operationResponse = authoriseApplePay(charge, walletAuthorisationRequest);
-                        break;
-                    case GOOGLE_PAY:
-                        operationResponse = authoriseGooglePay(charge, walletAuthorisationRequest);
-                        break;
-                    default:
-                        throw new UnsupportedOperationException(format("Wallet type %s not recognised", walletAuthorisationRequest.getWalletType()));
+                    case APPLE_PAY -> operationResponse = authoriseApplePay(charge, walletAuthorisationRequest);
+                    case GOOGLE_PAY -> operationResponse = authoriseGooglePay(charge, walletAuthorisationRequest);
+                    default -> throw new UnsupportedOperationException(format("Wallet type %s not recognised", walletAuthorisationRequest.getWalletType()));
```
alexbishop1: Should be able to do this:

```java
GatewayResponse<BaseAuthoriseResponse> operationResponse = switch (walletAuthorisationRequest.getWalletType()) {
    case APPLE_PAY -> authoriseApplePay(charge, walletAuthorisationRequest);
    case GOOGLE_PAY -> authoriseGooglePay(charge, walletAuthorisationRequest);
    default -> throw new UnsupportedOperationException(format("Wallet type %s not recognised", walletAuthorisationRequest.getWalletType()));
};
```

alexbishop1 (changes requested):
Nice work! One comment where I think we can make a small bit of code more elegant and another comment which reveals I don’t understand JPA as well as I should do.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
SandorArpa: It's detached. We could get away with just passing on external id

src/main/java/uk/gov/pay/connector/wallets/WalletAuthoriseService.java
```diff
@@ -75,16 +75,10 @@ public GatewayResponse<BaseAuthoriseResponse> authorise(String chargeId, WalletA
...
-                    case GOOGLE_PAY:
-                        operationResponse = authoriseGooglePay(charge, walletAuthorisationRequest);
-                        break;
-                    default:
-                        throw new UnsupportedOperationException(format("Wallet type %s not recognised", walletAuthorisationRequest.getWalletType()));
-                }
+                operationResponse = switch (walletAuthorisationRequest.getWalletType()) {
+                    case APPLE_PAY -> authoriseApplePay(charge, walletAuthorisationRequest);
+                    case GOOGLE_PAY -> authoriseGooglePay(charge, walletAuthorisationRequest);
+                };
```
alexbishop1, line 81: 🎉

alexbishop1 (approved):

PR #5632, opened 2024-11-08
PP-13240 refactor wallet payment

SandorArpa:
## WHAT YOU DID

GDS Java style guide states:
"Optional is intended to only to represent the absence of a result: do not use it for fields or method parameters."

- refactor `WalletAuthoriseService` -> `processGatewayAuthorisationResponse` not to use Optional in a method parameter.


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -733,8 +733,8 @@ public ChargeEntity updateChargePostWalletAuthorisation(String chargeExternalId,
                                                             AuthCardDetails authCardDetails,
                                                             WalletType walletType,
                                                             String emailAddress,
-                                                            Optional<Auth3dsRequiredEntity> auth3dsRequiredDetails) {
-        return updateChargeAndEmitEventPostAuthorisation(chargeExternalId, status, authCardDetails, transactionId, auth3dsRequiredDetails.orElse(null), sessionIdentifier,
+                                                            Auth3dsRequiredEntity auth3dsRequiredDetails) {
```
marcotranchino, line 736: Would it be helpful to add `@Nullable` before Auth3dsRequiredEntity auth3dsRequiredDetails ?
I am reading that it is not necessary but it might provide clarity. 
In the codebase we only use `@Nullable` a couple of times, though, so perhaps not so useful.

marcotranchino (approved):
LGTM!  👍 

(I just left one comment more for my learning that anything else.)

PR #5643, opened 2024-11-19
PP_13241 Add historicalEventEmitter for requires3ds

hjvoid:
## WHAT YOU DID
When the Historical Event Emitter is emitting an event for a charge in connector and the `charge_events` table shows either `AUTHORISATION SUCCESS` or `AUTHORISATION REJECTED` and the charge's `requires_3ds` value is false it should include a `GATEWAY_DOES_NOT_REQUIRE_3DS_AUTHORISATION` event.

## How to test

- Start `pay local` using this branch in connector.
- Make new user, login and  make a demo payment
- Then make a POST request to the historical event emmiter using the following url `http://localhost:9300/v1/tasks/historical-event-emitter?start_id=1&max_id=100&record_type=charge&do_not_retry_emit_until_duration=7200` and make sure you receive a 200 response. 
 - Then if you check the logs in connector you should find `"Gateway does not require 3DS Authorisation event emitted for [chargeExternalId=<your_charge_id]"` in the logs.


src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -118,6 +121,25 @@ public void processPaymentEvents(ChargeEntity charge, boolean forceEmission) {
         processUserEmailCollectedEvent(charge, chargeEventEntities, forceEmission);
         processGatewayTransactionIdSetEvent(charge, chargeEventEntities, forceEmission);
         process3DSVersionEvent(charge,chargeEventEntities, forceEmission);
+        processGateway3DSAuthorisation(charge, chargeEventEntities, forceEmission);
+    }
+
+    private void processGateway3DSAuthorisation(ChargeEntity charge, List<ChargeEventEntity> chargeEventEntities, boolean forceEmission) {
```
alexbishop1: Looks like this is only doing the `GATEWAY_FOES_NOT_REQUIRE_3DS_AUTHORISATION` event (I assume we already have something for the `GATEWAY_REQUIRES_3DS_AUTHORISATION` event) so the method could probably be renamed

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -118,6 +121,25 @@ public void processPaymentEvents(ChargeEntity charge, boolean forceEmission) {
...
+        if(charge.getRequires3ds() != null && !charge.getRequires3ds()) {
```
alexbishop1: `if (Boolean.FALSE.equals(charge.getRequires3ds())` will do the same thing (it’s null-safe) and describes the intent a bit better

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -118,6 +121,25 @@ public void processPaymentEvents(ChargeEntity charge, boolean forceEmission) {
...
+            var chargeEvent = chargeEventEntities
+                    .stream()
+                    .filter(event -> event.getStatus() == AUTHORISATION_SUCCESS || event.getStatus() == AUTHORISATION_REJECTED)
+                    .findFirst();
+
+            chargeEvent.ifPresent(event -> {
+                Event doesNotRequire3DSAuthorisation = GatewayDoesNotRequire3dsAuthorisation.from(event.getChargeEntity(), event.getChargeEntity().getCreatedDate());
```
alexbishop1: I don’t think `event.getChargeEntity().getCreatedDate()` is right: I think you want the timestamp of the charge event

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -118,6 +121,25 @@ public void processPaymentEvents(ChargeEntity charge, boolean forceEmission) {
...
+                boolean hasBeenEmittedBefore = emittedEventDao.hasBeenEmittedBefore(doesNotRequire3DSAuthorisation);
+                if(forceEmission || !hasBeenEmittedBefore) {
+                    eventService.emitAndRecordEvent(doesNotRequire3DSAuthorisation, getDoNotRetryEmitUntilDate());
+                    logger.info("Gateway does not require 3DS Authorisation event emitted for [chargeExternalId={}]", charge.getExternalId());
+                }
+            });
+        }
```
alexbishop1, line 141: You can actually do this all in one stream operation and reach functional nirvana:

```java
chargeEventEntities
    .stream()
    .filter(chargeEventEntity ->
            chargeEventEntity.getStatus() == AUTHORISATION_SUCCESS || chargeEventEntity.getStatus() == AUTHORISATION_REJECTED)
    .findFirst()
    .map(chargeEventEntity ->
            GatewayDoesNotRequire3dsAuthorisation.from(chargeEventEntity.getChargeEntity(),
                    chargeEventEntity.getChargeEntity().getCreatedDate())
    ).filter(gatewayDoesNotRequire3dsAuthorisation ->
            forceEmission || !emittedEventDao.hasBeenEmittedBefore(gatewayDoesNotRequire3dsAuthorisation)
    ).ifPresent(gatewayDoesNotRequire3dsAuthorisation -> {
        eventService.emitAndRecordEvent(gatewayDoesNotRequire3dsAuthorisation, getDoNotRetryEmitUntilDate());
        logger.info("Gateway does not require 3DS Authorisation event emitted for [chargeExternalId={}]", charge.getExternalId());
    });
```

src/main/java/uk/gov/pay/connector/tasks/service/ChargeParityChecker.java
```diff
@@ -287,23 +288,25 @@ private boolean matchAuthorisationSummary(ChargeEntity chargeEntity, LedgerTrans
...
+                && transaction.getAuthorisationSummary().getThreeDSecure() != null
+                && transaction.getAuthorisationSummary().getThreeDSecure().isRequired();
+        
+        if (!chargeRequires3ds && chargeRequires3dsDetailsIsNull && !transactionRequires3ds) {
+            return true;
+        }
+        if (!chargeRequires3ds && transactionRequires3ds){
+            return isEquals(chargeEntity.getRequires3ds(), transaction.getAuthorisationSummary().getThreeDSecure().isRequired(), "authorisation_summary.three_d_secure.required");
+        }
+        if (chargeRequires3ds && !transactionRequires3ds) {
```
alexbishop1: As we just discussed on a call, this logic is complex.I think we might be able to use enums to separate out determining what data connector has what data ledger has from evaluating whether or not they match:

```java
enum ConnectorAuthorisationSummaryState {
    CONNECTOR_HAS_REQUIRES_3DS_TRUE,
    CONNECTOR_HAS_REQUIRES_3DS_FALSE,
    CONNECTOR_HAS_REQUIRES_3DS_NULL_BUT_HAS_3DS_REQUIRED_DETAILS,
    CONNECTOR_HAS_REQUIRES_3DS_NULL_AND_NO_3DS_REQUIRED_DETAILS
}

private static ConnectorAuthorisationSummaryState calculateConnectorAuthorisationSummaryState(ChargeEntity chargeEntity) {
    if (chargeEntity.getRequires3ds() == null) {
        if (chargeEntity.get3dsRequiredDetails() == null) {
            return CONNECTOR_HAS_REQUIRES_3DS_NULL_AND_NO_3DS_REQUIRED_DETAILS;
        } else {
            return CONNECTOR_HAS_REQUIRES_3DS_NULL_BUT_HAS_3DS_REQUIRED_DETAILS;
        }
    } else if (chargeEntity.getRequires3ds()) {
        return CONNECTOR_HAS_REQUIRES_3DS_TRUE;
    } else {
        return CONNECTOR_HAS_REQUIRES_3DS_FALSE;
    }
}

enum LedgerAuthorisationSummaryState {
    LEDGER_HAS_AUTHORISATION_SUMMARY_WITH_THREE_D_S_REQUIRED_TRUE,
    LEDGER_HAS_AUTHORISATION_SUMMARY_WITH_THREE_D_S_REQUIRED_FALSE,
    LEDGER_HAS_NO_AUTHORISATION_SUMMARY,
    LEDGER_HAS_SOMETHING_COMPLETELY_DIFFERENT
}

private LedgerAuthorisationSummaryState calculateLedgerAuthorisationSummaryState(LedgerTransaction transaction) {
    if (transaction.getAuthorisationSummary() == null) {
        return LEDGER_HAS_NO_AUTHORISATION_SUMMARY;
    } else {
        if (transaction.getAuthorisationSummary().getThreeDSecure() == null) {
            return LEDGER_HAS_SOMETHING_COMPLETELY_DIFFERENT;
        } else {
            if (transaction.getAuthorisationSummary().getThreeDSecure().isRequired()) {
                return LEDGER_HAS_AUTHORISATION_SUMMARY_WITH_THREE_D_S_REQUIRED_TRUE;
            } else {
                return LEDGER_HAS_AUTHORISATION_SUMMARY_WITH_THREE_D_S_REQUIRED_FALSE;
            }
        }
    }
}

private boolean matchAuthorisationSummary(ChargeEntity chargeEntity, LedgerTransaction transaction) {
    if (chargeEntity.getCreatedDate().isBefore(CHECK_AUTHORISATION_SUMMARY_PARITY_AFTER_DATE)) {
        return true;
    }

    ConnectorAuthorisationSummaryState connectorAuthorisationSummaryState = calculateConnectorAuthorisationSummaryState(chargeEntity);
    LedgerAuthorisationSummaryState ledgerAuthorisationSummaryState = calculateLedgerAuthorisationSummaryState(transaction);

    return switch (connectorAuthorisationSummaryState) {
        case CONNECTOR_HAS_REQUIRES_3DS_NULL_AND_NO_3DS_REQUIRED_DETAILS -> {
            if (ledgerAuthorisationSummaryState != LEDGER_HAS_NO_AUTHORISATION_SUMMARY) {
                logger.info("Field value does not match between ledger and connector [field_name={}]", "authorisation_summary.three_d_secure.required",
                        kv(FIELD_NAME, "authorisation_summary.three_d_secure.required"));
                yield false;
            }
            yield true;
        }
        case CONNECTOR_HAS_REQUIRES_3DS_FALSE -> {
            if (ledgerAuthorisationSummaryState != LEDGER_HAS_AUTHORISATION_SUMMARY_WITH_THREE_D_S_REQUIRED_FALSE) {
                logger.info("Field value does not match between ledger and connector [field_name={}]", "authorisation_summary.three_d_secure.required",
                        kv(FIELD_NAME, "authorisation_summary.three_d_secure.required"));
                yield false;
            }
            yield true;
        }
        case CONNECTOR_HAS_REQUIRES_3DS_TRUE, CONNECTOR_HAS_REQUIRES_3DS_NULL_BUT_HAS_3DS_REQUIRED_DETAILS -> {
            if (ledgerAuthorisationSummaryState != LEDGER_HAS_AUTHORISATION_SUMMARY_WITH_THREE_D_S_REQUIRED_TRUE) {
                logger.info("Field value does not match between ledger and connector [field_name={}]", "authorisation_summary.three_d_secure.required",
                        kv(FIELD_NAME, "authorisation_summary.three_d_secure.required"));
                yield false;
            }
            yield isEquals(chargeEntity.get3dsRequiredDetails().getThreeDsVersion(),
                    transaction.getAuthorisationSummary().getThreeDSecure().getVersion(), "authorisation_summary.three_d_secure.version");
        }
    };
}
```

src/test/java/uk/gov/pay/connector/events/HistoricalEventEmitterServiceTest.java
```diff
@@ -724,6 +726,48 @@ void shouldEmitGateway3dsInfoObtainedEvent_whenThreeDsVersionIsMissingFromLedger
...
+
+        chargeEntity.getEvents().clear();
+        chargeEntity.getEvents().add(successEvent);
+
+        when(chargeDao.findMaxId()).thenReturn(1L);
+        when(chargeDao.findById(1L)).thenReturn(Optional.of(chargeEntity));
+        when(emittedEventDao.hasBeenEmittedBefore(any())).thenReturn(false);
+
+        historicalEventEmitterService.emitHistoricEventsById(1L, OptionalLong.empty(), 1L);
+        verify(eventService, times(1)).emitAndRecordEvent(any(GatewayDoesNotRequire3dsAuthorisation.class), isNotNull());
```
alexbishop1: `times(1)` is redundant (it’s the default)

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -362,6 +427,73 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+
+        assertThat(parityCheckStatus, is(DATA_MISMATCH));
+
+        verify(mockAppender).doAppend(loggingEventArgumentCaptor.capture());
+        List<LoggingEvent> logStatement = loggingEventArgumentCaptor.getAllValues();
+        assertThat(logStatement.getFirst().getFormattedMessage(), is("Field value does not match between ledger and connector [field_name=authorisation_summary.three_d_secure.required]"));
+    }
+    
+    @ParameterizedTest
+    @MethodSource("valuesForParityCheck")
```
alexbishop1: If you name the method the same as the test then you don’t need a `@MethodSource` annotation

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -362,6 +427,73 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+    void parityCheck_shouldReturnDataMismatchIfThreeDSecureRequiredMismatch(
+            Boolean is3dsRequired,
+            ParityCheckStatus status,
+            Boolean chargeEntitySwitch
```
alexbishop1: Is this really whether it has 3DS data or not? And can it be a `boolean` rather than a `Boolean`?

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -362,6 +427,73 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+                .withAuthorisationSummary(authorisationSummary)
+                .build();
+
+        ParityCheckStatus parityCheckStatus = chargeParityChecker.checkParity(chargeEntityType, transaction);
+
+        assertThat(parityCheckStatus, is(status));
+    }
+    
+    @Test
+    void parityCheck_shouldReturnMatchForThreeDSecureRequired(){
```
alexbishop1: I’m not convinced this is right: I think if ledger has `authorisation_summary.three_d_secure.required` as `false` then connector should have `requires3ds` as `false` too

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
```diff
@@ -118,6 +119,26 @@ public void processPaymentEvents(ChargeEntity charge, boolean forceEmission) {
...
+        processGatewayDoesNotRequire3DSAuthorisation(charge, chargeEventEntities, forceEmission);
+    }
+
+    private void processGatewayDoesNotRequire3DSAuthorisation(ChargeEntity charge, List<ChargeEventEntity> chargeEventEntities, boolean forceEmission) {
+        if (Boolean.FALSE.equals(charge.getRequires3ds())) {
+            chargeEventEntities
+                    .stream()
+                    .filter(chargeEventEntity ->
+                            chargeEventEntity.getStatus() == AUTHORISATION_SUCCESS || chargeEventEntity.getStatus() == AUTHORISATION_REJECTED)
+                    .findFirst()
```
alexbishop1, line 131: Not sure we can guarantee what order we get the events in in the list — maybe sort it by what `getUpdated()` returns first?

src/main/java/uk/gov/pay/connector/tasks/service/ChargeParityChecker.java
```diff
@@ -283,29 +292,71 @@ private boolean matchRefundSummary(ChargeEntity chargeEntity, LedgerTransaction
...
+                }
+                yield true;
+            }
+            case CONNECTOR_HAS_REQUIRES_3DS_FALSE -> {
+                if (ledgerAuthorisationSummaryState != LEDGER_HAS_AUTHORISATION_SUMMARY_WITH_THREE_D_S_REQUIRED_FALSE) {
+                    logger.info("Field value does not match between ledger and connector [field_name={}]", "authorisation_summary.three_d_secure.required",
+                            kv(FIELD_NAME, "authorisation_summary.three_d_secure.required"));
+                    yield false;
+                }
+                yield true;
```
alexbishop1: Now I think about it, we should probably check that connector’s `chargeEntity.get3dsRequiredDetails().getThreeDsVersion()` and ledger’s `transaction.getAuthorisationSummary().getThreeDSecure().getVersion()` are both null here

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -362,6 +427,96 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+        List<LoggingEvent> logStatement = loggingEventArgumentCaptor.getAllValues();
+        assertThat(logStatement.getFirst().getFormattedMessage(), is("Field value does not match between ledger and connector [field_name=authorisation_summary.three_d_secure.required]"));
+    }
+    
+    @ParameterizedTest
+    @MethodSource
+    void parityCheck_shouldReturnDataMismatchIfThreeDSecureRequiredMismatch(
+            boolean is3dsRequired,
+            ParityCheckStatus status,
+            boolean chargeEntitySwitch
```
alexbishop1: I think rather than using a boolean here we could instead pass in the desired `ChargeEntity`

alexbishop1 (changes requested):
Looks good. There’s one extra place where I think we might need to do a version check.

src/main/java/uk/gov/pay/connector/tasks/HistoricalEventEmitter.java
hjvoid: I've added this in, but I think it may already being sorted via the method on line 114 `getSortedChargeEvents(charge)`?

alexbishop1: That method name does sound like a giveaway! OK, please remove the sorting I asked you to add!

src/test/java/uk/gov/pay/connector/events/HistoricalEventEmitterServiceTest.java
```diff
@@ -724,6 +726,48 @@ void shouldEmitGateway3dsInfoObtainedEvent_whenThreeDsVersionIsMissingFromLedger
...
```
alexbishop1: Unnecessary `times(1)`

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -79,8 +85,8 @@ class ChargeParityCheckerTest {
     @Captor
     ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor;
     
-    private ChargeEntity chargeEntity;
-    private ChargeEntity chargeEntityWith3ds;
+    private static ChargeEntity chargeEntity;
+    private static ChargeEntity chargeEntityWith3ds;
```
alexbishop1: Is it possible to avoid making these static? It imperils the independence of the tests. It seems these are static because they’re used in the static factory `@MethodSource` methods. Could we instead have these factory methods create entirely new `ChargeEntity` objects?

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -362,6 +427,96 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+    }
+    
+    private static Stream<Arguments> parityCheck_shouldMatchIfLedgerAndConnectorHaveThreeDSecureRequired() {
+        
+        return Stream.of(
+                Arguments.of(true, true),
+                Arguments.of(false, false)
+        );
+    }
+
```
alexbishop1: So it looks like we have the following combinations tested in the tests that are added by this PR:

| Connector `requires_3ds` | Connector `get3dsRequiredDetails()` | Ledger `AuthorisationSummary` |
|--------|--------|--------|
| `null` | Not `null` | `null` | Cell |
| `null` | Not `null` | Not `null` but no `three_d_secure` |
| `null` | Not `null` | `three_d_secure.required` is `false` |
| `null` | Not `null` | `three_d_secure.version` does not match |
| `null` | `null` | `three_d_secure.required` is `true` | 
| `null` | Not `null` | `three_d_secure.required` is `false` |
| `null` | `null`  | `three_d_secure.required` is `true` | 
| `true` | Not `null` | `three_d_secure.required` is `true` |
| `false` | Not `null` | `three_d_secure.required` is `false` |
| `null` | `null` | `null` | 

Can we add a few more? We do not seem to have a test for the matching `null`/Not `null`/`three_d_secure.required` is `true` combo. There are very few tests where connector’s `requires_3ds` is not `null` and also few tests that check the version.

alexbishop1 (changes requested):
Sorry this took so long. Looks good overall but I think we could test a few more combinations. Using a coverage checker may be a good way of checking we hit all the branches of the parity check.

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
hjvoid: Have added an extra test for the `true` and `false` values of `requires_3ds`. Also added additional test for positive match of `getThreeDSecure().setVersion`. We already have a negative check on line 394.

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -362,6 +427,196 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+        private static ChargeEntity createChargeEntity(Boolean isRequired) {
+            List<ChargeEventEntity> chargeEvents = createChargeEvents();
+
+            return aValidChargeEntity()
+                    .withStatus(CAPTURED)
+                    .withEvents(chargeEvents)
+                    .withRequires3ds(isRequired)
+                    .build();
+        }
```
alexbishop1: Ooh, I like this because it makes it quite explicit what combinations we’re testing. However in the case of the `ChargeEntity`, there’s not just whether `requires3ds` is `true`, `false` or `null`, there’s also whether `get3dsRequiredDetails()` returns `null` or not. So if my maths is right there’s basically six relevant different states a `ChargeEntity` can be in (though some of those — like `requires3ds` being `true` but `get3dsRequiredDetails()` returning `null`) are nonsensical.

Could we try to make these fine methods explicitly consider setting up `get3dsRequiredDetails()` too? This would make the combinations easier to see.

If we had similar methods for creating the `LedgerTransaction` (with or without the `authorisation_summary`, `authorisation_summary.three_d_secure` etc.), we should be able to fairly easily check all the relevant combinations, which would give us greater confidence there’s no edge cases missed in the tests.

alexbishop1 (changes requested):
Thanks for making those changes. I think the tests are getting clearer and I like that IntelliJ’s code coverage checker thinks the tests are executing every path.

I do think we could do a bit more to make sure we’re testing all the various combinations though, as described in the comment below.

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -361,6 +367,202 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+        when(mockProviders.byName(any())).thenReturn(new SandboxPaymentProvider());
+        ParityCheckStatus parityCheckStatus = chargeParityChecker.checkParity(chargeEntity, ledgerTransaction);
+        assertThat(parityCheckStatus, is(EXISTS_IN_LEDGER));
+    }
+    
+    private static Stream<Arguments> parityCheck_shouldReturnMatchIf3dsDataMatchesExactly() {
+        ChargeEntity chargeEntityWith3dsRequiredNull = ChargeEntityFactory.createWith3dsRequiredNull();
+        ChargeEntity chargeEntityWith3dsRequiredTrue = ChargeEntityFactory.createWith3dsRequired(true);
+        ChargeEntity chargeEntityWith3dsRequiredFalse = ChargeEntityFactory.createWith3dsRequired(false);
+        ChargeEntity chargeEntityWith3dsRequiredDetailsBut3dsRequiredNull = ChargeEntityFactory.createWith3dsRequiredDetailsBut3dsRequiredNull();
```
alexbishop1: Feels like these could maybe be fields

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -361,6 +367,202 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+                        LedgerTransactionFactory.createWith3dsRequiredTrue(chargeEntityWith3dsRequiredDetailsBut3dsRequiredNull)
+                ),
+                Arguments.of(
+                        chargeEntityWith3dsRequiredNull,
+                        LedgerTransactionFactory.createWithNullAuthorisationSummary(chargeEntityWith3dsRequiredNull)
+                )
+        );
+    }
+    
+    private static Stream<Arguments> parityCheck_shouldReturnDataMismatchFor3dsDataDiscrepencies() {
```
alexbishop1: Can this go next to its test?

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -361,6 +367,202 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+                        "[field_name=authorisation_summary.three_d_secure.required]"
+                ),
+                Arguments.of(
+                        chargeEntityWith3dsRequiredDetailsBut3dsRequiredNull,
+                        LedgerTransactionFactory.createWith3dsRequiredFalse(chargeEntityWith3dsRequiredDetailsBut3dsRequiredNull),
+                        "[field_name=authorisation_summary.three_d_secure.required]"
+                ),
+                Arguments.of(
+                        chargeEntityWith3dsRequiredDetailsBut3dsRequiredNull,
+                        LedgerTransactionFactory.createWith3dSecureVersionMismatch(chargeEntityWith3dsRequiredDetailsBut3dsRequiredNull),
```
alexbishop1: This is the only one with a version mismatch but I suspect we also want it for other casers like `requires3ds` in connector and `requires_3ds` in ledger being `true`

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -361,6 +367,202 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+                    .build();
+
+            ChargeEventEntity chargeEventCaptureSubmitted = aValidChargeEventEntity()
+                    .withCharge(null)
+                    .withChargeStatus(CAPTURE_SUBMITTED)
+                    .withTimestamp(parse("2016-01-26T13:23:55Z"))
+                    .build();
+
+            return List.of(chargeEventEntity, chargeEventCaptured, chargeEventCaptureSubmitted);
+        }
```
alexbishop1: Do we actually need these? Or will the parity checker cry if we use an empty list?

alexbishop1 (changes requested):
This is looking good. I think we might want one or two more version mismatch tests but other that it’s just minor things.

hjvoid:
@alexbishop1 All updated with changes when you're ready 🙂

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -361,6 +372,194 @@ void parityCheck_shouldMatchIfCreatedBeforeDateToCheckForAuthorisationSummaryPar
...
+        assertThat(logStatement.getFirst().getFormattedMessage(), is("Field value does not match between ledger and connector " + fieldName));
+    }
+
+    private static Stream<Arguments> parityCheck_shouldReturnDataMismatchFor3dsDataDiscrepencies() {
+        return Stream.of(
+                Arguments.of(
+                        CHARGE_ENTITY_WITH_3DS_DETAILS_BUT_REQUIRED_NULL,
+                        LedgerTransactionFactory.createWithNullAuthorisationSummary(CHARGE_ENTITY_WITH_3DS_DETAILS_BUT_REQUIRED_NULL),
+                        "[field_name=authorisation_summary.three_d_secure.required]"
+                ),
```
alexbishop1: I’d be tempted to also make the `LedgerTransaction`s constants too!

alexbishop1 (dismissed):
Awesome, thanks for persevering to get this into shape!

hjvoid:
@alexbishop1  `LedgerTransaction`s updated 🚀 💥

alexbishop1 (approved):
Lovely!

PR #5651, opened 2024-11-25
PP-13315: Return 422 when return url is not https and gateway account is live

oswaldquek:
As part of this commit,

* AllowedStrings and AllowedStringsValidator are sneakily deleted
* A generic ChargeException that can take an ErrorIdentifier and http response status code is created. There is scope to delete a bunch of custom Exceptions by using this class but this refactor can go into another PR.


DomBelcher (approved):
LGTM

PR #5672, opened 2024-12-18
PP-13432 use correct property value

SandorArpa:
## WHAT YOU DID

Use correct property value in GATEWAY_3DS_EXEMPTION_RESULT_OBTAINED event.


alexbishop1 (commented):
Is there a test we can have to catch this?

alexbishop1 (approved):

PR #5682, opened 2025-01-14
Add temporary debug logging for refund notifications on test merchant

marcotranchino:
This change adds additional logging to inspect the complete notification payload from Worldpay specifically for refund notifications (REFUND_FAILED and SENT_FOR_REFUND statuses) on the test merchant account (GBSRECABOFFGDSGBPTEST).

The extra logging will help investigate where Worldpay includes the refund request ID in their notifications, since we discovered that the 'reference' field is being used for:
- authorization code (for successful refunds)
- failure reason (for failed refunds)

The logging is strictly limited to the test merchant to avoid exposing PII in production environments.

Further information in JIRA.

https://payments-platform.atlassian.net/browse/PP-13416

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
@@ -82,6 +82,9 @@ public class WorldpayNotificationService {
             logger.debug("Payload: {}", payload);
             notification = XMLUnmarshaller.unmarshall(payload, WorldpayNotification.class);
             logger.info("Parsed {} notification: {}", PAYMENT_GATEWAY_NAME, notification);
+            if (isTemporaryLoggingRequiredOnTestOnly(notification)) {
+                logger.info("{} notification {} is being inspected (on test only) for PP-13416. Payload: {}", PAYMENT_GATEWAY_NAME, notification, payload);
+            }
         } catch (XMLUnmarshallerException e) {
             logger.error("{} notification parsing failed: {}", PAYMENT_GATEWAY_NAME, e);
             return true;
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -161,6 +164,13 @@ private boolean isErrorNotification(WorldpayNotification notification) {
         return "ERROR".equals(notification.getStatus());
     }
 
+    // TODO: Temporary logging for PP-13416
+    private boolean isTemporaryLoggingRequiredOnTestOnly(WorldpayNotification notification) {
+        return ("REFUND_FAILED".equals(notification.getStatus()) ||
+        "SENT_FOR_REFUND".equals(notification.getStatus()))
+        && notification.getMerchantCode().equals("GBSRECABOFFGDSGBPTEST");
```
alexbishop1: Can we do `"GBSRECABOFFGDSGBPTEST".equals(notification.getMerchantCode())` just in case there’s a null hiding there?

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
marcotranchino: Amended, thanks!

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -161,6 +164,13 @@ private boolean isErrorNotification(WorldpayNotification notification) {
         return "ERROR".equals(notification.getStatus());
     }
 
+    // TODO: Temporary logging for PP-13416
+    private boolean isTemporaryLoggingRequiredOnTestOnly(WorldpayNotification notification) {
+        return ("REFUND_FAILED".equals(notification.getStatus()) ||
+        "SENT_FOR_REFUND".equals(notification.getStatus()))
+        && "GBSRECABOFFGDSGBPTEST".equals(notification.getMerchantCode());
```
kbottla, line 171: Is the merchant code sensitive? It is for a test account but we could use env variable for this.

marcotranchino: ~~It will only be in Splunk and only for less than 24 hours.
I don't think that the merchant code is sensitive.~~

marcotranchino: We have discussed that this is a test account only.
It would have been better to use an env variable.
But it will be quicker to now merge this PR, get the 2 payloads we need for the project, and then revert this PR.

PR #5683, opened 2025-01-14
Revert "Add temporary debug logging for refund notifications on test merchant"

marcotranchino:
Reverts alphagov/pay-connector#5682

After having collected the necessary information, we can now revert the PR
that introduced the additionally logging.


marcotranchino:
Will do this at the end of the project work. Ticket: https://payments-platform.atlassian.net/browse/PP-13584

PR #5690, opened 2025-01-16
PP-13359 Prevent Stripe API payments under 30p

james-peacock-ho:
Context: to prevent capture errors due to fees exceeding transaction amount, we are preventing the creation of Stripe API payments of less than 30p. Similar restrictions for payment links will be considered separately.
- implement logic for the above.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -149,12 +149,14 @@ import static uk.gov.service.payments.commons.model.AuthorisationMode.AGREEMENT;
 import static uk.gov.service.payments.commons.model.AuthorisationMode.MOTO_API;
 import static uk.gov.service.payments.commons.model.ErrorIdentifier.AGREEMENT_NOT_ACTIVE;
 import static uk.gov.service.payments.commons.model.ErrorIdentifier.AGREEMENT_NOT_FOUND;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -1168,14 +1170,20 @@ private void checkMotoApiAuthorisationModeAllowed(GatewayAccountEntity gatewayAc
...
-
+    
     private void checkIfGatewayAccountDisabled(GatewayAccountEntity gatewayAccount) {
         if (gatewayAccount.isDisabled()) {
             throw new GatewayAccountDisabledException("Attempt to create a charge for a disabled gateway account");
         }
     }
 
+    private void checkIfAmountBelowMinimum(Source source, Long amount, GatewayAccountEntity gatewayAccount, String paymentProvider) {
+        if (source == CARD_API && amount < 30 && PaymentGatewayName.valueFrom(paymentProvider) == PaymentGatewayName.STRIPE) {
```
DomBelcher: I think it's worth making the 30 value a constant, something like `MINIMUM_STRIPE_PAYMENT_AMOUNT`

james-peacock-ho: Good point. I've updated this. I haven't updated tests because if the limit changes we will need to change the error message anyway.

DomBelcher (approved):
Nice work

PR #5697, opened 2025-01-22
Introduce new reference attribute in Worldpay Notification: `refundAuthorisationReference`

marcotranchino:
With this change, we are introducing a new reference attribute in the Worldpay
notification, i.e. in our representation of the notification we receive from
Worldpay: refundAuthorisationReference.

This is parsed from the XML element `journalReference` with type
`refund_authorisation`.

We have already been parsing an other reference attribute, parsing it from the
XML element `journalReference` with type `capture`.

An example for the first is:

```
<journalReference type="capture" reference="7500jujg0e8hjap4m9885f8ska"/>
```
 An example for the second is:
 
 ```
 <journalReference type="refund_authorisation" reference="987654"/>
```

This change makes sure that we avoid parsing the wrong reference values, now
that Worldpay will start sending the new reference attribute for successful
refunds.

In the future, it might be good to rename our first internal `reference`
attribute to `captureReference`, to make it clearer that it is parsed from
the journalReference XML element with type `capture`, however we are keeping
it named as `reference`, for now, to limit the scope of this change.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-13419

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -787,7 +787,7 @@
         "filename": "src/test/java/uk/gov/pay/connector/util/WorldpayXMLUnmarshallerTest.java",
         "hashed_secret": "c6b43e7d8e68a66c283fd8760118b89592f6498d",
         "is_verified": false,
-        "line_number": 123
+        "line_number": 125
       }
     ],
     "src/test/java/uk/gov/pay/connector/wallets/applepay/ApplePayAuthRequestBuilder.java": [
@@ -1101,5 +1101,5 @@
       }
     ]
   },
-  "generated_at": "2025-01-22T15:39:59Z"
+  "generated_at": "2025-01-23T09:12:11Z"
 }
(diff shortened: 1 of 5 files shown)


alexbishop1 (approved):
Looks straightforward enough.

We may wish to consider adopting the `WrappedStringValue` pattern we use elsewhere in the code to enhance type safety.

We might also want to ponder if it makes sense to convert `WorldpayNotification` into a record.

marcotranchino:
Thanks, Alex!

I'll work on both suggestions straightaway.

SandorArpa (approved):
LGTM

marcotranchino:
Thank you, Sandor!

alexbishop1:
> I'll work on both suggestions straightaway.

It’s not necessarily something we need to do right now, just something to consider if we decide to do a bit of a larger refactor as part of this overall project of work.

PR #5701, opened 2025-01-24
PP-13419 Handle new successful refund notification

marcotranchino:
With this change, we are no longer ignoring the SENT_FOR_REFUND notification sent by Worldpay after a refund submission.

The notification is now included in the REFUND_STATUSES and examined: if it contains a RefundAuthorisationReference, then it means that the refund submission was successful, otherwise the notification is ignored.

Since we are now processing different types of notifications arriving after a successful refund, we are also introducing a check in the Refund Notification Processor, to avoid setting a refund as successful multiple times, also avoiding multiple emails to the end user.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-13419

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
@@ -67,6 +67,18 @@ public class RefundNotificationProcessor {
         RefundEntity refundEntity = optionalRefundEntity.get();
         RefundStatus oldStatus = refundEntity.getStatus();
 
+        if (isRefundTransitionRedundant(oldStatus, newStatus)) {
+            logger.info("Notification received for refund [{}] is redundant and therefore ignored because refund is already in state [{}]",
+                    refundEntity.getExternalId(), oldStatus);
+            return;
+        }
+
+        if (isRefundTransitionIllegal(oldStatus, newStatus)) {
+            logger.error("Notification received for refund [{}] is unexpected and therefore ignored because it would cause an illegal state transition to [{}] as refund is already in state [{}]",
+                    refundEntity.getExternalId(), newStatus, oldStatus);
+            return;
+        }
+
         refundService.transitionRefundState(refundEntity, gatewayAccountEntity, newStatus, charge);
 
         if (RefundStatus.REFUNDED.equals(newStatus)) {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
```diff
@@ -67,6 +67,18 @@ public void invoke(PaymentGatewayName gatewayName, RefundStatus newStatus, Gatew
...
         RefundStatus oldStatus = refundEntity.getStatus();
 
+        if (isRefundTransitionRedundant(oldStatus, newStatus)) {
+            logger.info("Notification received for refund [{}] is redundant and therefore ignored because refund is already in state [{}]",
+                    refundEntity.getExternalId(), oldStatus);
+            return;
+        }
+
+        if (isRefundTransitionIllegal(oldStatus, newStatus)) {
+            logger.error("Notification received for refund [{}] is unexpected and therefore ignored because it would cause an illegal state transition to [{}] as refund is already in state [{}]",
```
marcotranchino, line 79: We have a separate ticket to handle this scenario and get alerted: [[Should] Alert Pay Support in case of illegal state transitions](https://payments-platform.atlassian.net/browse/PP-13497)

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -146,8 +145,13 @@ public boolean handleNotificationFor(String ipAddress, String payload) {
             }
             chargeNotificationProcessor.invoke(notification.getTransactionId(), charge, CAPTURED, notification.getGatewayEventDate());
         } else if (isRefundNotification(notification)) {
-            refundNotificationProcessor.invoke(PaymentGatewayName.WORLDPAY, newRefundStatus(notification), gatewayAccountEntity,
-                    notification.getReference(), notification.getTransactionId(), charge);
+            if(isOfflineRefund(notification)){
```
marcotranchino: Move into method to improve readability.

marcotranchino: Implemented.

SandorArpa (approved):
LOKTM

marcotranchino:
Thank you, @SandorArpa :)

PR #5703, opened 2025-01-31
Temporarily log message to test Splunk Alert

marcotranchino:
With this change, we are temporarily adding a log message that will be written only for the Worldpay Test account and only for the SENT_FOR_REFUND notifications without the refund authorisation reference.

This is needed to test the alert which will be triggered in case of illegal state transitions.

Further information in Jira.

## Test

This can be tested in the Worldpay test account by refunding a payment with the specific payment amount causing the simulated refund to happen offline.


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -83,6 +83,10 @@ public boolean handleNotificationFor(String ipAddress, String payload) {
             logger.info("Parsed {} notification: {}", PAYMENT_GATEWAY_NAME, notification);
             if (isTemporaryLoggingRequiredOnTestOnly(notification)) {
```
marcotranchino, line 84: The temporary logging is only executed if we are processing notifications received for the Worldpay Test Account.

SandorArpa (approved):
LOKTM

PR #5704, opened 2025-01-31
PP-13511: Reject payment of <30p for a Stripe account

oswaldquek:
CARD_EXTERNAL_TELEPHONE is excluded as that is a special case, see https://manual.payments.service.gov.uk/manual/pay-features/telephone-payment-notifications.html


DomBelcher (approved):
LGTM

PR #5707, opened 2025-02-03
PP-13445 Parse the reason for a refund failure from the notification

marcotranchino:
With this change, we are adding the definition for a new attribute in the
Worldpay Notification class in order to use it for the reason for a refund
failure which is included in the journal XML element with journalType
`REFUND_FAILED`.

The attribute is actually the description for notifications for any
journalType.

For this reason, we are introducing the attribute as `description` and then
also adding a dedicated getter than returns the refund failed description for
notifications of type REFUND_FAILED.

This will be logged and then used by the engineer on Support to determine what
to do when a refund request fails.

We will start informing the Support team with a Splunk Alert[1] linked to the
parsed notification.

Depending on the reason for the failure, the Support Team will be able to ask
the Service Admins to either retry the refund (for example for a payment
older than two years) or to find an alternative method outside of Pay
(for example if the credit card had expired).

Further information in Jira[2].

[1]

[2]
https://payments-platform.atlassian.net/browse/PP-13445

### DIFF
--- a/src/test/java/uk/gov/pay/connector/util/WorldpayXMLUnmarshallerTest.java
+++ b/src/test/java/uk/gov/pay/connector/util/WorldpayXMLUnmarshallerTest.java
@@ -59,9 +59,11 @@ class WorldpayXMLUnmarshallerTest {
     void shouldUnmarshallANotification() throws Exception {
         String transactionId = "MyUniqueTransactionId!";
         String status = "CAPTURED";
+        String description = "CAPTURED DESCRIPTION";
         String successPayload = TestTemplateResourceLoader.load(WORLDPAY_NOTIFICATION)
                 .replace("{{transactionId}}", transactionId)
                 .replace("{{status}}", status)
+                .replace("{{description}}", description)
                 .replace("{{refund-ref}}", "REFUND-REF")
                 .replace("{{refund-authorisation-reference}}", "REFUND-AUTHORISATION-REFERENCE")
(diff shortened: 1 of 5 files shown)


AbdirizakIdris (dismissed):
Looks good to me!

SandorArpa (changes requested):
I can't see the test where the new field `description` is actually parsed and checked 🤔  It should be in `WorldpayXMLUnmarshallerTest`.

src/test/java/uk/gov/pay/connector/util/WorldpayXMLUnmarshallerTest.java
```diff
@@ -70,6 +72,7 @@ void shouldUnmarshallANotification() throws Exception {
                 .replace("{{bookingDateYear}}", "2017");
         WorldpayNotification response = XMLUnmarshaller.unmarshall(successPayload, WorldpayNotification.class);
         assertThat(response.getStatus(), is(status));
+        assertThat(response.getDescription(), is(description));
```
marcotranchino, line 75: Test added, @SandorArpa , thanks! :)

marcotranchino:
Thank you, @AbdirizakIdris and @SandorArpa , for the review! 

I have added a new test for the unmarshalling of the notification to check the new description property.

Ready for re-review! :)

SandorArpa (approved):
LGTM

PR #5712, opened 2025-02-10
PP-13584 Remove payload inspection needed for testing

marcotranchino:
With this commit, we are removing some temporary logging that we put in place to inspect the full payload sent by Worldpay in the refund notifications.

We are now parsing additional properties from the full payload into our notification object, so there is no longer any need to keep this temporary logging.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-13584


SandorArpa (approved):
LOKTM

PR #5713, opened 2025-02-11
PP-13623 Set transaction ID for Worldpay refunds on creation

alexbishop1:
For Worldpay, the gateway transaction ID (how Worldpay refers to the refund) is always the same as the external ID of the refund. Set the gateway transaction ID as soon as the refund entity is created, rather than when we process the response to the refund request to Worldpay, to avoid a possible race condition if we receive a refund notification from Worldpay immediately.

To do this, introduce a `RefundEntityFactory` interface so the code for different payment gateways can supply one with the desired behaviour.


marcotranchino (dismissed):
I'm confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

yes

SandorArpa (dismissed):

SandorArpa (approved):

PR #5715, opened 2025-02-12
PP-13638 Prevent refunds' transitions from terminal states

marcotranchino:
With this change, we are fixing a problem that is only happening on the
Worldpay Test account.

The Worldpay refund simulation is such that they send us a refund outcome
notification before completing the response for our refund request
submission.

Our implementation is resilient and is able to set the refunds correctly as
REFUNDED or REFUND_ERROR (depending on magic values for the simulation),
however when Worldpay completes the response from our refund request
submission, we would be setting the refunds as REFUND_SUBMITTED although they
are already in their terminal state (REFUNDED or REFUND_ERROR).

With this change, we are overcoming the problem by skipping a state transition
which would move the refunds from their terminal state of REFUNDED or
REFUND_ERROR.

With separate work, we have already introduced a Splunk Alert to trigger a
Zendesk ticket in case of illegal state transitions from a terminal state to
another terminal state[1].

And we also have already in place alerts for refunds stuck in submitted
status.

So now every scenario is taken care of.

Further information in JIRA[2].

[1]

[2]
https://payments-platform.atlassian.net/browse/PP-13638


SandorArpa (approved):
LOKTM

PR #5720, opened 2025-02-14
PP-13428 refactor pact test states

SandorArpa:
## WHAT YOU DID

JS pact tests are not using params. To be able to use certain pact states, we need to hard code some params.


marcotranchino (approved):
LGTM 👍

PR #5738, opened 2025-02-28
PP-13353: Exclude disabled gateway accounts

oswaldquek:
This commit fixes the following bug (some context of how the [selfservice app](https://github.com/alphagov/pay-selfservice/) works is required):

* When creating a new service and selecting to create a Stripe test account in selfservice, then in connector a sandbox account is created and immediately disabled when the stripe test account is created
* At this point calling connector's `GET /v1/api/service/{serviceId}/account/{accountType}` endpoint still works with selfservice because Stripe is the priority so that's returned
* Now use the selfservice switching functionality to switch to Worldpay
* Now when selfservice calls connector's `GET /v1/api/service/{serviceId}/account/{accountType}` endpoint connector returns the old sandbox one as that takes precedence over the Worldpay one which causes an error as the sandbox account was never associated with the service in the adminusers database to begin with.
    
The solution to the above is to filter out disabled gateway accounts which this commit fixes.

The solution caused some test breakages, but upon consulting with the team, we 
have come to the agreement that disabled gateway accounts should effectively be
extinct. If there is a need to PATCH a gateway account from disabled to
enabled, the `PATCH /v1/api/accounts/{accountId}` would need to be used.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/service/GatewayAccountService.java
@@ -195,16 +195,17 @@ public class GatewayAccountService {
     }
 
     public Optional<GatewayAccountEntity> getGatewayAccountByServiceIdAndAccountType(String serviceId, GatewayAccountType accountType) {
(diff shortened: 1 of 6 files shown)


james-peacock-ho (approved):
LGTM

PR #5739, opened 2025-03-03
PP-13600 fix refund availability status when dispute is won in service favour

SandorArpa:
## WHAT YOU DID

When a dispute is raised by a user, we update the transaction’s refund availability to unavailable but, when the dispute is won the refundability status should be available.

Refund availability status should be set to available when a dispute is won.


marcotranchino (approved):
LGTM 👍

PR #5781, opened 2025-04-25
PP-13923 send email to Stripe if enabled

SandorArpa:
## WHAT YOU DID

If existing `send_payer_email_to_gateway` setting is true and email has been collected send it to Stripe with payment method.


SandorArpa:
It's been hanging around for a while. We shall re-open them if we ever decide to implement them.

PR #5786, opened 2025-04-29
PP-14023 triage todo comments.

SandorArpa:
## WHAT YOU DID

Triage of the 2025 ITHC.

Removed all comments.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/refund/resource/RefundReversalResource.java
+++ b/src/main/java/uk/gov/pay/connector/refund/resource/RefundReversalResource.java
@@ -54,8 +54,7 @@ public class RefundReversalResource {
     @Operation(
             summary = "Fix failed refund charge",
             responses = {
-                    @ApiResponse(responseCode = "200", description = "OK",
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/refund/resource/RefundReversalResource.java
```diff
@@ -54,8 +54,7 @@ public RefundReversalResource(ChargeService chargeService, GatewayAccountDao gat
     @Operation(
             summary = "Fix failed refund charge",
             responses = {
-                    @ApiResponse(responseCode = "200", description = "OK",
-                            content = @Content(schema = @Schema(example = "{ TODO -next pr}"))),
```
marcotranchino, line 58: Is this the same kind of TODO that we want to remove?

SandorArpa: It's a TODO that's unnecessary as the Response is only a code and not including an entity.

marcotranchino (approved):

PR #5796, opened 2025-05-07
PP-1233 spike how we can apply our own 3ds on stripe

AbdirizakIdris:
## Context
- This PR is to  Look into if/how we could request 3DS for Stripe at a service level
 
  - Add new column for `force_3ds`
  - Add `"payment_method_options[card][request_three_d_secure]"=any` param to Stripe Payment intent

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
@@ -28,6 +28,7 @@ public class StripePaymentIntentRequest extends StripePostRequest {
     private final String customerId;
     private boolean offSession;
     private String tokenId;
+    private boolean required3ds;
 
     private StripePaymentIntentRequest(
             GatewayAccountEntity gatewayAccount,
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/connector/gateway/stripe/request/StripePaymentIntentRequest.java
```diff
@@ -55,6 +56,7 @@ private StripePaymentIntentRequest(
         this.customerId = customerId;
         this.offSession = offSession;
         this.tokenId = tokenId;
+        this.required3ds = gatewayAccount.isForce3ds() || gatewayAccount.isRequires3ds();
```
alexbishop1, line 59: I think this should be a `&&` rather than a `||` — `isRequires3ds()` (not the best name any more) is whether the gateway account supports 3DS at all, which is a prerequisite for `isForce3ds()` to work

alexbishop1 (changes requested):
One small logic thing but otherwise this looks straightforward enough

GabrielBoardman-MadeTech:
Stripe is going away so closing this PR

PR #5805, opened 2025-05-14
Add ip address to metric counter for grafana

hjvoid:
## WHAT YOU DID
Implemented grafana metric label if ip address is sent with authorisation summary.


AbdirizakIdris (approved):
Looks good to me!

PR #5813, opened 2025-05-20
PP-14114 Fix refund authorisation code casting

marcotranchino:
With this change, we are updating the Worldpay Notification Service in order to accept alphanumeric values sent by Worldpay for the refund authorisation code when there has been a fast refund authorisation.

Previously we had been accepting numeric values only.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
@@ -40,6 +40,7 @@ public class WorldpayNotificationService {
             "SETTLED",
             "SETTLED_BY_MERCHANT"
     );
+    public static final String NULL_AUTHORISATION_CODE = "null";
     private static final List<String> REFUND_STATUSES = ImmutableList.of("REFUNDED", "REFUNDED_BY_MERCHANT", "REFUND_FAILED", "SENT_FOR_REFUND");
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
@@ -196,15 +197,10 @@ public class WorldpayNotificationService {
         if (!"SENT_FOR_REFUND".equals(notification.getStatus())) {
             return false;
         }
-
-        boolean hasRefundAuthorisationCode = false;
-        try {
-            Integer.parseInt(notification.getRefundAuthorisationReference());
-            hasRefundAuthorisationCode = true;
-        } catch (NumberFormatException e) {
-            hasRefundAuthorisationCode = false;
+        if (NULL_AUTHORISATION_CODE.equals(notification.getRefundAuthorisationReference())) {
+            return true;
         }
-        return !hasRefundAuthorisationCode;
+        return false;
     }
 
     public String notificationDomain() {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -196,15 +196,10 @@ private boolean isOfflineRefund(WorldpayNotification notification) {
...
-        try {
-            Integer.parseInt(notification.getRefundAuthorisationReference());
-            hasRefundAuthorisationCode = true;
-        } catch (NumberFormatException e) {
-            hasRefundAuthorisationCode = false;
+        if ("null".equals(notification.getRefundAuthorisationReference())) {
+            return true;
         }
-        return !hasRefundAuthorisationCode;
+        return false;
```
alexbishop1: This is me dreaming of `WorldpayNotification` being a record so we could use a switch expression with pattern matching → 😴

But until then, this looks fine!

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
```diff
@@ -275,7 +275,7 @@ void shouldIgnoreNotificationWhenStatusIsSentForRefundWithoutAuthorisationCode()
         setUpChargeServiceToReturnCharge(Optional.of(charge));
         setUpGatewayAccountServiceToReturnGatewayAccountEntity(Optional.of(gatewayAccountEntity));
         String status = "SENT_FOR_REFUND";
-        String refundAuthorisationCode = "";
+        String refundAuthorisationCode = "null";
```
alexbishop1: Perhaps this test should be renamed to be `shouldIgnoreNotificationWhenStatusIsSentForRefundWithStringOfNullAuthorisationCode`?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
```diff
@@ -284,7 +284,24 @@ void shouldIgnoreNotificationWhenStatusIsSentForRefundWithoutAuthorisationCode()
 
         verifyNoInteractions(mockChargeNotificationProcessor);
         verifyNoInteractions(mockRefundNotificationProcessor);
-        }
+    }
+
+    @Test
+    void shouldNotIgnoreNotificationWhenStatusIsSentForRefundWithAlphanumericAuthorisationCode() {
```
alexbishop1, line 290: I think this could merged into the test below: the distinction between purely numeric and alphanumeric authorisation codes is historic now. We could have a parametrised test with a few different codes if we want to make sure that both purely numeric and alphanumeric authorisation codes and possibly ones with trailing spaces etc. all work.

alexbishop1 (changes requested):
Looks good, just a couple of suggestions.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
marcotranchino: Amended, thanks!

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
marcotranchino: Sounds good, amended, thanks!

marcotranchino:
I have made the changes, Alex, thanks!
I have additionally assigned the "null" string to a static variable.

alexbishop1 (approved):

PR #5817, opened 2025-05-21
PP-14114 Fix refund authorisation code casting - version 2

marcotranchino:
With this change, we are updating the Worldpay Notification Service in order
to accept alphanumeric values sent by Worldpay for the refund authorisation
code when there has been a fast refund authorisation.

Previously we had been accepting numeric values only.

Note that  this change was attempted yesterday,  however there was a bug for
which it treated offline refunds as successful refunds[1], interpreting a
null reference code as an authorisation code.  The bug was quickly reverted[2].

We are now introducing the functionality again.

[1]
https://github.com/alphagov/pay-connector/pull/5813

[2]
https://github.com/alphagov/pay-connector/pull/5814

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
@@ -196,15 +196,11 @@ public class WorldpayNotificationService {
         if (!"SENT_FOR_REFUND".equals(notification.getStatus())) {
             return false;
         }
-
-        boolean hasRefundAuthorisationCode = false;
-        try {
-            Integer.parseInt(notification.getRefundAuthorisationReference());
-            hasRefundAuthorisationCode = true;
-        } catch (NumberFormatException e) {
-            hasRefundAuthorisationCode = false;
-        }
-        return !hasRefundAuthorisationCode;
+        return switch (notification.getRefundAuthorisationReference()) {
+            case null -> true;
+            case "null" -> true;
+            default -> false;
+        };
     }
 
     public String notificationDomain() {
--- a/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
@@ -271,11 +271,11 @@ class WorldpayNotificationServiceTest {
     }
 
     @Test
-    void shouldIgnoreNotificationWhenStatusIsSentForRefundWithoutAuthorisationCode() {
+    void shouldIgnoreNotificationWhenStatusIsSentForRefundWithNullAuthorisationCode() {
         setUpChargeServiceToReturnCharge(Optional.of(charge));
         setUpGatewayAccountServiceToReturnGatewayAccountEntity(Optional.of(gatewayAccountEntity));
         String status = "SENT_FOR_REFUND";
-        String refundAuthorisationCode = "";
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -196,15 +198,10 @@ private boolean isOfflineRefund(WorldpayNotification notification) {
...
-        try {
-            Integer.parseInt(notification.getRefundAuthorisationReference());
-            hasRefundAuthorisationCode = true;
-        } catch (NumberFormatException e) {
-            hasRefundAuthorisationCode = false;
+        if (notification.getRefundAuthorisationReference() == NULL_AUTHORISATION_CODE || NULL_STRING_AUTHORISATION_CODE.equals(notification.getRefundAuthorisationReference())) {
+            return true;
         }
-        return !hasRefundAuthorisationCode;
+        return false;
```
alexbishop1: I think this might be clearer using a switch expression with pattern matching and literal values:

```java
return switch (notification.getRefundAuthorisationReference()) {
    case null -> true;
    case "null" -> true;
    default -> false;
};
```

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
```diff
@@ -362,10 +379,13 @@ private static String sampleWorldpayNotification(
             String bookingDateDay,
             String bookingDateMonth,
             String bookingDateYear) {
+
+        String safeRefundAuthRef = refundAuthorisationReference == null ? "null" : refundAuthorisationReference;
```
alexbishop1: Not sure I follow this — it seems `refundAuthorisationReference` will never be null (it’s going to be `"ALPHANUM3R1C"`)

alexbishop1 (changes requested):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
marcotranchino: Ah yes, a lot nicer, thanks! :)

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationServiceTest.java
marcotranchino: I have amended this, thanks!

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
```diff
@@ -40,6 +40,8 @@ public class WorldpayNotificationService {
             "SETTLED",
             "SETTLED_BY_MERCHANT"
     );
+    public static final String NULL_AUTHORISATION_CODE = null;
+    public static final String NULL_STRING_AUTHORISATION_CODE = "null";
```
alexbishop1: I’d kill these and just inline them wherever they’re used

alexbishop1 (dismissed):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayNotificationService.java
marcotranchino: OK

alexbishop1 (approved):

PR #5875, opened 2025-07-03
PP-14229 Add gateway account details to MDC

kbottla:
## WHAT YOU DID
- Added gateway account details to MDC during payout reconciliation so we can better identify payouts by gateway account type and associated gateway account


DomBelcher (approved):
LGTM

PR #5879, opened 2025-07-04
PP-13880 Add unit tests for PaymentEventWithoutDetails classes

alexbishop1:
Add unit tests for subclasses of `PaymentEventWithoutDetails`. They are very repetitive and basically just test serialisation but will give us reassurance as we migrate our event classes to records.


hjvoid (approved):
LGTM

PR #5885, opened 2025-07-16
PP-13917 Expose paymentConfirmation to frontend

AbdirizakIdris:
## WHAT YOU DID
 - add `paymentConfirmationEmailEnabled` to FrontendChargeResponse


SandorArpa (changes requested):
Can we add some tests to ChargeService to prove that it works, eg true false, null?

AbdirizakIdris:
> Can we add some tests to ChargeService to prove that it works, eg true false, null?

Yes. Let me add it.

SandorArpa (approved):
LGTM

PR #5904, opened 2025-08-01
PP-14361 refactor apple pay and google pay

SandorArpa:
## WHAT YOU DID

Refactor WorldpayWalletAuthorisationHandler to only send payer's email for Google wallet payemnts.


SandorArpa:
It's been hanging around for a while. We shall re-open them if we ever decide to implement them.

PR #5911, opened 2025-08-06
PP-14318 Allow Worldpay Refund status change from error to refunded

marcotranchino:
With this change, we are preventing a problem that occurred only once since we went live with Faster Refunds, but that requires an engineer on support to fix, correcting the data.

We had implemented a block for Connector to prevent refund status changes from REFUNDED to REFUND_FAILED and also from REFUND_FAILED to REFUNDED.

When notifications arrive from Worldpay indicating the change of refund status which looks like the transitions above, then Connector would prevent the state transition and would log a message that would then be picked up by Splunk creating an alert for Support to investigate[1].

While we want this behaviour to continue unchanged for notifications that would transition a refund from REFUNDED to REFUND_FAILED, we have realised that the opposite transition should be allowed.

This is because it can happen (although rarely) that a connection timeout with the PSP causes a Refund status to be set as REFUND_FAILED.

In those rare cases, we want to allow a subsequent REFUNDED notification to change the refund status.

Further information in the JIRA ticket[2].

[1]
https://manual.payments.service.gov.uk/manual/support/worldpay-refund-illegal-state-transition.html

[2]
https://payments-platform.atlassian.net/browse/PP-14318

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
@@ -107,6 +107,6 @@ public class RefundNotificationProcessor {
     }
 
     private boolean isRefundTransitionIllegal(RefundStatus oldStatus, RefundStatus newStatus) {
-            return (oldStatus == REFUNDED && newStatus == REFUND_ERROR) || (oldStatus == REFUND_ERROR && newStatus == REFUNDED);
+            return (oldStatus == REFUNDED && newStatus == REFUND_ERROR);
     }
 }
--- a/src/test/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessorTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessorTest.java
@@ -31,6 +31,7 @@ import java.util.List;
 import java.util.Optional;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.not;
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessor.java
```diff
@@ -107,6 +107,6 @@ private boolean isRefundTransitionRedundant(RefundStatus oldStatus, RefundStatus
     }
 
     private boolean isRefundTransitionIllegal(RefundStatus oldStatus, RefundStatus newStatus) {
-            return (oldStatus == REFUNDED && newStatus == REFUND_ERROR) || (oldStatus == REFUND_ERROR && newStatus == REFUNDED);
+            return (oldStatus == REFUNDED && newStatus == REFUND_ERROR);
```
kbottla, line 110: Can we not just set the refund status to SUBMITTED if a refund times out and rely on the notification for the status change? Doesn't work the same way for Stripe refunds though, as we don't handle notifications for Stripe refunds.

It will be confusing in the UI (selfservice) if we show both ERROR & REFUNDED events.

marcotranchino:
Alternative PR: https://github.com/alphagov/pay-connector/pull/5912

marcotranchino:
We have agreed to proceed with the following alternative approach: [Set Worldpay refund as submitted when gateway timeout](https://github.com/alphagov/pay-connector/pull/5912)

We have decided to optimistically set the refund as submitted in such cases,
given that we have already a mechanism in place to deal with refunds stuck in
submitted status.

We are also logging a specific message in Splunk in order to be able to count
how many times this happens.

PR #5912, opened 2025-08-07
PP-14318 Set Worldpay refund as submitted when gateway timeout

marcotranchino:
With this change, we are addressing a problem that happened recently when the
Worldpay endpoint didn't respond in time to a refund submission request,
leading to Connector setting the refund failed when in fact it had been
correctly submitted.

We have decided to optimistically set the refund as submitted in such cases,
given that we have already a mechanism in place to deal with refunds stuck in
submitted status.

We are also logging a specific message in Splunk in order to be able to count
how many times this happens.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayRefundHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayRefundHandler.java
@@ -3,15 +3,20 @@ package uk.gov.pay.connector.gateway.worldpay;
 import com.google.inject.name.Named;
 import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.GatewayException.GatewayConnectionTimeoutException;
 import uk.gov.pay.connector.gateway.GatewayOrder;
 import uk.gov.pay.connector.gateway.RefundHandler;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayRefundHandler.java
```diff
@@ -41,6 +47,11 @@ public GatewayRefundResponse refund(RefundGatewayRequest request) {
                     buildRefundOrder(request), 
                     getWorldpayAuthHeader(request.getGatewayCredentials(), request.getAuthorisationMode(), request.isForRecurringPayment()));
             return GatewayRefundResponse.fromBaseRefundResponse(unmarshallResponse(response, WorldpayRefundResponse.class), PENDING);
+        } catch (GatewayConnectionTimeoutException e) {
+            logger.info("Despite a Worldpay Gateway connection timeout error, we are optimistically setting refund {} as SUBMITTED.",
```
kbottla, line 51: Can the refund external_id be a key-value pair?

kbottla (approved):

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayRefundHandler.java
marcotranchino: Thank you, @kbottla . Improved in a separate PR: https://github.com/alphagov/pay-connector/pull/5916

PR #5914, opened 2025-08-08
PP-14366 fix recurring payment template

SandorArpa:
## WHAT YOU DID

When we switched email sending for recurring card payments on our smoke test caught an error: `(error code: 2, error: The content of element type "shopper" must match "(shopperEmailAddress?,authenticatedShopperID?,...`

The DTD says:
```
<!ELEMENT shopperEmailAddress		(#PCDATA )>
<!ELEMENT authenticatedShopperID	(#PCDATA )>
```

Refactor the template to have the child elements <shopper> in the right order.


alexbishop1 (approved):
Nice and simple!

PR #5927, opened 2025-08-18
PP-13991 refactor config to use dropwizard-sentry 4

SandorArpa:
## WHAT YOU DID

We can migrate to 4.x now we use Dropwizard 4, which would allow us to remove some custom code.


marcotranchino (approved):
LGTM 👍

PR #5938, opened 2025-08-26
PP-14408 set recurring payment reason as RECURRING

SandorArpa:
## WHAT YOU DID

When the authorisation request for a first (customer-initiated) recurring payment is sent to Worldpay then set the value of the attribute named `customerInitiatedReason` to RECURRING.

When the authorisation request for a recurring payment using stored details (merchant-initiated) is sent to Worldpay then set the value of the attribute named `merchantInitiatedReason` to RECURRING.


alexbishop1 (approved):
Nice and simple.

PR #5946, opened 2025-08-28
PP-14378 null pointer exception on email

Rafiq-Gasmi:
NOTE: makeChargeSubmitCaptureAndCheckSettlementSummary Test has been Disabled, The issue is not an app issue but how the test works [PP-14430](https://payments-platform.atlassian.net/browse/PP-14430) 

## WHAT YOU DID
A NullPointerException occurred in Connector when the PATCH /v1/frontend/charges/<charge_id> endpoint was called by Frontend to update the email address on a charge.

Adding null checks before accessing email fields or methods.
Refactoring code to handle missing or optional email values.
Adding tests to cover scenarios where email is null or absent.
Avoids potential NullPointerException if "true".equalsIgnoreCase(variable) is used.
Simpler and more readable, directly returns the result of the validation. rather than return the negation of invalid (!invalid), so it returns true if valid.
made the String to a ChargeParamValidator enum more Concise, thread-safe, with explicit key mapping and is more maintainable.

## How to test

How should it be reviewed? Check tests, check code. run test scenario with null email
What feedback do you need? works or not!
If manual testing is needed, give suggested testing steps.

### Logging

NA
### Documentation

NA

[PP-14430]: https://payments-platform.atlassian.net/browse/PP-14430?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/uk/gov/pay/connector/common/validator/ApiValidators.java
+++ b/src/main/java/uk/gov/pay/connector/common/validator/ApiValidators.java
@@ -6,7 +6,6 @@ import uk.gov.service.payments.commons.model.SupportedLanguage;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
 import java.util.Arrays;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -29,27 +28,23 @@ public class ApiValidators {
                 return email == null || email.length() <= MAXIMUM_FIELDS_SIZE.get(EMAIL_KEY);
             }
         },
-
         AMOUNT(AMOUNT_KEY) {
             @Override
             boolean validate(String amount) {
-                int amountValue;
                 try {
-                    amountValue  = Integer.parseInt(amount);
+                    int amountValue = Integer.parseInt(amount);
+                    return amountValue >= MIN_AMOUNT && amountValue <= MAX_AMOUNT;
                 } catch (NumberFormatException e) {
                     return false;
                 }
-                return MIN_AMOUNT <= amountValue && MAX_AMOUNT >= amountValue;
             }
         },
-        
         DELAYED_CAPTURE(DELAYED_CAPTURE_KEY) {
             @Override
             boolean validate(String delayedCapture) {
-                return Boolean.TRUE.toString().equals(delayedCapture) || Boolean.FALSE.toString().equals(delayedCapture);
+                return "true".equalsIgnoreCase(delayedCapture) || "false".equalsIgnoreCase(delayedCapture);
             }
         },
-
         LANGUAGE(LANGUAGE_KEY) {
             @Override
             boolean validate(String iso639AlphaTwoCode) {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/common/validator/ApiValidators.java
```diff
@@ -98,11 +83,9 @@ public static Optional<List<String>> validateChargeParams(Map<String, String> in
...
 
     public static boolean validateChargePatchParams(PatchRequestBuilder.PatchRequest chargePatchRequest) {
-        boolean invalid = ChargeParamValidator.fromString(chargePatchRequest.getPath())
-                .map(validator -> !validator.validate(chargePatchRequest.getValue()))
+        return ChargeParamValidator.fromString(chargePatchRequest.getPath())
+                .map(validator -> validator.validate(chargePatchRequest.getValue()))
                 .orElse(false);
-
-        return !invalid;
     }
```
DomBelcher: I think this changes the behaviour if the result of `ChargeParamValidator.fromString` is empty, current behaviour appears to be it would return `true` but this would return `false`

src/main/java/uk/gov/pay/connector/common/validator/ApiValidators.java
```diff
@@ -26,30 +25,26 @@ private enum ChargeParamValidator {
         EMAIL(EMAIL_KEY) {
             @Override
             boolean validate(String email) {
-                return email.length() <= MAXIMUM_FIELDS_SIZE.get(EMAIL_KEY);
+                return email == null || email.length() <= MAXIMUM_FIELDS_SIZE.get(EMAIL_KEY);
```
DomBelcher: i'm not sure whether this is the behaviour we want if the email is `null`, I think the email should always have a value so if it is `null` it should probably return `false` instead

src/main/java/uk/gov/pay/connector/common/validator/ApiValidators.java
Rafiq-Gasmi: Correct, fixed!

src/main/java/uk/gov/pay/connector/common/validator/ApiValidators.java
Rafiq-Gasmi: Fair point, Fixed

src/main/java/uk/gov/pay/connector/common/validator/ApiValidators.java
```diff
@@ -26,30 +25,26 @@ private enum ChargeParamValidator {
...
                 }
-                return MIN_AMOUNT <= amountValue && MAX_AMOUNT >= amountValue;
             }
         },
-        
         DELAYED_CAPTURE(DELAYED_CAPTURE_KEY) {
             @Override
             boolean validate(String delayedCapture) {
-                return Boolean.TRUE.toString().equals(delayedCapture) || Boolean.FALSE.toString().equals(delayedCapture);
+                return "true".equalsIgnoreCase(delayedCapture) || "false".equalsIgnoreCase(delayedCapture);
```
nlsteers, line 45: do we need to check for nullness here? If `delayedCapture` is null this will throw a NPE

nlsteers: looks like this isn't the case

nlsteers (dismissed):

nlsteers (approved):
please squash all commits when merging this one

PR #5952, opened 2025-09-04
PP-14364 test for email and ip sent if CIT transaction

hjvoid:
## WHAT YOU DID

‘CIT’ stands for 'Consumer Initiated Transaction' so in recurring payment space it is the initial payment (first one) where customer is present. 
We want to make sure that if a recurring CIT payment is being authorised and the gateway account has 3ds enabled, `send_payer_ip_address_to_gateway` and `send_payer_email_to_gateway` are both `true`, when the authorisation request is snt to Worldpay the email address and IP address are included. The logging for the authorisation is already taken care of in `AuthorisationREquestSummaryStructuredLogging`.


iqbalgds (approved):
LGTM

alexbishop1 (approved):

PR #5955, opened 2025-09-08
PP-14365 do not send email and ip if MIT transaction

hjvoid:
## WHAT YOU DID
'MIT' stands for 'Merchant Initiated Transaction' so in recurring payment space is all and any subsequent payments where customer is not present. MIT are considered as cardholder not present therefore the collection of email and IP addresses are not expected by Visa.

We want to make sure that if a recurring CIT payment is being authorised then `send_payer_ip_address_to_gateway` and `send_payer_email_to_gateway` are both true, when the authorisation request is sent to Worldpay the email address and IP address are **NOT** included. We can be sure that the logging of the authorisation will not contain the email or IP address as the `stringifyLogMessage` method that takes care of this is not invoked when we call `authoriseUserNotPresent`.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
@@ -605,6 +605,44 @@ class WorldpayAuthoriseHandlerTest {
     }
 
     @Test
+    void should_not_send_email_and_IP_address_to_worldpay_when_a_MIT_recurring_payment_is_initiated() throws Exception {
+        when(authorisationSuccessResponse.getEntity()).thenReturn(load(WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE));
+        PaymentInstrumentEntity paymentInstrument = new PaymentInstrumentEntity.PaymentInstrumentEntityBuilder()
+                .withRecurringAuthToken(Map.of(
+                        WORLDPAY_RECURRING_AUTH_TOKEN_PAYMENT_TOKEN_ID_KEY, "test-payment-token-123456",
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/connector/util/TestTemplateResourceLoader.java
```diff
@@ -39,6 +39,7 @@ public class TestTemplateResourceLoader {
     public static final String WORLDPAY_VALID_AUTHORISE_WORLDPAY_REQUEST_MIN_ADDRESS = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-request-min-address.xml";
     
     public static final String WORLDPAY_VALID_AUTHORISE_RECURRING_WORLDPAY_REQUEST_WITH_SCHEME_IDENTIFIER = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-recurring-request-with-scheme-identifier.xml";
+    public static final String WORLDPAY_VALID_AUTHORISE_RECURRING_WORLDPAY_REQUEST_WITH_SCHEME_IDENTIFIER_EMAIL_AND_IP_ADDRESS = WORLDPAY_BASE_NAME + "/valid-authorise-worldpay-recurring-request-with-scheme-identifier-email-and-ip-address.xml";
```
alexbishop1: Feels a bit weird for this to be all email and IP address when it doesn’t actually include either of those! Is there an existing template we can reuse?

alexbishop1 (changes requested):

src/test/java/uk/gov/pay/connector/util/TestTemplateResourceLoader.java
hjvoid: Yeah there is - I thought that it would be clearer if we used a separate one but very happy to change it back 👍

alexbishop1 (approved):

PR #5958, opened 2025-09-11
PP-13353 only filter disabled test accounts

nlsteers:
## WHAT YOU DID

Previously, we filtered any disabled gateway account when querying via service external id and account type, this behaviour has been amended to only filter test gateway accounts that have been disabled. 

We still want services to be able to access their account via the admintool in the case of us disabling it.


DomBelcher (approved):
send it

PR #5969, opened 2025-09-16
PP-14432 add new column to charges table

SandorArpa:
## WHAT YOU DID

Add a new nullable column to the charges table called agreement_payment_type of type VARCHAR(30).


hjvoid (approved):
LGTM👍

PR #5970, opened 2025-09-16
PP-14432 add new column (agreement_payment_type) to ChargeEntity

SandorArpa:
## WHAT YOU DID

Add new enum variable agreementPaymentType as column agreement_payment_type to ChargeEntityt.


hjvoid (approved):
LGTM 👍

PR #5976, opened 2025-09-18
PP-14550 reconfigure dependabot ignore

SandorArpa:
## WHAT YOU DID

Dropwizard released v5. We don't want to upgrade to this version just yet.

Remove `org.dhatim:dropwizard-sentry` ignore. We are using v4 that comes bundled with Dropwizard 4.


hjvoid (approved):
LGTM

PR #5980, opened 2025-09-19
PP-14556 add temporary logs

SandorArpa:
## WHAT YOU DID

Add some extra temporary logging lines to debug a moto_api issue, see ZD-6226228


alexbishop1 (approved):
It’s not beautiful but it’s only temporary.

PR #5984, opened 2025-09-22
PP-14556 add some more temporary logs

SandorArpa:
## WHAT YOU DID

Ugly but useful temp logs


src/main/java/uk/gov/pay/connector/paymentprocessor/service/AuthorisationService.java
```diff
@@ -67,6 +67,8 @@ private <T> T executeAuthorise(Supplier<T> authorisationSupplier, int timeoutInM
             case IN_PROGRESS:
                 throw new AuthorisationExecutorTimedOutException("Timeout while waiting for authorisation to complete");
             default:
+                logger.info("Exception occurred while doing authorisation left: {}", executeResult.getLeft());
```
kbottla, line 70: Don't think these logs will be useful as requests are not making it this far.

PR #6013, opened 2025-10-13
PP-14433 return agreement_payment_type reason

SandorArpa:
## WHAT YOU DID

Connector returns agreement_payment_type reason property in payment response objects when available.


hjvoid (approved):
LGTM👌

PR #6029, opened 2025-10-23
PP-14634 Fix null pointer when Stripe account requirements null

alexbishop1:
When processing a Stripe account updated notification, handle the requirements in the data object provided by Stripe being null by assuming that the requirements being null means all requirements are satisfied.


SandorArpa (approved):
LGTM

DomBelcher (approved):
LGTM

PR #6038, opened 2025-10-30
PP-14444 When authorising a Worldpay recurring card payment include a…

hjvoid:
## WHAT YOU DID
When authorising a Worldpay recurring card payment we should send a reason. If the value of `agreementPaymentType` in the database is `null` then we should default to the value of `RECURRING`. The behaviour should be the same regardless of whether this is a Customer Initiated action or Merchant Initiated action.

## How to test

- How should it be reviewed? 
- What feedback do you need? 
  - I've not included tests for recurring transaction without scheme Identifiers because as far as I can tell this should never happen but happy to be told otherwise (taking my influence from the Worldpay docs [here](https://docs.worldpay.com/apis/wpg/industryschemeextras/storedcredentials)), which expresses the below:

>   When sending the original payment you may not know if a subsequent payment is needed. Since you will need the <transactionIdentifier> for any follow-up payments, we recommend that you request to always have this returned for industry-specific reasons. Contact us to enable this feature.

### Logging
The `Authorisation with` logging line of the structured logging key `agreement_payment_type` should have the corresponding type.


src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthorisationRequestSummary.java
```diff
@@ -80,5 +83,10 @@ public Optional<Exemption3ds> corporateExemptionResult() {
     
     @Override
     public Presence email() { return email; }
+    
+    @Override
+    public Optional<AgreementPaymentType> agreementPaymentType() { 
+        return Optional.of(agreementPaymentType != null ? agreementPaymentType : AgreementPaymentType.RECURRING);
```
alexbishop1: Not sure we want this. We default to recurring earlier on so defaulting again here should not be necessary — I think just returning an empty optional is the right thing to do.

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -32,8 +32,12 @@
                     </#if>
                 </CARD-SSL>
                 <#if savePaymentInstrumentToAgreement>
+                <#if agreementPaymentType??>
+                <storedCredentials usage="FIRST" customerInitiatedReason="${agreementPaymentType}"/>
+                <#else>
                 <storedCredentials usage="FIRST" customerInitiatedReason="RECURRING"/>
```
alexbishop1: Maybe if we do not have it we should just not include the `customerInitiatedReason`? We’re getting into “this should never happen” territory here (because we default to recurring earlier on).

src/main/resources/templates/worldpay/WorldpayAuthoriseRecurringOrderTemplate.xml
```diff
@@ -10,7 +10,11 @@
                 <TOKEN-SSL tokenScope="shopper">
                     <paymentTokenID>${paymentTokenId}</paymentTokenID>
                 </TOKEN-SSL>
+                <#if agreementPaymentType??>
+                <storedCredentials usage="USED" merchantInitiatedReason="${agreementPaymentType}">
+                <#else>
                 <storedCredentials usage="USED" merchantInitiatedReason="RECURRING">
```
alexbishop1: Same comment as above applies here.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -545,6 +555,46 @@ void should_not_include_email_when_not_present_and_send_email_to_gateway_enabled
...
+    @MethodSource
+    void should_send_reason_for_recurring_payment_to_worldpay_when_a_CIT_recurring_payment_is_initiated_with_an_agreement_payment_type(AgreementPaymentType agreementPaymentType, String worldpayTemplate) throws Exception {
+        when(authorisationSuccessResponse.getEntity()).thenReturn(load(WORLDPAY_AUTHORISATION_SUCCESS_RESPONSE));
+
+        ChargeEntity chargeEntity = chargeEntityFixture
+                .withExternalId("test-chargeId-789")
+                .withAmount(500L)
+                .withDescription("This is a description")
+                .withTransactionId("transaction-id")
+                .withEmail("test@email.com")
```
alexbishop1: Please use `.test` domain names — `test@email.com` may be a real email address.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -545,6 +555,46 @@ void should_not_include_email_when_not_present_and_send_email_to_gateway_enabled
...
+
+        worldpayAuthoriseHandler.authorise(getCardAuthorisationRequest(chargeEntity, "127.0.0.1"), DO_NOT_SEND_EXEMPTION_REQUEST);
+
+        ArgumentCaptor<GatewayOrder> gatewayOrderArgumentCaptor = ArgumentCaptor.forClass(GatewayOrder.class);
+        verify(authoriseClient).postRequestFor(eq(WORLDPAY_URL), eq(WORLDPAY), eq("test"), gatewayOrderArgumentCaptor.capture(), anyMap());
+        assertXMLEqual(load(worldpayTemplate),
+                gatewayOrderArgumentCaptor.getValue().getPayload());
+    }
+
+    private static Stream<Arguments> should_send_reason_for_recurring_payment_to_worldpay_when_a_CIT_recurring_payment_is_initiated_with_an_agreement_payment_type() {
```
alexbishop1: You might be able to use `@CsvSource` to do this without needing a separate method.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
```diff
@@ -560,6 +567,34 @@ void testVariationsOfSendPayerEmailAndSendPayerIPAddress(boolean is3dsEnabled, b
         assertXMLEqual(TestTemplateResourceLoader.load(testTemplatePath), actualRequest.getPayload());
     }
 
+    @ParameterizedTest
+    @MethodSource
+    void shouldGenerateValidAuthoriseRecurringOrderRequestWithSchemeIdentifierAndWithAgreementPaymentType(AgreementPaymentType agreementPaymentType, String worldpayTemplate) throws Exception {
```
alexbishop1: Feels like this duplicates the new tests in the previous file a little bit. Might be a good idea to work out where this properly belongs by looking at tests for similar functionality.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -782,7 +782,7 @@ void should_retry_without_exemption_flag_when_authorising_with_exemption_flag_re
         List<LoggingEvent> logs = loggingEventArgumentCaptor.getAllValues();
         assertTrue(logs.stream().anyMatch(loggingEvent -> {
             String log = format("Authorisation with billing address and without email address and with 3DS data and without device data " +
-                    "collection result for %s", chargeEntity.getExternalId());
+                    "collection result and with agreement payment type of recurring for %s", chargeEntity.getExternalId());
```
alexbishop1: We definitely do not want to be including stuff about the agreement payment type for a non-recurring payment but I think the suggestion about not defaulting to recurring earlier should mostly take care of this.

src/test/resources/templates/worldpay/valid-authorise-worldpay-request-setup-agreement-with-instalment-agreement-payment-type.xml
```diff
@@ -0,0 +1,43 @@
...
+                            <city>London</city>
+                            <countryCode>GB</countryCode>
+                        </address>
+                    </cardAddress>
+                </CARD-SSL>
+                <storedCredentials usage="FIRST" customerInitiatedReason="INSTALMENT"/>
+                <session id="test-chargeId-789" shopperIPAddress="127.0.0.1"/>
+            </paymentDetails>
+            <shopper>
+                <shopperEmailAddress>test@email.com</shopperEmailAddress>
```
alexbishop1: Another email that should be `.test` (others in the other files too).

alexbishop1 (changes requested):
Looks good overall and I like that we test the mapping to the final XML. There’s a couple of bits of logic relating to defaulting the value I think should be re-examined and I think there’s some test duplication.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
hjvoid: After conversation with Sandor I decided to keep these tests and remove the ones in `WorldpayAuthorisationRequestSummaryTest`

src/main/resources/templates/worldpay/WorldpayAuthoriseOrderTemplate.xml
```diff
@@ -32,7 +32,9 @@
                     </#if>
                 </CARD-SSL>
                 <#if savePaymentInstrumentToAgreement>
-                <storedCredentials usage="FIRST" customerInitiatedReason="RECURRING"/>
+                <#if agreementPaymentType??>
+                <storedCredentials usage="FIRST" customerInitiatedReason="${agreementPaymentType}"/>
+                </#if>
```
alexbishop1, line 39: We’re in this-can-never-happen territory here, but in the case where there’s no `agreementPaymentType`, I think we still want to include `<storedCredentials usage="FIRST"/>` or the credentials will not be saved — Worldpay’s docs at https://docs.worldpay.com/apis/wpg/industryschemeextras/storedcredentials say `customerInitiatedReason` has to be included but they also give examples without it.

src/main/resources/templates/worldpay/WorldpayAuthoriseRecurringOrderTemplate.xml
```diff
@@ -10,7 +10,9 @@
                 <TOKEN-SSL tokenScope="shopper">
                     <paymentTokenID>${paymentTokenId}</paymentTokenID>
                 </TOKEN-SSL>
-                <storedCredentials usage="USED" merchantInitiatedReason="RECURRING">
+                <#if agreementPaymentType??>
+                <storedCredentials usage="USED" merchantInitiatedReason="${agreementPaymentType}">
+                </#if>
```
alexbishop1, line 17: Same as above.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthorisationRequestSummaryTest.java
```diff
@@ -155,5 +160,4 @@ void corporateExemptionHasNoResult() {
         var worldpayAuthorisationRequestSummary = new WorldpayAuthorisationRequestSummary(mockChargeEntity, mockAuthCardDetails, false);
         assertThat(worldpayAuthorisationRequestSummary.corporateExemptionResult(), is(Optional.empty()));
     }
-
```
alexbishop1, line 158: Is anything actually changing in this file?

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -545,6 +552,41 @@ void should_not_include_email_when_not_present_and_send_email_to_gateway_enabled
         assertXMLEqual(load(WORLDPAY_VALID_AUTHORISE_WORLDPAY_REQUEST_INCLUDING_3DS_WITHOUT_IP_ADDRESS),
                 gatewayOrderArgumentCaptor.getValue().getPayload());
     }
+    
+    @ParameterizedTest
+    @CsvSource(nullValues = "null", textBlock = """
```
alexbishop1: No need for `nullValues = "null"` since none of the values are ever null.

Nice use of text blocks.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -604,6 +648,50 @@ void should_send_email_and_IP_address_to_worldpay_when_a_CIT_recurring_payment_i
                 gatewayOrderArgumentCaptor.getValue().getPayload());
     }
 
+    @ParameterizedTest
+    @MethodSource
```
alexbishop1: I think this can be made to use `@CsvSource` like the last one.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderRequestBuilderTest.java
alexbishop1: Still seems like there’s duplication with `WorldpayAuthoriseHandlerTest` here.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -305,6 +311,61 @@ void do_authorise_should_respond_with_3ds_response_for_3ds_orders_with_worldpay_
 
         verify(mockEventService, never()).emitAndRecordEvent(any(GatewayDoesNotRequire3dsAuthorisation.class));
     }
+    
+    @Test
+    void authorise_with_agreement_payment_type_of_null_without_log_message() throws Exception {
```
alexbishop1: The name of this test makes it sound a bit like there’s no log message at all rather than no agreement payment type of the log message — perhaps `agreement_payment_type_not_logged_if_null` or similar?

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -305,6 +311,61 @@ void do_authorise_should_respond_with_3ds_response_for_3ds_orders_with_worldpay_
...
+
+        ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
+        verify(mockAppender, times(1)).doAppend(loggingEventArgumentCaptor.capture());
+        String log = loggingEventArgumentCaptor.getAllValues().get(0).getMessage();
+        assertTrue(log.contains("Authorisation with billing address and without email address and with 3DS data and without device data collection result "));
+        assertTrue(log.contains("Worldpay authorisation response (orderCode: transaction-id, lastEvent: AUTHORISED) .'. AUTHORISATION SUCCESS -> AUTHORISATION SUCCESS"));
+    }
+
+    @ParameterizedTest
+    @MethodSource
```
alexbishop1: I think this could be `@CsvSource`.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/WorldpayCardAuthoriseServiceTest.java
```diff
@@ -305,6 +311,61 @@ void do_authorise_should_respond_with_3ds_response_for_3ds_orders_with_worldpay_
...
+
+        assertTrue(response.getAuthoriseStatus().isPresent());
+        assertThat(response.getAuthoriseStatus().get(), is(BaseAuthoriseResponse.AuthoriseStatus.AUTHORISED));
+        assertThat(charge.getStatus(), is(AUTHORISATION_SUCCESS.getValue()));
+        assertThat(charge.getRequires3ds(), is(false));
+
+        ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
+        verify(mockAppender, times(1)).doAppend(loggingEventArgumentCaptor.capture());
+        String log = loggingEventArgumentCaptor.getAllValues().get(0).getMessage();
+        assertTrue(log.contains("Authorisation with billing address and without email address and with 3DS data and without device data collection result and with agreement payment type of " + value));
```
alexbishop1: This can maybe be trimmed a bit: we don’t care about whether things like “billing address” or “3DS data” are in the logging for this test and we would not want this test to break if we tweak the wording of those other things.

alexbishop1: Can we also test the structured logging data here?

alexbishop1 (changes requested):
This looks like it’s almost there. Mostly minor comments.

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthorisationRequestSummaryTest.java
SandorArpa: Oppsie, I'll add a test

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
```diff
@@ -604,6 +648,46 @@ void should_send_email_and_IP_address_to_worldpay_when_a_CIT_recurring_payment_i
                 gatewayOrderArgumentCaptor.getValue().getPayload());
     }
 
+    @ParameterizedTest
+    @CsvSource(textBlock = """
+            RECURRING, templates/worldpay/valid-authorise-worldpay-recurring-request-with-scheme-identifier-and-agreement-payment-type-recurring.xml
+            INSTALMENT, templates/worldpay/valid-authorise-worldpay-recurring-request-with-scheme-identifier-and-agreement-payment-type-instalment.xml
+            UNSCHEDULED, templates/worldpay/valid-authorise-worldpay-recurring-request-with-scheme-identifier-and-agreement-payment-type-unscheduled.xml
```
alexbishop1, line 655: Pondering if we should have a test for the this-can-never-happen null case.

alexbishop1 (approved):

src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayAuthoriseHandlerTest.java
SandorArpa: Keep pondering and in case you deemed that it's necessary then when you're reviewing the story, you can send it back to add the this-can-never-happen null case. That gives you a couple of days at least.

PR #6086, opened 2025-12-05
PP-14757 Refactor create charge method

ralphm10:
## WHAT YOU DID
Refactor `createCharge` method to reduce cognitive complexity.
Separate PR to be raised for refactor of `checkAgreementOptions`

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
@@ -323,7 +323,7 @@ public class ChargeService {
     private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, Long accountId, String idempotencyKey) {
         return gatewayAccountDao.findById(accountId).map(gatewayAccount -> {
 
-            checkReturnUrl(chargeRequest, accountId, gatewayAccount);
+            checkReturnUrl(chargeRequest, gatewayAccount);
 
             GatewayAccountCredentialsEntity gatewayAccountCredential =
                     getGatewayAccountCredentialsEntity(chargeRequest, gatewayAccount);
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -412,6 +376,63 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
...
+                .withSavePaymentInstrumentToAgreement(chargeRequest.getSavePaymentInstrumentToAgreement())
+                .withAgreementEntity(agreementEntity)
+                .withAuthorisationMode(chargeRequest.getAuthorisationMode())
+                .withAgreementPaymentType(agreementPaymentType);
+
+        chargeRequest.getReturnUrl().ifPresent(chargeEntityBuilder::withReturnUrl);
+        return chargeEntityBuilder.build();
+    }
+
+    private void checkReturnUrl(ChargeCreateRequest chargeRequest, Long accountId, GatewayAccountEntity gatewayAccount) {
```
DomBelcher: you could omit the `accountId` parameter since it's already present on the `gatewayAccount`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -412,6 +376,63 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
...
+        chargeRequest.getReturnUrl().ifPresent(returnUrl -> {
+            if (gatewayAccount.isLive() && !returnUrl.startsWith("https://")) {
+                throw new ChargeException(format("Gateway account %d is LIVE, but is configured to use a " +
+                        "non-https return_url", accountId), NON_HTTPS_RETURN_URL_NOT_ALLOWED_FOR_A_LIVE_ACCOUNT, 422);
+            }
+        });
```
alexbishop1, line 433: This could be changed to this:

```java
if (gatewayAccount.isLive()) {
    chargeRequest.getReturnUrl().filter(returnUrl -> returnUrl.startsWith("https://"))
            .orElseThrow(() -> new ChargeException(format("Gateway account %d is LIVE, "
                    "but is configured to use a non-https return_url", accountId),
                    NON_HTTPS_RETURN_URL_NOT_ALLOWED_FOR_A_LIVE_ACCOUNT, 422));
}
```

… but I can’t decide if it’s clearer or not.

alexbishop1: Wait, that will throw the exception if there’s no return URL, so forget about it.

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
```diff
@@ -412,6 +376,63 @@ private Optional<ChargeEntity> createCharge(ChargeCreateRequest chargeRequest, L
...
+
+    private void checkGatewayAccount(ChargeCreateRequest chargeRequest, GatewayAccountEntity gatewayAccount, AuthorisationMode authorisationMode) {
+        checkIfGatewayAccountDisabled(gatewayAccount);
+
+        if (authorisationMode == MOTO_API) {
+            checkMotoApiAuthorisationModeAllowed(gatewayAccount);
+        } else {
+            checkIfMotoPaymentsAllowed(chargeRequest.isMoto(), gatewayAccount);
+        }
+    }
```
DomBelcher: would this be simpler if it was inlined in `checkChargeValidity`? or alternatively perhaps called from `createCharge` directly? also you could remove the `authorisationMode` parameter since this exists on `chargeRequest`

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
ralphm10: Updated

src/main/java/uk/gov/pay/connector/charge/service/ChargeService.java
ralphm10: Updated the parameter - thanks. 
Otherwise I think I'm happy with the way it reads as is

DomBelcher (approved):

PR #6105, opened 2025-12-12
PP-14728 - Removing deprecated jws.getBody()

Rafiq-Gasmi:
## WHAT YOU DID
Removing deprecated jws.getBody() and replace with jws.getPayload()

## How to test

- How should it be reviewed?  Look at the code and the replaced code is apprpriate 
- What feedback do you need? PR comments


alexbishop1 (approved):
Looks like `getPayload()` is an exact drop-in replacement for `getBody()` and this change only affects tests so it’s safe.

PR #6193, opened 2026-02-17
PP-14249 Update account resource to return 404 for invalid account type

ralphm10:
## WHAT YOU DID
Update account resource to return 404 for invalid account type


ralphm10:
closing as this is handled in self-service - implemented by same ticket PP-14249

PR #6215, opened 2026-03-05
PP-14844 add gateway_rejection_reason to ChargeEntity

SandorArpa:
## WHAT YOU DID

Add `gateway_rejection_reason` to ChargeEntity.
It should be set by a setter as it's not something that being set at creation time.


marcotranchino (approved):
LGTM 👍

alexbishop1 (approved):

PR #6227, opened 2026-03-12
PP-15075 Add 222100000000000 as sandbox test card number

alexbishop1:
Add 222100000000000 as a valid sandbox test card number because it is listed in our docs.

Replace 5555555555554444, which is no longer recognised by card ID nor listed in our docs.


SandorArpa (approved):
LGTM

PR #6239, opened 2026-03-23
PP-14844 store gateway rejection reason

SandorArpa:
## WHAT YOU DID

We want to store the unmapped rejection code and rejection message that we receive from the PSP, e.g. "fraudulent", into a new field in the charges DB table for Connector.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/response/BaseAuthoriseResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/response/BaseAuthoriseResponse.java
@@ -26,6 +26,14 @@ public interface BaseAuthoriseResponse extends BaseResponse {
                 .filter(authoriseStatus -> authoriseStatus == AuthoriseStatus.REJECTED)
                 .map(authoriseStatus -> MappedAuthorisationRejectedReason.UNCATEGORISED);
     }
+    
+    default Optional<String> getGatewayRejectionReason() {
+        return Optional.empty();
+    }
+    
+    default boolean isSoftDecline() {
+        return false;
+    }
 
     default Optional<Map<String, String>> getGatewayRecurringAuthToken() {
         return Optional.empty();
--- a/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
@@ -10,6 +10,7 @@ import uk.gov.service.payments.commons.model.CardExpiryDate;
 
 import java.util.Optional;
 import java.util.StringJoiner;
+import java.util.stream.Stream;
 
 import static uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse.AuthoriseStatus.ERROR;
 import static uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse.AuthoriseStatus.REJECTED;
@@ -59,11 +60,7 @@ public class StripeAuthorisationFailedResponse implements BaseAuthoriseResponse 
             return Optional.empty();
         }
 
-        var mappedAuthorisationRejectedReason = Optional.ofNullable(errorResponse)
-                .map(StripeErrorResponse::getError)
-                .flatMap(StripeErrorResponse.Error::getStripePaymentIntent)
-                .flatMap(StripePaymentIntent::getLastPaymentError)
-                .map(LastPaymentError::getDeclineCode)
+        var mappedAuthorisationRejectedReason = getGatewayRejectionReason()
                 .map(StripeAuthorisationRejectedCodeMapper::toMappedAuthorisationRejectionReason)
                 .orElse(MappedAuthorisationRejectedReason.UNCATEGORISED);
 
@@ -88,6 +85,15 @@ public class StripeAuthorisationFailedResponse implements BaseAuthoriseResponse 
     @Override
     public String getErrorMessage() {
         return null;
+    }
+    
+    @Override
+    public Optional<String> getGatewayRejectionReason() {
+        return  Optional.ofNullable(errorResponse)
+                .map(StripeErrorResponse::getError)
+                .flatMap(StripeErrorResponse.Error::getStripePaymentIntent)
+                .flatMap(StripePaymentIntent::getLastPaymentError)
+                .map(LastPaymentError::getDeclineCode);
     }
 
     @Override
--- a/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.connector.gateway.worldpay;
 
+import jakarta.xml.bind.annotation.XmlRootElement;
 import org.eclipse.persistence.oxm.annotations.XmlPath;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -11,7 +12,6 @@ import uk.gov.pay.connector.gateway.model.response.BaseCancelResponse;
 import uk.gov.pay.connector.gateway.model.response.BaseInquiryResponse;
 import uk.gov.service.payments.commons.model.CardExpiryDate;
 
-import jakarta.xml.bind.annotation.XmlRootElement;
 import java.time.YearMonth;
 import java.util.HashMap;
 import java.util.Map;
@@ -19,6 +19,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.StringJoiner;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 import static java.lang.String.format;
 import static java.util.function.Predicate.not;
@@ -122,6 +123,7 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
         this.challengeAcsUrl = challengeAcsUrl != null ? challengeAcsUrl.trim() : null;
     }
 
+    @Override
     public boolean isSoftDecline() {
         return Optional.ofNullable(lastEvent).map("REFUSED"::equals).orElse(false)
                 && Optional.ofNullable(exemptionResponseResult).map(SOFT_DECLINE_EXEMPTION_RESPONSE_RESULTS::contains).orElse(false);
@@ -272,6 +274,13 @@ public class WorldpayOrderStatusResponse implements BaseAuthoriseResponse, BaseC
                 .orElse(MappedAuthorisationRejectedReason.UNCATEGORISED);
 
         return Optional.of(mappedAuthorisationRejectedReason);
+    }
+
+    @Override
+    public Optional<String> getGatewayRejectionReason() {
+        return Stream.of(Optional.ofNullable(getRefusedReturnCode()), Optional.ofNullable(getRefusedReturnCodeDescription()))
+                .flatMap(Optional::stream)
+                .reduce((errorCode, errorMessage) -> errorCode + " " + errorMessage);
     }
 
     @Override
--- a/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
+++ b/src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
@@ -30,6 +30,7 @@ import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayReques
 import uk.gov.pay.connector.gateway.model.request.RecurringPaymentAuthorisationGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseAuthoriseResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
+import uk.gov.pay.connector.gateway.worldpay.WorldpayOrderStatusResponse;
 import uk.gov.pay.connector.logging.AuthorisationLogger;
 import uk.gov.pay.connector.paymentprocessor.api.AuthorisationResponse;
 import uk.gov.pay.connector.paymentprocessor.exception.AuthorisationExecutorTimedOutException;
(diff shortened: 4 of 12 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/response/BaseAuthoriseResponse.java
```diff
@@ -26,6 +26,10 @@ default Optional<MappedAuthorisationRejectedReason> getMappedAuthorisationReject
                 .filter(authoriseStatus -> authoriseStatus == AuthoriseStatus.REJECTED)
                 .map(authoriseStatus -> MappedAuthorisationRejectedReason.UNCATEGORISED);
     }
+    
+    default String getGatewayRejectionReason() {
+        return null;
+    }
```
alexbishop1: I think this should probably return an `Optional<String>` (we do use return optionals from other getters in this interface so there’s precedent).

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
```diff
@@ -89,6 +90,14 @@ public String getErrorCode() {
...
         return null;
     }
+    
+    @Override
+    public String getGatewayRejectionReason() {
+        return Stream.of(Optional.ofNullable(errorResponse.getError().getCode()), Optional.ofNullable(errorResponse.getError().getMessage()))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .reduce((errorCode, errorMessage) -> errorCode + " " + errorMessage).orElse(null);
+    }
```
alexbishop1, line 97: I’m not sure this is quite what we want. According to https://docs.stripe.com/api/authentication?lang=java there’s a `code`, which sounds like a generic error code, but also a `decline_code`, which sounds like it might be more specific. The `getMappedAuthorisationRejectedReason()` method uses the `decline_code` (with some gymnastics to get it from the right place) so that code could be reused.

I’m also not sure the `message` is going to give us anything that’s not already in the `decline_code` so maybe that could be omitted.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
```diff
@@ -274,6 +275,14 @@ public Optional<MappedAuthorisationRejectedReason> getMappedAuthorisationRejecte
         return Optional.of(mappedAuthorisationRejectedReason);
     }
 
+    @Override
+    public String getGatewayRejectionReason() {
+        return Stream.of(Optional.ofNullable(refusedReturnCode), Optional.ofNullable(refusedReturnCodeDescription))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .reduce((errorCode, errorMessage) -> errorCode + " " + errorMessage).orElse(null);
+    }
```
alexbishop1, line 284: I think this might break if, say, `refusedReturnCode` is not null but `refusedReturnCodeDescription` is null.

We might want to just do something along the lines of this instead:

```java
if (refusedReturnCode != null) {
    if (refusedReturnCodeDescription == null) {
        return refusedReturnCode;
    }
    return refusedReturnCode + ' ' + refusedReturnCodeDescription;
}
```

We may also want to take some inspiration from `return refusedReturnCode()` and check the status and that the `refusedReturnCode` is not blank.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -219,6 +222,15 @@ private AuthorisationResponse updateChargePostAuthorisation(AuthCardDetails auth
...
                 .flatMap(baseAuthoriseResponse ->
                         baseAuthoriseResponse.getMappedAuthorisationRejectedReason().map(MappedAuthorisationRejectedReason::name));
+
+        Optional<String> gatewayRejectionReason = Optional.empty();
+        
+        if (newStatus == AUTHORISATION_REJECTED && 
+                !(charge.getPaymentProvider().equals("worldpay") && operationResponse.getBaseResponse().isPresent() &&
+                    ((WorldpayOrderStatusResponse)operationResponse.getBaseResponse().get()).isSoftDecline())) {
+                gatewayRejectionReason = operationResponse.getBaseResponse().map(BaseAuthoriseResponse::getGatewayRejectionReason);
+        }
```
alexbishop1, line 231: This makes me seriously think maybe we should promote `isSoftDecline()` to be part of the `BaseAuthoriseResponse` interface (with a `default` implementation returning `false` — which seems OK because soft declines are an industry-specific thing, even if only Worldpay makes us directly deal with them). This would remove the Worldpay-specific logic from `CardAuthoriseService`.

If we’re not going to do that, I’d create a private `isWorldpaySoftDecline()` method like the below and use it:

```java
private boolean isWorldpaySoftDecline(GatewayResponse<BaseAuthoriseResponse> operationResponse) {
    return operationResponse.getBaseResponse()
            .filter(baseAuthoriseResponse -> baseAuthoriseResponse instanceof WorldpayOrderStatusResponse)
            .map(WorldpayOrderStatusResponse.class::cast)
            .map(WorldpayOrderStatusResponse::isSoftDecline)
            .orElse(false);
}
```

src/test/java/uk/gov/pay/connector/charge/service/ChargeServicePostAuthorisationTest.java
```diff
@@ -309,10 +309,21 @@ void updateChargePostAuthorisationWithSavePaymentInstrumentSavesInstrument() {
...
     }
 
+    @Test
+    void updateChargePostAuthorisationWithGatewayRejectionReasonSavesGatewayRejectionReason() {
+        var chargeEntity = chargeEntityFixture.build();
+        when(mockAuthCardDetailsToCardDetailsEntityConverter.convert(authCardDetails)).thenReturn(mockCardDetailsEntity);
+        when(mockChargeDao.findByExternalId(EXTERNAL_ID)).thenReturn(Optional.of(chargeEntity));
+
+        chargeService.updateChargePostCardAuthorisation(EXTERNAL_ID, AUTHORISATION_SUCCESS, TRANSACTION_ID, auth3dsRequiredEntity,
+                PROVIDER_SESSION_IDENTIFIER, authCardDetails, TOKEN, null, null, "42 fraudulent");
```
alexbishop1: Where’s this `42 fraudulent` coming from? I assume it’s in some existing mock response defined elsewhere?

alexbishop1 (changes requested):
Looks fine overall. A few comments about some of the logic and some of the code organisation.

src/test/java/uk/gov/pay/connector/charge/service/ChargeServicePostAuthorisationTest.java
SandorArpa: I'm not sure I understand your question. It's coming from
`updateChargePostAuthorisation(AuthCardDetails authCardDetails, 
                                                                ChargeEntity charge, 
                                                                ChargeStatus newStatus)`
where the test method is called from and the value of the gatewayRejectionReason is retrieved

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
SandorArpa: I've fully tested it in scratch and it doesn't break.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
SandorArpa: That's what the story says to collect and save

src/test/java/uk/gov/pay/connector/charge/service/ChargeServicePostAuthorisationTest.java
SandorArpa: TBH, I don't think we need this test. Come to think of it I originally added to show Marco where the save happens. I might just remove it as this save is in two integration tests.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
alexbishop1: I think the story may have overlooked this.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
alexbishop1: What does it return in that case? Is it `null` or a string like `"5 null"`?

src/test/java/uk/gov/pay/connector/charge/service/ChargeServicePostAuthorisationTest.java
alexbishop1: I meant where does `42 fraudulent` come from because I can’t see it anywhere else in the test. If you think it’s covered by other tests then that’s great.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
marcotranchino: rejection reasons

Currently we are saving this in Splunk.

marcotranchino: [From that documentation ](https://docs.stripe.com/api/authentication?lang=java) it would seem the decline_code would not include invalid request errors, api errors, idempotency errors, network decline codes.

marcotranchino: Interesting that in the example above expired_card is a declined_by_network rejection rather than an issuer_declined. Maybe in this case the network has rejected it without sending it to the bank. But actually there is also outcome.type set as issuer_declined.... Mmhhh

alexbishop1: I assume we wouldn’t want invalid request errors, API errors or idempotency errors recorded as decline reasons. Hopefully we’d determine the response is an error and not a decline in that case. I think (based off reading https://docs.stripe.com/declines/codes), Stripe always provide their own `decline_code` when there’s a network decline code.

marcotranchino: OK 👍

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
SandorArpa: a string like `5`

SandorArpa: I've added a parameterised test to `WorldpayOrderStatusResponseTest` called `get_gateway_rejection_reason_should_return_reduced_values`

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
```diff
@@ -89,6 +90,19 @@ public String getErrorCode() {
...
+    public Optional<String> getGatewayRejectionReason() {
+        var declineCode = Optional.ofNullable(errorResponse)
+                .map(StripeErrorResponse::getError)
+                .flatMap(StripeErrorResponse.Error::getStripePaymentIntent)
+                .flatMap(StripePaymentIntent::getLastPaymentError)
+                .map(LastPaymentError::getDeclineCode)
+                .orElse(null);
+        return Stream.of(Optional.ofNullable(declineCode), Optional.ofNullable(errorResponse.getError().getMessage()))
+                .flatMap(Optional::stream)
+                .reduce((errorCode, errorMessage) -> errorCode + " " + errorMessage);
```
alexbishop1: Are keeping the message? It’s not in the original story. For Worldpay we get a numeric code and a human-readable explanation but for Stripe the code is already human-readable so I’m not sure the message will give us anything else.

src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java
alexbishop1: That’s great.

src/main/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseService.java
```diff
@@ -219,6 +222,14 @@ private AuthorisationResponse updateChargePostAuthorisation(AuthCardDetails auth
         Optional<String> mayBeRejectedReason = operationResponse.getBaseResponse()
                 .flatMap(baseAuthoriseResponse ->
                         baseAuthoriseResponse.getMappedAuthorisationRejectedReason().map(MappedAuthorisationRejectedReason::name));
+
+        Optional<String> gatewayRejectionReason = Optional.empty();
+        
+        if (newStatus == AUTHORISATION_REJECTED && 
+                !(operationResponse.getBaseResponse().isPresent() && operationResponse.getBaseResponse().get().isSoftDecline())) {
+                gatewayRejectionReason = operationResponse.getBaseResponse().flatMap(BaseAuthoriseResponse::getGatewayRejectionReason);
+        }
```
alexbishop1, line 231: This is much nicer.

The second part of the `if` could be replaced by `!(operationResponse.getBaseResponse().map(BaseAuthoriseResponse::isSoftDecline).orElse(false))` if you like.

alexbishop1 (commented):
Looks good. One take-it-or-leave-it suggestion and a query about the error message for Stripe.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
SandorArpa: I'd keep it just so we are consistent with WP code, I guess. @marcotranchino what do you think?

marcotranchino: I think either way is fine.

src/main/java/uk/gov/pay/connector/gateway/stripe/json/StripeAuthorisationFailedResponse.java
```diff
@@ -89,6 +90,15 @@ public String getErrorCode() {
     public String getErrorMessage() {
         return null;
     }
+    
+    @Override
+    public Optional<String> getGatewayRejectionReason() {
```
alexbishop1, line 91: Can we make the `getMappedAuthorisationRejectedReason()` method reuse this (that is, `getMappedAuthorisationRejectedReason()` calls `getGatewayRejectionReason()` at some point)? Be good to minimise duplication.

alexbishop1 (changes requested):
One bit where I think we can minimise some duplication then I think we’re good.

alexbishop1 (approved):

PR #6253, opened 2026-03-31
PP-15127-rewrite-pact-consumer-tests-in-conventional-style

AndrzejProchyra:
## WHAT YOU DID
Rewrote `CardIdServiceConsumerTest` in the conventional [documented style](https://docs.pact.io/implementation_guides/jvm/consumer/junit), specifying the pact fragments using the DSL.

The intent of the work on  this ticket is to update the Pact tests to JUnit 5. Our current approach to using the Pact consumer library is to extend the library's [PactProviderRule](https://docs.pact.io/implementation_guides/jvm/consumer/junit#using-the-pact-junit-rule) as `PayPactProviderRule` and override its behaviour (we pull this rule in via our `testing-dropwizard-5` artifact). Our custom rule reads pacts from JSON files in the file tree (in `src/test/resources/pacts`). The conventional approach is to use the provided DSL to specify pact fragments programmatically right in the test file.

Our current approach is problematic for the JUnit 4 -> JUnit 5 upgrade because there is no clear upgrade path. The JUnit 5 library uses the JUnit 5 extension mechanism, rather than JUnit 4 rules.

Other disadvantages of our current approach:
* It obscures the test, since the expected behaviour of the provider is supplied in a separate JSON file, with no clear connection between one and the other
* It is confusing for developers familiar with Pact, and for developers learning how to use Pact, or just reading the library documentation

This PR therefore rewrites one of Connector's consumer tests in the "conventional" style, allowing for a smoother upgrade path. Perhaps our custom approach can then be re-applied to the JUnit 5 tests as a separate piece of work, if it is found to be advantageous.

There are two differences between the pact emitted by the rewritten test and the original test:
1. The body of the expected response has been reordered - however, the order does not matter to the provider-side tests
2. The interaction requests specify a header of `"Content-Type": "application/json"`. This is because, by default, the library infers a `Content-Type` header of `application/json; charset=UTF-8`, but Connector does not specify the charset in its request to Cardid. Therefore we need to add an override to the simpler header structure.

image

## How to test

- Ensure that the test has been rewritten without obscuring the intent, or weakening/over-specifying the contract


src/test/java/uk/gov/pay/connector/client/cardid/model/CardidServiceConsumerTest.java
```diff
@@ -12,56 +18,86 @@
...
 public class CardidServiceConsumerTest {
-    
+
+    private static final String FOUND_CARD_NUMBER = "2221000000000000";
+    private static final String NOT_FOUND_CARD_NUMBER = "1000000000000000";
+
     @Rule
-    public PayPactProviderRule cardidRule = new PayPactProviderRule("cardid", this);
-    
+    public PactProviderRule mockCardId = new PactProviderRule("cardid", this);
```
ralphm10: would it be clearer to call this `mockProvider` / `mockTestProvider` perhaps? 
I understand why it is called `mockCardId`, which is correct, however I at first misunderstood as being a mock "card", rather than a mock for the provider application.  
But that might just be me.

src/test/java/uk/gov/pay/connector/client/cardid/model/CardidServiceConsumerTest.java
```diff
@@ -12,56 +18,86 @@
...
+    public void setUp() {
+        when(configuration.getCardidBaseUrl()).thenReturn(mockCardId.getUrl());
         Client client = RestClientFactory.buildClient(new RestClientConfig(), null);
         cardidService = new CardidService(client, configuration);
     }
 
+    @Pact(consumer = "connector")
+    public RequestResponsePact cardNumberFound(PactDslWithProvider builder) {
+        return builder
+                .uponReceiving("a get card information request when the card number is found in the BIN ranges")
```
ralphm10: Is this possibly a bit confusing, because the request is a POST? 
Could it just read `.uponReceiving("a card information request...` ? 
Again, maybe just me.

ralphm10 (dismissed):
LGTM

A couple of questions but will leave up to you to decide.

src/test/java/uk/gov/pay/connector/client/cardid/model/CardidServiceConsumerTest.java
AndrzejProchyra: Yes, I agree. `mockProvider` does make it clearer.

src/test/java/uk/gov/pay/connector/client/cardid/model/CardidServiceConsumerTest.java
AndrzejProchyra: Yes, that makes sense. Thank you.

ralphm10 (approved):

PR #6256, opened 2026-04-01
PP-14845 send rejection_reason in event

SandorArpa:
## WHAT YOU DID

Send gateway_rejection_reason in AuthorisationRejected event's details. Refactor AuthorisationRejectedEventDetails to also include the gatewayRejectionReason as a string.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedEventDetails.java
@@ -1,20 +1,15 @@
 package uk.gov.pay.connector.events.eventdetails.charge;
 
-import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.events.eventdetails.EventDetails;
 
-public class AuthorisationRejectedEventDetails extends EventDetails {
-    private final boolean canRetry;
+public abstract class AuthorisationRejectedEventDetails extends EventDetails {
+    protected final String gatewayRejectionReason;
 
-    private AuthorisationRejectedEventDetails(boolean canRetry) {
-        this.canRetry = canRetry;
+    protected AuthorisationRejectedEventDetails(String gatewayRejectionReason) {
+        this.gatewayRejectionReason = gatewayRejectionReason;
     }
 
-    public static AuthorisationRejectedEventDetails from(ChargeEntity charge) {
-        return new AuthorisationRejectedEventDetails(charge.getCanRetry());
-    }
-
-    public boolean getCanRetry() {
-        return canRetry;
+    public String getGatewayRejectionReason() {
+        return gatewayRejectionReason;
     }
 }
--- a/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedWebPaymentEventDetails.java
+++ b/src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedWebPaymentEventDetails.java
@@ -1,0 +1,13 @@
+package uk.gov.pay.connector.events.eventdetails.charge;
+
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+
+public class AuthorisationRejectedWebPaymentEventDetails extends AuthorisationRejectedEventDetails {
+    private AuthorisationRejectedWebPaymentEventDetails(String gatewayRejectionReason) {
+        super(gatewayRejectionReason);
+    }
+
+    public static AuthorisationRejectedWebPaymentEventDetails from(ChargeEntity charge) {
+        return new AuthorisationRejectedWebPaymentEventDetails(charge.getGatewayRejectionReason());
+    }
+}
--- a/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationRejected.java
+++ b/src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationRejected.java
@@ -2,7 +2,9 @@ package uk.gov.pay.connector.events.model.charge;
 
 import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
 import uk.gov.pay.connector.chargeevent.model.domain.ChargeEventEntity;
+import uk.gov.pay.connector.events.eventdetails.charge.AuthorisationRejectedAgreementPaymentEventDetails;
 import uk.gov.pay.connector.events.eventdetails.charge.AuthorisationRejectedEventDetails;
+import uk.gov.pay.connector.events.eventdetails.charge.AuthorisationRejectedWebPaymentEventDetails;
 import uk.gov.service.payments.commons.model.AuthorisationMode;
 
 import java.time.Instant;
@@ -18,10 +20,10 @@ public class AuthorisationRejected extends PaymentEvent {
     }
 
     private AuthorisationRejected(String serviceId, boolean live, Long gatewayAccountId, String resourceExternalId,
-                                 AuthorisationRejectedEventDetails eventDetails, Instant timestamp) {
+                                  AuthorisationRejectedEventDetails eventDetails, Instant timestamp) {
         super(serviceId, live, gatewayAccountId, resourceExternalId, eventDetails, timestamp);
     }
-
+    
     public static AuthorisationRejected from(ChargeEventEntity chargeEvent) {
         ChargeEntity charge = chargeEvent.getChargeEntity();
         if (charge.getAuthorisationMode() == AuthorisationMode.AGREEMENT) {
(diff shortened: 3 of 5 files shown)


src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedEventDetails.java
```diff
@@ -4,17 +4,23 @@
 import uk.gov.pay.connector.events.eventdetails.EventDetails;
 
 public class AuthorisationRejectedEventDetails extends EventDetails {
-    private final boolean canRetry;
+    private final Boolean canRetry;
```
alexbishop1: Why did this change from a primitive?

src/test/java/uk/gov/pay/connector/events/model/refund/EventFactoryTest.java
```diff
@@ -510,8 +520,16 @@ void shouldCreateCorrectEventWithoutEventDetailsForAuthorisationRejectedWebAutho
...
         assertThat(event.getResourceExternalId(), is(chargeEventEntity.getChargeEntity().getExternalId()));
+
+        if (event.getEventDetails() instanceof AuthorisationRejectedEventDetails eventDetails) {
+            assertThat(eventDetails.getCanRetry(), is(canRetry));
+            assertThat(eventDetails.getGatewayRejectionReason(), is(gatewayRejectionReason));
+        } else if (event.getEventDetails() instanceof EmptyEventDetails) {
+            assertTrue(true);
+        } else {
+            assertTrue(false);
+        }
```
alexbishop1, line 496: This feels a little odd. I wonder if we’d be better off splitting this into multiple tests.

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedEventDetails.java
SandorArpa: Because it can be null, depending on whether it's a recurring or simple card payment.

src/test/java/uk/gov/pay/connector/events/model/refund/EventFactoryTest.java
SandorArpa: The else {} is not strictly needed TBH, I just thought that if the event details class changes we catch it.

SandorArpa: Perhaps:

`if (event.getEventDetails() instanceof AuthorisationRejectedEventDetails eventDetails) {
            assertThat(eventDetails.getCanRetry(), is(canRetry));
            assertThat(eventDetails.getGatewayRejectionReason(), is(gatewayRejectionReason));
        } else {
            assertThat(event.getEventDetails(), instanceOf(EmptyEventDetails.class));
        }`

will do.

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationRejected.java
```diff
@@ -18,7 +19,12 @@ private AuthorisationRejected(String serviceId, boolean live, Long gatewayAccoun
     }
 
     private AuthorisationRejected(String serviceId, boolean live, Long gatewayAccountId, String resourceExternalId,
-                                 AuthorisationRejectedEventDetails eventDetails, Instant timestamp) {
+                                  AuthorisationRejectedAgreementPaymentEventDetails eventDetails, Instant timestamp) {
+        super(serviceId, live, gatewayAccountId, resourceExternalId, eventDetails, timestamp);
+    }
+
+    private AuthorisationRejected(String serviceId, boolean live, Long gatewayAccountId, String resourceExternalId,
```
alexbishop1: I think we can still just have one private constructor here that takes an `AuthorisationRejectedEventDetails` (rather than one of its concrete implementations).

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationRejected.java
```diff
@@ -30,7 +36,16 @@ public static AuthorisationRejected from(ChargeEventEntity chargeEvent) {
                     charge.getGatewayAccount().isLive(),
                     charge.getGatewayAccount().getId(),
                     charge.getExternalId(),
-                    AuthorisationRejectedEventDetails.from(charge),
+                    AuthorisationRejectedAgreementPaymentEventDetails.from(charge),
+                    chargeEvent.getUpdated().toInstant()
+            );
+        } else if (charge.getAuthorisationMode() == AuthorisationMode.WEB && charge.getGatewayRejectionReason() != null) {
```
alexbishop1: Not sure the `charge.getAuthorisationMode() == AuthorisationMode.WEB` is necessary: we’d want to send a reason if we have one for other authorisation modes too.

src/test/java/uk/gov/pay/connector/events/model/refund/EventFactoryTest.java
```diff
@@ -473,19 +475,28 @@ void shouldCreateCorrectEventWithEventDetailsForAuthorisationRejectedAgreementAu
...
+    @ParameterizedTest
+    @CsvSource(useHeadersInDisplayName = true, nullValues = "null", textBlock = """
+        canRetry, gateway rejection reason, authorisation mode
+        true, null, AGREEMENT
+        null, fraudulent, WEB
+        null, null, WEB
+        false, lost_card, WEB
+        true, fraudulent, AGREEMENT
+    """)
+    void shouldCreateCorrectEventForAuthorisationRejectedWebAuthorisationMode(Boolean canRetry, String gatewayRejectionReason, AuthorisationMode authorisationMode) throws Exception {
```
alexbishop1, line 457: Do we still need the previous test (`shouldCreateCorrectEventWithEventDetailsForAuthorisationRejectedAgreementAuthorisationMode`) now we have this one?

src/test/java/uk/gov/pay/connector/events/model/refund/EventFactoryTest.java
```diff
@@ -510,8 +521,16 @@ void shouldCreateCorrectEventWithoutEventDetailsForAuthorisationRejectedWebAutho
...
         assertThat(event.getResourceExternalId(), is(chargeEventEntity.getChargeEntity().getExternalId()));
+
+        if (event.getEventDetails() instanceof AuthorisationRejectedAgreementPaymentEventDetails eventDetails) {
+            assertThat(eventDetails.getCanRetry(), is(canRetry));
+            assertThat(eventDetails.getGatewayRejectionReason(), is(gatewayRejectionReason));
+        } else if (event.getEventDetails() instanceof AuthorisationRejectedWebPaymentEventDetails eventDetails) {
+            assertThat(eventDetails.getGatewayRejectionReason(), is(gatewayRejectionReason));
+        } else  {
+            assertThat(event.getEventDetails(), instanceOf(EmptyEventDetails.class));
+        }
```
alexbishop1, line 496: This ended up much nicer. Could use a `switch` here if we like.

alexbishop1 (changes requested):
Although it’s novel, I think introducing different variants of `AuthorisationRejectedEventDetails` is a good approach here. Minor comments.

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationRejected.java
SandorArpa: No, we can't. We need a concrete extension of EventDetails. AuthorisationRejectedEventDetails is now an interface.

src/test/java/uk/gov/pay/connector/events/model/refund/EventFactoryTest.java
SandorArpa: If it were the codebase I'd use switch, however it's a test so I prefer an if else if else approach. I might be too oldskul.

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationRejected.java
alexbishop1: If `EventDetails` were an interface (which it could be because it has basically nothing in it) we could do this but that’s something for another day. As discussed elsewhere, `AuthorisationRejectedEventDetails` is now an abstract class, which will do for now.

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedWebPaymentEventDetails.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.pay.connector.events.eventdetails.charge;
+
+import uk.gov.pay.connector.charge.model.domain.ChargeEntity;
+
+public class AuthorisationRejectedWebPaymentEventDetails extends AuthorisationRejectedEventDetails {
```
alexbishop1, line 5: I suppose we could have `AuthorisationRejectedEventDetails` be a non-abstract class and use that instead of `AuthorisationRejectedWebPaymentEventDetails` but let’s keep it like this because it will make migrating all this to records easier in the future.

alexbishop1 (approved):

src/main/java/uk/gov/pay/connector/events/model/charge/AuthorisationRejected.java
SandorArpa: I agree. We shall make EventDetails an interface when re-factoring events to be records.

src/main/java/uk/gov/pay/connector/events/eventdetails/charge/AuthorisationRejectedWebPaymentEventDetails.java
SandorArpa: I fully agree

PR #6257, opened 2026-04-01
PP-15127: Use XMLUnit for XML assertions

AndrzejProchyra:
## WHAT YOU DID
1. Ran `mvn tidy:pom`, i.e the `pom` goal of the [Tidy Maven Plugin](https://www.mojohaus.org/tidy-maven-plugin/) to ensure that the POM conforms to the [POM Code Convention](https://maven.apache.org/developers/conventions/code.html#POM_Code_Convention). In practice this has:
* Added an XML declaration
* Updated the schema location to the modern form
* Formatted the artifact coordinates according to convention
* Added empty lines between sections

2. Added the `xmlunit-assertj3` dependency in `test` scope. This [XMLUnit](https://www.xmlunit.org/) dependency adds a custom AssertJ assertion, for expressive fluent assertions and excellent failure messages, enabling an automated diff view in IntelliJ.

3. Replaced test assertions that previously used `wiremock.org.custommonkey.xmlunit.XMLAssert.assertXMLEqual` from a transitive dependency with explicit, modern, XMLUnit assertions. Running XML assertions does not now call out to fetch the Worldpay XML schema on every run, and has speeded up these tests.

4. Reformatted the files and made small fixes suggested by IntelliJ

Note that I have changed one fixture. This is because the comparison here is more rigorous than the previous XML string comparison. The test `WorldpayOrderRequestBuilderTest#shouldGenerateValidCaptureOrderRequest` failed when rewritten because we expect the `amount` element to look like this:

```xml
<amount currencyCode="GBP" debitCreditIndicator="credit" exponent="2" value="500"/>
```
but we, in fact, produce this:

```xml
<amount currencyCode="GBP" exponent="2" value="500"/>
```
Installing the Worldpay schema in IntelliJ and examining `src/test/resources/templates/worldpay/valid-capture-worldpay-request.xml` showed that the default value of the `debitCreditIndicator` attribute in this case is `"credit"`, and so it is redundant. I have removed it.

An example of a failure message using the old assertion:

```
junit.framework.AssertionFailedError: wiremock.org.custommonkey.xmlunit.Diff
[different] Expected attribute value 'transaction-id' but was 'DIFFERENCE!' - comparing <order orderCode="transaction-id"...> at /paymentService[1]/submit[1]/order[1]/@orderCode to <order orderCode="DIFFERENCE!"...> at /paymentService[1]/submit[1]/order[1]/@orderCode
```
and with the AssertJ-enabled XMLUnit (I have abbreviated the pretty-printed XML):

```
Expecting:
 <control instance> and <test instance> to be identical
Expected attribute value 'transaction-id' but was 'DIFFERENCE!' - comparing <order orderCode="transaction-id"...> at /paymentService[1]/submit[1]/order[1]/@orderCode to <order orderCode="DIFFERENCE!"...> at /paymentService[1]/submit[1]/order[1]/@orderCode
expected:<<order orderCode="transaction-id">
            ...
        </order>> but was:<<order orderCode="DIFFERENCE!">
           ...
        </order>>>
Comparison Failure: 
<Click to see difference>
```
The diff window can then be viewed by clicking on the link.

The tests could still be improved, making use of XMLUnit's features, but I have not attempted this here.

## How to test

The tests should be like-for-like and not be obscured or weakened.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -54,6 +54,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.xmlunit</groupId>
+                <artifactId>xmlunit-bom</artifactId>
+                <version>2.11.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
     <dependencies>
@@ -425,13 +432,11 @@
         <dependency>
             <groupId>org.xmlunit</groupId>
             <artifactId>xmlunit-matchers</artifactId>
-            <version>2.11.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.xmlunit</groupId>
             <artifactId>xmlunit-assertj3</artifactId>
-            <version>2.11.0</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -428,6 +428,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.xmlunit</groupId>
+            <artifactId>xmlunit-assertj3</artifactId>
+            <version>2.11.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>uk.gov.service.payments</groupId>
             <artifactId>testing-dropwizard-5</artifactId>
             <version>${pay-java-commons.version}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -1,9 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
+
     <groupId>uk.gov.pay</groupId>
-    <version>0.1-SNAPSHOT</version>
     <artifactId>pay-connector</artifactId>
+    <version>0.1-SNAPSHOT</version>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -54,7 +56,6 @@
             </dependency>
         </dependencies>
     </dependencyManagement>
-
     <dependencies>
         <!-- Main dependencies that are imported from one of the BOMs specified
              in <dependencyManagement> so no explicit versions needed -->
@@ -440,6 +441,7 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
     <build>
         <plugins>
             <plugin>
@@ -613,6 +615,7 @@
             </plugin>
         </plugins>
     </build>
+
     <profiles>
         <profile>
             <id>default</id>
--- a/pom.xml
+++ b/pom.xml
@@ -349,6 +349,16 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.xmlunit</groupId>
+            <artifactId>xmlunit-matchers</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.xmlunit</groupId>
+            <artifactId>xmlunit-assertj3</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.testcontainers</groupId>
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/pay/connector/gateway/worldpay/WorldpayPaymentProviderTest.java
```diff
@@ -307,16 +307,6 @@ private void verifyLoggingTypeOf3dsExemption(Exemption3dsType exemption3dsType,
...
 
-    private void verifyLoggingTypeOf3dsExemption(Exemption3dsType exemption3dsType, String externalId, int timesLoggingCalled) {
-        ArgumentCaptor<LoggingEvent> loggingEventArgumentCaptor = ArgumentCaptor.forClass(LoggingEvent.class);
-        verify(mockAppender, times(timesLoggingCalled)).doAppend(loggingEventArgumentCaptor.capture());
-        List<LoggingEvent> logs = loggingEventArgumentCaptor.getAllValues();
-        assertTrue(logs.stream().anyMatch(loggingEvent -> {
-            String log = format("Requesting %s exemption - charge_external_id=%s", exemption3dsType.name(), externalId);
-            return loggingEvent.getFormattedMessage().contains(log);
-        }));
-    }
```
alexbishop1, line 318: Checked and this was never used.

src/test/resources/templates/worldpay/valid-capture-worldpay-request.xml
```diff
@@ -6,7 +6,7 @@
         <orderModification orderCode="MyUniqueTransactionId!">
             <capture>
                 <date dayOfMonth="23" month="02" year="2013"/>
-                <amount value="500" currencyCode="GBP" exponent="2" debitCreditIndicator="credit"/>
+                <amount value="500" currencyCode="GBP" exponent="2"/>
```
alexbishop1, line 9: `debitCreditIndicator` never appears in `WorldpayCaptureOrderTemplate.xml` (the actual XML we send to Worldpay) so removing it here improves the verisimilitude.

pom.xml
```diff
@@ -427,13 +428,20 @@
             <version>2.11.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.xmlunit</groupId>
+            <artifactId>xmlunit-assertj3</artifactId>
+            <version>2.11.0</version>
+            <scope>test</scope>
+        </dependency>
```
alexbishop1: It looks like there’s an [xmlunit-bom](https://central.sonatype.com/artifact/org.xmlunit/xmlunit-bom/2.11.0/overview). Since we’re now using two XMLUnit dependencies, it would probably be good to use this.

alexbishop1 (changes requested):
This looks great.

It looks like there’s an XMLUnit BOM we can use but apart from that no change requests.

pom.xml
AndrzejProchyra: Good idea. I've added the BOM and removed individual `version` elements

alexbishop1: Can you move the version-less XMLUnit dependencies into the `<!-- Test dependencies that are imported from one of the BOMs … -->` section of the doc? We like to keep them separate so it’s clear when we’re not specifying a version because it’s inherited from the POM rather than because we’re yoloing and using the latest version.

alexbishop1 (changes requested):

pom.xml
AndrzejProchyra: Done!

pom.xml
```diff
@@ -340,6 +348,22 @@
...
+        <dependency>
+            <groupId>org.xmlunit</groupId>
+            <artifactId>xmlunit-matchers</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.hamcrest</groupId>
+                    <artifactId>hamcrest-core</artifactId>
+                </exclusion>
+            </exclusions>
```
alexbishop1, line 360: We discussed the unfortunate need for this.

pom.xml
```diff
@@ -373,12 +397,6 @@
             <version>3.0</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.hamcrest</groupId>
-            <artifactId>hamcrest-library</artifactId>
-            <version>3.0</version>
-            <scope>test</scope>
-        </dependency>
```
alexbishop1, line 381: This is deprecated in favour of the plain `hamcrest` artifact (which we already import) so agreed it should get the boot.

pom.xml
```diff
@@ -402,6 +420,12 @@
             <artifactId>hamcrest-date</artifactId>
             <version>2.0.8</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.hamcrest</groupId>
+                    <artifactId>hamcrest-core</artifactId>
+                </exclusion>
+            </exclusions>
```
alexbishop1, line 428: We also discussed the unfortunate need for this.

alexbishop1 (approved):

PR #6263, opened 2026-04-07
PP-15127 use LogUnit for logging tests

AndrzejProchyra:
## WHAT YOU DID
* Added the [LogUnit](https://github.com/netmikey/logunit) library. LogUnit is listed as a [JUnit Jupiter extension in the JUnit wiki
](https://github.com/junit-team/junit-framework/wiki/Third-party-Extensions#junit-jupiter-extensions).
* Added explicit dependency on [AssertJ](https://assertj.github.io/doc/) for fluent assertions (already a transitive dependency in test scope).
* Rewrote all logging tests to use LogUnit
* Minor IDE inspection fixes/reformats

### Observations
Since the API of LogUnit provides built-in assertions and exposes the SLF4J `LoggingEvent`, rather than the low-level detail of each log provider, there are several ways to rewrite some assertions. For example, in `StripeNotificationServiceTest#shouldLogTheIdOfThePayoutCreatedEvent_whenItIsReceived` we assert on the logging event's arguments. A stronger assertion may be:

```java
assertThat(loggingEvent.getArguments())                                                      
        .asInstanceOf(list(ObjectAppendingMarker.class))                                                
        .containsExactly(new ObjectAppendingMarker("stripe_connect_account_id", "connect_account_id"));
```

but perhaps it suffices to assert on the marker's `toString`

```java
assertThat(loggingEvent.getArguments())                        
        .extracting(Object::toString)                                     
        .containsExactly("stripe_connect_account_id=connect_account_id");
```

## How to test

* The tests should all pass
* The intent of each assertion should not have been weakened or obscured


src/test/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementServiceTest.java
```diff
@@ -132,12 +115,13 @@ void logsErrorIfChargeDoesNotHavePaymentInstrument() {
...
-        assertThat(loggingEvent.getFormattedMessage(),
-                is("Expected charge " + chargeEntity.getExternalId() + " to have a payment instrument but it does not have one"));
-
+        logs.assertContains(
+                event ->
+                        event.getLevel().equals(ERROR) &&
+                                event.getMessage().equals("Expected charge " +
+                                        chargeEntity.getExternalId() +
+                                        " to have a payment instrument but it does not have one"),
+                "Expected ERROR not found.");
```
ralphm10: could we expand this a little? `Expected ERROR event not found` perhaps?

src/test/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessorTest.java
```diff
@@ -48,154 +41,124 @@ class RefundNotificationProcessorTest {
...
+        when(refundService.findByChargeExternalIdAndGatewayTransactionId(charge.getExternalId(), REFUND_GATEWAY_TRANSACTION_ID)).thenReturn(optionalRefundEntity);
 
-        refundNotificationProcessor.invoke(paymentGatewayName, RefundStatus.REFUND_ERROR, gatewayAccountEntity, refundGatewayTransactionId, transactionId, charge);
+        refundNotificationProcessor.invoke(paymentGatewayName, RefundStatus.REFUND_ERROR, gatewayAccountEntity, REFUND_GATEWAY_TRANSACTION_ID, TRANSACTION_ID, charge);
 
-        verify(mockAppender).doAppend(loggingEventArgumentCaptor.capture());
-        List<LoggingEvent> logStatement = loggingEventArgumentCaptor.getAllValues();
-        String expectedLogMessageForSplunkAlert = "Refund request record set as failed (REFUND_ERROR)";
-        assertThat(logStatement.getFirst().getFormattedMessage(), containsString(expectedLogMessageForSplunkAlert));
+        logs.assertContains("Refund request record set as failed (REFUND_ERROR)");
```
ralphm10, line 105: it's great how much code we have removed by adding log unit

src/test/java/uk/gov/pay/connector/gateway/processor/RefundNotificationProcessorTest.java
```diff
@@ -48,154 +41,124 @@ class RefundNotificationProcessorTest {
...
+        refundNotificationProcessor.invoke(paymentGatewayName, RefundStatus.REFUNDED, gatewayAccountEntity, "unknown", TRANSACTION_ID, charge);
 
-        List<LoggingEvent> logStatement = loggingEventArgumentCaptor.getAllValues();
         String expectedLogMessage = String.format("%s notification '%s' could not be used to update refund (associated refund entity not found) for charge [%s]",
                 paymentGatewayName,
                 "unknown",
                 charge.getExternalId());
-
-        assertThat(logStatement.getFirst().getFormattedMessage(), is(expectedLogMessage));
+        logs.assertContains(expectedLogMessage);
```
ralphm10, line 145: and I like how this is much more succinct than previously

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceAuthoriseTest.java
```diff
@@ -493,14 +483,13 @@ private void verify422AndErrorMessage(Response response, String expectedErrorMes
...
-        assertThat(logEvents.stream().anyMatch(e -> e.getFormattedMessage().contains("Received encrypted payload for charge with id")), is(false));
+
+    private void verifyReceiptOfPayloadNotLogged() {
+        Assertions.assertThat(logs.size())
+                .isZero();
     }
-    
-    
+
+
```
ralphm10: we can probably lose this line (I know the change is just from autoformatting)

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
```diff
@@ -293,7 +278,7 @@ private GatewayResponse mockProviderRespondedSuccessfullyResponse(String transac
                 .build();
     }
 
-    private GatewayResponse mockAuthErrorResponse(AuthoriseStatus authoriseStatus, String errorCode) {
+    private GatewayResponse mockAuthErrorResponse(String errorCode) {
```
ralphm10: we can just inline this as the value is always `error-code` (thanks IJ for the suggestion)

src/test/java/uk/gov/pay/connector/queue/tasks/TaskQueueServiceTest.java
```diff
@@ -201,11 +194,10 @@ void shouldSwallowExceptionThrownWhenAddingToQueue() throws Exception {
...
 
-        verify(mockAppender).doAppend(loggingEventArgumentCaptor.capture());
-
-        LoggingEvent loggingEvent = loggingEventArgumentCaptor.getValue();
-        assertThat(loggingEvent.getLevel(), is(Level.WARN));
-        assertThat(loggingEvent.getMessage(), is("Error adding payment task message to queue"));
+        logs.assertContains(event ->
+                        event.getLevel().equals(WARN) &&
+                                event.getMessage().equals("Error adding payment task message to queue"),
+                "Expected WARN event not found.");
```
ralphm10, line 200: yes, I prefer this way! (regarding previous comment about an ERROR event)

src/test/java/uk/gov/pay/connector/tasks/service/ChargeParityCheckerTest.java
```diff
@@ -609,7 +587,7 @@ private static Stream<Arguments> parityCheck_shouldReturnMismatchIfExemption3dsD
         );
     }
 
-    
+
```
ralphm10: again probably don't need this line

src/test/java/uk/gov/pay/connector/wallets/WalletAuthoriseServiceTest.java
```diff
@@ -432,8 +406,8 @@ void shouldRespondAuthorisationError() throws Exception {
         providerWillError();
         GatewayResponse authResponse = mockAuthResponse(null, AuthoriseStatus.ERROR, "error-code", null);
         when(mockedPaymentProvider.authoriseApplePay(any(ApplePayAuthorisationGatewayRequest.class))).thenReturn(authResponse);
-        
-        
+
+
```
ralphm10: and this one

ralphm10 (commented):
Overall I am very happy with this change. 
The addition of log unit makes the test code much more readable and succinct, and well done for tidying up the code as well. Great work. 
I have added a few very minor comments. 

My only question is around the use of AssertJ - as per first comment on https://github.com/alphagov/pay-webhooks/pull/1860
I think @kbottla would need to confirm

src/test/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementServiceTest.java
AndrzejProchyra: Yes, that is clearer! But I'm not now sure if we need to filter and supply a failure message at all. If instead we set up the `LogCapturer` for the ERROR log level only, thus
```java
@RegisterExtension                                                    
LogCapturer errorLogs = LogCapturer.create()                          
        .captureForType(LinkPaymentInstrumentToAgreementService.class)
        .forLevel(ERROR);                                             
```
we can assert thus
```java
errorLogs.assertContains("Expected charge " + chargeEntity.getExternalId() +
        " to have a payment instrument but it does not have one");          
```

A failed assertion then looks like this
```
org.opentest4j.AssertionFailedError: Contain the string <Expected charge fhm3o3igcd6aqeisqpftk18kpb to have a payment instrument but it does not have one> ==> None of the 1 captured log events matched the filter predicate
```

The current code results in the following failed assertion message
```
org.opentest4j.AssertionFailedError: Expected ERROR not found. ==> None of the 1 captured log events matched the filter predicate
```
I don't think we gain anything using the current approach.

What do you think?

src/test/java/uk/gov/pay/connector/paymentprocessor/resource/CardResourceAuthoriseTest.java
AndrzejProchyra: Agreed. Perhaps a more rigorous auto-formatter on a pre-commit hook would help with such things.

src/test/java/uk/gov/pay/connector/paymentprocessor/service/CardAuthoriseServiceTest.java
AndrzejProchyra: That is true! I am also disturbed by the use of `GatewayResponse` as a raw type in this code. However, the problem lies in the production code.

src/test/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementServiceTest.java
ralphm10: Yes I agree that your suggestion is much clearer

ralphm10 (approved):
I am satisfied that since my initial review, all AssertJ assertions have been replaced with Hamcrest equivalents. 

Great work.

PR #6276, opened 2026-04-16
PP-15163 Add contents:read permission to CodeQL workflow

kbottla:
## WHAT YOU DID
- Adds `content:read` permissions to the workflow, so that the permission can be passed down to the CodeQL job


DomBelcher (approved):
LGTM

PR #6292, opened 2026-04-28
PP-15197 Create Adyen gateway account in connector

kbottla:
## WHAT YOU DID

- Enables creating Adyen gateway accounts in Connector.
- Accepts `legal_entity_id` (Organisation ID) and `store_id` (to use to associate service to payments) as part of credentials. These are not mandatory


Pippa101 (approved):
LGTM!

PR #6305, opened 2026-05-05
PP-15355 Validate adyen notification domain

kbottla:
## WHAT YOU DID
- Validates Adyen notification IP address by doing reverse DNS lookup and matching the allowed notification domain.
- Minor improvements to inject AdyenGatewayConfig and extract custom ConnectorModule to use across tests


AbdirizakIdris (approved):
Looks good to me!

PR #6313, opened 2026-05-06
PP-15084 Make CardAuthorisationGatewayRequest constructor private

kbottla:
## WHAT YOU DID
- The constructor was public to be used only by tests, now tests use the static factory method, and the constructor is private


Pippa101 (approved):

PR #6317, opened 2026-05-07
PP-15084 Authorise an Adyen payment

kbottla:
## WHAT YOU DID

- Authorises an Adyen payment with minimal payload - only includes card details and billing address (doesn't include browser info required for 3DS payments).
- Doesn't handle unhappy paths or exceptions yet
- Introduces `logunit-logback` test dependency, which provides a custom JUnit extension and simplifies testing logging behaviour. It is also used in webhooks

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -404,6 +404,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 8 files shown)


Pippa101 (approved):

PR #6319, opened 2026-05-07
PP-15084 Handle Adyen authorisation gateway exceptions

kbottla:
## WHAT YOU DID
- Handles Gateway exceptions during payment authorisation with Adyen, and added tests for authorisation declines
- For payment declines or errors on Adyen's end, Adyen responds with a 200 HTTP status code and the status code is mapped to an internal status.
- Generates and returns AuthorisationSummary for logging based on the current data included in the authorisation request.


Pippa101 (approved):

PR #6329, opened 2026-05-12
PP-15262 Add Adyen Capture Webhook Notification To Task Queue

AbdirizakIdris:
## WHAT YOU DID
- Add Capture notification to connector task queue
- Ignore non Capture notifications and not add to task queue
- Handle error for when task queue submission fails and return 500 status code

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
@@ -5,6 +5,10 @@ import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.classic.spi.LoggingEvent;
 import ch.qos.logback.core.Appender;
+import com.adyen.model.notification.NotificationRequest;
+import com.adyen.model.notification.NotificationRequestItem;
+import com.adyen.notification.WebhookHandler;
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
```diff
@@ -230,4 +319,19 @@ private HmacKeys getHmacKeys(String... testKey) {
 
         return new HmacKeys(pair);
     }
+    
+    private String getNotificationWithValidHmacSignature(String eventCode) throws SignatureException, IOException {
```
AbdirizakIdris: Have to throw `SignatureException` and `IOException` exceptions because `handleNotificationJson` and `calculateHMAC` methods throw them. Unless I put a try catch block or these exceptions IntelliJ complains.

kbottla: It is unclear whether the main code is throwing exceptions or the test helpers are.

Can the exceptions be caught within the test helper, so there won't be confusion about tests throwing exceptions?

AbdirizakIdris: Ok I catch the  exceptions in the helper now to make it clear

kbottla (approved):
LGTM!

PR #6356, opened 2026-05-21
PP-15146: factor out an AdyenCheckoutMockClient

AndrzejProchyra:
## WHAT YOU DID
* Factored out `AdyenCheckoutMockClient`, since we have added `AdyenMockClient` subclasses for each Adyen API. Cancellation-specific stubbings will be added to this client.
* Pulled PATCH stubbing method up to `AdyenMockClient`
* Made some non-mutated private fields in `AppWithPostgressAndSqsExtension` final

## Observation
This pattern will lead to a WireMock server per Adyen API. We could use composition, rather than inheritance, to use one `WireMockServer` instance for Adyen.

### DIFF
--- a/src/test/java/uk/gov/pay/connector/rules/AdyenMockClient.java
+++ b/src/test/java/uk/gov/pay/connector/rules/AdyenMockClient.java
@@ -5,22 +5,32 @@ import uk.gov.pay.connector.util.TestTemplateResourceLoader;
 
 import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.github.tomakehurst.wiremock.client.WireMock.matching;
+import static com.github.tomakehurst.wiremock.client.WireMock.patch;
 import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
 import static jakarta.ws.rs.core.HttpHeaders.CONTENT_TYPE;
 import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
 import static org.apache.http.HttpStatus.SC_INTERNAL_SERVER_ERROR;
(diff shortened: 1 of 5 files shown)


src/test/java/uk/gov/pay/connector/rules/AdyenMockClient.java
```diff
@@ -5,22 +5,32 @@
...
 import static com.github.tomakehurst.wiremock.client.WireMock.post;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
 import static jakarta.ws.rs.core.HttpHeaders.CONTENT_TYPE;
 import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
 import static org.apache.http.HttpStatus.SC_INTERNAL_SERVER_ERROR;
-import static org.apache.http.HttpStatus.SC_OK;
 import static uk.gov.pay.connector.util.TestTemplateResourceLoader.ADYEN_ERROR_RESPONSE;
 
 public class AdyenMockClient {
+
```
kbottla, line 17: To prevent this class from being used for future mocks, should we make it abstract?

AndrzejProchyra: Yes, that makes sense. We can also make the `WireMockServer` private and the constructor package visible.

AndrzejProchyra: Should we put these in their own package? `connector.rules` does not really make sense now, since it does not contain a set of JUnit 4 `TestRule`s any more.

kbottla: It's a good point. These are also not rules, and it looks like all stubs or mock clients ended up in this package over time. We may need to look into them separately.

kbottla (approved):

PR #6363, opened 2026-05-26
PP-15385 Add resource to switch to Adyen test account

ralphm10:
## WHAT YOU DID

Add resource to switch from Stripe to Adyen test account and handle the below "unhappy" scenarios:

- Adyen responds with an error response to any request
- the current active provider of the account the switch is being requested for, is not Stripe
- no test account exists for the service switch is being requested for

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/AdyenAccountResource.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/resource/AdyenAccountResource.java
@@ -40,7 +40,7 @@ public class AdyenAccountResource {
     @Path("/v1/api/service/{serviceId}/request-adyen-test-account")
     @Produces(APPLICATION_JSON)
     @Operation(
-            summary = "Creates a Adyen Test Account " +
+            summary = "Creates an Adyen Test Account " +
                     "and associated entities",
             responses = {
                     @ApiResponse(responseCode = "201", description = "OK"),
@@ -77,5 +77,31 @@ public class AdyenAccountResource {
                 "account_holder_id", adyenCredentials.accountHolderId(),
                 "balance_account_id", adyenCredentials.balanceAccountId());
         return Response.ok(response).build();
+    }
+
+    @POST
+    @Path("/v1/api/service/{serviceId}/switch-to-adyen-test-account")
+    @Produces(APPLICATION_JSON)
+    @Operation(
+            summary = "Disables Stripe test account, " +
+                    "creates an Adyen Test Account " +
+                    "and associated entities",
+            responses = {
+                    @ApiResponse(responseCode = "201", description = "OK"),
+                    @ApiResponse(responseCode = "409", description = "Adyen account already exists"),
+                    @ApiResponse(responseCode = "502", description = "Bad gateway"),
+            }
+    )
+    public Response switchToAdyenTestAccount(
+            @Parameter(example = "service-external-id-123", description = "Service ID")
+            @PathParam("serviceId") String serviceId,
+            Map<String, String> payload) {
+        
+        gatewayAccountService.throwIfNoStripeTestAccount(serviceId);
+
+        var serviceName = payload.get("service_name");
+        adyenTestAccountService.createTestAccount(serviceName);
+
+        return Response.ok().build();
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccount/resource/AdyenAccountResource.java
```diff
@@ -40,7 +40,7 @@ public AdyenAccountResource(AdyenTestAccountService adyenTestAccountService,
     @Path("/v1/api/service/{serviceId}/request-adyen-test-account")
     @Produces(APPLICATION_JSON)
     @Operation(
-            summary = "Creates a Adyen Test Account " +
+            summary = "Creates an Adyen Test Account " +
                     "and associated entities",
```
AndrzejProchyra, line 44: Nit: Does this need to be over two lines?

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/AdyenAccountResource.java
```diff
@@ -78,4 +78,30 @@ public Response requestAdyenTestAccount(
...
+            @Parameter(example = "service-external-id-123", description = "Service ID")
+            @PathParam("serviceId") String serviceId,
+            Map<String, String> payload) {
+        
+        gatewayAccountService.throwIfNoStripeTestAccount(serviceId);
+
+        var serviceName = payload.get("service_name");
+        adyenTestAccountService.createTestAccount(serviceName);
+
+        return Response.ok().build();
```
AndrzejProchyra, line 103: Should this be `200 OK`, or `201 Created`? We document the "OK" response as `201 Created` above. Is this intentional?

src/test/java/uk/gov/pay/connector/it/resources/adyen/AdyenAccountResourceIT.java
```diff
@@ -176,20 +179,88 @@ void requestingTestAccountShouldReturnConflictWhenTestAccountExistsForService()
...
+    void switchTestAccountShouldReturnBadGatewayWhenAdyenRespondsWithError() {
+        var serviceId = RandomIdGenerator.randomUuid();
+
+        app.getDatabaseFixtures()
+                .aTestAccount()
+                .withServiceId(serviceId)
+                .withPaymentProvider(STRIPE.getName())
+                .insert();
+
+        app.getAdyenKycMockClient().mockError("/legalEntities");
```
AndrzejProchyra, line 213: What happens if the error happens later in the process?

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/AdyenAccountResource.java
ralphm10: updated

src/main/java/uk/gov/pay/connector/gatewayaccount/resource/AdyenAccountResource.java
ralphm10: good spot - updated

src/test/java/uk/gov/pay/connector/it/resources/adyen/AdyenAccountResourceIT.java
ralphm10: It should be the broadly the same - 502 response and no credentials switched (or created) 
In terms of requests, it would make them up until the error occurs, then make no subsequent ones.

AndrzejProchyra: Would this lead to an inconsistent state? Do we need to undo the previous actions in the "transaction"?

ralphm10: I don't believe this will cause any issues. 
For each request to Adyen, we receive a unique identifier with a successful response. For example, the creation of an organisation through `https://kyc-test.adyen.com/lem/v3/legalEntities` might return `LE32CTD223227N5PFQ8XW6RTR`. A subsequent call with the same inputs might return `LE3227C223222H5J4DCLW9VBV`.
Because we do not store the returned credentials in our database and create a gateway account until all API calls have succeeded, this should not cause an inconsistent state.
The consequence of any failed API calls would be that entities are created in the Adyen test environment which are never used. This should not cause any problems.

AndrzejProchyra (approved):

PR #6366, opened 2026-05-27
PP-15146: make handler fields and constructors similar

AndrzejProchyra:
## WHAT YOU DID
Made the fields and constructors of the Adyen cancel, authorise and capture handlers have a similar shape, in order to highlight the similarity.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthoriseHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthoriseHandler.java
@@ -22,18 +22,18 @@ public class AdyenAuthoriseHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(AdyenAuthoriseHandler.class);
 
-    private final GatewayClient gatewayClient;
-    private final AdyenGatewayConfig adyenGatewayConfig;
+    private final GatewayClient client;
+    private final AdyenGatewayConfig config;
+    private final AdyenRequestFactory requestFactory;
     private final JsonObjectMapper jsonObjectMapper;
-    private final AdyenRequestFactory adyenRequestFactory;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthoriseHandler.java
```diff
@@ -22,18 +22,18 @@ public class AdyenAuthoriseHandler {
 
     private static final Logger logger = LoggerFactory.getLogger(AdyenAuthoriseHandler.class);
 
-    private final GatewayClient gatewayClient;
```
kbottla, line 25: I think the naming of `gatewayClient`, `adyenGatewayConfig`, and `adyenRequestFactory` is clear to understand when reading the code and carries the context. Making them generic will make it hard to relate them (within methods) quickly to classes unless methods are simple, like in `AdyenRequestUtil`

AndrzejProchyra: Should we relate fields and variables to classes or to roles? I have renamed them to match the classes.

kbottla (approved):

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthoriseHandler.java
kbottla: Relating to roles is fine, but in this case, class names carry both the role and the context, which makes the code easier to scan and understand within the method without needing to look at the types.

AndrzejProchyra: If the roles are not clear from the context, maybe there is a problem with the clarity of the context...

PR #6375, opened 2026-05-29
PP-15264-handle-capture-webhook-messages-from-task-queue-edit

AbdirizakIdris:
## WHAT YOU DID
- change `LOGGER.error` to `LOGGER.warn` in accordance to PP-15264.


Pippa101 (approved):
LGTM

PR #6382, opened 2026-06-01
PP-15385 Add an endpoint to switch Stripe test accounts to Adyen

ralphm10:
## WHAT YOU DID

- Add method to switch from Stripe to Adyen test account
- Complete implementation of resource, partially done in https://github.com/alphagov/pay-connector/pull/6363
- Add tests to verify successful account switch


kbottla (approved):
LGTM!

PR #6384, opened 2026-06-02
PP-15146: parameterise response type of `GatewayResponseBuilder#withGatewayError`

AndrzejProchyra:
## WHAT YOU DID
`GatewayResponseBuilder#withGatewayError` returns `GatewayResponseBuilder` as a raw type, bypassing compile-time type checking and causing the compiler to emit "unchecked conversion" warnings.

`GatewayResponseBuilder#withGatewayError` should return the same type of `GatewayResponseBuilder` as the builder on which it is called.

There are just a few places where we need to properly parameterise the `GatewayResponseBuilder`, either by declaring a parameterised variable or using a type witness in the call to the `responseBuilder()` static creation method.


kbottla (approved):
LGTM!

PR #6389, opened 2026-06-03
 PP-15148 send 3ds info to adyen and map response toto 3ds required

AbdirizakIdris:
## WHAT YOU DID

 - Add `http_method_3ds field` to the charges table which is needed to be saved on the charge.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/model/domain/Auth3dsRequiredEntity.java
+++ b/src/main/java/uk/gov/pay/connector/charge/model/domain/Auth3dsRequiredEntity.java
@@ -17,7 +17,7 @@ public class Auth3dsRequiredEntity {
 
     @Column(name = "md_3ds")
     private String md;
-    
+
     @Column(name = "worldpay_challenge_acs_url_3ds")
     private String worldpayChallengeAcsUrl;
     
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -2100,4 +2100,12 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add http_method_3ds to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="http_method_3ds" type="varchar(10)">
+                <constraints nullable="true"/>
+            </column>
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/connector/charge/model/domain/Auth3dsRequiredEntity.java
```diff
@@ -17,6 +17,9 @@ public class Auth3dsRequiredEntity {
 
     @Column(name = "md_3ds")
     private String md;
+
+    @Column(name = "http_method_3ds")
+    private String httpMethod3ds;
```
kbottla: Class should only be updated after the DB migration is released and the column is added to the table

src/main/resources/migrations.xml
```diff
@@ -682,6 +682,14 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add http_method_3ds to charges table" author="">
+        <addColumn tableName="charges">
+            <column name="http_method_3ds" type="text">
```
kbottla: This can be a varchar(10) as HTTP method values are fixed in size

src/main/resources/migrations.xml
```diff
@@ -682,6 +682,14 @@
         </addColumn>
     </changeSet>
 
+    <changeSet id="add http_method_3ds to charges table" author="">
```
kbottla: Any reason this is added in between changesets? It can be added at the end of the XML file

src/main/java/uk/gov/pay/connector/charge/model/domain/Auth3dsRequiredEntity.java
AbdirizakIdris: Noted I'll remove it now and add it after migration is run.

src/main/resources/migrations.xml
AbdirizakIdris: I'll move it to the end

kbottla (approved):
LGTM!

PR #6396, opened 2026-06-04
PP-15119 convert Worldpay MOTO authorisation request record into XML

SandorArpa:
## WHAT YOU DID

Use Freemarker to convert the record WorldpayMotoAuthoriseRequest into an xml.

- Use TemplateBuilder as we eventually want to get rid of OrderRequestBuilder
- Use `ftlx` file extension for the template to ensure automatic escape (see https://freemarker.apache.org/docs/pgui_config_outputformatsautoesc.html)
- The code is adapted so that it can take a record as input (currently it can only take a TemplateData class)
- None of this new stuff is actually called from anywhere yet

### DIFF
--- a/src/main/resources/templates/worldpay/WorldpayAuthoriseMotoOrderTemplate.ftlx
+++ b/src/main/resources/templates/worldpay/WorldpayAuthoriseMotoOrderTemplate.ftlx
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE paymentService PUBLIC "-//WorldPay//DTD WorldPay PaymentService v1//EN"
+        "http://dtd.worldpay.com/paymentService_v1.dtd">
+<paymentService version="1.4" merchantCode="${merchantCode}">
+    <submit>
+        <order orderCode="${orderCode}">
+            <description>${description}</description>
+            <amount currencyCode="GBP" exponent="2" value="${amountInPence}"/>
+            <paymentDetails>
+                <CARD-SSL>
+                    <cardNumber>${cardNumber}</cardNumber>
+                    <expiryDate>
+                        <date month="${expiryDateMonthTwoDigits}" year="${expiryDateYearFourDigits}"/>
+                    </expiryDate>
+                    <cardHolderName>${cardholderName}</cardHolderName>
+                    <cvc>${cvc}</cvc>
+                </CARD-SSL>
+            </paymentDetails>
+        </order>
+    </submit>
+</paymentService>
--- a/src/test/java/uk/gov/pay/connector/gateway/templates/TemplateBuilderTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/templates/TemplateBuilderTest.java
@@ -1,0 +1,98 @@
+package uk.gov.pay.connector.gateway.templates;
+
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.xmlunit.assertj3.XmlAssert;
+import uk.gov.pay.connector.gateway.model.request.records.WorldpayMotoAuthoriseRequest;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.WORLDPAY_VALID_AUTHORISE_WORLDPAY_MOTO_AUTHORISATION_REQUEST;
+import static uk.gov.pay.connector.util.TestTemplateResourceLoader.load;
+
+class TemplateBuilderTest {
+
+    private static final String CARD_NUMBER = "4242424242424242";
+    private static final String EXPIRY_DATE_MONTH = "11";
+    private static final String EXPIRY_DATE_YEAR = "2030";
+    private static final String CARDHOLDER_NAME = "Alec Barley";
+    private static final String CVC = "123";
+    private static final String ORDER_CODE = "MyUniqueTransactionId";
+    private static final String DESCRIPTION = "My description";
+    private static final String USERNAME = "username"; // pragma: allowlist secret
+    private static final String PASSWORD = "password"; // pragma: allowlist secret
+    private static final String MERCHANT_CODE = "MERCHANTCODE";
+    private static final long AMOUNT_IN_PENCE = 2000L;
+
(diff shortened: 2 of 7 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/request/records/WorldpayRecordRequest.java
```diff
@@ -0,0 +1,4 @@
+package uk.gov.pay.connector.gateway.model.request.records;
+
+public interface WorldpayRecordRequest {
```
alexbishop1: I think we could just call this `WorldpayRequest`.

src/main/resources/templates/worldpay/WorldpayAuthoriseMotoOrderTemplate.ftlx
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE paymentService PUBLIC "-//WorldPay//DTD WorldPay PaymentService v1//EN"
+        "http://dtd.worldpay.com/paymentService_v1.dtd">
+<paymentService version="1.4" merchantCode="${merchantCode}">
+    <submit>
+        <order orderCode="${orderCode?xml}">
+            <description>${description?xml}</description>
+            <amount currencyCode="GBP" exponent="2" value="${amountInPence}"/>
```
alexbishop1, line 8: I was going to suggest that we use `?xml` on every variable just to be safe for now and look at enabling automatic escaping in the future.

However, it looks like auto-escaping can be enabled on a per-file basis by just changing the file extension. See https://freemarker.apache.org/docs/pgui_config_outputformatsautoesc.html for details. So if we rename this file to have a file extension of `ftlx` rather than `xml` and remove all the `?xml` built-ins, it should just escape everything for us.

We’d need to test this, of course, but this could be as simple as changing the merchant code in the test to be, say, `MERCHANT"CODE"` and making the cardholder name be, say, `Marley & Marley` (I picked those two because one is an attribute and one is text content) and checking they appear as `MERCHANT&quot;CODE&quot;` and `Marley &amp; Marley` (or possibly their numerical entity equivalents) respectively in the output.

Since it’s a simple change I think we should start as we mean to go on and do it now.

src/test/java/uk/gov/pay/connector/gateway/templates/TemplateBuilderTest.java
```diff
@@ -0,0 +1,74 @@
...
+    String expiryDateMonth = "11";
+    String expiryDateYear = "2030";
+    String cardholderName = "Alec Barley";
+    String cvc = "123";
+    String orderCode = "MyUniqueTransactionId";
+    String description = "My description";
+    String username = "username"; // pragma: allowlist secret
+    String password = "password"; // pragma: allowlist secret
+    String merchantCode = "MERCHANTCODE";
+    long amountInPence = 2000L;
```
alexbishop1: These can be constants.

src/test/java/uk/gov/pay/connector/gateway/templates/TemplateBuilderTest.java
```diff
@@ -0,0 +1,74 @@
...
+            assertThat(thrown.getMessage(), is("Could not load template /worldpay/NonExistentOrderTemplate.xml in dir /templates"));
+        }
+        
+        @Test
+        void shouldThrowRuntimeExceptionWhenCannotRenderTemplate() {
+            var thrown = assertThrows(RuntimeException.class, () -> {
+                TemplateBuilder templateBuilder = new TemplateBuilder("/worldpay/WorldpayCancelOrderTemplate.xml");
+                WorldpayMotoAuthoriseRequest motoOrder = createWorldpayMotoAuthoriseRequest();
+                templateBuilder.buildWith(motoOrder);
+            });
```
alexbishop1: I think if we did it like this:

```java
TemplateBuilder templateBuilder = new TemplateBuilder("/worldpay/WorldpayCancelOrderTemplate.xml");
WorldpayMotoAuthoriseRequest motoOrder = createWorldpayMotoAuthoriseRequest();
var thrown = assertThrows(RuntimeException.class, () -> templateBuilder.buildWith(motoOrder));
```

… it would be clearer precisely which bit we expect to throw the exception.

alexbishop1 (changes requested):
This looks good.

Main request is related to auto-escaping (which we probably should have decided on earlier, sorry) and the rest are minor.

src/test/java/uk/gov/pay/connector/gateway/templates/TemplateBuilderTest.java
```diff
@@ -0,0 +1,98 @@
...
+    private static final String expiryDateMonth = "11";
+    private static final String expiryDateYear = "2030";
+    private static final String cardholderName = "Alec Barley";
+    private static final String cvc = "123";
+    private static final String orderCode = "MyUniqueTransactionId";
+    private static final String description = "My description";
+    private static final String username = "username"; // pragma: allowlist secret
+    private static final String password = "password"; // pragma: allowlist secret
+    private static final String merchantCode = "MERCHANTCODE";
+    private static final long amountInPence = 2000L;
```
alexbishop1: `SCREAMING_SNAKE_CASE` for constant names, please.

src/test/java/uk/gov/pay/connector/gateway/templates/TemplateBuilderTest.java
```diff
@@ -0,0 +1,98 @@
...
+            XmlAssert.assertThat(templateBuilder.buildWith(motoOrder))
+                    .and(load(WORLDPAY_VALID_AUTHORISE_WORLDPAY_MOTO_AUTHORISATION_REQUEST))
+                    .areIdentical();
+        }
+    }
+    
+    @Nested
+    class TemplateBuilderAppliesAutoEscape {
+        @Test
+        void shouldGenerateValidAuthoriseOrderRequestWithAutoEscape() {
```
alexbishop1, line 45: This could just be merged into the previous test (by making some of the constants have characters that need escaping in them) but I’ll leave that up to you.

alexbishop1 (changes requested):
Looks great! One or two small things.

src/test/java/uk/gov/pay/connector/gateway/templates/TemplateBuilderTest.java
SandorArpa: I'd like to leave them separate. It's two different kind of tests and merging them might make them not so straightforward to understand.

alexbishop1 (approved):

PR #6399, opened 2026-06-05
PP-15264 Add missing Transactional annotation and integration test

kbottla:
## WHAT YOU DID
- Added `Transactional` annotation to the AdyenWebhookTaskHandler, so a new transaction starts and changes to the charge status are committed.
- Also, added an integration test which tests that the charge status is correctly updated in the database

with @christiandutton-mt, @ralphm10, @joshuapook-MT


ralphm10 (approved):
LGTM

PR #6401, opened 2026-06-08
PP-15146: Introduce `AdyenCancelResponse`

AndrzejProchyra:
## WHAT YOU DID
Introduced `AdyenCancelResponse` and mapping from a happy-path response and an error response.

Additionally:
* Added logging for failed cancellation as per Stripe

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
@@ -54,7 +54,7 @@ public class AdyenCancelHandler {
                     .build();
         } catch (GatewayErrorException e) {
             LOGGER.atError()
-                    .setMessage("Cancel failed for gateway transaction id {}. Charge External Id: {}.")
+                    .setMessage("Cancel failed for gateway transaction ID {}, external charge ID: {}.")
                     .addArgument(request.getTransactionId())
                     .addArgument(request.getExternalChargeId())
                     .log();
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
@@ -5,6 +5,7 @@ import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayException;
 import uk.gov.pay.connector.gateway.adyen.AdyenRequestFactory;
 import uk.gov.pay.connector.gateway.adyen.request.AdyenCancelRequest;
+import uk.gov.pay.connector.gateway.adyen.response.AdyenCancelResponse;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseCancelResponse;
 import uk.gov.pay.connector.gateway.model.response.GatewayResponse;
@@ -45,26 +46,8 @@ public class AdyenCancelHandler {
         } catch (GatewayException e) {
             return responseBuilder.withGatewayError(e.toGatewayError()).build();
         }
-        return responseBuilder.withResponse(new BaseCancelResponse() {
-            @Override
-            public String getTransactionId() {
-                return "";
-            }
-
-            @Override
-            public CancelStatus cancelStatus() {
-                return SUBMITTED;
-            }
-
-            @Override
-            public String getErrorCode() {
-                return "";
-            }
-
-            @Override
-            public String getErrorMessage() {
-                return "";
-            }
-        }).build();
+        return responseBuilder.withResponse(
+                        new AdyenCancelResponse("", SUBMITTED, "", ""))
+                .build();
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
@@ -3,9 +3,11 @@ package uk.gov.pay.connector.gateway.adyen.handler;
 import uk.gov.pay.connector.app.adyen.AdyenGatewayConfig;
 import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.GatewayException.GatewayErrorException;
 import uk.gov.pay.connector.gateway.adyen.AdyenRequestFactory;
 import uk.gov.pay.connector.gateway.adyen.request.AdyenCancelRequest;
 import uk.gov.pay.connector.gateway.adyen.response.AdyenCancelResponse;
+import uk.gov.pay.connector.gateway.adyen.response.json.AdyenError;
 import uk.gov.pay.connector.gateway.adyen.response.json.CancelResponseBody;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.response.BaseCancelResponse;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
```diff
@@ -41,30 +48,21 @@ public GatewayResponse<BaseCancelResponse> cancel(CancelGatewayRequest request)
                 request.getGatewayAccount().getType(),
                 jsonObjectMapper);
         try {
-            gatewayClient.postRequestFor(cancelRequest);
+            var jsonResponse = gatewayClient.postRequestFor(cancelRequest).getEntity();
+            var cancelResponse = jsonObjectMapper.getObject(jsonResponse, CancelResponseBody.class);
+            return responseBuilder.withResponse(AdyenCancelResponse.from(cancelResponse))
+                    .build();
+        } catch (GatewayErrorException e) {
+            LOGGER.atError()
```
kbottla: Can values be logged as kv pair?

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
```diff
@@ -41,30 +48,21 @@ public GatewayResponse<BaseCancelResponse> cancel(CancelGatewayRequest request)
...
+                    .setMessage("Cancel failed for gateway transaction ID {}, external charge ID: {}.")
+                    .addArgument(request.getTransactionId())
+                    .addArgument(request.getExternalChargeId())
+                    .log();
+            var adyenError = jsonObjectMapper.getObject(e.getResponseFromGateway(), AdyenError.class);
```
kbottla: This method could throw an exception if JSON is invalid. Can the exception be handled too?

src/main/java/uk/gov/pay/connector/gateway/adyen/response/AdyenCancelResponse.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    public String getErrorCode() {
+        return errorCode();
+    }
+
+    @Override
+    public String getErrorMessage() {
+        return errorMessage();
+    }
```
kbottla, line 46: Can toString() be added to this? toString() on response is being called here https://github.com/alphagov/pay-connector/blob/bc3e83e6c060c9bc59db354cc0abcc3ce518aa73/src/main/java/uk/gov/pay/connector/charge/service/ChargeCancelService.java#L97

src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandlerTest.java
```diff
@@ -172,9 +196,33 @@ void should_return_SUBMITTED_cancel_status() {
         assertThat(gatewayResponse.getBaseResponse().get().cancelStatus(), is(SUBMITTED));
     }
 
+    @Test
+    void should_map_transactionId_from_the_Adyen_response_paymentPspReference() throws GatewayException {
+        var paymentPspReference = "payment-psp-reference-from-adyen";
+        var adyenResponse = """
+                {
+                  "merchantAccount": "GovernmentDigitalService_UK",
```
kbottla:
```suggestion
                  "merchantAccount": "testMerchantAccount",
```

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
AndrzejProchyra: Yes, that is a good idea. 

For [Stripe we log at the ERROR level here](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/stripe/handler/StripeCancelHandler.java#L61), but for the other Adyen handlers we log at WARN. Should this be WARN?

What should be logged?

`AdyenCaptureHandler` logs:
* `CaptureGatewayRequest#gatewayTransactionId` as `"provider_payment_id"`
* `CaptureGatewayRequest#externalId` as `"payment_external_id"`
* The status code from the Adyen error as `"status_code"`
* The `GatewayErrorException` message as `"gateway_error"`

`AdyenRefundHandler` logs:
* `RefundGatewayRequest#refundExternalId` as `"refund_external_id"`
* `RefundGatewayRequest#chargeExternalId` as `"payment_external_id"`
* The status code from the `GatewayErrorException` error as `"status_code"`
* The `GatewayErrorException` message as `"gateway_error"`

`AdyenAuthoriseHandler` does not distinguish between `GatewayException` types and logs the exception in a generic way.

kbottla: The important one is `payment_external_id`, which can be used to track all logs. The rest of them depend on the context.

For Cancel, we can log `payment_external_id`, `status_code`, `gateway_error`

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
AndrzejProchyra: If the JSON is invalid, we catch the `IOException` from Jackson's `ObjectMapper`, log and throw a `WebApplicationException` in our `JsonObjectMapper`. We do not catch these for Stripe and allow the normal `WebApplicationException` process to take place. Should we change this behaviour?

src/main/java/uk/gov/pay/connector/gateway/adyen/response/AdyenCancelResponse.java
AndrzejProchyra: A Java record has an implicit `toString()`. So

```java
new AdyenCancelResponse(
                "a-transaction-id",
                CancelStatus.SUBMITTED,
                "an-error-code",
                "an-error-type",
                "an-error-message");
```

has a string representation of

"AdyenCancelResponse[transactionId=a-transaction-id, cancelStatus=SUBMITTED, errorCode=an-error-code, errorType=an-error-type, errorMessage=an-error-message]"

Should we **override** the `toString()`?

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
kbottla: For successful responses and errors returned by Adyen, we expect a valid JSON. There may be intermittent errors (CDN, network issues) that may return invalid JSON.

src/main/java/uk/gov/pay/connector/gateway/adyen/response/AdyenCancelResponse.java
kbottla: The default string representation looks fine to me

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
AndrzejProchyra: OK, how about this?

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandler.java
AndrzejProchyra: Yes, but what should we do on invalid JSON? 

* Allow the `WebApplicationException` thrown by `JsonObjectMapper#getObject()` to propagate, as we do for Stripe?
* Catch the `WebApplicationException` in `AdyenCancelHandler#cancel()`, log the failure for a second time and return a `GatewayResponse<BaseCancelResponse>` with the error?
* Something else?

kbottla: We can catch and return the gateway error, so that the charge status will be set to the relevant CANCEL_ERROR status.

Don't think WebApplicationException error is being handled, and it will result in a 500 API error with charge state left in non-terminal state.

AndrzejProchyra: OK, so the Stripe behaviour is not what we want? So here we should catch `WebApplicationException`?

AndrzejProchyra: OK, how about this?

kbottla (approved):

PR #6404, opened 2026-06-09
PP-15431 Added Idempotency Id mapping to Adyen requests

joshuapook-MT:
## WHAT YOU DID
- Added Idempotency Id mapping to Adyen requests for authorise, capture, cancel and refund
- Idempotency key is made up of the operation name prepended to charge external id (for authorise, capture and cancel) and refund external id (for refund) to ensure it's always unique for each operation

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtil.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtil.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.gateway.adyen.utils;
 
 import uk.gov.pay.connector.app.adyen.AdyenGatewayConfig;
+import uk.gov.pay.connector.gateway.GatewayOperation;
 import uk.gov.pay.connector.gateway.model.OrderRequestType;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
@@ -43,8 +44,8 @@ public class AdyenRequestUtil {
         return URI.create(getBaseCheckoutUrl(config, request.getGatewayAccount().isLive()) + path);
     }
 
-    public static Map<String, String> getHeaders(AdyenGatewayConfig config, boolean isLive, OrderRequestType requestType, String idempotencyKey) {
-        return Map.of("X-API-Key", getCompanyApiKey(config, isLive), 
-                "Idempotency-Key", format("%s-%s", requestType.toString(), idempotencyKey));
+    public static Map<String, String> getHeaders(AdyenGatewayConfig config, boolean isLive, GatewayOperation requestType, String idempotencyKey) {
+        return Map.of("X-API-Key", getCompanyApiKey(config, isLive),
+                "Idempotency-Key", format("%s-%s", requestType.getConfigKey(), idempotencyKey));
     }
 }
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtil.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtil.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.gateway.adyen.utils;
 
 import uk.gov.pay.connector.app.adyen.AdyenGatewayConfig;
+import uk.gov.pay.connector.gateway.model.OrderRequestType;
 import uk.gov.pay.connector.gateway.model.request.CancelGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CaptureGatewayRequest;
 import uk.gov.pay.connector.gateway.model.request.CardAuthorisationGatewayRequest;
@@ -10,6 +11,7 @@ import uk.gov.pay.connector.gateway.model.request.RefundGatewayRequest;
 import java.net.URI;
 import java.util.Map;
 
+import static java.lang.String.format;
 import static uk.gov.pay.connector.gateway.adyen.utils.AdyenConfigUtil.getBaseCheckoutUrl;
 import static uk.gov.pay.connector.gateway.adyen.utils.AdyenConfigUtil.getCompanyApiKey;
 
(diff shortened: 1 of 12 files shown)


src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthoriseHandlerTest.java
```diff
@@ -30,9 +30,8 @@
 
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.hasProperty;
-import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.*;
```
DomBelcher: As a rule we avoid using wildcard imports - you can set intellij to do this by setting the threshold to be really high

src/test/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtilTest.java
AndrzejProchyra: We should probably rename this test now to something like

```suggestion
    void should_create_API_key_and_idempotency_key_headers_for_checkout_URL() {
```

Or create a new test to make it clear how this works.

src/main/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtil.java
```diff
@@ -41,7 +43,8 @@ private static URI getUrl(AdyenGatewayConfig config, GatewayRequest request, Str
         return URI.create(getBaseCheckoutUrl(config, request.getGatewayAccount().isLive()) + path);
     }
 
-    public static Map<String, String> getHeaders(AdyenGatewayConfig config, boolean isLive) {
-        return Map.of("X-API-Key", getCompanyApiKey(config, isLive));
+    public static Map<String, String> getHeaders(AdyenGatewayConfig config, boolean isLive, OrderRequestType requestType, String idempotencyKey) {
```
DomBelcher: I like the inclusion of the `requestType` as a parameter rather than leaving it up to the calling code to correctly format the idempotency key, nice 👍

AndrzejProchyra: Could we pass the whole `GatewayRequest` here? Then we could do
* `request.getRequestType()` instead of `requestType`
* `request.getGatewayAccount().isLive()` instead of `isLive`

Unfortunately `GatewayRequest` does not expose a suitable idempotency key so that would still need to be passed in.

src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthoriseHandlerTest.java
```diff
@@ -121,6 +122,9 @@ void should_send_request_to_Adyen_with_no_billing_address() throws GatewayExcept
         then(mockClient).should().postRequestFor(captor.capture());
         String payload = captor.getValue().getGatewayOrder().getPayload();
         JsonAssert.with(payload).assertNotDefined("$.billingAddress");
+        var headers = captor.getValue().getHeaders();
+        assertThat(headers, hasEntry("X-API-Key", TEST_API_KEY));
+        assertThat(headers, hasEntry("Idempotency-Key", "authorise-" + authoriseRequest.getGovUkPayPaymentId()));
```
AndrzejProchyra: Should we make these assertions in a separate test?

src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandlerTest.java
```diff
@@ -141,6 +141,7 @@ void should_POST_with_company_API_key_as_X_API_Key_header(GatewayAccountType gat
                 .postRequestFor(captor.capture());
         var headers = captor.getValue().getHeaders();
         assertThat(headers, hasEntry("X-API-Key", expectedApiKey));
+        assertThat(headers, hasEntry("Idempotency-Key", "cancel-" + request.getExternalChargeId()));
```
AndrzejProchyra, line 144: Should this go in a separate test?

src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenRefundHandlerTest.java
```diff
@@ -145,6 +147,10 @@ void shouldSendCorrectPayloadToAdyenForPartialRefund() throws Exception {
                 .assertThat("$.amount.currency", is("GBP"))
                 .assertThat("$.reference", is(REFUND_EXTERNAL_ID))
                 .assertThat("$.store", is(STORE_ID));
+        var headers = captor.getValue().getHeaders();
+        assertThat(headers, hasEntry("X-API-Key", TEST_API_KEY));
+        assertThat(headers, hasEntry("Idempotency-Key", "refund-" + request.getRefundExternalId()));
```
AndrzejProchyra: Should these go in a separate test?

src/main/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtil.java
joshuapook-MT: You could to be fair. I would say you're giving the method way too much scope for what it needs that way though. I hadn't seen that getRequestType method, so I've made the compromise of using that to pass the value in rather than the enum.

src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthoriseHandlerTest.java
joshuapook-MT: Added new test and reverted this one

src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenCancelHandlerTest.java
joshuapook-MT: I've just updated the name appropriately :)

src/test/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenRefundHandlerTest.java
joshuapook-MT: Done :)

src/main/java/uk/gov/pay/connector/gateway/adyen/utils/AdyenRequestUtil.java
AndrzejProchyra: Well, the method is "getting headers for the request". 

We would also not need to pass in the `AdyenGatewayConfig` if this method was on an object that had a reference to the config, and the method could know how to generate the whole idempotency key from a request if we exposed the proper ID in a uniform way. The alternative is a method that takes 4 parameters and requires the client to know where to get them from. i.e. the client needs to know a lot about how it works to use this.

AndrzejProchyra (approved):

PR #6405, opened 2026-06-09
PP-15433 Add Adyen gateway account setup tasks table

christiandutton-mt:
## WHAT YOU DID
Created gateway_accounts_adyen_setup table in migrations.xml with the following columns - 

- id, serial, primary key

- task, varchar(50), non nullable

- status, varchar(50), non nullable

- gateway_account_id, bigint, non-nullable

constraints:

gateway_account_id foreign key to the gateway_accounts table

Index:

Add an index for the field gateway_account_id

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -2108,4 +2108,29 @@
         </addColumn>
     </changeSet>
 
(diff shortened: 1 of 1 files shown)


kbottla (approved):
LGTM!

PR #6414, opened 2026-06-12
PP-14481 Use KV pairs in logs

ralphm10:
## WHAT YOU DID

- update logs to use key value pairs for resource IDs


kbottla (approved):
LGTM!

PR #6423, opened 2026-06-15
PP-15165 Update test configs logging level to INFO

kbottla:
## WHAT
- Reverts alphagov/pay-connector#6278 as Connector repo is now public


Rafiq-Gasmi (approved):
LGTM

PR #6428, opened 2026-06-16
PP-15146 Overwrite to string in AdyenCancelResponse

ralphm10:
## WHAT YOU DID

- Overwrite `toString()` method in Adyen Cancel Response
- This simplifies logging and makes consistent with [WorldpayCancelResponse.java](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayCancelResponse.java#L31:L42)


kbottla (approved):
LGTM!

PR #6430, opened 2026-06-17
PP-15472 - added all payment events to join task queue logic

christiandutton-mt:
## WHAT YOU DID

- Added enum to hold all adyen payment events
- Integrated using enum in existing logic (for capture) to check if the event code exists if the enum - if it does, proceed with adding notification to task queue.
- Updated singular test (for capture) to run against all of the values in the enum. 
- Updated test that was previously checking "AUTHORISATION" event to NOT be added to the task queue to now check for null, empty and "SOME_INVALID_VALUE"
- Renamed `HANDLE_ADYEN_WEBHOOK_NOTIFICATION` to `HANDLE_ADYEN_PAYMENTS_WEBHOOK_NOTIFICATION` and updated usages.


kbottla (dismissed):
LGTM!

kbottla (approved):

PR #6444, opened 2026-06-23
PP-15514 Added gateway_account_credential_id to gateway_accounts_adyen_setup table

joshuapook-MT:
## WHAT YOU DID
- Added gateway_account_credential_id as a foreign key to gateway_accounts_adyen_setup table
- Added mappings to set the credentials Id when a test account is created 
- Added mappings to set the credentials Id when a stripe account is switched to an adyen account

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccount/model/AdyenAccountSetupTaskEntity.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccount/model/AdyenAccountSetupTaskEntity.java
@@ -11,6 +11,8 @@ import jakarta.persistence.JoinColumn;
 import jakarta.persistence.ManyToOne;
 import jakarta.persistence.SequenceGenerator;
 import jakarta.persistence.Table;
+import uk.gov.pay.connector.gatewayaccountcredentials.model.GatewayAccountCredentialsEntity;
+import static uk.gov.pay.connector.gateway.PaymentGatewayName.ADYEN;
 
 @Entity
 @Table(name = "gateway_accounts_adyen_setup")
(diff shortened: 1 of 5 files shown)


src/main/resources/migrations.xml
```diff
@@ -2133,4 +2133,16 @@
         </createIndex>
     </changeSet>
 
+    <changeSet id="add gateway_account_credential_id to gateway_accounts_adyen_setup table" author="">
```
kbottla, line 2136: DB changes should be their own PR, followed by a code change PR

kbottla (changes requested):

src/main/java/uk/gov/pay/connector/gatewayaccount/model/AdyenAccountSetupTaskEntity.java
```diff
@@ -38,6 +44,7 @@ public AdyenAccountSetupTaskEntity(GatewayAccountEntity gatewayAccount, AdyenAcc
         this.gatewayAccount = gatewayAccount;
         this.task = task;
         this.status = status;
+        this.gatewayAccountCredential = gatewayAccount.getGatewayAccountCredentialsEntity(ADYEN.getName());
```
kbottla: Can we not add business logic here? It would be good if service classes responsible for creating tasks could pass the credentials entity. 

It is also possible that there can be multiple credentials for Adyen (Adyen-to-Adyen switch), so getGatewayAccountCredentialsEntity may not return the expected credential.

src/main/resources/migrations.xml
```diff
@@ -2140,7 +2140,7 @@
                         foreignKeyName="fk__gateway_account_credentials_adyen_setup_gateway_accounts"
                         referencedTableName="gateway_account_credentials"
                         referencedColumnNames="id"
-                        nullable="true"/>
+                        nullable="false"/>
```
kbottla: This should be a new change set and can be its own PR.

kbottla (approved):
LGTM!

PR #6450, opened 2026-06-25
PP-15120 AuthorisationLogger logs WorldpayAuthoriseRequest

alexbishop1:
Introduce a new `WorldpayAuthoriseRequest` interface, which extends `WorldpayRequest` and is implemented by `WorldpayMotoAuthoriseRequest`.

Enhance `AuthorisationLogger` so that it can log `WorldpayAuthoriseRequest` records (just `WorldpayMotoAuthoriseRequest` for now but this will be expanded in the future) by using a new `WorldpayAuthoriseRequestLogGenerator` class, which takes a `WorldpayAuthoriseRequest` and produces both the request part of the log line string and also the structured logging arguments (this is in contrast to the existing `AuthorisationRequestSummaryStringifier` and `AuthorisationRequestSummaryStructuredLogging` classes, which generate the two parts separately).

This approach bypasses the need for an intermediary `AuthorisationRequestSummary` class, instead using the `WorldpayAuthoriseRequest` record directly.

Some data that’s logged, like whether or not a corporate card is being used, are not part of the request sent to Worldpay so have to be retrieved from the `AuthCardDetails` object.

The existing `AuthorisationLogger` functionality, such as logging an `AuthorisationRequestSummary`, remains unchanged.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/model/request/records/WorldpayAuthoriseRequest.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/model/request/records/WorldpayAuthoriseRequest.java
@@ -1,0 +1,4 @@
+package uk.gov.pay.connector.gateway.model.request.records;
+
+public interface WorldpayAuthoriseRequest extends WorldpayRequest {
+}
--- a/src/main/java/uk/gov/pay/connector/gateway/util/WorldpayAuthoriseRequestLogGenerator.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/util/WorldpayAuthoriseRequestLogGenerator.java
@@ -1,0 +1,53 @@
+package uk.gov.pay.connector.gateway.util;
+
+
+import net.logstash.logback.argument.StructuredArgument;
+import uk.gov.pay.connector.gateway.model.AuthCardDetails;
+import uk.gov.pay.connector.gateway.model.request.records.WorldpayAuthoriseRequest;
+import uk.gov.pay.connector.gateway.model.request.records.WorldpayMotoAuthoriseRequest;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.StringJoiner;
+
+import static net.logstash.logback.argument.StructuredArguments.kv;
+import static uk.gov.pay.connector.gateway.util.AuthorisationRequestSummaryStructuredLogging.BILLING_ADDRESS;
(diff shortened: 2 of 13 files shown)


src/main/java/uk/gov/pay/connector/gateway/model/request/records/WorldpayAuthoriseRequest.java
```diff
@@ -0,0 +1,4 @@
+package uk.gov.pay.connector.gateway.model.request.records;
+
+public interface WorldpayAuthoriseRequest extends WorldpayRequest {
+}
```
GabrielBoardman-MadeTech, line 4: Ignore if this is team standard, but I’d normally keep empty interface braces on one line

src/main/java/uk/gov/pay/connector/gateway/util/WorldpayAuthoriseRequestLogGenerator.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private AuthorisationRequestLog generate(WorldpayMotoAuthoriseRequest worldpayMotoAuthoriseRequest, AuthCardDetails authCardDetails) {
+        List<StructuredArgument> structuredArguments = new ArrayList<>();
+
+        var stringJoiner = new StringJoiner(" and ", " ", "");
+
+        stringJoiner.add("without billing address");
+        structuredArguments.add(kv(BILLING_ADDRESS, false));
+
+        if (authCardDetails.isCorporateCard())  {
```
GabrielBoardman-MadeTech: Two spaces between closing bracket and opening curly

src/main/java/uk/gov/pay/connector/gateway/model/request/records/WorldpayAuthoriseRequest.java
alexbishop1: Not sure we have a standard but I checked and everywhere in connector where we have an interface or class with no body we put the closing brace on its own line

src/main/java/uk/gov/pay/connector/gateway/util/WorldpayAuthoriseRequestLogGenerator.java
alexbishop1: How did that sneak in there? Will fix!

GabrielBoardman-MadeTech (approved):
LGTM

PR #6453, opened 2026-06-29
PP-15121 Refactor WorldpayAuthoriseHandler — extract private method

alexbishop1:
Extract the part of the `authorise(…)` method that invokes the `authoriseClient` to send the request and then handles to response into its own reusable private method.


GabrielBoardman-MadeTech (approved):

PR #6454, opened 2026-06-29
PP-15148-send-3ds-info-to-adyen-and-map-to-3ds-response-parse-ints

AbdirizakIdris:
##   Context

 - From review it was found:
	- Some strings are converted to an integer when building the browserInfo object. There is a possibility that these will result in an exception when the frontend doesn’t send data (for null values- when JavaScript is disabled, and invalid integers for some reason )

##   What you did

- Add safe parsing for browser information fields that are expected to be integers (colourDepth, screenHeight, screenWidth and timeZoneOffset)

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
@@ -1,5 +1,7 @@
 package uk.gov.pay.connector.gateway.adyen;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.common.model.domain.Address;
 import uk.gov.pay.connector.gateway.adyen.request.json.Amount;
@@ -25,12 +27,14 @@ import uk.gov.pay.connector.northamericaregion.NorthAmericanRegionMapper;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 
 import static uk.gov.pay.connector.gateway.adyen.utils.AdyenConfigUtil.getMerchantAccountId;
 
 public class AdyenRequestFactory {
 
     private final ConnectorConfiguration configuration;
+    private final Logger LOGGER = LoggerFactory.getLogger(AdyenRequestFactory.class);
 
     public AdyenRequestFactory(ConnectorConfiguration configuration) {
         this.configuration = configuration;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
```diff
@@ -125,18 +129,27 @@ private static AdyenCredentials mapToAdyenCredentials(GatewayCredentials gateway
...
-                authCardDetails.getJsTimezoneOffsetMins().map(Integer::valueOf).orElse(null),
+                authCardDetails.getJsNavigatorLanguage().orElse(null),
+                parseInteger(authCardDetails.getJsScreenHeight()),
+                parseInteger(authCardDetails.getJsScreenWidth()),
+                parseInteger(authCardDetails.getJsTimezoneOffsetMins()),
                 authCardDetails.getUserAgentHeader()
         );
     }
+    
+    private Integer parseInteger(Optional<String> value) {
```
kbottla: Can the value be passed instead of Optional? Optional is intended for use as a return type. See API note in **[Java docs](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/util/Optional.html)**

Due to the Optional being passed, when JavaScript is disabled, values will always be null, and logged as warnings

src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
```diff
@@ -125,18 +129,27 @@ private static AdyenCredentials mapToAdyenCredentials(GatewayCredentials gateway
         }
         return (AdyenCredentials) gatewayCredentials;
     }
-
+    
     private BrowserInfo mapToBrowserInfo(AuthCardDetails authCardDetails) {
         return new BrowserInfo(
                 authCardDetails.getAcceptHeader(),
-                authCardDetails.getJsScreenColorDepth().map(Integer::valueOf).orElse(null),
+                parseInteger(authCardDetails.getJsScreenColorDepth()),
```
kbottla: We could do something like this to avoid passing Optional as an argument.

```suggestion
                authCardDetails.getJsScreenColorDepth().flatMap(this::parseInteger).orElse(null),
```

src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
```diff
@@ -125,18 +129,26 @@ private static AdyenCredentials mapToAdyenCredentials(GatewayCredentials gateway
...
+                authCardDetails.getJsTimezoneOffsetMins().flatMap(this::parseInteger).orElse(null),
                 authCardDetails.getUserAgentHeader()
         );
     }
+
+    private Optional<Integer> parseInteger(String value) {
+        try {
+            return Optional.of(Integer.valueOf(value));
+        } catch (NumberFormatException e) {
+            LOGGER.warn("Unable to parse browser info integer '{}'", value, e);
```
kbottla, line 150: It would be good to identify the property for which the warning is logged and use kv pair for logging. It can be in a separate PR.

Something like below:

```suggestion
            LOGGER.atWarn()
                    .setMessage("Unable to parse browser info integer")
                    .addKeyValue("property", property)
                    .addKeyValue("value", value)
                    .setCause(e)
                    .log();
```

kbottla (approved):
LGTM!

Added a comment about logging which can be addressed in a separate PR.

src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
AbdirizakIdris: Ok I'll add better logging in the next pr

PR #6455, opened 2026-06-29
PP-15515 Updated idempotency key to use OrderRequestType for Adyen requests

joshuapook-MT:
## WHAT YOU DID
- Updated idempotency key to use OrderRequestType for Adyen requests


Pippa101 (approved):
LGTM

PR #6458, opened 2026-06-29
PP15417 Fix additional browser info API field name

kbottla:
## WHAT YOU DID

- Updated field name `js_timezone_offset_min` to `js_timezone_offset_mins`, which is what the Frontend is sending to Connector https://github.com/alphagov/pay-frontend/blob/26243da0120be74d222b75d2f0404ee964ce0a68/app/services/normalise-charge.js#L187


Pippa101 (approved):
LGTM

PR #6468, opened 2026-06-30
PP-15121 Rename AuthoriseRequest to MotoApiAuthoriseRequest

alexbishop1:
This will free up the generic name `AuthoriseRequest` for other uses.


GabrielBoardman-MadeTech (dismissed):

GabrielBoardman-MadeTech (approved):

PR #6471, opened 2026-07-01
PP-15121 Introduce AuthoriseRequest sealed interface

alexbishop1:
Introduce `AuthoriseRequest` interface that is extended by the existing `WorldpayAuthoriseRequest` interface to avoid having PSP-specific types used in generic code.

Make `AuthoriseRequest` and `WorldpayAuthoriseRequest` sealed interfaces so we can take advantage of features like exhaustiveness checking in switch statements.


SandorArpa (approved):

PR #6476, opened 2026-07-02
PP-15121 Call AuthorisationLogger with AuthoriseRequest

alexbishop1:
If we create an `AuthoriseRequest` record when authorising a payment, call `AuthorisationLogger` with it and use it as the source of metrics because this can provide more accurate data about what was actually sent to the gateway.

Since `AuthoriseRequest` records are never currently created, this new code is not used yet.


GabrielBoardman-MadeTech (approved):

PR #6477, opened 2026-07-02
PP-14487 - Control Adyen 3DS handler response and improve test coverage

Rafiq-Gasmi:
## WHAT YOU DID
- Use sterilised body to explicitly control what is returned
- Add WireMock verification to test if expected fields are sent
- Simplify exception handling with explicit imports
- Override toString method for Authorise3dsResponseBody with unit tests
- 
## How to test

- What feedback do you need?  PR comments/DM
- If manual testing is needed, give suggested testing steps. NA

### Logging

- [X] only emit log lines at ERROR level which require immediate attention from a support engineer. These will trigger a zendesk alert.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthorise3dsHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthorise3dsHandler.java
@@ -6,6 +6,7 @@ import uk.gov.pay.connector.app.ConnectorConfiguration;
 import uk.gov.pay.connector.app.adyen.AdyenGatewayConfig;
 import uk.gov.pay.connector.gateway.GatewayClient;
 import uk.gov.pay.connector.gateway.GatewayException;
+import uk.gov.pay.connector.gateway.GatewayException.GatewayErrorException;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthorise3dsHandler.java
```diff
@@ -67,11 +68,11 @@ public Gateway3DSAuthorisationResponse authorise3dsResponse(Auth3dsResponseGatew
             var mappedStatus = mapStatus(responseBody.resultCode());
 
             return Gateway3DSAuthorisationResponse.of(
-                    jsonResponse,
+                    responseBody.toString(),
```
kbottla, line 71: Do we have a toString() implementation and a relevant test similar to https://github.com/alphagov/pay-connector/blob/efde32cbc80fc9187c5e7d8482133b46f2d8f16f/src/main/java/uk/gov/pay/connector/gateway/worldpay/WorldpayOrderStatusResponse.java#L311?

kbottla (approved):
LGTM!

PR #6479, opened 2026-07-03
PP-15121 Add AuthoriseRequestFactory and friends

alexbishop1:
Add `AuthoriseRequestFactory` and `WorldpayAuthoriseRequest` factory. The former figures out if it’s a Worldpay payment and calls the latter. The latter figures out if it’s a web MOTO payment and calls the `WorldpayMotoAuthoriseRequestFactory`. If not, an empty optional is returned. This will allow us to gradually increase the range of `AuthoriseRequest` implementations we support.


GabrielBoardman-MadeTech (approved):

PR #6480, opened 2026-07-03
PP-15121 Use PaymentGatewayName.WORLDPAY.getName() not .toString()

alexbishop1:
Use `PaymentGatewayName.WORLDPAY.getName()` (which returns a string like “worldpay” rather than `PaymentGatewayName.WORLDPAY.toString()` (which returns a string like “WORLDPAY”).


SandorArpa (approved):

SandorArpa (approved):

PR #6493, opened 2026-07-08
PP-15121 Use WorldpayMotoAuthoriseRequest when appropriate

alexbishop1:
If a payment:
- is Worldpay
- is authorisation mode WEB
- is MOTO

… then authorise it by creating a `WorldpayMotoAuthoriseRequest` record.


GabrielBoardman-MadeTech (approved):

PR #6498, opened 2026-07-09
PP-15510 update open API spec for Adyen account setup resource

ralphm10:
## WHAT YOU DID

-  update open API spec to use AdyenAccountSetupResponse and provide example response object in schema


Pippa101 (approved):
LGTM

PR #6499, opened 2026-07-09
PP-15121 Build multi-release JAR

alexbishop1:
Configure the Maven shade plugin to produce a multi-release JAR, so dependencies that use multi-release JARs to can enable certain features on newer Java versions (such as support for records in Freemarker).


GabrielBoardman-MadeTech (approved):

PR #6503, opened 2026-07-10
PP-15511 Add an endpoint to patch adyen setup task

ralphm10:
## WHAT YOU DID

- Add an endpoint to patch adyen setup task
- Add integration and unit tests for task that does/does not exist, and multiple update requests
- Renamed `StripeSetupPatchRequest` to `AccountSetupPatchRequest`, converted to record and used for Adyen patch requests

**NB. this PR covers happy paths only, error handling to be addressed in subsequent PRs**


Pippa101 (approved):
LGTM

PR #6511, opened 2026-07-14
PP-15511 Add integration tests for unprocessable entries in patch request

ralphm10:
## WHAT YOU DID

PP-15511

- Add integration tests for unprocessable entries in patch request
- Add `@Valid` annotation to resource
- Add `@Valid` annotation to value field of `AccountSetupPatchRequest` 
- Add `@AllowedStrings` annotation, allowed strings and error message to value field of `AccountSetupPatchRequest` 
  - This includes valid fields of true/false which are used in the case of updating Stripe account setup tasks as `AccountSetupPatchRequest` is now used for both Adyen and Stripe accounts

**NB. Because `@Valid` annotation is used, `AdyenAccountSetupResource` now returns `422` (unprocessable content) as opposed to `400` (Bad Request) as defined in scenarios. This matches existing behaviour of Stripe endpoint - to return 400 instead would require adding more code**

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -196,14 +196,14 @@
         "filename": "openapi/connector_spec.yaml",
         "hashed_secret": "40a714e4a70aa9b75e73dc27be3a978fa98ee4e3",
         "is_verified": false,
-        "line_number": 6093
+        "line_number": 6102
       },
       {
         "type": "Hex High Entropy String",
         "filename": "openapi/connector_spec.yaml",
         "hashed_secret": "ec969b7613112e3df2b11d8f64ce37f0d20bf29d",
         "is_verified": false,
-        "line_number": 6104
+        "line_number": 6113
       }
     ],
     "src/main/java/uk/gov/pay/connector/agreement/resource/AgreementsApiResource.java": [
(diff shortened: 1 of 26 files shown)


DomBelcher (commented):
A few thoughts, not directly related to this PR:

- What if the value is `true` or `false` in an Adyen request?
- Is adding the validation to something used by the Stripe endpoint going to cause any issues?
- given those, would it be better to have a separate `AdyenAccountSetupPatchRequest` class for this endpoint, and leave the existing one for the Stripe setup endpoint?

ralphm10:
> A few thoughts, not directly related to this PR:
> 
> * What if the value is `true` or `false` in an Adyen request?
> * Is adding the validation to something used by the Stripe endpoint going to cause any issues?
> * given those, would it be better to have a separate `AdyenAccountSetupPatchRequest` class for this endpoint, and leave the existing one for the Stripe setup endpoint?

All good points!
We were trying to avoid creating another `SetupPatchRequest` class if possibly which would essentially be identical but agree this is probably necessary and the best solution.

DomBelcher (commented):
Nice work

DomBelcher (approved):

PR #6517, opened 2026-07-15
PP-14497 - Create a new Adyen tokens webhook endpoint 

Rafiq-Gasmi:
## WHAT YOU DID
Create a new Adyen tokens webhook endpoint in Connector to handle recurring token events
Unit Test and IT

## How to test

- What feedback do you need? PR comments/ DM
- If manual testing is needed, give suggested testing steps. NA

## Code review checklist
[x] Clean code
[x] Logging
[x] fail fast 

### Logging


Rafiq-Gasmi:
split the ticket

PR #6526, opened 2026-07-17
PP-15617 Worldpay web MOTO payments allow payment instrument setup

alexbishop1:
Restore the ability to save a payment instrument to an agreement when authorising a Worldpay MOTO payment with authorisation mode `WEB`.

This fixes a regression introduced by PP-15121.

The fix involves not creating a `WorldpayMotoAuthoriseRequest` and instead going down the classic path if `save_payment_instrument_to_agreement` is `true`.

Also reorder the checks when deciding whether to create a `WorldpayMotoAuthoriseRequest` record so that it’s more likely to bail out early, which is the common case.

This does not affect MOTO authorisation API payments (authorisation mode of `MOTO_API`), which have never supported saving payment instruments to agreements.


marcotranchino (approved):

SandorArpa (approved):
LOKTM

PR #6533, opened 2026-07-21
PP-15568: Add check for empty recurringAuthToken so PaymentInstrument cannot be set to Active unless token is present.

Pippa101:
## WHAT YOU DID
- Added conditional to check if `recurringAuthToken` is not empty before linking `paymentInstrument` to `agreement` and setting `paymentInstrument` as `ACTIVE`
- updated mocking for tests that assumed `recurringAuthToken` would be present but had not mocked it out. 

### Logging

- I added logging at info level

### DIFF
--- a/src/main/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementService.java
+++ b/src/main/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementService.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.connector.charge.service;
 
 import com.google.inject.persist.Transactional;
+import jakarta.inject.Inject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.agreement.model.AgreementEntity;
@@ -13,10 +14,11 @@ import uk.gov.pay.connector.paymentinstrument.model.PaymentInstrumentEntity;
 import uk.gov.pay.connector.paymentinstrument.model.PaymentInstrumentStatus;
 import uk.gov.pay.connector.queue.tasks.TaskQueueService;
 
-import jakarta.inject.Inject;
 import java.time.InstantSource;
 import java.util.List;
+import java.util.Map;
 
+import static java.util.function.Predicate.not;
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static uk.gov.service.payments.logging.LoggingKeys.AGREEMENT_EXTERNAL_ID;
 import static uk.gov.service.payments.logging.LoggingKeys.PAYMENT_INSTRUMENT_EXTERNAL_ID;
@@ -33,7 +35,7 @@ public class LinkPaymentInstrumentToAgreementService {
     @Inject
     public LinkPaymentInstrumentToAgreementService(PaymentInstrumentDao paymentInstrumentDao,
                                                    LedgerService ledgerService,
-                                                   TaskQueueService taskQueueService, 
+                                                   TaskQueueService taskQueueService,
                                                    InstantSource instantSource) {
         this.paymentInstrumentDao = paymentInstrumentDao;
         this.ledgerService = ledgerService;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementService.java
```diff
@@ -45,17 +45,23 @@ public LinkPaymentInstrumentToAgreementService(PaymentInstrumentDao paymentInstr
...
-                paymentInstrumentEntity.setAgreementExternalId(agreementEntity.getExternalId());
-                paymentInstrumentEntity.setStatus(PaymentInstrumentStatus.ACTIVE);
-                ledgerService.postEvent(List.of(
-                        AgreementSetUp.from(agreementEntity, instantSource.instant()),
-                        PaymentInstrumentConfirmed.from(agreementEntity, instantSource.instant())
-                ));
-                LOGGER.info("Agreement successfully set up with payment instrument",
-                        kv(AGREEMENT_EXTERNAL_ID, agreementEntity.getExternalId()),
-                        kv(PAYMENT_INSTRUMENT_EXTERNAL_ID, paymentInstrumentEntity.getExternalId()));
+                if (paymentInstrumentEntity.getRecurringAuthToken().isEmpty() || paymentInstrumentEntity.getRecurringAuthToken().get().isEmpty()) {
```
alexbishop1: `paymentInstrumentEntity.getRecurringAuthToken().ifPresentOrElse(…)` would work here.

alexbishop1 (changes requested):
Logic and tests look fine, just one suggestion on making the code a little neater.

src/main/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementService.java
```diff
@@ -45,29 +45,36 @@ public LinkPaymentInstrumentToAgreementService(PaymentInstrumentDao paymentInstr
...
-                paymentInstrumentEntity.setStatus(PaymentInstrumentStatus.ACTIVE);
-                ledgerService.postEvent(List.of(
-                        AgreementSetUp.from(agreementEntity, instantSource.instant()),
-                        PaymentInstrumentConfirmed.from(agreementEntity, instantSource.instant())
-                ));
-                LOGGER.info("Agreement successfully set up with payment instrument",
-                        kv(AGREEMENT_EXTERNAL_ID, agreementEntity.getExternalId()),
-                        kv(PAYMENT_INSTRUMENT_EXTERNAL_ID, paymentInstrumentEntity.getExternalId()));
+                paymentInstrumentEntity.getRecurringAuthToken()
+                        .filter(authTokenMap -> !authTokenMap.isEmpty())
```
alexbishop1: We could also do `.filter(not(Map::isEmpty))` here.

src/main/java/uk/gov/pay/connector/charge/service/LinkPaymentInstrumentToAgreementService.java
```diff
@@ -45,29 +45,36 @@ public LinkPaymentInstrumentToAgreementService(PaymentInstrumentDao paymentInstr
...
+                            paymentInstrumentEntity.setStatus(PaymentInstrumentStatus.ACTIVE);
+                            ledgerService.postEvent(List.of(
+                                    AgreementSetUp.from(agreementEntity, instantSource.instant()),
+                                    PaymentInstrumentConfirmed.from(agreementEntity, instantSource.instant())
+                            ));
+                            LOGGER.info("Agreement successfully set up with payment instrument",
+                                    kv(AGREEMENT_EXTERNAL_ID, agreementEntity.getExternalId()),
+                                    kv(PAYMENT_INSTRUMENT_EXTERNAL_ID, paymentInstrumentEntity.getExternalId()));
+                        },
+                        () -> LOGGER.info("Payment instrument doesn't have a token associated. Not linked the payment instrument to the agreement",
```
alexbishop1: Minor nit but it looks like this is inconsistent with what we do for the other `ifPresentOrElse(…)` calls in this method, where we have `}, () -> LOGGER.info(…)`.

It also looks like the whole block indented by an extra four spaces compared to the other ones. It would be a little more consistent if the whole method looked like this:

```java
    @Transactional
    public void linkPaymentInstrumentFromChargeToAgreement(ChargeEntity chargeEntity) {
        chargeEntity.getPaymentInstrument().ifPresentOrElse(paymentInstrumentEntity -> {
            chargeEntity.getAgreement().ifPresentOrElse(agreementEntity -> {
                paymentInstrumentEntity.getRecurringAuthToken().filter(not(Map::isEmpty)).ifPresentOrElse(_ -> {
                    cancelActivePaymentInstruments(agreementEntity);
                    agreementEntity.setPaymentInstrument(paymentInstrumentEntity);
                    paymentInstrumentEntity.setAgreementExternalId(agreementEntity.getExternalId());
                    paymentInstrumentEntity.setStatus(PaymentInstrumentStatus.ACTIVE);
                    ledgerService.postEvent(List.of(
                            AgreementSetUp.from(agreementEntity, instantSource.instant()),
                            PaymentInstrumentConfirmed.from(agreementEntity, instantSource.instant())
                    ));
                    LOGGER.info("Agreement successfully set up with payment instrument",
                            kv(AGREEMENT_EXTERNAL_ID, agreementEntity.getExternalId()),
                            kv(PAYMENT_INSTRUMENT_EXTERNAL_ID, paymentInstrumentEntity.getExternalId()));
                }, () -> LOGGER.info("Payment instrument doesn't have a token associated. Not linked the payment instrument to the agreement",
                        kv(AGREEMENT_EXTERNAL_ID, agreementEntity.getExternalId()),
                        kv(PAYMENT_INSTRUMENT_EXTERNAL_ID, paymentInstrumentEntity.getExternalId())));
            }, () -> LOGGER.error("Expected charge {} to have an agreement but it does not have one", chargeEntity.getExternalId()));
        }, () -> LOGGER.error("Expected charge {} to have a payment instrument but it does not have one", chargeEntity.getExternalId()));
    }
```

I know formatting like this isn’t usually worth sweating over but in this method there’s quite a lot of nesting going on so I think it’s worth doing what we can to make it as easy as possible to follow.

alexbishop1 (changes requested):
Looks good, just a couple of suggestions to make it a little easier to follow.

alexbishop1 (approved):
💯

PR #6565, opened 2026-07-29
PP-15645 Modify gateway request records type hierarchy

alexbishop1:
Modify the gateway request records type hierarchy to better fit our needs.

New type hierarchy:

```
                   AuthoriseRequest
                          |
                          v
CardAuthoriseRequest           ApplePayAuthoriseRequest
         |                                |
         v                                v
WorldpayCardAuthoriseRequest   AdyenApplePayAuthoriseRequest
         |
         v
WorldpayMotoAuthoriseRequest
```

Additionally, every Worldpay record implements `WorldpayAuthoriseRequest` and every Adyen record implements `AdyenAuthoriseRequest`, which extend from `WorldpayRequest` and `AdyenRequest` respectively.

All interfaces are sealed except for `WorldpayRequest` and `AdyenRequest` so it’s easy to require exhaustiveness when handling, say, all `AuthoriseRequest`s (but we do not yet have a case where it’s necessary to exhaustively handle all `WorldpayRequest`s or `AdyenRequest`s).

Rename factories etc. to reflect the new hierarchy.

`AuthorisationLogger` currently takes a `CardAuthoriseRequest`. It should really take an `AuthoriseRequest` but we do not yet handle `AdyenApplePayAuthoriseRequest` in `AuthorisationLogger` — when we do we can change it from taking a `CardAuthoriseRequest` to an `AuthoriseRequest`.


GabrielBoardman-MadeTech (approved):

PR #6575, opened 2026-07-30
PP-15645 Extract JSON object mapper from AdyenAuthorisationRequest

alexbishop1:
Extract JSON object mapper from `AdyenAuthorisationRequest` and make `AdyenAuthorisationRequest` hold a `GatewayOrder` rather than an `AuthoriseRequestPayload` (because we want to support types other than `AuthoriseRequestPayload` such as `AdyenApplePayAuthoriseRequest`).

Introduce `AuthoriseRequestPayloadToGatewayOrderConverter` to convert an `AuthoriseRequestPayload` to a `GatewayOrder` (including the JSON object mapping).

Since `AdyenAuthorisationRequest` is now a pure data class with no JSON object mapping functionality itself, its tests have been moved to `AuthoriseRequestPayloadToGatewayOrderConverter`.


GabrielBoardman-MadeTech (approved):

PR #6578, opened 2026-07-31
PP-15645 Convert GatewayOrder to record

alexbishop1:
Override `toString()` to not include the payload because it may contain card numbers or other sensitive information etc.


GabrielBoardman-MadeTech (approved):

PR #6583, opened 2026-07-31
PP-15645 Make WalletAuthoriseService create ApplePayAuthoriseRequest …

hjvoid:
Make WalletAuthoriseService create ApplePayAuthoriseRequest where supported.
When we can create an apple authorisation request, fallback on other route if not an Adyen payment.


GabrielBoardman-MadeTech (approved):

PR #6586, opened 2026-08-03
PP-15644 Move classes into different packages

alexbishop1:
Move some classes, mostly relating to sending authorisation requests to gateways, into different packages to better organise them.

The records in `uk.gov.pay.connector.gateway.model.request.records` implement sealed interfaces, which means they all need to stay in the same package because permitted classes must be in the same package in the unnamed module or in the same named module.


GabrielBoardman-MadeTech (approved):

PR #6591, opened 2026-08-04
PP-15511 Update error response when credential does not exist

ralphm10:
## WHAT YOU DID

- Update error response to 400 when credential does not exist for Adyen account type


DomBelcher (approved):
LGTM

PR #6592, opened 2026-08-04
PP-15647 - Added AdyenAuthoriseRequestLogGenerator So that we can use…

GabrielBoardman-MadeTech:
What we did:
- Added `AdyenAuthoriseRequestLogGenerator` So that we can use `AdyenApplePayAuthoriseRequest`
- Gave `AuthorisationLogger` the capability to work with `AdyenAuthoriseRequest`
- Added and updated tests where necessary


alexbishop1 (approved):

PR #6595, opened 2026-08-04
PP-15648 Replace ternary operators with switch expressions

alexbishop1:
When determining when a metric label should be `"success"` or `"failure"`, use a switch expression rather than a ternary operator.

When determining whether a metric label should be `"apple-pay"` or `"google-pay"`, use a switch expression rather than a ternary operator, which would have the incorrect behaviour if a third wallet type is added.


marcotranchino (dismissed):

GabrielBoardman-MadeTech (approved):

PR #6597, opened 2026-08-05
Pp 15688 alter frontend url removing trailing slash from tests

hjvoid:
## WHAT YOU DID
- Removed trailing slash from frontend url for all tests


alexbishop1 (approved):

PR #6598, opened 2026-08-05
PP-15682: Add charge_external_id column to payment_instruments table

Pippa101:
## WHAT YOU DID

Add `charge_external_id` column to `payment_instruments table` with field type varchar(26) and nullable

### DIFF
--- a/src/main/resources/migrations.xml
+++ b/src/main/resources/migrations.xml
@@ -2149,4 +2149,16 @@
         <addNotNullConstraint tableName="gateway_accounts_adyen_setup" columnName="gateway_account_credential_id"/>
     </changeSet>
 
+    <changeSet id="add charge_external_id column to payment_instruments table" author="">
+        <addColumn tableName="payment_instruments">
+            <column name="charge_external_id" type="varchar(26)">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="create index for charge_external_id on payment_instruments" runInTransaction="false" author="">
+        <sql>CREATE INDEX CONCURRENTLY idx_payment_instruments_charge_external_id ON payment_instruments (charge_external_id);</sql>
+    </changeSet>
+
 </databaseChangeLog>


alexbishop1 (dismissed):

src/main/resources/migrations.xml
```diff
@@ -2149,4 +2149,19 @@
...
+        <addColumn tableName="payment_instruments">
+            <column name="charge_external_id" type="varchar(26)">
+                <constraints
+                        foreignKeyName="fk__payment_instruments_charges"
+                        referencedTableName="charges"
+                        referencedColumnNames="external_id"
+                        nullable="true"/>
+            </column>
+        </addColumn>
+        <createIndex tableName="payment_instruments" indexName="idx_payment_instruments_charge_external_id">
```
kbottla: Will this create the index concurrently? Otherwise, the table will be locked while the index is being created and may cause some errors.

kbottla: Could add another changeset similar to https://github.com/alphagov/pay-connector/blob/088a749110b0da2932946b9bf958cfa6ce87a406/src/main/resources/migrations.xml#L2083 to be on the safe side

Pippa101: Will do that now

Pippa101: updated

src/main/resources/migrations.xml
```diff
@@ -2081,6 +2081,7 @@
             CREATE INDEX CONCURRENTLY idx_one_off_merchant_code ON gateway_account_credentials (one_off_merchant_code);
             CREATE INDEX CONCURRENTLY idx_recurring_cit_merchant_code ON gateway_account_credentials (recurring_cit_merchant_code);
             CREATE INDEX CONCURRENTLY idx_recurring_mit_merchant_code ON gateway_account_credentials (recurring_mit_merchant_code);
+            CREATE INDEX CONCURRENTLY idx_payment_instruments_charge_external_id ON payment_instruments (charge_external_id);
```
alexbishop1: I think this wants to be in its own changeset, after the `add charge_external_id column to payment_instruments table` one.

alexbishop1 (changes requested):

src/main/resources/migrations.xml
Pippa101: Thanks Alex, I've updated

src/main/resources/migrations.xml
```diff
@@ -2149,4 +2149,23 @@
...
+        <addColumn tableName="payment_instruments">
+            <column name="charge_external_id" type="varchar(26)">
+                <constraints
+                        foreignKeyName="fk__payment_instruments_charges"
+                        referencedTableName="charges"
+                        referencedColumnNames="external_id"
+                        nullable="true"/>
+            </column>
+        </addColumn>
+        <createIndex tableName="payment_instruments" indexName="idx_payment_instruments_charge_external_id">
```
alexbishop1: We don’t want this `<createIndex>` if we have the `CREATE INDEX CONCURRENTLY` later.

alexbishop1 (changes requested):

src/main/resources/migrations.xml
Pippa101: okay to clarify it should look like:

```
    <changeSet id="add charge_external_id column to payment_instruments table" author="">
        <addColumn tableName="payment_instruments">
            <column name="charge_external_id" type="varchar(26)">
                <constraints
                        foreignKeyName="fk__payment_instruments_charges"
                        referencedTableName="charges"
                        referencedColumnNames="external_id"
                        nullable="true"/>
            </column>
        </addColumn>
    </changeSet>

    <changeSet id="create index for charge_external_id on payment_instruments" author="">
        <sql>CREATE INDEX CONCURRENTLY idx_payment_instruments_charge_external_id ON payment_instruments (charge_external_id);</sql>
    </changeSet>
```

Do I need to add another `<changeSet>` if I delete the first `createIndex`?

alexbishop1 (dismissed):

src/main/resources/migrations.xml
alexbishop1: Honestly not sure. Probably fine to have two.

src/main/resources/migrations.xml
```diff
@@ -2149,4 +2149,20 @@
...
+                <constraints
+                        foreignKeyName="fk__payment_instruments_charges"
+                        referencedTableName="charges"
+                        referencedColumnNames="external_id"
+                        nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="create index for charge_external_id on payment_instruments" author="">
```
kbottla:
```suggestion
    <changeSet id="create index for charge_external_id on payment_instruments" runInTransaction="false" author="">
```

src/main/resources/migrations.xml
```diff
@@ -2149,4 +2149,20 @@
         <addNotNullConstraint tableName="gateway_accounts_adyen_setup" columnName="gateway_account_credential_id"/>
     </changeSet>
 
+    <changeSet id="add charge_external_id column to payment_instruments table" author="">
+        <addColumn tableName="payment_instruments">
+            <column name="charge_external_id" type="varchar(26)">
+                <constraints
+                        foreignKeyName="fk__payment_instruments_charges"
```
kbottla: Sorry, overlooked this the first time. Let's not create a foreign key constraint, as charges will be expunged once they reach the terminal state, but payment instruments will remain in Connector.

alexbishop1 (approved):

PR #6600, opened 2026-08-06
PP-15644 Rename gateway request records to end in “Payload”

alexbishop1:
Rename gateway request records — specifically `AdyenApplePayAuthoriseRequest` and `WorldpayMotoAuthoriseRequest` — to instead end in `Payload` — that is: `AdyenApplePayAuthorisePayload` and `WorldpayMotoAuthorisePayload` — in order to be more consistent with other Adyen records that represent payloads sent to Adyen and to make them more distinguishable from interfaces like `AdyenAuthoriseRequest` or `WorldpayCardAuthoriseRequest` that due to the limitations of sealed classes have to be in the same package.

We’re using `AdyenApplePayAuthorisePayload` or similar rather than `AdyenApplePayAuthoriseRequestPayload` because the latter is rather long.

New convention:
- Interfaces implemented by gateway request records end in `Request`
- Gateway request records that implement the aforementioned interfaces end in `Payload`


GabrielBoardman-MadeTech (approved):

PR #6611, opened 2026-08-10
PP-15681: Update shopperReference for recurring Adyen payments to be a unique combination of agreementExternalId and chargeExternalId

Pippa101:
## WHAT YOU DID
- created a method in AdyenRequestFactory which is only used for recurring payments to build the new shopper reference with a structure of `agreementExternalId-chargeExternalId`
  - When setting up a recurring payment, the charge_external_id of the payment being authorised is appended to the `shopperReference` for Adyen.
  - When taking a recurring payment, the charge_external_id on the payment instrument is appended to the `shopperReference` for Adyen.

## How to test

I have updated tests for recurring payments with Adyen to show the new change.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
@@ -101,7 +101,7 @@ public class AdyenRequestFactory {
         var recurringAuthToken = paymentInstrument.getRecurringAuthToken()
                 .orElseThrow(() -> new IllegalArgumentException("Payment instrument does not have recurring auth token set"));
 
-        String shopperReference = request.getAgreementId();
+        String shopperReference = createShopperReferenceForRecurringPayments(request);
         String storedPaymentMethodId = recurringAuthToken.get(STORED_PAYMENT_METHOD_ID);
 
         if (shopperReference == null || storedPaymentMethodId.isBlank()) {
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/connector/gateway/adyen/AdyenRequestFactory.java
```diff
@@ -129,6 +129,16 @@ public AuthoriseRequestPayload createRecurringPaymentRequest(RecurringPaymentAut
         );
     }
 
+    private String createShopperReferenceForRecurringPayments(RecurringPaymentAuthorisationGatewayRequest request) {
+        var chargeExternalId = request.getPaymentInstrument().isPresent()
+                ? request.getPaymentInstrument().get().getChargeExternalId()
+                : request.getGovUkPayPaymentId();
+        if (chargeExternalId == null || request.getAgreementId() == null) {
+            return null;
```
kbottla, line 137: Since both should be available at all times, can an exception be thrown if either is null instead of returning null?

Pippa101: An `IllegalArgumentException` is thrown on line 108 if shopperReference is null, I can refactor it so that it's thrown in `createShopperReferenceForRecurringPayments()` instead?

kbottla: If it is already handled, I don't mind refactoring it or leaving it as is on line 108.

kbottla (approved):
LGTM!

PR #6613, opened 2026-08-11
PP-15557 add recurring token webhooks to task queue

AbdirizakIdris:
## WHAT YOU DID
- Added support for event types: 
  -` recurring.token.created` and `recurring.token.disabled`
- ignore unsupported token webhook event types and log the notification details
- prevent notifications with an invalid HMAC signature from being added to the task queu


Pippa101 (approved):
LGTM

PR #6616, opened 2026-08-11
PP-15635 Enable creation of Adyen switching credentials

ralphm10:
## WHAT YOU DID

- Update `GatewayAccountCredentialsRequestValidator.java` to accept Adyen gateway accounts
- Update `GatewayAccountCredentialsService.java` to throw an error for Adyen test accounts

These changes allow for the creation of a switching credential for Adyen live accounts only, when calling 
POST `/v1/api/accounts/{accountId}/credentials` or POST `/v1/api/service/{serviceId}/account/{accountType}/credentials`

 We already have functionaly to create Adyen test accounts. This is done via POST `/v1/api/service/{serviceId}/request-adyen-test-account` for new accounts, or POST `/v1/api/service/{serviceId}/switch-to-adyen-test-account` for existing Stripe accounts. 

Calls to either of these endpoints will fail if an Adyen credential exists in any state, so while unlikely to happen, we should guard against this. Additionally, when creating Adyen test accounts using the existing endpoints, we receive credentials from Adyen, so we should prevent test accounts being created with an empty credential object.

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
+++ b/src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
@@ -8,7 +8,9 @@ import jakarta.ws.rs.BadRequestException;
 import jakarta.ws.rs.WebApplicationException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.pay.connector.charge.exception.ConflictWebApplicationException;
 import uk.gov.pay.connector.charge.model.domain.Charge;
+import uk.gov.pay.connector.common.exception.ConflictRuntimeException;
 import uk.gov.pay.connector.gateway.PaymentGatewayName;
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountCredentialsNotFoundException;
 import uk.gov.pay.connector.gatewayaccount.exception.GatewayAccountNotFoundException;
@@ -104,8 +106,8 @@ public class GatewayAccountCredentialsService {
         boolean credentialsPrePopulated = !credentials.isEmpty();
         var gatewayAccountType = GatewayAccountType.fromString(gatewayAccountEntity.getType());
         
-        if (paymentGatewayName == ADYEN && gatewayAccountType == TEST && !credentialsPrePopulated) {
-            throw new BadRequestException("Switching credential cannot be created for an Adyen test account.");
+        if (paymentGatewayName == ADYEN && gatewayAccountType == TEST && credentials.isEmpty()) {
+            throw new ConflictWebApplicationException("Cannot create empty Adyen credentials on a test account.");
         }
 
         if (paymentGatewayName == SANDBOX ||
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -103,6 +103,10 @@ private GatewayAccountCredentialState calculateStateForNewCredentials(GatewayAcc
         boolean isFirstCredentials = !gatewayAccountCredentialsDao.hasActiveCredentials(gatewayAccountEntity.getId());
         boolean credentialsPrePopulated = !credentials.isEmpty();
         var gatewayAccountType = GatewayAccountType.fromString(gatewayAccountEntity.getType());
+        
+        if (paymentGatewayName == ADYEN && gatewayAccountType == TEST && !credentialsPrePopulated) {
+            throw new BadRequestException("Switching credential cannot be created for an Adyen test account.");
```
DomBelcher: `409 Conflict` may be a more accurate response, but it's not overly important

DomBelcher: Also I think this prevents creating an Adyen credential for any test account, rather than creating a new credential on an Adyen test account. This is fine because we assume test accounts never switch PSP, but the error message should reflect this, e.g. `Cannot create new Adyen credentials on test account`. 

I guess technically it's new, empty Adyen which credentials can't be added to a test account - we use these for provider switching, and may colloquially refer to them as "switching credentials" but perhaps it's more accurate to say `Cannot create empty Adyen credentials...`.

src/main/java/uk/gov/pay/connector/gatewayaccountcredentials/service/GatewayAccountCredentialsService.java
```diff
@@ -103,6 +103,10 @@ private GatewayAccountCredentialState calculateStateForNewCredentials(GatewayAcc
         boolean isFirstCredentials = !gatewayAccountCredentialsDao.hasActiveCredentials(gatewayAccountEntity.getId());
         boolean credentialsPrePopulated = !credentials.isEmpty();
         var gatewayAccountType = GatewayAccountType.fromString(gatewayAccountEntity.getType());
+        
+        if (paymentGatewayName == ADYEN && gatewayAccountType == TEST && !credentialsPrePopulated) {
```
DomBelcher: `!credentialsPrePopulated` is indirectly a double negation, so maybe `... && credentials.isEmpty()` would be clearer, even if it requires calling that method a second time - although I assume it's not an expensive method call.

src/test/java/uk/gov/pay/connector/it/resources/GatewayAccountResourceCreateIT.java
DomBelcher: Maybe this test name should be updated to something like `should_create_live_Adyen_...`

DomBelcher (approved):
Nice work

PR #6619, opened 2026-08-11
PP-15725 Extract BrowserInfo creation into factory

joshuapook-MT:
- Added BrowserDataFor3ds interface
- Added AdyenBrowserInfoFactory and implemented create method with defaults
- Updated BrowserInfo to AdyenBrowserInfo
- Updated AdyenRequestFactory to use new AdyenBrowserInfoFactory for creating browser info


alexbishop1 (approved):

PR #6622, opened 2026-08-13
PP-15715 Implement BrowserDataFor3ds for GooglePayPaymentInfo

alexbishop1:
Implement the `BrowserDataFor3ds` interface for`GooglePayPaymentInfo`. This adds support for deserialising additional browser information that’s used by 3D Secure when sent from frontend (which will collect it from the paying user’s browser). Continue to support this data not being present because it is not currently sent and even when it is, there’s no guarantees it will be sent (because, for example, JavaScript could be disabled or the browser could give us something completely bonkers that fails frontend’s validation).


hjvoid (approved):

PR #6623, opened 2026-08-13
PP-15715 Rationalise duplicate getters in GooglePayPaymentInfo

alexbishop1:
Remove getters in `GooglePayPaymentInfo` that return the same fields as the getters specified in the `BrowserDataFor3ds` interface. This requires us to add `@JsonProperty` annotations to some fields because the getter names no longer match the JSON property names.


hjvoid (dismissed):

hjvoid (dismissed):

hjvoid (approved):
LGTM

PR #6625, opened 2026-08-14
PP-15716 - Create AdyenGooglePayAuthRequest record

GabrielBoardman-MadeTech:
- Created a new record with only the parameters we wish to pass from frontend
- Added a test for the record


alexbishop1 (approved):

PR #6634, opened 2026-08-17
PP-15717 - Add new route that uses AdyenGooglePayAuthRequest

GabrielBoardman-MadeTech:
-Added a new route to CardResource
-Added tests for WalletService
-Updated the openApi spec


hjvoid (approved):

PR #6638, opened 2026-08-19
PP-15717 - Update OpenAPI docs

GabrielBoardman-MadeTech:
-Update field annotations so that connector_spec.yaml is auto generated correctly


hjvoid (approved):

PR #6655, opened 2026-08-25
PP-15740 Update OpenAPI spec for Adyen Notification endpoint with examples

joshuapook-MT:
- Added examples to the Adyen notification endpoint definition

### DIFF
--- a/src/main/java/uk/gov/pay/connector/webhook/resource/NotificationResource.java
+++ b/src/main/java/uk/gov/pay/connector/webhook/resource/NotificationResource.java
@@ -1,7 +1,11 @@
 package uk.gov.pay.connector.webhook.resource;
 
+import com.adyen.model.notification.NotificationRequest;
 import io.swagger.v3.oas.annotations.Operation;
 import io.swagger.v3.oas.annotations.Parameter;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.parameters.RequestBody;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/connector/webhook/resource/NotificationResource.java
```diff
@@ -147,6 +154,64 @@ public Response authoriseStripeNotifications(String notification,
                     @ApiResponse(responseCode = "415", description = "Unsupported Media Type - Unsupported content type")
             }
     )
+    @RequestBody(
+            description = "Adyen notification payload",
+            required = true,
+            content = @Content(
+                    mediaType = APPLICATION_JSON,
+                    schema = @Schema(oneOf = {NotificationRequest.class, AdyenTokenNotification.class}),
```
Pippa101: You don't need the examples if you're using schema like you are. So you can delete the examples list, this is what the open api spec looks like if you don't have examples and only have `schema = @Schema(oneOf = {NotificationRequest.class, AdyenTokenNotification.class})`

Image

Image

However if you want to keep the examples that's fine with me also :)

joshuapook-MT: If I remove the examples, maven just shows this:
Screenshot 2026-08-25 at 12 51 21

joshuapook-MT: Alright, explicit examples removed

Pippa101 (approved):
LGTM

PR #6656, opened 2026-08-25
PP-15683 Refactor AdyenTokenWebhookNotificationHandler to remove unnecessary calls to DB

joshuapook-MT:
In AdyenTokenWebhookNotificationHandler:
- Added condition to check if the storedPaymentMethodId is already set as the recurring auth token before setting it
- Moved call to get webhook payment instrument until it's definitely needed


kbottla (commented):
LGTM!

kbottla (approved):

PR #6672, opened 2026-09-02
PP-15837 Consistent naming for payload gateway request records

alexbishop1:
In a previous commit, we renamed concrete gateway request record types (as opposed to interfaces) so their names ended with `Payload` rather than `Request` (leaving `Request` to indicate an interface implemented by one or more of the `Payload` records). Finish this up by renaming the factories and fixtures to match the `Payload`/`Request` naming consistently.


GabrielBoardman-MadeTech (approved):

PR #6691, opened 2026-09-07
PP-15856 Revert Adyen specific changes when creating payment instruments

kbottla:
## WHAT YOU DID
- Reverted Adyen specific changes introduced to create payment instruments with an empty token and disabled relevant tests. Adyen-specific logic can be in payment provider code and will be addressed separately.
- Added a missing Stripe test setting up an agreement that requires 3DS. This is a real-world scenario


Pippa101 (approved):
LGTM

PR #6696, opened 2026-09-08
PP-15789: add method to delegate HMAC validation based on Adyen notification structure

Pippa101:
## WHAT YOU DID

- Added a method to delegate HMAC valdiation to existing validator methods based on the Adyen notification's field `usesAdyenNotificationItem`. This is so the HMAC validator has 1 generic point of entry to the class and logic is now seperated from the service (this will be implemented in the next PR). 

### Logging

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationValidator.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationValidator.java
@@ -6,6 +6,7 @@ import jakarta.inject.Inject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.connector.app.adyen.AdyenGatewayConfig;
+import uk.gov.pay.connector.gateway.adyen.webhook.model.AdyenWebhookNotification;
 import uk.gov.pay.connector.gateway.exception.AdyenNotificationException;
 import uk.gov.pay.connector.util.IpDomainMatcher;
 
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationValidator.java
```diff
@@ -23,12 +25,41 @@ public class AdyenNotificationValidator {
...
+
+    public boolean validateHmacSignature(AdyenWebhookNotification adyenWebhookNotification,
+                                         String payload,
+                                         String hmacSignature) {
+        String hmacKey = getHmacKeyForWebhookType(adyenGatewayConfig,
+                adyenWebhookNotification.event().getWebhookType(),
+                adyenWebhookNotification.isLive());
+
+        if (adyenWebhookNotification.usesAdyenNotificationItem()) {
+            NotificationRequestItem notificationRequestItem = adyenWebhookDeserialiser.deserialiseAndGetNotificationItem(payload);
```
joshuapook-MT, line 50: It would be better to make an override to pass this in instead, I think? We ideally don't want to be deserialising the payload twice. 

At that point, it might be this method isn't needed though. 

Probably worth having a chat in the morning as I might be missing at what point this will be called :)

joshuapook-MT: Spoke to Pippa and understand why the deserialisation is needed here.

joshuapook-MT (approved):

PR #6698, opened 2026-09-09
PP-15789: Refactor AdyenNotificationService to delegate deserialisation and webhook type sorting logic to helper classes

Pippa101:
## WHAT YOU DID
- refactored `AdyenNotificationService` to delegate deserialisation and webhook type sorting logic to helper classes made in previous PRs as part of refactoring adyen webhooks epic. 
https://payments-platform.atlassian.net/browse/PP-15738
- updated tests and webhook handlers that were previously using the deserialisation method in `AdyenNotificationService` to now use `AdyenWebhookDeserialiser` class.

### Logging

### DIFF
--- a/src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
+++ b/src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
@@ -9,32 +9,20 @@ import com.adyen.model.notification.NotificationRequest;
 import com.adyen.model.notification.NotificationRequestItem;
 import com.adyen.notification.WebhookHandler;
 import com.adyen.util.HMACValidator;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import jakarta.ws.rs.WebApplicationException;
-import org.junit.Assert;
 import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.EnumSource;
-import org.junit.jupiter.params.provider.NullAndEmptySource;
-import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.ArgumentCaptor;
 import org.mockito.Captor;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 import org.slf4j.LoggerFactory;
-import uk.gov.pay.connector.app.adyen.AdyenGatewayConfig;
-import uk.gov.pay.connector.app.adyen.HmacKeys;
-import uk.gov.pay.connector.app.adyen.HmacKeys.WebhookHmacKeyPair;
-import uk.gov.pay.connector.app.adyen.WebhookHmacKeys;
-import uk.gov.pay.connector.gateway.adyen.response.AdyenTokenNotification;
+import uk.gov.pay.connector.gateway.adyen.webhook.model.AdyenWebhookEvent;
+import uk.gov.pay.connector.gateway.adyen.webhook.model.AdyenWebhookNotification;
 import uk.gov.pay.connector.gateway.exception.AdyenNotificationException;
 import uk.gov.pay.connector.queue.tasks.TaskQueueService;
-import uk.gov.pay.connector.queue.tasks.TaskType;
 import uk.gov.pay.connector.queue.tasks.model.Task;
-import uk.gov.pay.connector.util.JsonObjectMapper;
 import uk.gov.pay.connector.util.TestTemplateResourceLoader;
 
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
```diff
@@ -135,123 +127,54 @@ void shouldRejectNotificationWhenExceptionIsThrown() {
...
-        void shouldThrowWebApplicationExceptionWhenSendingPaymentNotificationToTaskQueueFails(AdyenPaymentEvent eventCode) {
-            when(mockAdyenNotificationValidator.isValidIpAddress("5.6.7.8")).thenReturn(true);
-            when(mockAdyenGatewayConfig.getHmacKeys()).thenReturn(getHmacKeys());
-            when(mockAdyenNotificationValidator.isValidHmac(any(), any())).thenReturn(true);
+        assertFalse(result);
+        verifyNoInteractions(mockTaskQueueService);
+    }
 
+    @ParameterizedTest
+    @EnumSource(AdyenWebhookEvent.class)
```
kbottla: We don't need a parameterised test, as the test checks the method's behaviour when adding a task to the SQS queue fails. We are not testing AdyenWebhookEvent specific behaviour

src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
```diff
@@ -270,16 +193,16 @@ void shouldThrowWebApplicationExceptionWhenSendingPaymentNotificationToTaskQueue
...
-        @EnumSource(AdyenPaymentEvent.class)
-        void shouldAddValidPaymentNotificationToTaskQueue(AdyenPaymentEvent eventCode) {
-            when(mockAdyenNotificationValidator.isValidIpAddress("5.6.7.8")).thenReturn(true);
-            when(mockAdyenGatewayConfig.getHmacKeys()).thenReturn(getHmacKeys());
-            when(mockAdyenNotificationValidator.isValidHmac(any(), any())).thenReturn(true);
-
-
+    @ParameterizedTest
+    @EnumSource(AdyenWebhookEvent.class)
+    void shouldAddValidNotificationToTaskQueue(AdyenWebhookEvent eventCode) {
```
kbottla: This test can be split into

shouldAddValidNotificationToTaskQueue - for a valid unignored event
shouldNotAddIgnoredEventToTaskQueue

src/test/java/uk/gov/pay/connector/gateway/adyen/webhook/AdyenNotificationServiceTest.java
```diff
@@ -270,16 +193,16 @@ void shouldThrowWebApplicationExceptionWhenSendingPaymentNotificationToTaskQueue
...
+        if (!eventCode.isIgnored()) {
             String payload = getNotificationWithValidHmacSignature(eventCode.toString());
```
kbottla: This could cause confusion, as the ADYEN_NOTIFICATION template is loaded for each eventCode, but the structure is different for some events.

src/test/java/uk/gov/pay/connector/it/resources/adyen/AdyenNotificationResourceIT.java
```diff
@@ -83,7 +83,7 @@ void shouldReturn500WhenInvalidJsonPayload() {
                 .contentType(APPLICATION_JSON)
                 .post(NOTIFICATION_PATH)
                 .then()
-                .statusCode(500);
+                .statusCode(403);
```
kbottla: Any reason this is changed to 403? If the JSON is invalid, a 403 Forbidden error doesn't seem correct

Pippa101: From what I can find, in all Adyen IT tests we never expect to return a 500, when we do get a 500 we convert it to a Forbidden status [see this test](https://github.com/govuk-pay/pay-connector/blob/master/src/test/java/uk/gov/pay/connector/it/resources/adyen/AdyenCardResourceAuthoriseIT.java#L231). Because we're handling `UnparseableAdyenWebhookException` in `AdyenNotificationService` it falls into the  `return forbiddenErrorResponse()` in `NotificationResource` [here](https://github.com/govuk-pay/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/webhook/resource/NotificationResource.java#L170)

kbottla: Other IT tests are part of the payment journey, and for different status codes returned, the frontend interprets and presents a specific page to users.

For Webhooks, we return 
- 415 - for unsupported media or content type
- 403 - when verification fails (IP, HMAC or similar)
- 500 - For issues processing the valid payload (parsing, SQS errors, unknown webhooks, and so on)

kbottla (approved):
LGTM!

PR #6700, opened 2026-09-10
PP-15789: Delete unused adyen event enums and remove uses from Adyen webhook handlers

Pippa101:
## WHAT YOU DID

- deleted unused Adyen webhook Payments and Token event enums as they have now been combined into 1 enum `AdyenWebhookEvent`.
- removed implementations of unused enums and replaced with values from `AdyenWebhookEvent`


kbottla (approved):
LGTM!

PR #6702, opened 2026-09-11
PP-15856 Return an empty map for recurring_auth_token when storedPaymentMethodId is not present

kbottla:
## WHAT

- Returns an empty map for the recurring_auth_token when storedPaymentMethodId is not present after a recurring payment is authorised.
    - A payment instrument is created with state 'CREATED' when an empty map is returned by the payment provider
    - When storedPaymentMethodId is not available in the authorisation response, we expect it to be available in the `recurring.token.created` webhook processed.
- Enabled relevant integration tests that were disabled in https://github.com/govuk-pay/pay-connector/pull/6690

### DIFF
--- a/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthorise3dsHandler.java
+++ b/src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthorise3dsHandler.java
@@ -22,6 +22,7 @@ import static org.apache.commons.lang3.StringUtils.isBlank;
 import static uk.gov.pay.connector.gateway.adyen.utils.AdyenRequestUtil.get3dsAuthUrl;
 import static uk.gov.pay.connector.gateway.adyen.utils.AdyenRequestUtil.getHeaders;
 import static uk.gov.pay.connector.gateway.model.OrderRequestType.AUTHORISE_3DS;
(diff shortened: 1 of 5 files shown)


joshuapook-MT (approved):

src/main/java/uk/gov/pay/connector/gateway/adyen/handler/AdyenAuthorise3dsHandler.java
```diff
@@ -74,6 +75,10 @@ public Gateway3DSAuthorisationResponse authorise3dsResponse(Auth3dsResponseGatew
             Map<String, String> recurringAuthToken = storedPaymentMethodId == null ? null
                     : Map.of("storedPaymentMethodId", storedPaymentMethodId);
 
+            if (AUTHORISED.equals(mappedStatus) && recurringAuthToken == null) {
+                recurringAuthToken = Map.of();
+            }
```
alexbishop1, line 80: I wonder if this might be a bit clearer (I think the logic is equivalent):

```java
Map<String, String> recurringAuthToken = null;
if (storedPaymentMethodId != null) {
    recurringAuthToken = Map.of("storedPaymentMethodId", storedPaymentMethodId);
} else if (AUTHORISED.equals(mappedStatus)) {
    recurringAuthToken = Map.of();
}
```

PR #6703, opened 2026-09-11
[PP-15703] Add workflow to generate transitive dependency graph for dependabot

foxmaltas-gds:
## WHAT YOU DID
Adds a github actions workflow to keep the java transitive dependencies up to date

## How to test
This will be run on push/merge to the `master` branch. Once this has been merged, the workflow will run

### Logging

### Documentation

### DIFF
--- a/.github/workflows/generate-transitive-dependency-graph.yml
+++ b/.github/workflows/generate-transitive-dependency-graph.yml
@@ -1,0 +1,9 @@
+name: Generate java transitive dependency graph
+on:
+  push:
+    branches:
+      - "master"
+
+jobs:
+  generate-dependency-graph:
+    uses: govuk-pay/pay-ci/.github/workflows/_run-java-transitive-dependency-update.yml@master


.github/workflows/generate-transitive-dependancy-graph.yml
```diff
@@ -0,0 +1,8 @@
+name: Generate java transitive dependency graph
+on:
+  push:
+    branches: [ "master" ]
+
+jobs:
+  generate-dependency-graph:
+    uses: govuk-pay/pay-ci/.github/workflows/_run-java-transitive-dependancy-update.yml@master
```
benjamb: Oh, I've only just noticed the file has a typo!

.github/workflows/generate-transitive-dependancy-graph.yml
```diff
@@ -0,0 +1,8 @@
+name: Generate java transitive dependency graph
+on:
+  push:
+    branches: [ "master" ]
```
benjamb: It's slightly more idiomatic to do:

```suggestion
    branches:
      - master
```
but this is fine.

benjamb (changes requested):

.github/workflows/generate-transitive-dependancy-graph.yml
foxmaltas-gds: Typo has been corrected (both in the reusable workflow and in this file)

.github/workflows/generate-transitive-dependancy-graph.yml
foxmaltas-gds: Made this change

benjamb (approved):

PR #6715, opened 2026-09-15
PP-15855 Update Adyen test account service

kbottla:
## WHAT YOU DID
- Updated Adyen test service to fix signing the PCI questionnaire and add wallet payments to Stores 
- Added 
     - Config to link the SplitConfigurationId to Stores. The split config profile is created separately with rules for different cards (API docs https://docs.adyen.com/platforms/automatic-split-configuration/create-split-configuration#create-split-configuration)  
          - We only need a test ID, as Connector is responsible for creating Adyen test accounts only. 
     - Custom payout schedule for test accounts


Pippa101 (approved):
LGTM

PR #6721, opened 2026-09-17
PP-15735 Remove unused graphite metrics

felixmcd-mt:
## WHAT YOU DID

- Removed the unused metrics graphite dependency and associated configuration from Connector.


src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
```diff
@@ -102,9 +99,8 @@ public class ConnectorApp extends Application<ConnectorConfiguration> {
     private static final Logger logger = LoggerFactory.getLogger(ConnectorApp.class);
 
     public static final boolean NON_STRICT_VARIABLE_SUBSTITUTOR = false;
-
-    private static final String SERVICE_METRICS_NODE = "connector";
-    private static final int GRAPHITE_SENDING_PERIOD_SECONDS = 10;
+    
+    private static final int METRICS_SENDING_INTERVAL_SECONDS = 10;
```
AndrzejProchyra: Should this be 5 now, and we don't divide by 2 where we use it on line 230? 

Does "interval" or "period" make more sense here? `ScheduledExecutorService` calls it `period`. That makes more sense to me because this value represents an "unanchored" amount of time between two events that repeat at a frequency of 1/period, whereas an interval is the distance between two specific points in time, right?

AndrzejProchyra (commented):
Looks good! Just a small comment about the constant.

src/main/java/uk/gov/pay/connector/app/ConnectorApp.java
felixmcd-mt: Period does make more sense, and yes it’s cleaner to remove the division within scheduleAtFixedRate and use the period value directly

AndrzejProchyra (approved):

